PR #2182, opened 2014-03-20
Read only flag

joshmoore:
First steps toward a proper "read-only" (or "published") state. With this, the `EventContext` object properly records whether or not any write action (i.e. INSERT, UPDATE, DELETE) is possible. Further PRs can make use of this state to set the server into shutdown-mode.

There is little to no use currently of the `r-----` flag, so this branch as it stands should have no impact on existing tests or functionality. As the one integration test shows, though, any newly created group starting with `r-` will be locked, or published, in that no database INSERTs can take place.

As it stands, this branch may not be compatible with the `dev_5_0` branch because of the change in the model objections. If the general functional is considered to be correct, we can evaluate how best to achieve `5.0`-compatibility.


joshmoore:
Closing until this can be worked on further.

PR #2309, opened 2014-04-15
Add debugging of NPE in InteractiveProcessorI

joshmoore:
Intended for tracking down the problem on gretzky.
May not need to be merged.

cc @manics


joshmoore:
--rebased-to #3414

PR #2906, opened 2014-08-01
Server dialog

jburel:
Review the layout of the server dialog.
Problem noticed while running the client from the command line

To test:
- Start insight using the version matching your OS. e.g. mac
- Click on the wrench to display the server dialog. Check that the display is correct.
- Download the Linux client package and use the scripts from the command line to start insight
- Click on the wrench to display the server dialog. Check that the display is correct.


jburel:
--rebased-to #3375

PR #2953, opened 2014-08-18
Download spw original files

will-moore:
This adds the "Download" menu (and fileset paths) to SPW images when wells are selected.

To test:
- Import a Plate.
- Also create a plate from a Dataset with Dataset_To_Plate.py script (simpler if images are not MIFs).
- Select ONE well from each plate:
  - Test the "Show paths" button
  - Download the Original files for the well. This will be a single image for the created plate and the whole plate for the imported plate.
  - Export as png/jpeg/tiff. Check that you get the correct image for the current Plate 'index'.
- Select MULTIPLE wells from each plate:
  - Test paths, download original and export png/jpeg/tiff as described above.


sbesson:
--rebased-to #3351

PR #2954, opened 2014-08-18
min-max set

jburel:
Always set the windows start-end to global min/max.
This should speed up the generation of rendering settings
To test this PR:
- Import few images
- Check that for each channel the start-end is set to the min/max
- Select images imported previously (not the newly imported on)
- Click "reset defaults"
- Check that for each channel start-end=min, max


jburel:
--rebased-to #3365

PR #3033, opened 2014-09-15
Float re

jburel:
Add support for floating point image
To test this PR
- make sure that the unit tests under `server/test/omeis.providers.re.utests` are green
- Import floating point images. a selection can be found under `squig/team/will/float0-data.zip`
- View the image in OMERO, modify some values e.g. pixels window.
- View the same image in ImageJ. 

Note that the pyramid generation for floating point images (>4kx4k)will not work. This will not be addressed in this PR.


jburel:
Due to time, merge with 32-bit PR (conflicting PR)

--rebased-to #3366

PR #3086, opened 2014-10-01
Create image by tiling 12459

will-moore:
This adds tiling support to Blitz Gateway conn.createImageFromNumpySeq() allowing it to use numpy planes larger than 3k x 3k.
However, this is only true if the image has zero or one of dimensions Z / C / T > 1.
E.g. it can handle a multi-channel image OR a z-stack OR timelapse, but not any combination of these.

To test:
 ```
$ cd OmeroPy
$ ./setup.py test -s test/integration/gatewaytest/test_create_image.py
```

Also try creating a "BIG" ROI (over 3k x 3k) on a big image (multichannel, single Z, single C) and run the ```Images from ROIs.py``` script.
This should create a new Big tiled image.
NB: numpy arrays for each of the 3 planes will be handled as a individual 2D arrays, so this is a limitation on very big ROIs.


joshmoore:
12459 has been unscheduled. Closing this for now.

PR #3141, opened 2014-10-27
jstree upgrade (Functionality Review Only)

dpwrussell:
This is a huge PR addressing the update of jstree which had far reaching consequences. Because much of the code needed to be rewritten, I have also fixed a number of other bugs in the webclient at the same time and switched to a fully JSON REST API powered tree. Obviously it would be preferable to have done that in separate PRs, but that would have required coding a lot of broken behaviour from scratch only to then fix it.

As it's such a large PR, I am looking to evaluate the functionality first, fix any obvious problems, and then I will refactor this PR into logical chunks for what will probably be a hefty code/full functionality review.

Trello Link: https://trello.com/c/tly4JAme/63-jquery-and-jstree-3

Testing instructions
--------------------------
Basically just compare existing webclient to this.

There are differences, most of which are upgrades (e.g. child numbers now update when copying/moving objects) or bug fixes (e.g. tree now supports having the same object in multiple places without unpredictable behaviour), but if there are mistakes, it would be good to find them now.

I am aware that webtagging does not work yet, this is because it was relying on webclient internals (unavoidable in current plugins realistically and also unavoidable to break it here) and I have yet to update the public version.

Known Issues
-------------------
Permissions have not yet been updated with the hack required for http://trac.openmicroscopy.org.uk/ome/ticket/9800

The All-Users display is not consistent with current webclient. However, it is much more consistent with Insight which might be more desirable.


joshmoore:
Contacted @dpwrussell. This work is being passed off to @will-moore. They decided that a reworking/cleanup of the branch won't necessarily make the next steps any easier, and so there's nothing else coming here. Closing.

NB: I've added the `LATER` flag so it's easier to find it when the time comes.

PR #3157, opened 2014-10-30
Reset password

jburel:
Do not allow to change LDAP password

To test 
- Use the first user listed under https://www.openmicroscopy.org/site/team/Password
- Check that the LDAP info are displayed and you cannot change the password
- Log out
- log in as an admin
- Go to the Admin tab
- Expand Test Group and select the user used previously
- Check that the LDAP info are displayed. Check that you can neither modify the password  nor the login name.


components/insight/SRC/org/openmicroscopy/shoola/env/data/AdminServiceImpl.java
```diff
@@ -872,4 +877,14 @@ public boolean isSystemUser(long userID, String key)
...
     }
+
+    /**
+     * Implemented as specified by {@link AdminService}.
+     * @see AdminService#lookupLdapAuthExperimenter(SecurityContext, long)
+     */
+    public String lookupLdapAuthExperimenter(SecurityContext ctx, long userID)
+            throws DSOutOfServiceException, DSAccessException
+    {
+        return gateway.lookupLdapAuthExperimenter(ctx, userID);
```
joshmoore: So is this method now unused?

jburel: No we have to use it. The boolean flag on user only tells us if LDAP or not but in order to display the details we need the call.
Other option will be to replace the boolean flag by the details i.e. string in DB.

joshmoore: Ah, understood. Storing it in the DB would get us back into the de-synchronization trouble we had previously. So barring a larger reworking (like **all** users are in LDAP), this will have to stay.

PR #3181, opened 2014-11-11
Update to Hibernate 3.6 (See trac2570)

joshmoore:
As an attempted fix for https://trac.openmicroscopy.org.uk/ome/ticket/12601,
I tried upgrading to Hibernate 3.6. This didn't fix the issue caused by enums
(which was eventually solved in https://github.com/openmicroscopy/openmicroscopy/pull/3146
by moving to proper enums), but having an updated version of Hibernate is
useful itself.

No particular testing is necessary, though this should likely stay in the queue
for sometime. To initially avoid any problems with corrupt Lucene indexes, I'm
putting this into breaking first. Once this code has run against a server, to rollback
requires deleting the `/OMERO/FullText` directory.

 * [ ] fix search
 * [ ] remove model warnings


mtbc:
I've really no idea why this is being problematic. I am curious as to why a `HibernateTransactionManager` won't work instead, perhaps some bad interaction with our custom `SessionHandler`, but even if it did work it may be of no use, it's just that I've already managed to refute the other hypotheses I've generated about what's going wrong here. Given the apparent lack of others with this problem in upgrading Hibernate and Hibernate Search, perhaps the culprit is in something custom and weird, unless there's simply some bit of misconfiguration that didn't previously cause problems.

joshmoore:
Closing until this can be worked on further.

PR #3208, opened 2014-11-20
Update the OMERO install script to test Web login with CSRF (rebased onto develop)

sbesson:
This is the same as gh-3185 but rebased onto develop.

----

This commit patches OMERO-homebrew-install.sh to retrieve the CSRF token and
then pass it to the POST data in a second call testing the login.

To test this PR, check the http://ci.openmicroscopy.org/job/OME-5.0-merge-homebrew/ passes while testing the development Web server.

/cc @aleksandra-tarkowska

### DIFF
--- a/docs/hudson/OMERO-homebrew-install.sh
+++ b/docs/hudson/OMERO-homebrew-install.sh
@@ -151,8 +151,10 @@ bin/omero web start
 
 # Test simple Web connection
 brew install wget
-post_data="username=root&password=$ROOT_PASSWORD&server=1&noredirect=1"
-resp=$(wget --post-data $post_data http://localhost:$HTTPPORT/webclient/login/)
+wget --keep-session-cookies --save-cookies cookies.txt http://localhost:$HTTPPORT/webclient/login/ -O csrf_index.html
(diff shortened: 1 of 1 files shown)


sbesson:
You anticipated my question :) For now 5.1 versions will be suffixed with `51`. The ideal scenario would be to have support for aliases in Homebrew taps and then flip the symlink. Last time I tried this wasn't supported. I can see it both ways:
- either `omero` -> `omero50` and `omero51` -> `omero`
- or `omero` -> `omero50` and we use versioned formulas

joshmoore:
Looking at https://github.com/sbesson/homebrew-alt/tree/5.1.0-m3/Formula, having each version present would certainly seem to make sense. I wonder if a copy of the stable version to the version-less name (e.g. `omero.rb`) would suffice and not be too burdensome and/or confusing.

sbesson:
So `omero50`, `omero51` and make a hard-copy of whatever is stable? That would work for me.

joshmoore:
It's slightly wonky, but I think the cleanest.

sbesson:
Sounds good. I will open a request once this PR and https://github.com/ome/homebrew-alt/pull/69 are merged.

joshmoore:
:thumbsup:

Issue #3231, opened 2014-11-28
Attachments from local do not work in Web


pwalczysko:
Bug found during testing of https://github.com/openmicroscopy/openmicroscopy/pull/3187#issuecomment-64394345

NOTE: THIS PROBLEM IS NOT on 5.0.6 !!! - nightshade is okay and does attachments allright.

![screen shot 2014-11-26 at 14 17 47](https://cloud.githubusercontent.com/assets/2478303/5202675/e1af2ff8-7577-11e4-8c04-be448faad6a2.png)
Bug: Attaching files from local disk is not working, instead produces `no files chosen` response
- select a Project as user-4 (read-only-1)
- Attach file
- in the dialog, select local file
- use the Browse button, select a valid local file (a screenshot and a csv file was attampted)
- click `Accept` -> the file is not attached, insted a message gets produced in the Attachments (see screenshot)

```
http://trout.openmicroscopy.org/merge/webclient/annotate_file/?project=559&index=0


<script type="text/javascript">
    $(document).ready(function(){
        $(":button:contains('Accept')",
            $("#choose_attachments_form").parent()).prop("disabled", true).addClass(
                'ui-state-disabled');

        $('#id_annotation_file, #id_files').change(
            function(){
                if ($(this).val()){
                    $(":button:contains('Accept')",
                        $("#choose_attachments_form").parent()
                            ).prop("disabled", false).removeClass('ui-state-disabled');
                }
                else {
                    $(":button:contains('Accept')",
                        $("#choose_attachments_form").parent()).prop("disabled", true).addClass(
                            'ui-state-disabled');
            }
        });

    });
</script>

<div>

    <h1>Upload a new file:</h1>
    <span>Choose a local file to upload and attach</span>
    <input id="id_annotation_file" name="annotation_file" type="file" />
</div>

<p>
    <h1>And/or attach an existing File:</h1>
    <span>Select the attachments to add</span>

    <div><select multiple="multiple" class="existing" id="id_files" name="files" size="8">
<option value="46502">Batch_Image_Export.zip</option>
<option value="48709">Batch_Image_Export.zip</option>
<option value="48758">Batch_Image_Export.zip</option>
<option value="47096">error report.docx</option>
</select></div>
    <!-- hidden fields used to specify the objects we're tagging -->
    <div class="hiddenField"></div>
    <div class="hiddenField"></div>
    <div class="hiddenField"><select multiple="multiple" id="id_project" name="project" size="10">
[...]

PR #3237, opened 2014-12-02
Fix web attachment error msg develop

zeb:
This PR attempts to make alert messages more explicit in case of failed file annotations upload in OMERO.web.

Testing requires nginx with a config such that:
* either no `client_max_body_size` is configured, hence using the default 1MB value
* or explicitly set to `client_max_body_size 1M;`

Test cases:
* Standard file annotation:
  * create a dataset and attach a file larger than 1MB via the webclient from the local filesystem
  * the browser's alert popup should display "Upload failed [413 Request body too large]" instead of "undefined"
* Batch file annotation:
  * create two datasets, select both in the left panel and repeat the file upload >1M
  * an alert popup should be displayed with a message as above instead of the error being silently ignored

Note: 
the issue is reproducible on `5.0.6` but testing on `develop` currently depends on https://trac.openmicroscopy.org/ome/ticket/12663 first (see #3231).

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/batch_annotate.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/batch_annotate.html
@@ -187,6 +187,7 @@
                     },
                     error: function(data){
                         $("#fileann_spinner").hide();
+                        alert("Upload failed [" + data.status + " " + data.statusText + "]");
                     }
                 });
                 // prepare dialog for choosing file to attach...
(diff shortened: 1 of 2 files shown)


joshmoore:
Any next step here?

will-moore:
I think the ON_HOLD label can be removed, since the local file upload issue is fixed.

joshmoore:
Note: "ON_HOLD" has no effect on snoopy. This has been in the build for ~2 months. My biggest question at the moment is whether or not this is wanted/needed on 5.0.

manics:
Good to merge. Uploading a large file results in an alert: `Upload failed [413 Request Entity Too Large]` instead of `Undefined`. Batch file annotations display the alert.

jburel:
--no-rebase

PR #3267, opened 2014-12-10
Mt big images (rebased onto dev_5_0)

jburel:
This is the same as gh-3134 but rebased onto dev_5_0.

----

Fix some issues with ROI and big images
To test:
* Import a big image (>4kx4k plane) and some non big images.
* Open the measurement tool "results" tab is displayed
* Draw a rectangle (or any other shape)
* Check that the "results" tab is displayed and columns populated. (big images)
* Check that the formatting of the results tab. (cc @dominikl)
* Check that the area is now displayed when drawing a rectangle for example
* for big images, check that the font size changes according to the resolution level. (default was always 12)


dominikl:
Works like expected

PR #3269, opened 2014-12-10
Play movie (rebased onto dev_5_0)

jburel:
This is the same as gh-2908 but rebased onto dev_5_0.

----

Fix few bugs related to movie player cf. https://trello.com/c/zmxUtwWp/63-movie-player-issues
To Test:
* Select an image with one z, Make sure the Error no longer occurs
* Select an image with several z and t
* Modify the value in the text field. No longer required to press Enter
* When moving to another field. Check that the value of the previous field is modified only if the value is not in the correct range.


dominikl:
Works like expected!

PR #3282, opened 2014-12-15
Make movie web

will-moore:
This adds "Make Movie" functionality to the web client.
It should replicate the Insight functionality.

Test on images with sizeT > 1 or sizeZ > 1 or both.

 - Test that Make Movie option is enabled ONLY when a single image with sizeT > 1 or sizeZ > 1 is selected.
 - Check that the Make Movie dialog corresponds to Insight's when the same image is selected.
 - Check that edits to parameters in the UI have the expected effect in the exported movie (check it's the same as Insight).

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
@@ -208,8 +208,15 @@ class BaseContainer(BaseController):
         elif objDict is not None:
             if 'image' in objDict or 'dataset' in objDict:
                 thumbnailFig['enabled'] = True
+
+        makeMovie = {'id': 'MakeMovie', 'name': 'Make Movie', 'enabled': False,
+            'tooltip': "Create a movie of the image"}
+        if self.image and (self.image.getSizeT() > 0 or self.image.getSizeZ() > 0):
+            makeMovie['enabled'] = True
+
         figureScripts.append(splitView)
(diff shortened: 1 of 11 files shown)


will-moore:
I think we should stick with spelling "Scalebar" as one word, but happy to change if this isn't the consensus.
I've fixed the handling of images named with path/to/image.tiff. Now we just use the last name in the path. This provides a name that can be safely used without having to subsequently handle the path. Insight should probably change to doing the same. However, the change to make the script actually use this value will have to come in a different PR (scripts repo).

gusferguson:
@will-moore 

Tested using http://trout.openmicroscopy.org/merge/webclient/?# user-3 read-only-1

Downloaded file now saved with name from script text box.

will-moore:
@jburel If you're happy with this, could you merge please? Thanks.

jburel:
Adding comment to https://trello.com/c/ll8b4rFP/62-script-improvement
so we remember to unify/

joshmoore:
Do any of the JS function renamings here need to be documented? /cc @hflynn

will-moore:
I don't know of any JS documentation or external use of these methods, but it is possible that @aleksandra-tarkowska 's embedding of the viewer might use some of these. I'll check...

gusferguson:
@will-moore @jburel - testing https://github.com/openmicroscopy/openmicroscopy/pull/3518
- still Scalebar in Web and Scale Bar in Insight

PR #3292, opened 2014-12-17
keep our annotations on others' data

mtbc:
With this fix, it is no longer possible to remove one's annotations from others' data by moving one's own annotated data. Furthermore, if others annotate our data and we move it, their comments should move too (unless also attached to something not moving!) but never their own tags.

cc: @pwalczysko
--no-rebase

### DIFF
--- a/components/blitz/resources/ome/services/blitz-graph-rules.xml
+++ b/components/blitz/resources/ome/services/blitz-graph-rules.xml
@@ -165,7 +165,7 @@
         <bean parent="graphPolicyRule" p:matches="POFM:PixelsOriginalFileMap.parent = OF:[E]{r}, POFM.child = Pixels[E]/o"
                                        p:changes="OF:{a}"/>
         <bean parent="graphPolicyRule" p:matches="L:ILink[!D].child = C:!Job[E]{r}/!d" p:changes="C:{a}"/>
-        <bean parent="graphPolicyRule" p:matches="L:ILink[!D].parent = [E]/o, L.child = C:!Pixels[E]{r}" p:changes="C:{a}"/>
+        <bean parent="graphPolicyRule" p:matches="L:ILink[!D].parent = [E], L.child = C:!Pixels[E]{r}" p:changes="C:{a}"/>
         <bean parent="graphPolicyRule" p:matches="!ILink[E]{ia} = X:!ILink[E]{r}" p:changes="X:{a}"/>
         <bean parent="graphPolicyRule" p:matches="X:[E]{r} == [E]{a}" p:changes="X:{a}"/>
         <bean parent="graphPolicyRule" p:matches="[DI].instrument == I:Instrument[E]{r}" p:changes="I:{a}"/>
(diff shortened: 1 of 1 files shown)


mtbc:
So, sometimes we should never move others' annotations with an image, e.g., tags, sometimes we always should, e.g., comments. @pwalczysko, of the different kinds of annotation, which are the kinds we should move when we move an image, and which aren't? (@jburel: Is this exactly what sharable / non-sharable is about?)

I think the possible annotations are: Boolean, Comment, Double, File, List, Long, Map, Tag, Term, Timestamp, XML.

I figure, for instance, that others' File, Tag, Term annotations we shouldn't move, but Boolean, Comment, Double, Long, Timestamp we should? What about the others?

jburel:
Technically all annotations can be shared, 
we decided to add rules in our clients e.g. comment we won't share, boolean
Nothing prevents others to share comment outside our clients.

mtbc:
Okay, thank you @jburel, for @pwalczysko to decide then which behavior he desires for which annotations, but no hurry as I'm seeing if I can reproduce his "lost ROIs" issue.

manics:
@mtbc A very large can of worms ... annotations can be created by users or automatically (e.g. scripts, some other post-processing algorithm), and the expected behaviour for a user-created annotation might not be the same as that for a machine-created one.

mtbc:
@manics: Within this PR I'm not touching *that* one with a bargepole. :smiley: If it helps, with the graphs reimplementation, you should be able to put those annotations in a different namespace and, *with moves and deletes by scripts,* have those annotations treated differently from what we might usually expect. If that isn't sufficient, certainly get something on to the DevX board to cause discussion.

Except, it's within this PR if you want to argue against @pwalczysko's thought that, if a different user puts their tag on my image (and no other) and a comment on my image, when I move that image to another group then the comment should move too but not the tag.

mtbc:
@pwalczysko: I am not managing to lose any ROIs unexpectedly. Can you tell me a way to reproduce the issue on trout? The simpler the better.

mtbc:
@pwalczysko: With the next build you should find that comments and tags work as you would like.

jburel:
@manics: in some cases, we have used ns to differentiate annotation e.g. rating is a long that we handle differently. We will use the same strategy for the "client" mapAnnotation

pwalczysko:
Still having problems with "case 5". The Tags and Annotations do not move, which is expected, but the Comments DO NOT move. 
Setup: an owner in Read-annotate-1 group makes a comment on an image of a user. Then the user moves the image into another Read-annotate group (ra2). The comment gets lost.

pwalczysko:
The simple workflow from the top comment https://github.com/openmicroscopy/openmicroscopy/pull/3292#issuecomment-67357312 still works as expected as of today.

pwalczysko:
Except for the problem with the comments see https://github.com/openmicroscopy/openmicroscopy/pull/3292#issuecomment-68891226 there are no problems here.

mtbc:
@pwalczysko: Surprisingly, your outstanding problem does not appear to be a regression. With a setup from your `man_test_create_users_modif` and a server config with omero.graphs.wrap=false, so that the previous graphs code operates (not the reimplementation), others' comments are *still* lost.

If I try fixing this, integration tests break that expected the previous behavior. (For instance, @jburel, in `AnnotationMoveTest` we do *not* expect non-sharable annotations from RWRA to RWRA to be moved.)

Given time pressures, I'd much like to leave improvements in functionality that break the existing tests to after 5.1.0: I'd suggest we merge this PR and you note this issue for the graphs work UX wishlist (at least make sure your spreadsheets already covers it).

pwalczysko:
I agree. Happy with merging this.

jburel:
Merging. Current implementation does not yet reflect what is the expected behaviour cf. graph work.

mtbc:
The RWRA to RWRA issue may be related to http://trac.openmicroscopy.org/ome/ticket/12023.

PR #3295, opened 2014-12-18
MapAnnotation Pojo and Tests 

dominikl:
Adds the Pojo for MapAnnotations and the integration tests.
Test: Check that the (annotation related) integration tests pass

(This PR also adds a namespace constant for client created MapAnnotations to Constants.ice /cc @will-moore  )

--no-rebase
--depends-on #3334

### DIFF
--- a/components/blitz/resources/omero/Constants.ice
+++ b/components/blitz/resources/omero/Constants.ice
@@ -158,6 +158,7 @@ module omero {
         const string NSINSIGHTTAGSET = "openmicroscopy.org/omero/insight/tagset";
         const string NSINSIGHTRATING = "openmicroscopy.org/omero/insight/rating";
         const string NSMOVIE = "openmicroscopy.org/omero/movie";
+        const string NSCLIENTMAPANNOTATION = "openmicroscopy.org/omero/client/mapAnnotation";
     };
 
     /**
--- a/components/tools/OmeroJava/test/integration/AbstractServerTest.java
+++ b/components/tools/OmeroJava/test/integration/AbstractServerTest.java
@@ -92,6 +92,9 @@ import omero.model.LightSourceAnnotationLink;
 import omero.model.LightSourceAnnotationLinkI;
 import omero.model.LongAnnotation;
 import omero.model.LongAnnotationI;
+import omero.model.MapAnnotation;
+import omero.model.MapAnnotationI;
+import omero.model.NamedValue;
 import omero.model.OriginalFile;
 import omero.model.OriginalFileAnnotationLink;
 import omero.model.OriginalFileAnnotationLinkI;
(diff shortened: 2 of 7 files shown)


jburel:
Discussed with @dominikl, removing the exclude label so we can test the UI work.

dominikl:
--depends-on #3325

mtbc:
Appears to break a bunch of integration tests, e.g., in DeleteServiceTest.

components/blitz/resources/omero/Constants.ice
```diff
@@ -158,6 +158,7 @@ module omero {
         const string NSINSIGHTTAGSET = "openmicroscopy.org/omero/insight/tagset";
         const string NSINSIGHTRATING = "openmicroscopy.org/omero/insight/rating";
         const string NSMOVIE = "openmicroscopy.org/omero/movie";
+        const string NSCLIENTMAPANNOTATION = "openmicroscopy.org/omero/client/mapAnnotation";
```
joshmoore: I'm not quite sure about the use of this namespace. How does this differ from just any map annotation?

jburel: The Idea was to have one map annotation that the user will be editing. One per user.
Other could be generated during acquisition are not editable in clients.
This was discussed in MapAnnotation mockup

dominikl:
Looks like MapAnnotations can't be deleted or moved. I assume it's not an issue of the tests, but the delete and move methods have to be adjusted to handle MapAnnotations. Can someone give me a hint where to do this (I'm still not really familiar with the server side code)?

components/tools/OmeroJava/test/integration/AbstractServerTest.java
```diff
@@ -1604,10 +1615,20 @@ public void run(omero.util.TileData data, int z, int c, int t,
...
+		ma.setMapValue(values);
+		if (ns != null)
+			b.setNs(omero.rtypes.rstring(ns));
+		ma = (MapAnnotation) iUpdate.saveAndReturnObject(ma);
+        
         ids.add(c.getId().getValue());
         ids.add(l.getId().getValue());
         ids.add(b.getId().getValue());
-
+        ids.add(ma.getId().getValue());
```
mtbc: also need to add `ma` into code below to actually create the link?

dominikl:
Travis is continuously failing, without a sensible error message (as far as I can see); don't know why.

components/tools/OmeroJava/test/integration/DeleteServiceTest.java
```diff
@@ -201,7 +201,10 @@
 
     /** Identifies the File. */
     public static final String REF_FILE = "/FileAnnotation";
-
+    
+    /** Identifies the File. */
```
mtbc: the file?

dominikl:
Fixed the wrong comment. If it's ok otherwise, would be great if it could be merged soon, so I can open my Insight UI PR for the MapAnnotations to look at.

joshmoore:
:+1: for the comment fixed. Any discussion on the namespace can take place elsewhere. Merging.

PR #3296, opened 2014-12-18
Integration library

sbesson:
See https://trello.com/c/8IkxJlSD/323-integration-library

The `library.py` file under `OmeroPy` is used for integration tests by many components including `OmeroFS` and `OmeroWeb`. To simplify upcoming work on this test library, it is now migrated under `components/tests` as a standalone package. The hardcoded `PYTHONPATH` patching in `OmeroWeb` is removed from `setup.py` and the library is added to the path in `test_setup.py`. Note this may have consequences on low-level `py.test` usage (where the integration library needs to be manually added to the `PYTHONPATH`).

--no-rebase


ximenesuk:
Everything runs okay from both `setup.py` and `py.test`. I'll add a clause to https://www.openmicroscopy.org/site/support/omero5.1/developers/testing.html#running-tests-directly regarding the additional Python path needed for the `py.test`.

Otherwise good to merge.

ximenesuk:
Does this need any further review? The change looks trivial.

PR #3301, opened 2014-12-18
Analyze shape

jburel:
This PR allows to analyse the shape not linked to any plane. The plane currently displayed is in that case selected
To test
- Login as user-4/Test Group
- Open the OME-TIFF with ROIs/mouseWithROI.ome.tif
- Open the measurement, Do not save the ROI.
- Select one the shape
- Go to the Graph, Intensity View, Intensity Result View. Check that the shape is analysed.


jburel:
--rebased-to #3306

PR #3300, opened 2014-12-18
Roi ellipse

jburel:
Fix display of ROI

To test:
- Open the Measurement tool
- Draw an ellipse
- Click Save. 
- Check that the ellipse does not "jump" to the top-left corner.

PR #3305, opened 2015-01-05
Roi ellipse (rebased onto dev_5_0)

jburel:
This is the same as gh-3300 but rebased onto dev_5_0.

----

Fix display of ROI

To test:
 * Open the Measurement tool
 * Draw an ellipse
 * Click Save. 
 * Check that the ellipse does not "jump" to the top-left corner.


jburel:
--rebased-from #3300

dominikl:
Works like expected

PR #3306, opened 2015-01-05
Analyze shape (rebased onto dev_5_0)

jburel:
This is the same as gh-3301 but rebased onto dev_5_0.

----

This PR allows to analyse the shape not linked to any plane. The plane currently displayed is in that case selected
To test
 * Login as user-4/Test Group
 * Open the OME-TIFF with ROIs/mouseWithROI.ome.tif
 * Open the measurement, Do not save the ROI.
 * Select one the shape
 * Go to the Graph, Intensity View, Intensity Result View. Check that the shape is analysed.


jburel:
--rebased-from #3301

dominikl:
Works like expected

PR #3308, opened 2015-01-05
Original metadata loading

will-moore:
This fixes timeouts when loading original metadata (see https://trac.openmicroscopy.org.uk/ome/ticket/12625). It also moves the metadata loading to "on demand" so that original metadata is not loaded unless needed.

To test:
 - Open "Acquisition" tab for an image (also test well / image) and expand the "Original Metadata" panel.
 - Original metadata should be loaded when panel is expanded, and search filters should work as before.
 - When an omero.LockTimeout is thrown, a 'try again' message is shown and clicking this should load as before (see screenshot). NB: don't know of a good way to test this. I just added ``` raise omero.LockTimeout()``` to the Blitz Gateway image.loadOriginalMetadata.

screen shot 2015-01-05 at 15 02 39


gusferguson:
@will-moore 

Tested using http://trout.openmicroscopy.org/merge/webclient/?# user-3 read-only-1

With images
- original metadata displayed as expected
- filtering of original metadata worked as expected

With plate-image - using 5.0 screen > mode1_z000_t000_im1_1.tif > Run 1204 > mode1_z000_t000_im1_1.tif [Well A3]
- first time progress bar ran for about 2 minutes and then data started appearing
- initially no filter text box
- when filter text box appeared, attempted to filter while progress indicator still spinning.
- got error message below and reverted back to no selection in data manager and empty centre pane
- logged out and in again
- tried again and progress indicator ran for a few seconds then got same error message but no reversion to default data manager
- third attempt to replicate after looking at other images - metadata started appearing after about 1 minute
- progress bar still spinning after ten minutes
- tried a filter again while spinning
- crash out to default data manager view but no error message this time

With plate-image INMAC384-DAPI-CM-eGFP_59223_1 > Run 1153 > INMAC384-DAPI-CM-eGFP_59223_1.xdce [Well E-4, Field #1]
- reports "No Original Metadata for this image"

With plate-image - using MFGTMP_081128200001_A02f239d1.C01
- original metadata appeared immediately and filter worked as expected

{{{

  File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/django/core/handlers/base.py", line 114, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)

  File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/omeroweb/decorators.py", line 452, in wrapped
    retval = f(request, *args, **kwargs)

  File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/omeroweb/decorators.py", line 502, in wrapper
    context = f(request, *args, **kwargs)

  File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/omeroweb/webclient/views.py", line 855, in load_metadata_details
    manager.annotationList()

  File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/omeroweb/webclient/controller/container.py", line 430, in annotationList
    annClass = ann._obj.__class__

<WSGIRequest
path:/merge/webclient/metadata_details/project/1211/,
GET:<QueryDict: {}>,
POST:<QueryDict: {}>,
COOKIES:{'_ga': 'GA1.2.766626352.1413448784',
 'csrftoken': '9MuY0QjG2t50EwhfR0omj6rvgb3QAbQ6',
 'phpbb3_a9mfv_k': '',
 'phpbb3_a9mfv_sid': '7b8e96435dcb1dc66f25b5ec60f87aa9',
 'phpbb3_a9mfv_u': '1',
 'sessionid': 'skjptuk4ju57ini6luopqa30m1tipu7g',
 'style_cookie': 'null'},
META:{'CONTENT_LENGTH': '',
 'CONTENT_TYPE': '',
 u'CSRF_COOKIE': u'9MuY0QjG2t50EwhfR0omj6rvgb3QAbQ6',
 'HTTPS': '',
 'HTTP_ACCEPT': 'text/html, */*; q=0.01',
 'HTTP_ACCEPT_ENCODING': 'gzip, deflate',
 'HTTP_ACCEPT_LANGUAGE': 'en-us',
 'HTTP_CONNECTION': 'keep-alive',
 'HTTP_COOKIE': 'csrftoken=9MuY0QjG2t50EwhfR0omj6rvgb3QAbQ6; sessionid=skjptuk4ju57ini6luopqa30m1tipu7g; _ga=GA1.2.766626352.1413448784; phpbb3_a9mfv_k=; phpbb3_a9mfv_sid=7b8e96435dcb1dc66f25b5ec60f87aa9; phpbb3_a9mfv_u=1; style_cookie=null',
 'HTTP_HOST': 'trout.openmicroscopy.org',
 'HTTP_REFERER': 'http://trout.openmicroscopy.org/merge/webclient/?',
 'HTTP_USER_AGENT': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) AppleWebKit/600.2.5 (KHTML, like Gecko) Version/8.0.2 Safari/600.2.5',
 'HTTP_X_REQUESTED_WITH': 'XMLHttpRequest',
 'PATH_INFO': u'/webclient/metadata_details/project/1211/',
 'QUERY_STRING': '',
 'REQUEST_METHOD': 'GET',
 'SCRIPT_INFO': '/merge',
 'SCRIPT_NAME': u'/merge',
 'SERVER_NAME': 'ome-c6100-3',
 'SERVER_PORT': '80',
 'SERVER_PROTOCOL': 'HTTP/1.1',
 'wsgi.errors': <flup.server.fcgi_base.TeeOutputStream object at 0x5231a90>,
 'wsgi.input': <flup.server.fcgi_base.InputStream object at 0x5231310>,
 'wsgi.multiprocess': True,
 'wsgi.multithread': False,
 'wsgi.run_once': False,
 'wsgi.url_scheme': 'http',
 'wsgi.version': (1, 0)}>
}}}

will-moore:
For the "5.0 screen" above, I'm also seeing problems, but a different error message.
```
  File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/omero/gateway/__init__.py", line 7143, in loadOriginalMetadata
    for l, m in ((global_metadata, rsp.globalMetadata),

AttributeError: 'NoneType' object has no attribute 'globalMetadata'
```

will-moore:
@gusferguson The reason that particular plate is causing problems is that the Global Metadata has 177421 rows, which make it slow to get the data from OMERO and to turn this into html and return it. It also breaks the jQuery plugin for quicksearch and my browser is practically brought to it's knees! I've handled the error I was seeing, so now we just get the Timeout handling I described. We could consider giving the callback more time to get the data from OMERO before timing out, but then we still have the other problems.
I don't know exactly why you're seeing the exception above (couldn't reproduce it myself) but I've added some logging so if we see it again we may have some more clues.

gusferguson:
@will-moore 
Sounds like a edge case - happy to merge it then if you are.

gusferguson:
@will-moore 

Tested using http://trout.openmicroscopy.org/merge/webclient/?# user-3 read-annotate-1

Images and plates all working as expected except for the problem plate mode1_z000_t000_im1_1.tif.

In Safari
- first attempt got time out notice after about 1 minute
- retry - acquisition data progress gif ran for 40 s and then text box and metadata displayed but progress gift continued spinning
- attempt at filter crashed out back to default data manger 

In Chrome
- replicated Safari

In Firefox
- received warning in Console Logging of Firefox:
"Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help http://xhr.spec.whatwg.org/"
- same sequence as Safari
- when tried filter while progress gif still spinning (after 5 minutes) - froze after typing 2 letters
- whole of Firefox frozen - had to force quit
- when I restarted Firefox the session was still active as I was still logged in when the Web client reopened
- no error message from OS regarding FF force-quit
- I was able to replicate this

It appears that the timeout worked the first time but not subsequently in Safari, and not at all in the other browsers.
If filtering while metadata still be ing fetched is causing problems is it possible to disable the filter text box until all data is returned to prevent crash out?

will-moore:
@gusferguson I don't think it's "filtering while metadata still being fetched" that's the problem. You can only filter once the data has loaded. The spinner you see after the data has arrived is actually the 'quicksearch' filter spinner, which is normally hidden when the quicksearch plugin is set-up. I could simply disable the quicksearch plugin if the number of rows is "large" (E.g. over 100 000) and hide the spinner and filter field. If we were going to see a lot of files like this, I could spend time trying to optimise the plugin, or pick a different one. But if this is really an edge case then we'll just display the data (and hopefully prevent crash etc). OK?

gusferguson:
@will-moore 
Sounds good - I will leave it in your hands.
Good to merge from my POV.

will-moore:
@gusferguson That commit should improve things a bit for the very large Original Metadata cases (>100000 rows). It's still slow to load them, but then we don't try to quicksearch them (hide the field instead).

gusferguson:
@will-moore 

For most images working as expected.

For the troublesome screen - mode1_z000_t000_im1_1.tif 
- progress spinner for 1 minute and then original metadata appeared
- no text box for filtering
- could scroll and view all metadata
- no crash out to data manager as before

Behaves as expected.
Good to merge.

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_acquisition.html
```diff
@@ -70,6 +61,49 @@
...
+                        var url = $("#orig_metadata_link").prop('href');
+                        $("#orig_metadata")
+                            .html(loadingHtml)
+                            .load(url, function( response, status, xhr ) {
+                                if (xhr.status == 408) {
+                                    origMetadataLoaded = false;
+                                    $("#orig_metadata").html("Timeout. <a href='#' class='load_original_metadata'>Try again</a>");
+                                } else {
+                                    // setup quick search fields (but not if MANY rows, since it crashes!)
+                                    if ($("#global_metadata tr").length < 100000) {
```
joshmoore: Does this number (`length < 100000`) need to be reviewed / made configurable?

will-moore: @joshmoore This is really just to catch the 'edge case' that @gusferguson was testing with. I haven't tried to test the performance of the quicksearch plugin with various different list sizes, but certainly could do.

gusferguson:
@will-moore 

Behaves as expected.
For the troublesome screen - mode1_z000_t000_im1_1.tif data appears after 46 seconds.

Good to merge.

jburel:
@will-moore: could we make the  limit (i.e. 100000) a parameter than we could easily configure?

will-moore:
@jburel Why would you want to configure it?

jburel:
It is an arbitrary value, as we have seen in insight when limiting preview after a given size w/o being to parametrize it was not well received.
In various places when choosing an arbitrary value, we are trying to make it easily configurable
e.g. insight: displaying table instead of thumbnails view if more than 100 (value picked but easily modified), server: threshold for big images.
Offering the ability to easily modify this type of values is important

will-moore:
In this case it is not an arbitrary value. I've tested the jQuery quicksearch plugin with different sizes of Original Metadata lists and the plugin is broken for lists over 60000.
I don't want to spend a lot of time investigating this and fixing the plugin or choosing a different one since this is an edge case to have such a large list and and this functionality is a "bonus" anyway (not essential and not supported in Insight).

jburel:
Having the value set in a common place via an external parameter is easier to find if/when the plugin is updated (not necessary by you). Nothing to do with a "bonus" feature or if implemented in insight.

Especially when there is a current effort to do e.g. https://github.com/openmicroscopy/openmicroscopy/pull/3340 (this is for common parameters) but this is more as a principle
I do not want to go on about it but we should try to follow that path as a general rule.

will-moore:
@jburel You OK to merge this, so I can add "Download original metadata" based on this?

will-moore:
@jburel Can you merge this if you're happy with it?

jburel:
Discussed with @will-moore.
Not blocking this PR, need to review the quick search plugin later to see if the limit can be removed.
Merging.

PR #3309, opened 2015-01-05
LightSource Attenuation as percent

will-moore:
Follow up for metadata panel display, see: https://trello.com/c/PiLauXgp/47-units-client-follow-up

To test (using units.ome.tif to test mostly):
 - Web should now do the same as Insight and show Light Source attenuation of 0.8 as "Attenuation (%): 80".
 - LightSource type should be displayed for Arc and Filament objects.
 - ND filter (%) should also be converted from fraction to percent, and rounded to 2 significant figures. Used a different DV file to test this.

screen shot 2015-01-05 at 16 55 35

screen shot 2015-01-05 at 16 13 38

screen shot 2015-01-05 at 15 26 57

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/forms.py
@@ -1883,7 +1883,7 @@ class MetadataLightSourceForm(forms.Form):
             self.fields['attenuation'] = forms.CharField(
                 max_length=100,
                 widget=forms.TextInput(attrs={'size':25, 'onchange':'javascript:saveMetadata('+str(lightSourceSettings.id)+', \'attenuation\', this.value);'}),
-                initial=lightSourceSettings.attenuation,
+                initial=lightSourceSettings.attenuation * 100,
                 label="Attenuation (%)",
                 required=False)
         else:
--- a/components/tools/OmeroWeb/omeroweb/webclient/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/forms.py
@@ -364,11 +364,15 @@ class MetadataChannelForm(forms.Form):
         
         # ndFilter
         try:
-            if logicalCh is not None:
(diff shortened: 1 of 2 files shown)


dominikl:
Light Source attenuation is shown as percentage, but some rounding errors happen:
screen shot 2015-01-06 at 09 02 45

will-moore:
That last commit fixes the display of NDFilter when it's 100% as in this screenshot:

screen shot 2015-01-06 at 12 42 28

will-moore:
@dominikl Lightsouce attenuation should be fixed now.

dominikl:
Works like expected.

jburel:
Merging. Thanks

joshmoore:
Sorry for coming to this late, but is it an issue to be setting `getcontext()` without rolling it back? https://docs.python.org/2/library/decimal.html#decimal.localcontext might be a better option. /cc @knabar @chris-allan

knabar:
@joshmoore Additionally, the change made by adjusting the context outside of the method where it's needed could be overwritten elsewhere

joshmoore:
Agreed. @will-moore: you ok to open a fix?

chris-allan:
Just further adding that meddling with the decimal context is absolutely not a good idea. That has far reaching implications across the entire codebase. Explicitly performing string conversions with the correct format string (`%.2f` in the case of two desired decimal places) and/or rounding would be much preferred.

will-moore:
OK, I was looking for a way to format to 2 significant figures (rather than 2 decimal places).

chris-allan:
Completely understood. However, assigning a specific precision to a decimal context is unlikely to give you the desired visual uniformity if that's your intention. Fractional percentages will still be displayed below 10%.

If all you want to do is avoid representation error (https://docs.python.org/2/tutorial/floatingpoint.html#representation-error) then what you want to do is use the format string and round or just truncate for display with `int()`.

will-moore:
@chris-allan Next attempt at formatting is #3339

PR #3311, opened 2015-01-05
Paste rendering settings on selected wells

dominikl:
See Ticket: [12441](http://trac.openmicroscopy.org.uk/ome/ticket/12441)

To test, check the following rendering settings workflow with Plate/Wells (expected to work like Dataset/Images) 
- Modify the rendering settings of a particular well (of a plate you are not the owner of, important for the last check)
- Copy the rendering settings
- Select a subset of other wells and paste the rendering settings
- Reset some of the wells by "Set Original and Save"
- Reset some of the wells by "Set Owner's and Save"

--no-rebase

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/ManageRndSettingsAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/ManageRndSettingsAction.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.dataBrowser.actions.ManageRndSettingsAction 
  *
  *------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


dominikl:
As @will-moore noticed, if multiple images/wells are selected, that acquisition and preview tab should be disabled. I'll fix that in another PR, already created a ticket for it [12694](http://trac.openmicroscopy.org.uk/ome/ticket/12694)

will-moore:
All working as described.
A couple of minor points: Probably should disable Acquisition and Preview tabs when multiple wells selected (as discussed). Also, selecting a range of wells would be much easier if Shift-click was supported.
But functionality above works fine. Good to merge.

jburel:
Merging since RFE will be done in another PR.

PR #3186, opened 2014-11-13
Roi inspector

jburel:
This PR fixes 2 more issues with measurement tool
To test
- Open the measurement tool
- Draw an ellipse. 
  - Go to the inspector tab. Check the scale proportionally is a checkbox and not N/A
- Edit the width and height of the ellipse in the inspector tab.
- Check that the shape of the ellipse is modified.

PR #3312, opened 2015-01-06
Roi inspector (rebased onto dev_5_0)

jburel:
This is the same as gh-3186 but rebased onto dev_5_0.

----

This PR fixes 2 more issues with measurement tool
To test
* Open the measurement tool
* Draw an ellipse. 
 * Go to the inspector tab. Check the scale proportionally is a checkbox and not N/A
* Edit the width and height of the ellipse in the inspector tab.
* Check that the shape of the ellipse is modified.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/FigureTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/FigureTableModel.java
@@ -257,7 +257,7 @@ public class FigureTableModel
 	 */
 	public void setValueAt(Object value, int row, int col)
 	{
-		if (col == 0) return;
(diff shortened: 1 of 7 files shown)


jburel:
--rebased-from #3186

pwalczysko:
All works as expected when tested with tests written in the header of this PR.

RFE: Possibly for another PR - the unexpected behaviour starts when you save the ellipse region you have just created. When the saved region is reselected and then the width and height is edited and changed, nothing happens. Also the area displayed when you first drawed the ellipse is gone after save, even when you tick ``Show Measurements`` box. Note that for example ``line width`` parameter DOES work after Save of the region, in contrast.

sbesson:
Since this is a rebase, merging.

PR #3314, opened 2015-01-06
Metadata followup

will-moore:
This fixes a few more points from https://trello.com/c/PiLauXgp/47-units-client-follow-up
To Test:

 - Channel Excitation and Emission should have units in brackets.
 - Display of Serial Number & Lot Number for Microscope and Objective should be fixed (compare with Insight)
 - Pump should be displayed for Lasers (test using alternate.ome from units testing).


dominikl:
Point 1 and 2 work like expected, but could not check 3, because alternate units image can't be opened.
Don't know where my QA report has gone, but the error message was:
```
File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/omeroweb/webclient/forms.py", line 1977, in __init__
    lsAttn = Decimal(lsAttn)/1

  File "/usr/lib64/python2.6/decimal.py", line 649, in __new__
    "First convert the float to a string")

TypeError: Cannot convert float to Decimal.  First convert the float to a string
```

will-moore:
@dominikl That should fix the bug you're seeing above (which was actually from #3309)

dominikl:
Good to merge

jburel:
Thanks Merging

PR #3310, opened 2015-01-05
Unmark broken tests

ximenesuk:
This PR unmarks some tests that were marked as broken but have either been consistently passing or failing very, very intermittently. The intermittent tests are marked as such mainly for documentation reasons.

Minimal review necessary, mainly that the CI tests are green.

The passing 'broken' tests excluded from this PR are those in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/test_permissions.py#L979 which probably need a more considered review. /cc @joshmoore @dpwrussell 

--no-rebase

PR #3318, opened 2015-01-07
Itest improvements

sbesson:
This PR refactors the `ITest` class in the Python library to perform the setup at the class level instead of the method level. This should not affect the validity of the tests but mostly result in a performance improvements while running the integration tests due to the reduction of the setup overhead.

Specific test cases which were relying on the previous per-method setup should be corrected e.g. by using `self.new_clients()`.

To test this PR, check the Python/Web integration  tests are green and that the full time required to complete the Python integration tests is significantly reduced.

/cc @ximenesuk

### DIFF
--- a/components/tests/python/library/__init__.py
+++ b/components/tests/python/library/__init__.py
@@ -98,34 +98,6 @@ class ITest(object):
         cls.update = cls.sf.getUpdateService()
         cls.query = cls.sf.getQueryService()
 
-    # def setup_method(self, method):
-    #
-    #     self.OmeroPy = self.omeropydir()
-    #
-    #     self.__clients = Clients()
-    #
-    #     p = Ice.createProperties(sys.argv)
-    #     rootpass = p.getProperty("omero.rootpass")
-    #
-    #     name = None
-    #     if rootpass:
-    #         self.root = omero.client()  # ok because adds self
-    #         self.__clients.add(self.root)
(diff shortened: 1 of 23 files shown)


ximenesuk:
Is there anymore to come on this? The job looks pretty good. I'll take a look later by means of review.

joshmoore:
Sorry, missed @sbesson's ping last night. Removing breaking.

ximenesuk:
This looks fine, the vast majority of tests are passing as normal and the build seems faster. There may be oddities related to setup and teardown, but given the coverage of this PR I'd say it is good to merge and then pick up those oddities in later PRs.

sbesson:
Last commit should improve script tests output after marking additional tests as not broken in https://github.com/openmicroscopy/openmicroscopy/pull/3310.

sbesson:
See https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/342/testReport/ and https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration/buildTimeTrend.

ximenesuk:
+1

Locally I still got a fail but I was getting those intermittently anyhow! I can retest further or we can merge and then look at any subsequent issues?

sbesson:
I would be in favor of merging this. I will create a card to investigate the `test_scripts` workaround.

sbesson:
--no-rebase

PR #3319, opened 2015-01-07
Modify setup.py usage in Python testing

ximenesuk:
This PR tries to fix a long-standing bugbear for some of us on the Python side, it allows the `-s` option to be passed through from 'setup.py`. It implements a `-t` option (not used anywhere else in `py.test`) for the test path. It also removes the inherited options (only used for `unittest`) which led to a confusing help listing.

To test:
 + firstly the CI tests should run okay
 + `./setup.py test --help` should no longer display the old `-m`, `-s` and `-r` options at the top of the `PyTest` option list
 + `./setup.py test -t <path to tests>` should run a folder of tests
 + `./setup.py test -t <path to tests> -s` should run a folder of tests and output any stdout with the tests to the console.

This sill
--no-rebase

### DIFF
--- a/components/tools/python.xml
+++ b/components/tools/python.xml
@@ -109,7 +109,7 @@ build_year = "${python_build_year}"
                 <arg value="test"/>
                 <optionalArgs/>
                 <arg value="-v"/>
-                <arg value="-s"/>
+                <arg value="-t"/>
                 <arg file="@{target}"/>
                 <arg value="--junitxml=${testreports.dir}/junit-results.xml"/>
             </setup_py>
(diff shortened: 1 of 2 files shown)


joshmoore:
What jobs needed updating for this? Just thinking about possible impact. I also see this:

```
$ git grep " -s " | grep setup.py
components/tools/OmeroPy/setup.py:      ./setup.py test -s test/gatewaytest
```

which might be worth looking at eventually.

ximenesuk:
Gah! Good catch. I didn't realise we had docs in the `setup.py`. I think this pre-dates the refactoring of the code to `test_setup.py`. Given we now have this documented on the webpages I'd suggest just dropping this chunk as `setup.py` is more general than just running `py.test`.

As far as I know there is no wider impact, the jobs all access this through the build file and that has been updated.

PR #3320, opened 2015-01-07
Shift click range selection for wells

dominikl:
Enables shift+click range selection for wells and disables acquisition and preview tab if multiple images/wells are selected. (Follow-up to PR #3311)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorUI.java
@@ -316,11 +316,10 @@ class EditorUI
     {
         ImageData img = model.getImage();
         if (img == null) return;
-        tabPane.setEnabledAt(ACQUISITION_INDEX, img.getId() > 0);
         boolean multi = model.isMultiSelection();
         boolean preview = model.isPreviewAvailable();
         tabPane.setEnabledAt(RND_INDEX, preview && !multi);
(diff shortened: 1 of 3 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorUI.java
```diff
@@ -317,8 +317,10 @@ void handleImageSelection()
         ImageData img = model.getImage();
         if (img == null) return;
         tabPane.setEnabledAt(ACQUISITION_INDEX, img.getId() > 0);
+        boolean multi = model.isMultiSelection();
         boolean preview = model.isPreviewAvailable();
-        tabPane.setEnabledAt(RND_INDEX, preview);
+        tabPane.setEnabledAt(RND_INDEX, preview && !multi);
+        tabPane.setEnabledAt(ACQUISITION_INDEX, !multi);
```
mtbc: this doesn't render line 319 obsolete? (I really don't know)

mtbc:
Functionality works fine so, apart from code question above, looking good. :+1:

dominikl:
@mtbc That's true. I removed the obsolete line. I don't think a re-review is necessary because of that.

mtbc:
We don't still need the `img.getId() > 0` condition?

dominikl:
You're right, to be safe I added the ```img.getId() > 0``` again, although I can't see a case, in which this wouldn't be true.

mtbc:
Thanks, code now looks good to me.

jburel:
Travis is now green. Merging

PR #3322, opened 2015-01-08
Rating web

will-moore:
This is the updated version of previous PR: #3057

Exploring a new way of saving space with ratings. I now show the user's own rating first, then all other ratings are grouped by value, with brackets showing the count of each and tooltip showing who added them (like Insight does with tags).
When you have not added a rating, the 0 stars image will not be surrounded by a block (as in previous PR) but will still "flow" to save space (won't have a line to itself as before).
Hovering over your own rating shows a "X" to remove it.

NB: This is not ready for review yet.

screen shot 2015-01-08 at 13 46 57


gusferguson:
@will-moore 

Tested using http://trout.openmicroscopy.org/merge/webclient/ user-3, user-5, user-9, user-11, user-12 Safari and Chrome on Mac.

The ratings re displayed correctly, but I am not happy with the whole paradigm.

The idea of listing the ratings and numbers is to give a single glance appreciation of how the ratings are distributed as on Amazon:

safari

This is not achieved with the suggested display as
1. you have to hunt and pick and process what you are looking at - "is that 3 or 4 stars” and they are on different lines and yet not aligned.
2. the owner’s rating doesn’t count in the score for that number:

safari

Here user-3 (owner) and user-5 have rated as 4 and yet at a glance it looks like just one 4 rating exists.

If we go this way I think:
1. the actual mechanism for rating and the display of the ratings needs to be clearly differentiated
2. the ratings need to be stacked and aligned as in the amazon example.

This is clearly going to not help with space saving, so we are trying to square circles here.

We could have the “Rate” button to the right and a simple table - no bars just “5 star: 2” etc. on the left. We hide this table if there are no ratings and take the space hit when there are ratings.

Any of this will leave us completely up the creek for displaying the ratings in batch annotation mode. We would be looking at a table per image/container. Nightmare.

I think this needs a lot more thought. I am happy to mock up the table idea if that would help.

will-moore:
I think we're not aiming to copy amazon or similar but should treat ratings a bit more like tags: It's most likely that your own rating is the only one there, and if there are others, they're still not as important as your own one. I don't think we're interested in averages (but could add that if needed). One line per rating is what I had before, but that was deemed to take up too much space so I moved to doing it more like tags (like in the scoping mock-ups).

gusferguson:
@will-moore 

Tested using http://trout.openmicroscopy.org/merge/webclient/ user-3, user-9, user-11, user-12 read-annotate-1

I am not at all happy with how this is going.

If you rate and look at it as different users you get completely different views - it is extremely difficult for a user to understand what is being shown.

These are the 4 different views of the four users who have rated the image:

screenshots1

There really needs to be a better distinction between the mechanism for the user to rate and the display of all the ratings.
There also needs to be consistency in the display of the numbers of each rating.

I think this needs to be properly scoped and discussed - there are lots of ways of doing it and more than your and my opinion needs to feed into this. Here is just one option:

ratings2

If you want a quick fix then I suggest you bite the bullet and do the average rating like Insight until such time that we have had time to come up with another agreed solution.

jburel:
It is better to scope it properly, otherwise we will waste a lot of time (that we currently do not have!)

For now we should implement a simple solution: either average like insight or only the user rating (no average). The second option will require an insight change.

joshmoore:
This now conflicts with `origin/develop` and so is no longer in the daily builds. @will-moore : are you ok to implement the insight average rating as mentioned in https://github.com/openmicroscopy/openmicroscopy/pull/3322#issuecomment-69553116 ?

will-moore:
@joshmoore OK, I'll just display an average rating for now.

gusferguson:
@will-moore 

Tested using http://trout.openmicroscopy.org/merge/webclient/ user-3 and user-11 Safari and Chrome on Mac.

Behaves as expected.
Good to merge.

PR #3323, opened 2015-01-08
5.0 graphs implementation deprecation

mtbc:
After #3228 the previous `spec.xml`-based graphs implementation is no longer being actively tested. Deprecation warnings are added accordingly: `omero.graphs.wrap` no longer available in 5.2, API-compatibility layer no longer available in 5.3 (to give people more time to update clients and scripts).

@hflynn: If merged, I must update `omero/developers/Server/ObjectGraphs.txt` accordingly.

--no-rebase

### DIFF
--- a/components/blitz/resources/omero/cmd/Graphs.ice
+++ b/components/blitz/resources/omero/cmd/Graphs.ice
@@ -40,8 +40,8 @@ module omero {
 
         };
 
-        ["deprecated:omero::cmd::GraphModify is deprecated"]
-        ["java:type:java.util.ArrayList<omero.cmd.GraphModify>:java.util.List<omero.cmd.GraphModify>"]
+        ["deprecated:omero::cmd::GraphModify is deprecated",
+         "java:type:java.util.ArrayList<omero.cmd.GraphModify>:java.util.List<omero.cmd.GraphModify>"]
         sequence<GraphModify> GraphModifyList;
 
         /**
--- a/components/blitz/resources/omero/cmd/Graphs.ice
+++ b/components/blitz/resources/omero/cmd/Graphs.ice
@@ -16,6 +16,7 @@ module omero {
         /**
          *
          **/
+        ["deprecated:use omero::cmd::GraphModify2 instead"]
         class GraphModify extends Request {
             string type;
             long id;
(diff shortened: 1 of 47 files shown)


joshmoore:
In general, this looks correct, but I am slightly concerned with the build usability after this since I spent a goodly amount of time reducing the deprecated warnings. I wonder if migration to another component with a quieted build would be sensible. Something like `components/server-deprecated`.

mtbc:
I could add `@SuppressWarnings( "deprecation" )` to the actually deprecated classes to avoid warnings within that set. Outside that set, we probably *want* the warnings, to tell us where to change clients, etc. accordingly. Would that make sense?

joshmoore:
Within the implementation, I'm not overly concerned, but there's little you can do about the Ice-generated deprecation warnings.

mtbc:
Caused by `Graphs.ice`? We probably do want those warnings, bah. Happy to go with whatever you think. I'll add some annotations to the Java classes anyway, every little helps.

mtbc:
If it helps, I can divert effort into rewriting clients to not use the deprecated requests.

joshmoore:
@mtbc : sorry, have been a bit inundated. I don't think there's anything wrong with this PR, and perhaps we should just go ahead and merge. Mostly I've just been thinking about (or trying to find time to do so) some better way to handle this. I agree completely that we need the `deprecated` markers, but both due to Java and Ice this makes many things far less readable.

mtbc:
Despite the many files changed I expect that the risk of conflicts with other PRs is low, though if you prefer I could move the `Graphs.ice` commit to a separate PR that could stay open for now. Feel free to take more time to possibly get time to think about it first, if you like, or merge and adjust afterward if necessary.

mtbc:
Piping builds through `sed '/deprecation/{N;N;d;}'` can help. :smiley:

joshmoore:
Shall we teach ant to do that? :smile:

mtbc:
Anything further to do on this?

joshmoore:
Merging for m4. Thanks, @mtbc. I still would like to find a better alternative for dealing with Ice but might as well have this in.

mtbc:
Sorry if you had to hold your nose as you pushed the button!

joshmoore:
I did! But that's alright. Nothing you could do about it.

PR #3324, opened 2015-01-08
reimplement PublicRepositoryI.deletePaths using Delete2

mtbc:
File deletion should still work. Make sure the code changes look sensible and that the integration tests still pass.

--no-rebase as Delete2 is not available in 5.0


ximenesuk:
There didn't seem to be any problem deleting OriginalFiles directly or indirectly. Good to merge.

PR #3325, opened 2015-01-08
Reload MapAnnotations with NamedValue list initialized (temporary PR)

dominikl:
This is just a temporary fix for loading MapAnnotations (with the NamedValue list initialized).
--on-hold  (has to be done properly in an other PR)

### DIFF
--- a/components/server/src/ome/logic/MetadataImpl.java
+++ b/components/server/src/ome/logic/MetadataImpl.java
@@ -1,5 +1,5 @@
 /*
- *   Copyright (C) 2006-2014 University of Dundee & Open Microscopy Environment.
+ *   Copyright (C) 2006-2015 University of Dundee & Open Microscopy Environment.
  *   All rights reserved.
  *
  *   Use is subject to license terms supplied in LICENSE.txt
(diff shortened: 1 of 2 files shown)


joshmoore:
To copy the comment from https://github.com/dominikl/openmicroscopy/commit/a43d1bbc764ff93fa80c146f0b9e855fbd0b77a1 :+1: 

There are a few options/alternatives I can think of for this:

First, do this via a group lookup, i.e. where a.id in (:ids). This would require collecting all the map annotations, loading if the set is non-empty, and then finding which is which for the reload, but would likely be more performant if a number of annotations are being loaded. This is more code but should be straight-forward.
My second thought is not possible: http://stackoverflow.com/questions/9548317/eager-loading-of-sub-class-collections-in-hql
Another option would be to simply always eagerly fetch maps, which would be a change to the code-gen. /cc @jburel @mtbc @will-moore @aleksandra-tarkowska

dominikl:
I'd prefer option 3. But I'm not sure if we have maps somewhere where this could be a problem.

jburel:
Option will be certainly better but we also to be careful not to "load too much". 
Annotations are loaded every time a new object is selected.

jburel:
Added label so it is visually easy for person preparing the standup that the PR is not included.

joshmoore:
Agreed. That would be the biggest negative of gh-3334. We've never really done eager loading before. If we start to feel that the impact would be too great, I can try [fetch profiles](https://docs.jboss.org/hibernate/orm/3.6/reference/en-US/html/performance.html#performance-fetching-profiles) (which we've also never used before).

joshmoore:
Removed `--exclude` from the description so we don't inadvertently leave it. That being said, we should probably collapse the discussion here to one PR. Better to close?

dominikl:
As the java MapAnnotation integration tests passed with PR #3334 , I close this PR.

PR #3326, opened 2015-01-09
Apache 2.4 mod_proxy_fcgi template generation

manics:
Support for generating a configuration file for Apache 2.4 with mod_proxy_fcgi (should be installed by default).
https://trello.com/c/ROD6R0L7/180-omero-web-custom-prefix
https://trello.com/c/GLXxdWyl/128-server-installation-doc-and-supported-versions

Testing: `dist/bin/omero web config apache-fcgi`
with the usual config variations of `force_script_name` and `static_url`

Also contains web cli tests from @sbesson

Also test maintenance.html in OMERO.server/var/ with/without a prefix when using Nginx.

Additional changes (2015-01-15):
- Nginx maintenance page removed and replaced with 502 error handler (should respect presence of a /prefix)
- `omero web config nginx` changed to be more consistent with apache
  - `nginx` now generates the system config file instead of a user config
  - `--system` flag removed
  - `nginx-development` replaces the previous (non-system) nginx target

--no-rebase


etc/apache-mod_proxy_fcgi.conf.template
```diff
@@ -0,0 +1,77 @@
+###
```
aleksandra-tarkowska: As a number of templates started growing and perhaps we will add more and more, does it make sense to put them in etc/templates or similar?

joshmoore: :+1: on a subdir from me, @aleksandra-tarkowska

components/tools/OmeroPy/src/omero/plugins/web.py
```diff
@@ -212,6 +214,28 @@ def config(self, args):
...
+                        679, "Apache mod_proxy_fcgi requires fastcgi-tcp")
+                fastcgi_external = '%s:%s' % (
+                    settings.APPLICATION_SERVER_HOST,
+                    settings.APPLICATION_SERVER_PORT)
+                d["FASTCGI_EXTERNAL"] = fastcgi_external
+
+                # Forcing a prefix for the entire server makes it easier to
+                # setup the proxy redirects
+                if d["FORCE_SCRIPT_NAME"] == '/':
+                    d["FORCE_SCRIPT_NAME"] = '/omero'
```
aleksandra-tarkowska: sure but you still should be able to deploy on http://host/webclient, right?

manics: You'd need separate `RewriteRule` and/or `ProxyPass` statements for each prefix (webclient, webgateway, webadmin, etc), otherwise requests for `/static` will be treated as a cgi request.

etc/apache-mod_proxy_fcgi.conf.template
manics: This or a separate PR?

joshmoore: I'd probably just do it here unless you think it's going to be disruptive. Only impact should be on `web.py`, though, no?

aleksandra-tarkowska: And documentation ! ;-)

sbesson: I have a local branch refactoring `web.py` in order to add CLI unit tests for `web config`. happy to do that as part of that branch and/or open a PR against this one.

sbesson: Note we also have a `robot.template` so this plugin would likely be affected by a `templates` subdirectory.

rleigh-dundee:
Functionally, this all appears to work OK for me.  Tested on MacOS 10.10 with homebrew apache.  I do think that the setup documentation could document:

- Which apache modules to enable (mod_rewrite, mod_proxy and mod_proxy_fcgi I think)
- Where to place the generated apache config fragment
- How to source the fragment (and how to set it up for a virtualhost rather than the main site for a shared webserver?)

As a non-web-expert this was stuff which I could have found useful to get it working.

components/tools/OmeroPy/src/omero/plugins/web.py
manics: Actually it might be possible to use `!(static)` in the RewriteRule, I'll give it a go

manics:
Rewritten to allow an empty prefix e.g. `http://localhost/webclient/`
Note I've deliberately taken out the redirect from `/` to `$force_script_name` if set, since if a sysadmin has set a prefix this is probably because there are other applications running on the same server so we shouldn't redirect the root (arguably this could also be removed from the standard apache fastcgi conf too).

manics:
I'll update the documentation once everyone's happy with this PR.

components/tools/OmeroPy/test/unit/clitest/test_web.py
```diff
@@ -0,0 +1,134 @@
...
+                                static_prefix + '/static/', raising=False)
+        self.args += ["config", "nginx"]
+        if system:
+            self.args += ["--system"]
+        if http:
+            self.args += ["--http", str(http)]
+        self.add_templates_dir()
+        self.cli.invoke(self.args, strict=True)
+        o, e = capsys.readouterr()
+        print o
```
sbesson: Oops. This was used for debugging and can be removed.

manics:
@sbesson I'm done, do you have anything else to add?

sbesson:
No I think unit tests are in a better place now. I will be leaving reviewers to propose suggestion/enhancements if applicable.

manics:
Last two commits ensure the maintenance.html page appears under `$force_script_name` instead of at the web-server's root

rleigh-dundee:
All looks fine to me, and good to merge.

I tested with apache 2.4, so haven't tested the maintenance page.

sbesson:
@aleksandra-tarkowska: you okay with the state of this PR?

etc/templates/nginx.conf.template
```diff
@@ -39,7 +39,7 @@ http {
 
         location %(FORCE_SCRIPT_NAME)s {
             if (-f %(ROOT)s/var/maintenance.html) {
```
aleksandra-tarkowska: Just a small comment, I never managed to get that maintenance working as this is handling 503 rather then 502 Bad Gateway. On the prod servers I have
```
    error_page 502  /500.html;
    
    location ~* ^/(50*\.html)$ {
        root /usr/local/etc/nginx/error;
    }
```
We should command to admin handle it in a global config

sbesson:
From 

```
 $ git grep "config nginx"
components/tools/OmeroPy/src/omero/plugins/web.py:    omero web config nginx --http=8000 >> nginx.conf
docs/hudson/OMERO-homebrew-install.sh:bin/omero web config nginx --http $HTTPPORT > $(bin/brew --prefix omero51)/etc/nginx.conf
docs/install/VM/setup_nginx.sh:/home/omero/OMERO.server/bin/omero web config nginx --http 8080 > /home/omero/OMERO.server/omero-web-nginx.conf
etc/nginx.conf.template:# for inclusion in a system-wide nginx configuration see omero web config nginx --system
```

A couple of other places need to be modified including the development nginx template. Note the `OMERO-homebrew-install.sh` is currently used as part of the daily Homebrew job and thus will need to be modified after the next 5.1.0-m4.
Finally, if we are removing `--system` from the command, should we rename the template?

snoopycrimecop:
Conflicting PR. Removed from build [OME-5.1-merge-push#702](http://ci.openmicroscopy.org/job/OME-5.1-merge-push/702/). See the [console output](http://ci.openmicroscopy.org/job/OME-5.1-merge-push/702/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - docs/hudson/OMERO-homebrew-install.sh

snoopycrimecop:
Conflicting PR. Removed from build [OME-5.1-merge-push#703](http://ci.openmicroscopy.org/job/OME-5.1-merge-push/703/). See the [console output](http://ci.openmicroscopy.org/job/OME-5.1-merge-push/703/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - docs/hudson/OMERO-homebrew-install.sh

manics:
Conflicting reverted and handled in a new PR to avoid rebasing and losing history here:
--depends-on #3354

components/tools/OmeroPy/src/omero/plugins/web.py
```diff
@@ -82,15 +84,18 @@ def _configure(self, parser):
...
+            "  nginx-development: Standalone user-run Nginx server\n"
+            "  apache: Apache 2.2 with mod_fastcgi\n"
+            "  apache-fcgi: Apache 2.4+ with mod_proxy_fcgi\n")
+        config.add_argument("type", choices=(
+            "nginx", "nginx-development", "apache", "apache-fcgi"))
         config.add_argument(
             "--http", type=int,
             help="HTTP port for web server (not fastcgi)")
         config.add_argument(
-            "--system", action="store_true",
```
joshmoore: Considering that this is likely what most folks have, would making this the default and adding a `--development` not be generally less breaking?

manics: `-` not `+` :-)

joshmoore: Yes, but up until now the only way to get "system" functionality was to pass `--system` in which case we can assume there are a number of scripts out there that pass it, right? If so, and we **leave** `--system` making it the default, only users of (to be added) `--development` will be broken.

rleigh-dundee: Agreed, this should definitely be the default.

In the case of `nginx-development`, could `--development` be inferred?  If that's true, is this option needed at all?

manics: @joshmoore I took out all the flags because they were inconsistent- they only worked for nginx and not apache. I suspect few people would script the omero config web-server config with every upgrade since it requires a mix of commands being run by the `omero` and `root` system users (nginx conf directory shouldn't be writeable by `omero`). If you think people are likely to have it scripted I could add `--system` as a noop.

joshmoore: You'll know better than I. Was mostly just the hope of forcing the fewest number of people to change their installation scripts.

manics: How about if I add it back and make it output a deprecated message?

manics:
Three commits added... and hopefully that's it.

components/tools/OmeroPy/src/omero/plugins/web.py
joshmoore: Thanks, @manics. Certainly works for me.

etc/templates/error/502.html
```diff
@@ -0,0 +1,81 @@
...
+        </div>
+
+        <div id="login" class="standard_form inlined">
+	        <p>Site is currently down for maintenance. Please try again later.</p>
+
+        </div>
+
+        <div id="login-footer">
+	        <p>
+		        &copy; 2007-2014 University of Dundee &amp; Open Microscopy Environment<br>
```
sbesson: `2014` -> `2015`

aleksandra-tarkowska: @sbesson is there any way we could maintain current year in static files?

aleksandra-tarkowska: @joshmoore @sbesson is there any way we could maintain current year in static files?

sbesson: Sorry what do you mean? As a `php` or `script` line in the HTML itself that would return the current year or as a automated way to update the source code?

mtbc: We have something else (in web somewhere?) that takes the year of when the build was run, I think?

aleksandra-tarkowska: I mean if ant build could replace it some how? JavaScript will be an option as well

aleksandra-tarkowska: @mtbc in web is set by Python

sbesson:
Tested the deployment of OMERO.server-5.1.0-m3-597-adf4b70-ice35-b387 with ports prefixes by `bin/omero admin ports --prefix 1` under OSX 10.9.5 with nginx 1.6.2:
- using `bin/omero web config nginx-development > etc/nginx`
- using `bin/omero web config nginx > etc/nginx > /usr/local/etc/nginx/conf.d/omero.conf`

In both cases:
- tested the Web deployment with and without a prefix (`omero.web.force_script_name`/`omero.web.static_url`)
- tested the maintenance page by calling `bin/omero web stop`

From the `nginx` point of view, everything worked as expected.

etc/templates/error/502.html
sbesson: More generally, the question is should this page reflect the build date or the current date? Assuming OMERO.web is substituting the current date year, I would be inclined to use the same strategy.

aleksandra-tarkowska: right that make sense, sorted in https://github.com/manics/openmicroscopy/pull/8

mtbc: It makes no sense at all for an intellectual property claim? (Build date doesn't much either, but current date is even more obviously wrong.)

aleksandra-tarkowska: In OMERO.web we display current year for when it's running. If that policy has to change then that should be discussed with everyone and change globally

manics:
Last commit from @aleksandra-tarkowska

etc/templates/error/502.html
mtbc: Okay, I'll add a trello card somewhere; can handle outwith this PR.

aleksandra-tarkowska: My apology I lied to you guys. In web we do ```build_year``` that's when my original question came from ;-) . I will update that to 2015 and that's all. Sorry for the confusion

rleigh-dundee:
Everything works fine with apache-fcgi, not that the latest changes affect it really.

joshmoore:
Merging with 2 RFEs on devx:
 * general copyright handling 
 * 502s in apache

Note: https://github.com/manics/openmicroscopy/compare/openmicroscopy:ddac350...54698564bac482fb949405ff7c3af05d7e8669a7?diff=unified&name=apache_mod_proxy_fcgi&w=1 from @aleksandra-tarkowska is untested, but considering the amount of change there the impact should minimal. A follow-up PR can adjust based on what gets decided re: copyrights.

mtbc:
Well, at least copyrights aren't current-day after all. :smiley:

PR #3328, opened 2015-01-10
fix #12701: combine facade requests even for different target types

mtbc:
A weakness in the graphs reimplementation's API compatibility layer prevented Insight users from being able to move sets of images between groups where that set included both a MIF and an image script-generated from the MIF. For testing, projecting from `Beta Catenin.lif`, then moving the lot after selecting the individual images, should now work in both Insight and web, and other moves and deletes should continue to work.

Fixes http://trac.openmicroscopy.org/ome/ticket/12701.
--no-rebase


pwalczysko:
Did not work for me.

mtbc:
This works fine when moving into orphaned images but not when moving into a dataset! I shall investigate.

mtbc:
(Again, it works fine in web!)

pwalczysko:

PR #3330, opened 2015-01-12
Use 5.0 versions of OMERO and Homebrew formulas

sbesson:
See also https://github.com/sbesson/openmicroscopy/commit/4321538e4f0abb0db6ca26c54a3ca9911afac526. This PR modifies the `OMERO-homebrew-install.sh` installation script under `dev_5_0` to use the new formulas introduced by https://github.com/ome/homebrew-alt/pull/70.

--no-rebase

### DIFF
--- a/docs/hudson/OMERO-homebrew-install.sh
+++ b/docs/hudson/OMERO-homebrew-install.sh
@@ -76,30 +76,30 @@ cd /usr/local
 ###################################################################
 
 # Install Bio-Formats
-bin/brew install bioformats
-VERBOSE=1 bin/brew test bioformats
+bin/brew install bioformats50
+VERBOSE=1 bin/brew test bioformats50
 
 ###################################################################
 # OMERO installation
 ###################################################################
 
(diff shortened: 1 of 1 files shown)


sbesson:
See https://ci.openmicroscopy.org/job/OME-5.0-merge-homebrew/666/

joshmoore:
So looks good and is what I had in mind. Guess the only worry is that now nothing is testing/verifying the non-versioned formulae. Do we just live with that for the moment?

sbesson:
Agreed, I see 2 ways to tackle this:
- either make the formulas names configurable and extend the testing matrix of the daily job
- or at the `homebrew-alt` level add some checkpoint (Travis?) ensuring `omero` and `omero5x` only differ by the class definition line

joshmoore:
The latter seems sufficient. And then when we move to omero51 etc. travis only needs to be bumped.

sbesson:
See last commits on https://github.com/ome/homebrew-alt/pull/70

PR #3332, opened 2015-01-12
Insight progress bar fix

dominikl:
Fixes a bug introduced with PR #3320 , see also [Ticket 12699](http://trac.openmicroscopy.org.uk/ome/ticket/12699)
Test: Open a Dataset with several Images (or a Plate with several Wells), check that the progress bar (status bar in central panel) progresses up to 100% and then disappears again. (Before: Progressbar stopped at 50%).


jburel:
works as expected.

PR #3334, opened 2015-01-12
Enable eager fetching of maps

joshmoore:
Since the MapField properties are of little use
if the map itself has not been loaded, they are
now actively loaded. Queries need not include a
"left outer join fetch" clause.

Testing:
 * integration tests should be green.
 * some of @dominikl's work should pass **without** gh-3325

### DIFF
--- a/components/tools/OmeroPy/test/integration/test_model51.py
+++ b/components/tools/OmeroPy/test/integration/test_model51.py
@@ -101,13 +101,12 @@ class TestModel51(lib.ITest):
         m = g.getConfigAsMap()
         assert m["foo"] == "bar"
 
-    def assertMapAnnotations(self, anns, length=2):
+    def assertMapAnnotation(self, anns, mid):
         m = None
-        assert length == len(anns)
-        for idx in range(length):
-            a = anns[idx]
+        for a in anns:
             if isinstance(a, omero.model.MapAnnotationI):
-                m = anns[idx]
+                if a.id.val == mid:
+                    m = a
         assert m
         assert "foo" == m.getMapValue()[0].name
         assert "bar" == m.getMapValue()[0].value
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/test/integration/test_model51.py
```diff
@@ -100,3 +100,55 @@ def testAsMapMethod(self):
...
+
+    def testMapEagerFetch(self):
+        m = omero.model.MapAnnotationI()
+        m.setMapValue(
+            [NV("foo", "bar")]
+        )
+        m = self.update.saveAndReturnObject(m)
+        anns = self.query.findAllByQuery(
+            "select m from MapAnnotation m ",
+            None)
```
mtbc: Would it be a problem here if some other integration testing had left other map annotations around?

mtbc:
Looks like next integration test build indeed ought not have #3325 included.

components/tools/OmeroPy/test/integration/test_model51.py
joshmoore: This should be a new user in a new group, but I'm happy to add filtering. I wasn't quite sure of the status of @sbesson's re-working to know what the new "Right Way" was (to be).

mtbc: Ah, yes, group filtering, good point. I'm satisfied anyway. :smiley:

sbesson: See https://github.com/openmicroscopy/openmicroscopy/pull/3318/files#diff-cc72fc55d66d83492366c49e1420fff4R699 for instance. With the new `ITest` class, `self.client` will be shared by all test methods in the class. If the test has any assumption, it is possible to create `new_user_and_client` or `new_client` using the library methods and use this client instead.

mtbc:
Presently we have a test failure in https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/test.integration.test_model51/TestModel51/testMapEagerFetch/

joshmoore:
Pushed.

mtbc:
Well, that seems to have helped the tests at least!

will-moore:
Sorry - forgot to review this yesterday. It's certainly working for me in my current Map Annotation PR and gateway tests. Good to merge from me.

joshmoore:
Merging. If we see any issues with slow client performance, we may have to re-evaluate.

joshmoore:
--no-rebase

PR #3337, opened 2015-01-13
OMERO.insight Java Application Bundler task for Mac OS X

chris-allan:
This pull request adds support for the Java Application Bundler (https://java.net/projects/appbundler) and its open source fork (https://bitbucket.org/infinitekind/appbundler). This bundler and corresponding launcher are intended, at least as far as we're concerned, as a transition point for Java 1.7+ on Mac OS X.

The high level `zips` target can be run to trigger the execution of the Java Application Bundler specific `dist-osx-appbundler` task as follows:

```
./build.py -f components/insight/build.xml zips
```

You will likely find that this is best run after a current build run via `./build.py clean build`. The new `OMERO.insight.app` is then available in the following subdirectory:

```
components/insight/OUT/dist/appbundler
```

Due to the fact that execution sets the "current working directory" to the user's home directory initial testing can be performed as follows:

```
components/insight/OUT/dist/appbundler/OMERO.insight.app/Contents/MacOS/OMERO.insight \
    container.xml `pwd`/components/insight/OUT/app
```

If you have errors, launcher debugging can be turned on with an environment variable (https://java.net/projects/appbundler/lists/dev/archive/2012-11/message/2):

```
export JAVA_LAUNCHER_VERBOSE=1
```

Work will have to be done on the following before this can be merged:

* Resolving issues with the way that OMERO.insight locates configuration files

/cc @jburel, @rleigh-dundee


joshmoore:
@chris-allan : before merging, I'd propose we force push away the jar and either use some existing maven repository or copy it to our "manual" repository.

joshmoore:
Having looked at https://github.com/chris-allan/openmicroscopy/commit/92ebe4c9e661d6625490f21400b831511b209b60 now, perhaps maybe the way to go is to set up a job that automatically builds and pushes to http://artifacts.openmicroscopy.org /cc @sbesson

sbesson:
With regard to packaging and assuming this new app bundler will eventually be the long term solution for the Insight application under OS X, a suggestion would be to:
- rename the old OMERO.insight e.g. as OMERO.insight_legacy (.app folder, zips etc etc)
- bundle both legacy and non-legacy apps in the OMERO.clients (and deprecate legacy in a later release)
- do we need the same strategy for editor and importer?
- do we backport openGL removal on this line to reduce the build time, artifact list?

joshmoore:
Perhaps `OMERO.insight_Java6` ?

rleigh-dundee:
Bundling the new and legacy .app bundles is convenient, but these are quite large due to each having a complete copy of all the jars, so having them separate does have its advantages.

sbesson:
Separate OMERO.insight bundles should also be available via the downloads page. The only place where we would have doubled size are the OMERO.clients for mac. As noted above this is a transition as we expect to drop the `_Java6` bundle eventually.
Would be good to have @gusferguson opinion on this, the major questions we need to answer in terms of packaging is:
- what is the less confusing for our users?
- how to properly document this (PR against ome-help?)

gusferguson:
I would **really** like to see the Editor and Importer disappear from the bundles - they are very confusing for users.
We do 32 and 64 bit versions for Windows - so I presume 2 versions for Mac will be much the same overhead. As long as it is clear what is what the user probably won't care about size.
On the _legacy or _Java6 issue - I feel the _Java6 is much more intuitive and self-evident than _legacy - so my vote would go there.
Documentation will be fun - just have to see what it looks like and try and work out a way to make it obvious what a particular user needs.

jburel:
@gusferguson: editor has been removed for 5.1. we will have to keep it for 5.0.x

chris-allan:
Two `.app`'s, `OMERO.insight_Java6.app` and `OMERO.insight_Java7+.app`, will now be included in the client ZIPs. We still have the issue of configuration file loading but that will have to be taken care of by @jburel.

jburel:
great. I will start working on it tomorrow. Final adjustment on imageJ work prior to demo tomorrow.

joshmoore:
(travis kicked)

mtbc:
Ready for review again?

jburel:
@mtbc: suggestions have been implemented. investigating solutions to get the config files picking up. Probably no need to list tomorrow

mtbc:
Thank you, @jburel. Good luck with it.

jburel:
PR opened against this PR.
Maybe for 5.1, we may want to include (e.g. under Contents/Resources) the config directory.
This will require an adjustment to dist.xml

jburel:
@mtbc: to be listed tomorrow

rleigh-dundee:
Test platform: MacOSX 10.10 with Java 1.7.0_51.  The _Java6 app (untested since I don't have Java6) warns than I need a Java6 VM.  The new app works just fine, starts up and lets me log in, do stuff etc.  However, there is this annoying UI glitch:

screen shot 2015-01-21 at 09 34 56

which goes away on resize.  Maybe Java7 is not firing a resize event so it needs doing manually when the window is mapped?

screen shot 2015-01-21 at 09 35 09

dominikl:
Built with Java 8, ran with Java 7, didn't notice any problems.
@rleigh-dundee Looks like to be a similar problem to #3174 (just had to make another call to pack() to fix this). For reproducing the problem in #3174 I needed exactly this setup, compile with Java 8, run with Java 7. Unfortunately this is not sufficient to reproduce the problem with the server dialog, can't see that, looks normal for me.

jburel:
The UI issue will be covered in another PR.

rleigh-dundee:
From the build POV, looks good to merge.  The cosmetic issues (application suffixes, UI) do need further work.

joshmoore:
Waiting on `appbundler` modifications from @sbesson.

manics:
`$ open components/insight/OUT/dist/OMERO.insight_Java7+.app` fails:
screen shot 2015-01-21 at 15 08 40

`$ open components/insight/OUT/dist/OMERO.insight_Java7+.app --args container.xml $PWD/components/insight/OUT/app` works

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.9.5
BuildVersion:   13F34
$ javac -version
javac 1.8.0_31
```

sbesson:
Jar uploaded to our artifactory and accessible to Ivy via

```
<dependency org="ome" name="appbundler" rev="1.0.0">
    <artifact name="appbundler" ext="jar"/>
</dependency>
```

manics:
However the CI artifact works:
https://ci.openmicroscopy.org/job/OMERO-5.0-merge-build/lastSuccessfulBuild/ICE=3.5,jdk=8_LATEST,label=octopus/artifact/src/target/OMERO.clients-5.0.6-119-0d5f460-ice35-b209.mac.zip

jburel:
Build locally with java 1.7 no problem, will try with 1.8

chris-allan:
That is expected @manics. The configuration files are not copied in to the `dist` directory so those `.app`'s will not work directly from there.

jburel:
The idea (cf. a previous comment) is to move the config folder under Contents/Resources.
This should have been done a while ago.

jburel:
for UI issue see https://github.com/openmicroscopy/openmicroscopy/pull/3375
Fixed in develop but not rebased.

chris-allan:
This is going to require some more reworking. In adding the above three commits I've noticed that OMERO.editor and OMERO.importer `.app`'s have been broken by the commits from @jburel above. This is because each of these is delineated by a command line option to specify the `container.xml` instance to use.

We'll need some time tomorrow to work through all of this.

jburel:
@chris-allan: Will have a look either today or over the weekend

jburel:
PR opened https://github.com/chris-allan/openmicroscopy/pull/9

chris-allan:
Commits from @jburel now included which fix issues with picking up the correct `container.xml` and with the use of command line options. Just addressing Travis build failures and then this will be ready to be tested.

jburel:
To test the last commits:
 * Check that you can start the importer and editor.

joshmoore:
On 10.10, all three Java7+ apps launch for me.

sbesson:
The thress Java7+ apps also work on 10.9.5 for me

pwalczysko:
Tested OMERO.clients.zip on ice 3.3, 3.4 and 3.5 on the octopus build. 
Mac OS X 10.8
java version "1.7.0_51"
Java(TM) SE Runtime Environment (build 1.7.0_51-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.51-b03, mixed mode)

All passed fine (6 clients per ice version tested, 2x ed, 2x imp, 2x insight each).

joshmoore:
10.6 / Java_6: :ok: (all three)
10.6 / Java_7+: double click instantly exists; no warning or error.

rleigh-dundee:
Also tested all three Java 7 applications on 10.10 and it's working fine.  If we have the split into 10.6 and 10.7+ zips, that will also remove the need for the special application suffixes.

chris-allan:
Force pushed this without JARs added to `lib/repository` and resolution of artifacts thanks to work from @sbesson.

```
 ______________
< Artifactory! >
 --------------
        \   ^__^
         \  (**)\_______
            (__)\       )\/\
             U  ||----w |
                ||     ||
```

sbesson:
Thanks guys. Merging

sbesson:
--rebased-to #3417

PR #3338, opened 2015-01-13
Mapannotations insight

dominikl:
Adds MapAnnotations to Insight.


components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/MapAnnotationsComponent.java
```diff
@@ -0,0 +1,383 @@
+package org.openmicroscopy.shoola.agents.metadata.editor;
```
mtbc: Copyright header.

will-moore:
If you remove the last row from a Map Annotation, should probably delete the annotation? (still need to do this in web).
Don't allow to remove rows if canEdit() is false (although server allows it - @joshmoore is looking into this).
Need to separate the different map annotations a bit, and give some indication of who added what.
Following @aleksandra-tarkowska suggestions, the Add Key / Add Value row in web appears different (see below)
When entering Key/Value, Tab or Enter should move you onto the next field (creating a new row if needed).

screen shot 2015-01-16 at 14 44 13

mtbc:
May not want to delete it automatically: the map annotation may itself have a file attached or somesuch and then the file gets separated from what the map annotation annotates.

will-moore:
@mtbc Any annotations not added by the client are ignored by the client on a number of objects, E.g ROIS, rating annotations, tags, comments etc. The clients assume that since the client added these, they should be able to delete them too.
For other map annotations (that don't have the 'client' namespace) these are not edited by the client at-all, so any annotations added to these will be safe (although not displayed in the clients).

will-moore:
@dominikl I added a custom map annotation to http://trout.openmicroscopy.org/merge/webclient/?show=project-1223 (user-3)

dominikl:
Thanks @will-moore !

pwalczysko:
RFE: When Plus is used, it seems to default to addition of new row on the TOP of the table. 
  - select the last row
  - click Plus
  - row at the bottom of the table will be added but
  - click Plus again -> row at the top of the table is added -> surprising

pwalczysko:
Copy to another image - empty table

  - select a row in the table in an image
  - nagivate to another image (which does not have any table yet)
  - in the warning window select Yes, save changes
  - arriving at the new image, the table is not populated at all (see screenshot), nowhere to paste the copied row -> not expected

screen shot 2015-01-22 at 12 25 49

pwalczysko:
When already populated cell is selected in the table, and double-clicked, the editing is in ADD mode, meaning the text which was originally in the cell has not been highlighted and overwritten, instead the newly typed text is being added at the end of the original string -> not in line with Web and not very helpful, because default would be to overwrite the item completely in majority of cases.

dominikl:
@pwalczysko The problems you mentioned should be fixed; maybe not for "Copy to another image - empty table" I couldn't see this. Can you show me an example? Cells are now automatically in 'overwrite' mode, i. e. you switch to a cell and start editing, the existing text will be overwritten. Unfortunately you don't have a visual indication for that (the selected text is not marked as selected and there is no caret).

will-moore:
I can currently edit 'non-client' map annotations. These shouldn't be editable by clients (even though permissions allow it).

screen shot 2015-01-23 at 10 07 59

will-moore:
I see the map annotation saving is now on-hold while editing the table, but it would be nice if it auto-saved when the table lost focus. Otherwise you need to click "Save" at the top of the right panel or you get the "Do you want to save?" pop-up when you try to navigate away, both of which are a bit annoying.

will-moore:
If I click the X button with no rows selected I get a terminal NPE:
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.lang.NullPointerException
	at org.openmicroscopy.shoola.agents.metadata.editor.MapAnnotationsComponent.deleteSelection(MapAnnotationsComponent.java:542)
	at org.openmicroscopy.shoola.agents.metadata.editor.MapAnnotationsComponent.access$800(MapAnnotationsComponent.java:74)
```

will-moore:
Another bug with not saving immediately is if I try to switch user, I get a "really disconnect?" dialog which also tries to save the changes, but if I click Yes, then I get a login screen AND a modal "app will now exit" popup. I am forced to log in and then quit ;)

screen shot 2015-01-23 at 10 26 22

will-moore:
Minor bug - If I am editing another user's Map annotation and click + to add a row, then there are no selected rows, so that subsequent clicks of the + won't add a row to the same Map annotation but will instead switch to adding them to MY Map annotation. Probably best to automatically select the newly added row.

dominikl:
@will-moore How did you manage to get an enabled X button with no row selected?
Wrt "Auto-save on table lost focus", I think this would be inconsistent. E. g. you edit a MapAnnotation switch to another one, the changes will be saved (without confirmation); now you edit the MapAnnotation and switch to another image/project/... then you'll get asked if you want to save the changes (which I think is quite good, because this gives the possibilty to 'undo' your modifications).

will-moore:
@dominikl I never said the X was enabled when I clicked it ;).
I think saving without confirmation is OK for now (that's what we're doing in web).
Alternative is to add a Save icon to the map annotation toolbar. But I just don't think users will think to click the Save button at the top of the panel. Even when they know they have to, it's still kinda inconvenient, compared to the small number of times anyone will want to edit without saving.
I guess an "Undo" functionality might be nice, but probably more work than is worth it just now.

jburel:
* The "added by" should be added only if not yours. I think that what we agreed during last meeting.
* The green plus button is not great. @gusferguson did you have a better one
* Starting with an empty map, click on +, then click on "Add key". It is very difficult to see what I am editing since I have 2 white rows w/o border for editing box
* Edit key, use tab key to move to value cell does not work.
* Single click vs double click for edit. I think we decided to go for single click
* Enter a value. Press enter, a new row is created but it is not automatically editable.
* Click on the empty "Add key, Add value" row.  Select copy, then paste rows with the default text are added above the selected row. Not expected.

jburel:
@dominikl: few issues noticed. I review further if you wish.

dominikl:
The problem is, that the 'edit' mode isn't really indicated. You can click on a cell (single click) and start editing, the cell just doesn't show the caret. That also means, tab (and enter) key is actually working, it moves the focus to the next cell, which you can start editing instantly; just again, there's no indication like a caret. I'll try to fix that, but it's really tricky...
I also noticed another open issue: An empty MapAnnotation is not deleted, but it should get deleted.

jburel:
@dominikl: edit mode is tricky I can sympathise with that..

dominikl:
Fixed some bugs; empty MapAnnotations are now deleted; tried to improve cell editing a bit (however, I think this is the closest I can get to the Web behaviour).
Some notes wrt editing: Cells are editable on single click (= selection), but the caret is only shown when double clicked. This seems to be the default Java cell editing behaviour, couldn't find a way to overwrite this. A cell marks itself as modified as soon as it looses focus (move to different cell, hit enter/tab, ..). If you edit a cell and directly switch to another item (image, dataset, ...) the modification is not saved (same in Web). Pressing enter while in double click editing mode, confirms the change, but does not move the focus to the next cell, i. e. if you start typing again you overwrite the cell's content again.

jburel:
* Delete: when we delete the last row, we should probably have an "empty" default row
* Add map annotation on another user data (no annotation for that use), values are saved but when I moved back to my data: no empty map annotation displayed see screenshot
addrowthenmove

jburel:
Map annotation added by user-4 on project owned by user-11
Nothing displayed when logged in as user-11
user4
user11

jburel:
Delete workflow is a bit strange:
group read-write. User-4 added map annotation to user-11 project
User-11  delete the rows. The border added by user-4 is still visible
Then click elsewhere and the delete dialog pops up.

dominikl:
Review: Check that the problems mentioned in @jburel 's last two comments have been fixed.

Note: The delete dialog will still popup, but now directly after removing the last row of a MapAnnotation; this is because I use the delete 'Activity' to delete the Annotation. Is there another way, kind of a 'silent' delete without notification @jburel ?

jburel:
you could try not to register this delete activity in the UI
i.e. ```DeleteActivityParam.setUIRegister```

dominikl:
Thanks @jburel , that did the trick. There's still a tiny notification in the bottom right corner, but I think that's ok.

jburel:
We were planning to remove that one anyway. So if we want to go ahead that's fine.

dominikl:
Just fixed an issue discovered by @pwalczysko yesterday. Removed MapAnnotations from metadata panel, if multiple items are selected (not supported, yet).
Is that 'good to merge' or shall we have another review?

jburel:
Let's have a final check tomorrow

jburel:
Still few UI things to fix before the release, but it certainly usable and getting closer
I will add comment to Trello card.
Merging

PR #3339, opened 2015-01-13
Don't use Decimal for percent_fraction formatting

will-moore:
This removes the use of Decimal for formatting of percent fraction values (from #3309) for ND filter and attenuation.

Formatting should be as follows:
```
100 -> "100"
80.5 -> "81"
20 -> "20"
1.0 -> "1"
0.5 -> "0.5"
0.25 -> "0.3"
0 -> "0.0"
```

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/forms.py
@@ -3,7 +3,7 @@
 # 
 # 
 # 
-# Copyright (c) 2008-2011 University of Dundee.
+# Copyright (c) 2008-2015 University of Dundee.
 # 
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -39,6 +39,7 @@ from custom_forms import MetadataModelChoiceField, \
 from omeroweb.webadmin.custom_forms import ExperimenterModelChoiceField, \
                         ExperimenterModelMultipleChoiceField, \
                         GroupModelMultipleChoiceField, GroupModelChoiceField
+from omeroweb.webclient.webclient_utils import formatPercentFraction
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


chris-allan:
Technically looks great and is much improved over the #3309 implementation.

joshmoore:
@gusferguson : look ok to you?

gusferguson:
@joshmoore 
Looks fine from my POV.

chris-allan:
The only additional suggestion I can think of right now would be to put this in a method or function called `format_for_display()` or similar so that you can use the same logic wherever it's required. That would also allow you to set up some unit tests to explicitly check for and implicitly document the desired numeric to string representation translation. But, that's nitpicking. Not required for a merge.

dominikl:
Viewing alternate.ome (read-annotate-1/user-3/0801) I noticed two flaws:
- Attenuation is shown as 2 digits, but with different color (black instead of grey like the other properties)
- Rounding issues for Transmittance property

screen shot 2015-01-15 at 10 14 53

chris-allan:
Then it's probably time to add the utility method outlined above @dominikl and apply it more thoroughly. Every floating point number has the possibility of displaying representation error (https://docs.python.org/2/tutorial/floatingpoint.html#representation-error) when converting to a string.

will-moore:
Fixed a few things here.
I created a utils method for formatting percent fraction values.
This is in webclient_utils.py and is being tested by a new unit test.
The Transmittance field was not being formatted as a percent fraction.
Also fixed the appearance of the Attenuation field (wasn't being disabled before).

Unit test run like:
```
OmeroWeb wmoore$ ./setup.py test -t test/unit/test_util.py
....
test/unit/test_util.py
...
============== 1 passed in 0.01 seconds ===========
```

screen shot 2015-01-15 at 11 48 25

chris-allan:
Looks great @will-moore. As long as Travis is happy, I am.

mtbc:
Travis is happy. :smiley:

PR #3340, opened 2015-01-14
Properties cleanup

aleksandra-tarkowska:
This PR centralize common client properties.

 - ```omero.web.scripts_to_ignore``` list of scripts shouldn't be displayed in the client. It became:
    - omero.client.scripts_to_ignore
 - ```omero.web.ui.tree.orphaned``` allows to customize name and description of the Orphaned Images container. It became:
    - omero.client.ui.tree.orphans.name
    - omero.client.ui.tree.orphans.description
 - ```omero.web.ui.menu.dropdown``` allows to control which part of the User/Group menu should be displayed. It became:
    - omero.client.ui.menu.dropdown.leaders
    - omero.client.ui.menu.dropdown.leaders.enabled
    - omero.client.ui.menu.dropdown.colleagues
    - omero.client.ui.menu.dropdown.colleagues.enabled
    - omero.client.ui.menu.dropdown.everyone
    - omero.client.ui.menu.dropdown.everyone.enabled

To test use ```bin/omero config set ...``` to alter UI.

cc: @dominikl @jburel could you apply them in Insight in corresponding PR?
If you need any help please let me know

--no-rebase

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -592,17 +592,6 @@ CUSTOM_SETTINGS_MAPPINGS = {
          leave_none_unset_int,
          ("Configuration options for the viewer. -1: zoom in fully,"
           " 0: zoom out fully, unset: zoom to fit window")],
-    "omero.web.scripts_to_ignore":
-        ["SCRIPTS_TO_IGNORE",
-         ('["/omero/figure_scripts/Movie_Figure.py", '
-          '"/omero/figure_scripts/Split_View_Figure.py", '
-          '"/omero/figure_scripts/Thumbnail_Figure.py", '
-          '"/omero/figure_scripts/ROI_Split_Figure.py", '
-          ' "/omero/export_scripts/Make_Movie.py",'
-          '"/omero/setup_scripts/FLIM_initialise.py", '
-          '"/omero/import_scripts/Populate_ROI.py"]'),
-         parse_paths,
-         None],
 
 }
 
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -606,14 +606,6 @@ DEVELOPMENT_SETTINGS_MAPPINGS = {
     # "DESCRIPTION":"This is a virtual container with orphaned images. These
     # images are not linked anywhere. Just drag them to the selected
     # container."}'
-    "omero.web.ui.tree.orphaned":
-        ["UI_TREE_ORPHANED",
-         ('{"NAME":"Orphaned images",'
-          ' "DESCRIPTION":"This is a virtual container with orphaned images.'
-          ' These images are not linked anywhere. Just drag them to the'
-          ' selected container."}'),
-         json.loads,
-         None],
     "omero.web.webstart_admins_only":
         ["WEBSTART_ADMINS_ONLY", "false", parse_boolean, None],
     "omero.web.webadmin.enable_email":
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -240,8 +240,11 @@ CUSTOM_HOST = CUSTOM_SETTINGS.get("Ice.Default.Host", "localhost")
 INTERNAL_SETTINGS_MAPPING = {
     "omero.qa.feedback":
         ["FEEDBACK_URL", "http://qa.openmicroscopy.org.uk", str, None],
-    "omero.upgrades.url":
-        ["UPGRADES_URL", "http://upgrade.openmicroscopy.org.uk/", str, None],
+    "omero.upgrade-check.enabled":
+        ["UPGRADE_CHECK",
+         "true",
+         parse_boolean,
+         "A boolean that turns on/off upgrade check performed by OMERO.web."],
 
     # Allowed hosts:
     # https://docs.djangoproject.com/en/1.5/ref/settings/#allowed-hosts
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -240,11 +240,8 @@ CUSTOM_HOST = CUSTOM_SETTINGS.get("Ice.Default.Host", "localhost")
 INTERNAL_SETTINGS_MAPPING = {
     "omero.qa.feedback":
         ["FEEDBACK_URL", "http://qa.openmicroscopy.org.uk", str, None],
-    "omero.web.upgrade-check.enabled":
-        ["UPGRADE_CHECK",
-         "true",
-         parse_boolean,
-         "A boolean that turns on/off upgrade check performed by OMERO.web."],
+    "omero.web.upgrade-check.url":
+        ["UPGRADES_URL", "http://upgrade.openmicroscopy.org.uk/", str, None],
 
     # Allowed hosts:
     # https://docs.djangoproject.com/en/1.5/ref/settings/#allowed-hosts
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -554,12 +554,6 @@ CUSTOM_SETTINGS_MAPPINGS = {
           "E.g. ``'[[\"Webtest\", \"webtest_index\"], [\"Homepage\","
           " \"http://...\", {\"title\": \"Homepage\", \"target\": \"new\"}"
           " ]]'``")],
-    "omero.web.ui.menu.dropdown":
-        ["UI_MENU_DROPDOWN",
-         ('{"LEADERS": "Owners", "COLLEAGUES": "Members", '
-          '"ALL": "All members"}'),
-         json.loads,
-         "Shows/hides users in dropdown menu."],
     "omero.web.ui.right_plugins":
         ["RIGHT_PLUGINS",
          ('[["Acquisition",'
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -240,8 +240,8 @@ CUSTOM_HOST = CUSTOM_SETTINGS.get("Ice.Default.Host", "localhost")
 INTERNAL_SETTINGS_MAPPING = {
     "omero.qa.feedback":
         ["FEEDBACK_URL", "http://qa.openmicroscopy.org.uk", str, None],
-    "omero.web.upgrade-check.url":
-        ["UPGRADES_URL", "http://upgrade.openmicroscopy.org.uk/", str, None],
+    "omero.web.upgrades.url":
+        ["UPGRADES_URL", None, leave_none_unset, None],
 
     # Allowed hosts:
     # https://docs.djangoproject.com/en/1.5/ref/settings/#allowed-hosts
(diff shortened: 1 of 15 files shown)


etc/omero.properties
```diff
@@ -638,6 +638,38 @@ omero.ldap.new_user_group=default
 ## omero.ldap.new_user_group=:bean:myNewUserGroupMapperBean
 
 #############################################
+## OMERO client properties:
+#############################################
+
+## Serverside scripts used in IScript service Clients shouldn't display.
```
sbesson: The double `##` means the property description will not be parsed to auto-generate the documentation page. Can we revert to `#` instead?

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -240,8 +240,11 @@ def leave_none_unset_int(s):
 INTERNAL_SETTINGS_MAPPING = {
     "omero.qa.feedback":
         ["FEEDBACK_URL", "http://qa.openmicroscopy.org.uk", str, None],
-    "omero.upgrades.url":
-        ["UPGRADES_URL", "http://upgrade.openmicroscopy.org.uk/", str, None],
+    "omero.upgrade-check.enabled":
```
joshmoore: Could probably use some discussion here. Either it's web in name and use, or it's everything.

etc/omero.properties
```diff
@@ -638,6 +638,38 @@ omero.ldap.new_user_group=default
...
+#############################################
+
+# Serverside scripts used in IScript service Clients shouldn't display.
+omero.client.scripts_to_ignore=/omero/figure_scripts/Movie_Figure.py,\
+/omero/figure_scripts/Split_View_Figure.py,\
+/omero/figure_scripts/Thumbnail_Figure.py,\
+/omero/figure_scripts/ROI_Split_Figure.py,/omero/export_scripts/Make_Movie.py,\
+/omero/setup_scripts/FLIM_initialise.py,/omero/import_scripts/Populate_ROI.py
+
+# Name of the "Orphand images" container located in client tree data manager.
```
mtbc: Orphaned (also below)

etc/omero.properties
```diff
@@ -638,6 +638,38 @@ omero.ldap.new_user_group=default
 ## omero.ldap.new_user_group=:bean:myNewUserGroupMapperBean
 
 #############################################
+## OMERO client properties:
+#############################################
+
+# Serverside scripts used in IScript service Clients shouldn't display.
```
mtbc: maybe hyphenate "Server-side"

etc/omero.properties
```diff
@@ -638,6 +638,38 @@ omero.ldap.new_user_group=default
...
+
+# Name of the "Orphand images" container located in client tree data manager.
+omero.client.ui.tree.orphans.name=Orphaned images
+
+# Description of the "Orphand images" container.
+omero.client.ui.tree.orphans.description=This is a virtual container with orphaned images. These images are not linked anywhere. Just drag them to the selected container.
+
+# Client dropdown menu leader label.
+omero.client.ui.menu.dropdown.leaders=Owners
+## Flag to show/hides leader.
```
mtbc: hide

etc/omero.properties
```diff
@@ -638,6 +638,38 @@ omero.ldap.new_user_group=default
...
+omero.client.scripts_to_ignore=/omero/figure_scripts/Movie_Figure.py,\
+/omero/figure_scripts/Split_View_Figure.py,\
+/omero/figure_scripts/Thumbnail_Figure.py,\
+/omero/figure_scripts/ROI_Split_Figure.py,/omero/export_scripts/Make_Movie.py,\
+/omero/setup_scripts/FLIM_initialise.py,/omero/import_scripts/Populate_ROI.py
+
+# Name of the "Orphaned images" container located in client tree data manager.
+omero.client.ui.tree.orphans.name=Orphaned images
+
+# Description of the "Orphand images" container.
```
mtbc: Orphaned

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -168,6 +170,24 @@ def changeActiveGroup(self, gid): # TODO: should be moved to ISession
             logger.error(traceback.format_exc())
             return False
     
+    def getOrphandContainerSettings(self):
```
mtbc: Orphaned

jburel:
I remember @sbesson indicating that in some institutions the term "datasets" was not suitable. I cannot remember what was the suggested value. 
This should be another good candidate to add to the list.

jburel:
UPGRADE_CHECK could probably go to common.

jburel:
omero.client.scripts_to_ignore=/omero/figure_scripts/Movie_Figure.py,\
+/omero/figure_scripts/Split_View_Figure.py,\
+/omero/figure_scripts/Thumbnail_Figure.py,\
+/omero/figure_scripts/ROI_Split_Figure.py,/omero/export_scripts/Make_Movie.py,\
+/omero/setup_scripts/FLIM_initialise.py,/omero/import_scripts/Populate_ROI.py

The logic to exclude seems to be b/c Wed does not have a customized UI.
The scripts in that list to ignore are probably
```/omero/setup_scripts/FLIM_initialise.py,/omero/import_scripts/Populate_ROI.py```
Also some of the utils scripts should be excluded.

**Comment by aleksandra-tarkowska** [comment]
@jburel UPGRADE_CHECK this is web specific flag in case someone like @chris-allan would like to exclude it only on web. This will not go to common. If you wish to disable UpgadeCheck globally you have already that available. https://www.openmicroscopy.org/site/support/omero5.1/sysadmins/UpgradeCheck.html#disabling.

**Comment by chris-allan** [comment]
Just to expand on this further, OMERO.web performs an explicit version string equivalency check on login. This is in addition to the upgrade check itself. For those of us who deploy OMERO.web separately and regularly, enforced version checking can become very troublesome so it is *very* convenient to be able to disable it.

For clarity the property could be called `omero.web.version-check.enabled`, `omero.web.exact-version-check.enabled` or similar.

**Comment by aleksandra-tarkowska** [comment]
@chris-allan but everywhere we call it "upgrade check", including python library. Version check has a different meaning, we have also check_version to match the client and server version. It may be confusing. I would stay with ```omero.web.upgrade-check.enabled``` unless you really have a strong feeling about that ;-)

**Comment by joshmoore** [comment]
@aleksandra-tarkowska : I think what @chris-allan is getting at is there might need to be more control in that option (and adjustment of the naming). Perhaps something like:
 * `omero.web.upgrade_check=strict (default)` Current behavior, i.e. check and require patch equality
 * `omero.web.upgrade_check=lenient` Log upgrades but do not prevent startup
 * `omero.web.upgrade_check=off` Don't access the URL

**Comment by aleksandra-tarkowska** [comment]
@joshmoore @chris-allan could you create a trello card for that I will handle it in a separate PR then

**Comment by joshmoore** [comment]
Sorry, in talking to Chris, the strict/lenient requirement is a secondary issue that we can handle elsewhere. The biggest impact here (now that I think I understand it) is that this removes the ability to configure the URL which is used for upgrade checks. There are a couple of ways we could get that back here, ranging from straight-forward to more complex. If you'd like to do that for Monday, likely this can be simply tested and go in. Otherwise, we can discuss our options and plan next week.

File: components/tools/OmeroWeb/omeroweb/webadmin/webadmin_utils.py [hunk 8]
```diff
@@ -3,6 +3,9 @@
 import logging
 
 from django.core.validators import validate_email
+from django.conf import settings
+
+from omero.util.upgrade_check import UpgradeCheck
```
chris-allan: Please move this back into `upgradeCheck()`. The scoped import was done to not leak `omero` packages into `webadmin_utils` so that it could be imported without `omero` on the `PYTHONPATH`.

aleksandra-tarkowska:
upgrade check reverted. I just changed name including prefix

sbesson:
A couple of Python files are chmod'ed from `755` to `644`. Is that on purpose?

aleksandra-tarkowska:
Yes, some of them were originally commuted with a wrong permissions.

aleksandra-tarkowska:
according to our doc in empty value disable upgrade check on the server side https://www.openmicroscopy.org/site/support/omero5.1/sysadmins/UpgradeCheck.html#disabling. We cannot have the same name in web because empty ```omero.upgrades.url``` will use default value and do not disable anything. That is a bug.

It is true we do not use ```-``` we use ```_```. I would prefer to call it ```omero.web.upgrades.url``` then. This value modifies the URL used by OMERO.web when user log in. My setting it web login calls:

```
2015-01-19 13:18:27,009 DEBUG [                 omero.util.UpgradeCheck] (proc.06410) run:125 Attempting to connect to http://domain.omero/?os.version=Darwin+Ke...
```

joshmoore:
Like the name. I guess my only additional wish would be that if `omero.web.upgrades.url` is not present, that the value from `omero.upgrades.url` gets used, i.e. that a sysadmin doesn't have to set **both** of them if they are trying to change rather than disable. This may not work with the bug you are talking about, in which case maybe something else is needed.

aleksandra-tarkowska:
@joshmoore that should do the job now. By default server, where connection is acquired will give the right ```omero.upgrades.url```. It will simplify the process for sysadmin if would like to change it. For OMERO.web only setting ```omero.web.upgrades.url``` will always take a priority. I hope that make sense now.

sbesson:
NB: OMERO.web is failing with the same stackstrace as https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-web/lastCompletedBuild/testReport/test.integration.test_csrf/TestCsrf/test_forgot_password/. Excluding to rebuild the merge server/client.

aleksandra-tarkowska:
sorry that should resolve the issue:
```
2015-01-20 10:01:08,372 DEBUG [                 omero.util.UpgradeCheck] (proc.05262) run:125 Attempting to connect to http://upgrade.openmicroscopy.org.uk?os.version=Darwin...
```

aleksandra-tarkowska:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-web/ seems to be back to stable

components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -137,8 +139,13 @@ def login(request):
...
                 if userGroupId in conn.getEventContext().memberOfGroups:
                     request.session['connector'] = connector
-                    upgradeCheck()
-
+                    # UpgradeCheck URL should be loaded from the server or loaded
+                    # omero.web.upgrades.url allows to customize web only
+                    try:
+                        upgrades_url = settings.UPGRADES_URL
+                    except:
+                        upgrades_url = conn.getUpgradesUrl()
```
joshmoore: This should only be called when a user is asked to login, right?

aleksandra-tarkowska: Yes, this is login view.

joshmoore:
Great, thanks @aleksandra-tarkowska. Merging, and leaving Java client side issues to @jburel and @dominikl at which point we may need further slight modifications here.

dominikl:
Sorry in case I missed that in the discussion: Do we have a policy that says that every property relevant for the clients starts with omero.client.*? I just wonder if I can rely on that on the client side.

aleksandra-tarkowska:
@dominikl that was my intention

PR #3342, opened 2015-01-14
General pane followup

dominikl:
Follow-up PR to the General Pane alignment, see [Trello](https://trello.com/c/OBUND1Za/82-general-pane-client-alignment)
- Show creation date for Datasets, Projects, etc.
- Show Annotation tooltips like Web does (display namespace, description, etc.)
- Filters aren't wrapped into 'Light Path' titled border anymore

Also implements [Trac 12681](https://trac.openmicroscopy.org.uk/ome/ticket/12681) : The image ID can be selected and copied with CTRL+C

### DIFF
--- a/components/blitz/src/pojos/AnnotationData.java
+++ b/components/blitz/src/pojos/AnnotationData.java
@@ -1,7 +1,7 @@
 /*
  * pojos.AnnotationData
  *
- *   Copyright 2006 University of Dundee. All rights reserved.
+ *   Copyright 2006-2015 University of Dundee. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  */
 
@@ -105,6 +105,28 @@ public abstract class AnnotationData extends DataObject {
         return timeOfEvent(getDetails().getCreationEvent());
     }
 
+	/**
+	 * Retrieves the {@link Annotation#getDescription() description} of the
+	 * underlying {@link Annotation} instance.
+	 * 
+	 * @return See above
+	 */
+	public String getDescription() {
+		omero.RString desc = asAnnotation().getDescription();
+		return desc == null ? null : desc.getValue();
+	}
+
+	/**
+	 * Sets the description of the underlying {@link Annotation} instance.
+	 * 
+	 * @param description
+	 *            The description
+	 */
+	public void setDescription(String description) {
+		asAnnotation().setDescription(
+				description == null ? null : rstring(description));
+	}
+    
     /**
      * Sets the annotation value. If the improper content is given for the
      * current {@link Annotation}, then an
--- a/components/blitz/src/pojos/DataObject.java
+++ b/components/blitz/src/pojos/DataObject.java
@@ -1,7 +1,7 @@
 /*
  * pojos.DataObject
  *
- *   Copyright 2006 University of Dundee. All rights reserved.
+ *   Copyright 2006-2015 University of Dundee. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  */
 
@@ -351,7 +351,7 @@ public abstract class DataObject {
 
     protected Timestamp timeOfEvent(Event event) {
         if (event == null || !event.isLoaded() || event.getTime() == null) {
-            throw new IllegalStateException("Event does not contain timestamp.");
+            return null;
         }
         return new Timestamp(event.getTime().getValue());
     }
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/view/SearchResultTable.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/view/SearchResultTable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014 University of Dundee & Open Microscopy Environment.
+ * Copyright (C) 2014-2015 University of Dundee & Open Microscopy Environment.
  * All rights reserved.
  *
  * This program is free software; you can redistribute it and/or modify
@@ -70,10 +70,6 @@ import pojos.ImageData;
  * @since 5.0
  */
 public class SearchResultTable extends JXTable {
-
-    /** Defines the format how the date is shown */
-    private static final DateFormat DATE_FORMAT = new SimpleDateFormat(
-            "yyyy-MM-dd HH:mm:ss"); //2013-04-20 18:13:43
     
     /** Reference to the DataBrowserModel */
     private AdvancedResultSearchModel model;
@@ -285,7 +281,7 @@ public class SearchResultTable extends JXTable {
                 JLabel l = new JLabel((Icon) value);
                 p.add(l);
             } else if (value instanceof Date) {
-                JLabel l = new JLabel(DATE_FORMAT.format((Date)value));
+                JLabel l = new JLabel(UIUtilities.formatDefaultDate(((Date)value)));
                 p.add(l);
             }
             else {
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/DocComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/DocComponent.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.metadata.editor.DocComponent 
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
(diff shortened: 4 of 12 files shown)


gusferguson:
@dominikl @jburel @will-moore @pwalczysko 

Tested using OMERO.insight-5.1.0-m3-826-cb9c45c-ice34-b393-mac on trout and http://trout.openmicroscopy.org/merge/webclient/ Mac Safari  user-3 read-annotate-1

Also implements Trac 12681 : The image ID can be selected and copied with CTRL+C
- works as expected

Show creation date for Datasets, Projects, etc.
- creation date title is shown, but value shows N/A even for containers created today
- see screenshot 1

Show Annotation tooltips like Web does (display namespace, description, etc.)
- annotations have tooltips displayed but ...
- there are huge differences between what is shown in the Web tooltips and the Insight tooltips in terms of what is shown, how it is shown, order of items and naming of items.
- see screenshot 2

Filters aren't wrapped into 'Light Path' titled border anymore
- no idea on how to test this - need some guidance

On the display of of information in the General Pane in general:
- there are still a couple of important differences and a number of less important differences
- it's seems a pity to have spent so much time and effort on updating, aligning and adjusting the General Pane and yet still be left with so much disparity

important difference - pixel size
- label different - Insight (XY) - Web (XYZ) 
- Insight shows (nm) - Web shows (µm) and therefore different decimal points with rounding so that the accuracy level is low in Web compared with Insight - screenshot 3 
- when this field zero/not available - Insight does not display it and Web does - screenshot 4
- when µm are being displayed by both Insight has 3 decimal places, Web 2 - screenshot 5

possibly less important differences:
- if acquisition date not present Insight displays blank field, Web has no entry - screenshot 6
- time is displayed in different modes - Insight 12 hour no seconds - Web 24 hrs with seconds
- pixel type is still displayed differently
- see screenshot 7

It seems that a number of these differences only show up with particular types of data and when information is missing, so a really comprehensive go-through of all elements with all data types and sources is probably needed to catch them all. 
It might be better for me to get this done for attention after the 5.1 release - say what y'all think is best.


omero_insight 5

tooltips

omero_insight 2

omero_insight 3

omero_insight

omero_insight 4

omero_insight 7

jburel:
* Dimension and Pixels Size: Insight does more check that web. If for example there is not a Z value, then Dimension (Z) is not displayed. Same for the pixels size if it is 0. We do not display.
Web should be a bit smarter and follow the insight logic (see point below)
* Acquisition date in insight should be hidden if not set. So we follow the same logic. 
* The pixel Type needs to be unified (item on trello card)

dominikl:
- Creation Date is always N/A because the DataObject's creation event is always null; issue with loading the objects from the server. @jburel You had an idea how to fix that, if I remember right?
- Pixel type. Just to confirm (not really clear on the trello board discussion): That means Insight should take over the Web naming of pixel types?
- Dates: I'll remove the Acquisition Date label, if there is no acquisition date. But what is the date format, we shall use throughout? YYYY-MM-DD hh:mm:ss (Web) or MM/DD/YY hh:mm am/pm (Insight)? Guess I'll go for the Web format, as this is also used in the search results, right?
- Tooltips: It just appeared a bit odd to me, to remove the additional information we already show in Insight's tooltips, so I just extended it.

dominikl:
I just noticed a weird thing: Clicking on a search result item, the creation date is not null. Still investigating this.

dominikl:
I don't know why the creation date on DataObjects isn't initialized. I'll tackle this problem within another PR, on the UI side (which this PR is about) it should be fine. To test the creation date label for now, you have to click on a Dataset/Project in the search result table, curiously there it is initialized.

jburel:
@dominikl: for pixels type, yes. Date should match the search see https://trello.com/c/2vdtlj9e/267-date-format-and-display

dominikl:
@gusferguson The 'lightpath' issue is the last checkbox on the trello card in the Insight section. Filter metadata was wrapped within a 'Lightpath' titled border panel, now it just displayed consecutively without the titled border. Unfortunately I can't remember, which file I used for testing; I think it was one of generated files we used for units testing.

gusferguson:
@jburel - do you know what I can use to test the filter metadata-light path aspect?

@dominikl - is it worth testing this PR again today or should we just wait until you have sorted everything?

dominikl:
@gusferguson  Happy to wait with the review, until the creation date issue is sorted out. @jburel Could you give me a hint about the DataObject/Creation date initialization? I couldn't track it down, yet.

mtbc:
Random guess: related to some server container service or Insight gateway code where the HQL needs an extra join on that property?

jburel:
@mtbc: that's what is required. @dominikl: I will show you sorry for not doing that earlier

jburel:
@gusferguson: I will import a file for you

dominikl:
@jburel Could you have a look at the last commit? It works, but I'm not sure if this really is the right place for this change.

jburel:
@dominikl: That the call we are using. You will have to adjust the integration tests or add some more to be sure nothing is broken

dominikl:
Added an integration test to check if the creationDate is loaded. All other tests are still passing, so the modification of the HQL query doesn't seem to affect any other methods.

jburel:
@dominikl: good thanks

dominikl:
Differences still left: 
- Insight does a transformation to a reasonable unit for the pixels size. Web apparently uses the plain value.
- Insight uses the default locale for number formatting, which seems to round to 3 decimal places. Web uses two.
Which way did we agree on for these two points, @jburel @gusferguson ? I'm struggling a bit to find this information between all the tickets, trello boards and PRs.
A third point:
- Shall I really remove the additional information in the annotation tooltips? It's different from Web, but I don't think, showing additional information is a harm.

jburel:
* Point 1: we agree to use reasonable units for sizeX/Y etc and time. Web needs to be adjusted cc @will-moore 
 * Point 2: I do not think we discussed it. 2 decimals seem fine.
 * Point 3: I will leave it as it is,Maybe add more in Web cc @will-moore

dominikl:
Reverted my changes for the point (trello card) "Metadata, lightpath: only display elements in the order, filters-dichroic-filter" again, because I think I got it wrong. Need some more clarification what this point actually means.

jburel:
@dominikl: the change was correct. The comment was there to make sure 
we need to display elements composing the lightPath in the specified order. This "reflects" the physical arrangement

dominikl:
Ok, removed my last commit again. 
@gusferguson Test image for the "lightpath" issue: user-3/read-annotate1/general pane alignment follow up/1h_after_pdt.oib (thanks @will-moore ). If I got it right, the problem was, that "excitation filter (if present) - dichroic (if present)  - emission filter (if present) " were shown inside a titled component named "Lightpath" (see screenshot 1), now there are just shown consecutively without the titled border (see screenshot 2), but still in the order excitation - dichroic - emission. 
Before:
screen shot 2015-02-04 at 14 09 55
Now:
screen shot 2015-02-04 at 14 09 05

I think this PR is now really ready for a (hopefully final) re-review. Due to server side changes, also make sure that the integration test jobs are still passing.

gusferguson:
**Original tests:**

Show creation date for Datasets, Projects, etc.
- works as expected.

Show Annotation tooltips like Web does (display namespace, description, etc.):
- not consistent yet
- still large difference between Insight and Web in what is and isn’t shown, what the order is and what they are labelled as.
- see screenshot 3
- I am happy with leaving it for now - until I can scope out a consistent approach across all tooltips.

Filters aren't wrapped into 'Light Path' titled border anymore
- works as expected.

**Other issues that arose:**

No acquisition date
- not appearing if blank
- works as expected

Time/date display mode
- showing YYYY-MM-DD hh:mm:ss
- consistent across Insight search and Web General pane and search
- works as expected

Pixel type display different from Web
- now consistent with web
- works as expected

Pixel size display
- displaying to 2 decimal places if they are there, but not if less
- inconsistent with web which displays 2 decimal places even if whole number
- see screenshots 1 and 2
- I think this should be done the way web does it - to 2 decimal places even if they are 0 - makes it much easier to read.

I think that is all the points covered.

So apart from the display of 2 decimal places even if 0, I think this is good to merge.

safari

safari

untitled

jburel:
adjusting the tooltip should be straightforward. Only one place to adjust.

dominikl:
Fixed the decimal places issue. Leaving the tooltip issue for now, until we agreed on a common format.

jburel:
@dominikl the last change looks good
comment added to card https://trello.com/c/OBUND1Za/82-general-pane-client-alignment
Not do be done in this PR.
Merging

PR #3343, opened 2015-01-14
os version in upgrade check

aleksandra-tarkowska:
@rleigh-dundee asked to provide more information about OS in UpgradeCheck. This is because original implementation uses only ```platform.version()```. This change should give more details to registry.

Mac OS X:
```
>>> "%s-%s" % (platform.platform(),platform.mac_ver()[0])
'Darwin-12.5.0-x86_64-i386-64bit-10.8.5'
```

Linux:
```
>>> platform.platform()
'Linux-2.6.32-431.29.2.el6.x86_64-x86_64-with-centos-6.5-Final'
```

Windows:
```
>>> platform.platform()
'Windows-2008Server-6.0.6002-SP2'
```

Test library locally logging to web calling python directly:

```
from omero.util.upgrade_check import UpgradeCheck
check = UpgradeCheck("OMERO.test")
check.run()
```

```
2015-01-14 16:26:10,994 DEBUG [                 omero.util.UpgradeCheck] (proc.31949) run:136 Attempting to connect to http://upgrade.openmicroscopy.org.uk/?os.version=Darwin-12.5.0-x86_64-i386-64bit-10.8.5&os.arch=x86_64&version=5.1.0-m4-ice34-SNAPSHOT&python.compiler=GCC+4.2.1+Compatible+Apple+Clang+4.0+%28tags%2FApple%2Fclang-418.0.60%29&python.build=%28%27default%27%2C+%27Oct+11+2012+20%3A14%3A37%27%29&os.name=Darwin&python.version=2.7.2
```

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/upgrade_check.py
+++ b/components/tools/OmeroPy/src/omero/util/upgrade_check.py
@@ -94,7 +94,7 @@ class UpgradeCheck(object):
     def getOSVersion(self):
         try:
             if len(platform.mac_ver()[0]) > 0:
-                version = "%s-%s" % (platform.platform(),
+                version = "%s;%s" % (platform.platform(),
                                      platform.mac_ver()[0])
             else:
                 version = platform.platform()
--- a/components/tools/OmeroPy/src/omero/util/upgrade_check.py
+++ b/components/tools/OmeroPy/src/omero/util/upgrade_check.py
@@ -91,6 +91,17 @@ class UpgradeCheck(object):
         self.upgradeUrl = results
         self.exc = e
 
+    def getOSVersion(self):
+        try:
+            if len(platform.mac_ver()[0]) > 0:
+                version = "%s-%s" % (platform.platform(),
+                                     platform.mac_ver()[0])
+            else:
+                version = platform.platform()
+        except:
+            version = platform.platform()
+        return version
+
     def run(self):
         """
         If the {@link #url} has been set to null or the empty string, then no
@@ -110,7 +121,7 @@ class UpgradeCheck(object):
             params["version"] = self.version
             params["os.name"] = platform.system()
             params["os.arch"] = platform.machine()
-            params["os.version"] = platform.version()
+            params["os.version"] = self.getOSVersion()
             params["python.version"] = platform.python_version()
             params["python.compiler"] = platform.python_compiler()
             params["python.build"] = platform.python_build()


components/tools/OmeroPy/src/omero/util/upgrade_check.py
```diff
@@ -91,6 +91,17 @@ def _set(self, results, e):
         self.upgradeUrl = results
         self.exc = e
 
+    def getOSVersion(self):
+        try:
+            if len(platform.mac_ver()[0]) > 0:
+                version = "%s-%s" % (platform.platform(),
```
rleigh-dundee: I thought we were going to use `;` as the separator here so we could use a regex to separate the information?

aleksandra-tarkowska:
@rleigh-dundee done

rleigh-dundee:
I'm having trouble testing this; I don't get any debugging output.  How are you running the test?  i.e. working directory/environment etc.

rleigh-dundee:
This looks fine, thanks.

For future work, unifying the reported data for both python and java would be useful; it's currently quite different.  And splitting up this information into separate columns would also make it much easier to query.

aleksandra-tarkowska:
@rleigh-dundee as discussed these changes would require model changes in registry. I will take it to the account during upgrade.

rleigh-dundee:
Good to merge.

joshmoore:
Thanks, guys. Merging.

@aleksandra-tarkowska : based on what you've been looking at recently, would it make sense to backport this, or will complicate things?

aleksandra-tarkowska:
--rebased-to #3367

aleksandra-tarkowska:
@rleigh-dundee I forgot about that before, should Java library alter that value as well?

rleigh-dundee:
@aleksandra-tarkowska It would certainly be nice if the Python and Java code submitted the same information, so it's meaningful to query it.

aleksandra-tarkowska:
@joshmoore @jburel @dominikl could you also make sure java UpgradeCheck is setting right value?

joshmoore:
The extent available in Java is:
```
>>> p = java.lang.System.getProperties()
>>> for x in p.keySet():
...     if x.startswith("os"): print x, p.get(x)
...
os.arch x86_64
os.name Mac OS X
os.version 10.10.2
```

mtbc:
In case it's of any use, the corresponding values from Java for my laptop are, respectively,
- amd64
- Linux
- 3.2.0-4-amd64.

rleigh-dundee:
For MacOS X that's fine.  For Linux, it's not very useful.  We're missing distribution+version; the kernel version isn't very useful in comparison.  Adding that would be very helpful; there are a few places to check to determine this e.g. `lsb-release -a`, /etc/redhat-release /etc/debian_version.  Not perfect or comprehensive, but good enough for our needs.

mtbc:
Could `/etc/issue` be useful across distributions?

rleigh-dundee:
Not generally, since this is intended for customisation by the admin.  The formal interface for this is `lsb-release`, but it's not necessarily available by default on all systems without LSB support being installed.  It's installed on Ubuntu by default though.  However, may be useful as a fallback in the absence of lsb-release or /etc/redhat|debian*

mtbc:
True. I have it locally as `lsb_release`, not `lsb-release`.

joshmoore:
cF. https://github.com/aurbroszniowski/os-platform-finder

PR #3345, opened 2015-01-15
adjust createdb commands to specify Unicode databases

mtbc:
https://trello.com/c/Fv1HZVfs/779-virt-microscope-pg-9-for-testing reveals the need for a UTF8-encoded database for OMERO since 5.1 patch 12.

PostgreSQL 8.4 references can be updated once version prerequisites suggested via http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-January/004982.html are approved, so for now they are left alone.

See https://github.com/openmicroscopy/ome-documentation/pull/1070#discussion_r22995970

--no-rebase

### DIFF
--- a/docs/QUICKSTART.bat
+++ b/docs/QUICKSTART.bat
@@ -61,7 +61,7 @@ createuser -S -D -R -U postgres omero
 if errorlevel 1 echo Already exists?
 
 echo Creating %OMERO_CONFIG% db
-createdb -O omero -U postgres %OMERO_CONFIG%
+createdb -E UTF8 -O omero -U postgres %OMERO_CONFIG%
 if errorlevel 1 goto ERROR
 
 echo Adding pgsql to db
--- a/docs/hudson/OMERO-homebrew-install.sh
+++ b/docs/hudson/OMERO-homebrew-install.sh
@@ -111,7 +111,7 @@ export ICE_CONFIG=$(bin/brew --prefix omero51)/etc/ice.config
 if [ -d "$PSQL_DIR" ]; then
     rm -rf $PSQL_DIR
 fi
-bin/initdb $PSQL_DIR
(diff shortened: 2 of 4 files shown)


sbesson:
See https://ci.openmicroscopy.org/job/OME-5.1-merge-homebrew/17/

```
++ bin/createdb -E UTF8 -O db_user omero_database
createdb: database creation failed: ERROR:  new encoding (UTF8) is incompatible with the encoding of the template database (SQL_ASCII)
```

mtbc:
One could consider this a feature: The database that was being previously created is unsuitable for running OMERO 5.1.0!

Whether inside or outside this PR, what do we suggest? Does it work in that environment if you add in a `-T template0` option too? If so, should we adopt that more widely in https://github.com/openmicroscopy/ome-documentation/pull/1070 too? I'm venturing a little out of my depth here so comment by @kennethgillen or @rleigh-dundee is welcome.

manics:
I think the docs should reflect what works today with our recommended software versions. The current createdb command should already work, so si there an OS X specific problem? Was the postgres used in the homebew job out of date, or installed strangely?

rleigh-dundee:
Agreed that the above error is reasonable to get.  It indicates that the database cluster itself (rather than just the omero db inside the cluster) needs replacement.

sbesson:
@mtbc: can you change 2 lines above in the script to use the encoding option? 
```
initdb -E UTF8 /usr/local/var/postgres
```
This should work with a fresh Homebrew PostgreSQL.

mtbc:
Do any of the other scripts need a similar change?

sbesson:
http://ci.openmicroscopy.org/view/4.4/job/OME-5.1-merge-homebrew/ is green with these changes in. So :+1: from the perspective of these scripts. I don't see any other scripts in these repos that would require an update /cc @manics

PR #3346, opened 2015-01-15
Scripts/Downloads menu

dominikl:
Implements the m3 changes for scripts/downloads menu, see [Trello card](https://trello.com/c/3n7Plb19/189-scripts-menus-download-options).
Check for:
- 'Download Original MetaData' removed from download menu
- Download menu is now a single-level menu like Web
- Term 'Export' is used instead of 'Save As'

Task 'Delete FileAnnotation after Download' will be tackled in another PR.

--no-rebase

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
@@ -259,14 +259,7 @@ class ToolBar
         }
         exportAsOmeTiffItem.setEnabled(b);
         saveAsMenu.add(exportAsOmeTiffItem);
-        JMenu menu = new JMenu();
-        menu.setIcon(icons.getIcon(IconManager.SAVE_AS));
-        menu.setText("Save as...");
-        menu.setToolTipText("Save the images at full size as JPEG. PNG or" +
-                "TIFF.");
         ActionListener l = new ActionListener() {
-
-
             public void actionPerformed(ActionEvent e) {
                 int index = Integer.parseInt(e.getActionCommand());
                 controller.saveAs(index);
@@ -281,14 +274,14 @@ class ToolBar
                 ho instanceof WellSampleData || ho instanceof DatasetData);
         while (i.hasNext()) {
             e = i.next();
-            item = new JMenuItem();
-            item.setText(e.getValue());
+            item = new JMenuItem(icons.getIcon(
+                    IconManager.EXPORT_AS_OMETIFF));
+            item.setText("Export As "+e.getValue());
             item.addActionListener(l);
             item.setActionCommand(""+e.getKey());
             item.setEnabled(enabled);
-            menu.add(item);
+            saveAsMenu.add(item);
         }
-        saveAsMenu.add(menu);
         setRootObject();
     	return saveAsMenu;
     }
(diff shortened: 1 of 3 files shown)


joshmoore:
Bumped due to romio failure. That's getting annoying...

components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
```diff
@@ -302,14 +274,14 @@ public void actionPerformed(ActionEvent e) {
                 ho instanceof WellSampleData || ho instanceof DatasetData);
         while (i.hasNext()) {
             e = i.next();
-            item = new JMenuItem();
-            item.setText(e.getValue());
+            item = new JMenuItem(icons.getIcon(
+                    IconManager.EXPORT_AS_OMETIFF));
+            item.setText("Export As "+e.getValue());
```
mtbc: maybe lowercase "as" (same for `PublishingDialog`'s export message)

gusferguson:
Tested using OMERO.insight-5.1.0-m3-424-5397b0e-ice35-b382-mac trout user-3 read-annotate-1

Menu displays as expected - see screenshot 1

"Export As" of whole dataset as JPEG/PNG/TIFF
- first attempt saves to selected folder as Batch_Image_Export.zip
- subsequent attempts appear to download as Batch_Image_Export.zip according to Activity menu, but actually fail to appear in selected folder
- same applies for any "Export As" - if there is a Batch_Image_Export.zip file in the selected folder there is no download despite what Activities menu says - see screenshot 2

If there is no existing Batch_Image_Export.zip in the destination folder then:
- “Export As” works for dataset as expected
- “Download” works as expected for individual images
- “Export As” OME-TIFF/JPEG/PNG/TIFF works for individual images

This is a regression from 5.0.6 where the second Batch_Image_Export.zip is downloaded and saved with an incremented number appended - Batch_Image_Export_(1).zip etc

I think that files should be named for the dataset or image for download - not a generic name of “Batch_Image_Export.zip” used.
Ideally the name should be used as the default and the user should have an option to change the name in the destination selection dialog.
At the bare minimum if the “Export As” is going to fail because of an existing file of the same name then the Activity window needs to reflect this.

omero insight

finder

dominikl:
Thanks for figuring this out. I had so many Batch_Image_Export_(xx).zips in my testing folder that I didn't notice, that it's actually not working if the zip file already exists. With respect to name 'Batch_Image_Export.zip': I don't think I can do something about that within this PR, because that's the name/file which comes from the batch export script Insight launches (I'm not really sure how this scripts works, I have to admit).

jburel:
Export as JPEG, PNG, Tiff, should have "..." after the text since a dialog pops up.

jburel:
The "zips" number is now updated.
Changing the name will require some adjustment to the sequence. 
@dominikl: do you think you could have a look? This could be done in another PR.

dominikl:
Added '...' to the 'Export as' menus. Don't think it needs another review for that, does it? 
@jburel  Yes, I'll have a look at the name issue in another PR.

jburel:
@dominikl: we just need to make everything is green tomorrow then we can merge.

jburel:
@gusferguson: are you happy with the changes now?

gusferguson:
@dominikl @jburel  Looks good to me … I am happy thanks.

PR #3350, opened 2015-01-15
Add CLI info, listusers and listgroups subcommands

sbesson:
Fixes https://trac.openmicroscopy.org/ome/ticket/12589

This PR adds 
- a `info` subcommand to both `UserControl` and `GroupControl` plugins. These commands have the same behavior as the `list` subcommand for a subset of users/groups. The following commands should now be accepted

  ```
bin/omero user info   # Give info about the context user
bin/omero user info user-1
bin/omero user info user-1 --user-id 3 --user-name user-4
bin/omero group info   # Give info about the context user
bin/omero group info read-write1
bin/omero group info read-write1 --group-id 3 --group-name read-annotate-1
```

- a `bin/omero group listusers` and a `bin/omero user listgroups` subcommands. Both subcommands either take no argument (using the context) or a single argument. The following calls should work:

  ```
bin/omero group listusers   # Return the users of the context group
bin/omero group listusers read-write1
bin/omero user listgroups   # Return the groups of the context user
bin/omero user listgroups user-6
```

To test this PR:
- check the commands above are working in a multi-user/multi-group context
- check the integration tests are passing both for the new commands as well as the `user list` and `group list` subcommands which were refactored as part of this PR

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/test/unit/clitest/test_user.py
+++ b/components/tools/OmeroPy/test/unit/clitest/test_user.py
@@ -23,7 +23,8 @@ from omero.plugins.user import UserControl
 from omero.cli import CLI
 import pytest
 
-subcommands = ['add', 'list', 'password', 'email', 'joingroup', 'leavegroup']
+subcommands = ['add', 'list', 'password', 'email', 'joingroup', 'leavegroup',
+               'info']
 
 
 class TestUser(object):
--- a/components/tools/OmeroPy/test/unit/clitest/test_user.py
+++ b/components/tools/OmeroPy/test/unit/clitest/test_user.py
@@ -24,7 +24,7 @@ from omero.cli import CLI
 import pytest
 
 subcommands = ['add', 'list', 'password', 'email', 'joingroup', 'leavegroup',
-               'info']
+               'info', 'listgroups']
 
 
 class TestUser(object):
--- a/components/tests/python/library/__init__.py
+++ b/components/tests/python/library/__init__.py
@@ -78,8 +78,6 @@ class ITest(object):
 
         p = Ice.createProperties(sys.argv)
         rootpass = p.getProperty("omero.rootpass")
-        if not rootpass:
-            raise Exception("No omero.rootpass found in ICE_CONFIG")
 
         try:
             cls.root = omero.client()  # ok because adds self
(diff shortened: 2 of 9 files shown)


components/tools/OmeroPy/test/unit/clitest/test_user.py
```diff
@@ -23,7 +23,8 @@
 from omero.cli import CLI
 import pytest
 
-subcommands = ['add', 'list', 'password', 'email', 'joingroup', 'leavegroup']
+subcommands = ['add', 'list', 'password', 'email', 'joingroup', 'leavegroup',
```
joshmoore: Wonder if this could be added to the "BaseControl" API? Somehow it should be possibly to figure out what commands are available...

sbesson: Definitely a sensible idea, will try to work on that in a separate branch.

joshmoore:
Various `user info` points:
 * under `info` I expected more than just a listing of groups. Perhaps this is really `groups`, and `info` could list the likes of LDAP DN, email, institution, etc. Perhaps (later) with an `--edit` option?
 * The output of `users info --long` can be fairly hard to use since it is so wide and `--style=csv --long` isn't parseable (though this problem likely exists currently for other output).
 * eventually having `info [username | userid]` would make sense.

joshmoore:
Some of above also applies to `group` just swapped. Only other question at this point is why the difference between the defaults of `--long` and `--count`?

sbesson:
On the last question, one of the relevant PRs was https://github.com/openmicroscopy/openmicroscopy/pull/514. As far as I can tell, the main reason for doing this is that `group user` with `--long` would result in a very long line for the `user` group. If we homogeneize, I would suggest we use `--count` as the default across the board.

sbesson:
Excluding to let #3356 be reviewed. More work coming / PR description will be updated.

sbesson:
PR description/title updated to describe the new subcommands.

ximenesuk:
I've managed to briefly check out the commands and most of the options. I don't know if there is more to come but I'll take another look tomorrow once the later commits have been tested by the CI system.

ximenesuk:
This looks fine to me. Everything seems to work as expected. Good to merge.

PR #3351, opened 2015-01-15
Download spw original files dev50

will-moore:
This is https://github.com/openmicroscopy/openmicroscopy/pull/2953 rebased back to dev_5_0 (discussed last Tuesday meeting).

This adds the "Download" menu (and fileset paths) to SPW images when wells are selected.

To test:

Import a Plate.
Also create a plate from a Dataset with Dataset_To_Plate.py script (simpler if images are not MIFs).
Select ONE well from each plate:
Test the "Show paths" button
Download the Original files for the well. This will be a single image for the created plate and the whole plate for the imported plate.
Export as png/jpeg/tiff. Check that you get the correct image for the current Plate 'index'.
Select MULTIPLE wells from each plate:
Test paths, download original and export png/jpeg/tiff as described above.

screen shot 2015-01-15 at 13 46 26

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -255,6 +255,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
     "omero.web.server_list": ["SERVER_LIST", '[["%s", 4064, "omero"]]' % CUSTOM_HOST, json.loads],
     # Configuration options for the viewer. -1: zoom in fully, 0: zoom out fully, unset: zoom to fit window
     "omero.web.viewer.initial_zoom_level": ["VIEWER_INITIAL_ZOOM_LEVEL", None, leave_none_unset_int],
+    "omero.web.plate_download.enabled": ["PLATE_DOWNLOAD_ENABLED", "false", parse_boolean],
     # the following parameters configure when to show/hide the 'Volume viewer' icon in the Image metadata panel
     "omero.web.open_astex_max_side": ["OPEN_ASTEX_MAX_SIDE", 400, int],
     "omero.web.open_astex_min_side": ["OPEN_ASTEX_MIN_SIDE", 20, int],
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
@@ -181,12 +181,21 @@ class BaseContainer(BaseController):
         elif self.acquisition:
             return self.acquisition._obj.plate.id.val
         
-    def isDownloadDisabled(self):
+    def isDownloadDisabled(self, objDict=None):
         """
-        Returns True only if we have an SPW object and
+        Returns True only if we have an SPW object(s) and
         settings.PLATE_DOWNLOAD_ENABLED is false
         """
-        if (self.screen is None and self.acquisition is None and
+        # As used in batch_annotate panel
+        if objDict is not None:
+            spwData = False
+            for spw in ('screen', 'plate', 'well', 'acquisition'):
+                if spw in objDict and len(objDict[spw]) > 0:
+                    spwData = True
+            if not spwData:
+                return False
+        # As used in metadata_general panel
+        elif (self.screen is None and self.acquisition is None and
                 self.plate is None and self.well is None):
             return False
         if hasattr(settings, 'PLATE_DOWNLOAD_ENABLED'):
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
@@ -181,6 +181,16 @@ class BaseContainer(BaseController):
         elif self.acquisition:
             return self.acquisition._obj.plate.id.val
         
+    def isDownloadDisabled(self):
+        """
+        Returns True only if we have an SPW object and
+        settings.PLATE_DOWNLOAD_ENABLED is false
+        """
+        if (self.screen is None and self.acquisition is None and
+                self.plate is None and self.well is None):
+            return False
+        if hasattr(settings, 'PLATE_DOWNLOAD_ENABLED'):
+            return (not settings.PLATE_DOWNLOAD_ENABLED)
 
     def listFigureScripts(self, objDict=None):
         """
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -90,7 +90,7 @@
                 Download...</a>
             </li>
             {% endif %}
-            {% if image %}
+            {% if image and not omeTiffDisabled %}
             <li>
                 <a id="create-ometiff" href="{% url 'ome_tiff_script' image.id %}" 
                 title="Create OME-TIFF File for Download">Export as OME-TIFF...</a>
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -81,7 +81,7 @@
                     href="{% url 'archived_files' %}?{{ obj_string }}"
                   {% else %}
                     href="#"
-                    onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&fileCount={{ filesetInfo.count }}');"
+                    onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&fileCount={{ filesetInfo.count }}&index={{ index }}');"
                   {% endifequal %}
                 {% endifequal %}
                 title="Download {{ filesetInfo.count }} original imported file{{ filesetInfo.count|pluralize:'s as zip'}} ({{ filesetInfo.size|filesizeformat }})"
@@ -122,17 +122,17 @@
               {% if filesetInfo %}
               <li>
                 <a href="#"
-                  onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&format=jpeg');"
+                  onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&format=jpeg&index={{ index }}');"
                   title="Download as JPEGs (zip)">Save As JPEG</a>
               </li>
               <li>
                 <a href="#"
-                  onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&format=png');"
+                  onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&format=png&index={{ index }}');"
                   title="Download as PNGs (zip)">Save As PNG</a>
               </li>
               <li>
                 <a href="#"
-                  onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&format=tif');"
+                  onClick="return OME.openPopup('{% url 'download_placeholder' %}?ids={{ link_string }}&format=tif&index={{ index }}');"
                   title="Download as TIFFs (zip)">Save As TIFF</a>
               </li>
               {% endif %}
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -45,7 +45,7 @@
     <!-- This is used for 'batch annotate' panel with 'filesetCount'... -->
     <!-- ... and for single images -->
 
-    {% if manager.image or filesetInfo %}
+    {% if image or filesetInfo %}
     <!-- download options -->
     <div class="toolbar_dropdown">
         <button class="btn silver btn_download" title="Download Image as...">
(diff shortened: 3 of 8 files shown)


dominikl:
Works like expected

joshmoore:
@will-moore : was this intended for @imunro's use case?

imunro:
I would imagine it was, as I asked for it and the test data shown above is mine also.

joshmoore:
Then, does it do everything you need?

imunro:
I've never worked out a good way to test the public downloads. Even if I could set up a public user on a test server it wouldn't replicate our configuration, firewalls etc ... open to suggestions.

will-moore:
If the download of data from an Image currently works for your public deployment, then I would expect this to work too.

imunro:
That (Image data download) currently works exactly as hoped.

joshmoore:
@pwalczysko / @chris-allan : anything else from you guys?

chris-allan:
At a minimum I would propose testing on some larger Plates before merging. This has some very real potential pitfalls, especially resource exhaustion. Enabling this on a system with a large Plates available to the public user raises serious denial of service concerns.

imunro:
I can provide larger files for test if required.

will-moore:
Of course it would be unwise to allow public access to resource-intensive operations, but that's true of any functionality (E.g. running scripts, chgrp etc). That's not a reason to block this PR (which is already in develop). It's just the responsibility of admins to only allow public access to 'safe' operations.

imunro:
FYI I have imported a much larger plate (of real data) to https://cisbic.bioinformatics.ic.ac.uk/omero/webclient/?show=dataset-7604
I have imported this into a dataset so it can be downloaded. If you want to use this file for test simply download and re-import into a screen.

sbesson:
In the same spirit as https://github.com/openmicroscopy/openmicroscopy/pull/3340, a way to allow admins to selectively enable this functionality would be to add a server-side (or Web) property allowing the plate download from the client (turned off by default?).

chris-allan:
I fully disagree @will-moore and had I been consulted on #2953 I would have said the same thing I've said above. That PR should be rolled back and this should not be merged as it stands.

This PR is not the same as any of the pieces of functionality you mention as they would not be available to public users or are all relatively easy to disable by URL restriction. None of those features expose you to server physical resource exhaustion (drive space or memory) either. Enabling such functionality by default is dangerous at best. It is so easy to end up in a denial of service scenario with this feature. Click the link 5 times with even a medium size set of data and your OMERO.web will be completely unusable until gigabyte size ZIPs are prepared in temporary directories after downloading potentially 1000s of files. Absolutely not acceptable.

Add to all of this the default 60 second request timeout and I'd venture to say that even on a medium size Plate you wouldn't be able to prepare the ZIP in time for successful download. The code in this PR is unlikely to work on anything but the most trivial of datasets.

You can't download datasets via OMERO.web for exactly these reasons @imunro.

imunro:
Sorry I'm a little confused here as to the difference (in terms of  DOS issues) between downloading a file that has been imported to a dataset (currently enabled)  and downloading the same file when imported to a screen. Could somebody clarify please.

mtbc:
I wonder if it's safe to enable this function only for single-file plates.

imunro:
@mtbc that would work for us.

chris-allan:
The difference @imunro, is that an Image imported into a Dataset is more often than not a single file or a small number of files. We have resolution to **exactly** which files on disk make up that Image or Multi-Image File (MIF). An Image as imported into a Plate is indistinguishable from the entire Plate as far as files on disk are concerned that OMERO has registered, via the Fileset mechanism, to belong to it.

Bio-Formats, and by extension OMERO, does not have single Well, WellSample (Field) or Image resolution for files that belong to that object. Even if we did, most Plates would make absolutely no sense without **all** their associated files. So, you're stuck downloading the **entire** Plate. Such a Plate is often many 100s of MB or GBs in size and may contain 1000s of files.

As an example we'll examine this rather modest 96-well Plate acquired as part of the HMS LINCS project served up via OMERO and OMERO.web 5.0.6:
 * http://lincs-omero.hms.harvard.edu/webclient/?show=plate-459

Just the binary data associated with this Plate (1024 x 1024; 2 bytes per pixel; 3 channels; 96 wells; 4 fields):
 * `1024 * 1024 * 2 * 3 * 96 * 4 = 2415919104 bytes` (**2.25GB**)

As you might expect, such a set of data takes a signifiant amount of time and resources to prepare as a single ZIP for download. As implemented it can also only be used for that singular download action.

imunro:
Thanks @chris-allan  that makes things clearer. It's certainly possible for us to continue storing our Plate data as sets of individual ome-tiffs (one per well sample with the Row and Column encoded in the file name and stored in a dataset) as we currently do.   A clear decision as to whether this capability is likely to be added in the not-too-distant-future would be helpful. (I'm assuming here that your reservations apply just as much to 5.1 as they do to 5.0.7). Worst case from our POV would be to proceed as described above and then find plate download enabled in a month or two.

chris-allan:
Correct @imunro, my concerns are completely version agnostic.

will-moore:
Following discussion this morning, I will add a boolean setting: ```omero.web.plate_download.enabled``` which is False by default and will only allow SPW download if this is set to True.
One extra thought - Even if this IS enabled, it seems a good idea to disable OME-TIFF generation for SPW data under all conditions, right? cc @imunro @chris-allan

imunro:
Sounds ok to me our plate data is OME-tiff in the first place so I've no idea what that would do .. generate an ome-tiff for each well-sample ?

chris-allan:
Definitely a good idea to disable export under all conditions I would think @will-moore.

will-moore:
To test the latest commits:
 - Download of SPW data should be disabled by default, both when a single well is selected and when multiple wells are selected (batch annotate panel). Download of non-SPW images should still be working.
 - After configuration and restart of web:
``` $ bin/omero config set omero.web.plate_download.enabled True ```
 - Now, download should be enabled for single or multiple selected wells, BUT OME-TIFF export should still be disabled for all SPW data.
 - Finally, check that OME-TIFF is still enabled for non-SPW data.

**Comment by mtbc** [comment]
Travis failure.

**Comment by chris-allan** [comment]
Travis build re-queued.

**Comment by mtbc** [comment]
Does seem to work well enough for @imunro's needs. :+1: Especially, the download/export options are enabled or disabled appropriately, so is already good to merge if needed for 5.0.7.

Some RFEs:

1. in multi-well selection, could still generate good zip file name using common prefix of image names or, easier, the plate name
1. directory structure of plate (below fileset template prefix) is not preserved in zip, useless for many SPW formats, especially those with identically named files in different directories.

**Comment by imunro** [comment]
During discussions about this, @chris-allan described the test data I've provided  as "plates that aren't really a plate".  I'm concerned that this means we are storing our data inappropriately. The history here is that we needed a file-format for our (non-plate) proprietary data purely in order to have an "original file" that can be downloaded from OMERO & chose ome-tiff, assuming that would work well with OMERO.
 I then attempted to do the same with plate data. My understanding, at the time, was that each file in the file set required the full plate description. This meant that having one one-tiff per well-sample was horribly inefficient (there being much more ome-xml than data in each file). We therefore settled on one big ome-tiff per plate. There was some discussion of 'binary' (possibly wrong terminology) ome-tiff where each file has a pointer to one file containing the ome-xml but I was advised that this capability had not been extensively tested. 

If anyone can suggest a better approach then  this would be a good time to do so, before we acquire too much data from our plate reader.

**Comment by chris-allan** [comment]
I was talking about forcing it into a Dataset @imunro as well as testing this and other pull requests using synthetic Plates created by the `Dataset_To_Plate.py` script as outlined in this PRs description. I wasn't specifically referring to your choice of OME-TIFF structure at all. Your data is most definitely a Plate and I can't see any issue with the way you are working.

The important thing to understand is that you are, to my knowledge, the only group we've ever seen that has the luxury of having their Plate data as a single file. What is in this PR is not going to work, either from a technical or usability perspective, for the vast majority of Plate based data. This is what @mtbc is referring to.

**Comment by imunro** [comment]
Ok thanks for clarifying. NB the data was forced into a dataset simply so that people could download the  file from our server for testing on your server.

**Comment by chris-allan** [comment]
For reference, there is no reason to do that other than the convenience of a button to click in the UI. The URL structure for the download of archived files (always available in OMERO 5) is as follows:

* `.../webgateway/archived_files/download/<image_id>/`

In your case you can get the file in question by following this link (`Image:114522`), or linking to any Image in that Dataset:

* https://cisbic.bioinformatics.ic.ac.uk/omero/webgateway/archived_files/download/114522/

**Comment by joshmoore** [comment]
Removed "rebased-from: #2953" from the description since this will need re-backporting, i.e. this PR needs at least 2 `rebased-*` comments.

**Comment by will-moore** [comment]
--rebased-to #3412

**Comment by chris-allan** [comment]
Before we merge this I'd like to see the disabling conditionals extended to the download view and integration tests added to cover that. We are then sure we're not exposing this haphazardly. The URLs are easily predicted as above.

**Comment by will-moore** [comment]
OK - I'll disable urls like  ```/webgateway/archived_files/download/?image=501``` and ``` /webgateway/archived_files/download/500/ ``` if they point to plates and the ```omero.web.plate_download.enabled ``` setting is false. Return a 404 or 403?

**Comment by chris-allan** [comment]
Sounds good. `404` seems fine.

**Comment by imunro** [comment]
So there was a way of downloading these files, although we didn't know it , and I've now persuaded you to disable it.!  Excellent!

**Comment by chris-allan** [comment]
There was and there wasn't. It didn't handle Well based targets for download until this PR.

**Comment by will-moore** [comment]
@chris-allan I've disabled SPW download in webgateway.views.py and added an integration test that checks we get a 404 response for the download url with an SPW image.
Currently, trying to download a non-SPW test image returns 500 because it has no original archived files. This confirms that the SPW check only applies to SPW images (although it might be nice to have the download actually working in the tests?).

**Comment by sbesson** [comment]
The `library.py` has quite a few ultities to import fake images via the importer, see `importMIF`, `importSingleImage`, `importSingleImageWithCompanion`.

**Comment by chris-allan** [comment]
I would argue that we should not knowingly return, and thus test for, a `500` status code. `404` seems like the most reasonable reaction to a missing set of original archived files. As @sbesson has outlined, there are also existing `library.py` utilities to create more thoroughly populated Images so that you don't have to re-create the structure yourself.

**Comment by will-moore** [comment]
@chris-allan @joshmoore Agree that it would be right to make those changes, and I'll start on it now. But is this going to hold up the 5.0.7 RC1 build?
If so, is it worth waiting or are the current tests enough for the release (they do at least confirm we get a 404 where expected)?

**Comment by chris-allan** [comment]
I'll defer to @joshmoore and @jburel on 5.0.7 decisions.

Regarding the code, the tests do seem sufficient to confirm that we get a `404` status code where appropriate and a `500` is certainly no worse than the current state of 5.0.6. That said, it does seem rather silly to not fix this properly when we have integration tests now that expose a rather glaring deficiency.

**Comment by jburel** [comment]
I think we should not put in dev_5_0 + It is targeted to a specific use case
I will focus on develop only and have it solid there so we can put it out with more confidence.
It is a bit too late for 5.0.7.

**Comment by will-moore** [comment]
@jburel I've fixed the test. If this doesn't go in 5.0.7 then all the above work is wasted and @imunro will still not be able to download SPW data!

**Comment by imunro** [comment]
I haven't thoroughly tested yet but it looks as if the 
.../webgateway/archived_files/download/<image_id>/  approach that Chris mentioned might allow us to download SPW data after a lll. Doesn't help with the wasted work I'm afraid.

**Comment by will-moore** [comment]
@imunro You might find that the link above does allow you to download your plate, but users browsing the data in the webclient will not have any way to access those links.

**Comment by jburel** [comment]
It won't be wasted since it has improved the code in develop.

**Comment by chris-allan** [comment]
As long as the integration tests pass I would be happy to merge this as is. Can someone record somewhere that we do need to address the `500` status code when making this request where we do not have archived original files?

**Comment by will-moore** [comment]
@chris-allan I added a comment to the develop rebase of this PR: #3412 and will do it in that PR.

**Comment by chris-allan** [comment]
:+1:

**Comment by imunro** [comment]
@will-moore Thanks. I got that the new work is to put the functionality  into the GUI but just being able to send a working link would give us a work-around we could use.

**Comment by jburel** [comment]
See failure
https://ci.openmicroscopy.org/view/5.0/job/OMERO-5.0-merge-integration-web/240/testReport/junit/test.integration.test_download/TestDownload/test_image_download/

**Comment by will-moore** [comment]
Seems like the test library import_image() is failing with:
```
out = 'OMERO_HOME=../../../dist cannot be found'
```

sbesson:
The error above is caused by https://github.com/openmicroscopy/openmicroscopy/pull/3191. I would assume using an an absolute path as `OMERO_HOME` would be enough as a first fix to restore the logic. I will add a fix and a simple test in https://github.com/openmicroscopy/openmicroscopy/pull/3405.

will-moore:
So, just to be clear; Is this going to be in 5.0.7 or not?

imunro:
I'd been wondering the same.

jburel:
@will-moore, @imunro: need to fix the path in order to be sure that the test passes.

jburel:
This was not included in rc1, if we want to have it:
 * Test must be green. Cherry-pick commit 0c9c972  from @sbesson 
 * we will need an rc2.
 * This implies that we cannot release not Thursday.
cc @pwalczysko, @sbesson

imunro:
This sounds very much like a "no, This won't be in 5.0.7 to me" ?

jburel:
Having an rc2 is not a problem (we might need one), fix to this PR must be made first in order to be considered.

will-moore:
@jburel Sorry, I was thinking that @sbesson had added that fix to another PR, but I realise now that was only for develop.
Just cherry-picked https://github.com/openmicroscopy/openmicroscopy/commit/0c9c9726029c20a27c82702186b958ea018e20e3 above.

sbesson:
https://ci.openmicroscopy.org/job/OMERO-5.0-merge-integration-web/242/testReport/test.integration.test_download/TestDownload/test_image_download/?. Thanks. Merging

PR #3352, opened 2015-01-15
Delete FileAnnotation after image export

dominikl:
If an image is exported as JPEG, PNG or TIFF the images are zipped into a Batch_Image_Export.zip file (attached to the image), which is then downloaded. Previously this file was kept and stayed attached to the image. Now it is removed after the download has finished.
Test: Make sure the export as JPEG/PNG/TIFF functionality still works and that the FileAnnotation with the zip file is deleted after it was downloaded.

--no-rebase


mtbc:
This does seem to work well, except that if `/tmp/Batch_Image_Export.zip` already exists then the download claims to have happened but I can't find the new zip anywhere.

jburel:
Comments added to PR #3346, zip name issue has been fixed.

mtbc:
Good to merge, or needs another review?

jburel:
Merging, was waiting on @dominikl to check if he wanted to do further work on the "zip naming front".

PR #3358, opened 2015-01-19
combine some SQL update statements into fewer

mtbc:
Attempts to reduce the number of times the same tables are updated, to decrease upgrade time. Read the new SQL carefully in case of mistakes. Try the upgrade with a database with data in it, both from 5.0 and from early 5.1.


joshmoore:
@sbesson : so now the question -- a `release-upgrade` like build for `merge-*`?

joshmoore:
Oops, this isn't mergeable.

joshmoore:
Waiting on the performance of https://ci.openmicroscopy.org/job/OMERO-5.1-merge-upgrade/VERSION=5.0.0,label=dogfish/198/console

joshmoore:
> new SQL carefully in case of mistakes.

Would code-generation of the upgrades have been useful? Would it still make sense to compare?

sql/psql/OMERO5.1DEV__17/OMERO5.0__0.sql
```diff
@@ -1649,10 +1668,6 @@ ALTER TABLE transmittancerange
 
 ALTER TABLE experimentergroup ADD COLUMN ldap BOOL NOT NULL DEFAULT false;
 
--- Set "ldap" value for each group to be false
-
-UPDATE experimentergroup SET ldap = false;
```
joshmoore: Where does this removal come from?

joshmoore: Ah, the `NOT NULL DEFAULT`.

mtbc:
I think the changes are small enough that by-hand-and-eye suffices for this PR but, in general, certainly for anything bigger, a script for generating the updates sounds like a fine idea.

mtbc:
(In my previous job where we'd usually drop the indices for the upgrade, we weren't doing the UPDATE-conflation that this PR does; it's plausible that either approach suffices.)

sbesson:
As per https://ci.openmicroscopy.org/job/OMERO-5.1-merge-upgrade/VERSION=5.0.0,label=dogfish/198/, the DB ugprade took 2h30 as opposed to 23hr from the latest OMERO on the same node (see https://ci.openmicroscopy.org/job/OMERO-5.1-release-upgrade/VERSION=5.0.0,label=dogfish/17/). Assuming the code changes make sense, the performance improvement is certainly significant enough to merge this PR.

joshmoore:
I've read over this multiple times and have found no issues. (Considering the fiddliness of the code, I may have overlooked something). I'm merging with the assumption that if there are any wrong defaults here, they will be found out during testing. Thanks, @mtbc.

PR #3362, opened 2015-01-20
maintain the namespace table

mtbc:
Now the `namespace` table in the database always contains at least all the namespaces currently used by any annotation. To test, try to violate that, whether by fiddling with the annotation table or the namespace table directly in the database, upgrading from an older version which already had plenty of annotations, or some other course of action. The system should frustrate your every attempt.

@jburel: Also adds name fields for annotations, namespaces, and ROIs. Annotation names should be searchable.

--no-rebase


sbesson:
From https://ci.openmicroscopy.org/job/OMERO-5.1-breaking-deploy/206/console
```
DB version (OMERO5.1DEV__18) does not match omero.properties (OMERO5.1DEV__19). Please apply a db upgrade.
See http://www.openmicroscopy.org/site/support/omero5/sysadmins/server-upgrade.html
*
```

Are we missing an upgrade path ?

joshmoore:
NB: I'm not sure we want to delete the NS when the last annotation is deleted. Especially not if it's been configured by the user but up for discussion.

components/server/src/ome/security/SystemTypes.java
```diff
@@ -78,6 +79,8 @@ public boolean isSystemType(Class<?> klass) {
             return true;
         } else if (IGlobal.class.isAssignableFrom(klass)) {
             return true;
+        } else if (Namespace.class.isAssignableFrom(klass)) {
```
joshmoore: Depending on the interpretation that's wanted, you could add `global="true"` to the mapping file and then this line wouldn't be necessary.

mtbc:
Last NS indeed oughtn't be deleted when the last annotation is deleted.

components/server/src/ome/security/SystemTypes.java
mtbc: The codegen still seems to generate some kind of annotation count table for the namespace but without then replacing it with a view. I'm guessing that globals aren't expected to be annotatable.

joshmoore: group is annotatable and global (for example): https://github.com/openmicroscopy/openmicroscopy/blob/v5.1.0-m3/components/model/resources/mappings/meta.ome.xml#L12

mtbc: Huh, I I'll have look into why we don't have a count table for that then.

mtbc: We do. While, say, `count_Image_annotationLinks_by_owner` is a view, `count_Experimenter_annotationLinks_by_owner` is a table. Is that okay: should I go ahead and make `Namespace` global and have a `count_Namespace_annotationLinks_by_owner` table?

mtbc: (I wonder if I could spot all such count tables and turn them to views in the PSQL footer script.)

joshmoore: Unless there's a reason not to from @jburel, @manics, etc., I'd go for global and we do whatever cleanup of those tables & views that's necessary/appropriate.

joshmoore:
Card looks well covered. The only question we didn't recently cover is whether or not a "bool automatic" field is wanted for those namespaces created by the trigger. (Or "source"?) /cc @jburel

mtbc:
Good question. I wonder if non-automatic namespace creation would be likely to leave some obvious difference in some other existing column anyway.

joshmoore:
I was thinking that an automatically created namespace would be hidden by default. Perhaps not enough though.

mtbc:
@joshmoore, I'm happy to implement whatever you and @jburel decide.

joshmoore:
Based on:
 * https://ci.openmicroscopy.org/view/Breaking/job/OMERO-5.1-breaking-upgrade/229/
 * https://ci.openmicroscopy.org/view/Breaking/job/OMERO-5.1-breaking-integration-python/177/
 * https://ci.openmicroscopy.org/view/Breaking/job/OMERO-5.1-breaking-integration-java/178/
Removing breaking.

joshmoore:
Note: decision was to not worry about the `automatic` or `triggered` field since that information can be semi-deduced from a combination of the presence or absence of an eventlog entry and a possibly null value in the `display` column.

joshmoore:
Based on https://ci.openmicroscopy.org/view/5.1/job/OMERO-5.1-merge-integration-java/351/ et al merging.

PR #3363, opened 2015-01-20
More roi fixes

jburel:
Fixes some issues with the ROI/measurement tool:
* Multi-Z and/or multi-T images: If the ROI tool was opened once, moving the Z and/or T slider continuously reopenend the ROI dialog, even it has been closed before. See https://www.openmicroscopy.org/qa2/qa/feedback/9527/  (this has already been reviewed by @aleksandra-tarkowska in #3070 , probably no need to review again)
* Wrong measurements with images with non-micron pixel sizes. Measurements were always calculated in micron size, even if the reference unit was nm. To test: Open an image with nm pixel size, draw a line and check that the distance measurement is correct, also check that unit and value is correct on the results tab. See https://www.openmicroscopy.org/qa2/qa/feedback/9560/
* Only PNG export was available via the context menu on the LineProfile chart (now PNG and JPEG is available - it uses the same action as the "Export" button); reported by @pwalczysko, see Trello card https://trello.com/c/h0PLjn5r/54-more-roi-fixes

I created a bit of a mess with my different ROI fix branches/PRs. So I moved all relevant commits into this PR and going to close #3070.

--rebased-from #3084


jburel:
Travis restarted

joshmoore:
247. Restarted.

pwalczysko:
Ad test2 in the header (``Wrong measurements with images....``):
The measurement is correct, but the header in the Results table of the Measurement tools clashes with the units in the rows (Angstrom in header, micrometers in the rows):

screen shot 2015-01-23 at 12 47 39

pwalczysko:
Ad https://github.com/openmicroscopy/openmicroscopy/pull/3363#issuecomment-71188797 - when I read the QA feedback, even if we remove the Angstrom unit from the column header, what the user is asking for is actually to thave the "proper" units displayed (= NOT micrometers)
``We have an EM image that has a physical pixel size of 1.502nm. If you draw a line in the measurement tool it reports the length in the wrong units (i.e. in um instead of nm) despite the correct units in the title of the column.``

pwalczysko:
Except the units problem all works fine here.

jburel:
@dominikl: will all the units work done on 5.1, it might be better to keep that one develop only
What do you think?

dominikl:
Probably. As these are just minor annoying bugs, a rebase might not be necessary. Unfortunately I also can't remember how this PR (respectively the original one) is related to unit PRs.

jburel:
Closing in that case.

PR #3372, opened 2015-01-21
Do not load graphs by default in getProjects/getDatasets

sbesson:
This PR should modify the default behavior of `getProjects()` and `getDatasets` to return unloaded graphs by default.

--no-rebase

### DIFF
--- a/components/tools/OmeroM/src/io/getProjects.m
+++ b/components/tools/OmeroM/src/io/getProjects.m
@@ -12,7 +12,8 @@ function [projects, datasets] = getProjects(session, varargin)
 %
 %   projects = getProjects(session, ids, loaded) returns all the projects
 %   identified by the input ids in the context of the session group. If
-%   loaded is False, the images attached to the  datasets are not loaded.
+%   loaded is false, the images attached to the  datasets are not loaded.
+%   Default: false.
 %
 %   projects = getProjects(session, 'owner', ownerId) returns all the
 %   projects owned by the input owner in the context of the session group.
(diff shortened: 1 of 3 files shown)


components/tools/OmeroM/src/io/getProjects.m
```diff
@@ -12,7 +12,8 @@
 %
 %   projects = getProjects(session, ids, loaded) returns all the projects
 %   identified by the input ids in the context of the session group. If
-%   loaded is False, the images attached to the  datasets are not loaded.
+%   loaded is false, the images attached to the  datasets are not loaded.
```
mtbc: double space

joshmoore:
https://ci.openmicroscopy.org/view/5.1/job/OMERO-5.1-merge-training/209/console looks good:

```
Listing projects and orphaned datasets
Found 1 projects
  Project 1: Project 1 (490) - 2 datasets
    Dataset 1: Dataset 1-1 (451)
    Dataset 2: Dataset 1-2 (452)
Found 2 orphaned datasets
  Orphaned dataset 1: Orphaned dataset 1 (453)
  Orphaned dataset 2: Orphaned dataset 2 (454)

Listing datasets
Found 4 datasets

Reading dataset with loaded images
Found 1 images in dataset 452 using getDatasets()
Found 1 images in dataset 452 using getImages()
...
```
The only caveat is that depending on the rest of my work for http://trac.openmicroscopy.org.uk/ome/ticket/5858 the invocations which take a `true` may eventually throw an exception if too many elements are present. Either we wait on that, or just keep it in mind.

sbesson:
I think the exception throwing is orthogonal to this change. At the moment a call with `true` with too many elements will likely result in the client hanging but if we have a well-defined exception it might be worth adding to the function header.

jburel:
@sbesson: will you do the potential change in this PR?

sbesson:
No strong preference. This mostly depends on the timeline for fixing 5858.

joshmoore:
There's something of a Hibernate blocker for 5858. Since that then (largely) depends on a Hibernate update, I'd say don't wait on that.

jburel:
With the large number of opened PRs, merging this one.
Improvement can be done later on.

PR #3373, opened 2015-01-22
Serverside omero client properties

dominikl:
~~Adds a ConfigService to the Registry through which server side stored omero client properties can be requested.~~ Loads the server side omero client properties into the Registry. As first example the name for the orphaned images folder is taken from the server side omero client properties

( Note: Also removes some unused code from DataServiceFactory )

Test:
Change the name of the orphaned images folder with `./omero config set omero.client.ui.tree.orphans.name "foo"` restart server and client, make sure the modified name is shown in the client (cross check with Web).

--no-rebase


joshmoore:
@dominikl : worth a chat on this?

jburel:
sorry forgot about that PR. @dominikl

dominikl:
Shall we have a call about this?

jburel:
@dominikl: will tomorrow work for you?

dominikl:
@jburel Sure, anytime tomorrow.

dominikl:
If this is a reasonable way to retrieve the properties from the server, the 'on hold' label could be removed. Any objections?

dominikl:
Because there was a merge conflict with #3506 , it would be good to check again, that connecting via Session ID still works.

jburel:
It works:
 2 things
 * are you going to add other properties e.g members, owners for users display
 * we no longer have ref to Service in the gateway. in theory we should be able to connect to other services. It might be easier to load the "config" map when we first create the session.

dominikl:
@jburel Ok, I'll overhaul this, so that the whole config map is loaded when the session is initialized.
Other properties: Is there a trello card about this (which properties exists, are supported/should be supported in the clients, etc.)?

jburel:
@dominikl see https://trello.com/c/llOSq41B/180-common-settings-on-the-serverside

joshmoore:
NB: @aleksandra-tarkowska also recently mentioned needing to load several properties at once. If `IConfig.getConfigValues(String prefix)` would substantially help, we could investigate.

dominikl:
@joshmoore Yes, a method like this is definitely needed, because I can't see another way, how I should get all properties at once. I have no idea how the `ome.logic.ConfigImpl` gets the data from database and/or properties file, so probably can't make this before the 'feature freeze' tomorrow.

joshmoore:
I'll try to look into this for tomorrow morning.

joshmoore:
See gh-3576

dominikl:
If I don't use a ConfigService wrapper for IConfigPrx I can't make the properties lookup via the Registry, because the Agent Registries are already initialized at the startup, but I can only load the properties after the user has logged in (session has been created). So where to put the properties (and look them up from), directly to the Container? @jburel

jburel:
you could still use the container registry
i.e. ```registry.bind(config_string, map)``` then link to agents registry in the ```DataserviceFactory```

dominikl:
--depends-on #3576

joshmoore:
Sorry, @dominikl : you'll need to cherry-pick https://github.com/joshmoore/openmicroscopy/commit/a6101e74a54c30e8852a435d0c6b3122626f642f to make travis happy.

aleksandra-tarkowska:
works as expected omero insightscreensnapz003

jburel:
@aleksandra-tarkowska: could you also test with no property set up? This led to a NPE reported by @chris-allan

dominikl:
Would be good if this could be merged, then I could start to tackle other tickets which need this functionality. Or is the merge postponed due to code/feature freeze for 5.1.0?

joshmoore:
I'm not sure about the NPE status from @jburel, but otherwise, no objections from me.

jburel:
Merging. @dominikl when adding support for other properties, add a ```if null``` check

PR #3374, opened 2015-01-22
Disable extra artifacts with package-extra=false

joshmoore:
The -javadoc and -sources packages are currently
built by default via the package targets dependency
on package-extra. By setting package-extra=false
either on the command-line or in etc/local.properties,
these sections can be disabled. On my system, that
takes a rebuild (without clean) from 5'45" to 3'30".

/cc @jburel @sbesson 

Testing
---------
The intent is that this change should improve the overall health of travis builds. If this build is green, then likely adding it to the regular 5.0 build is fine. If all the other builds remain green (incl. release!), then things are working well. Developers may want to set `package-extra=false` locally. Note: this may cause the daily merge to fail unexpectedly since nothing will have built javadocs until that point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,14 @@ before_install:
     - sudo pip install scc pytest
     - scc travis-merge
     - if [[ $BUILD == 'build-python' ]]; then travis_retry sudo pip install flake8; fi
+    - if [[ $BUILD == 'build-python' ]]; then ./components/tools/travis-build py-flake8; fi
 
-install:
-
-# retry the build due to:
+# retries the build due to:
 # https://github.com/travis-ci/travis-ci/issues/2507
+install:
+    - if [[ $BUILD == 'build-python' ]]; then travis_retry ./components/tools/travis-build py-build; fi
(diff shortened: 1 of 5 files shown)


mtbc:
It's kind of ironic that travis failed so badly here!

joshmoore:
@mtbc: it is! At least on this third try it's not failing on the blitz compile, so I'm hesitant to try that as a next step:

```
Compiling 155 source files to /home/travis/build/openmicroscopy/openmicroscopy/components/model/target/classes
```

joshmoore:
Note: ee0b58d only passed with a manual retry.

joshmoore:
This didn't seem to do any good for travis. To not disrupt 5.0.7, I'm going to close and rebase to develop. @mtbc, if it's of sufficient use there (in terms of speeding up your builds) we can re-evaluate opening this.

mtbc:
I'll certainly appreciate it on develop. :+1:  to rebasing.

joshmoore:
--rebased-to #3408

PR #3376, opened 2015-01-22
Add option to hide passwords in bin/omero config get

sbesson:
With this PR, the following command

```
$ bin/omero config get --hide-password
```

should return the server configuration and hide the values of all keys ending with `.pass` or `.password` (replacing them with '*'). Also check the unit tests are passing.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -52,7 +52,7 @@ def getprefs(args, dir):
     """
     if not isinstance(args, list):
         raise Exception("Not a list")
-    cmd = ["prefs"]+list(args)
+    cmd = ["prefs"] + list(args)
     return omero.java.run(cmd, chdir=dir)
 
 
@@ -136,6 +136,9 @@ class PrefsControl(WriteableConfigControl):
         get.set_defaults(func=self.get)
         get.add_argument(
             "KEY", nargs="*", help="Names of keys in the current profile")
+        get.add_argument(
+            "--hide-password", action="store_true",
+            help="Hide values of password keys in the current profile")
 
         set = parser.add(
             sub, self.set,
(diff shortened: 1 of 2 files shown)


manics:
Would it be better to invert the flag (default is to hide)?

components/tools/OmeroPy/src/omero/plugins/prefs.py
```diff
@@ -264,13 +267,18 @@ def get(self, args, config):
...
+        is_password = lambda x: x.endswith('.pass') or x.endswith('.password')
         for k in keys:
             if k not in orig:
                 continue
             if args.KEY and len(args.KEY) == 1:
                 self.ctx.out(config[k])
             else:
-                self.ctx.out("%s=%s" % (k, config[k]))
+                if (hide_password and is_password(k)):
+                    self.ctx.out("%s=%s" % (k, '*' * len(config[k])))
```
manics: This still reveals the password length. OTOH it's useful to know if the field is empty.

sbesson: Yes, I am not in favor of completely removing the line for debugging purposes. To further reduce entropy, I could use a fixed length, e.g. '*' *8

joshmoore: :+1: on a fixed length.

manics: How about `'*' * 8 if config[k] else ''`

joshmoore:
> Would it be better to invert the flag (default is to hide)?

Mostly a question of where it will be used. Anything internal would then need to know when it's asking for a pass. I think of this mostly as a cut-n-paste help from the documentation; other tools like `git-config` don't try to actively hide sensitive information.

manics:
Wouldn't anything internal be using `ConfigService`?

mtbc:
`test_prefs.py` passes locally for me.

mtbc:
Passwords do indeed get obscured when `--hide-password` is used, except if the password is an empty string. If that is okay, then good to merge.

sbesson:
Improvements can be appended to https://trello.com/c/DqRxeeWO/352-bin-omero-config-no-passwords

sbesson:
--no-rebase

PR #3366, opened 2015-01-20
32 bit and float (rebased onto dev_5_0)

jburel:
This is the same as gh-2955 but rebased onto dev_5_0.

----

Add support for  32-bits images (pixels type)
 * Import the following images. 
    * test_image_good/ubm
    * from skyking/gatan/chip/
   * from skyking/gatan/qa-7642
 * Check that you can view the images.
 * Open the images in ImageJ or Fiji using BioFormats to validate.

Images imported w/o this PR can be found under 
trout/ port 14064 (latest), user-1 private-1 Project imported 18-08-14 to test 32 bit

Add support for floating point image
To test this PR

make sure that the unit tests under server/test/omeis.providers.re.utests are green
Import floating point images. a selection can be found under squig/team/will/float0-data.zip
View the image in OMERO, modify some values e.g. pixels window.
View the same image in ImageJ.
Note that the pyramid generation for floating point images (>4kx4k)will not work. This will not be addressed in this PR.
                
--rebased-from #2955 
--rebased-from #3033

PR #3377, opened 2015-01-22
Float rendering web dev 5 0

will-moore:
This is PR #3039 rebased onto dev_5_0.

Some small changes to support floating point data ranges in the channel sliders in Preview tab and main image viewer.
To test:
- Open a float pixel-type image in Preview panel. Test various images with very small and large dynamic ranges.
 - Start positions of slider handles are Integers (see above) but sliding handles should give a reasonable number of floating point increments (100 currently for full range of slider).
 - Save will save floating point start/end to DB (although this can't be verified because of bug above).
 - Launch Image Viewer from preview panel should apply the current floating point slider values to the image viewer.

--rebased-from #3039

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5813,7 +5813,7 @@ class _ChannelWrapper (BlitzObjectWrapper):
         :rtype:     int
         """
 
-        return int(self._re.getChannelWindowStart(self._idx, self._conn.SERVICE_OPTS))
+        return self._re.getChannelWindowStart(self._idx, self._conn.SERVICE_OPTS)
 
     def setWindowStart (self, val):
         self.setWindow(val, self.getWindowEnd())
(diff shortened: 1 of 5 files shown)


dominikl:
The channel sliders are working with the float values; however the image I used for testing (user-3/read-only/test/EMD-2225.map) stays black, I guess this is because the backend changes for float images aren't included in current octopus build?

dominikl:
With PR #3366 merged floating point images are displayed (were black before), however the preview pane looks pretty empty (no channel slider, just the preview) and the channel edit pane in full viewer looks very different from develop branch. Is this expected? See screenshot.
screen shot 2015-01-28 at 08 50 08

jburel:
@dominikl: FYI in dev_5_0, the web preview is limited.

dominikl:
Ok, if this layout is expected, then this is good to merge. The channel slider in full viewer works as expected.

will-moore:
@dominikl Yes, that's expected. All the preview pane work and corresponding changes to main viewer are 5.1 only.

PR #3378, opened 2015-01-23
Backported UI support for float images

dominikl:
Gathered all UI changes for float images from develop branch together and rebased them to dev_5_0; this mainly affects the TwoKnobSlider.
Review: Check that the channel sliders work with int and float images. But also check other places where the TwoKnobSlider is used, e. g. the movie dialog.

To be able test this PR, #3366 should be included in the build.
/cc @jburel

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/util/player/MoviePlayerControl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/util/player/MoviePlayerControl.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.imviewer.util.player.MoviePlayerControl
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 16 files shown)


joshmoore:
Just restarted due to 247.

dominikl:
I started Travis several times by re-comitting my last commit, it always fails. Does anyone know what the problem is?

jburel:
Java was green,
failure on python front "The command "sudo pip install scc pytest" failed and exited with 2 during ."

dominikl:
Ok. @joshmoore  just confirmed, that the dev_5_0 PRs will be in the merge build anyway, so shouldn't be a problem then.

jburel:
@dominikl: user-3, looking at map file.
Full range should be turned off for double or float

jburel:
Full range is now disabled.
I tried to Set min/max and Reset and Save from menu in Data Manager. It did not work.

dominikl:
It actually works, but Set min/max doesn't update the thumbnail and Reset and save doesn't update the preview. Shall I try to fix that within this PR, and you exclude this PR (to get your PR merged) @jburel , or merge this one and I'll open another PR?

dominikl:
Actually, shouldn't set min/max also be disabled for float images?

jburel:
Why do you want to disable it?
Actually my PR does not conflict with yours.
It is probably to dangerous to fix it at the last minute

dominikl:
Ok. Ah no, forget about the min/max disabling, just got a bit confused.

jburel:
I was looking at a different PR, this one and #3406 are conflicting.

dominikl:
@jburel Is there still something left, I should do within this PR? But as you want to start the RC build today, it might be already to late...

jburel:
I think the other points are minor so I will not do anything more in this PR.
The major controls are in place so we can modify setting for float/double.
Merging

PR #3335, opened 2015-01-13
Add cp and cp_rm transfer implementations

joshmoore:
As suggested by Niko on the forums, cp and cp_rm
implementations make local transfer of files possible
even across filesystem boundaries. The trade-off is
of course that in the case of a failed transfer there
will be increased space used which will need to be
manually cleaned up.

See: https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7699

/cc @ximenesuk

PR #3379, opened 2015-01-23
Add cp and cp_rm transfer implementations (rebased onto dev_5_0)

joshmoore:
This is the same as gh-3335 but rebased onto dev_5_0.

----

As suggested by Niko on the forums, cp and cp_rm
implementations make local transfer of files possible
even across filesystem boundaries. The trade-off is
of course that in the case of a failed transfer there
will be increased space used which will need to be
manually cleaned up.

See: https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7699

/cc @ximenesuk


joshmoore:
--rebased-from #3335

ximenesuk:
This all looks good to merge.

joshmoore:
Thanks, @ximenesuk. Responding on https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7699 as well.

PR #3380, opened 2015-01-23
Copy bftools to bin/ with omero

joshmoore:
Briefly discussed with @melissalinkert this PR. In order to clean off of another branch and have a proper discussion, I'm opening as a standalone PR. The goal of this would be to allow testing of the the BF jars built with OMERO by running the standard bftools scripts under bin. E.g.:
```
$ cd dist
$ bin/omero version
5.1.0-m3-ice35-SNAPSHOT
$ touch a.fake
$ bin/bfconvert a.fake a.ome.tiff
a.fake
FakeReader initializing a.fake
[Simulated data] -> a.ome.tiff [OME-TIFF]
        Converted 1/1 planes (100%)
[done]
1.018s elapsed (22.0+496.0ms per plane, 424ms overhead)
```

Originally I had intended to write a wrapper plugin (`bin/omero bf` or similar) but this is a likely much simpler fix.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -748,7 +748,7 @@ To get started using Eclipse, execute "./build.py build-dev" and import the top-
     Top-level helpers (shared helpers are in components/antlib/resources)
     =====================================================================
     -->
-    <target name="build-dist" depends="init,copy-licenses,copy-history,copy-etc,copy-sql,copy-server,copy-client,update-version,create-workdirs"
+    <target name="build-dist" depends="init,copy-licenses,copy-history,copy-etc,copy-sql,copy-bftools,copy-server,copy-client,update-version,create-workdirs"
         description="Copy all components to dist/; called at the end of build-* targets"/>
 
     <target name="copy-licenses" depends="init">
@@ -778,6 +778,16 @@ To get started using Eclipse, execute "./build.py build-dev" and import the top-
                 <fileset dir="${omero.home}">
                     <include name="sql/**"/>
                 </fileset>
+            </copy>
+    </target>
+
+    <target name="copy-bftools" depends="init">
+            <copy todir="${dist.dir}/bin">
+                <fileset dir="${bioformats.comp}/tools"/>
+            </copy>
+            <chmod dir="${dist.dir}/bin" perm="a+x" includes="*"/>
+            <copy todir="${dist.dir}/bin">
+                <fileset file="${bioformats.comp}/artifacts/bioformats_package.jar"/>
             </copy>
     </target>
 


build.xml
```diff
@@ -811,6 +811,16 @@ To get started using Eclipse, execute "./build.py build-dev" and import the top-
             </copy>
     </target>
 
+    <target name="copy-bftools" depends="init">
+            <copy todir="${dist.dir}/bin">
+                <fileset dir="${bioformats.comp}/tools"/>
+            </copy>
+            <chmod dir="${dist.dir}/bin" perm="a+x" includes="*"/>
+            <copy todir="${dist.dir}/bin">
+                <fileset file="${bioformats.comp}/artifacts/bioformats_package.jar"/>
```
sbesson: I wonder if it's not possible to fix `config.sh` at the low-level Bio-Formats level to reuse the JARs already copied under `lib/server`. I will have a go at this if that makes sense.

joshmoore: Yup, had considered that. Would certainly save some space. That being said, it might be that having package.jar in with the clients will get more testing for it.

mtbc:
This works fine for me, I like it.

joshmoore:
@melissalinkert / @sbesson : alternative suggestions?

sbesson:
Sorry. Forgot to follow up on this thread. I tried to use the server JARs directly but this failed because the JARs are named differently between OMERO and Bio-Formats. For now the server size is increased from 209 MB to 227 MB with this change which I can certainly live with.

My only suggestion would be to include this library under `lib/` (or `lib/xxx`) instead of `bin/` and fix the low-level `config.sh` to allow the library to be found.

sbesson:
Additionally the permissions are lost during the server zipping process. While unzipping the merge server, all non-`omero` executables have `644` permissions. I suppose the `release-{zip,tar}` target needs to be modified to have all the content of `bin/*` with `filemode=755`

joshmoore:
Shall we just close? Leave open for testing in the merge build?

build.xml
joshmoore: Force pushing, which will hide this conversation.

mtbc:
Want to keep on hold for a while more?

joshmoore:
Yes. I'm hoping/assuming that @sbesson & co. will do some magic such that we don't need another copy of the jars in the zip.

sbesson:
Another related question here is whether we should bundle the Bio-Formats command-line tools and upload them to artifactory as part of the Maven deployment phase /cc @melissalinkert

melissalinkert:
Might not hurt to try, but I don't know how easy it is to deploy a Zip - I don't think we're doing that anywhere else at the moment.

https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Packaging suggests this may not be easy, though http://stackoverflow.com/a/6440903 claims to work.

joshmoore:
--rebased-to #4229

joshmoore:
This is no longer valid since the submodule is gone. We'd need to download bftools.zip from artifactory.

PR #3381, opened 2015-01-23
Fix ice.config modification when using bin/omero admin ports --prefix

sbesson:
Without this PR, calling `admin ports` appends `omero.port=PORT` to the `ice.config`. This results in the following calls:

```
bin/omero admin ports --prefix=1
bin/omero admin ports --prefix=1 --revert
```

will append the following tow lines

```
omero.port=14064
omero.port=4064
```

This PR fixes the `ice.config` modification logic in `bin/omero admin ports` by:
1- looking for an `omero.port=xxx` line in ICE_CONFIG and replacing it using the new port (as for the other configuration files)
2- if not present, looking for the `## omero.port=xxx` line which ships with the default `ice.config` and uncommenting it
3- if 2 is not applicable, appending `omero.port=xxx` as done previously

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/install/change_ports.py
+++ b/components/tools/OmeroPy/src/omero/install/change_ports.py
@@ -29,24 +29,30 @@ def line_has_port(line, port):
 
 
 def update_ice_config(f, f_glacier2, t_glacier2):
-    port_pattern = re.compile('^omero.port=%s$' % f_glacier2, re.MULTILINE)
-    port_pattern2 = re.compile('^## omero.port=\d+$', re.MULTILINE)
+    old_port = "omero.port=%s" % f_glacier2
+    new_port = "omero.port=%s" % t_glacier2
+    glacier_port_pattern = re.compile('^%s$' % old_port, re.MULTILINE)
+    general_port_pattern = re.compile('^omero.port=\d+$', re.MULTILINE)
+    commented_port_pattern = re.compile('^## omero.port=\d+$', re.MULTILINE)
 
     with open(f, 'r') as r:
         text = r.read()
 
-    (new_text, n) = port_pattern.subn("omero.port=%s\n" % t_glacier2, text)
+    (new_text, n) = glacier_port_pattern.subn(new_port, text)
     if n != 0:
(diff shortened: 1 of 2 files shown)


mtbc:
This seems to work fine and the tests pass.

If I try a few `--revert`s for wrong prefixes, then I get a growing number of `omero.port=4064` lines; want to fix that here too? Otherwise good to merge.

sbesson:
@mtbc: you mean the following scenario?

```
bin/omero admin ports --prefix=1
bin/omero admin ports --prefix=2 --revert
```
No this is not covered as we speak and it may be a good occasion to do so (with tests). Will push additional commits.

mtbc:
Right, that one. :+1:

sbesson:
Pushed.

mtbc:
Tests look good, e.g.,
```
test/unit/clitest/test_admin.py::TestAdminPorts::testRevert[True-None] PASSED
test/unit/clitest/test_admin.py::TestAdminPorts::testRevert[True-1] PASSED
test/unit/clitest/test_admin.py::TestAdminPorts::testRevert[True-2] PASSED
test/unit/clitest/test_admin.py::TestAdminPorts::testRevert[False-None] PASSED
test/unit/clitest/test_admin.py::TestAdminPorts::testRevert[False-1] PASSED
test/unit/clitest/test_admin.py::TestAdminPorts::testRevert[False-2] PASSED
test/unit/clitest/test_admin.py::TestAdminPorts::testFailingRevert[True] PASSED
test/unit/clitest/test_admin.py::TestAdminPorts::testFailingRevert[False] PASSED
```

mtbc:
That is working great. Only complaint now is RFE,
```
bin/omero admin ports --prefix=2
bin/omero admin ports --prefix=3
```
leaves them prefixed with 2. As far as I'm concerned, this is already good to merge.

sbesson:
Agreed, switching between prefixes requires an intermediate `--revert` step at the moment. We can always improve this behavior but its impact is quite limited.

joshmoore:
Assuming the `--revert` improvement has been captured, merging.

PR #3382, opened 2015-01-23
Remove emma from build system (See #994)

joshmoore:
Other code-coverage libraries (like http://www.eclemma.org/jacoco/)
are now available. We have not actively used emma for so long that
it is better to simply remove all references and re-add when the
time comes.

There should be no builds calling this code so removal should leave
all jobs green. /cc @sbesson


sbesson:
All builds remained green with this change. Merging.

joshmoore:
--no-rebase # Though I do have a dev_5_0 commit if anyone is interested.

PR #3383, opened 2015-01-23
Index map annotations (Fix #12318)

joshmoore:
In order for the name-value pairs attach to objects
to be searchable by Lucene, a handler for the map
annotations needed to be added.

To discuss/do:
 * [ ] Which other map columns to index?
 * [x] Whether or not "map.key" is useful?
 * [x] Add documentation to the developers/Modules/Search.html page (https://github.com/openmicroscopy/ome-documentation/pull/1113)

To test:
 * Add a map annotation to an object and search for either the key or the value.
 * Additionally, searching for "key:value" should also return the object.

### DIFF
--- a/components/server/src/ome/services/fulltext/FullTextBridge.java
+++ b/components/server/src/ome/services/fulltext/FullTextBridge.java
@@ -17,11 +17,13 @@ import ome.model.ILink;
 import ome.model.IObject;
 import ome.model.annotations.Annotation;
 import ome.model.annotations.FileAnnotation;
+import ome.model.annotations.MapAnnotation;
 import ome.model.annotations.TagAnnotation;
 import ome.model.annotations.TextAnnotation;
 import ome.model.annotations.TermAnnotation;
 import ome.model.core.OriginalFile;
 import ome.model.internal.Details;
+import ome.model.internal.NamedValue;
 import ome.model.internal.Permissions;
 import ome.model.meta.Event;
 import ome.model.meta.Experimenter;
@@ -207,6 +209,10 @@ public class FullTextBridge extends BridgeHelper {
                     FileAnnotation fileAnnotation = (FileAnnotation) annotation;
                     handleFileAnnotation(document, opts,
                             fileAnnotation);
+                } else if (annotation instanceof MapAnnotation) {
+                    MapAnnotation fileAnnotation = (MapAnnotation) annotation;
+                    handleMapAnnotation(document, opts,
+                            fileAnnotation);
                 }
             }
         }
@@ -352,6 +358,30 @@ public class FullTextBridge extends BridgeHelper {
                 add(document, "file.mimetype", file.getMimetype(), opts);
             }
             addContents(document, "file.contents", file, files, parsers, opts);
+        }
+    }
+
+    /**
+     * Creates {@link Field} instances for {@link MapAnnotation} named-value
+     * pair.
+     *
+     * @param document
+     * @param store
+     * @param index
+     * @param boost
+     * @param mapAnnotation
+     */
+    private void handleMapAnnotation(final Document document,
+            final LuceneOptions opts, MapAnnotation mapAnnotation) {
+
+        List<NamedValue> nvs = mapAnnotation.getMapValue();
+        if (nvs != null && nvs.size() > 0) {
+            for (NamedValue nv : nvs) {
+                if (nv != null) {
+                    add(document, nv.getName(), nv.getValue(), opts);
+                    add(document, "map.key", nv.getName(), opts);
+                }
+            }
         }
     }
 
--- a/components/server/src/ome/services/fulltext/FullTextBridge.java
+++ b/components/server/src/ome/services/fulltext/FullTextBridge.java
@@ -43,7 +43,7 @@ import org.hibernate.search.bridge.builtin.DateBridge;
  * Primary definition of what will be indexed via Hibernate Search. This class
  * is delegated to by the {@link DetailsFieldBridge}, and further delegates to
  * classes as defined under "SearchBridges".
- * 
+ *
  * @author Josh Moore, josh at glencoesoftware.com
  * @since 3.0-Beta3
  * @DEV.TODO insert/update OR delete regular type OR annotated type OR
@@ -81,7 +81,7 @@ public class FullTextBridge extends BridgeHelper {
 
     /**
      * Main constructor.
-     * 
+     *
      * @param files
      *            {@link OriginalFileServce} for getting access to binary files.
      * @param parsers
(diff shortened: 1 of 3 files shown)


components/server/src/ome/services/fulltext/FullTextBridge.java
```diff
@@ -356,6 +362,31 @@ private void handleFileAnnotation(final Document document,
...
+     * @param mapAnnotation
+     */
+    private void handleMapAnnotation(final Document document,
+            final LuceneOptions opts, MapAnnotation mapAnnotation) {
+logger().error("handling map annotation");
+        List<NamedValue> nvs = mapAnnotation.getMapValue();
+        if (nvs != null && nvs.size() > 0) {
+            for (NamedValue nv : nvs) {
+                if (nv != null) {
+logger().error("{}={}", nv.getName(), nv.getValue());
```
mtbc: "in progress" things to be removed before final merge, I guess

joshmoore:
Thanks, @mtbc. Pushed.

components/server/src/ome/services/fulltext/FullTextBridge.java
```diff
@@ -356,6 +362,31 @@ private void handleFileAnnotation(final Document document,
...
+                    add(document, nv.getName(), nv.getValue(), opts);
```
mtbc: I wonder if we want a standard prefix before map keys. (Even also using the map name, if it has a name?) So, they don't collide with other search field names.

joshmoore: Happy to entertain concrete suggestions. My thinking was that even a collision would still return proper search results. A prefix will require users to know to add it. *Shrugs* Both perhaps?

dominikl:
Search works like expected. I just wonder what the "map.key" is about.

joshmoore:
Using `[ NamedValue("foo", "bar") ]` as an example, *without* `map,key`

 * `foo:bar` would return a result and
 * `bar` (i.e. the default in `_combined`) would return a result

but `foo` alone would not. *With* `map.key`, the following all return:

 * `foo:bar`
 * `bar`
 * `foo`
 * `map.key:foo`

Is this useful?

dominikl:
Ah ok, yes, I think it makes sense; one would expect to find something by just searching for a name/key only.

joshmoore:
Would "has_key:foo" be better/clearer?

joshmoore:
@dominikl @pwalczysko @jburel @will-moore : any thoughts on this?

pwalczysko:
@joshmoore : Sorry, not sure about 2 things
  - does the user type this in the search box (e.g. ``has_key:foo`` ) ?
  - not sure what the difference between ``has_key:foo`` and ``map.key:foo`` really is (seems a bit the same ?)
  - searching for ``foo:bar`` (or rather in more realistic terms ``temperature:37`` is the main case to cater for in my opinion -> we shouild not make the user to have to know the whole syntax with the colon, the search for ``temperature 37`` or similar should give him something as well

will-moore:
If I search for ```foo=bar``` would I get key:value of ``` foo: blah ``` in my search results? If so, is it guaranteed to come AFTER ``` foo: bar ``` in the results?
What's the difference between searching for ``` foo:bar ``` and ``` foo=bar ```?

will-moore:
Do we need any changes on the client side for this PR? Any changes in how we parse queries?

joshmoore:
> Do we need any changes on the client side for this PR? Any changes in how we parse queries?

I don't think so. The only thing you would need to decide is if you ever want to work with `has_key`.

> If I search for foo=bar would I get key:value of foo: blah in my search results? If so, is it guaranteed to come AFTER foo: bar in the results?

Do you mean `foo=bar` or `foo:bar`? I haven't done anything to support the former. If you're saying the client will parse out the "=" and therefore this will actually search for "foo bar" then yes, you could also get results for "foo blah" and there's no guarantee as to the ordering, though in general, I'd assume more matches come earlier.

> What's the difference between searching for foo:bar and foo=bar?

The former is proper Lucene syntax. The second (if it works) is an odd by-product of our string parsing.

> does the user type this in the search box (e.g. has_key:foo ) ?

I'm not sure if that's allowed by the clients, but certainly on the CLI you could type this to find all images that have a certain key. (At the moment, `map.key:foo` though). Alternatively, just `foo` would also return those images.

> not sure what the difference between has_key:foo and map.key:foo really is (seems a bit the same ?)

Yes, it's just the question of which name is most understandable.

> searching for foo:bar (or rather in more realistic terms temperature:37 is the main case to cater for in my opinion -> we shouild not make the user to have to know the whole syntax with the colon, the search for temperature 37 or similar should give him something as well

I believe this to be the case, yes.

pwalczysko:
@joshmoore : Thanks. ``has_key:foo`` sounds more understandable.

joshmoore:
Pushed.

dominikl:
If I understood this right, the name "map.key" or "has_key" doesn't matter on the client side (just an issue on the backend), so there's nothing specific to test for that in the clients, right?
If so, this PR is good to merge. Checked the search functionality again; search results are like I would expect. And I can confirm, if I search for "foo:bar" (or "foo=bar", doesn't matter because it's reduced to "foo bar" in both cases), it will find results where key=foo and value=bar, as well as results where  just key or value=foo respectively where just key or value=bar; but results which contain both terms come first.

joshmoore:
Just to be clear: so there's currently no way to make use of "X:" from the clients at the momonet? i.e. it's parsed away?

dominikl:
Yes, at the moment all non-alphanummeric characters (except _ are replaced with a space). Ok that actually means I didn't understand this right, "has_key:xxx" is an expression the user should be able to type in order to search for a specific key only?

mtbc:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/test.integration.test_search/TestSearch/test_map_annotations/

joshmoore:
For me all the tests are now green. I pushed a few more to looking into https://github.com/openmicroscopy/openmicroscopy/pull/3383#issuecomment-74233445 (namely https://github.com/joshmoore/openmicroscopy/commit/9b22776709aa304e9dcbb6e0e17be93c8da237c7) which all seem to be fine. (Minimal) documentation has been opened. Assuming we don't want to start indexing more in this PR, I'd say it's good to go once the next run of the integration tests are green unless there are any questions/objections on the client-side.

joshmoore:
NB: re: "Which other map columns to index?" -- the only other 2 objects which would be useful to index onto the image would be the GenericLightSource and ImagingEnvironment. At the moment, though, we haven't done *anything* with those objects, so likely that would be more effort than we want to expend.

joshmoore:
Comment fixed. As I said, my assumption is that all these tests are green. If it's better to have this in m4, I'm happy to handle any test adjustments in subsequent PRs.

PR #3384, opened 2015-01-23
Disable companion attachments (Fix #12519)

joshmoore:
Annotations with namespace NSCOMPANIONFILE are no
longer needed since all the files in a single fileset
are grouped in the database and are parseable via
Bio-Formats. Clients can continue looking for old
companion annotations but no assumption should be
made that all images will have them.

### DIFF
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -780,58 +780,6 @@ public class OMEROMetadataStoreClient
         iUpdate.saveObject(file);
     }
 
-    /**
-     * Uses the {@link ManagedImportRequestI#reader reader} object to obtain
-     * a list of companion files from the current image. Resolves the name of
-     * the file into an OriginalFile and tries to link to the image in the DB.
-     * @param imageList A list of image to which companion files will be
-     *                  attached.
-     */
-    public void attachCompanionFilesToImage(Fileset fs, List<Image> imageList)
-        throws ServerError {
-
-        final String[] companionFiles = reader.getUsedFiles(true);
-        if (companionFiles == null || companionFiles.length == 0) {
-            return; // EARLY EXIT
-        }
-
-        final List<List<IObject>> linkBatches = new ArrayList<List<IObject>>();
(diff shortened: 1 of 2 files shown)


mtbc:
Well, this certainly seems to work fine. It does seem to mean that in the client I don't see a list of companion files any more; that's okay / expected?

joshmoore:
@mtbc : in which location? We might need to update a query to support that. /cc @jburel @dominikl @will-moore

mtbc:
It's over in the metadata panel on the right, maybe under the acquisition tab?

jburel:
Acquisition tab that's where they are displayed if any. I will have to review the annotation count (on-going issue)

joshmoore:
@jburel : let me know if you want me to track down what queries are currently being made.

jburel:
I will have a look at that next.

joshmoore:
@jburel : do you want me or someone else to look into something here?

joshmoore:
@jburel : ping. Still planning on this for 5.1.0?

jburel:
will work on it now that i have the imagej out of the way

jburel:
There is no assumptions that the image will have it since it was not the case prior to FS, i.e. only certain file format will have a companion file e.g. DV.
Checked web and insight, no error noticed.

joshmoore:
i.e. we should have these into the testing plan for 5.1.0?

jburel:
It won't hurt to have it check.
The thing to test is:
 * Import file like lei or a dv with a companion file and check the display
 * compare with previously imported file of similar type

cc @pwalczysko

jburel:
Just imported a file so it is no longer possible to view the log of a DV. This could be a surprise for user so we will have to handle that case cc @will-moore @gusferguson @pwalczysko

pwalczysko:
Works fine with dv. New display really does not have companion file, but no problems with the file detected.

joshmoore:
So an RFE re: displaying logs based on the fileset? Anything else?

jburel:
that should it. not a blocker for that PR since we can have access to the log file when the user downloads the image.

joshmoore:
Jotted down the namespace for 5.2 cleanup. Merging.

PR #3385, opened 2015-01-23
Fix jvmcfg unittests

sbesson:
As reported by @mtbc, the unit tests for the JVM configuration can vary depending on the level where they run, i.e. the following commands report different results

```
$ ./build.py -f components/tools/OmeroPy/build.xml test -DTEST=test/unit/test_jvmcfg.py
$ cd components/tools/OmeroPy && ./setup.py test -t test/unit/test_jvmcfg.py -v
```

This is due to one of the tests using the `server.jar` to calculate the system memory and not the other (because of a path issue). In all cases, some of the tests in the `testjvmcfg.json` have assumptions on the memory configuration of the system. This PR is monkeypatching `Strategy._system_memory_mb_java()` to return the default value. In a more advanced version of this PR, these settings could be made variable and also stored in the JSON.
Additionally, the fixtures are now properly named by settings the `ids` keyword.

To test this PR, check both top-level and OmeroPy level tests pass.

### DIFF
--- a/components/tools/travis-build
+++ b/components/tools/travis-build
@@ -34,7 +34,7 @@ build_java()
 
 build_python()
 {
-    flake8 -v .
+    flake8 .
     ./build.py build-default
     ./build.py -py test -Dtest.with.fail=true -DMARK="not broken"
     ./build.py -fs test -Dtest.with.fail=true -DMARK="not broken"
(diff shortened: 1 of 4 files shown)


components/tools/travis-build
```diff
@@ -34,7 +34,7 @@ build_java()
 
 build_python()
 {
-    flake8 -v .
+    flake8 .
```
mtbc: I assume this reducing verbosity is intended.

sbesson: Yes. Verbosity was useful to check what components were flake8'd but since we have now most of the Python code covered, this option results in more noise than necessary while trying to identify the cause of Travis failures (See commit message). With this change, flake8 failures should now render as https://travis-ci.org/openmicroscopy/openmicroscopy/jobs/48101273.
/cc @ximenesuk @joshmoore

mtbc:
Tests pass both ways. Good to merge.

PR #3386, opened 2015-01-24
Add extra event to return logfile id early.

ximenesuk:
This is a PR for @jburel's evaluation. An extra `POST_UPLOAD` event, `IMPORT_STARTED` has been added and propagated to give the logfile ID to the client as early as possible. Previously the ID was sent after the metadata import step. For direct testing the cli importer should show something like the following on the console:
```
...
2015-01-23 09:15:05,016 11413      [      main] INFO   ormats.importer.cli.LoggingImportMonitor - FILESET_UPLOAD_END
2015-01-23 09:15:05,614 12011      [      main] INFO   ormats.importer.cli.LoggingImportMonitor - IMPORT_STARTED Import started. Logfile: 152
2015-01-23 09:15:07,561 13958      [l.Client-0] INFO   ormats.importer.cli.LoggingImportMonitor - METADATA_IMPORTED Step: 1 of 5  Logfile: 152
...
```
For testing within Insight an image file that fails early on import would be useful. @pwalczysko may be able to provide details.

To test this PR, you can use the following image on squig
```
data_repo/test_images_broken/broken_images_scenario/fails_at_processing/wrong-detector-settings-mk1.ome
```

### DIFF
--- a/components/blitz/src/ome/formats/importer/ImportEvent.java
+++ b/components/blitz/src/ome/formats/importer/ImportEvent.java
@@ -410,6 +410,22 @@ public class ImportEvent {
     // These extra PROGRESS_EVENT classes are added to allow some meaningful
     // event reporting under FS rather than abusing the ones above
 
+    public static class IMPORT_STARTED extends POST_UPLOAD_EVENT {
+        public IMPORT_STARTED(int index, String filename, IObject target,
+                Long pixId, int series, ImportSize size,
+                Integer numDone, Integer total, Long fsId) {
+            super(index, filename, target, pixId, series, size, numDone, total, fsId);
+        }
+
+        @Override
+        public String toLog() {
+            StringBuilder sb = new StringBuilder();
+            sb.append(getClass().getSimpleName());
+            sb.append(String.format(" Logfile: %d", logFileId));
+            return sb.toString();
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


jburel:
@ximenesuk Thanks, I will try it

joshmoore:
My networks misbehaving this morning and so I've had some interesting effects:
 * download of log files taking a *very* long-time. The user doesn't get any sense of the fact that this could take a while. Perhaps a status somewhere or an incremental download?
 * I don't seem able to cancel the upload of tubhistwt.ome from `test_images_broken`
 * The order of which files start uploading is somewhat random, so that upper files are still in "Pending".

That being said, the code looks good, and I'm definitely getting log file IDs. I don't know if I had a failure at *precisely* the right point to prevent it from happening, but I'd think this is already an improvement. :+1:

mtbc:
When we do switch to compressing them, we may want to make the compressed version available for download so that decompression can happen client-side.

jburel:
@joshmoore: in general, we should notify that there is an on-going download (similar to view error)
This can be done in another PR as RFE.

joshmoore:
Agreed (...that compression is going to be lots of fun :smile:)

ximenesuk:
--no-rebase

PR #3388, opened 2015-01-26
Fix 'Save to All' in web preview panel

will-moore:
See: https://trac.openmicroscopy.org/ome/ticket/12697

Fixes the "Save to all" button in web preview panel.
To test:

 - Choose an image with other similar ones in the same Dataset
 - Edit rendering settings in Preview Panel and click "Save to all".
 - Other compatible images in the Dataset should be updated (thumbnails) and
    the rdef thumbnail in the Preview panel should also update.


gusferguson:
@will-moore 

Tested using http://trout.openmicroscopy.org/merge/webclient/ user-3 read-annotate-1 Safari Mac

Behaves as expected.
Good to merge.

PR #3389, opened 2015-01-26
remove unused and wrong import

mtbc:
Trivial fix to remove an emory import that wasn't used anyway.


joshmoore:
Rather than fight with travis to get this into the build, I'm simply going to merge it (crazy, I know!)

PR #3390, opened 2015-01-26
remove unused and wrong import (rebased from dev_5_0)

mtbc:
Trivial fix to remove an emory import that wasn't used anyway.

--rebased-from #3389


joshmoore:
Not even waiting for travis to yell at me/us. Looks good.

PR #3391, opened 2015-01-26
Fix edit name & description and jsTree update

will-moore:
See: https://trac.openmicroscopy.org.uk/ome/ticket/12666

To test:
 - Edit P/D/I name & description in web.
 - Check that jsTree updates appropriately
 - Check that name & description fields update correctly and don't get over-ridden by each other etc.


gusferguson:
@will-moore

Tested using http://trout.openmicroscopy.org/merge/webclient/ user-3 read-annotate-1 Safari Mac.

Save works as expected with name and description fields data tree updating correctly with name.
So over-writing/over-riding occurred even saving with both edit boxes active.
Good to merge.

PR #3392, opened 2015-01-27
Download menus web

will-moore:
Updates Download Menu. See https://trello.com/c/3n7Plb19/189-scripts-menus-download-options

To test:
 - Select an image and compare Download options with Insight and trello card above.
 - Also check Download menu with other objects selected P/D and SPW data.
 - Check menu with multiple Images and other objects selected (batch download).

screen shot 2015-01-27 at 11 43 43

screen shot 2015-01-27 at 11 49 35

 - Also check buttons in the Activities panel after running a script E.g. ThumbnailFigure on an image with a long name. (fixes: https://github.com/openmicroscopy/openmicroscopy/pull/2970#issuecomment-55227394)

screen shot 2015-01-27 at 11 32 26


gusferguson:
@will-moore

Tested using http://trout.openmicroscopy.org/merge/webclient/ user-3 read-annotate-1 Safari Mac and OMERO.insight-5.1.0-m3-755-21e9c32-ice35-b392-mac

There is no downloads menu button when a dataset is selected. 
See screenshot 1.

There are both an active and a greyed-out "Download Original Metadata" item appearing for both images that have metadata and those that don't.
See screenshots 2 and 3.

Downloading original metadata when there is none gives a text file with: 
[Global Metadata]
[Series Metadata] 
in it.
When there is metadata, that is as expected.

With multiple selection of images - "Download Original Metadata" only appears once and is greyed out whether there is metadata or not. 
See screenshot 4.

Activity window buttons and layout is fine with long names.
See screenshot 5.

omero insight

safari

safari

webclient

safari

will-moore:
@gusferguson Download is not supported in web for Projects or Datasets. Need to consider what to offer re: discussion on the dangers of Plate download in web: cc @chris-allan @jburel https://github.com/openmicroscopy/openmicroscopy/pull/3351

It's kinda tricky to test for the lack of Original Metadata when enabling or disabling menu items. As far as I can tell, the only way to tell if you've go no Original Metadata is to read all the original metadata and see how long it is. This is too expensive to do for every image we show in right panel.
Presumably not many users will try to download Original Metadata unless they expect it to contain something useful (won't try it for a png).

I'm not sure what an Original Metadata download for multiple images would look like, or if we should offer it. E.g. is it multiple text files in a zip? Probably best to disable it for both client with multiple images selected.

joshmoore:
Might be worth a brief discussion with @melissalinkert. We can certainly expand the API so you can get a `List` of `OriginalMetadata` responses. Possibly nulling if they are effectively empty?

will-moore:
Maybe something to consider for future work. If it can be made a very lightweight call then great, but we'd still have to access the original file on disk, right?
I don't think it's needed for this PR anyway.

joshmoore:
It'd certainly have to access the `.bfmemo` file on disk minimally.

will-moore:
@gusferguson I fixed the duplicate "Download Original Metadata" menu items. You OK with suggestions that we leave Project/Dataset download, disabling Original Metadata download (when none), and batch Original Metadata download for future work?

gusferguson:
@will-moore - makes sense - fine with that.

gusferguson:
@will-moore 

Behaves as expected.
Good to merge.

jburel:
Small point that you may want to fix in this PR.
In the publishing menu, each item launches a dialog so it should have ```...``` to be in synch with the rest.

will-moore:
Added ... (see screenshot).

screen shot 2015-02-04 at 11 29 41

jburel:
Waiting on travis to be green.

jburel:
Travis is green merging.

PR #3393, opened 2015-01-27
Sort users on last name in the webclient group dropdown

PaulVanSchayck:
The Users/Groups dropdown in the webclient is not sorted on anything at the moment. This will sort the list on last name.
I'm not sure this is the preferable way to fix this, or whether you would like this configurable. 

How to test: check whether the dropdown is sorted on last name.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2754,10 +2754,10 @@ class _BlitzGateway (object):
                 leaders = [self.getUser()]
             else:
                 colleagues = [self.getUser()]
-                
+
         colleagues.sort(key=lambda x: x.getLastName().lower())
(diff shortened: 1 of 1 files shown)


joshmoore:
Added `include` so this will make it into our test builds. Thanks @PaulVanSchayck. /cc @will-moore @aleksandra-tarkowska

mtbc:
Works well, thank you for the contribution.

Small suggestion: could sort first by first name, so it effectively becomes a secondary key for the last-name sort.

will-moore:
Code looks fine. I guess if you wanted to sort by first name as a secondary key, you could do something like:
```
colleagues.sort(key=lambda x: "%s %s" % (x.getLastName().lower(), x.getFirstName().lower()) )
```

sbesson:
Thanks @PaulVanSchayck. Merging.

sbesson:
--no-rebase

PR #3396, opened 2015-01-27
Split view figure, z/t plane

dominikl:
Fixes [Trac 12709](https://trac.openmicroscopy.org.uk/ome/ticket/12709)

Note: But this will introduce a slight inconsistency. Z/T is part of the rendering settings. With this PR you can now save changes to Z/T, but you cannot undo/redo them like the other rendering settings.

Test: 
Select a multi-Z/T image. Modify the Z and/or T plane (full viewer and/or preview pane). Make sure the save button is enabled, but don't save. Open the 'Split View Figure' dialog. Make sure that the preview shows the saved Z/T plane and not your current non-saved modification of the plane.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/actions/SaveRndSettingsAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/actions/SaveRndSettingsAction.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.imviewer.actions.SaveRndSettingsAction
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2008 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
@@ -61,7 +61,7 @@ public class SaveRndSettingsAction
     /** Sets the flag depending on the permissions of the group. */
     private void handleChange()
     {
-    	if (model.isOriginalSettings()) {
+    	if (model.isOriginalSettings(true)) {
     		setEnabled(false);
     	} else {
     		setEnabled(model.canAnnotate());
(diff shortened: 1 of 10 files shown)


mtbc:
Doesn't seem to be working for me; @dominikl will check what's going on.

dominikl:
For some reason, I thought yesterday it worked. The enabling of the save button should work, but the preview still shows the current settings and not the saved ones. Unfortunately this looks like one of these apparent simple problems, but which are nearly impossible to do. Maybe I have some kind of inspiration tomorrow, otherwise I'll just close this PR again. Or is it worth merging nonetheless, so that at least the save button is enabled?

dominikl:
No idea, how I can retrieve the original/unmodified rendering settings, so I just reset the settings in the FigureDialog, should do the job.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.1.0-m3-910-704dd72-ice35-b399-mac_Java6 trout user-3 read-annotate-1

If Z section or T point or both are changed then split view clicked without saving then the new Z and/or T are shown in the split view as expected.
See screenshot 1.

If settings saved with button or on closing - then re-opened - it shows  the new Z and/or T as expected.

But if Z and T are changed - not saved - split view done - image closed - "Don't save" selected - the new Z and T selections are still in fact "saved" and displayed in the Preview pane, although the thumbnails in the centre and preview pane revert to "Imported".
See screenshot 2.

If I then select "Imported" in the Preview pane, the preview view returns to imported (default) Z and T - but if I re-open in the Full Viewer I am still shown the changed Z and T in the Full Viewer.
See screenshot 3.

If I then click on "Imported" in the rendering settings pane of the Full Viewer - the image returns to "imported" but the sliders for Z and T do not return to the default positions although the counters do.
See screenshot 4. 

(in this image shown default was Z=3 T=1 and was reset to Z-4 and T=42 for the test).

It seems that the changes to try and get the desired behaviour has had unintended consequences to the saving and rendering in both the Data Manager Preview pane and the Full Viewer when reopening the image.

omero insight

user-3 user-3 connected to trout openmicroscopy org

omero insight

omero insight

dominikl:
After changing Z/T and always getting asked if the settings should be saved, I remember that it actually was a deliberate change, that Z/T changes don't trigger the 'dirty' state. Shall I just close this, but what about [Ticket 12709](https://trac.openmicroscopy.org.uk/ome/ticket/12709)?
/cc @jburel

jburel:
@dominikl: we will need to review that, I will close it for now and rethink it for 5.1.1

dominikl:
Yes, I agree. Sorry to everyone reviewing this PR, should have thought about this before starting the PR.

PR #3400, opened 2015-01-27
Add permissions tests for CLI tag plugin

sbesson:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/11628

This PR;
- moves the permissions fixtures to the common library and adds a `get_name()` method to return a short name used in fixtures (instead of `fixturexxx`)
- modifies the CLI tag creation commands to return the created object under the form `TagAnnotation:xxx` in the standard output
- adds try/except statement in the CLI `tag link` subcommand to handle permission exceptions
- adds permissions test for the `tag link` subcommand using the PermissionsFixture

### DIFF
--- a/components/tools/OmeroPy/test/integration/clitest/test_tag.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_tag.py
@@ -22,6 +22,8 @@
 import pytest
 import omero
 from omero.plugins.tag import TagControl
+from omero.cli import NonZeroReturnCode
+from library import PFS
 from test.integration.clitest.cli import CLITest
 from omero.rtypes import rstring, rlong
 from omero.util.temp_files import create_path
@@ -41,11 +43,16 @@ class AbstractTagTest(CLITest):
         self.teardown_mock()
 
     @classmethod
+    def new_tag(self, name=""):
+        tag = omero.model.TagAnnotationI()
+        tag.textValue = omero.rtypes.rstring("%s" % name)
+        return tag
+
+    @classmethod
     def create_tags(self, ntags, name):
         tag_ids = []
         for i in list(xrange(ntags)):
-            tag = omero.model.TagAnnotationI()
-            tag.textValue = omero.rtypes.rstring("%s - %s" % (name, i))
+            tag = self.new_tag("%s - %s" % (name, i))
             tag = self.update.saveAndReturnObject(tag)
             if ntags > 1:
                 tag_ids.append(tag.id.val)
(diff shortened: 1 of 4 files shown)


components/tools/OmeroPy/test/integration/clitest/test_tag.py
```diff
@@ -41,11 +44,16 @@ def teardown_method(self, method):
         self.teardown_mock()
 
     @classmethod
+    def new_tag(self, name=""):
```
ximenesuk: Is this method likely to be useful elsewhere?

components/tools/OmeroPy/test/integration/test_permissions.py
```diff
@@ -975,7 +887,7 @@ def assertPerms(self, perms, fixture):
         assert expected_arr == found_arr
 
     @pytest.mark.broken(ticket="12474")
-    @pytest.mark.parametrize("fixture", PFS)
+    @pytest.mark.parametrize("fixture", lib.PFS)
```
ximenesuk: Is there a reason for the difference in the change to parameterisation  between this case...

components/tools/OmeroPy/test/integration/test_permissions.py
```diff
@@ -993,7 +905,8 @@ def testProjectionPermissions(self, fixture):
         else:
             self.assertPerms(perms, fixture)
 
-    @pytest.mark.parametrize("fixture", PFS)
+    @pytest.mark.parametrize("fixture", lib.PFS,
```
ximenesuk: ...and this one?

components/tools/OmeroPy/test/integration/test_permissions.py
sbesson: No I should likely use `ids=` in this parametrization here as well if that makes sense.

components/tools/OmeroPy/test/integration/clitest/test_tag.py
sbesson: Hard to tell, this could arguably be moved to the upstream library but it is also a 2-line method.

ximenesuk: Yes, I did find a similar method in OmeroWeb tests but that works very differently. I guess it should stay here until it is used somewhere else.

ximenesuk:
Looks good overall. Should probably check the test builds (including broken) tomorrow before merging

ximenesuk:
The relevant tests look good, either passing or broken in the same way. Good to merge.

sbesson:
--no-rebase

PR #3402, opened 2015-01-28
Colorizer

joshmoore:
If the image for a pixels object has a CommentAnnotation in the namespace `openmicroscopy.org/ome/io/LUT` then the string value (`textValue`) will be passed to an `ImageJLutSource` and applied to the reader wrapper stack via a `Colorizer` instance. This will make it possible for the clients to manipulate the bytes which are returned from Bio-Formats before, e.g., rendering takes place.

--depends-on openmicroscopy/bioformats#1551

To test, add such an annotation to an existing image:
```
$ dist/bin/omero obj new CommentAnnotation ns=openmicroscopy.org/ome/io/LUT textValue=ice
CommentAnnotation:1
$ dist/bin/omero obj new ImageAnnotationLink parent=Image:1 child=CommentAnnotation:1
ImageAnnotationLink:1
```

It is also possible to add the annotation during import, in which case the colorization will be applied to the thumbnail (see https://github.com/openmicroscopy/openmicroscopy/pull/3402#issuecomment-75202493):
```
bin/omero import --annotation_text ice --annotation_ns openmicroscopy.org/ome/io/LUT ...
```


joshmoore:
Excluded due to https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/test.integration.test_exporter/TestExporter/testBasic/ etc.

components/romio/src/ome/io/nio/PixelsService.java
```diff
@@ -628,9 +633,14 @@ protected String getOriginalFilePath(Pixels pixels)
...
+        Map<String, String> params = resolver.getPixelsParams(pixels);
+        return getSeries(params);
+    }
+
+    protected int getSeries(Map<String, String> params)
+    {
         try
         {
-            Map<String, String> params = resolver.getPixelsParams(pixels);
             return Integer.valueOf(params.get("image_no"));
```
mtbc: This'll be changing, we'll conflict if we're not careful!

components/romio/src/ome/io/nio/PixelsService.java
```diff
@@ -639,6 +649,30 @@ protected int getSeries(Pixels pixels)
...
+     * Retrieves the series for a given set of pixels.
+     * @param pixels Set of pixels to return the series for.
+     * @return The series as specified by the pixels parameters or
+     * <code>0</code> (the first series).
+     */
+    protected LutSource getLutSource(Map<String, String> params)
+    {
+        String lut = null;
+        for (Map.Entry<String, String> param : params.entrySet()) {
+            if (param.getKey().startsWith("lut")) {
```
mtbc: I wonder if this is specific enough.

mtbc:
The code looks generally sensible. I'm not familiar with this, though, so there may be some conception of the API, and its usage, that I'm wholly missing, so I'd certainly welcome @jburel's second opinion here.

components/romio/src/ome/io/nio/PixelsService.java
joshmoore: Definitely *not*. This was awaiting our storage decision. It will be converted to an annotation with a proper namespace.

pwalczysko:
Note: This caused all-imports-failure on trout today. Maybe breaking ?

mtbc:
Looks like the --depends-on is important to review too, regarding API.

joshmoore:
@will-moore @dominikl @jburel : see the instructions in the description for what needs to happen for a LUT to be activated in the UI. At the moment (for some unknown reason), what displays with the LUT own is not significantly different from without, but in terms of the API that the clients can expect, this is what I have at the moment.

As this currently stands, the `CommentAnnotation` appears in the UI as "ice" (or whatever) and is deletable, but otherwise not editable. Also, there's no way for the user to add one from the UI.

Feedback welcome.

will-moore:
Your description says the "pixels object has CommentAnntoation" but in fact you mean the Image has the Comment on it?
So, the clients should not display annotations with the ```openmicroscopy.org/ome/io/LUT``` namespace?
I'm guessing that "ice" is the name of an ImageJ lookup table?

jburel:
```
addString("ns", "openmicroscopy.org/ome/io/LUT"));
```
should be replaced by 
```omero.constants.namespaces.NSLOOKUPTABLE.value```
if we keep the name space in Constants.ice

jburel:
@will-moore: annotation is linked to the image. cf. code

jburel:
could use ```CollectionUtils.isNotEmpty``` instead of ```rv != null && rv.size()```

joshmoore:
> Your description says the "pixels object has CommentAnntoation" but in fact you mean the Image has the Comment on it?

Sorry, yes. I forgot when writing the description that we took annotations off Pixels. (Description updated)

> So, the clients should not display annotations with the openmicroscopy.org/ome/io/LUT namespace?

If that's possible. Alternatively, I move to a different type of annotation (XML? map?)

> I'm guessing that "ice" is the name of an ImageJ lookup table?

Yes.

> should be replaced by omero.constants.namespaces.NSLOOKUPTABLE.value

Those are available in the internal server code.

jburel:
Sorry my question was more: Are we going to keep them in Constants.ice?

joshmoore:
@jburel : for use in the clients, yes. Constants.ice is great.

components/romio/src/ome/io/nio/PixelsService.java
joshmoore: Yup, we already are. i.e. waiting on #3446

jburel:
All annotations are now displayed in the clients so it does not matter which one is used. I am wondering if we will need more granularity to the ns 
some LUT could be hidden i.e. part of the original image, some LUT will be available to users in UI.

components/romio/src/ome/io/nio/PixelsService.java
mtbc: https://ci.openmicroscopy.org/job/OMERO-5.1-merge-deploy-win/ is now green at least.

melissalinkert:
I suspect that at least part of the problem is that ```Colorizer``` is higher up in the stack than ```ChannelSeparator```, so reversing that order might be a place to start.  That may require overriding ```getSizeC()``` in ```Colorizer``` to return 3 times as many channels when a non-null ```LutSource``` is set.

If that doesn't work, the only other thing I can think of off-hand is to add ```Colorizer``` to the import reader stack, and set the ```LutSource``` to ```new ImageJLutSource("grays")```.  It wouldn't completely surprise me if there was some weirdness with having the two reader stacks out of sync.

The annotation does seem to be detected and correctly turned into a ```LutSource```, though, so that part of the PR at least should be fine.

joshmoore:
Results of moving `Colorizer` to the bottom of the stack:
 * showinf seems to like the new setup
 * bfconvert fails with `ArrayIndexOutOfBoundsException: 2359296`
 * OMEROWrapper still shows no results

I'll try `grays` now and see where that gets me, then push everything for a discussion later.

joshmoore:
Reconfiguring the `ReaderWrapper` stack to consist of:
 * Memoizer
 * Colorizer (except during metadata import)
 * ...rest...
solved all display issues:

screen shot 2015-02-20 at 08 57 59

jburel:
@joshmoore: looks great

will-moore:
I've added the 'ice' annotation as described above to image https://trout.openmicroscopy.org/merge/webclient/?show=image-9487 (user-3) but don't see any difference in web or insight.

From the discussion above, it seems that if you apply a LUT to a single channel images, you will get a 3-channel RGB image in the clients? Is that right?
Seems that really you'd still want a single channel image with a different appearance on rendering.

joshmoore:
Ah, sorry, @will-moore. Should have mentioned that this only works on int8 data at the moment. I'll look into tackling int16 next. As for the number of channels, this is related to:

 * http://trac.openmicroscopy.org/ome/ticket/9789 and
 * http://trac.openmicroscopy.org/ome/ticket/9810

which we'll need to discuss.

will-moore:
OK, I tried it on a RGB jpeg I just imported, with unexpected results (see screenshot). Then I tried it on a single-channel version of this image created with the Channel_Offsets script https://trout.openmicroscopy.org/merge/webclient/?show=image-44906, but I didn't see any Colorizer effect on it.

screen shot 2015-02-23 at 11 28 20

will-moore:
Will there be any way to choose a LUT per channel instead of per image? E.g. If I want to apply LUT to a fluorescence channel and still have a transmitted light channel as greyscale etc.

joshmoore:
Re: differing by channel, the 2 options I can think of at this point are:
 * move to map annotation and have some form of syntax for specifying per channel (will the user need help?)
 * actually allow adding annotations to a channel from the UI.

will-moore:
What level of usability are we aiming for in the clients? As far as I can see this is not going to be usable at all in the clients without some changes (you can't even specify a namespace in the clients).
The most usable option would be to have a "LUT picker" for each channel (in the same way as we have colour pickers for each channel).
Does the rendering engine support LUTs per channel? What's going on with the 3-channel image with LUT above?

joshmoore:
This is all pre-rendering engine. It works on the bytes returned directly from the filesystem.

will-moore:
Is it not possible to do this just in the Rendering Engine?
E.g. I imagine you'd want to change / toggle LUTs on a channel without having to reset the RE. Also as mentioned above, we want the RE, client UIs etc to know that this is actually a 1-channel image, not a multi-channel image simply because of an applied LUT.

pwalczysko:
Agree with @will-moore here. As it was presented to me, it looks like we are heading toward having out of 1-channel image a 3-channel image just because it has an applied LUT. This will hardly be acceptable as a workflow in UI.

joshmoore:
Excluded to let a ML-PR through. We'll probably need to have a longer discussion on this anyway.

joshmoore:
Closing until we have more time for this.

PR #3403, opened 2015-01-28
Misc cleanups

joshmoore:
While looking into https://github.com/openmicroscopy/openmicroscopy/pull/3402, there were a number of (confusing) uses of `OMEROWrapper` along with various whitespace issues which I cleaned up on that branch. Rather than complicate 3402, I'm opening them here as a separate PR. Other than the one question in b5ded34c, this should have no impact.


mtbc:
Looks good to me, and not yet noticed it seeming to cause problems through being in the merge.

PR #3404, opened 2015-01-28
OMERO model object reference generator

mtbc:
Introduces a standalone component of OMERO.server that generates the model object reference as at https://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Model/EveryObject.html. This is an initial pull request to get the code into the repository. In the longer term it will probably have to be modified to ease the job of bringing this functionality into an automatic documentation update job. It can be run by configuring the server with `bin/omero config set`, and its database, appropriately such that it can start but, instead of starting it, do something like,
```
java -cp lib/server/\* `bin/omero config get | awk '{print"-D"$1}'` ome.services.graphs.GraphPathReport EveryObject.txt
```
and the `EveryObject.txt` Sphinx document will be generated, without an introductory paragraph.

### DIFF
--- a/components/server/src/ome/services/graphs/GraphPathBean.java
+++ b/components/server/src/ome/services/graphs/GraphPathBean.java
@@ -77,7 +77,7 @@ public class GraphPathBean extends OnContextRefreshedEventListener {
      * @author m.t.b.carroll@dundee.ac.uk
      * @since 5.1.0
      */
-    private static class PropertyDetails {
+    /* private GraphPathReport */ static class PropertyDetails {
         final String holder;
         final List<String> path;
         final Type type;
--- a/components/server/src/ome/services/graphs/GraphPathReport.java
+++ b/components/server/src/ome/services/graphs/GraphPathReport.java
@@ -131,7 +131,7 @@ public class GraphPathReport {
      * @return a Sphinx label for that class
      */
     private static String labelFor(String className) {
-        return Joiner.on('.').join("omero.model", className);
+        return Joiner.on('.').join("Hibernate definition corresponding to class omero.model", className);
     }
 
     /**
@@ -140,7 +140,7 @@ public class GraphPathReport {
      * @return a Sphinx label for that class property
      */
     private static String labelFor(String className, String propertyName) {
-        return Joiner.on('.').join("omero.model", className, propertyName);
+        return Joiner.on('.').join("Hibernate definition corresponding to property omero.model", className, propertyName);
     }
 
     /**
--- a/components/server/src/ome/services/graphs/GraphPathReport.java
+++ b/components/server/src/ome/services/graphs/GraphPathReport.java
@@ -67,9 +67,10 @@ import ome.tools.hibernate.ListAsSQLArrayUserType;
 
 /**
  * A standalone tool for producing a summary of the Hibernate object mapping for our Sphinx documentation. One may invoke it with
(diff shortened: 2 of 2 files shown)


sbesson:
My prefrence here would to create a dedicated CLI subcommand which can handle the classapth and all the configuration retrieval (like `admin reindex`). Any input on the name?  `bin/omero obj graphreport` ?

mtbc:
That would be great. Your suggested name looks good too.

joshmoore:
I'm for the command and name. A minor RFE it would be good if the description were so formatted that I knew what could sanely be passed to `bin/omero obj`

rleigh-dundee:
The generated code looks pretty good.  I did notice a minor issue with the link names: https://gist.githubusercontent.com/rleigh-dundee/b2e9fb1baf2660f4c71f/raw/5a0247520816921098cf216b61a662f5ec7c3a59/test%20rst

The "Used by" section includes suffixes on the names e.g. `DatasetAnnotationLink.child`, `.group` etc.  Is that intentional?

The Properties are using "|" as the leading symbol.  This is fine now.  If more information is ever added then a description list or table layout might be a useful improvement to make (but not needed now).

I would also suggest that the anchor names should have a prefix such as `db_table`, `java_class` or something else unique to prevent them conflicting with links elsewhere in the documentation.  For example, other language bindings which might use similar names.

mtbc:
The suffixes are definitely intentional, to make it easy to find the relevant propert{y,ies} after following the link. (I would have linked directly to the properties if I didn't need extra blank lines among them to achieve that.)

Is it possible to make link names page-local? (Probably not. Half of my effort in this was in battling Sphinx.)

I have no idea what's going on with the link names in your gist. Any suggestions?

rleigh-dundee:
The :source: link names are broken, probably because I generated a standalone test page where this extension wasn't valid.

rleigh-dundee:
I don't think page-local names are possible since they are accessible for cross-referencing anywhere in the whole document.

mtbc:
Oh! Okay. Thanks. You don't think the `omero-model-` anchor prefixes are unique enough? They are corresponding to the OMERO omero::model::whatever objects in Ice, hence the prefix.

mtbc:
(The Hibernate mapping is generated from the OMERO model, the DB has extra stuff that Hibernate doesn't know about.)

rleigh-dundee:
@mtbc That's fine I think.  I was looking at `.. _omero.model.AcquisitionMode:` and similar and thinking that might be shared with python, so could use some additional unique qualifier.

rleigh-dundee:
One other thought was that if we are (or could) use COMMENT ON to describe all tables, fields, etc. in the database, these could be pulled out and used to document the various properties and classes in the generated docs.  Assuming hibernate can access them.  That's outside the scope of this PR, just a thought for the future.

mtbc:
If the `omero.model` code generation could put these comments into a Java runtime annotation on classes and methods, I could try to reflect on those in this class.

mtbc:
@rleigh-dundee: I hope this improves the anchor names suitably.

rleigh-dundee:
@mtbc I don't think sphinx will like the spaces here.  Something like `hibernate_def_` would do fine.

mtbc:
It seems to actually work okay?

rleigh-dundee:
Just tested, and surprisingly it seems to work for both HTML and LaTeX (it normalises the name with hyphens).  It's still a little wordy for use in cross-references outside this autogenerated page though, if it's going to be referenced from the developer docs.

rleigh-dundee:
This looks fine and the generated code also looks fine.

PR #3405, opened 2015-01-28
OMERO.web integration tests

sbesson:
Following the work of https://github.com/openmicroscopy/openmicroscopy/pull/3318, this PR refactors the OMERO.web integration tests to make use of the new features of the `ITest` library which uses the `setup_class/teardown_class` strategy to create connections. Main changes are described below

- a local `weblibrary.py` is created defining the `IWebTest` class derived from `ITest` which initializes Django test clients
- all redundant fixtures (`itest`, `client`, `update_service`) are removed
- all other fixtures are moved as methods of the test class
- `test_settings.py` is moved to the unit tests and should be run by Travis

This PR should:
- not affect the outcome of any test i.e. all integration tests should still be passing
- increase completion time for the entire test suite especially for tests re-using the same client/connection like `test_show.py`

/cc @will-moore @aleksandra-tarkowska @chris-allan 

--no-rebase


ximenesuk:
I've not read every line, I mainly looked at the library class, but on the whole it looks okay. Assuming tests are allhappy then I'm happy.

sbesson:
Last 2 commits should fix the  regression introduced in https://github.com/openmicroscopy/openmicroscopy/pull/3191 so that:
- the Web integration tests can use the import functionalities of the library like `importSingleImage()`
- `OMERO_HOME` is still properly set so that the internal `settings.py` logic can still pick the correct server list 

To test these last 2 changes, 
- start your local server on a non-default prefix (`bin/omero admin ports --prefix 1`) 
- run the Web integration tests
- check that `TestSimple.testSingleImage()` 
- check that all tests dependent on the value of `omero.web.server_list`, e.g. `TestCsrf` are passing

ximenesuk:
Everything looks good to me. I don't know if this needs review by other webby people or not.

will-moore:
@ximenesuk If you're happy, I'm happy ;)

sbesson:
--no-rebase

PR #3407, opened 2015-01-28
Split OS X packaging into 2 sets

sbesson:
Follow up of https://github.com/openmicroscopy/openmicroscopy/pull/3337, this PR splits the zip bundles for the insight/importer/editor clients and the clients bundles for Mac OS X into two: Java 6 and Java 7+. This should result in twice more OSX client artifacts but reduce the size of these artifacts by a factor of 2.

The zip bundles are now suffixed by `_Java6` or `_Java7+`. So do the applications stored in them. Alternatively we could only suffix the zip bundles and preserve `OMERO.insight`, `OMERO.importer`, `OMERO.editor` as the app name since we do not bundle different Java versions together.

### DIFF
--- a/components/insight/build/dist.xml
+++ b/components/insight/build/dist.xml
@@ -848,22 +848,33 @@
   </jarbundler>
   <property name="dist.zip.prefix.importer.osx"
             value="${distImporter.bundle.name}-${dist.bundle.version}-mac"/>
-      <zip destfile="${dist.dir}/${dist.zip.prefix.importer.osx}.zip">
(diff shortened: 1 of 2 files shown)


dominikl:
Works like expected, there is now a separated zip for Java6 and Java7+ for each Mac client; see octopus artifacts, but also tested locally with build target 'release-clients'.

joshmoore:
Now onto the downloads page! :smile:

sbesson:
--rebased-to #3417

PR #3414, opened 2015-01-29
Add debugging of NPE in InteractiveProcessorI (rebased onto develop)

joshmoore:
This is the same as gh-2309 but rebased onto develop.

----

Intended for tracking down the problem on gretzky.
May not need to be merged.

cc @manics


joshmoore:
--rebased-from #2309

PR #3416, opened 2015-01-29
Fix _execute_child on Windows Python 2.7+

manics:
Copied from https://github.com/mozilla/addon-sdk/pull/1379
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12320

Testing: Scripts should now work on Windows with Python 2.7 (previously they only worked on 2.6). CI server is still being setup so you'll have to test on a local VM (or wait...).

### DIFF
--- a/components/tools/OmeroPy/src/omero_ext/killableprocess.py
+++ b/components/tools/OmeroPy/src/omero_ext/killableprocess.py
@@ -103,7 +103,12 @@ class Popen(subprocess.Popen):
 
     if mswindows:
         def _execute_child(self, *args_tuple):
-            # workaround for bug 958609
+            # The arguments to this internal Python function changed on
+            # Windows in 2.7.6
+            # - https://trac.openmicroscopy.org.uk/ome/ticket/12320
+            # Upstream bug and fix:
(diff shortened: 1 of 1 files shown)


manics:
--on-hold

chris-allan:
The commit message is certainly thorough but it would probably be good to link to the bug, fix and rationale explicitly, with URLs, in the actual code.

Fix itself looks good but will obviously require Windows testing which I won't be able to provide.

manics:
@chris-allan comment added

chris-allan:
Looks great.

:+1:

rleigh-dundee:
Change looks fine itself.  I can't run the server on my system since it's all set up to use virtualenvs and the server can't use them on Windows due to the way it starts up as a service and requires everything to be in the global environment for the system.  (Couldn't we allow it to run a script to customise the service environment at startup?)

joshmoore:
> (Couldn't we allow it to run a script to customise the service environment at startup?)

Yes, and I think something similar is what we need for Docker anyway.

But merging this.

manics:
--no-rebase

PR #3423, opened 2015-02-02
Increase web zooming rate to 5% for non pyramid images

PaulVanSchayck:
For us, the zooming rate of 1% for non pyramid images in the webclient is agonizing slow. It's much slower in scroll zooming of pyramided files (which for us increase by rate of 20% per zoom). 

This increases that rate to 5%.


aleksandra-tarkowska:
Thanks @PaulVanSchayck for highlighting the issue. Although, that parameter should be configurable from omero properties to allow others to specify their own step.

sbesson:
Discussed with @aleksandra-tarkowska, she will take over these changes and make the zooming rate configurable via the OMERO properties. Closing this PR.

PR #3424, opened 2015-02-02
Remove OMERO-homebrew-install.sh script

sbesson:
The Homebrew installation script has been migrated to the
https://github.com/ome/omero-install repository.


sbesson:
--rebased-to #3425

PR #3427, opened 2015-02-02
Omero to imagej

jburel:
To test this PR
 * Update the omero-imagej plugin
 * Open an image in imageJ
 * Draw few ROIs and add them either as overlays or to the ROIManager
 * Click on the Save ImageJ results button (in DataManager toolbar)
 * Check that The rois are saved and linked to the image.

Test 2
 * Open an image in ImageJ (not already imported)
 * Click on the Save ImageJ results button.
 * A dialog should pop up asking the user wants to import the image first. 

cc @bramalingam @pwalczysko

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/MainIJPlugin.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/MainIJPlugin.java
@@ -52,6 +52,7 @@ import javax.swing.event.HyperlinkListener;
 
 
 
+
 //Third-party libraries
 import ij.IJ;
 import ij.ImageJ;
@@ -60,11 +61,13 @@ import ij.plugin.PlugIn;
 
 
 
+
 //Application-internal dependencies
 import org.openmicroscopy.shoola.env.Container;
 import org.openmicroscopy.shoola.env.LookupNames;
 import org.openmicroscopy.shoola.env.config.PluginInfo;
 import org.openmicroscopy.shoola.env.data.DataServicesFactory;
+import org.openmicroscopy.shoola.env.event.SaveEvent;
 import org.openmicroscopy.shoola.env.init.StartupException;
 import org.openmicroscopy.shoola.env.log.LogMessage;
 import org.openmicroscopy.shoola.util.ui.MacOSMenuHandler;
@@ -234,6 +237,7 @@ implements PlugIn
         String home = "";
         String configFile = null;
         int index = LookupNames.IMAGE_J;
+        boolean save = false;
         if (args != null) {
             String[] values = args.split(" ");
             List<String> l = new ArrayList<String>();
@@ -241,8 +245,12 @@ implements PlugIn
                 String v = values[i];
                 if (v.startsWith("imageJ")) {
                     String[] k = v.split("=");
-                    if (k.length == 2 && k[1].equals("import")) {
-                        index = LookupNames.IMAGE_J_IMPORT;
+                    if (k.length == 2) {
+                        if (k[1].equals("import")) {
+                            index = LookupNames.IMAGE_J_IMPORT;
+                        } else if (k[1].equals("save")) {
+                            save = true;
+                        }
                     }
                 } else l.add(v);
             }
@@ -260,6 +268,10 @@ implements PlugIn
         }
         try {
             container = Container.startupInPluginMode(home, configFile, index);
+            if (save) {
+                container.getRegistry().getEventBus().post(
+                        new SaveEvent(LookupNames.IMAGE_J));
+            }
             attachListeners();
         } catch (StartupException e) {
             showMessage(e.getPlugin());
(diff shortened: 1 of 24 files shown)


bramalingam:
Everything works as expected. 

The 'Save ImageJ results button' in the data manager toolbar, is the only one that might need some reviewing. Cant it be part of the ImageJ Plugin's menu itself (under the OMERO list)? That would make the workflow more intuitive and keep all options under the same menu. This would also allow the user to do all his work under the ImageJ menu's instead of switching between Insight and ImageJ.

Just a suggestion, please correct me if am wrong.

jburel:
@bramalingam: Option added to the plugins menu.

pwalczysko:
Works, but I have problems with multi-point tool in ImageJ - after saving this into OMERO, the multipoints are lost. The singly done points are okay. Also there is an entry in ROI Manager for the multipoint ROI, but it has just one shape. 

  - create a single point ROI in IJ on image opened from OMERO and add to overlay with red line and thicker line width
  - crteate also multipoint ROI in IJ (second option on the point region tool) and add to overlay (the same properties as above)
  - use the button in Insigth ``Save Results from ImageJ`` 
  - other regions are okay, with the properties preserved, just the multipoints are lost (compare screenshots below, image is taken in IJ, the ROI does not list the multipoints as multishapes, and these are not displayed in the image in OMERO)

screen shot 2015-02-04 at 12 16 03
screen shot 2015-02-04 at 12 16 15

pwalczysko:
Note that the multipoint case seems to be quite important, as this was the first thing Graeme started to do when I told him to start an analysis in IJ (he was marking cells in the image).

pwalczysko:
Aha, web is showing them right - probably the multi-shapes issue in Insight

screen shot 2015-02-04 at 12 23 26

jburel:
@pwalczysko: multiple shapes belonging to the same roi won't work in insight. This will require some major changes in the measurement tool. Not doable for 5.1

pwalczysko:
@jburel @bramalingam : I guess it would be worth to create one roi per point in the meantime.Just tried to do multiple points in the image by using the single point tool, and it is not very pleasant - you have to click every time to add the point to ROI manager (or overlay)

pwalczysko:
It works really nicely. 

Two point for discussion: 
  1. How about a message saying ``2 regions were saved ?`` - I got sometimes confused about whether I have really clicked the menu item, there is no immediate reaction in OMERO insight either (need to refresh to update the ROI counter in Insight)
  2. Repeated reopening of the same image in IJ. When Overlay workflow is used, all is perfect. When ROI Manager workflow is used, all works as expected, but there is a bit of an issue with duplication of regions - I have no warning, I just add some more regions, and the old regions will be "duplicated" because they will be in the ROI Manager in IJ in the list as well. Not sure what is the best solution, I undertand that this happens because the regions get automatically listed in the ROI manager. Strictly a workflow issue.

pwalczysko:
@gusferguson ? - see point 2 in https://github.com/openmicroscopy/openmicroscopy/pull/3427#issuecomment-72851332 please

jburel:
if we decide to go that way, this won't be done in this PR.
We will have to unify various parts of the code.

jburel:
Point 2 is a problem of the ROI manager itself. Not sure we want to go there.

jburel:
Point 1: I was debating maybe keeping the saving dialog up/ having  a Save spinner.
For the message, if we have multiple IJ windows opened, I will have to add more code in order to do an update count. I was trying to re-use what we had. Happy to do it if we decide to opt for that solution.

pwalczysko:
@jburel @gusferguson : The spinner or keeping up the saving dialog would do it for me - maybe the message could be in the saving dialog ?

(Do we want to discuss on some other platform now ?)

jburel:
@pwalczysko: will try something in the dialog

gusferguson:
@jburel @pwalczysko - I think I need to see how things are happening and where we are at before I can be helpful - I am a bit behind the curve.

jburel:
Will demo it tomorrow before making any more changes.

jburel:
Discussed with @gusferguson and @pwalczysko today, we will use the activity dialog.

jburel:
@gusferguson, @pwalczysko: saving roi now shows up in activity dialog.

pwalczysko:
The wokflow works as expected. 
I would just suggest to reformulate the dialog (see screenshot below)
``Do you wish to import selected image(s) which are not saved in OMERO into OMERO ?``
screen shot 2015-02-18 at 11 34 42
I realize that even my reformulation is clanky, but I could not understand the formulation in the winow at all. 
Any better idea @gusferguson ?

pwalczysko:
Happy for the RFE https://github.com/openmicroscopy/openmicroscopy/pull/3427#issuecomment-74851162 to be dealt with in another PR.

jburel:
I will fix the wording in another coming PR: roi and MIF.
Happy to merge it @pwalczysko?

gusferguson:
I would go for:

"Do you wish to import any selected images not already saved in OMERO to the OMERO.server?"

pwalczysko:
Happy to merge. (@jburel please go for the formultaion suggested by @gusferguson , reads much better than mine.)

jburel:
Thanks
Merging

jburel:
--no-rebase

PR #3398, opened 2015-01-27
Column can be a group or user

chris-allan:
In OMERO.webadmin the drive space statistics table can either be by group or by user. This PR includes a very simple visual change to the column header from `User` to `Group/User`. As the content is loaded by AJAX, changing the header itself based on the type of data being displayed is much more involved.

To test, open up OMERO.webadmin and examine the drive space statistics page (`.../webadmin/stats/`). The header for the leftmost column should be `Group/User`:

screenshot 1

PR #3428, opened 2015-02-02
Column can be a group or user (rebased onto develop)

jburel:
This is the same as gh-3398 but rebased onto develop.

----

In OMERO.webadmin the drive space statistics table can either be by group or by user. This PR includes a very simple visual change to the column header from `User` to `Group/User`. As the content is loaded by AJAX, changing the header itself based on the type of data being displayed is much more involved.

To test, open up OMERO.webadmin and examine the drive space statistics page (`.../webadmin/stats/`). The header for the leftmost column should be `Group/User`:

screenshot 1


jburel:
--rebased-from #3398

ximenesuk:
Looks good to merge.

PR #3430, opened 2015-02-03
Add CLI subcommand to generate graph report

sbesson:
Follow up of https://github.com/openmicroscopy/openmicroscopy/pull/3404, this PR:
- adds a `bin/omero obj graphreport` command to generate the Sphinx documentation
- refactors the `ObjControl` class to implement proper subcommands
- converts the hidden `--file` argument of `ObjControl` into a `load` subcommand
- creates an intermediate `ConfigControl` class which contains the interface required for accessing the configuration in read-only mode with `WriteableConfigControl` being a subclass of it

To test this PR,
- check the tests are passing and the `bin/omero obj new/update` commands are working as expected
- check the `bin/omero obj load` and `bin/omero obj graphreport` subcommand
- check the `bin/omero admin reindex` logic is unmodified

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -311,7 +311,9 @@ Bash examples:
                           default=[sys.stdin])
 
         graphreport = parser.add(
-            sub, self.graphreport, "Generate")
+            sub, self.graphreport,
+            help="Generate a full report of the object graph in"
+            " reStructuredText format")
         graphreport.add_argument(
             "--source", help="Configuration file to be used. Default:"
             " etc/grid/config.xml")
@@ -374,8 +376,10 @@ Bash examples:
         import os
         import omero.java
         server_dir = self.ctx.dir / "lib" / "server"
+        log_config_file = self.ctx.dir / "etc" / "logback-cli.xml"
+        logback = "-Dlogback.configurationFile=%s" % log_config_file
         classpath = [file.abspath() for file in server_dir.files("*.jar")]
-        xargs = ["-cp", os.pathsep.join(classpath)]
+        xargs = [logback, "-cp", os.pathsep.join(classpath)]
 
         cfg = config.as_map()
         config.close()  # Early close. See #9800
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -372,7 +372,7 @@ Bash examples:
 
     @with_config
     def graphreport(self, args, config):
-
+        """Generate a report of the object graph in reStructuredText format"""
         import os
         import omero.java
         server_dir = self.ctx.dir / "lib" / "server"
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -379,21 +379,7 @@ Bash examples:
 
         cfg = config.as_map()
         config.close()  # Early close. See #9800
-        for x in ("name", "user", "host", "port"):
-            # NOT passing password on command-line
-            k = "omero.db.%s" % x
-            if k in cfg:
-                v = cfg[k]
-                xargs.append("-D%s=%s" % (k, v))
-
-        # Pass omero.db.pass using JAVA_OPTS environment variable
-        if "omero.db.pass" in cfg:
-            dbpassargs = "-Domero.db.pass=%s" % cfg["omero.db.pass"]
-            if "JAVA_OPTS" not in os.environ:
-                os.environ['JAVA_OPTS'] = dbpassargs
-            else:
-                os.environ['JAVA_OPTS'] = "%s %s" % (
-                    os.environ.get('JAVA_OPTS'), dbpassargs)
+        self.set_db_arguments(cfg, xargs)
 
         cmd = ["ome.services.graphs.GraphPathReport", args.file]
 
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -32,8 +32,9 @@ import sys
 import shlex
 
 from omero_ext.argparse import FileType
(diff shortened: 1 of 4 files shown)


components/tools/OmeroPy/src/omero/plugins/obj.py
```diff
@@ -297,38 +297,58 @@ def _configure(self, parser):
...
+            sub, self.update, "Update an existing OMERO object")
+
+        load = parser.add(
+            sub, self.load,
+            "Load a series of commands from a file or the stdin")
+        load.add_argument("file", nargs="*", type=FileType("r"),
+                          default=[sys.stdin])
+
+        graphreport = parser.add(
+            sub, self.graphreport, "Generate")
```
ximenesuk: "Generate" what?

components/tools/OmeroPy/src/omero/plugins/obj.py
```diff
@@ -297,38 +297,58 @@ def _configure(self, parser):
...
+        graphreport.add_argument(
+            "--source", help="Configuration file to be used. Default:"
+            " etc/grid/config.xml")
+        graphreport.add_argument(
```
ximenesuk: Does this need to be required? Might someone want to use stdout?

sbesson: Would make sense to me. As of https://github.com/openmicroscopy/openmicroscopy/pull/3404/files#diff-7799624cb91ccbbdc8b02f18f640f769R428, it seems the unrelying Java class requires a valid file input and does not support stdout as of now /cc @mtbc

mtbc: Indeed. All sorts of logging guff gets dumped out to the terminal window. That would have to be fixed first.

ximenesuk:
This still isn't working for me locally and amongst the exceptions I find this a bit worrying given that my data dir is not /OMERO

```
2015-02-06 11:05:12,705 18825      [      main] WARN   ate.search.store.DirectoryProviderHelper - Index directory not found, creating: '/OMERO'
...
Caused by: org.hibernate.search.SearchException: Unable to create index directory: /OMERO for index FullText
```

I assume the problem here is upstream rather than in this PR

sbesson:
Ahh thanks for the stack trace, I tested locally and used the default `/OMERO` so the `omero.data.dir` is not passed to the Java command but looking at this, it looks like it is required. I will move to another repo and check.

sbesson:
@ximenesuk: last commit should fix issues with non-default data directories.

components/tools/OmeroPy/test/integration/clitest/test_obj.py
```diff
@@ -118,7 +121,8 @@ def test_link_annotation(self):
             new CommentAnnotation ns=test textValue=foo
```
joshmoore: Don't these need prefixing with `obj` for a load?

joshmoore: Ah, I see. Sorry for the noise. We have a new level of `load`.

ximenesuk:
@sbesson Thanks that now seems to work. Can I just check that this effectively generates a report for the local build - ie the server itself is irrelevant?

mtbc:
The database it connects to will have to match.

ximenesuk:
Will have to match what? I guess I have realised I don't really understand what this is meant to do!

mtbc:
The DB patch of the database will have to match that of your local build.

ximenesuk:
@mtbc Thanks.

I do, in that case, wonder why this is an `obj` subcommand at all. To me it doesn't sit easily alongside the others. Before I looked at #3404 I naively assumed this told me something about my objects that I had created. I now also see that the name was discussed there so I defer to that but still feel it is misplaced.

sbesson:
@ximenesuk: I definitely hesitated between `obj` and `admin`. Thoughts about `bin/omero admin graphreport`?

ximenesuk:
It feels more like an `admin` command to me sitting alongside things like `diagnostics` makes more sense, especially as it is exposed in `obj -h`.

(Sorry to come to the conversation late on this.)

joshmoore:
Think I mentioned this elsewhere, but I can understand leaving it here as long as it also produces a format that's usable by the `bin/omero obj` users. i.e.
 * `bin/omero obj graphreport` (or just "list"?) shows me what objects I can create & how
 * `bin/omero obj graphreport --sphinx` does the current default.

sbesson:
@mtbc: is there an easy way to retrieve the first RFE from @joshmoore or would that be a separate Jave class?

mtbc:
It might ultimately be the same class but to be useful it could probably do with its own card and scoping and suchlike. For instance, rather than generating a few thousand lines of output, might one prefer a command that lists the objects that can be created, and then for any of those another command will provide some details of its properties? And would it be okay if it still glosses over important details of tricky object properties like enumerations and the new map annotations? (My motivation for the report is in querying, not constructing, and in that it's already served me well; I had limited interest in non-mapped-object value types.)

mtbc:
Before much further work it might also be best to refactor the Java class to use an expanded graph path bean (available since 5.1) rather than querying Hibernate directly.

joshmoore:
@mtbc : agreed on this needing some scoping. Shall we just suppress the output of `graphreport` in the help in order to keep this a private API for the moment? Then we can go on to such lovely things as:
```
bin/omero obj sample Project # --> "Project name=X description=Y"
```
etc.

mtbc:
Yes, that might be wise.

sbesson:
Shall I close this PR and wait for more scoping/upstream Java patches then?

mtbc:
If you need this PR's subcommand to do the report autogen for docs then I would say, let's press ahead with it and just push the RFE. My pending refactoring of the Java class won't affect its output.

sbesson:
Closing this PR. In its current state, the requirements for having the graph report auto-generated i.e. a fully set-up server that is stopped by the command, would involve too much logic addition to the autogeneration script/job. On top of that, the API is meant to be frozen for 5.1.0 soon.

PR #3431, opened 2015-02-03
Fix windows tables

manics:
Attempt to fix https://trac.openmicroscopy.org.uk/ome/ticket/12321
(i.e. Tables should now work on Windows).

Note this requires modifications to `omero_ext/portalocker.py`... what's the best way of doing this?

Requires manual testing (no Windows integration tests at the moment.... hopefully soon)

Definitely needs someone to carefully go over the change in locking (/cc @joshmoore)

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/tables.py
+++ b/components/tools/OmeroPy/src/omero/tables.py
@@ -237,8 +237,9 @@ class HdfStorage(object):
             return tables.openFile(str(self.__hdf_path), mode=mode,
                                    title="OMERO HDF Measurement Storage",
                                    rootUEP="/")
-        except (tables.HDF5ExtError, IOError):
-            msg = "HDFStorage initialized with bad path: %s" % self.__hdf_path
+        except (tables.HDF5ExtError, IOError) as e:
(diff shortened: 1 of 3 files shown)


joshmoore:
@manics : changes directly to portalocker are probably fine. If possible, try to maintain compatibility with the original. Re: testing -- probably would be best to make sure we have some concurrent test which checks the ordering of calls and tries to corrupt the file. (I don't remember if we have something like that for `hdfstorage` yet. Probably not)

manics:
I've added a couple more unit tests.

joshmoore:
Reading through all the changes again, I can't see any faults. A couple of thoughts I had were:
 * `remove` could possibly include the `close` but *shrugs*
 * It might be possible to get down to a single map containing the state of `HDFLIST`, but that would be a larger change.
 * If there's still a use for the portalocker methods, then they could redirect to the new methods. However, if they are dangerous, then better to remove.

So based on https://s3.amazonaws.com/archive.travis-ci.org/jobs/50077763/log.txt
```
test/unit/tablestest/test_hdfstorage.py::TestHdfStorage::testLocking PASSED
```
I'm merging.

PR #3420, opened 2015-01-30
Chgrp not owner

will-moore:
This fixes https://trac.openmicroscopy.org.uk/ome/ticket/12198, https://trac.openmicroscopy.org.uk/ome/ticket/10588 and
https://trac.openmicroscopy.org.uk/ome/ticket/12305 as requested on forums: https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7484

To test:
 - Login as root (or any Admin).
 - Browse to another user's data (possibly in a group that root is not a member of).
 - Choose data in tree and right-click -> Move to Group...
 - Should see list of groups that the data owner belongs to (not including current group).
 - Click on a group. Now you should see list of containers (same as existing behaviour).
 - Can choose an existing container OR click "New" to create a new container (if moving Dataset/Images/Plates)
 - Click "New..." and enter name... Click OK.
 - Data should be moved to the chosen group and placed in new container.

 Also:
 - Test that the workflow above works when trying to move split MIFs.
 - Test workflow as owner of the data.
 - Chgrp shouldn't be enabled for group owners (only data owners or Admins).

PR #3433, opened 2015-02-03
Delete - KEEP children

will-moore:
See: https://trac.openmicroscopy.org.uk/ome/ticket/12719

This fixes deletion of child objects when the parents are deleted in the jsTree.
Also includes a new Robot framework test that tests deletion of Project and Dataset (which fails without the first commit).
To test:
 - Delete a Project containing a Dataset. Check the Dataset and Images are deleted too (not orphaned).
 - Delete a Screen/Plate and check that the images are deleted (not found as orphans).

### DIFF
--- a/components/tests/ui/resources/web/tree.txt
+++ b/components/tests/ui/resources/web/tree.txt
@@ -61,7 +61,8 @@ Create Project
     [Return]    ${pid}
 
 Create Dataset
-    ${did}=     Create Container    dataset
+    [Arguments]     ${newName}=testCreateContainerRobot
+    ${did}=     Create Container    dataset    ${newName}
     [Return]    ${did}
 
 Create Button Should Be Enabled
--- a/components/tests/ui/testcases/web/post_test.txt
+++ b/components/tests/ui/testcases/web/post_test.txt
@@ -25,6 +25,30 @@ Test Delete Project
     Page Should Not Contain Element         id=project-${pid}
 
 
+Test Delete Project Dataset
+    [Documentation]     Create and Delete a Project containing a Dataset
+
+    # Clear any activities from earlier tests etc.
+    Click Element                           id=launch_activities
+    Click Element                           id=clear_activities
+    Select Experimenter
+    ${pid}=                                 Create project      robot test delete
+    ${did}=                                 Create Dataset      robot test deleteChildren
+    Click Element                           refreshButton
+    Wait Until Page Contains Element        id=project-${pid}
+    Click Element                           css=#project-${pid}>a
+    Click Element                           id=deleteButton
+    Wait Until Page Contains Element        id=delete-dialog-form
+    Click Element                           xpath=//button/span[contains(text(),'Yes')]
+    # Wait for activities to show job done, then refresh tree...
+    Wait Until Page Contains Element        xpath=//span[@id='jobstatus'][contains(text(),'1')]     20
+    Click Element                           refreshButton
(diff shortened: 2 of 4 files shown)


dominikl:
Datasets/Images are deleted if Project is deleted; so are Images if Screen/Plate is deleted. I. e. works like expected. Some robot frameworks tests have status "failed", don't know if this is related to this PR.

chris-allan:
The Plate deletion method invocation arguments look to be all that have been changed so far in `container.py`. Am I missing something obvious?

will-moore:
@chris-allan The change to deleteItem("Plate") in container.py was my first attempt to fix the bug, before I found the real bug. The critical fix is actually in containers.html where ```child=on``` was being used to specify that children should be deleted. The bug arose from 346b4a7e773b82e64725a13fc3d0a89831052d99 where the python util method toBoolean() only returned True if the value was actually 'true' (previously 'on' had returned True).
I left the container.py change as it seemed to make sense anyway, although we don't use deleteItem() for Plates now.

mtbc:
Thank you very much for finding the culprit and fixing this, @will-moore.

chris-allan:
A ha! That makes sense in reading the `containers.html` changes @will-moore. Thanks for the detailed explanation. After looking at 346b4a7e773b82e64725a13fc3d0a89831052d99 it's certainly possible that similar things have happened elsewhere and that other people are relying on the old semantics. `toBoolean()` is used in various webadmin and webclient view methods.

I signed off on and we merged #3340 without investigating more thoroughly the implications to the wider code base. @aleksandra-tarkowska, what was the reason for the change to the `toBoolean()` semantics?

aleksandra-tarkowska:
@chris-allan that was related to @jburel comment on https://github.com/aleksandra-tarkowska/openmicroscopy/commit/6f6ea0280cd8855a8ae4f7271e83090e21bff64c#comments. That is true ```on``` is missed, I will open another PR and fix it as we may have more issues somewhere else. Sorry about that.

chris-allan:
With the changes @aleksandra-tarkowska has put into #3441, do you want to revert the `on` --> `true` change @will-moore or make that change so things are more explicit? I don't have a strong opinion either way.

will-moore:
@chris-allan I think I'll leave it as 'true'. We only have 'on' because I was reproducing a checkbox in a form, but it's not as clear as 'true' in the code.

chris-allan:
Completely fair, @will-moore. As @dominikl is happy with the result this is good to merge as far as I'm concerned.

:+1:

sbesson:
Note we need to fix the results of https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-robotframework/146/console before getting this merged.

will-moore:
I added some robot framework test fixes to https://github.com/openmicroscopy/openmicroscopy/pull/3420

will-moore:
The new Robot test "Test Delete Project Dataset" is passing in FF and Chrome: https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-robotframework/147/console
Good to merge?

sbesson:
:+1: from the robot perspective (assuming remaining broken tests unreleated to this PR will be fixed elsewhere). Just to be sure, does this PR depend on #3441 or does the `on->true` change allow it to be merged now?

will-moore:
@sbesson ``` on -> true ``` is certainly sufficient to fix the bug above and for the Robot tests to pass. So #3441 is an independent fix/improvement. This can be merged now.

PR #3435, opened 2015-02-04
Update fallback IPython to 1.2.1

manics:
This is the most recent release that supports Python 2.6
Should fix https://trac.openmicroscopy.org.uk/ome/ticket/12322 (*Bug: Windows bin\omero shell: TypeError: _isdir() takes exactly 1 argument (0 given)*)

Testing: Check `omero shell` works on all OSes when IPython is ***not*** already installed.

--no-rebase


sbesson:
Tested on CentOS 6.5/Pythonn 2.6.6, Windows R2012/Python 2.7 and OS X 10.9.6/Python 2.7.9. All worked as expected. I don't have a Windows/Python 2.6 available for testing but all the systems above succesfully launched `omero shell` without `ipython` installed.

joshmoore:
I'd be inclined to merge this and think about adding `bin/omero shell` to the list of things that need automated testing (across our platforms).

manics:
I think there might be a trello card somewhere for CLI smoke-tests (along the lines of shell/batch scripts to check a server's running, create users/groups, import, hql... the sort of stuff that would make testing 20 different server configurations easier). But I can't find it using Trello search...

jburel:
https://trello.com/c/Sr8wIfvd/50-ci-testing-of-ci-node-configuration-changes is it what you are after @manics?

joshmoore:
@manics: ping? Any other objections to merging?

manics:
No objections.

PR #3436, opened 2015-02-04
File opening

jburel:
Use Process builder to open external application and file.

This should fix issue on Linux 
see https://trac.openmicroscopy.org.uk/ome/ticket/12028
and https://trac.openmicroscopy.org.uk/ome/ticket/12013

To test:
 * Open a file on linux
 * Open script using webstart.


mtbc:
For both tickets, I still can't view anything under Linux XFCE4; nothing pops up at all. The activities window mentions only the initial creation of the split view figure. I do not get the exception that Niko mentioned.

jburel:
Closing for now

PR #3421, opened 2015-01-30
5-0-7 version history

hflynn:
Based on release announcement

PR #3439, opened 2015-02-05
5-0-7 version history (rebased onto develop)

sbesson:
This is the same as gh-3421 but rebased onto develop.

----

Based on release announcement


sbesson:
--rebased-from #3421

joshmoore:
:+1:

PR #3410, opened 2015-01-28
fix #10969: rearrange DoAll's in cli.py

mtbc:
`bin/omero chgrp` should now work fine for MIFs.

See http://trac.openmicroscopy.org/ome/ticket/10969 and http://trac.openmicroscopy.org/ome/ticket/11966.

--no-rebase

PR #3440, opened 2015-02-05
Show GraphException message

dominikl:
Show the GraphException message, if a move or delete action in fails (see [Ticket 12722](https://trac.openmicroscopy.org/ome/ticket/12722) ).

Test: Provoke a failed move and/or delete action; check that the GraphException message providing more details about the cause is shown in the activity status.

/cc @mtbc 
--no-rebase

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/ui/DataTransferActivity.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/ui/DataTransferActivity.java
@@ -26,8 +26,10 @@ package org.openmicroscopy.shoola.env.ui;
 //Java imports
 import javax.swing.Icon;
 
+
+import omero.cmd.GraphException;
+
 //Third-party libraries
-import org.apache.commons.lang.StringUtils;
 
 //Application-internal dependencies
 import org.openmicroscopy.shoola.env.config.Registry;
@@ -109,9 +111,9 @@ public class DataTransferActivity
 	{
 		if (result instanceof ProcessReport) {
 			type.setText(DESCRIPTION_ERROR);
-			String graphExMsg = getGraphExceptionMessage((ProcessReport) result);
-			if(!StringUtils.isEmpty(graphExMsg)) 
-				messageLabel.setText(messageLabel.getText()+" [Details: "+graphExMsg+"]");
+			GraphException ex = ((ProcessReport) result).getGraphException();
+			if (ex != null) 
+				messageLabel.setText(messageLabel.getText()+" [Details: "+ex.message+"]");
 			notifyActivityError();
 		} else {
 			type.setText(DESCRIPTION_END);
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/ui/DataTransferActivity.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/ui/DataTransferActivity.java
@@ -113,7 +113,7 @@ public class DataTransferActivity
 			type.setText(DESCRIPTION_ERROR);
 			GraphException ex = ((ProcessReport) result).getGraphException();
 			if (ex != null) 
-				messageLabel.setText(messageLabel.getText()+" [Details: "+ex.message+"]");
+				messageLabel.setText(messageLabel.getText()+" - "+ex.message);
 			notifyActivityError();
 		} else {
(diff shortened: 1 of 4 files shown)


mtbc:
That was quick! #3410 remains unmerged -- might want to hold off until that goes in so you don't have to unpick it from the stacktrace.

dominikl:
ON_HOLD : Waiting for #3410 being merged.

dominikl:
Rebased after #3410 has been merged. Ready for review now.

components/insight/SRC/org/openmicroscopy/shoola/env/ui/DataTransferActivity.java
```diff
@@ -109,6 +111,9 @@ protected void notifyActivityEnd()
 	{
 		if (result instanceof ProcessReport) {
 			type.setText(DESCRIPTION_ERROR);
+			GraphException ex = ((ProcessReport) result).getGraphException();
+			if (ex != null) 
+				messageLabel.setText(messageLabel.getText()+" [Details: "+ex.message+"]");
```
mtbc: I should mention that the message may well also end with a `]`

dominikl: Is it better to use curved brackets? Or no brackets at all?

mtbc: maybe none at all is best, you never know what other messages might be added server-side

dominikl:
What's the problem with Travis again?

joshmoore:
Travis looks happy enough.

gusferguson:
Tested using OMERO.insight-5.1.0-m3-1131-2b7bc2c-ice35-b409-mac_Java7+ trout user-3 read-annotate-1

Encountered 2 bugs.

https://trac.openmicroscopy.org.uk/ome/ticket/12733
https://trac.openmicroscopy.org.uk/ome/ticket/12734

Following the testing suggested by Mark in the ticket:
- moving a projected image without it’s parent gives rise to the warning shown in screenshot 2.
- “move 1 object - may not split images that share instrument[6251]" is still a pretty user-unfriendly and opaque warning
- Suggest something like: "The move could not be completed as the image is part of a group."
- If you need a more generic message then keep it really generic "The operation could not be completed."
- if an error code is to be quoted (usually not really helpful from user POV) it should be indicated what it is: " [Error code: 6251]”.
- another issue with this is that after the failed move, the projected image (the one I tried to move) disappears from the dataset even though it has not been moved. This is rather alarming as it makes you think the data has been lost in cyberspace between groups! You have to refresh the data manager to see that it is still there.
- see screenshot 2.

I could not attempt to delete an annotation belonging to another user on their own image or my image as there is no way to do this in Insight.

main

main

mtbc:
The 6251 is actually the id of the specific Instrument instance that's causing the problem.

It's good if at least the clueful users get access to these messages from the server because they really help in pinpointing the actual problem; unfortunately there isn't an easy message to give that will be usefully actionable by a naïve user, who probably don't even know that Instrument objects with ID numbers even exist, let alone how to find the images that share that instrument. If the messages oughtn't be displayed by default in the dialog, hopefully they still could still make it into the Insight log file or be submittable via QA or something, as a real problem with failed moves/deletes in Insight is how the client says nothing explanatory about the failure (web is much better in allowing the error to be seen).

The messages probably aren't going to improve for 5.1.0; it could be possible to create a much better UI experience for move/delete failures where, for instance in the projection case, Insight could recognize a specific kind of exception from the server and could then query for which other images share the instrument and could then helpfully guide the user to them or offer to include them in the move, but that would be a non-trivial RFE.

dominikl:
How shall we proceed? I'd suggest to merge, because this PR is just about forwarding the exception messages to the user. I think improving the messages themselves should go into a separate PR. Same for the disappearing of the items in the browser although they've not been successfully deleted.

mtbc:
I'd concur: merge and file RFEs.

mtbc:
Probably could remove "on hold" label unless further discussion is required.

joshmoore:
@gusferguson : any objections to merging as suggested?

gusferguson:
@joshmoore - Happy to merge.

PR #3445, opened 2015-02-05
Tagging ome tiff

jburel:
Fix problems when tagging ome-tiff at import see https://trac.openmicroscopy.org.uk/ome/ticket/12164
Images were not tagged
To test:
 * Select an ome-tiff
 * Select or create a new tag.
 * Import the image.
 * Check that the image is tagged.


pwalczysko:
Works as expected.

joshmoore:
@jburel : a quick question -- why is this OME-TIFF specific?

jburel:
it was not ome-tiff specific, just it happens to be highlighted with that specific format due to the check.

joshmoore:
Gotcha. Thanks.

PR #3447, opened 2015-02-05
Customizable batch image export filename

dominikl:
See [Ticket 11895](https://trac.openmicroscopy.org/ome/ticket/11895)

Test: Select some images and export them as JPG, PNG or TIFF. You can now choose a name for the generated batch export zip file. Default name is "Batch_Image_Export.zip". Make sure that if a file with that name already exists, the default name will be incremented to "Batch_Image_Export(1).zip" and so on. Also check if the file gets overwritten if you select a file which already exists (and that you have to confirm this).


gusferguson:
@dominikl 

Tested with OMERO.insight-5.1.0-m3-1037-eba9904-ice34-b402-mac_Java7+ trout user-3 read-annotate-1

Export as PNG/JPEG etc from thumbnail or plate view works as expected.
Dialog allows renaming of file or increments Batch_Image_Export.zip with number in brackets - Batch_Image_Export(1).zip etc.
Good to merge.

PR #3449, opened 2015-02-05
CLI nginx template improvements

sbesson:
See https://trello.com/c/llYcBf5H/408-omero-web-nginx-client-max-body-size and https://trello.com/c/lm5ZOYF0/412-omero-web-nginx-default-port-change-to-80-instead-of-8080.

The CLI web plugin and nginx templates has been modified to:
- reduce the complexity of the `config()` method
- group all nginx arguments into an argument group
- use `8080` as the default for the development nginx and `80` as the default for the system nginx
- set `max_client_body_size` and allow it to be control via CLI (defaulting to 0, i.e. no size check)

Unit tests have been updated to reflect these changes, test all combinations but also test the backwards compatible `bin/omero web config nginx --system`.
The improvements in this PR can be manually checked by running the following commands.

```
bin/omero web config -h  # should group all nginx arguments

bin/omero web config nginx   # should give listen 80;
bin/omero web config nginx --http 8081  # should give listen 8081;
bin/omero web config nginx-development   # should give listen 8080;
bin/omero web config nginx-development --http 8081  # should give listen 8081;

bin/omero web config nginx-development   # should give client_max_body_size 0;
bin/omero web config nginx-development --max-body-size 0   # should give client_max_body_size 0;
bin/omero web config nginx-development --max-body-size 1m   # should give client_max_body_size 1m;

# same commands as the last block with nginx instead of nginx-development
```

/cc @manics
--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/web.py
+++ b/components/tools/OmeroPy/src/omero/plugins/web.py
@@ -234,7 +234,8 @@ class WebControl(BaseControl):
             "ROOT": self.ctx.dir,
             "OMEROWEBROOT": self._get_python_dir() / "omeroweb",
             "STATIC_URL": settings.STATIC_URL.rstrip("/"),
-            "NOW": str(datetime.now())}
+            "NOW": str(datetime.now()),
+            "HTTPORT": port}
 
         try:
             d["FORCE_SCRIPT_NAME"] = settings.FORCE_SCRIPT_NAME.rstrip("/")
@@ -242,7 +243,6 @@ class WebControl(BaseControl):
             d["FORCE_SCRIPT_NAME"] = "/"
 
         if server in ("nginx", "nginx-development"):
-            d["HTTPPORT"] = port
             d = self._set_nginx_fastcgi(d, settings)
 
         if server == "apache":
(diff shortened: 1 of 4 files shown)


manics:
Remove the old `client_max_body_size 0;` from nginx-development.conf.template https://github.com/openmicroscopy/openmicroscopy/pull/3449/files#diff-56dc20a5722077514a2173680ca302fdR25

Otherwise looks good, generates configs are identical to before apart from the expected additions.

sbesson:
@manics: good point. One related question, in the `nginx-development` template we have

```
    sendfile on;
    send_timeout 60s;
    client_max_body_size 0;
```

in the `http` block. Both additional configurations work either in `http` or `server` context  (http://nginx.org/en/docs/http/ngx_http_core_module.html#send_timeout and http://nginx.org/en/docs/http/ngx_http_core_module.html#sendfile). Should we impose these defaults in the `server` block of both templates as well?
/cc @aleksandra-tarkowska

manics:
http://nginx.org/en/docs/http/ngx_http_core_module.html#send_timeout
`Default: 	send_timeout 60s;`
Could we remove this from the development template?

manics:
Looks good to me.

PR #3451, opened 2015-02-06
Fixed a ClassCastException

dominikl:
Same as b8319c9fdb ; Fixes the ClassCastException in Projection view, [Ticket 12724](https://trac.openmicroscopy.org.uk/ome/ticket/12724)


jburel:
Crash is no longer happening.  Also tested with various figure dialogs, no error noticed.
@pwalczysko is going to try it now too

pwalczysko:
Tested Projection and tabs in acquisition pane, float and dv images with multi-z. 
Both Mac and Win 7. 
Also plates with multi-z.
All projections working as expected. 
Split view also fine.
All seems good here.

PR #3452, opened 2015-02-06
Fixed potential NPE

dominikl:
See [Ticket 12547](https://trac.openmicroscopy.org.uk/ome/ticket/12547)

I don't know how to test this; but a code review should be sufficient (it's just an additional check for `null`, and `password` is the only Object which could trigger a `NullPointerException` in this `toString()` method)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/login/UserCredentials.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/login/UserCredentials.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.env.data.login.UserCredentials
  *
  *------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mtbc:
Code change looks sensible. :+1: Don't know if you want to log anything if it *is* `null`?

jburel:
I won't do anything. I need to find work I did for connection using only session id. 
we might have a ``` null``` password in that case.

joshmoore:
Guess the question is what it means for the password to be null. If no one knows, logging at warn might make sense.

jburel:
There is no ref to logger at that level, a ref will have to be specified  or it will have to be registered when the class is initialized.

joshmoore:
Fair enough. Unless there's a precedent of adding "# TODO on refactoring", I'd say let's merge and move on.

mtbc:
Fine with me.

PR #3453, opened 2015-02-06
Bugfixes: Change user's default group

dominikl:
Fixes [Ticket 12378](http://trac.openmicroscopy.org.uk/ome/ticket/12378)
Also fixed: Save button did not become enabled if just the default group was changed.

Test: As an admin user try to change the default group of another user to a group the admin user is not a member of. This previously triggered the `DSOutOfServiceException: Can't modify the current group.` exception.


pwalczysko:
Works as expected.

PR #3455, opened 2015-02-06
fix #12179: adjust Insight JNLP specification for webstart

mtbc:
Fixes http://trac.openmicroscopy.org.uk/ome/ticket/12179.

* Bumps Java requirement up to 1.6.
* Removes preference for creating a desktop icon.

--no-rebase


dominikl:
Works fine with IcedTea 1.7 (on Debian), still a warning that the signature could not be verified (but that's expected with trout builds, isn't it?), no warning about the desktop launcher anymore; so good to merge.

mtbc:
I've already been enjoying having this PR in the merge build, one less dialog to click each time in testing. :smiley:

PR #3462, opened 2015-02-07
Impose pep8 1.5.7 for Travis validation

sbesson:
A new release of PEP8 (1.6.0) is out which invalids our Python code base. This commit should ensure Travis does not fail systematically

PR #3463, opened 2015-02-07
Impose pep8 1.5.7 for Travis validation (rebased onto dev_5_0)

sbesson:
This is the same as gh-3462 but rebased onto dev_5_0.

----

A new release of PEP8 (1.6.0) is out which invalids our Python code base. This commit should ensure Travis does not fail systematically


sbesson:
--rebased-from #3462

sbesson:
Merging to prevent Travis to be (completely) broken on this line.

PR #3467, opened 2015-02-09
Fixed typo(?) in HistogramPlot

dominikl:
While investigating [Ticket 12657](https://trac.openmicroscopy.org.uk/ome/ticket/12657) I noticed an if statement, which doesn't make sense, maybe a typo. I think the purpose of the statement is, to make sure that all three lists are not null and have the same size, therefore the 'and' should be an 'or'.

Test: Just review the changed if statement; apart from that, make sure the histogram plot (ROI/Measurement tool) still works.
Note: This isn't a fix for the ticket, but I'll close the ticket, because I can't reproduce it.

--no-rebase


joshmoore:
`Graph Pane > Histogram` is working fine. Certainly the logic (`||` rather than `&&`) makes much more sense. :thumbsup:

PR #3469, opened 2015-02-09
Show in place imported images

will-moore:
This replicates Insight's indication of images that are imported with "in place" import.

To test:
 - import an image with transfer, E.g: ``` $ bin/omero import -- --transfer=ln_s my_file.dv ```
 - Check that this image shows the link icon to the right of Image ID in right panel, with tooltip.
 - Check other images or P/D etc don't show this.

screen shot 2015-02-09 at 16 34 46

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/fs/__init__.py
+++ b/components/tools/OmeroPy/src/omero/fs/__init__.py
@@ -2,3 +2,17 @@
 # -*- coding: utf-8 -*-
 import IceImport
 IceImport.load("omero_FS_ice")
+
+#
+# Copied from:
+# blitz/src/ome/formats/importer/transfers/AbstractFileTransfer.java
+#
+TRANSFERS = {
+    "ome.formats.importer.transfers.CopyFileTransfer": "cp",
+    "ome.formats.importer.transfers.CopyMoveFileTransfer": "cp_rm",
+    "ome.formats.importer.transfers.HardlinkFileTransfer": "ln",
+    "ome.formats.importer.transfers.MoveFileTransfer": "ln_rm",
+    "ome.formats.importer.transfers.SymlinkFileTransfer": "ln_s",
+    "ome.formats.importer.transfers.UploadRmFileTransfer": "upload_rm",
+    "ome.formats.importer.transfers.UploadFileTransfer": "",
+    }
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2819,15 +2819,25 @@ class _BlitzGateway (object):
 
         params = omero.sys.ParametersI()
         params.addIds(imageIds)
+
         query = "select distinct(fse) from FilesetEntry as fse "\
-                "left outer join fse.fileset as fs "\
+                "left outer join fetch fse.fileset as fs "\
+                "left outer join fetch fs.annotationLinks as link "\
+                "left outer join fetch link.child as a "\
                 "left outer join fetch fse.originalFile as f "\
                 "left outer join fs.images as image where image.id in (:ids)"
         queryService = self.getQueryService()
         fsinfo = queryService.findAllByQuery(query, params, self.SERVICE_OPTS)
         fsCount = len(fsinfo)
+        annNs = []
+        for fse in fsinfo:
+            for l in fse.fileset.copyAnnotationLinks():
+                annNs.append(l.child.ns.val)
         fsSize = sum([f.originalFile.getSize().val for f in fsinfo])
-        filesetFileInfo = {'count': fsCount, 'size': fsSize}
+        filesetFileInfo = {'fileset': True,
+                           'count': fsCount,
+                           'size': fsSize,
+                           'annNs': annNs}
         return filesetFileInfo
 
     def getArchivedFilesInfo(self, imageIds):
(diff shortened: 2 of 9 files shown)


joshmoore:
Note: this might be the time to classify which transfer types are dangerous or not:

 * `ln`, `ln_rm`, `cp_rm`, `cp` should all be fine.
 * `ln_s` : likely the only well-known transfer which would leave data outside.
 * anything user defined should be considered at risk.

mtbc:
```
./components/tools/OmeroPy/src/omero/gateway/__init__.py:8696:21: E126 continuation line over-indented for hanging indent
```

will-moore:
@joshmoore You want some indication in the UI about what type transfer type has been used? Is that stored anywhere?

joshmoore:
@will-moore : the annotation that you are already checking will contain the long-string form of the transfer (see https://github.com/openmicroscopy/openmicroscopy/blob/v.5.0.8/components/tools/OmeroPy/src/omero/plugins/fs.py#L50). For ln, ln_rm, cp_rm, cp, I would think we don't need as strong of a warning (if any at all). /cc @ximenesuk @mtbc

will-moore:
Maybe we should simply state the type of transfer that was used, and leave the warnings to the docs. Trying to convey the complexity of this issue in a tooltip is bound to leave most people confused and no-one sufficiently warned.

ximenesuk:
Can a tooltip link to the docs?

jburel:
FYI: @gusferguson is currently reviewing tooltip see https://trello.com/c/HnLbU7Xp/319-scope-standardisation-of-tooltips-across-and-within-clients

will-moore:
After discussing this with @pwalczysko there are several issues" cc @gusferguson @dominikl @jburel.

 - Insight has 2 places where you can show path in managed repo: inPlace icon tooltip and ```../``` button.
 - Having the inPlace 'link' icon to the right of Image ID is probably too prominent since this is not a very important piece of info and will likely confuse regular users who have no knowledge of where the data is on the server.
 - The ```../``` button also has the P/D hierarchy which is hidden for anyone trying to find where this went, and we are also mixing client P/D/I "paths" with server-side managed repo paths and this is kinda confusing.

Suggested solutions:
 - Probably need to move the P/D hierarchy back to where it was, since it is critical functionality for Insight search results and if users can't find it then search is kinda broken. Alternative is to at least put a note in it's place saying "Hierarchy is now available from the top toolbar (under a 'Dataset' icon)" and put this in the toolbar itself. Web should follow whatever we decide for Insight. 
 - Maybe move the "in-place" info into the "paths" pop-up, since this info is very closely related. This will keep it out of the way of users who have no interest in server-side details. Here we can then display more info about the type of transfer etc.
 - I'm kinda going back on my "Link to the docs" idea, since the person who did the import should know about the docs anyway. And if we are not careful we will have every feature of the clients with a "link to the docs" beside it.

Let's discuss at the next opportunity.

jburel:
Let's review tomorrow, everybody should be in the office.

will-moore:
Following discussion today, I have removed the link icon from beside image-ID and instead show the command in the file-paths info panel (see screenshot).
Still TODO: need to collapse the "Imported from" and "Path(s) on server" sections if more than 2 paths.

screen shot 2015-02-12 at 23 14 30

gusferguson:
@will-moore - looks good, but can the horizontal dividing line go above the titles - with the From made into a title -

1 image file:
Horizontal line
Imported with: blah
Horizontal line
From:
blah/blah
Horizontal line
Path on server:
blah/blah/blah

Thanks!

will-moore:
@gusferguson In the case like above where it's not imported with --transfer=... (just a regular import) how do you want it to look then?
How about:
```
1 image file
------------------------
Imported From:
blah/blah
------------------------
Path on server:
blah/blah/blah
```

Then when imported with transfer, we have
```
1 image file
------------------------
Imported with --transfer=ln
------------------------
Imported From:
blah/blah
------------------------
Path on server:
blah/blah/blah
```

will-moore:
Fixed the layout of the paths panel as discussed.
If there's more than 2 paths, I just do a simple "show more" instead of a whole expandable panel. This gives the functionality we need while keeping it simple (re: @dominikl's comments on Insight).

screen shot 2015-02-16 at 15 26 21

screen shot 2015-02-16 at 15 27 19

gusferguson:
@will-moore - sorry I missed your query earlier.
Yes your layout for 
Imported with --transfer=ln
looks good.

dominikl:
Instead of the "Show more" I have a little +/- icon in Insight, but I think these little differences are ok, aren't they?

gusferguson:
Tested using https://trout.openmicroscopy.org/merge/webclient/ user-3 read-annotate-1 - own data and other’s data

Looking good.
Screenshots for reference cf Insight.
Good to merge.

omero insight

omero insight

omero insight

omero insight

joshmoore:
Location of `TRANSFERS` looks good to me. Only question I have for future-proofing is to what extent we consider this returned dictionary to be part of the public API. If so, it might make sense to plan for some of the changes in 5.2. /cc @mtbc

mtbc:
Would definitely be good to have a clear documented sense of what is public API. This transfer-method question probably needs a Trello card at least for some thinking-through for 5.2, if it should even be part of the OMERO model somehow.

PR #3470, opened 2015-02-10
Mac build

jburel:
Add the config folder under Contents/Resources

* Make sure the insight and importer start w/o a config directory next to the app

A zip is still generated, we may want to remove that too.

--no-rebase


sbesson:
Tested under OS X 10.9.5 the four available artifacts:

```
OMERO.importer-5.1.0-m3-1131-2b7bc2c-ice35-b409-mac_Java6.zip
OMERO.importer-5.1.0-m3-1131-2b7bc2c-ice35-b409-mac_Java7+.zip
OMERO.insight-5.1.0-m3-1131-2b7bc2c-ice35-b409-mac_Java6.zip
OMERO.insight-5.1.0-m3-1131-2b7bc2c-ice35-b409-mac_Java7+.zip
```

All zips contain an empty INSTALL file and an application. All applications contain a `config` folder under `Content/Reources`. All application launch fine and connect to a matching server. Tested values from the container: feedback is properly submitted and URL modification is correctly propagated. Good to merge.

With regard to ZIP, I am not in favor of dropping it for 2 reasons:
- even if small, there is some compression (70MB vs 80MB)
- the application does not have any version/build number/platform in its name. Is that something we want to modify?

jburel:
The INSTALL is not empty but the text could do with some updating. 
I am not sure of adding the version number to name.

jburel:
@sbesson: updated file.

jburel:
@sbesson: travis is green

PR #3471, opened 2015-02-10
Remove ref to fs.config.

jburel:
Remove reference to the fs.config file
We no longer use it.

--no-rebase


components/insight/SRC/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
```diff
@@ -45,11 +41,6 @@
 
 import javax.swing.JDialog;
 import javax.swing.JFrame;
-
-//Third-party libraries
-
-
-//Application-internal dependencies
```
mtbc: don't know if you wanted to remove this but not the Java imports comment

joshmoore:
Both the Mac_Java7+ and the Linux zips opened fine. Reviewing changes now...

joshmoore:
Do any of these files deserve removing?
```
$ git ls-files | grep insight | grep -i fs | wc -l
     100
```

jburel:
Ironically when some of the work started, the agent was named "fsimporter"
I have kept the filesystem view and related files if we decide to offer view of the repo itself (discussed a while ago!)

jburel:
not planning further changes on this PR.

PR #3473, opened 2015-02-10
Select next row after deleting map annotation rows

will-moore:
Fix issues from https://trello.com/c/Fx74klF5/17-map-annotation-schema-db

Only 1 change: Deletion of a row should select the next row (or previous row if the bottom row is deleted).

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
@@ -963,15 +963,52 @@
         
     
     
-    
-    
-    
-    
-    /* Modal Dialog */
-    
-   
-    
-           
+    /* Map Annotations */
+
+    .mapAnnContainer {
+        position: relative;
+        max-height: 250px;
+        overflow-y: auto;
+    }
+    .keyValueTable {
+        position: relative;
+        margin-bottom: 20px;
+        width: 98% !important;
+    }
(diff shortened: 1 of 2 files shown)


gusferguson:
@will-moore 

Deletion of row behaves as expected - next row selected.
From this point of view good to merge.

A couple of other issues I noticed:
- the X icon for deleting is out of alignment with the other icons in the toolbar - it looks as if it is top aligned and the others middle.
- can the toolbar be shifted a few pixels to the left so it doesn't overlap the scroll bar?
- it is possible to create empty key-value pairs - I thought this was not allowed?


safari

will-moore:
@gusferguson Fixed the alignment of the X button and shifted the toolbar left a bit.
The + button will create a new row, which will be empty, but will not be saved unless some key or value is added.
I prefer to allow the user to create a bunch of empty rows if they want (E.g. may want to create 4 rows that they know they need, then fill them in).
I don't think we gain much by disabling the + button under certain conditions. E.g. what conditions would you disable it and how would you tell the user what they have to do to enable it again?
Spreadsheet apps allow you to create a ton of empty rows before filling them in.

gusferguson:
@will-moore 
Happy with the empty rows if they are not being saved.

gusferguson:
@will-moore 

Looking good.
Behaves as expected.
Good to merge.

safari

jburel:
Thanks Merging

jburel:
--no-rebase

PR #3476, opened 2015-02-10
Memory settings error handling

sbesson:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12527

This PR should be tested via 2 scenarios:

1. from a merge server, download the `templates.xml` from a 5.0.2 or earlier server and put it under `etc/grid`, then either start the server or run `bin/omero admin jvmcfg`. 

2. set an improper strategy, e.g. `bin/omero config set omero.jvmcfg.strategy invalid_strategy` and run the same command as above (`bin/omero admin jvmcfg`). 

In both cases the CLI command should fail with an informative error message

Also check the new unit tests are green.

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/install/jvmcfg.py
+++ b/components/tools/OmeroPy/src/omero/install/jvmcfg.py
@@ -398,6 +398,10 @@ def adjust_settings(config, template_xml,
             ("pixeldata", pixeldata), ("repository", repository))
 
     for name, StrategyType in loop:
+        if name not in options:
+            raise Exception("Cannot find %s option" % name)
+
+    for name, StrategyType in loop:
         specific = strip_dict(m, suffix=name)
         defaults = strip_dict(m)
         settings = Settings(specific, defaults)
--- a/components/tools/OmeroPy/src/omero/install/jvmcfg.py
+++ b/components/tools/OmeroPy/src/omero/install/jvmcfg.py
@@ -399,7 +399,9 @@ def adjust_settings(config, template_xml,
 
     for name, StrategyType in loop:
         if name not in options:
-            raise Exception("Cannot find %s option" % name)
+            raise Exception(
+                "Cannot find %s option. Make sure templates.xml was "
+                "not copied from a 5.0.2 or older server" % name)
 
     for name, StrategyType in loop:
         specific = strip_dict(m, suffix=name)
--- a/components/tools/OmeroPy/src/omero/install/jvmcfg.py
+++ b/components/tools/OmeroPy/src/omero/install/jvmcfg.py
@@ -401,7 +401,7 @@ def adjust_settings(config, template_xml,
         if name not in options:
             raise Exception(
                 "Cannot find %s option. Make sure templates.xml was "
-                "not copied from a 5.0.2 or older server" % name)
+                "not copied from an older server" % name)
 
     for name, StrategyType in loop:
         specific = strip_dict(m, suffix=name)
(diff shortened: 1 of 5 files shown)


components/tools/OmeroPy/src/omero/install/jvmcfg.py
```diff
@@ -398,6 +398,12 @@ def adjust_settings(config, template_xml,
             ("pixeldata", pixeldata), ("repository", repository))
 
     for name, StrategyType in loop:
+        if name not in options:
+            raise Exception(
+                "Cannot find %s option. Make sure templates.xml was "
+                "not copied from a 5.0.2 or older server" % name)
```
manics: Is a specific version necessary? If you just say it's out of date it won't matter if the templates.xml is changed again in future.

components/tools/OmeroPy/test/unit/clitest/test_admin.py
```diff
@@ -332,3 +332,50 @@ def testExplicitPorts(self, registry, ssl, tcp, webserver, prefix):
         self.cli.invoke(self.args, strict=True)
 
         self.check_ice_config(**kwargs)
+
+
+class TestAdminJvmCfg(object):
+
+    @pytest.fixture(autouse=True)
+    def setup_method(self, monkeypatch):
```
ximenesuk: Ah, I see what you mean. This seems to be mixing the two idioms of `setup_method` and `fixture`. If `setup_method` as it is doesn't work for this use case then maybe a different name for the fixture method would avoid any confusion?

manics:
Otherwise looks good.

components/tools/OmeroPy/test/unit/clitest/test_admin.py
sbesson: Fair enough, I will look into creating a separate fixture

components/tools/OmeroPy/src/omero/install/jvmcfg.py
sbesson: Thinking about this again. I agree, if we make further changes to `templates.xml` in the future, we will need to make sure this is in-sync with the current server anyways.

Actually a related thought: should we `chmod 444 templates.xml` since this file should only be used in read-only mode? /cc @joshmoore

mtbc: If so, maybe mention in upgrade notes a link to the proper way to change the settings.

joshmoore: There are certainly options which are currently *only* configurable in templates.xml. I'm hesitant to say that we should prevent it from being edited. This makes upgrades very difficult/dangerous. The only solutions I've thought of so far are:
 * generate everything under `/etc` (in which case nothing is hand-editable)
 * put `etc/` into git and manage it like `lib/scripts`

manics: Is there a version field (like in config.xml)?

mtbc:
(The new unit tests are checked by Travis, right?)

mtbc:
All works pretty well, e.g.,
```
Cannot adjust memory settings in /home/mtbc/src/openmicroscopy/dist/etc/grid/templates.xml.
Bad strategy: invalid_strategy
```
and
```
Cannot adjust memory settings in /home/mtbc/src/openmicroscopy/dist/etc/grid/templates.xml.
Cannot find blitz option. Make sure templates.xml was not copied from an older server
```
It's unfortunate that the message lines run over 80 chars, but this PR is already **good to merge**.

PR #3481, opened 2015-02-12
Don't set info popup modal

dominikl:
Tiny PR which fixes [Ticket 12734](http://trac.openmicroscopy.org.uk/ome/ticket/12734)

Problem: Despite being modal, it could sometimes loose focus, move into the background and block the whole UI. I think there's no need for the info popup to be modal. Now it's just closed if it looses focus.
Best being tested by @gusferguson , because the bug seems to need quite a specific environment to appear.

--no-rebase


gusferguson:
@dominikl 

Works as expected.
Good to merge.

PR #3482, opened 2015-02-13
Reduce pyinotify read frequency and coalesce events

ximenesuk:
This is another test PR to look at the effect of `read_freq` on cpu usage. It doesn't need review.

@sbesson can you mark as onhold please?


ximenesuk:
I'm closing this for now to look at the performance without this PR.

ximenesuk:
I'm not convinced this provides a solution to the problem Niko is seeing. It seems to cause a permanent high load on the process.

PR #3484, opened 2015-02-13
General pane and File paths

dominikl:
This PR implements the changes to the General Pane in Insight, discussed yesterday with @will-moore @jburel @gusferguson  and @pwalczysko :
- Inplace import icon removed, instead the inplace import information is shown in the "File path" info popup.
- The information shown in the "File path" popup matches the information shown in Web, see #3469 
- "File Path" and "Located in" buttons aren't hidden in a sub menu anymore, but are both displayed in the toolbar. 
- The order of the icons in the toolbar matches the order in Web.
- The "Located in" panel at the bottom of the General Pane is back again.


components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
```diff
@@ -452,25 +416,31 @@ public void mouseReleased(MouseEvent e) {
...
+        MouseListener pathLocML = new MouseAdapter() {
+            @Override
+            public void mousePressed(MouseEvent e) {
+                location = e.getPoint();
+                component = (Component) e.getSource();
+            }
+        };
+		
+		pathButton = new JButton(icons.getIcon(IconManager.FILE_PATH));
+        pathButton.setToolTipText("Show file paths on server.");
```
mtbc: Why does this get a period but not the location button?

gusferguson:
@dominikl 

Tested with OMERO.insight-5.1.0-m3-1196-48fc421-ice34-b413-win trout user-3 read-annotate-1

Behaves as expected.
Layout of dialog is what is expected and Web should match it soon.
I need to do another icon for the file path as this one is looking very poor.
Will let you know when I have one.
Good to merge.

safari

safari

jburel:
@gusferguson: did you try with MIF?

gusferguson:
@jburel - No will do now.

gusferguson:
@dominikl @jburel - seems to be OK with MIF.

parallels desktop

dominikl:
Just added missing full stops to the tooltip texts (see @mtbc comment above), no new review needed for that.

mtbc:
Last commit looks fine to me.

jburel:
From last discussion, I understood that we will show all if no more than 2 otherwise the Path on server section will not be expanded by default cc @will-moore

dominikl:
So the "Path on server" section should be expandable/collapsable, within a TaskPane? I think having such a 'heavy weight' component in this rather light weight dialog popup would look a bit strange.

jburel:
Maybe just a +/> clickable label to expand not the full expandable component.
If it is too much work, then let's stick to what we have and we can revisit later

gusferguson:
@dominikl @jburel 

Tested using OMERO.insight-5.1.0-m3-1210-25c0de8-ice35-b414-mac_Java6 trout user-3 read-annotate-1 - own data and other’s data

In Insight
- when first opened dialog was centred in Data Manager pane. Subsequently moved over to left - couldn’t replicate or work out why - see screenshot 1
- I would prefer dialog to be aligned so that dialog and Data Manager right edges are aligned, rather than the fairly random placement at present - this would be more in keeping with Web.
- I think text needs to wrap in Insight as it does in Web - see screenshot 1
- in Insight button is available for screening data and dialog says “No information available”. For normal data if there is no information available then the button is greyed out. Consistency would be nice, but if it is difficult to grey out the button in Insight then it is not a deal-breaker.  Web button is not available in both cases when there is no data. Screenshot 2.
- Insight - truncation for >2 is fine, but I don’t think the “+” for show more in Insight works - it is not intuitive at all. I think it would be better to have a link or button below the truncated text as in Web saying “Show more …" - screenshots 3 and 4
- neither Insight or Web have ability to collapse again after “Show More” but I don’t think this is a problem as you can just use the button again and regenerate the dialog.

omero insight

omero insight

omero insight

omero insight

dominikl:
- “No information available”: I'm not really sure, if this is only the case for <= 5.0 imported data or 'archived' files... Anyway, I thought the usual case probably is, that the information is available. And assuming that the server path information dialog is a rarely used feature anyway, we could save another call to the server triggered by the metadata panel by just showing "No information available" if it has been clicked on instead of requesting the FileSet information in every case just for checking if the button should be enabled or not. 
- Collapse path list again: Should be possible in Insight. The plus icon turns into a minus icon after expansion, which collapses the list again, if clicked on. That's also why I went for the plus/minus icons, because "Show more", and then what "Show less"? 
- Have to check the opening location of the dialog again. Actually it should open next to 'Paths' button.

dominikl:
How shall I proceed here? Remove +/- and replace with "Show more" (and "Show less"?)? @jburel What do think about the "No information available" vs button disabled (which needs a call to the server every time the reference objects changes)?

jburel:
I will use the "Show xx" to be more explicit
and avoid another call since we have enough call already for images. So better to use "No info"

dominikl:
- Replaced +/- with show more/less
- Couldn't reproduce the location problem, the dialog opens with its top-right corner close to the "Show paths" button. Is it OS specific?
- Unfortunately getting the line wrap to work, would need a really horrible workaround. Could we omit that?

gusferguson:
@dominikl 

The location problem was a one off - write it down to temporary glitch.

Happy to leave wrapping - definitely don't want to embark on horrible workarounds at this stage!

gusferguson:
@dominikl 

Tested using OMERO.insight-5.1.0-m4-139-69fede3-ice34-b419-win user-3 read-annotate-1

Behaves as expected - "Show more ..." "Show less ..." works much better.
No location issues evident.
Good to merge.

jburel:
thanks

PR #3490, opened 2015-02-16
fixing deprecated warning, ticket #12472

aleksandra-tarkowska:
this should resolve deprecation warning. To test just see if searching works

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/search.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/search.py
@@ -97,6 +97,9 @@ class BaseSearch(BaseController):
                     self.containers[dt] = doSearch(dt)
                     resultCount += len(self.containers[dt])
         except Exception, x:
-            if hasattr(x, "message") and "TooManyClauses" in x.message:
-                self.searchError = "Please try to narrow down your query. The wildcard matched too many terms."
+            if hasattr(x, "message"):
+                if "TooManyClauses" in x.message:
+                    self.searchError = "Please try to narrow down your query. The wildcard matched too many terms."
+                else:
+                    self.searchError = "Your query for '%s' caused an error: %s." % (query, x.message)
         self.c_size = resultCount
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/search.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/search.py
@@ -25,6 +25,7 @@
 
 import datetime
 import time
+import omero
 
 from omero.rtypes import rtime
 from webclient.controller import BaseController
@@ -97,9 +98,11 @@ class BaseSearch(BaseController):
                     self.containers[dt] = doSearch(dt)
                     resultCount += len(self.containers[dt])
         except Exception, x:
-            if hasattr(x, "message"):
+            if isinstance(x, omero.ServerError):
                 if "TooManyClauses" in x.message:
                     self.searchError = "Please try to narrow down your query. The wildcard matched too many terms."
                 else:
                     self.searchError = "Your query for '%s' caused an error: %s." % (query, x.message)
+            else:
+                self.searchError = "Your query for '%s' caused an error: %s." % (query, str(x))
         self.c_size = resultCount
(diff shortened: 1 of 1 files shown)


joshmoore:
Searching for `a*` at https://trout.openmicroscopy.org/merge/webclient/search/
causes:
```
  File "/opt/hudson/workspace/OMERO-5.1-merge-deploy/src/dist/lib/python/omeroweb/webclient/controller/search.py", line 100, in search
    self.searchError = "Your query for '%s' caused an error: %s." % (query, x.reason)

AttributeError: 'InternalException' object has no attribute 'reason'
```

@aleksandra-tarkowska : did you try using https://docs.python.org/2/library/warnings.html ?

aleksandra-tarkowska:
My initial testing was based on raising exception within the code. It looks like due to change in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/gateway/__init__.py#L4102 we don't see deprecated warning anymore. I reverted to the original changes, but removed other as we only case about maxClauseCount

aleksandra-tarkowska:
the same is here:
 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webadmin/views.py#L577
 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webadmin/views.py#L584
 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/views.py#L2760
 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/views.py#L3118
 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/views.py#L3269
 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py#L131

components/tools/OmeroWeb/omeroweb/webclient/controller/search.py
```diff
@@ -97,10 +97,6 @@ def doSearch(searchType):
                     self.containers[dt] = doSearch(dt)
                     resultCount += len(self.containers[dt])
         except Exception, x:
-            msg = x.message
-            if "TooManyClauses" in msg:
+            if hasattr(x, "message") and "TooManyClauses" in x.message:
                 self.searchError = "Please try to narrow down your query. The wildcard matched too many terms."
-            else:
```
joshmoore: Why is this removed?

aleksandra-tarkowska:
Sorry just a mistake

aleksandra-tarkowska:
All omero.ServerError exceptions are safe as https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/resources/omero/ServerErrors.ice#L123
We may want to add some helper wrapper to validate which type of exceptions we are getting then we will have a better way of controlling it in one place

aleksandra-tarkowska:
All ```omero.ServerError``` exceptions are safe as https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/resources/omero/ServerErrors.ice#L123
We may want to add some helper wrapper to validate which type of exceptions we are getting, then we will have a better way of controlling it in one place. Something like:

```
class WebError(Exception):

    def __init__(self, exc):
        Exception.__init__(self, exc)
        if isinstance(exc, omero.ServerError):
            self._message = exc.message
        else:
            self._message = exc
        
    message = property(lambda self: self._message, 
                       lambda self, v: setattr(self, '_message', v))

    def __unicode__(self):
        return unicode(self.message)
```

@will-moore @knabar @chris-allan @sbesson @manics any thoughts?

manics:
Could you use `str(x)` for everything?

aleksandra-tarkowska:
Yes I did, but then you end up with the whole stack-trace in UI message. That is a maker problem

joshmoore:
Thanks for battling through this one, @aleksandra-tarkowska. New strategy makes sense to me (without having tested it yet)

joshmoore:
```
Please try to narrow down your query. The wildcard matched too many terms.
No results found
```
Assuming there's no easy way to get rid of the "No results found" on the page, I'd say :thumbsup: @aleksandra-tarkowska, what do you think?

aleksandra-tarkowska:
That is default icon. We would need to design new one for error. At this moment we really capture 1 error. Is that worth?

joshmoore:
Nope. If there's not something easy (like "returning empty string" or whatever), then let's not worry about it. Thanks.

PR #3496, opened 2015-02-18
Save "pending" MapAnnotations changes

dominikl:
Fixes [Ticket 12736](http://trac.openmicroscopy.org.uk/ome/ticket/12736)

If text was entered/modified within a MapAnnotation and the save button was hit, the latest changes weren't saved, unless the enter/tab button was pressed before to "confirm" the changes.

Test: Modify some keys/values of a MapAnnotation, don't hit enter/tab after the last change,but directly press save button (or switch do another DataObject in the browser - should automatically save the changes). Make sure also the last modification has been saved.

Note: The save button still doesn't become enabled on typing only; but as changes are saved automatically, I think we should remove the save button anyway in another PR (also to make it conform to Web).

--no-rebase

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/MapAnnotationsComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/MapAnnotationsComponent.java
@@ -456,6 +456,8 @@ public class MapAnnotationsComponent extends JPanel implements
 		t.addFocusListener(new FocusListener() {
 			@Override
 			public void focusLost(FocusEvent e) {
+                if (t.getCellEditor() != null)
+                    t.getCellEditor().stopCellEditing();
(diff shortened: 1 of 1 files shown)


pwalczysko:
Tested on trout with a custom build of @dominikl : all works as expected.

pwalczysko:
Although the first part here works as expected (i.e. the ticket has been solved as advertised), the second paragraph (when navigating away from unsaved text edits) does not work. 

I would suggest to still merge this PR, as it nicely solves the original problem.
I will create a follow-up ticket

dominikl:
Yes, I'll fix that in a followup PR, probably not for m4 though. Thanks @pwalczysko .

pwalczysko:
Created https://trac.openmicroscopy.org/ome/ticket/12744#ticket

With this, this PR is ready to merge.

pwalczysko:
Also after rebuild, this passed the test.

PR #3501, opened 2015-02-18
Automatically append / to omero.web.static_url

manics:
Automatically append / to omero.web.static_url if necessary instead of failing with a warning that a trailing slash is required. Note this also removes the unused `omeroweb.settings.remove_slash` method.

Testing: Test `omero.static.url` with/without a trailing /, and unset.

--no-rebase


aleksandra-tarkowska:
Thanks @manics for improving that, just these few small comments.

manics:
Thanks, updated.

aleksandra-tarkowska:
looks good, ready to merge

PR #3505, opened 2015-02-19
MapAnnotation bugfixes

dominikl:
Fixes several issues discovered during m4 testing:
- [Ticket 12747](https://trac.openmicroscopy.org.uk/ome/ticket/12747)
- [Ticket 12748](https://trac.openmicroscopy.org.uk/ome/ticket/12748)
- NPE when you tried to paste Key/Values when no MapAnnotation was selected and no own user table exists (e.g. admin annotating other user's DataObject in private group)
- Fixed bug which removed the text caret from cells (which had the effect, that you only could overwrite the cells content).


pwalczysko:
All four fixes are effective as we wanted:

1. Ctrl+V problem - no crash anymore
2. Problem of private group where no own annot but pasting without selection problem - no error anomore
3. Problem of copying rows on not-own data (but own annot) and editing them - no error anymore
4. Editing text (correcting typo) workflow problem - fixed

Ad 2. above -> it would be nice to disable the Paste icon in such case, but this is for another PR.

Please merge.

joshmoore:
Great, thanks guys!

PR #3506, opened 2015-02-19
Connect with session ID

jburel:
Allow to connect using session id.
A way to test the PR
 * Run code in eclipse for example and print out the session ID
 * launch the application
 * Copy and paste this id in user field, password can be anything but not empty
 * Click connect.

Note: I might go for empty password @gusferguson what do you think?

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
@@ -1823,11 +1823,28 @@ class OMEROGateway
 			else secureClient = new client(hostName);
 			secureClient.setAgent(agentName);
 			ServiceFactoryPrx entryEncrypted;
+
 			if (userName.equals(password)) {
-			    entryEncrypted = secureClient.joinSession(userName);
-	        } else {
-	            entryEncrypted = secureClient.createSession(userName, password);
-	        }
+			    ServiceFactoryPrx guestSession = secureClient.createSession("guest", "");
+			    boolean session = false;
+			    try {
+			        IAdminPrx prx = guestSession.getAdminService();
+		            ExperimenterData exp = (ExperimenterData) PojoMapper.asDataObject(
+		                    prx.lookupExperimenter(userName));
+		            session = exp == null;
+                } catch (Exception e) {
+                    session = true;
+                } finally {
+                    secureClient.closeSession();
+                }
+			    if (session) {
+			        entryEncrypted = secureClient.joinSession(userName);
+			    } else {
+			        entryEncrypted = secureClient.createSession(userName, password);
+			    }
+			} else {
+			    entryEncrypted = secureClient.createSession(userName, password);
+			}
 			serverVersion = entryEncrypted.getConfigService().getVersion();
 			String ip = null;
 	        try {
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
@@ -50,6 +50,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.openmicroscopy.shoola.util.CommonsLangUtils;
+
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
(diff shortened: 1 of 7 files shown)


components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -1814,13 +1816,18 @@ client createSession(String userName, String password, String hostName,
...
+		
 		try {
 		    // client must be cleaned up by caller.
 			if (port > 0) secureClient = new client(hostName, port);
 			else secureClient = new client(hostName);
 			secureClient.setAgent(agentName);
-			ServiceFactoryPrx entryEncrypted
-			    = secureClient.createSession(userName, password);
+			ServiceFactoryPrx entryEncrypted;
+			if (userName.equals(password)) {
```
manics: What if a user's password is the same as their username

mtbc: omero/omero? :smiley:

sbesson: or any user created by our integration tests at the moment

joshmoore: `joinSession(userName)` == `createSession(userName, userName)` so this should be fine, but you could try using the guest connection to look for a user of this name if you want to be really careful.

jburel:
I have to handle the case of username == password, (known) Just wanted to test it and mark it onhold

jburel:
i also need to remove session ID from user field.

jburel:
I will probably go ahead with @joshmoore suggestion and also empty password (or something like that)

jburel:
guest option does not seem to work.

components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
joshmoore: If guest isn't working, I really think this block or even the simpler `createSession(username, password)` is fine considering the implementation of `joinSession`:

https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/client.java#L635

gusferguson:
@jburel - I suspect the empty password would be the least confusing.

manics:
Created a CLI session, used this to log in with Insight, works as expected.

jburel:
@gusferguson: ok so I will review the log in screen.

jburel:
Enable login control if no password. @manics: could you test it again? this time no password needed.

components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -1814,13 +1813,29 @@ client createSession(String userName, String password, String hostName,
...
-			ServiceFactoryPrx entryEncrypted
-			    = secureClient.createSession(userName, password);
+		    if (port > 0) secureClient = new client(hostName, port);
+		    else secureClient = new client(hostName);
+		    secureClient.setAgent(agentName);
+		    ServiceFactoryPrx entryEncrypted;
+		    boolean session = true;
+		    try {
+		        //Check if it is a session first
+		        ServiceFactoryPrx guestSession = secureClient.createSession("guest", "guest");
```
aleksandra-tarkowska: isn't better to use ```getAdminService().getSecurityRoles()```

jburel: I am not connected at that point. That's the problem

dominikl:
Doesn't work for me. With session id + empty password I get "Failed to login" message, session id + random string as password works.

jburel:
@dominikl: thanks will check. I might have type something during my test

jburel:
@dominikl: done

manics:
It looks like the session-id is still saved in preferences- if you restart Insight the Username field is pre-populated with the last session-id.

jburel:
@manics: ok thanks.

jburel:
@manics: It is saved when the connection failed (it will always today)
If the connection is fine, then it should be kept. Only time we know it was a "session ID" credentials.

dominikl:
Works like expected. Can confirm, if the connection could be established the session-id is not saved in the preferences, on a Insight restart the user name field is empty.
What surprised me, I can still connect to the session, after I closed the original Insight instance (started from eclipse), which created the session. Shouldn't it be closed at this point, does Insight not close the sessions properly?
Anyway, would be good if this could be merged soon, so I can rebase #3513 .

jburel:
@dominikl: I discussed the session not being closed with @joshmoore.
This is outside the scope of this PR.
I have created a card. https://trello.com/c/NWT3qUma/430-client-del-session-still-alive

jburel:
--no-rebase

PR #3507, opened 2015-02-19
Improvements to disk usage plugin

ximenesuk:
This PR is intended to improve the output options for the disk usage plug-in now that size and file counts are available at a per-user, per-group, per-component level. Firstly it allows totals to be summed by any of the three parts above:
```
--sum_by {user,group,component} [{user,group,component} ...]`
```
which will sum the totals by one, two or all three options. Summing by all three options is the same as not using this flag at all, ie the default. This only has an effect when the `--report`` option is used.

The second addition is:
```
--pretty
```
This pretty-prints the sizes by making the human-readable. This works for both the single value and for the table when `--report` is used. It is, however, overridden by any `--units` setting for the single value. The other minor point is `--units` produces KiB, MiB, etc., while `--pretty` produces KB, MB, etc.

This PR is very much for comments.
 + I'm not wedded to the option names, I don't really like `--sum_by` even though it describes what it does.
 + Same goes with `--pretty`, I originally had `--human-readable`. Maybe adding single letter options would help?
 + Should `--units` apply to the table to, ie either forced units or human-readable across the board? The downside with this is that the smaller table entries may show as 0 for large units
 + KiB vs KB!? The method producing the pretty printing is used exclusively by the `fs` command as far as I can see so we could switch. Or add an `iec=False` argument, though that seems like overkill for what this does

TODO: Since these numbers pass through dictionaries the original ordering from the response object is lost. An order was never guaranteed but it tended to be ordered to the eye. So adding an ordering by columns option, probably pushed to TableBuilder, makes sense.

Testing instructions
--------------------------
See https://github.com/openmicroscopy/openmicroscopy/pull/3009 for basic instructions.
```
bin/omero fs usage -h
```
should provide more help. The two added options are `--pretty` and `--sum_by`. 

For `--pretty` try it with and without the `--units` option. The latter option should have precedence. Also try it with `--report`, the bytes column in the table should have more human readable numbers.

For  `--sum_by`, which only works in combination with `--report`, try it with all combinations of `user`, `group` and `component`. Using all three options should be the same as not using `--sum_by` at all. The final two columns should be right-justified for any use of `--report`.

Finally, check some other commands that produce tables, `bin/omero user list`, etc. The tables should be correctly formatted and all columns should be left justified.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 #
-# Copyright (C) 2014 Glencoe Software, Inc. All Rights Reserved.
+# Copyright (C) 2014-2015 Glencoe Software, Inc. All Rights Reserved.
 # Use is subject to license terms supplied in LICENSE.txt
 #
 # This program is free software; you can redistribute it and/or modify
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -822,21 +822,15 @@ Examples:
         """
         from omero.util.text import TableBuilder
 
-        sum_by = allCols = sortCols = ("user", "group", "component")
+        sum_by = ("user", "group", "component")
         if args.sum_by is not None:
-            sum_by = sortCols = args.sum_by
-        sortCols = list(sortCols)
-        sortCols.extend(["size", "files"])
+            sum_by = args.sum_by
+        showCols = list(sum_by)
+        showCols.extend(["size", "files"])
 
-        cols = []
-        align = ''
-        for col in allCols:
-            if col in sum_by:
-                cols.append(col)
-                align += 'l'
-        cols.extend(["size","files"])
+        align = 'l'*len(sum_by)
         align += 'rr'
-        tb = TableBuilder(*cols)
+        tb = TableBuilder(*showCols)
         tb.set_align(align)
         if args.style:
             tb.set_style(args.style)
@@ -886,7 +880,7 @@ Examples:
             keys = []
             for col in args.sort_by:
                 try:
-                    pos = sortCols.index(col)
+                    pos = showCols.index(col)
                     keys.append(pos)
                 except:
                     pass
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -286,7 +286,7 @@ class FsControl(CmdControl):
             "--units", choices="KMGTP",
             help="Units to use for disk usage")
         usage.add_argument(
-            "--pretty", action="store_true",
+            "--human_readable", action="store_true",
             help="Use most appropriate units")
         usage.add_argument(
             "--groups",  action="store_true",
@@ -787,7 +787,7 @@ Examples:
                 if args.units:
                     size = ("%d %siB"
                             % (self._to_units(size, args.units), args.units))
-                elif args.pretty:
+                elif args.human_readable:
                     size = filesizeformat(size)
                 self.ctx.out(
                     "Total disk usage: %s bytes in %d files"
@@ -813,7 +813,7 @@ Examples:
             if col in sum_by:
                 cols.append(col)
                 align += 'l'
-        if args.pretty:
+        if args.human_readable:
             cols.extend(["size", "files"])
         else:
             cols.extend(["size (bytes)", "files"])
@@ -859,7 +859,7 @@ Examples:
 
         for key in subtotals.keys():
             row = list(key)
-            if args.pretty:
+            if args.human_readable:
                 subtotals[key][0] = filesizeformat(subtotals[key][0])
             row.extend(subtotals[key])
             tb.row(*tuple(row))
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -39,6 +39,7 @@ from omero.rtypes import rstring
 from omero.rtypes import unwrap
 from omero.sys import Principal
 from omero.util.temp_files import create_path
+from omero.util.text import filesizeformat
 from omero.fs import TRANSFERS
 
 
@@ -285,6 +286,9 @@ class FsControl(CmdControl):
             "--units", choices="KMGTP",
             help="Units to use for disk usage")
         usage.add_argument(
+            "--pretty", action="store_true",
+            help="Use most appropriate units")
+        usage.add_argument(
             "--groups",  action="store_true",
             help="Print size for all current user's groups")
         usage.add_argument(
@@ -308,7 +312,6 @@ class FsControl(CmdControl):
     def _extended_info(self, client, row, values):
 
         from omero.cmd import ManageImageBinaries
-        from omero.util.text import filesizeformat
 
         rsp = None
         try:
@@ -354,7 +357,6 @@ Examples:
 
         from omero.rtypes import unwrap
         from omero.sys import ParametersI
-        from omero.util.text import filesizeformat
 
         select = (
             "select i.id, i.name, fs.id,"
@@ -810,7 +812,10 @@ Examples:
         for col in allCols:
             if col in sum_by:
                 cols.append(col)
-        cols.extend(["size (bytes)", "files"])
+        if args.pretty:
+            cols.extend(["size", "files"])
+        else:
+            cols.extend(["size (bytes)", "files"])
         tb = TableBuilder(*cols)
         if args.style:
             tb.set_style(args.style)
@@ -834,6 +839,8 @@ Examples:
 
         for key in subtotals.keys():
             row = list(key)
+            if args.pretty:
+                subtotals[key][0] = filesizeformat(subtotals[key][0])
             row.extend(subtotals[key])
             tb.row(*tuple(row))
 
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -294,7 +294,9 @@ class FsControl(CmdControl):
             help="Print size for all current user's groups")
         usage.add_argument(
             "obj", nargs="*",
-            help="Objects to be queried in the form '<Class>:<Id>[,<Id> ...]'")
+            help=("Objects to be queried in the form "
+                  "'<Class>:<Id>[,<Id> ...]', or '<Class>:*' "
+                  "to query all the objects of the given type "))
 
         for x in (images, sets):
             x.add_argument(
@@ -710,6 +712,10 @@ Examples:
     bin/omero fs usage --groups    # total usage for current user's groups
     # total usage for five images with minimal output
     bin/omero fs usage Image:1,2,3,4,5 --size_only
+    # total usage for all images with in a human readable format
+    bin/omero fs usage Image:* --human-readable
+    # total usage for all users broken down by user and group
+    bin/omero fs usage Experimenter:* --report --sum-by user group
     # total usage for two projects and one dataset Megabytes
     bin/omero fs usage Project:1,2 Dataset:5 --units M
     # in this last case if the dataset was within one of the projects
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -808,14 +808,18 @@ Examples:
             sum_by = args.sum_by
 
         cols = []
+        align = ''
         for col in allCols:
             if col in sum_by:
                 cols.append(col)
+                align += 'l'
         if args.pretty:
             cols.extend(["size", "files"])
         else:
             cols.extend(["size (bytes)", "files"])
+        align += 'rr'
         tb = TableBuilder(*cols)
+        tb.set_align(align)
         if args.style:
             tb.set_style(args.style)
 
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -785,7 +785,8 @@ Examples:
             else:
                 files = sum(rsp.totalFileCount.values())
                 if args.units:
-                    size = "%d %siB" % (self._to_units(size, args.units), args.units)
+                    size = ("%d %siB"
+                            % (self._to_units(size, args.units), args.units))
                 elif args.pretty:
                     size = filesizeformat(size)
                 self.ctx.out(
@@ -821,6 +822,7 @@ Examples:
         subtotals = {}
         for userGroup in rsp.bytesUsedByReferer.keys():
             for (element, size) in rsp.bytesUsedByReferer[userGroup].items():
+                files = rsp.fileCountByReferer[userGroup][element]
                 keyList = []
                 if "user" in sum_by:
                     keyList.append(userGroup.first)
@@ -828,12 +830,12 @@ Examples:
                     keyList.append(userGroup.second)
                 if "component" in sum_by:
                     keyList.append(element)
-                key=tuple(keyList)
+                key = tuple(keyList)
                 if key in subtotals.keys():
                     subtotals[key][0] += size
-                    subtotals[key][1] += rsp.fileCountByReferer[userGroup][element]
+                    subtotals[key][1] += files
                 else:
-                    subtotals[key] = [size, rsp.fileCountByReferer[userGroup][element]]
+                    subtotals[key] = [size, files]
 
         for key in subtotals.keys():
             row = list(key)
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -796,7 +796,7 @@ Examples:
         """
         err = self.get_error(rsp)
         if err:
-            self.ctx.err("Error: " + rsp.name)
+            self.ctx.err("Error: " + rsp.parameters['message'])
         else:
             size = sum(rsp.totalBytesUsed.values())
             if args.size_only:
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -824,16 +824,31 @@ Examples:
             tb.set_style(args.style)
 
         subtotals = {}
-        for userGroup in rsp.bytesUsedByReferer.keys():
-            for (element, size) in rsp.bytesUsedByReferer[userGroup].items():
-                files = rsp.fileCountByReferer[userGroup][element]
+        if "component" in sum_by:
+            for userGroup in rsp.bytesUsedByReferer.keys():
+                for (element, size) in rsp.bytesUsedByReferer[userGroup].items():
+                    files = rsp.fileCountByReferer[userGroup][element]
+                    keyList = []
+                    if "user" in sum_by:
+                        keyList.append(userGroup.first)
+                    if "group" in sum_by:
+                        keyList.append(userGroup.second)
+                    keyList.append(element)
+                    key = tuple(keyList)
+                    if key in subtotals.keys():
+                        subtotals[key][0] += size
+                        subtotals[key][1] += files
+                    else:
+                        subtotals[key] = [size, files]
+        else:
+            for userGroup in rsp.totalBytesUsed.keys():
+                size = rsp.totalBytesUsed[userGroup]
+                files = rsp.totalFileCount[userGroup]
                 keyList = []
                 if "user" in sum_by:
                     keyList.append(userGroup.first)
                 if "group" in sum_by:
                     keyList.append(userGroup.second)
-                if "component" in sum_by:
-                    keyList.append(element)
                 key = tuple(keyList)
                 if key in subtotals.keys():
                     subtotals[key][0] += size
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -767,7 +767,7 @@ Examples:
         oneK = 1024.0
         powers = {'K': 1, 'M': 2, 'G': 3, 'T': 4, 'P': 5}
         if units in powers.keys():
-            return round(size/oneK**powers[units], 2)
+            return round(size/oneK**powers[units], 1)
         else:
             raise ValueError("Unrecognized units: ", units)
 
@@ -786,7 +786,7 @@ Examples:
             else:
                 files = sum(rsp.totalFileCount.values())
                 if args.units:
-                    size = ("%d %siB"
+                    size = ("%s %siB"
                             % (self._to_units(size, args.units), args.units))
                 elif args.human_readable:
                     size = filesizeformat(size)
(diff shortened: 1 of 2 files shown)


ximenesuk:
These two commits add a right alignment option to the TableBuilder and then right align the final two columns of `fs usage`. This should not break use and alignment of TableBuilder in other plugins and default alignment (left for all columns) should still be apparent, ie `bin/omero user list` should still produce sensible output.

pwalczysko:
Docs problem here: As I did not know exactly how to execute the plugin in the first place, I tried to find it in our docs. 
I found that the feature was added in 5.0.2. And that it serves admins. Went to the sysadmin section of the docs. Because it is not in any of the headers, I did not find it.
Search did not help (you must know the exact wording, usage delivers too many hits, disk space usage delivers no hits).
Tried ``bin/omero -h`` - found nothing relevant in the output.

pwalczysko:
After discussion with @ximenesuk - please relist, more precise testing description is needed.

ximenesuk:
This last commit sums the totals differently if there is no breakdown by component. It uses the provided totals from the DiskUsage command. This is on the advice of @mtbc On a dirtyish (fs and pre-fs) system such as trout this should produce more accurate numbers as some files seem to be double-counted.

ximenesuk:
@pwalczysko raises some interesting points regarding documentation. Using search on something like 'disk usage' throws up lots of pages but not the correct ones. In addition using `bin/omero -h` only lists top level commands leaving someone guessing as to where to find subcommands. Maybe some sort of extended cli help would be useful? Either a tree view or a simple list of subcommands under each command? /cc @joshmoore @sbesson

sbesson:
Something we may want to review/expand but which does the trick is:

```
bin/omero help --all
```

ximenesuk:
@sbesson I didn't even realise that was there, @pwalczysko probably didn't either! It could certainly be more readable, I guess it is just concatenating all the calls to help?

sbesson:
@ximenesuk: agreed, this command is aggregating all help calls. Arguably we could have another form which is recursing through all subcommands (this would be the first step towards the resolution of https://trac.openmicroscopy.org/ome/ticket/10851).

In terms of usability, the question are:
- is `help` a good subcommand name? if not, do you have alternative suggestions?
- what would be a good syntax for commands vs commands + subcommands?
- what is the most appropriate way to present the otuput?

mtbc:
I wonder if there's a good place for all this output in the sysadmin docs so that it's more searchable.

sbesson:
One extreme possibility would certainly be to auto-generate some Sphinx documentation from this command (with modification) and use the `autogen_docs` logic to include the CLI auto-generated documentation in the documentation. Individual sections from each documentation (users/sysadmin/developers) could then refer to appropriate commands.

ximenesuk:
I think `help` is as good a name as any. The two options I can imagine, in addition to the current output of `--all` would be a full iteration, as you suggest, an a summary of subcommands for each command. So something like the current top-level help but with a line under each command,
```
...
fs     Filesystem utilities (or whatever)
        images, repos, sets, usage  (or whatever)
...
```
at least this way finding a subcommand would be relatively easy.

joshmoore:
```
./components/tools/OmeroPy/src/omero/plugins/fs.py:829:80: E501 line too long (81 > 79 characters)
```

ximenesuk:
Thanks @joshmoore not sure why that one slipped through.

pwalczysko:
Tried  ``bin/omero fs usage Experimenter:5 --report --pretty --units M``
This happened just because before reading the header I thought that ``--report`` is compulsory.
But now I think: Why is ``--units`` actually not working at all (no effect) in combination with ``--report`` ?
Also, in this situation (when ``--report`` is present) the precedence of ``--units`` before  ``--pretty`` does not seem to hold.

pwalczysko:
Test 2:

Using the command ``--sum_by group`` for user-6 in my local DB I get a discrepancy in the highest numbered group (user-6 has apparently data only in 3 groups, 3, 4, and 6). The number given by the ``--sum_by`` for group 6 is lower than the number counted using Excel sheet taken from the control command ``bin/omero fs usage``
See below (group 6 is listed in 1st column, the discrepancy starts at the 3rd numeral), the other two groups are counted precisely
347707415				964744369	20571106926
					
					
342204019				964744369	20571082804

pwalczysko:
Further to the comment above:
I do not understand where are the other users which have also data in the DB which user-6 (=Experimenter 7) can view.  For example user-2 has some data, as well as user test1, which can be seen by user-6. Why is just user-4 (= Experimenter 5) listed ? Just becuase he probably has some data in a group where user-6 also has some data ? How does admin (=user-6) actually find out the breakup of the whole DB per user, even of such users which do not have data which intersect in any manner with admins own data ?

```
[petr@LS25165 ~/Downloads/OMERO.server-5.1.0-m4-139-69fede3-ice35-b419]$ bin/omero fs usage  --report --sum_by group user
Using session f5096b82-7ab0-40c0-95d0-3ace4dab4787 (user-6@localhost:4064). Idle timeout: 10.0 min. Current group: read-only-1
Total disk usage: 21878031192 bytes in 5206 files
 user | group | size (bytes) | files 
------+-------+--------------+-------
 5    | 6     |       244970 |    98 
 7    | 3     |    964744369 |   722 
 7    | 6     |    341959049 |   498 
 7    | 4     |  20571082804 |  3888 
(4 rows)
```

ximenesuk:
On test 1 first: `--units` does not work for the report table, though if you use just this option with `--report` the total above the table will be shown with units. The reason for this is that the totals in the report table can vary by several orders of magnitude. This means that a unit that suits on component may not suit another. In fact for large units the smaller values may show as a misleading zero. This is a question I asked in the description. Also the interaction between `--units` and `--pretty` is a question I raise above.

pwalczysko:
Test 2 in total : the commands work, although I am not sure what they are showing actuall, see https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-75596603 and the summation seems to be not right see https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-75592179

ximenesuk:
On test 2: if you sum by either group, user or both -- but crucially not component -- then the totals are calculated differently to is components are detailed. This is because the response from the server contains totals by user/group and totals for each component separately. @mtbc may be able to express this better but there is some double-counting of some data depending on what binary files are present. Thus the totals by group/user are correct.

The reason user id 5 is there is because there may be data related to that user that belongs to `user-6`, again @mtbc may be able to explain this more clearly.

ximenesuk:
Re https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-75596603 again, @mtbc may be better explaining this but the disk usage isn't that which the user can "see".

ximenesuk:
@pwalczysko note that the plugin this PR tries to present the data given by the DiskUsage server command in useful ways. I hope that the numbers the plugin is presenting are correct according to the server response, but assuming they are then what those numbers mean is an issue for the server command to some extent. It may be that we really need to document what this thoroughly so that users will understand the tables output by this plugin.

pwalczysko:
@ximenesuk : Re https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-75598371 thanks, but genuinely do not understand the second part  of the answer. My worry is not that user id 5 is there, but rather all the users which are not there - how is this used by admins to find out how much in total each user/group has in the DB (user-6 in my test is an admin). Or is it not an aim of the plugin ? (what is then ?)

pwalczysko:
his PR is very much for comments.

    I'm not wedded to the option names, I don't really like --sum_by even though it describes what it does.

``I think sum_by is okay``
    Same goes with --pretty, I originally had --human-readable. Maybe adding single letter options would help?

``pretty is pretty awful. I think definitely it should be human-readable. Not sure about single letter options, as long as it is explained in help....``

    Should --units apply to the table to, ie either forced units or human-readable across the board? The downside with this is that the smaller table entries may show as 0 for large units

``Yes. If you go to the trouble to use the flag, then you probably understand the danger of being displayed zeros.``

    KiB vs KB!? The method producing the pretty printing is used exclusively by the fs command as far as I can see so we could switch. Or add an iec=False argument, though that seems like overkill for what this does

``Sorry, have to pass on this one. Need to ask and be advised about implications.``

ximenesuk:
@pwalczysko I don't know how many users you have on your system and how much data each of them have so I'm not sure what you are expecting. Some users have no data and as far as I know those zero amounts are not returned, again, @mtbc can hopefully confirm that.

pwalczysko:
Re https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-75601350 - I am mentioning only users which certainly do have data, like the test1 user. Hard to discuss though when I totally do not understand how the plugin actually gets its numbers in the first place.


Trusting that the numbers are summed correctly (although will probably need a session with @mtbc to even start understanding this) and that the users in the tables are displayed ``as they should be`` , although possibly this is not very useful atm as nobody understands it, this PR works as expected.

I have not answered the last question in the header of the PR, basically do not understand the implications of the question.

ximenesuk:
@pwalczysko thanks for the comments on my questions. I'll switch to `--human_readable` unless anyone else objects. I'm happy to apply `--units` to the table with the proviso that it may return zeroes. I will make `--units` and `--human-readable` mutually exclusive is possible, otherwise I'll have `--units` take precedence.

As for `KiB` vs `KB` I'd suggest a sit down and a cup of tea with @rleigh-dundee !

ximenesuk:
@pwalczysko to try to reproduce what you are seeing can I check that you have a pretty standard, `user-1`, `user-2`,... system and that users 1 through 6 have data? If so I'll try to look at this tomorrow morning by doing a few more imports.

ximenesuk:
Ah, sorry for being slow. In https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-75599795 the reason you only see certain users is that
```
bin/omero fs usage --report
```
made by `user-6` (id=7) is that same as:
```
bin/omero fs usage Experimenter:7 --report
```
ie not providing a series of objects defaults to the Experimenter object for the current user. For and admin to see all users they would have to list all users, something like
```
bin/omero fs usage Experimenter:1,2,3,4,5,6,7,8,9 --report
```
or a similar command across all groups.

It has been discussed whether a command like:
```
bin/omero fs usage Experimenter:* --report
```
to report on all users (visible to the current user) might be a useful shortcut.

ximenesuk:
@pwalczysko these last commits address some of the argument and formatting issues you raised. The final commit may or may not be useful. It unifies the number of decimal places between the two different unit defining options. Whether we want to truncate or round to the nearest unit or have a number of decimal places is open for debate.

mtbc:
(Yes, zero counts will be omitted from the response from the server.)

mtbc:
The current request is basically just,
```
class DiskUsage extends Request {
    omero::api::IdListMap objects;
};
```
Any thoughts on how you might like this API changed, so you don't have to enumerate user or group IDs? Nothing's jumping to my mind, but obviously *something*'s feasible. Of course, another option is that a helper could be added to the gateway to query the IDs.

mtbc:
If Alice comments on Bob's image, and Bob asks for his disk usage, it'll include Alice's comment under her user's count.

manics:
RFE (maybe unrelated to this PR?): consistency of help argument:
`omero help`, and `omero fs help` work, `omero fs usage help` doesn't (needs to be `--help`)

The use of `_` in argument names seems unconventional compared to most unix commands, could you use `-` instead?

Could an object without an ID by a synonym for all objects of that type, e.g. if an admin wants to get diskusages for all users could they run `omero fs usage --report --units M Experimenter --sum_by user`

ximenesuk:
@manics Thanks for the feedback.

Regarding the use of `help`: `omero fs help` doesn't really work, it just gives feedback because `help` is an unknown subcommand, `omero fs foobar` would give the same help. However, `help` is a top-level command and so `omero help` does work. See gh-3520 where any RFEs might be useful for the `help` command. 

We seem to use a mix of underscores and hyphens in cli help. I'd be happy to standardise to hyphens /cc @sbesson @joshmoore 

@mtbc does wonder if passing a null object id list could be (ab)used to return the totals for all objects.

ximenesuk:
Last commit moves to hyphens in options, definitely in the majority.

joshmoore:
NB: If is the general agreement, then we might want to do this minimally for `import` but keep the old versions for backwards compatibility.

sbesson:
Last RFE captured in https://trello.com/c/l5vppgeI

mtbc:
@ximenesuk: see #3528 for your `Experimenter:*` wish

pwalczysko:
@ximenesuk : siorry, still did not get to this PR today. Yes, I have the normal DB with user-1 through user-12. Most of the data are in the first 6 users. But also there are another users, like ``test1`` which have data as well.

ximenesuk:
@pwalczysko no problem. There will be more commits coming to address gh-3528 which will hopefully fix the issue you are seeing. As explained in https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-75615537 it is a current limitation that gh-3528 should lift.

ximenesuk:
This PR now
--depends-on #3528

ximenesuk:
@pwalczysko with this latest commits you can now do:
```
bin/omero fs usage Experimenter:* --report
```
and as an admin this should give the sort of output you were expecting above, ie all users. The `*` can be applied to any of the object types and mixed with specific ids of other types:
```
bin/omero fs usage Project:* Dataset:1,2,3 --report
```
in which case a union of the two object lists would be returned.

ximenesuk:
These last two commits should restore some order to the report table though it is not user configurable.

pwalczysko:
Bug: connecting to trout as user-6, and executing below I get an error

```
[petr@LS25165 ~/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424]$ bin/omero fs usage Group:* --report --sum-by user
Using session dc5fe1fb-27c0-4aa8-b0c4-f2b5a60a8059 (user-6@trout.opy.org:4064). Idle timeout: 10.0 min. Current group: read-annotate-1
failed: 'disk usage operation failed'
	message=Illegal query:SELECT id FROM Group
Group is not mapped [SELECT id FROM Group]
	stacktrace=ome.services.query.QueryException: Illegal query:SELECT id FROM Group
Group is not mapped [SELECT id FROM Group]
	at ome.services.query.StringQuery.buildQuery(StringQuery.java:84)
	at ome.services.query.Query.doInHibernate(Query.java:222)
	at org.springframework.orm.hibernate3.HibernateTemplate.doExecute(HibernateTemplate.java:406)
	at org.springframework.orm.hibernate3.HibernateTemplate.execute(HibernateTemplate.java:339)
	at ome.logic.QueryImpl.projection(QueryImpl.java:445)
	at omero.cmd.graphs.DiskUsageI.getDiskUsage(DiskUsageI.java:447)
	at omero.cmd.graphs.DiskUsageI.step(DiskUsageI.java:362)
	at omero.cmd.graphs.DiskUsageI.step(DiskUsageI.java:65)
	at omero.cmd.HandleI.steps(HandleI.java:436)
	at omero.cmd.HandleI$1.doWork(HandleI.java:365)
	at omero.cmd.HandleI$1.doWork(HandleI.java:361)
	at sun.reflect.GeneratedMethodAccessor278.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:307)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:183)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:150)
	at ome.services.util.Executor$Impl$Interceptor.invoke(Executor.java:566)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	at ome.security.basic.EventHandler.invoke(EventHandler.java:154)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	at org.springframework.orm.hibernate3.HibernateInterceptor.invoke(HibernateInterceptor.java:111)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:108)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	at ome.tools.hibernate.ProxyCleanupFilter$Interceptor.invoke(ProxyCleanupFilter.java:241)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	at ome.services.util.ServiceHandler.invoke(ServiceHandler.java:116)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:202)
	at com.sun.proxy.$Proxy72.doWork(Unknown Source)
	at ome.services.util.Executor$Impl.execute(Executor.java:447)
	at omero.cmd.HandleI.run(HandleI.java:359)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
	at ome.services.util.Executor$Impl$1.call(Executor.java:488)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:744)


    rv = omero.cli.argv()
  File "/Users/petr/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424/lib/python/omero/cli.py", line 1412, in argv
    cli.invoke(args[1:])
  File "/Users/petr/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424/lib/python/omero/cli.py", line 927, in invoke
  File "/Users/petr/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424/lib/python/omero/cli.py", line 1004, in onecmd
  File "/Users/petr/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424/lib/python/omero/cli.py", line 1086, in execute
    args.func(args)
  File "/Users/petr/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424/lib/python/omero/plugins/fs.py", line 753, in usage
    self._usage_report(req, rsp, status, args)
  File "/Users/petr/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424/lib/python/omero/plugins/fs.py", line 816, in _usage_report
    self._detailed_usage_report(req, rsp, status, args)
  File "/Users/petr/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424/lib/python/omero/plugins/fs.py", line 857, in _detailed_usage_report
    for userGroup in rsp.totalBytesUsed.keys():
AttributeError: 'ERR' object has no attribute 'totalBytesUsed'
[petr@LS25165 ~/Downloads/OMERO.server-5.1.0-m4-285-82fccf6-ice35-b424]$ 
```

pwalczysko:
Re https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-76235320 - I get the same error when I replace the ``Group:*`` by ``Bla:*`` which means that the object is not known - still think this should not end in an error. 
I got what I wanted by doing ``bin/omero fs usage ExperimenterGroup:* --report --sum-by user``

pwalczysko:
Why not call the Group a ``Group`` also as an object, why to use ``ExperimenterGroup`` ?

pwalczysko:
From  help output
```
# in this last case if the dataset was within one of the projects
    # then the size returned would be identical to:
    bin/omero fs usage Project:1,2 --units M
```
The second hash seems out of place, because it is just continuing the point&sentence (see formatting in the same paragraph above these lines.
Also reformulate ``...was within one of the projects 1 or 2...`` (add the ``1 or 2`` to clarify the projects in question)

pwalczysko:
From today's testing 1 RFE https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-76241431 and 1 Bug https://github.com/openmicroscopy/openmicroscopy/pull/3507#issuecomment-76235320 .
I think in general it is much improved, including the naming, help output (leave you guys to sort out the ``help`` ``-h`` and ``--help`` consistency issues) and usability for an admin.

ximenesuk:
@pwalczysko Thanks for the feedback. I'll fix trapping the error so that the user doesn't get this output and I'll tidy up the help output.

The names used here for the objects are those used by the underlying model, thus `Experimenter` and `ExperimenterGroup` where we might use `user` and `group` at a documentation level. I'd be loathed to allow alternative object names to be used in this plugin as it is. Maybe there's a way system-wide aliases could be used nearer to the server so that cli tools could hide these names for ordinary users. I'm sure @joshmoore could clarify that.

I have added some later commits which add `--sort-by` as an additional option which I'd be happy to have reviewed before fixing the issues above but I'll not re-list this PR today.

mtbc:
@ximenesuk: I could try to detect non-existent class names and throw some specific exception for you server-side if that would help, just let me know what's needed.

Indeed, the class names are exactly those at http://www.openmicroscopy.org/site/support/omero5.1/developers/Model/EveryObject.html. "ExperimenterGroup" crops up everywhere from Blitz logs to database tables, so OMERO server admins should know the term before long anyway, as will client/script developers from pages like http://downloads.openmicroscopy.org/omero/5.1.0-m4/api/slice2html/omero/api/IAdmin.html. I too would balk at confusing the widely used class names with an additional list of synonyms, because in many other ways of working with OMERO those synonyms still won't work. Perhaps a good compromise might be extra convenience options for disk usage like `--allusers` or `--allgroups` as an alternative to specifying the class names directly?

ximenesuk:
@mtbc I'll let you know if anything is needed command-side, I think the problem here is that after this exception the plugin is still trying to produce output which is a regression. I'll fix that and then see whether the command can be more helpful.

As for the object names we do, of course, have the commands `user` and `group`!

mtbc:
:+1:

pwalczysko:
@ximenesuk : If you mean that you would like to see the ``--sort-by`` flag tested, then I did it, I just did not write about it yesterday. Although your commits were late, there was one extra daytime rebuild of trout yesterday, and the flag worked nicely.

ximenesuk:
@pwalczysko Yes, that's what I meant, I wasn't sure of the timing of the commits, the builds and your testing. Thanks!

mtbc:
#3550 may be useful.

ximenesuk:
This now takes the error message from rsp.

--depends-on #3550

ximenesuk:
These later commits fix a usage example suggested by @pwalczysko and give a sensible error message. In particular using a non-existent class, `Blah:1` or `Blah:*` should give a message rather than a stacktrace.

pwalczysko:
With the last commit the `` bin/omero fs usage Group:* --report --sum-by user`` gives the output ``Error: class Group is unknown``.
Works as expected.

joshmoore:
:thumbsup: -- anything on the integration test side to await or record before merging?

ximenesuk:
Not for the moment from me, given how long this PR has run for I'd quite like to get it merged as is if possible. There is the dependency on gh-3550 to consider and that PR may need wider testing since it goes beyond the disk usage command.

ximenesuk:
--no-rebase

PR #3510, opened 2015-02-20
FileExclusion: new extension for the ImportLibrary

joshmoore:
Similar to the FileTransfer interface, well-known or third
party FileExclusion instances can be passed to the ImportLibrary
in order to influence the import process. Each exclusion is
given the chance to veto the import of a given ImportContainer
which has been chosen, e.g., by ImportCandidates. For example,
passing `--exclude=filename` searches for a file with the same
name on the server and then subsequently checks that the hash
matches. If it goes, the call to `importImage` exits early with
the log statement:

```
INFO   ....exclusions.FilenameExclusion - Checksum match for filename:
```

Testing:
 * `bin/omero import $SOMEFILE`
 * `bin/omero import $SOMEFILE # This should produce a duplicate`
 * `bin/omero import -- --exclude=filename $SOMEFILE # This should import nothing.`
 * `bin/omero import -- --exclude=filename $SOMEOTHERFILE # imports`
 * `bin/omero import -- --exclude=filename $SOMEOTHERFILE # does not import`

cc: @jwgwarren @ximenesuk @jburel @mtbc 

*API comments welcome as are ideas for new implementations to test that API*

### DIFF
--- a/components/blitz/src/ome/formats/importer/ImportEvent.java
+++ b/components/blitz/src/ome/formats/importer/ImportEvent.java
@@ -215,6 +215,12 @@ public class ImportEvent {
         }
     }
 
+    public static class FILESET_EXCLUSION extends FILE_UPLOAD_EVENT {
+        public FILESET_EXCLUSION(String filename, int fileIndex, int fileTotal) {
+            super(filename, fileIndex, fileTotal, 0l, 0l, null);
+        }
+    }
+
     public static class FILESET_UPLOAD_PREPARATION extends FILE_UPLOAD_EVENT {
         public FILESET_UPLOAD_PREPARATION(String filename, int fileIndex,
                 int fileTotal, Long uploadedBytes, Long contentLength,
--- a/components/blitz/src/ome/formats/importer/ImportLibrary.java
+++ b/components/blitz/src/ome/formats/importer/ImportLibrary.java
@@ -32,6 +32,8 @@ import loci.common.Location;
 import loci.formats.FormatException;
 import loci.formats.FormatReader;
 import ome.formats.OMEROMetadataStoreClient;
+import ome.formats.importer.ImportEvent.FILESET_EXCLUSION;
+import ome.formats.importer.exclusions.FileExclusion;
 import ome.formats.importer.transfers.FileTransfer;
 import ome.formats.importer.transfers.TransferState;
 import ome.formats.importer.transfers.UploadFileTransfer;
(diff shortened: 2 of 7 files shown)


sbesson:
In terms of testing, the scenario described in this PR can minimally be added as it is to `test/integration/clitest/test_import.py`.

ximenesuk:
And
 + `bin/omero import -- --exclude=filename $SOMEOTHERFILE`
should work the first time?

joshmoore:
> bin/omero import -- --exclude=filename $SOMEOTHERFILE should work the first time?

Yes.

jwgwarren:
And presumably this works on the directory level but doesn't exclude the directory but any files underneath that are already present?

joshmoore:
@jwgwarren : not at the moment. Can you explain more about how you would see that working?

jwgwarren:
so I'd like to be able to put a directory argument in like I can at the moment but as you have suggested  --exclude /mydir/ then it will compare the original upload path of the images owned by my user and in the same group and if path and filename/digest the same then don't upload it.

joshmoore:
`--exclude` shouldn't change whether or not you can pass the whole directory. Each fileset (as found by `bin/omero import -f`) will get checked against the filename and the checksum individually. I began with a version (`--exclude=fullpath`) but was unsure if it was going to handle path names well enough (i.e. differences between Windows & Unix). Certainly happy to look back into that.

ximenesuk:
Tested with individual files, previously imported and as-yet un-imported, whole folders including some or all existing imports, etc. Everything looks good including good feedback on the cli.

joshmoore:
Thanks, @ximenesuk. @jwgwarren, anything else you see pressing here, or is it more a matter of differing implementations that you may need, which could be handled in another PR?

ximenesuk:
I've not reviewed the code but functionally I'd say merge this and then add new implementations later.

ximenesuk:
Okay, one question after reading through the code. Should there be any output of the new event via the LoggingImportMonitor?

joshmoore:
Ah, yes. I was wondering why it hadn't appeared. Unknown events are only at DEBUG. Happy to add this:
```
2015-02-24 08:27:27,420 4043       [      main] INFO   ts.importer.exclusions.FilenameExclusion - Checksum match for filename: date.fake
2015-02-24 08:27:27,424 4047       [      main] INFO   ormats.importer.cli.LoggingImportMonitor - FILESET_EXCLUSION
```
but it does seem a bit duplicated.

ximenesuk:
Probably fine as it is, just sanity checking. The cli output is clear enough and the event is available if other clients need it. I'd say this is good to merge as is.

jwgwarren:
This all sounds great to me. We would prefer a very brief informative one line log message for our use case over a longer detailed one.

joshmoore:
Merging as is then. As @ximenesuk, Java clients can catch the event and handle it differently. @jwgwarren, if you have any specific ideas on what type of exclusion you need/want, don't hesitate to speak and/or open a PR :smile:

PR #3512, opened 2015-02-20
imageMarshal getTileSize if tiles

will-moore:
This is a simple workaround that allows the webclient to open the MicroManager file that fails with re.getTileSize() in https://trac.openmicroscopy.org/ome/ticket/12750

We do not need to call re.getTileSize() if it's not a big image.
Insight may already be applying this logic (or catching the exception) since this image is viewable in Insight.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -158,7 +158,6 @@ def imageMarshal(image, key=None):
 
     try:
         rv.update({
-            'tiles': tiles,
             'size': {'width': image.getSizeX(),
                      'height': image.getSizeY(),
                      'z': image.getSizeZ(),
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -97,7 +97,7 @@ def imageMarshal(image, key=None):
             'projectId': pr and pr.id or None,
             'projectDescription': pr and pr.description or '',
             'datasetName': ds and ds.name or 'Multiple',
-            'datasetId': ds and ds.id or '',
+            'datasetId': ds and ds.id or None,
             'datasetDescription': ds and ds.description or '',
             'wellSampleId': wellsample and wellsample.id or '',
             'wellId': well and well.id.val or '',
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -134,7 +134,9 @@ def imageMarshal(image, key=None):
 
     # big images
     tiles = image._re.requiresPixelsPyramid()
-    width, height = image._re.getTileSize()
+    width, height = 256, 256
+    if (tiles):
+        width, height = image._re.getTileSize()
     levels = image._re.getResolutionLevels()
     zoomLevelScaling = image.getZoomLevelScaling()
     nominalMagnification = image.getObjectiveSettings() is not None \
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -134,11 +134,18 @@ def imageMarshal(image, key=None):
 
     # big images
     tiles = image._re.requiresPixelsPyramid()
-    width, height = 256, 256
+    rv['tiles'] = tiles
     if (tiles):
         width, height = image._re.getTileSize()
-    levels = image._re.getResolutionLevels()
-    zoomLevelScaling = image.getZoomLevelScaling()
+        levels = image._re.getResolutionLevels()
+        zoomLevelScaling = image.getZoomLevelScaling()
+
+        rv.update({'tile_size': {'width': width,
+                                 'height': height},
+                   'levels': levels})
+        if zoomLevelScaling is not None:
+            rv['zoomLevelScaling'] = zoomLevelScaling
+
     nominalMagnification = image.getObjectiveSettings() is not None \
         and image.getObjectiveSettings().getObjective().getNominalMagnification() \
         or None
@@ -152,9 +159,6 @@ def imageMarshal(image, key=None):
     try:
         rv.update({
             'tiles': tiles,
-            'tile_size': {'width': width,
-                          'height': height},
-            'levels': levels,
             'size': {'width': image.getSizeX(),
                      'height': image.getSizeY(),
                      'z': image.getSizeZ(),
@@ -166,8 +170,6 @@ def imageMarshal(image, key=None):
             })
         if init_zoom is not None:
             rv['init_zoom'] = init_zoom
-        if zoomLevelScaling is not None:
-            rv.update({'zoomLevelScaling': zoomLevelScaling})
         if nominalMagnification is not None:
             rv.update({'nominalMagnification': nominalMagnification})
         try:
--- a/components/tools/OmeroWeb/test/integration/test_marshal.py
+++ b/components/tools/OmeroWeb/test/integration/test_marshal.py
@@ -104,9 +104,9 @@ class TestImgDetail(object):
         print imgData
         # Not a big image - tiles should be False with no other tiles metadata
         assert imgData['tiles'] == False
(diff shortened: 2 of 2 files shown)


aleksandra-tarkowska:
Thank @will-moore for picking it up. Although as discussed we should move the whole code related to pyramid images to one section to avoid unnecessary calls, like:

```
if (tiles):
    ...
```

will-moore:
Removed big image data from json if tiles is false.
Test that big image viewing and non-big image viewing still works OK.
json data at ``` webclient/imgData/14/ ``` should not include ```tile_size```, ```levels``` or ```zoomLevelScaling``` if the image is not big ('tiles' is false).

dominikl:
I could import and view the image, also other images (big/tiled images) worked fine. Couldn't notice a difference between Insight and Web with respect to these images, however I couldn't find the location of the equivalent code in Insight, so couldn't verify that the logic really is identical. May have to pass this on to @jburel

will-moore:
@dominikl I don't think it needs a direct comparison with Insight. As long as this is working, and makes sense in web, we are good to merge. OK with you @aleksandra-tarkowska?

components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
```diff
@@ -134,9 +134,18 @@ def imageMarshal(image, key=None):
 
     # big images
     tiles = image._re.requiresPixelsPyramid()
-    width, height = image._re.getTileSize()
-    levels = image._re.getResolutionLevels()
-    zoomLevelScaling = image.getZoomLevelScaling()
+    rv['tiles'] = tiles
+    if (tiles):
```
aleksandra-tarkowska: are you sure they will not throw AttributeError?

will-moore: Yes. You don't need a 'tiles' attribute to be present:
```
>>> rv = {}
>>> rv['tiles'] = True
```

aleksandra-tarkowska: I am asking about all of the attributes below, they were in try catch block. Are you sure it is not needed anymore?

will-moore: I don't know what the ``` AttributeError ``` is designed to catch, but the only code that was moved out of the try catch block are additions to the rv dict by rv.update() which cannot raise AttributeErrors.

aleksandra-tarkowska: @will-moore it raises when an attribute reference or assignment fails.

@chris-allan @knabar do you remember why ```AttributeError``` was cached on image object exactly? Is that when RE is None or any getters?

chris-allan: I'm not sure. You'd have to go through all the code paths. This code is ancient and goes well, well back to before 2012 (4.4.0) unchanged. The `try` block you're referring to is as it was when I originally moved all of this marshalling out of `omeroweb.webgateway.views` as part of #313 in 2012 so that shape marshalling could be unit tested.

I would encourage this PR to be unit and/or integration tested as well. Extending `test/unit/test_marshal.py` would be perfect.

will-moore: While I agree that more testing is generally better, I don't think that moving a couple of lines out of a AttributeError try catch requires a new test when those lines can't throw AttributeError. Also, it's not really a unit test because this code requires a ImageWrapper (Big Image) with rendering engine etc.

chris-allan: My request for tests is not specifically because of moving lines out of the `try` block. There are currently no tests for `imageMarshal()` at all. If we're going to change a marshalling method we should have tests for it. This is especially relevant when we know that `imageMarshal()` backs possibly the most significant and public JSON API endpoint we currently have (`.../imgData/`).

If you want to perform that testing by building an integration test against `.../imgData/` that seems fine.

will-moore:
@sbesson Are there any images with more metadata (channels etc) available from library than: ```itest.importSingleImage(client=client)```?

sbesson:
@will-moore: `import_image()` with no argument will import `tinyTest.d3d.dv`. Re the new integration test, any reason not to use the `IWebTest` (which does all the setup including the django_client, see `test_simple.py`)?

will-moore:
@sbesson I don't see IWebTest in ```test_simple.py``` https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/test/integration/test_simple.py

will-moore:
@sbesson Another thing is bugging me now... I'm editing code under components/tools/OmeroWeb/omeroweb/webgateway/marshal.py but these changes seem to be ignored by the tests, even though the corresponding marshal.pyc is created when I run the tests. Are the tests using a different copy of this code in a different location?

sbesson:
Sorry, wrong link indeed (although I should let `test_simple.py` use `IWebTest` to show minimal examples of usage). See https://github.com/sbesson/openmicroscopy/commit/4932dd17675f7bfc5487eb98ee1dccc60a265a35 for a conversion from the fixtures to the usage of `IWebTest`.

I think the Web tests infrastructure is currently using the code under `components/tools/target/lib/python`. To keep things in sync, I usually run `./build build-py` before running tests.

will-moore:
@sbesson Yes it is using code under that target. I'd looked under a bunch of other /target dirs but not that one!
However, running ```./build.py build-py``` didn't help. I don't know where it copied the code from, but it wasn't components/tools/OmeroWeb/omeroweb/webgateway/marshal.py! Anyway, I manually copied my code across and now it's working.

Minor comment: Slightly unexpected how lib.import_image() returns a list of PixelIds instead of the image that is returned by lib.importSingleImage(). Just need to load Pixels to get the imageId...

Cheers

will-moore:
@chris-allan Added a test with hard-coded checks of all values in webgateway/imgData json. Also made a tiny change to the imageMarshal itself for consistency: https://github.com/will-moore/openmicroscopy/commit/877d2f1ffa38217bb6f3f8a868a3d9c9cb6ae103

chris-allan:
Tests look great @will-moore! Very nice to have.

As @sbesson has tried to do with the other tests we should probably rationalise some of these fixtures so that they do not have to be copy and pasted into every set of integration tests. You might have a discussion with him to see if that's possible with what he has already done. Fixing that all completely is definitely not in scope for this PR but we should limit potential churn of fixture updates where possible based on what has already been done.

sbesson:
Seconding @chris-allan, I actually have a branch doing part of that https://github.com/sbesson/openmicroscopy/tree/webtests_cleanup. Once this PR is merged, I will open a cleanup PR. Happy to refactor this new marshal tests to use the library utilities there.

will-moore:
@sbesson When I tried to use IWebTest as in https://github.com/sbesson/openmicroscopy/commit/4932dd17675f7bfc5487eb98ee1dccc60a265a35

I got following error (so I've not pushed it):

```
django_client = <django.test.client.Client object at 0x10abe7d50>, request_url = '/webgateway/imgData/7518/', query_string = '', status_code = 200

    def _get_reponse(django_client, request_url, query_string, status_code=405):
        query_string = urlencode(query_string.items())
        response = django_client.get('%s?%s' % (request_url, query_string))
>       assert response.status_code == status_code
E       assert 500 == 200
E        +  where 500 = <omeroweb.http.HttpJavascriptResponseServerError object at 0x10d622dd0>.status_code

test/integration/weblibrary.py:93: AssertionError
```

sbesson:
@will-moore: can you push your code with the error to another branch for examination?

will-moore:
@sbesson OK, so it seems that I was getting the ```client``` the wrong way. Now it seems to be working. 
NB: I have used the ``` _get_reponse ``` so you'll need to rename it here when you open your PR. Thanks!

sbesson:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-web/lastCompletedBuild/testReport/test.integration.test_marshal/TestImgDetail/ is green and the refactoring of the test to use `IWebTest` was very much appreciated (will fix the `_get_reponse` in my cleanup PR once this is merged). @chris-allan @aleksandra-tarkowska : are you happy with the state of this PR?

chris-allan:
All good here. :+1:

PR #3514, opened 2015-02-20
Use 'Active Group' instead of 'Default Group' in myaccount form

will-moore:
Tiny fix for https://trello.com/c/GkdpN8wD/311-rfe-change-default-group-to-active-group-in-uis

In user's settings form, we use "Active Group" instead of "Default Group".
Anywhere in the UI where this needs changing?

screen shot 2015-02-20 at 14 21 58

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -295,6 +295,7 @@ class MyAccountForm(NonASCIIForm):
             if kwargs['initial']['default_group']:
                 pass
             self.fields['default_group'] = GroupModelChoiceField(
+                label="Active Group",
                 queryset=kwargs['initial']['groups'],
                 initial=kwargs['initial']['default_group'],
                 empty_label=None)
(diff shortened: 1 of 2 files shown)


aleksandra-tarkowska:
@will-moore I think there are few more places we use *Default group* term:

 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webadmin/custom_forms.py#L416
 - https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/myaccount.html#L59

will-moore:
After discussion with @pwalczysko and @aleksandra-tarkowska yesterday, we thought that instead of renaming "Default Group" to "Active Group", we should simply disable the switching of default group on log-out of web and Insight. Then, the user would always return to their "default group" on login and it would be clearer that this is different from their "active group".

Also, we noted that when an Admin is editing a user's profile in Web, they can't choose the user's default group. Which means the user has to do it themselves. This is just one more thing for users to learn, so we should re-show and test the default user widget (just hidden now). cc @gusferguson @jburel @dominikl

PR #3515, opened 2015-02-20
adjust graphs API to provide Collections-based Java interface

mtbc:
OMERO should still work, including moves and deletes, and tests should still pass: there should be no change in behavior.

*Update:* Now provides Insight with a utility class for constructing graphs requests.

--no-rebase

### DIFF
--- a/components/blitz/resources/omero/cmd/Graphs.ice
+++ b/components/blitz/resources/omero/cmd/Graphs.ice
@@ -201,6 +201,7 @@ module omero {
              * specific kinds of children.
              * Only the first applicable option takes effect.
              **/
+            ["java:type:java.util.ArrayList<omero.cmd.graphs.ChildOption>:java.util.List<omero.cmd.graphs.ChildOption>"]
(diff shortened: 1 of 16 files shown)


jburel:
Requests should probably go out of insight since it will be useful for other applications.
This could be done when refactoring gateway.

mtbc:
Agreed, it should be scooped up with the gateway when that goes wherever it's off to. (I assume it will do so naturally as the gateway will use it.)

dominikl:
I guess once this is merged I'm going to make the switch over in Insight from Chgrp to Chgrp2, etc., right?

mtbc:
Indeed. http://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Server/GraphsMigration.html should include some useful hints, but also of course feel free to ask.

dominikl:
Looks fine to me. Integration tests passed, did some Chgrps and Deletes, didn't notice any problems.

joshmoore:
@dominikl : all yours, then.

PR #3517, opened 2015-02-23
Changed default group name to 'active group'

dominikl:
Tiny PR changing the name of the 'Default Group' to 'Active Group'
See [Trello Card](https://trello.com/c/GkdpN8wD/311-rfe-change-default-group-to-active-group-in-uis)
(corresponding Web PR: #3514 )

### DIFF
--- a/components/blitz/src/pojos/ExperimenterData.java
+++ b/components/blitz/src/pojos/ExperimenterData.java
@@ -244,7 +244,7 @@ public class ExperimenterData extends DataObject {
      * 
      * @return See above.
      */
-    public GroupData getDefaultGroup() {
+    public GroupData getActiveGroup() {
     	List<GroupData> groups = getGroups();
     	if (groups == null || groups.size() == 0) return null;
         return groups.get(0);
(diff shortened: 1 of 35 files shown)


mtbc:
Also change the batch call description in `org.openmicroscopy.shoola.env.data.views.calls.AdminLoader.java`?

jburel:
Change in UI made
Problem about changing active group still exists, does not have to be done in this PR
  * log in as a user members of at least 2 groups
  * add the 2 groups to display
  * Change the active group.

as suggested by @mtbc, maybe update the description so it is less confusing.

dominikl:
To be more consistent I'm going to make some more changes within this PR, e. g. also change the method name in the ExperimenterData Pojo.

jburel:
@dominikl: if you change the pojo you will need to add a new method and mark the existing one "deprecated"

dominikl:
Thanks, nearly forgot that. The problem with the change of the active group is a bit trickier than I initially thought. I'll tackle this in another PR. But changing the logged in user's active group and confirming this via the Save button should work, what doesn't work is changing the active group and then switching to another object (which triggers the confirmation dialog, if you want to save the changes). But as I said, better to open a separate PR for that.

dominikl:
Closing after discussion

PR #3521, opened 2015-02-23
Auto save 1 second after Map Annotation Edit

will-moore:
Fix for https://trac.openmicroscopy.org/ome/ticket/12754

There is no nice way to catch the click on the Tree and save any current Map Annotation edits before clearing the right panel.
An alternative solution is to simply automatically save a second after the last edit.
We already save the whole Map Annotation to the server whenever you leave a Key/Value input field and the 1 second delay should limit the number of calls to reduce synchronous updates.

To test: 
 - simply try a number of rapid edits, moving between key/value fields, with short delays etc
 - try editing a field, then navigating away in the Tree while still editing the field. Change should still be saved (unless you are very quick)!

--no-rebase

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/mapannotations.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/mapannotations.html
@@ -28,7 +28,7 @@
         var $input,
             inputEdited = false,
             saveInProgress = false,
-            saveTimeout,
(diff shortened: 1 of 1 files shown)


gusferguson:
@will-moore 

Tested using https://trout.openmicroscopy.org/merge/webclient/ user-2 private-1

Behaves as expected.
Even with mouse poised over tree I could not click fast enough to prevent edit being saved - so solution works well.
Good to merge.

PR #3523, opened 2015-02-23
catch SecurityViolation from setFileId

mtbc:
Prompted by https://trello.com/c/sWlIbIsi/426-setfileid-exception.

--no-rebase


joshmoore:
Change looks fine though there's still the question of how to reproduce and what other situations could cause this to happen. I collected further logs (though from 5.0) today and attached to the card.

mtbc:
Well, I don't get anything in `master.err` if I try a big-image JPEG import with the current 5.1-latest bug. (It seemed worth a try!)

PR #3519, opened 2015-02-23
DB 5.1__0

joshmoore:
This is primarily a renaming of the DEV20 DB to the final release name but does include:

 * a minor fix to the pressure unit, renaming `kBar` to `kbar`
 * a widening of the map annotation type
 * an addition of a `lookupTable` column to channel and channelbinding
 * a restriction to PG >= 9.2
 * a bump to the PG JDBC driver

cc: @mtbc @qidane

PR #3524, opened 2015-02-23
Update PostgreSQL JDBC driver to 9.4

rleigh-dundee:
Testing: server should start and run correctly.  Test that binary data is stored and retrieved correctly.


mtbc:
Do we still actually need to include the new JAR in the local git repo?

joshmoore:
No we don't, and I've already bumped here: https://github.com/joshmoore/openmicroscopy/commit/209b9af72f3a121c946cf831415139da4c8511b9

Is it safe to go all the way to 9.4? Worth testing on breaking? Let's continuous this on gh-3519 since otherwise we'll conflict.

rleigh-dundee:
Why would it be any less safe than upgrading to 9.2?  The JDBC jar is compatible with all versions back to 7.x, so there's little reason not to use the latest jar.  It's backward compatible with all protocol versions, so by using a 9.1 or 9.2 jar, we make the 9.1 or 9.2 protocol the lowest common denominator, thereby precluding automatic use of any subsequent protocol enhancements.  That said, the last breaking change was the binary encoding stuff in 9.0/9.1 which was what we last updated the jar for; I'm not aware of any subsequent breaking changes. The main changes between 9.2 and 9.4 are a whole pile of bugfixes and performance enhancements, so it's probably worth upgrading for them since you'll get the benefit of them irrespective of the server version for the most part.

The changelog for the JDBC driver is here: https://jdbc.postgresql.org/documentation/changelog.html

PR #3525, opened 2015-02-23
Float big image

jburel:
Do not set the min max values to full range for big images of type float or double
We do not generate pyramid for big images of that pixels type
to test the PR:
 * Import ```team/will/float-data/mrc/AMI-public-data/GroEL/...ef.mrc``` team/will/float-data.zip
 * Check that the min max does not correspond is resp. ```181.491``` and ```627.354``` (was ```-2147483648``` and ```2147483647```)
 * Open the image in imageJ and check it matches the omero image.

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
@@ -655,11 +655,7 @@ public class ManagedImportRequestI extends ImportRequest implements IRequest {
         int maxPlaneSize = sizes.getMaxPlaneWidth() * sizes.getMaxPlaneHeight();
         if (((long) reader.getSizeX()
              * (long) reader.getSizeY()) > maxPlaneSize) {
-            int pixelType = reader.getPixelType();
-            if (pixelType != FormatTools.FLOAT &&
-                    pixelType != FormatTools.DOUBLE) {
-                    return null;
-            }
+            return null;
         }
         int bytesPerPixel = getBytesPerPixel(reader.getPixelType());
         MessageDigest md;
--- a/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
@@ -656,14 +656,9 @@ public class ManagedImportRequestI extends ImportRequest implements IRequest {
         if (((long) reader.getSizeX()
              * (long) reader.getSizeY()) > maxPlaneSize) {
             int pixelType = reader.getPixelType();
-            long[] minMax = FormatTools.defaultMinMax(pixelType);
-            for (int c = 0; c < reader.getSizeC(); c++) {
-                store.setChannelGlobalMinMax(
-                        c, minMax[0], minMax[1], series);
-            }
             if (pixelType != FormatTools.FLOAT &&
-                pixelType != FormatTools.DOUBLE) {
-                return null;
+                    pixelType != FormatTools.DOUBLE) {
+                    return null;
             }
         }
         int bytesPerPixel = getBytesPerPixel(reader.getPixelType());
@@ -692,7 +687,7 @@ public class ManagedImportRequestI extends ImportRequest implements IRequest {
 
 
     /**
-     * Read a plane to cause min/max valus to be calculated.
+     * Read a plane to cause min/max values to be calculated.
      *
      * @param size Sizes of the Pixels set.
      * @param z The Z-section offset to write to.
--- a/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
@@ -661,7 +661,10 @@ public class ManagedImportRequestI extends ImportRequest implements IRequest {
                 store.setChannelGlobalMinMax(
                         c, minMax[0], minMax[1], series);
             }
-            return null;
+            if (pixelType != FormatTools.FLOAT &&
+                pixelType != FormatTools.DOUBLE) {
+                return null;
+            }
         }
         int bytesPerPixel = getBytesPerPixel(reader.getPixelType());
         MessageDigest md;


dominikl:
Works like expected. Although I don't understand why, because the if check and return happens after the `store.setChannelGlobalMinMax(...)`, but I'm not familiar enough with this code.

jburel:
@dominikl:  the mix max will first be set to the full range of pixels type, it is then adjusted when the planar  values are read. For large planes, the values are not read

dominikl:
Ah ok... thanks for the explanation.
Good to merge.

components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
```diff
@@ -661,7 +661,10 @@ public MessageDigest parseData(
                 store.setChannelGlobalMinMax(
                         c, minMax[0], minMax[1], series);
             }
-            return null;
+            if (pixelType != FormatTools.FLOAT &&
```
joshmoore: I'm still not understanding this. It seems like this doesn't calculate in *fewer* cases as opposed to *more* cases. If there were an extra `return null`, say after the if, then it would skip more cases, right?

jburel: from what I gather (not my code), the global min and max are set to the min/max of the pixel type.
For large planes, the planes are not read (i.e. return null).In the case of float/double we do not generate pyramid so do not read any planes so the values are always set to min/max of pixel range.
The check added is to keep the same behaviour for other pixels types

joshmoore: Might be easier to talk it through but what I'm seeing (in pseudo-code) is:
```
  # Old
  if (big) {
    set_defaults();
    return null;
  }
  read_planes();
```
```
  # New
  if (big) {
    set_defaults();
    if (float || double) {
      return null;
    }
  }
  read_planes();
```
where I would think we wanted:
```
  # New
  if (big) {
    set_defaults();
    return null;
  } 
  if (float || double) {
    set_defaults();
    return null;
  }
  read_planes();
```
or perhaps just
```
  # New
  if (big || float || double) {
    ...
  }
```

jburel: let's discuss it. It will be easier

jburel:
Further investigation:
I tested with a gatan file (unint16) qualifying as a big image
 * During import the min max is set to the full pixels range.
 * During the pyramid generation, another stats info is created with new min/max. This stats info is then used. Note that the first stats info is never deleted when the image is deleted
For float or double, there is no pyramid generation so no new stats info created. So the modification in this PR allows to set the correct min max values.

jburel:
@dominikl:
to test the last commit
 * Import a gatan file e.g. ```biosamplefullframe.dm3```. Check that you do no have 2 entries created in Stats Info: one with full pixel range 
 * Then check the large float image

dominikl:
I can confirm that still everything works, but I couldn't see the two StatsInfo entries in the DB without the PR (tested locally with empty DB and develop branch), with and without PR there was just one entry after importing `biosamplefullframe.dm3`.
Note: Just noticed that the Preview is enabled in Insight for big float images, e. g. `team/will/float-data/mrc/AMI-public-data/GroEL/...ef.mrc`. Probably better to disable it, like it's done for int images, isn't it? Anyway, nothing to do with this PR, I can do this in another PR.

jburel:
did you try to expand dataset containing the image?  you will see the second entry w/o the PR

jburel:
@dominikl: happy for that one to be merged?

dominikl:
@jburel Yes. Sorry, this PR slipped out of my attention.

joshmoore:
Summarizing what I think is the change here (for docs & @pwalczysko):
 * we no longer create (extra) StatsInfo for big images
 * rather than skipping this method for all big images, we only skip for float/double big images.

If that's true, then is there any danger of https://github.com/openmicroscopy/openmicroscopy/pull/3525/files#diff-641e76080134ee29c700f964cf9f2501R677 consuming too much memory.

jburel:
@joshmoore: this could happen for potentially large float/double. the examples we have in the repo are single plane images around 4kx4k (EM)
This will have to be rolled back when we have pyramid support for float.
or we can introduce a tmp size limit for such image.

joshmoore:
Understood, I think there's just the potential for having an OOM for very large float/double images. Perhaps we just specify a slightly larger limit for these but make it **hard** and list even larger float/double images as "Unsupported"?

cc: @hflynn @chris-allan @pwalczysko @melissalinkert

jburel:
an option could be to use @chris-allan https://github.com/openmicroscopy/openmicroscopy/pull/3596.
no stats for float if large

joshmoore:
i.e. this method short-circuits even earlier? If so, that would prevent the possible OOM.

jburel:
we will do
```
if (((long) reader.getSizeX()
              * (long) reader.getSizeY()) > maxPlaneSize) {
return null;
         }
```
i.e. no stats info at this stage for big image

pwalczysko:
Retest acoording to the header of this PR.
Import went fine.
Full viewer of the image in Insight is fine.
Full viewer in Insight corresponds to thumb
Full Viewer in Insight corresponds to ImageJ view with the same Rendering settings.
Preview does load up and looks fine.
Thumbnail is present.


Min and Max:

In Preview it is populated as 

screen shot 2015-03-13 at 16 33 18

In Acquisition pane of Insight it is shown as 


screen shot 2015-03-13 at 16 33 27


In ImageJ after importing the image using Bio-formats and ticking the "Show metadata" box

screen shot 2015-03-13 at 16 42 35

pwalczysko:
In summary, all seems fine here (except that Insight is very slow when zomming in and out and changeing the rendering settings of this image, which is probably expected).

The only question is - are the numbers in Preveiw pane of Insight in Min and Max expected - see previous comment (screenshot) ?


screen shot 2015-03-13 at 16 33 18

jburel:
they are since we do not enter the stats. The values displayed are the full pixels range.

sbesson:
@jburel: i.e. ready to merge?

jburel:
No more changes to come. Ready to merge

jburel:
--no-rebase

PR #3529, opened 2015-02-24
Compare full generated web configs

manics:
This should make it easier to review the effects of any future config changes.

Reference configs exist for each web-server with defaults, and each web-server with all options together. Combinations of options continue to be tested by the existing tests.

### DIFF
--- a/components/tools/OmeroPy/test/unit/clitest/reference_templates/apache-fcgi-withoptions.conf
+++ b/components/tools/OmeroPy/test/unit/clitest/reference_templates/apache-fcgi-withoptions.conf
@@ -1,0 +1,78 @@
+###
+# apache config for omero
+# this file should be loaded *after* ssl.conf
+#
+# -D options to control configurations
+#  OmeroForceSSL - redirect all http requests to https
+
+###
+### Example SSL stanza for OMERO.web created 0000-00-00 00:00:00.000000
+###
+
+# Eliminate overlap warnings with the default ssl vhost
+# Requires SNI (http://wiki.apache.org/httpd/NameBasedSSLVHostsWithSNI)
+# support most later versions of mod_ssl and OSes will support it if you see
+# "You should not use name-based virtual hosts in conjunction with SSL!!"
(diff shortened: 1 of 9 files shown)


sbesson:
```
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateDefaults[nginx] PASSED
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateDefaults[nginx-development] PASSED
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateDefaults[apache] PASSED
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateDefaults[apache-fcgi] PASSED
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateWithOptions[server_type0] PASSED
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateWithOptions[server_type1] PASSED
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateWithOptions[server_type2] PASSED
     [exec] test/unit/clitest/test_web.py::TestWeb::testFullTemplateWithOptions[server_type3] PASSED
```

Looks good. Do we still need the per-line tests or is the goal to deprecate them in favor of the full templates tests?

manics:
Depends whether we want to keep testing combinations of all options (would require a lot of files, by my calculations 36 for `testNginxConfig` alone) or just a few.

sbesson:
Having full templates for each combination definitely does not scale well especially if we keep adding options in the future. I guess there is no harm in keeping both series of test. I wonder if dropping the numbered lines checks in favor of more generic `assert line in output` would simplify the maintenance of these tests since we have the full templates ones. Thoughts?

manics:
Yes, or something in between where we drop the line numbers but check the required lines appear in the expected order. I'll have a go.

manics:
@sbesson How do the last two commits look?

sbesson:
Certainly works for me. As long as Travis is happy, this looks ready to merge.

manics:
--no-rebase

PR #3530, opened 2015-02-24
Units conversion 2 (sympy)

joshmoore:
The new conversion code makes use of the http://www.sympy.org library for using a small number of symbolic equations to generate all the possible conversions between our various units. In general, this PR should be as complete as is currently possibly and not noticeably slower on conversions. Spot checking of some of the equations would be appreciated, as well as comments about both the `Conversion` API as well as the `BigResult` exception. (It currently is marked as a `RuntimeException` until @dominikl has a chance to look into it. I can either take a PR against this repo or that can be handled in a follow-up. Python does not have the same issue).

See: https://github.com/joshmoore/ome-units/commit/f983d16c55c42fa2a4dda1c9137cd0d427d5efe4

cc: @dominikl @mtbc @jburel @qidane @sbesson 
--no-rebase


components/blitz/src/omero/model/ElectricPotentialI.java
```diff
@@ -609,35 +620,28 @@ public ElectricPotentialI(double d, ome.model.enums.UnitsElectricPotential ul) {
...
-    public ElectricPotentialI(ElectricPotential value, String target) {
+    public ElectricPotentialI(ElectricPotential value, String target) throws BigResult {
        String source = value.getUnit().toString();
        if (target.equals(source)) {
            setValue(value.getValue());
            setUnit(value.getUnit());
         } else {
-            double[][] coeffs = conversions.get(source + ":" + target);
-            if (coeffs == null) {
+            Conversion conversion = conversions.get(source + ":" + target);
```
mtbc: If we have an array of many same-unit values to convert, it's a pity if we have to do this string concatenation and map lookup for every one in converting the lot, but I don't see an easy way around that and it may not be much cost compared to other construction costs anyway.

components/blitz/src/omero/model/TimeI.java
```diff
@@ -612,35 +755,28 @@ public TimeI(double d, ome.model.enums.UnitsTime ul) {
...
-
-                k = coeffs[1][0];
-                p = coeffs[1][1];
-                v += Math.pow(k, p) * orig;
-            } else {
-                throw new RuntimeException("coefficients of unknown length: " +  coeffs.length);
+            BigDecimal big = conversion.convert(orig);
+            double converted = big.doubleValue();
+            if (converted == Double.NEGATIVE_INFINITY ||
+                    converted == Double.POSITIVE_INFINITY) {
```
mtbc: any chance of a NaN?

mtbc: also `Double.isInfinite` if you prefer, fine either way

components/insight/SRC/org/openmicroscopy/shoola/util/roi/figures/MeasurePointFigure.java
```diff
@@ -207,7 +210,12 @@ public UnitPoint getMeasurementCentre()
      */
     public Length getMeasurementY()
     {
-        return new LengthI(transformY(getY()), getUnit());
+        try {
+            return new LengthI(transformY(getY()), getUnit());
+        } catch (BigResult result) {
+            // FIXME: temporarily returning infinity
```
mtbc: I'll assume this is on radar somewhere.

components/model/src/ome/model/units/Conversion.java
```diff
@@ -0,0 +1,269 @@
...
+            this.i = i;
+            this.s = null;
+        }
+
+        public Int(String s) {
+            this.s = s;
+            this.i = 0;
+        }
+
+        public BigDecimal convert(double original) {
```
mtbc: Might want to add to Javadoc that the argument's ignored.

components/model/src/ome/model/units/Conversion.java
```diff
@@ -0,0 +1,269 @@
...
+            this.denom = denom;
+            this.delay = false;
+        }
+
+        public Rat(Conversion...conversions) {
+            super(conversions);
+            this.num = 0;
+            this.denom = 0;
+            if (conversions.length != 2) {
+                throw new IllegalArgumentException("Too man conversions: " +
```
mtbc: many

components/tools/OmeroPy/src/omero/model/conversions.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class Conversion(object):
+    """
+    Base-functor like object which can be used for preparing complex
+    equations for converting from one unit to another. Primarily these
+    classes and static methods are used via code-generation. Sympy-generated
+    strings are placed directly into code. If the proper imports are in place,
+    then a top-level Conversion (usually of type Add or
+    Mul is returned from the evaluation.
```
mtbc: close parenthesis

components/tools/OmeroPy/test/unit/test_conversions.py
```diff
@@ -0,0 +1,94 @@
...
+    def testSimpleRat(self):
+        r = Rat(1, 3)
+        assert "(1 / 3)" == str(r)
+        x = r(-1)  # -1 is ignored
+        self.assertEquals(0.333333, x)
+
+    def testDelayedRat(self):
+        r = Rat(Int(1), Int(3))
+        x = r(-1)  # -1 is ignored
+        self.assertEquals(0.33333333, x)
```
mtbc: it's interesting we needed more decimal places here than in the previous test

components/tools/OmeroPy/test/unit/test_conversions.py
```diff
@@ -0,0 +1,94 @@
...
+from omero.model.conversions import Int
+from omero.model.conversions import Mul
+from omero.model.conversions import Pow
+from omero.model.conversions import Rat
+from omero.model.conversions import Sym
+
+
+class TestConversions(object):
+
+    def assertEquals(self, a, b, places=6):
```
ximenesuk: Hopefully this method will be confined to this class?

components/insight/SRC/org/openmicroscopy/shoola/util/roi/figures/MeasurePointFigure.java
joshmoore: That's for @dominikl

components/tools/OmeroPy/test/unit/test_conversions.py
joshmoore: Yup. Just a short-cut for setting `places`.

components/blitz/src/omero/model/ElectricPotentialI.java
joshmoore: All gone now.

components/blitz/src/omero/model/TimeI.java
joshmoore: Thanks.

joshmoore:
Finally tracked down what was going on in the integration tests. @will-moore : I have no idea why https://github.com/joshmoore/openmicroscopy/commit/1857fd4c41ace4adbf8420c766d394046035ff41 is now required, and so could use your interpretation/feedback.

joshmoore:
Changes from https://github.com/joshmoore/ome-units/commit/c66013ef1058c74ff607b0893667f06fbbd72ca5 have been applied now which should fix the inter-SI conversions. Angstrom has also been fixed.

jburel:
cc @mporter-gre since he might have to check for exception thrown

PR #3533, opened 2015-02-25
Fix gateway test helpers to point at downloads.openmicroscopy.org

sbesson:
This PR removes the need for a personal URL as the necessary files are made public.

To test it, check the gateway tests still pass.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/scripts/dbhelpers.py
+++ b/components/tools/OmeroPy/src/omero/gateway/scripts/dbhelpers.py
@@ -13,7 +13,7 @@ from types import StringTypes
 from path import path
 
 BASEPATH = os.path.dirname(os.path.abspath(__file__))
-TESTIMG_URL = 'http://users.openmicroscopy.org.uk/~cneves-x/'
+TESTIMG_URL = 'http://downloads.openmicroscopy.org/images/gateway_tests/'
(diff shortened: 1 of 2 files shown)


joshmoore:
Guess the next question would be if these are even necessary. Perhaps OMERO5 helps alleviate some of the need?!

sbesson:
`tinyTest.d3d.dv` is definitely in the repository (and used by other tests). Speaking to @melissalinkert, `big.tiff` was created from a fake file (meaning we could import the fake file directly). I would need to assess the content/usage of the last DV file.
In short, yes I think we can remove the need for external images in these tests. That being said, 1- I don't have the bandwidth to do it right now especally we have a Trello card for gateway testing cleanup (in the course of 5.1.x) and 2- I would like to minimally rebase this PR to `dev_5_0` so that integration tests do not depend on a personal URL in case of a 5.0.9.

joshmoore:
Understood & agreed.

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration-python/377/. Gateway tests failures are likely related to https://github.com/openmicroscopy/openmicroscopy/pull/3530 rather than this PR.

joshmoore:
Manual test of the URLs looks good. Merging.

sbesson:
--rebased-to #3540

PR #3537, opened 2015-02-26
Apache equivalent of Nginx 502 gateway down

manics:
a24923b..b8b1401: Show a nice error page if OMERO.web hasn't been started
- https://trello.com/c/mmpExksH/389-apache-502-handling
- https://github.com/openmicroscopy/openmicroscopy/pull/3326#issuecomment-70850090

140efe4: Fix error in test logic in https://github.com/openmicroscopy/openmicroscopy/pull/3529

 ab40f38 bc548e6: If `omero.web.prefix` has been set don't redirect `/` to this `prefix`. This is a change from the previous behaviour, but unifies the Apache 2.2 config with everything else and makes more sense (a prefix implies multiple sites).

Testing: Review the template changes in `components/tools/OmeroPy/test/unit/clitest/reference_templates/`
Test Apache 2.2 and 2.4 with/without prefixes, and check the following:
- OMERO.web and Apache running: Everything should work as normal
- OMERO.web stopped: Apache should show the OMERO maintenance page
- OMERO.web restarted: Everything should work again
- Generate a server exception in a web POST call: Error dialog should be displayed, and submission to QA should work
- Generate a server exception in a web GET call (easiest way is to add `assert False` into one of the omeroweb python files), Error dialog should be displayed

Note error codes seem to differ between servers
- Nginx returns error 502
- Apache 2.2/mod_fastcgi returns error 500
- Apache 2.4/mod_proxy_fcgi returns error 503

Since error 500 is also generated by Django check the Apache `ErrorDocument 500` doesn't cause problems with Django's error handler.

I haven't tested Apache 2.2 with fastcgi sockets... see also https://trello.com/c/FHk2rdFe/433-remove-non-tcp-fastcgi-apache-config for a proposal to drop it.

--no-rebase


manics:
--on-hold

joshmoore:
Haven't had a chance to spin up any of these today. The changes look like what I was thinking, though. Only thing that occurs to me is if the naming of `502.html` no longer holds, but that's not really important.

What's your shorted-list of steps to go from some image to each of these running? :smile:

manics:
I took the mual options for testing- start with the image from https://github.com/ome/omero-install/tree/master/linux/test/centos6-apache, upgrade to 5.1-m4, installed apache24 alongside (`yum install centos-release-SCL`, `yum install httpd24`), copied the test configs from `components/tools/OmeroPy/test/unit/clitest/reference_templates/` and then started/stopped `httpd`/`httpd24`.

manics:
502.html renamed.

I've noticed an existing bug with the nginx 502 page- you only see it if you're below the root (e.g. http://localhost:10080/webclient/), if you just go to http://localhost:10080 you don't see the maintenance page. This all works on Apache.

sbesson:
Would it be worth dropping the extra `$uri/` cf http://serverfault.com/a/382456? I can test using Nginx development tonight or tomorrow.

manics:
Good find, updated (works for me).

sbesson:
Tested locally using `OMERO.server-5.1.0-m5-173-76ab2e2-ice35-b429` with both development nginx and system nginx. Worked as expected.

aleksandra-tarkowska:
tested locally with apache Server version: Apache/2.2.26 (Unix) all works as expected

joshmoore:
Thanks, all. As mentioned elsewhere, :thumbsup: for getting this in, and then potentially doing a larger round of testing  (as mentioned on weekly testing card, @pwalczysko )

PR #3544, opened 2015-02-27
Don't change group on web logout

will-moore:
From: https://trello.com/c/GkdpN8wD/311-rfe-change-default-group-to-active-group-in-uis

To test:
 - Browse to different group data in web, then logout. See that when you log in again you back to your Default group and this isn't changed on logout.
 - Can also confirm this by changing default group in user settings.


dominikl:

PR #3545, opened 2015-02-27
Image viewer projection fixes

will-moore:
A couple of minor fixes, including http://trac.openmicroscopy.org.uk/ome/ticket/12758

To test:
 - Open the web image viewer. The "Normal" projection radio button should be checked.
 - For single Z-plane images, the "Max Intensity" projection should be disabled (also check it's not disabled in Z-stack images).


mtbc:
The radio buttons don't look much different when enabled or disabled, but this PR works just as described. Good to merge.

mtbc:
It's minor, but might be nice to merge this for the RC.

PR #3549, opened 2015-02-27
Carry maxmem setting through to Insight app build file.

ximenesuk:
This PR should stop builds failing due to running out of heap space. I'm not sure if there is a rigorous way to test this other than that a local default build should be successful. 

/cc @joshmoore in case he can suggest better testing.

### DIFF
--- a/components/insight/build/app.xml
+++ b/components/insight/build/app.xml
@@ -132,6 +132,7 @@
           depends="app-init"
           description="Compile the whole app.">
     <javac srcdir="${base.src.dir}" target="1.6"
+           memoryMaximumSize="${javac.maxmem}"
            source="1.6"
            destdir="${app.compiled.dir}"
            includeantruntime="no"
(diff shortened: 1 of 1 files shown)


joshmoore:
Perhaps: *setting the javac.maxmem setting to a very **low** value should cause the insight build to fail. Bumping it back up should then let the build pass.*

rleigh-dundee:
Java8u31/MacOS 10.10:

- 650m causes blitz failure
- 675m, 700m OK

OpenJDK7u75/Linux:

- 650m, 675m, 700m, 725m causes blitz failure
- 750m OK

I can't reproduce any insight build failures.  Setting too low results in a failure to build inside blitz.  The change looks reasonable though, and seems to be working as expected.

joshmoore:
Only remaining question from me was whether or not the regular insight build still worked, but since @jburel removed build.class that's no longer a worry! Merging.

ximenesuk:
I assume...

--no-rebase

PR #3543, opened 2015-02-27
Policy for download restriction

joshmoore:
See: https://trello.com/c/Y1YocYjU/323-download-button-restriction

Add support for limiting the download of original files. The basic functionality is that it is now possible to configure the server to limit access to original files, for example, preventing downloading the fileset.

## Client-side: Add Permissions.extendedRestrictions mechanism

In order to allow further canEdit/canDelete/...-like
functionality, a second field on Permissions has been
added which contains strings documented and set by
individual services at runtime. Failure to obey these
restrictions in the client will lead to an exception.

## Server-side: ome.security.policy.PolicyService

In order to support the new Permissions.extendedRestriction
client-side checks, this commit adds a server-side interface,
ome.security.policy.PolicyService, which is configured with
instances of ome.security.policy.Policy. Similar to the JVM's
SecurityManager and Permission objects, calling code can either
check if a call *would* fail, or simply let the service raise
a SecurityViolation.

Currently, the only prototypical restriction policy in place
is the DownloadPolicy.

cc: @mtbc @aleksandra-tarkowska @chris-allan @imunro 

## Testing

 * Check that with this PR merged but no special configuration that all downloads continue to work
 * Disable all downloads:
   * `bin/omero config set omero.policy.download none`
   * check that **nothing** is downloadable
   * e.g. `bin/omero upload some_file.txt && bin/omero download 32 -`
   * note: **import** will also be broken. It may be that this is a useless config.
 * Set the repository setting:
   * `bin/omero config set omero.policy.download repository`
   * check that the owner of a file can download it
   * but others in a `rwr---` or `rwra--` group can **not**

PR #3553, opened 2015-03-02
Initial sketch of the ome.security.policy package

joshmoore:
In order to support the new Permissions.extendedRestriction
client-side checks, this commit adds a server-side interface,
ome.security.policy.PolicyService, which is configured with
instances of ome.security.policy.Policy. Similar to the JVM's
SecurityManager and Permission objects, calling code can either
check if a call *would* fail, or simply let the service raise
a SecurityViolation.

Currently, the only prototypical restriction policy in place
is the DownloadPolicy.

--exclude
NOTE: This is intended for API comments only. The branch will likely be force-pushed.
cc: @mtbc @aleksandra-tarkowska

### DIFF
--- a/components/blitz/resources/omero/Constants.ice
+++ b/components/blitz/resources/omero/Constants.ice
@@ -250,6 +250,12 @@ module omero {
        **/
       const int ANNOTATERESTRICTION = 3;
 
+      /**
+       * Extended restriction name which may be applied to images and other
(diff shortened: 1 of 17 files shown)


joshmoore:
Closing in favor of https://github.com/openmicroscopy/openmicroscopy/pull/3543

PR #3558, opened 2015-03-03
CLI key session

sbesson:
Minor bug reported by @manics. Under certain conditions a local session file can be with no `omero.port=xxx` line while using the default port. When this session is re-attached via:

```
bin/omero session -s host -k key -p 4064
```

it results in a conflict between the old port (`None`) and the new port (`4064`). This PR adds a special case in `store.conflicts` to handle this particular case and implements the corresponding unit tests (which should pass).

--no-rebase


manics:
Fixes the problem I found yesterday, tests pass, good to merge.

PR #3531, opened 2015-02-25
Rendering toolbar update

aleksandra-tarkowska:
This PR cleanup stylesheets used for various toolbars. ```.toolbar``` is now in in ```ome.toolbar.css``` and can be used globally. Thanks to that rendering toolbar in right hand panel and image viewer benefits from hover, active, focus, etc styles. To test:

 - choose image, click preview tab, alter channel slider -> appropriate buttons (undo, save) should activate
 - click undo -> redu should activate, save should deactivate
 - click copy settings -> past should activate
 - choose compatible image, check if paste button is active, paste settings -> appropriate buttons (undo, save) should activate

test the same in image viewer

cc: @will-moore 

--no-rebase

PR #3560, opened 2015-03-04
Basket preview viewer fix

will-moore:
This fixes the preview panel viewer on the basket page, as reported https://github.com/openmicroscopy/openmicroscopy/pull/3531
but the bug was actually the missing infieldlabel.js javascript library on the basket page.
Also found that this lib was duplicated for search and history pages, being also present in their parent template.

To test:
 - Check that in-line labels work (the placeholder text is removed when you start to type) in right panel Comments field on basket page, search results page and history page.
 - Check that in-line labels work for top search field on basket page.
 - Check that image viewer works in Preview panel on basket page.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base.html
@@ -83,6 +83,10 @@
 
     <script>
         $(document).ready(function(){
+
+            // setup in-line labels on top search field
+            $("#top_search_field label").inFieldLabels();
+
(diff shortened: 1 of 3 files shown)


gusferguson:
@will-moore 

Tested with https://trout.openmicroscopy.org/merge/webclient/ user-3 and user-1.

Behaves as expected - all in-line labels work and image viewer works in Preview pane.
The smallest of RFEs - when next doing anything in this area - is it possible to vertically align the inline Search: label to middle - any typed text is slightly better aligned - see screenshots.
Good to merge.

safari

safari

PR #3566, opened 2015-03-04
add a "current step" to status of async server jobs

mtbc:
Addresses some of the discussion on https://trac.openmicroscopy.org/ome/ticket/10474. Ensure that https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/test.integration.test_delete/TestDelete/ passes.

--no-rebase

### DIFF
--- a/components/blitz/resources/omero/cmd/API.ice
+++ b/components/blitz/resources/omero/cmd/API.ice
@@ -42,6 +42,9 @@ module omero {
             StateList flags;
             StringMap parameters;
 
+            /** the latest step to be commenced, from 0 to steps-1 */
+            int currentStep;
+            /** the total number of steps */
             int steps;
             long startTime;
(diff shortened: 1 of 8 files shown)


joshmoore:
Looks correct. Tests are passing. The only thing that I would have liked to have done is played with https://github.com/openmicroscopy/openmicroscopy/blob/v.5.0.8/components/blitz/resources/omero/cmd/Basic.ice#L68 (with and without a `DoAll`) to see if things looked sensible. It might be that we just list that as a follow-up.

mtbc:
Makes sense. It's an awkward compromise between not wanting tests to take much longer than necessary and wanting tests not to fail if the server is slow. In practice, running locally, I found that the delete test's status queries did manage to actually hit each step, often multiple times.

PR #3567, opened 2015-03-04
Prevent property with trailing equal sign in config set

sbesson:
See https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7752#p15430

This PR should prevent an operation like
```
bin/omero config set omero.property= value
```

Apart from checking the unit tests are green in Travis, testing this PR should include running this command and testing `config.xml` remains unmodified.

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -285,10 +285,14 @@ class PrefsControl(WriteableConfigControl):
 
     @with_rw_config
     def set(self, args, config):
-        if "=" in args.KEY and args.VALUE is None:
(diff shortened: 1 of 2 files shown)


manics:
Is there a specification for allowed characters in keys? E.g. if it's limited to `[A-Za-z0-9\.]+` could you do a more thorough check?

joshmoore:
There's not really anything that's filtered. This is mostly a usability check in terms of "a=b" is *probably* not what you mean, since that means "remove the key named 'a=b'". If we added "bin/omero config unset a=b" and deprecate the one arg usage, perhaps this would get easier.

manics:
Good to merge then.

PR #3568, opened 2015-03-04
Remove fastcgi -socket from web template generation

manics:
See https://trello.com/c/FHk2rdFe/433-remove-non-tcp-fastcgi-apache-config

We never use or test FastCGI with sockets so there should be no visible effect of this PR. Just check that the generated templates are unchanged.

Note: FastCGI sockets continue to be usable with OMERO.web, but you'll have to create the web-server configuration file yourself.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/web.py
+++ b/components/tools/OmeroPy/src/omero/plugins/web.py
@@ -30,7 +30,6 @@ Configuration:
 Example Nginx developer usage:
 
     omero config set omero.web.debug true
-    omero config set omero.web.application_server fastcgi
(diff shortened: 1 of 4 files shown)


manics:
I've belatedly noticed nginx also has template generation code for fastcgi unix sockets. Should I remove it too since it's also never tested by us: https://github.com/manics/openmicroscopy/commit/6d5ae000d3947c02cbc1070b4c3b61208ae39190

/cc @chris-allan

chris-allan:
No issue with that from me.

manics:
--no-rebase

PR #3571, opened 2015-03-05
Right panel adjustments

dominikl:
This PR tackles a few problems with the metadata panel:
- Aligns the 'Located in' task pane like the other task panes (and not at the bottom) - *make sure the 'Located in' pane still works, does not stick on the bottom, and behaves like the other panes.*
- Brings back the download button in the toolbar, which I accidently removed in a previous PR - *check that the download button is there, and still works like expected*
- Rounds all values shown in Metadata panel to two decimal places (you can test the rounding for example with test_images_good/lei_martin_sp2; these images contain many metadata values with lots of decimal places) - *check that you don't see any metadata value with more than 2 decimal places*

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/GeneralPaneUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/GeneralPaneUI.java
@@ -195,10 +195,10 @@ class GeneralPaneUI
 		
 		add(annotationTaskPane, c);
 		c.gridy++;
-
-        UIUtilities.addFiller(this, c, true);
         
         add(browserTaskPane, c);
+        
+        UIUtilities.addFiller(this, c, true);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


jburel:
Logged as user-4
 * The Download option is greyed out for very_small.dv under ```03-10-13 user-4 imported by user-6>user-9 RA```.  I re-imported the image in another dataset. The option is not greyed out.

jburel:
The rest works as expected.

dominikl:
That's true, it's the case for all images in this dataset. Have to check that again tomorrow, can't log into trout with that branch at the moment.

dominikl:
@jburel The problem with the images in `03-10-13 user-4 imported by user-6>user-9 RA` is that the ImageData.isArchived() method returns false, in which case the Download option is disabled. Are these images which have been imported before 5.0 without archiv option, and therefore no original image exists for these files?

dominikl:
Review: Just last commit. Select a dv image with exposureTime/deltaT metadata series, open the channel metadata panel, make sure exposureTime/deltaT is displayed and no crash with ClassCastException thrown happens.

jburel:
@dominikl: do you have a good example?

dominikl:
I think almost all dv images in test_images_good have some exposure_time/deltaT data, but for example test11_R3D.dv certainly has (usually already imported on trout in archivedv)

jburel:
tested using ```SMN10ul03_R3D_D3D.dv``` user-3 on trout.
No crash noticed.

jburel:
Merging

PR #3574, opened 2015-03-05
Duplicate projects in tree

will-moore:
Fix for http://trac.openmicroscopy.org/ome/ticket/12771

When multiple Projects containing multiple Datasets are retrieved by projection in marshal_projects() they are initially P-D pairs,

```
p1 - d1
p1 - d2
p2 - d3
```
and are organised into hierarchies by sorting by Project name, then collapsing datasets if 2 adjacent Projects are the same. 
E.g.
```
p1 - d1
    - d2
p2 - d3
```

However, if we have many Projects with the same name, then the same Project will not be sorted beside each other and not fully collapsed.
E.g. 
```
p1 - d1
p2 - d3
p1 - d2
```
Fix is to simply sort by ID after sorting by Project name.

To test, check that multiple same-named Projects with multiple Datasets only appear a single time in the tree. See ticket above for links to existing data in trout.

Also applied the same fix to Screens and Plates, so need to test that they still load OK in tree.

Added a marshal test that fails without the bug-fix.
```
OmeroWeb wmoore$ ./setup.py test -t test/integration/test_tree.py -k test_marshal_projects_datasets_duplicates
...
>       assert len(marshaled) == len(projects_datasets)
E       assert 15 == 4
```
--no-rebase

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/tree.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/tree.py
@@ -339,7 +339,7 @@ def marshal_screens(conn, experimenter_id=None):
                left join splink.child plate
                left join plate.plateAcquisitions pa
         %s
-        order by lower(screen.name), lower(plate.name), pa.id
+        order by lower(screen.name), screen.id, lower(plate.name), pa.id
         """ % (where_clause)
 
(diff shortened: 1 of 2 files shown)


pwalczysko:
Checked both Projects and Screens on trout merge.
All works as expected.
Ready to merge.

PR #3576, opened 2015-03-05
IConfig Map methods

joshmoore:
Two methods have been added to IConfig to allow loading maps of values:
 * getConfigValues(String regex)
 * getConfigDefaults()

The first applies security settings to each value as if `IConfig.getConfigValue()` had been called individually. The second simply returns the contents of `etc/omero.properties`. There is *some* potential that a user has encoded something sensitive in this file (though they shouldn't). Shall we make it admin only?

Initially added for gh-3373 (and another PR from Ola)
cc: @dominikl @aleksandra-tarkowska 
--no-rebase

### DIFF
--- a/components/blitz/resources/omero/api/IConfig.ice
+++ b/components/blitz/resources/omero/api/IConfig.ice
@@ -24,6 +24,7 @@ module omero {
                 idempotent string getVersion() throws ServerError;
                 idempotent string getConfigValue(string key) throws ServerError;
                 idempotent omero::api::StringStringMap getConfigValues(string keyRegex) throws ServerError;
+                idempotent omero::api::StringStringMap getConfigDefaults() throws ServerError;
                 idempotent void setConfigValue(string key, string value) throws ServerError;
                 idempotent bool setConfigValueIfEquals(string key, string value, string test) throws ServerError;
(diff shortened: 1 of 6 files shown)


dominikl:
The two new methods are working like expected, thanks @joshmoore .

joshmoore:
@kennethgillen , @sbesson , @chris-allan : any thoughts on exposing `etc/omero.properties` to the world? If not, then happy to have this go in.

dominikl:
I actually just need the call `getConfigValues("omero\\.client\\..*")` being permitted for the user, everything else can be admin only.

manics:
Is it time to implement http://trac.openmicroscopy.org.uk/ome/ticket/10755#comment:5
> Moving forward, I would think bin/omero config will always be supported, across nodes, etc. One other option I've considered is making some namespace generally readable: omero_ext.*, omero.public.* or similar.

joshmoore:
Tempted to almost just add:
 * getClientConfigValues()
 * getClientConfigDefaults()

joshmoore:
@dominikl has cherry-picked these to gh-3373. Closing.

kennethgillen:
@joshmoore: no objections here.

PR #3579, opened 2015-03-06
Upgrade settings toplinks

will-moore:
This attempts to address the change in web top_links setting in config.xml on upgrade from 5.0.x to 5.1 or later.

The upgrade is working OK (see test) but the new VERSION '5.1.0' is not getting written to the config.xml so the top_links are getting extended on every startup!

### DIFF
--- a/components/tools/OmeroPy/src/omero/config.py
+++ b/components/tools/OmeroPy/src/omero/config.py
@@ -201,13 +201,15 @@ class ConfigXml(object):
                     if x.get("name") == "omero.web.ui.top_links":
                         val = x.get("value", "")
                         toplinks = json.loads(val)
-                        defaultlinks = [["Data", "webindex",
-                            {"title": "Browse Data via Projects, Tags etc"}],
+                        defaultlinks = [
+                            ["Data", "webindex",
+                                {"title":
+                                 "Browse Data via Projects, Tags etc"}],
                             ["History", "history",
-                            {"title": "History"}],
+                                {"title": "History"}],
                             ["Help", "http://help.openmicroscopy.org/",
-                            {"target": "new", "title":
-                            "Open OMERO user guide in a new tab"}]]
+                                {"target": "new", "title":
+                                    "Open OMERO user guide in a new tab"}]]
                         toplinks = defaultlinks + toplinks
                         val = json.dumps(toplinks)
                         x.set("value", val)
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/src/omero/config.py
```diff
@@ -188,9 +190,30 @@ def version(self, id=None):
...
 
+    def toplinks_check(self):
+        for k, v in self.properties(None, True):
+            version = self.version(k)
+            if version == "4.2.1" and v is not None:
+                for x in v.getchildren():
+                    if x.get("name") == "omero.web.ui.top_links":
+                        val = x.get("value", "")
+                        toplinks = json.loads(val)
+                        defaultlinks = [
```
sbesson: Should we rather use `from omeroweb import settings` and use the default value of `settings.TOP_LINKS` ?

joshmoore: no objections conceptually, just keep in mind that at this point settings.py could be in the process of importing config.py.

will-moore: Using the default value of settings.TOP_LINKS would be nice, but I think I'd have to access it manually via 
```
settings.CUSTOM_SETTINGS_MAPPINGS["omero.web.ui.top_links"][1] 
```
since ```settings.TOP_LINKS``` would give me the config.xml value that I'm trying to update!
Is it worth trying, or does @joshmoore's warning make it too risky?

will-moore:
@joshmoore Could you add "exclude" to this - until I know why the VERSION is failing to update to 5.1.0 in the config.xml? Any ideas?

joshmoore:
How about we try out Chris' recent suggestion? Close, and that way you can re-include without needing anyone to be involved.

will-moore:
@joshmoore So, it seems that XmlConfig.save() is writing the updated top_links to both the ```<properties id="__ACTIVE__">``` block AND the "default" block but the new "version number "5.1.0" is only getting written to the ```<properties id="__ACTIVE__">``` block. Then, when I follow the logic of ```version_check()``` to see if I need to update the top_links, this is checking the "default" properties for version number, not the "__ACTIVE__" properties, so it never recognises the updated version.

So, either I need  XmlConfig.save() to write "5.1.0" to both the "default" and "__ACTIVE__" blocks or I need the top_links update to always check the "__ACTIVE__" block.
I just wondered which way this is expected to work?

joshmoore:
@will-moore : I would have expected it to be rewritten in both blocks.

PR #3581, opened 2015-03-06
OMERO.matlab: map annotations

sbesson:
This PR exposes the new MapAnnotation in the OMERO.matlab toolbox

-  the `getAnnotationTypes` dictionary is extended to include maps
- a new `writeMapAnnotation` function is added to create maps from either strings or cell arrays of key/value pairs
- the `WriteData` example is extended to cover map creation/linking/retrieval

--no-rebase


sbesson:
See https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/248/console

manics:
Works as expected. Should there be a `get*MapAnnotations()` method? Or are they being deprecated to avoid too many methods?

sbesson:
@manics: yes the `getxxxyyyAnnotations()` table is to maintain (esp. help headers of all similar files need to be updated) and the table becomes unreadable. The [documentation](http://www.openmicroscopy.org/site/support/omero5.1/developers/Matlab.html#annotations) has been modified to point at the generic `getObjectAnnotations` function as the default one.

manics:
Good to merge then

PR #3585, opened 2015-03-07
Browse project

jburel:
Do not browse project with empty datasets.

see https://trac.openmicroscopy.org/ome/ticket/11952

To test
 * Select a project containing empty datasets.
 * Make sure the browse action is not enabled.


pwalczysko:
Sorry, I am not sure I understand here - does this mean I will not be able to manage my P/Ds in Insight just because the datasets happen to be empty ? How would I delete them then, or how would I fill them in ?

pwalczysko:
Aha, maybe it is just the Browse button disabled. I was horrified that I would not be able to expand a Project in the tree if Datasets are empty.

jburel:
Just the browse button not enabled if empty dataset

mtbc:
Works well, good to merge.

jburel:
--no-rebase

PR #3586, opened 2015-03-07
projection combobox

jburel:
Use normal combobox.
see https://trac.openmicroscopy.org/ome/ticket/11742
To test:
 * Select an image with several z
 * Open the full viewer.
 * Go to the projection tab.
 * Check that the options are displayed in a general combobox.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ControlPane.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ControlPane.java
@@ -526,8 +526,7 @@ class ControlPane
             names[index] = entry.getValue();
             index++;
         }
-        projectionTypesBox = EditorUtil.createComboBox(names, 0,
-                getBackground());
+        projectionTypesBox = new JComboBox(names);
         projectionTypesBox.setBackground(getBackground());
         projectionTypesBox.setToolTipText(PROJECTION_DESCRIPTION);
         projectionTypesBox.setActionCommand(""+TYPE);


rleigh-dundee:
With this PR click-drag-release now works.  However, keyboard navigation does not (focus, space to open; it auto-selects when you use the arrow keys without giving a chance to move and select the desired choice).  Why isn't this just a standard combo box?  What's so special about this one?

jburel:
it is just a combobox now

rleigh-dundee:
@jburel OK.  Is it grabbing keyboard events when it shouldn't though?

joshmoore:
@rleigh-dundee : was the behavior you are referring to also the state on latest? I.e. something that can be handled elsewhere.

rleigh-dundee:
@joshmoore Just tried to test on Windows with -latest and -merge.  Keyboard navigation is non-functional with -latest and partially functional with -merge.

In general, focus movement seems incomplete.  It's not always possible to identify the focussed widget, and not all widgets work.  For example, progressing along active toolbar icons works, but the Compression combobox does not (and it's another non-standard widget?--it's definitely different to the Intensity combobox).  Interpolate does not show focus but can be toggled.

For the Intensity combobox, using the arrow keys on it can change the setting, but
- it's not possible to drop it down with space/enter (or whatever is supposed to)
- movement has an immediate and unpredictable effect
- any change immediately switches focus to the Zoom slider (why?); why change focus at all?  This is likely the main problem.
- The "Every n'th slice" number entry also has the same focus-switching problems
- After this point in the tab Project works (but switches focus to Intensity), and after this it's not possible to tell what's focussed (not visible, and things don't have any obervable effect)

joshmoore:
All noted & appreciated, but is the summary that the behavior is better with this, and we can scope out keyboard usage later? cc: @gusferguson

rleigh-dundee:
@joshmoore Yes

joshmoore:
Awesome. Thank you. Leaving you to follow up on the need for a card with @gusferguson with @jburel out today.

jburel:
--no-rebase

PR #3589, opened 2015-03-08
Search plate acquisition

will-moore:
As in Insight, searching for Plate "Runs" should work the same as for Plates.
To test:
 - Tag a plate 'Run'
 - Search for the Tag - Run should be in results
 - Click it to display in right panel
 - Click "Browse" to browse to the run.


dominikl:
Search works like expected, results in Insight and Web are identical. Just a minor issue: The browse function opens the plate but doesn't populate the central panel and selects the run, like browse on an image within a dataset does.

joshmoore:
@will-moore : are you planning on modifying the opening here or elsewhere?

will-moore:
@joshmoore Sorry - not in this PR. It's not part of search functionality.

joshmoore:
Ok, leaving @will-moore to note that down in trello. Merging.

PR #3587, opened 2015-03-07
roi color

jburel:
Remove yellow background color around text.

Discussed during client meeting 05/03
cc @aleksandra-tarkowska 

To test: 
 * Draw few different shapes. 
 * Add some text
 * Check that there is no longer a yellow background.

PR #3590, opened 2015-03-09
ROI text background

aleksandra-tarkowska:
As a follow up https://github.com/openmicroscopy/openmicroscopy/pull/3587 web also removes yellow background behind the text.

To test:
 - draw ROIs (rectangular and elipse) with text, add Text ROI. Save
 - open the same image in web and check if text has no yellow background behind the text.
 - test if selected objects are marked with blue line overlapping border of ROI 

firefoxscreensnapz015

test with big image and see if line remain minimum half of the ROI line

--no-rebase


will-moore:
Yellow box is not shown.
I think that you need to fix the text colour, so that it matches the line colour for shapes, as Insight does.
Otherwise you have no control over how well this will show up in web. Also the highlighted line is very hard to see, as discussed (see screenshot).

screen shot 2015-03-10 at 15 48 12

aleksandra-tarkowska:
@will-moore try new changes

will-moore:
Looks good. Good to merge.

screen shot 2015-03-16 at 13 38 55

PR #3591, opened 2015-03-09
Filter MapAnnotations

dominikl:
Modifies MapAnnotation panel so that it takes the 'show all', 'added by me' and 'added by others' filter into account. Also adds a tooltip to the MapAnnotation panel (Tooltip is active on the title border around the MapAnnotations panel, which means the user's own MapAnnotation doesn't have a tooltip).

Test: Check that the "Show all", "Show added by me" and "Show added by others" filter in the Annotations panel works for MapAnnotations like it does for all other kinds of annotations. Check that the tooltip is shown when you hover of the titled border of another user's MapAnnotation.

See [Ticket 12752](https://trac.openmicroscopy.org/ome/ticket/12752)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AnnotationDataUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AnnotationDataUI.java
@@ -104,16 +104,16 @@ class AnnotationDataUI
 {
 
 	/** Indicates to display all the annotations linked to the object. */
-	private static final int				SHOW_ALL = 0;
+	public static final int				SHOW_ALL = 0;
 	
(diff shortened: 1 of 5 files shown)


pwalczysko:
The filtering works as expected.

The Tooltip: it gives a bit more than promised in the header of this PR. In fact, it is there also for your own Map Annotations. You just have to drive exactly onto the bordering line of the box which encompasses the annotation, then the tooltip pops up. For other user's annotations, the tooltip pops up when driven on "Added by..." text. 

I think this is ready to merge.

pwalczysko:
Found an issue, but unrelated to this PR. See https://trac.openmicroscopy.org/ome/ticket/12782. 
Ready to merge.

PR #3578, opened 2015-03-06
Fixing Download button

aleksandra-tarkowska:
https://github.com/openmicroscopy/openmicroscopy/pull/3543 follow up.

As long as PR #3593 is in, this should validate default settings and enable download button for image and file annotations but not allow to download plate. For custom config this can be tested as stated in #3593

**To test** go to Read-Ann group, choose user, choose own image and download. Repeat the same test in read-only and private group, try also with big image. Then test plate. Arrow button was turned on but you should see only export as, no Download.

Check if test passed
https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration-web/lastCompletedBuild/testReport/test.integration.test_download/TestDownload/test_spw_download/

--no-rebase

PR #3577, opened 2015-03-05
Advanced email configuration

joshmoore:
To follow-on from the recent mail improvements, this PR introduces a few build modifications as well as listener objects to allow for a more sysadmin-directed configuration of system notifications. One of the main benefits is that it is now possible to write **integration tests** for mail behaviors.

### Deployment changes ###
 * Support `classpath*:blitz/*.xml` such that Spring bean files can be added under etc/ and be picked up automatically.
 * Support `export OMERO_PROFILE=/opt/omero-config/` as another location for XML (Note the ending slash)

### Server-side code changes ###
 * Moved MailUtil to new `ome.services.mail` package (cc: @aleksandra-tarkowska)
 * Remove `ShapeChangeMessage` (unused)
 * Remove old spring xml files
 * Add `SqlAction.getUserEmailsByGroup()`
 * Add a `ShutdownMessage` raised in `ome.services.blitz.Entry`
 * Add `subethasmtp.jar` dependency, by default the runtime doesn't access it. Uncommenting or copying a blob from `spring-cfg/mail.xml` to some other location will start up a SMTP server.

### Functional email changes ###
 * Send emails during certain events **if activated**:
    * When a new user is added @ehrenfeu
    * When a server starts
    * When a server shuts down

### Test setup ###
 * Make a new directory for placing your config files (e.g. /tmp/cfg) and subdirectory blitz (/tmp/cfg/blitz/)
 * Set that directory as a location for loading config, `export OMERO_PROFILE=/tmp/cfg/`
 * Copy the two files from etc/blitz/ and rename:
   * `mv etc/blitz/mail-server.example /tmp/cfg/blitz`
   * `mv etc/blitz/mail-senders.example /tmp/cfg/blitz`
 * Restart the server
 * Make sure `root` has a password: `bin/omero obj update Experimenter:0 email=test@localhost`

### Testing ###
 * Run: `test/integration/test_mail.py`. Note: this is being run by the CI infrastructure as well.
 * Verify that one of the actions configured (comment on image or new user) create a MapAnnotation as expected.

--no-rebase

PR #3593, opened 2015-03-09
Restrict download (Take 2)

joshmoore:
This PR aims to build on gh-3543 by:
 * [x] allowing group settings (`ExperimenterGroup.confg`) to modify the `DownloadPolicy`
 * [x] ~~allow the `DownloadPolicy` to be driven by the size of files to be downloaded.~~

cc: @pwalczysko @aleksandra-tarkowska 

##### (A) +read,+write,+image (default) #####

 * For plates, in the UIs the Download and Download Original Metadata menus are missing. If a group enables `+plate`, then the UI menus will come back.
 * All the downloads of non-plate data should work as before.

##### (A2) +read,+write,+image,-plate #####

 * Like (A), but no override is possible for individual groups.

##### (B) -read,+write,+image,-plate #####

 * As (B) above, but download will be restricted for members of the group who:
   * are not the owners of an image (assuming a non-`rwrw--` group)
   * are not owners of the group
   * are not in the system group

##### (C) +read,+write,+image,+plate #####

 * Downloads should be **enabled** for all objects.
 * A single group can disable plate imports by setting `-plate`.

##### (D) -read,-write,-image,-plate #####

 * Downloads should be **disabled** for all objects. (Just setting `-write` is sufficient for this setting, which is equivalent to "none" from gh-3545)

#### CLI testing tip ####

With https://github.com/openmicroscopy/openmicroscopy/pull/3600, it's possible to set a particular flag on the group:
```
  # To enable plate download (regardless of server setting)
  bin/omero obj map-set ExperimenterGroup:$ID config omero.policy.binary_access +plate

  # To disable group-member download (regardless of server setting)
  bin/omero obj map-set ExperimenterGroup:$ID config omero.policy.binary_access -read

  Etc.
```
So, to test each of the scenarios above, use (in order):
 * (A) `bin/omero obj null ExperimenterGroup:$ID config`
 * (A2) `bin/omero obj map-set ExperimenterGroup:$ID config omero.policy.binary_access -plate`
 * (B) `bin/omero obj map-set ExperimenterGroup:$ID config omero.policy.binary_access -read`
 * (C) `bin/omero obj map-set ExperimenterGroup:$ID config omero.policy.binary_access +plate`
 * (D) `bin/omero obj map-set ExperimenterGroup:$ID config omero.policy.binary_access -write`

Note: *only one such value can be set at a time via `map-set`.*

### DIFF
--- a/components/model/src/ome/model/internal/NamedValue.java
+++ b/components/model/src/ome/model/internal/NamedValue.java
@@ -87,11 +87,33 @@ public class NamedValue implements Serializable, Primitive {
 
     @Override
     public int hashCode() {
-        final int prime = 31;
+        final int prime = 139;
         int result = 1;
         result = prime * result + ((name == null) ? 0 : name.hashCode());
         result = prime * result + ((value == null) ? 0 : value.hashCode());
         return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        NamedValue other = (NamedValue) obj;
+        if (name == null) {
+            if (other.name != null)
+                return false;
+        } else if (!name.equals(other.name))
+            return false;
+        if (value == null) {
+            if (other.value != null)
+                return false;
+        } else if (!value.equals(other.value))
+            return false;
+        return true;
     }
 
     @Override
--- a/components/model/src/ome/model/internal/Permissions.java
+++ b/components/model/src/ome/model/internal/Permissions.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
+import java.util.HashSet;
 import java.util.Set;
 
 import javax.persistence.Column;
@@ -158,7 +159,7 @@ public class Permissions implements Serializable {
                     "Make sure that you have not passed omero.group=-1 for a save without context");
         }
         this.revokeAll(p);
-        copyRestrictions(p.restrictions);
+        copyRestrictions(p.restrictions, p.extendedRestrictions);
     }
 
     // ~ Fields
@@ -300,6 +301,27 @@ public class Permissions implements Serializable {
         return isDisallow(restrictions, LINKRESTRICTION);
     }
 
+    public void addExtendedRestrictions(Set<String> extendedRestrictions) {
+        if (extendedRestrictions == null || extendedRestrictions.isEmpty()) {
+            return;
+        }
+
+        if (this.extendedRestrictions == null) {
+            this.extendedRestrictions = extendedRestrictions.toArray(
+                    new String[extendedRestrictions.size()]);
+        } else {
+            // Should be a much less likely case since these will
+            // likely not have been loaded/set yet.
+            Set<String> copy = new HashSet<String>();
+            for (String er : this.extendedRestrictions) {
+                copy.add(er);
+            }
+            copy.addAll(extendedRestrictions);
+            this.extendedRestrictions = copy.toArray(
+                    new String[copy.size()]);
+        }
+    }
+
     /**
      * Produce a copy of restrictions for use elsewhere.
      */
@@ -329,7 +351,16 @@ public class Permissions implements Serializable {
      * Safely copy the source array. If it is null or contains no "true" values,
      * then the restrictions field will remain null.
      */
-    public void copyRestrictions(final boolean[] source) {
+    public void copyRestrictions(final boolean[] source, String[] extendedRestrictions) {
+
+        if (extendedRestrictions == null || extendedRestrictions.length == 0) {
+            this.extendedRestrictions = null;
+        } else {
+            final int sz = extendedRestrictions.length;
+            this.extendedRestrictions = new String[sz];
+            System.arraycopy(extendedRestrictions, 0, this.extendedRestrictions, 0, sz);
+        }
+
         if (noTrues(source)) {
             this.restrictions = null;
         } else {
@@ -351,6 +382,7 @@ public class Permissions implements Serializable {
             this.extendedRestrictions = extendedRestrictions.toArray(
                     new String[extendedRestrictions.size()]);
         }
+
         if (allow == 15) { // Would be all false.
             this.restrictions = null;
             return;
--- a/components/blitz/resources/omero/Constants.ice
+++ b/components/blitz/resources/omero/Constants.ice
@@ -253,9 +253,9 @@ module omero {
       /**
        * Extended restriction name which may be applied to images and other
        * downloadable materials. This string can also be found in the
-       * ome.security.policy.DownloadPolicy class.
+       * ome.security.policy.BinaryAccessPolicy class.
        **/
-      const string DOWNLOAD = "RESTRICT-DOWNLOAD";
+      const string BINARYACCESS = "RESTRICT-BINARY-ACCESS";
     };
 
     module projection {
--- a/components/model/resources/mappings/acquisition.ome.xml
+++ b/components/model/resources/mappings/acquisition.ome.xml
@@ -105,6 +105,7 @@
 			<optional name="hasher" type="ome.model.enums.ChecksumAlgorithm"/>
 			<optional name="hash" type="string"/>
 			<optional name="mimetype" type="string"/>
+			<zeromany name="filesetEntries" type="ome.model.fs.FilesetEntry" inverse="originalFile"/>
 			<!--
 			Has a default set at the DB level of application/octet-stream.
 			Internal values in use are:
--- a/components/server/resources/ome/services/hibernate.xml
+++ b/components/server/resources/ome/services/hibernate.xml
@@ -96,7 +96,8 @@
     Scope: private
     </description>
     <constructor-arg ref="sessionHandler"/>
-    <constructor-arg ref="basicACLVoter"/>
+    <constructor-arg ref="aclVoter"/>
+    <constructor-arg ref="currentDetails"/>
   </bean>
   
   <bean id="sessionHandler" class="ome.tools.hibernate.SessionHandler">
--- a/components/server/resources/ome/services/sec-primitives.xml
+++ b/components/server/resources/ome/services/sec-primitives.xml
@@ -75,7 +75,7 @@
 
   <bean id="defaultPolicyService" class="ome.security.policy.DefaultPolicyService"/>
 
-  <bean id="RESTRICT-DOWNLOAD" class="ome.security.policy.DownloadPolicy">
+  <bean id="RESTRICT-BINARY-ACCESS" class="ome.security.policy.BinaryAccessPolicy">
     <constructor-arg>
       <set>
         <value>ome.model.annotations.FileAnnotation</value>
@@ -86,7 +86,7 @@
       </set>
     </constructor-arg>
     <constructor-arg ref="aclVoter"/>
-    <constructor-arg value="${omero.policy.download}"/>
+    <constructor-arg value="${omero.policy.binary_access}"/>
   </bean>
 
   <bean id="aclVoter" class="ome.security.CompositeACLVoter">
@@ -94,14 +94,14 @@
     <constructor-arg ref="sharingACLVoter"/>
     <constructor-arg ref="basicACLVoter"/>
   </bean>
-    
+
   <bean id="sharingACLVoter" class="ome.security.sharing.SharingACLVoter">
     <constructor-arg ref="systemTypes"/>
     <constructor-arg ref="currentDetails"/>
     <constructor-arg ref="shareStore"/>
     <constructor-arg ref="tokenHolder"/>
   </bean>
-  
+
   <bean id="basicACLVoter" class="ome.security.basic.BasicACLVoter">
     <constructor-arg ref="currentDetails"/>
     <constructor-arg ref="systemTypes"/>
@@ -111,7 +111,7 @@
     <constructor-arg ref="roles"/>
   </bean>
 
-  <bean name="sessionCache" class="ome.services.sessions.state.SessionCache">  
+  <bean name="sessionCache" class="ome.services.sessions.state.SessionCache">
     <property name="cacheManager"  ref="cacheManager"/>
     <property name="updateInterval" value="${omero.sessions.sync_force}"/><!-- ms -->
   </bean>
@@ -133,7 +133,7 @@
   <bean id="methodSecurity" class="ome.security.basic.BasicMethodSecurity">
     <property name="sessionManager" ref="sessionManager"/>
   </bean>
-  
+
   <!-- Throttling primitives -->
 
   <bean id="threadCounterFactory" class="ome.services.sessions.stats.CounterFactory">
@@ -145,7 +145,7 @@
   <bean id="sessionCounterFactory" class="ome.services.sessions.stats.CounterFactory"/>
 
   <!--
-  The following stats types 
+  The following stats types
   -->
 
   <bean id="perSessionStats" class="ome.services.sessions.stats.PerSessionStats">
--- a/components/server/src/ome/security/ACLEventListener.java
+++ b/components/server/src/ome/security/ACLEventListener.java
@@ -10,14 +10,14 @@ package ome.security;
 // Java imports
 
 // Third-party imports
+import java.util.Set;
+
 import ome.annotations.RevisionDate;
 import ome.annotations.RevisionNumber;
 import ome.conditions.SecurityViolation;
 import ome.model.IObject;
 import ome.tools.hibernate.HibernateUtils;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.hibernate.event.PostDeleteEvent;
 import org.hibernate.event.PostDeleteEventListener;
 import org.hibernate.event.PostInsertEvent;
@@ -34,6 +34,8 @@ import org.hibernate.event.PreLoadEvent;
 import org.hibernate.event.PreLoadEventListener;
 import org.hibernate.event.PreUpdateEvent;
 import org.hibernate.event.PreUpdateEventListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * responsible for intercepting all pre-INSERT, pre-UPDATE, pre-DELETE, and
@@ -103,6 +105,9 @@ public class ACLEventListener implements
             if (!aclVoter.allowLoad(event.getSession(), o.getClass(), o.getDetails(), o.getId())) {
                 aclVoter.throwLoadViolation(o);
             }
+            Set<String> restrictions = aclVoter.restrictions(o);
+            ((IObject) entity).getDetails().getPermissions()
+                .addExtendedRestrictions(restrictions);
         }
     }
 
--- a/components/server/src/ome/security/ACLVoter.java
+++ b/components/server/src/ome/security/ACLVoter.java
@@ -183,4 +183,10 @@ public interface ACLVoter {
      * @return
      */
     Set<String> restrictions(IObject object);
+
+    /**
+     * Gives the {@link ACLVoter} instance a chance to act on the {@link IObject}
+     * <em>after</em> the transaction but before finishing the AOP stack.
+     */
+    void postProcess(IObject obj);
 }
--- a/components/server/src/ome/security/ACLVoter.java
+++ b/components/server/src/ome/security/ACLVoter.java
@@ -12,7 +12,7 @@ package ome.security;
 // Third-party libraries
 
 // Application-internal dependencies
-import org.hibernate.Session;
+import java.util.Set;
 
 import ome.annotations.RevisionDate;
 import ome.annotations.RevisionNumber;
@@ -20,6 +20,8 @@ import ome.conditions.SecurityViolation;
 import ome.model.IObject;
 import ome.model.internal.Details;
 import ome.system.EventContext;
+
+import org.hibernate.Session;
 
 /**
  * helper security interface for all decisions on access control
@@ -173,4 +175,12 @@ public interface ACLVoter {
      */
     void throwDeleteViolation(IObject iObject) throws SecurityViolation;
 
+    /**
+     * Provide the active restrictions for this {@link IObject}.
+     *
+     * See {@link ome.security.policy.PolicyService for further details.
+     * @param object
+     * @return
+     */
+    Set<String> restrictions(IObject object);
 }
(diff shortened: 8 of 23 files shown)


components/model/src/ome/model/internal/NamedValue.java
```diff
@@ -95,6 +95,28 @@ public int hashCode() {
     }
 
     @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
```
mtbc: This is interesting. I was reading https://netbeans.org/bugzilla/show_bug.cgi?id=156994 and wondering if the choice here is related to `NamedValue` being mapped by Hibernate.

components/model/src/ome/model/internal/NamedValue.java
```diff
@@ -95,6 +95,28 @@ public int hashCode() {
...
+            if (other.name != null)
+                return false;
+        } else if (!name.equals(other.name))
+            return false;
+        if (value == null) {
+            if (other.value != null)
+                return false;
+        } else if (!value.equals(other.value))
+            return false;
+        return true;
```
mtbc: `StringUtils.equals` may help here.

components/model/src/ome/model/internal/Permissions.java
```diff
@@ -300,6 +301,27 @@ public boolean isDisallowLink() {
         return isDisallow(restrictions, LINKRESTRICTION);
     }
 
+    public void addExtendedRestrictions(Set<String> extendedRestrictions) {
+        if (extendedRestrictions == null || extendedRestrictions.isEmpty()) {
```
mtbc: `CollectionUtils.isEmpty` can be helpful for this kind of thing.

components/model/src/ome/model/internal/Permissions.java
```diff
@@ -300,6 +301,27 @@ public boolean isDisallowLink() {
...
+            return;
+        }
+
+        if (this.extendedRestrictions == null) {
+            this.extendedRestrictions = extendedRestrictions.toArray(
+                    new String[extendedRestrictions.size()]);
```
mtbc: Now you have me wondering if it might have been easier if the slice file generated a `List<String>` Java-side.

mtbc:
This looks like it would be a nice feature to have. :+1:

components/model/src/ome/model/internal/Permissions.java
joshmoore: It would be easier but less efficient, I think. Though now this is all getting complicated enough, that perhaps the few cycles would be worth it.

joshmoore:
Restarted travis after "romio no tests"

joshmoore:
@mtbc: please note https://github.com/joshmoore/openmicroscopy/commit/5b218f0eec3d1917ffc1f1ae94bb4026d3455bce which @aleksandra-tarkowska has said should be run past you w.r.t naming, specifically:
 * binary access v. download
 * +/- read, write, image, and plate

joshmoore:
Note: in discussion with @aleksandra-tarkowska, we pushed the concept of "restrict download by size"

mtbc:
Thank you, I appreciate it. That commit looks good to me. My main comment would just be that it's not clear to me from the comment on the configuration property how read/write interact with image/plate and how the order-dependent overriding works. It's possible that we're expressing something quite subtle and complex here that does support the expected use cases, and this may be a fine way to express it, it just may need that bit more explanation (even outwith this PR).

joshmoore:
@mtbc: Perhaps take a look at the just updated description (which can be moved to docs as needed). If that's not sufficient, it might need some slight remodeling.

mtbc:
Definitely still confusing, I'm afraid, though the examples in the PR description all follow the same ordering. @ximenesuk and I are developing theories as to what it all might mean, and perhaps I should look through the implementation and then decide if I think I understand the code and work back from that, but access control definitely isn't something we want sysadmins misconfiguring. Might be worth a chat at a moment that my head feels very clear. (I am fine with this not blocking the PR though.)

pwalczysko:
@joshmoore @mtbc : I even think I can understand this from the description in the header of the PR.

What is missing (and consequently what is unclear to me) is:

  - matching the ``tips....`` examples on the CLI with the 4 categories clearly defined above them
  - expanding on the ``map-add`` which is forthcoming (= ominous)

mtbc:
@pwalczysko: So, for instance, is it possible to configure the system to allow "read" users to download images but "write" users to download both images and plates? (And are all group members "write" users for read-write groups?)

pwalczysko:
@mtbc : Is there such a thing as a ``write-user`` ?  And ``read-user`` ? Do not think so. Do not see the header here as defining such a thing.

mtbc:
Sorry, shorthand for the documentation's "users who have READ access" or "users who have WRITE access".

joshmoore:
@mtbc : in your terms `write` == `basicAclVoter.allowUpdate()`

joshmoore:
> What is missing (and consequently what is unclear to me) is: ...

@pwalczysko : Understood. I'll will clarify this and then pass back to you.

mtbc:
@joshmoore: Main controversy here so far is if/when it makes sense to put a "read" or "write" *after* an "image" or "plate" in the list and what that would do.

pwalczysko:
@mtbc had an interesting idea: In case -read and -write are overpowering flags which ``go over everything`` then they might be want to be defined within a specific command ? Mixing all four together creates a dangerous confusion.

joshmoore:
At the moment, it would change nothing in the implementation. The default ordering currently shows how the implementation will consume, but that is not enforced. In other words, this is a `Set<String>` of options.

@pwalczysko : thoughts on the current description?

mtbc:
If it's unordered, what's the,

> earlier components which are disabled may override later ones

about?

joshmoore:
Perhaps, for etc/omero.properties, "the list above is ordered. Items like read and write are taken into account before image and plate. The value of binary_access, however, is unordered." Or shall I go all the way and use "order of evaluation" or some such?

pwalczysko:
@joshmoore : Re https://github.com/openmicroscopy/openmicroscopy/pull/3593#issuecomment-78059575 it sounds good, I understood all up to the word ``binary_access``. Then I got lost. Maybe if this word is explained or replaced, all becomes clear.

mtbc:
Perhaps something like, "If a component in the list above is disabled with a "-" prefix, that may override a "+" setting for components listed below it." My confusion was that I thought the "earlier" was to do with the ordering of components in the value of `omero.policy.binary_access` itself. It may also be worth adjusting phrasing to make it clearer that the value's components are unordered.

mtbc:
So, summarizing recent conversation: I'm fine with a "pessimistic wins" strategy, but, if the "broader" defaults must be the most permissive union of what is to be permitted anywhere more specifically, then an RFE (*not* blocking this PR) could be to allow something more restrictive to be configured to apply by default to newly created things. For instance, if I want to allow plate download from only a couple of groups, server default must include +plate but maybe new groups could automatically get a -plate before being manually adjusted.

joshmoore:
Unfortunately, I had an another idea that might do what @mtbc is requesting at the cost of *slightly* loosening the "pessimistic" rule that @pwalczysko so likes: should we allow a missing server configuration component to equal abstention? i.e. `+read,+write,+image` becomes the server default, which *still* disallows `plate`, but a group could then add `+plate` to enable. If the server setting were changed to `+read,+write,+image,-plate`, though, no group setting could overrule the server's pessimism.

mtbc:
I wonder how well that would extend to a system with more "levels" where these restrictions can also somehow annotate specific projects, datasets, images ...

joshmoore:
If you mean for the wider restriction system (i.e. beyond `BINARYACCESS`) I wouldn't worry about that too much, since it's intended to have more & varied implementations.

If you mean *just* for `canDownload` I would think we'd want to draw a clear line somewhere, likely at image, since none of the other objects you mention have binary data themselves, and the many-to-many nature will just make this impossible. For anyone who wants to implement something really special, I'd suggest they add their own `Policy` implementation and just disable our `BinaryAccessPolicy`.

pwalczysko:
@joshmoore : Re
Unfortunately, I had an another idea that might do what @mtbc is requesting at the cost of slightly loosening the "pessimistic" rule that @pwalczysko so likes: should we allow a missing server configuration component to equal abstention? i.e. +read,+write,+image becomes the server default, which still disallows plate, but a group could then add +plate to enable. If the server setting were changed to +read,+write,+image,-plate, though, no group setting could overrule the server's pessimism.

Clever. But needs really a good explanation, because slightly counterintuitive.

mtbc:
It has the issue that the admin may want the default for new groups to be tighter than `+read,+write,+image` but much depends on what actual use cases we anticipate in the longer term.

ximenesuk:
`±plate` :-)

aleksandra-tarkowska:
@joshmoore during the failed test investigation I have just realized that default config allows you to download plate even ```-plate```.

pwalczysko:
@aleksandra-tarkowska has just shown me the workflow to repeat https://github.com/openmicroscopy/openmicroscopy/pull/3593#issuecomment-78248859 - this is why I meant "lets keep it simple"...
  - go trout merge, user-4, 
  - download something else (image which is donwloadable) and
  - replace the image number in the url of the Download window for ``21107`` - this will download the mias plate, which should be forbidden

joshmoore:
> It has the issue that the admin may want the default for new groups to be tighter than +read,+write,+image but much depends on what actual use cases we anticipate in the longer term.

Really? I was thinking that if there's not an explicit `+plate`, plate download would be disabled. Isn't that the default you're talking about?

mtbc:
They may want new groups to have -read or -image or somesuch too.

joshmoore:
Then omit "+read"? Off by default, but a group can turn it on, no?

mtbc:
Ah, okay. Being able to omit them is probably also useful to mention in documentation / examples.

joshmoore:
It's not possible yet. But if the addition of "element omission" on top of "pessimism" solves everybody's wishes, then I'll push that here.

mtbc:
If it covers the motivating / likely use cases -- @pwalczysko will be clearer on those than I -- then go for it.

manics:
Reading through this it does seem a bit complicated due to mixing two different concepts (group-rw vs object-type). If it's just image and plate could you require all options to be explicitly enumerated {±readimage,±writeimage,±readplate,±writeplate}?

mtbc:
I wondered if it made sense to have two configuration properties, one for the user's access permissions to the model object, one for the kinds of model object.

manics:
That wouldn't allow `+write-plate,-read-plate`

joshmoore:
Seems like if we're going that far, we might as well do:

 * omero.policy.binary_access.plate = true/false/""
 * omero.policy.binary_access.image = true/false/""
 * omero.policy.binary_access.read = true/false/""
 * omero.policy.binary_access.write = true/false/""

i.e. just treat them as 4 different properties, which is what I was going for when the previous list.

joshmoore:
Pushed, but this doesn't take the one or more strings discussion into account.

manics:
Another consideration: Downloads of large attachments, e.g. HDF5 tables?

joshmoore:
"OriginalFile" objects are already taken into account by the policy, but we pushed the strict size limitation til post 5.1.0.

aleksandra-tarkowska:
tested default:

```
12/Mar/2015 12:22:02] "GET /webgateway/archived_files/download/?image=751&zipname=test.zip HTTP/1.1" 404 1674
```
Expected exception:
```
Cannot download file (id:None). Traceback (most recent call last): File "/Users/ola/OMERO/openmicroscopy/components/tools/OmeroWeb/omeroweb/webgateway/views.py", line 2111, in archived_files zipName = zip_archived_files(images, temp, zipName) File "/Users/ola/OMERO/openmicroscopy/dist/lib/python/omeroweb/webgateway/util.py", line 115, in zip_archived_files for chunk in a.getFileInChunks(): File "/Users/ola/OMERO/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 4658, in getFileInChunks yield store.read(0, long(size)) File "/Users/ola/OMERO/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 4119, in __call__ return self.handle_exception(e, *args, **kwargs) File "/Users/ola/OMERO/openmicroscopy/dist/lib/python/omeroweb/webclient/webclient_gateway.py", line 2069, in handle_exception e, *args, **kwargs) File "/Users/ola/OMERO/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 4116, in __call__ return self.f(*args, **kwargs) File "/Users/ola/OMERO/openmicroscopy/dist/lib/python/omero_api_RawFileStore_ice.py", line 162, in read return _M_omero.api.RawFileStore._op_read.invoke(self, ((position, length), _ctx)) SecurityViolation: exception ::omero::SecurityViolation { serverStackTrace = ome.conditions.SecurityViolation: Download is restricted for ome.model.core.OriginalFile:Id_1923 at ome.security.policy.BinaryAccessPolicy.checkRestriction(BinaryAccessPolicy.java:156) at ome.security.policy.DefaultPolicyService.checkRestriction(DefaultPolicyService.java:83) at ome.security.basic.BasicSecuritySystem.checkRestriction(BasicSecuritySystem.java:652) at ome.services.RawFileBean.read(RawFileBean.java:409) at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) at java.lang.reflect.Method.invoke(Method.java:483) at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:307) at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:183) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:150) at ome.security.basic.EventHandler.invoke(EventHandler.java:154) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at ome.tools.hibernate.SessionHandler.doStateful(SessionHandler.java:218) at ome.tools.hibernate.SessionHandler.invoke(SessionHandler.java:202) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:108) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at ome.tools.hibernate.ProxyCleanupFilter$Interceptor.invoke(ProxyCleanupFilter.java:249) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at ome.services.util.ServiceHandler.invoke(ServiceHandler.java:116) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:202) at com.sun.proxy.$Proxy94.read(Unknown Source) at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) at java.lang.reflect.Method.invoke(Method.java:483) at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:307) at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:183) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:150) at ome.security.basic.BasicSecurityWiring.invoke(BasicSecurityWiring.java:98) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at ome.services.blitz.fire.AopContextInitializer.invoke(AopContextInitializer.java:43) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:202) at com.sun.proxy.$Proxy94.read(Unknown Source) at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) at java.lang.reflect.Method.invoke(Method.java:483) at ome.services.blitz.util.IceMethodInvoker.invoke(IceMethodInvoker.java:179) at ome.services.throttling.Callback.run(Callback.java:56) at ome.services.throttling.InThreadThrottlingStrategy.callInvokerOnRawArgs(InThreadThrottlingStrategy.java:56) at ome.services.blitz.impl.AbstractAmdServant.callInvokerOnRawArgs(AbstractAmdServant.java:149) at ome.services.blitz.impl.RawFileStoreI.read_async(RawFileStoreI.java:83) at ome.services.blitz.repo.RepoRawFileStoreI.read_async(RepoRawFileStoreI.java:112) at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) at java.lang.reflect.Method.invoke(Method.java:483) at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:307) at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:183) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:150) at omero.cmd.CallContext.invoke(CallContext.java:78) at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172) at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:202) at com.sun.proxy.$Proxy95.read_async(Unknown Source) at omero.api._RawFileStoreTie.read_async(_RawFileStoreTie.java:78) at omero.api._RawFileStoreDisp.___read(_RawFileStoreDisp.java:249) at omero.api._RawFileStoreDisp.__dispatch(_RawFileStoreDisp.java:420) at IceInternal.Incoming.invoke(Incoming.java:222) at Ice.ConnectionI.invokeAll(ConnectionI.java:2482) at Ice.ConnectionI.dispatch(ConnectionI.java:1258) at Ice.ConnectionI.message(ConnectionI.java:1213) at IceInternal.ThreadPool.run(ThreadPool.java:321) at IceInternal.ThreadPool.access$300(ThreadPool.java:12) at IceInternal.ThreadPool$EventHandlerThread.run(ThreadPool.java:693) at java.lang.Thread.run(Thread.java:745) serverExceptionClass = ome.conditions.SecurityViolation message = Download is restricted for ome.model.core.OriginalFile:Id_1923 } 
```

```
[12/Mar/2015 12:22:47] "GET /webgateway/archived_files/download/?well=115&zipname=test.zip HTTP/1.1" 404 1673
[12/Mar/2015 12:22:18] "GET /webgateway/archived_files/download/?image=1&zipname=test.zip HTTP/1.1" 200 9589204
[12/Mar/2015 12:26:30] "GET /webgateway/archived_files/download/5/ HTTP/1.1" 200 3147776
```
:+1:

aleksandra-tarkowska:
Now what is not working:
 - it is possible to stream plate with default settings

pwalczysko:
The trick which Ola showed me yesterday about fooling the server into downloading a plate ends up with 404 now both locally and on trout, so this seems to be fine, but https://github.com/openmicroscopy/openmicroscopy/pull/3593#issuecomment-78478044 - I think a call between @joshmoore and @aleksandra-tarkowska is needed.
The default behaviour seems to be fine.

I guess I am not able to execute the ``...bin/omero obj map-set ExperimenterGroup:$ID omero.policy.binary_access +plate`` command, getting output

```
usage: map-set OBJ FIELD KEY VALUE
```

aleksandra-tarkowska:
@pwalczysko, @joshmoore is going to prepare test for that and makes a decision if this will or not be fixed in 5.1.0. We will keep posting you.

joshmoore:
> usage: map-set OBJ FIELD KEY VALUE

My apologies, @pwalczysko . The commands in the description were missing a critical `... config ...`. Now added.

snoopycrimecop:
Conflicting PR. Removed from build [OME-5.1-merge-push#846](https://ci.openmicroscopy.org/job/OME-5.1-merge-push/846/). See the [console output](https://ci.openmicroscopy.org/job/OME-5.1-merge-push/846/consoleText) for more details.
Possible conflicts:
  - PR #3577 joshmoore 'Advanced email configuration'
    - components/tests/python/library/__init__.py
  - PR #3578 aleksandra-tarkowska 'Fixing Download button'
    - components/tests/python/library/__init__.py

joshmoore:
Last commit fixes the download issues locally that several have seen elsewhere (cc: @pwalczysko)
```
test/integration/test_download.py::TestDownload::test_spw_download PASSED
test/integration/test_download.py::TestDownload::test_orphaned_image_direct_download PASSED
test/integration/test_download.py::TestDownload::test_orphaned_image_download PASSED
test/integration/test_download.py::TestDownload::test_image_in_dataset_download PASSED
test/integration/test_download.py::TestDownload::test_image_in_dataset_in_project_download PASSED
test/integration/test_download.py::TestDownload::test_well_download PASSED
test/integration/test_download.py::TestDownload::test_attachement_download PASSED

============================================================================ 7 passed in 280.71 seconds =============================================================================
```

aleksandra-tarkowska:
tested locally, plate is now throwing an exception as well. Good to be merged

pwalczysko:
Tested on trout (just the default behaviour and the downloads problem). All as expected.
Relying on @aleksandra-tarkowska for the other tests.
Ready to merge.

sbesson:
Thanks all. Merging.

aleksandra-tarkowska:
Thanks all, uff

aleksandra-tarkowska:
I've been trying to put it to the action and is not clear to me how should I restrict download of original file but still allow downloading file annotations in read only group for non owner.

-read,+write,+image,-plate restrict both

Should that be +read,+write,-image,-plate?

Any help welcome

PR #3594, opened 2015-03-09
RFC: Deprecate IUpdate.deleteObject API method

mtbc:
Deprecates `IUpdate.deleteObject` arising from discussion on https://trac.openmicroscopy.org/ome/ticket/2683. Definitely make sure that Insight object-delete options still work: this isn't the button for deleting images, datasets, etc., but instead subtler uses, like in cut/unlink or in changing / removing shapes in ROIs. In checking deletion, ensure that *only* the correct object(s) was deleted.

--no-rebase

### DIFF
--- a/components/blitz/resources/omero/api/IUpdate.ice
+++ b/components/blitz/resources/omero/api/IUpdate.ice
@@ -28,6 +28,7 @@ module omero {
                 void saveArray(IObjectList graph) throws ServerError;
                 IObjectList saveAndReturnArray(IObjectList graph) throws ServerError;
                 omero::sys::LongList saveAndReturnIds(IObjectList graph) throws ServerError;
+                ["deprecated:use omero::cmd::Delete2 instead"]
                 void deleteObject(omero::model::IObject row) throws ServerError;
                 idempotent void indexObject(omero::model::IObject row) throws ServerError;
             };
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
@@ -35,6 +35,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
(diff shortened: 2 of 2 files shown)


mtbc:
Travis hasn't noticed yet.

dominikl:
Didn't notice any problems. Unlink and deleting shapes/rois also work like expected. Good to merge.

sbesson:
Is this PR related to the deletion issue in https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/253/console ?

mtbc:
I can't rule it out. Does that code use the Java gateway? What *is* it, can I see the code and run it? The job doesn't get a mention at http://www.openmicroscopy.org/site/support/contributing-staging/ci-omero.html. Anything interesting in the server log at the moment the test failed? Can chat when you're in.

sbesson:
Sure. Minimally the code is in the training example (https://github.com/openmicroscopy/openmicroscopy/blob/develop/examples/Training/matlab/DeleteData.m) and is using https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroM/src/delete/deleteObjects.m. Will check the server logs.

mtbc:
If it's using `Delete` requests directly in a `DoAll` then I don't think this PR would have affected it.

mtbc:
(Oh, it's using stuff I have locally in `examples/Training/`, interesting! I think I've hardly ever looked at that.)

mtbc:
There might be a useful clue in the details of the response to the DoAll, of course, whether it was OK or ERR. Server-side, this PR just marks an IUpdate method as deprecated.

joshmoore:
Sorry, I might have misread this thread. Is anything else being looked into here?

mtbc:
I'm not aware of any problems with it, but happy to relist if some aspect requires further review. I think @sbesson's exception turned out to be something else, perhaps one of the usual transient mysteries, but he can confirm.

sbesson:
Seconding @mtbc's comment, the delete error reported by the training jobs turned out to be non-reproducible i.e. unrelated to this PR.

joshmoore:
:+1: Thanks all.

PR #3596, opened 2015-03-09
No stats ui

jburel:
Fix some UI issues when no stats info for the image
To test
 * Import an image using the CLI see https://github.com/openmicroscopy/openmicroscopy/pull/3580
 * Go to preview in insight:
   *  check that channel slider is correctly displayed
   * Check that the min/max is not enabled.

Note that Web still has the min/max enabled, if click on min/max, it defaults to Full range.
is it the behaviour we want? @gusferguson @will-moore @pwalczysko


jburel:
Discussed with @pwalczysko and @gusferguson: we will keep the button active to keep both clients in synch.Min/max action will behave like Full range. 
cc @will-moore 

one more commit to come

dominikl:
Imported a bunch of dv images, once with "--no_stats_info" and once without. Channel sliders were displayed correctly, min/max and fullrange actions worked like expected (min/max behaves like fullrange in the case of "--no_stats_info" import). Good to merge.

jburel:
--no-rebase

PR #3597, opened 2015-03-09
Download dir names

will-moore:
See http://trac.openmicroscopy.org/ome/ticket/12776

When downloading multiple files with duplicate names, use a simple index to name sub-directories.
To test:
 - import the same image 3 times
 - Select the 3 images -> Download
 - Download the created zip and unzip.
 - Files from one image should be in the root directory, with two other directories /1/ and /2/ containing the files for the other images.


pwalczysko:
Works as expected.

PR #3598, opened 2015-03-09
Clean up mask rendering

chris-allan:
This PR cleans up mask rendering by adding a new `PlaneDefWithMasks` subclass of `PlaneDef` and deprecating the unused, ugly and broken `setOverlays()` method on the rendering engine. This was discussed as the best way forward for the 5.1.0 milestone with @jburel and @joshmoore.

There should be no affect on the clients as none of them were using this logic. Intentions are to expand and make use of this API in 5.1.1 or later. Getting the API changes in for 5.1.0 was paramount. For the moment the old code stands as it is, parts of it are fixed by #3580, and some new example code has been added that works against Mask shapes matching the previous rendering criteria associated with the current image.

Review should be performed by @jburel and @joshmoore.

/cc @emilroz

### DIFF
--- a/components/blitz/resources/omero/ROMIO.ice
+++ b/components/blitz/resources/omero/ROMIO.ice
@@ -11,6 +11,7 @@
 
 #include <omero/ServerErrors.ice>
 #include <Ice/BuiltinSequences.ice>
+#include <omero/Collections.ice>
 
 module omero
 {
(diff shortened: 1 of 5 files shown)


sbesson:
Discussed with @joshmoore and @jburel this morning. This cleans code with has been broken for years although testing it properly requires a specific context. Merging this for 5.1.0-m5

sbesson:
--no-rebase

PR #3601, opened 2015-03-10
last-minute database adjustments for 5.1

mtbc:
Adds indices for various columns and makes key-value maps all TEXT. Also, for upgrades causes existing plate runs to be indexed. See https://trello.com/c/zOzqLMBR/440-last-minute-db-upgrade.


mtbc:
Needs a breaking label, maybe also exclude depending on what else is in breaking.

joshmoore:
I'd imagine you'd be safe to run with this early if you wanted, which would also test https://github.com/openmicroscopy/openmicroscopy/pull/3598, say, during the meeting. cc: @sbesson

mtbc:
Passed upgrade job and deployment job just fine, and pretty sure the integration test job failure is unrelated.

joshmoore:
i.e. remove breaking?

mtbc:
I don't see why not. :smiley:

joshmoore:
Done.

sbesson:
NB: once this PR is merged https://github.com/openmicroscopy/ome-documentation/blob/develop/omero/sysadmins/server-upgrade.txt#L173 will need to be modifed accordingly.
Arguably we could work towards auto-generate this block by reading `omero.db.version` and `omero.db.patch` from the server. However I am not sure how to automatically retrieve the version/patch of the previous release server (use omego? /cc @manics).

manics:
`omego` reads the current version from the `dbpatch` table. If we're not sticking with patch=0 for releases I don't think there's any way to know.

sbesson:
Thanks for the heads up.

mtbc:
So, ideas for testing seem to include,

* Check that search now works for plate runs imported before today.
* Check that in Insight map annotation names and values accept strings longer than 255 characters.

mtbc:
It does look like trout did the reindex of the runs this morning,
```
2015-03-11 04:56:44,344 INFO  [   ome.services.fulltext.FullTextIndexer] (2-thread-2) INDEXED  847 objects in batch#1      [ 175294 ms.]  ~99% done (3099561 of 3099563)
```
(As of now there are 849 runs.)

pwalczysko:
@joshmoore @jburel @mtbc : It seems like the Run does not have a name stored in the database. But when you change some metadata (add description and Save) then the name DOES get saved. In this way, the original names of Runs, like ``Run 902`` will not be searchable just after import, but after you have changed the name or description the ``Run 902`` becomes a name in the DB and becomes searchable. 
@dominikl is verifying this hypothesis.

dominikl:
@pwalczysko Can confirm this, PlateAcquisitions (runs) don't have a name after import ("Run xyz" seems to be a kind of virtual name made up by Insight/Web), but once you enter/modify a description, the name is written to the DB (and picked up by the search indexer).
I think this should be handled at the Plate import level. But actually this isn't related to this PR.

pwalczysko:
The runs are indexed and searchable if something changes on them ( see comment https://github.com/openmicroscopy/openmicroscopy/pull/3601#issuecomment-78244223).
Even very long Key and Values strings are accepted in Insight. In Web, we have an issue https://trac.openmicroscopy.org/ome/ticket/12783 (strictly UI-side).

Other than the weird Run naming/indexing https://github.com/openmicroscopy/openmicroscopy/pull/3601#issuecomment-78247062 all works fine here.

joshmoore:
I can confirm that these are in fact "synthetic", i.e. not in the database. So, other than changing how they appear in the UI (italics?), not much can be done about these:

```
OMERO-5.1-merge-deploy=# select name, count(*) from plateacquisition group by name;
   name    | count
-----------+-------
           |   846
 Run 224 x |     1
 Run 903   |     1
 findme    |     1
 Run 902   |     1
(5 rows)
```

joshmoore:
If we assume the plate run name issue will be handled elsewhere, I think the only issue I'm worried about is @manics omego assumptions. Does that need discussion or just possibly a code fix later?

manics:
`omego` should just work since it just orders the list of files. MY comment was w.r.t. @sbesson's question about autogenerating docs.

joshmoore:
Ah, excellent. Not nearly as worrisome as a I thought. Thanks all. Merging.

mtbc:
@sbesson: Thank you for the suggestion: opened https://github.com/openmicroscopy/ome-documentation/pull/1136 accordingly.

PR #3603, opened 2015-03-10
RenderingEngine: lookupTable getters/setters

joshmoore:
These methods will properly set the lookupTable
field on ChannelBinding. For the moment, that is
the extent of implementation for 5.1.0. A next
step will be to teach the RenderingEngine how to
make use of these properties.

### DIFF
--- a/components/blitz/resources/omero/api/RenderingEngine.ice
+++ b/components/blitz/resources/omero/api/RenderingEngine.ice
@@ -59,6 +59,8 @@ module omero {
                 idempotent Ice::IntSeq getRGBA(int w) throws ServerError;
                 idempotent void setActive(int w, bool active) throws ServerError;
                 idempotent bool isActive(int w) throws ServerError;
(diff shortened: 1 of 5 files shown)


joshmoore:
Passing test pushed, but the value isn't being persisted. :/ cc: @jburel

jburel:
string will indicate a known look-up, Do we want to handle the case, where we pass a file, i.e. tmp upload of the file and use its value. Probably too much for now

joshmoore:
Definitely. I think even having the value stored in the DB is a plus at this point. Certainly don't intended to try to interpret that string at all. Any ideas on why it's not being saved? If not (and it's important), I'll dig later.

jburel:
@joshmoore: I will have a look

will-moore:
Looks good.

joshmoore:
Thanks, guys. Merging and we can fix the save in a follow-up if we decide it's worth it.

PR #3606, opened 2015-03-11
Examples units

will-moore:
Updated a couple of Python code examples for how to work with units, getting and setting pixel sizes.

To test, read examples, then run (edit image Ids and login credentials as needed)
```
$ cd examples/Training/python
$ python Read_Data.py
$ python Create_Image.py
```

### DIFF
--- a/examples/Training/python/Create_Image.py
+++ b/examples/Training/python/Create_Image.py
@@ -54,7 +54,7 @@ i = conn.createImageFromNumpySeq(
     planeGen(), "numpy image", sizeZ, sizeC, sizeT, description=desc,
     dataset=None)
 
-print "Created new image: ", i.getId(), i.getName()
+print  'Created new Image:%s Name:"%s"' % (i.getId(), i.getName())
 
 
 # Set the pixel size using units (new in 5.1.0)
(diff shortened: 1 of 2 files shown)


jburel:
I suspect the Java one will have to be adjusted too. @bramalingam  do you want to have a look?

bramalingam:
Sure. I can.. Please let me know what needs to be done.

examples/Training/python/Create_Image.py
```diff
@@ -53,6 +54,23 @@ def planeGen():
     planeGen(), "numpy image", sizeZ, sizeC, sizeT, description=desc,
     dataset=None)
 
+print "Created new image: ", i.getId(), i.getName()
```
ximenesuk: Since this is training material a nicer output would be good here, something more like
`Created new Image:6  Name:"CFPNEAT01_R3D.dv"`
as per the other script.

ximenesuk:
Other than the minor comment above both scripts run okay and changes look fine.

joshmoore:
```
./examples/Training/python/Create_Image.py:57:6: E271 multiple spaces after keyword
```

joshmoore:
Pushed a fix.

ximenesuk:
Looks good to go.

PR #3610, opened 2015-03-12
Import no pixels checksum

sbesson:
Follow-up of https://github.com/openmicroscopy/openmicroscopy/pull/3580. In the case of large multi-dimensional datasets, the plane checksum calculation performed in the `pixelData` step of import can be computationally intensive. This PR re-uses the the `checksumAlgorithm` set in the `ImportConfig` to skip this checksum calculation if `File-Size-64` is selected further reducing the in-place import of large datasets while skipping all optional steps.

In terms of client exposition, a new  `--skip` argument is added to the Python CLI import plugin and serves as a shortcut to the various advanced imported options allowing to speed up import. The only performance argument not included in this option is `--transfer` which may vary depending on the infrastructure (`ln` vs `ln_s`). In terms of testing, from this branch, the following import command was run on a local server against a 200GB SPIM dataset:

```
bin/omero import --skip=all /Volumes/ome/data_repo/from_CarlZeissImaging/ -- --transfer=ln_s
...
2015-03-12 12:04:58,894 1352476    [l.Client-0] INFO      ome.formats.importer.cli.ErrorHandler - Number of errors: 0

==> Summary
6 files uploaded, 1 fileset created, 4 images imported, 0 errors in 0:13:20.211
```

Once import was completed, images were browsed via  client and thumbnails were generated in < 1min (with no other process running)

Using this new semantics, the `ITest.missing_pyramid()` logic is modified to use a large fake file import with `--skip=all` set instead of the previous implementations. This dramatically speeds up all tests depending on missing pyramids (`test_rawpixelsstore.py` and `test_thumbs.py`) so that their `long_running` markers can be removed. These tests should be run as part of the daily integration job and are expected to be green.

Comments/questions:
- is the naming scheme sensible for the CLI?
- do we also want to expose `--transfer` in the regular help of `bin/omero import`?
- longer term, we may want to deprecate/remove the plane checksum computation at import time as this step is obsolete since 5.0.

--no-rebase


mtbc:
I like the idea of getting rid of the `pixels.sha1` column altogether.

joshmoore:
Re-ran travis due HDF5 error.

snoopycrimecop:
Conflicting PR. Removed from build [OME-5.1-merge-push#846](https://ci.openmicroscopy.org/job/OME-5.1-merge-push/846/). See the [console output](https://ci.openmicroscopy.org/job/OME-5.1-merge-push/846/consoleText) for more details.
Possible conflicts:
  - PR #3525 jburel 'Float big image'
    - components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java

snoopycrimecop:
Conflicting PR. Removed from build [OME-5.1-merge-push#847](https://ci.openmicroscopy.org/job/OME-5.1-merge-push/847/). See the [console output](https://ci.openmicroscopy.org/job/OME-5.1-merge-push/847/consoleText) for more details.
Possible conflicts:
  - PR #3525 jburel 'Float big image'
    - components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java

mtbc:
The code changes do look reasonable.

components/tools/OmeroPy/src/omero/plugins/import.py
```diff
@@ -255,15 +234,53 @@ def importer(self, args):
...
+
+        xargs = [logback, "-Xmx1024M", "-cp", os.pathsep.join(classpath)]
+
+        # Here we permit passing ---file=some_output_file in order to
+        # facilitate the omero.util.import_candidates.as_dictionary
+        # call. This may not always be necessary.
+        out = args.file
+        err = args.errs
+
+        if out:
```
chris-allan: `close()` these at some point? Defensively include a `try / finally`?

chris-allan:
Apart from some silly stylistic things mentioned above and discussed with @sbesson this all looks completely reasonable. :+1:

sbesson:
Last commits should further refactor `import.py` to close file handles within try/finally block (plus additional stylistic changes).
Additionally, `ITest.import_images()` is modified to support a `skip` optional argument, `ITest.missing_pyramid()` is rewritten to use `import_image(..., skip='all')` and the `long_running` markers are dropped from all Python tests using `missing_pyramid()` /cc @joshmoore @ximenesuk

sbesson:
Based on https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-breaking-integration-python/232/testReport/, removing `breaking` label.

ximenesuk:
Everything looks good. Although I didn't have a local 200GB image to play with a 200MB one was enough to show improvements, especially combined with in-place imports. The previously long running tests sped by!

Regarding the help. The `skip` option relies on the underlying checksum amongst other things. That means either exposing the `checksum` option in the main help - which would mean exposing other options too in all likelihood. Or alternatively, `skip` could be pushed back to the advanced help where it would sit naturally under the `Import speed` section. As it stands `skip` seems to be the only effectively advanced option exposed at the top level.

sbesson:
@ximenesuk: agreed the high-level exposure of the speed options requires additional work and scoping. My main rationale is that `advanced help` is really hidden for most of the users and complex to use. Here I am in favor of migrating a subset of these advanced options to a top-level `Import speed/performance` Python argument group which could minimally include `--skip` and `--transfer` arguments.

With regard to checksum vs skip options, I would still be inclined to expose a binary choice at the high-level. The advanced help would then allow the user to have more granularity on the choice of algorithm rather than an on/off choice (`SHA1/File-Size-64`).

Finally, the most unsatisfying piece of code in my PR is definitely the following line:
```
 if checksumAlgorithm.equals("File-Size-64")
```
Rather than binding this check to a particular algorithm, we might consider migrating this to the mapper class with a method like `ChecksumAlgorithmMapper.isFast(checksumAlgorithm)`. Thoughts @mtbc?

mtbc:
It's certainly a meritorious idea, but fastness is a sliding scale; people might think it means "sub-cryptographic" or something. A clearer distinction might be a name that distinguishes algorithms that actually look at the (whole?) file contents from ones that don't. (Calling File-Size-64 a "checksum" algorithm is generous in the extreme!)

sbesson:
Definitely! `isFast()` was rather for the example than a real suggestion. In the context of this PR, I concurr we want to classify computationally expensive algorithms. An issue here is that we are partly misusing the file-level `checksum_algorithm` option.

A third (more thorough) option would be to properly define at the Java level a new ``--no-plane-checksum`` advanced option, which would be used by the `PixelData` step and map this flag as well as the ``--checksum_algorithm`` choice (for the file upload only) to the top-level ``--skip={all,checksum}``.

sbesson:
Last set of commits implement a low-level `--no_pixels_checksum` option for disabling the pixels checksum computation as well as documentation of new Java options in the importer advanced-help. The high-level `--skip=checksum` behavior is unchanged and disables both file and pixels checksum computation.

components/tools/OmeroPy/src/omero/plugins/import.py
```diff
@@ -176,55 +176,39 @@ def _configure(self, parser):
...
+            self.command_args.extend(["-p", client.getProperty("omero.port")])
+            self.command_args.extend(["-k", client.getSessionId()])
+
+    def set_skip_arguments(self, args):
+        """Set the arguments to skip steps during import"""
+        if args.skip in ['all', 'checksum']:
+            self.command_args.append("--no_pixels_checksum")
+            self.command_args.append("--checksum_algorithm=File-Size-64")
+        if args.skip in ['all', 'thumbnails']:
+            self.command_args.append("--no_thumbnails")
```
ximenesuk: Is the `--no_thumbnails` option not exposed in the help? Or am I just missing it?

ximenesuk:
I notice that underscores are used in the new options. I guess this is for consistency but might this be a good point to move to hyphens in `import.py` especially with new options.

components/tools/OmeroPy/src/omero/plugins/import.py
sbesson: It is in the regular Java help. Arguably this could be migrated to the `Import speed` section

ximenesuk: I would be in favour of that for consistency.

ximenesuk:
temporary comment

--a-marker

ximenesuk:
This all looks okay. Will the Mac-only `test_import.py`be fixed in this PR?

sbesson:
@ximenesuk: if the changes above make sense, I am tempted to implement them in a separate PRs. All  comments should be captured in see https://trello.com/c/l5vppgeI/325-cli-import-features.

ximenesuk:
@sbesson that sounds reasonable. I don't know what the timescale is but having to move to hyphens in 5.1 would make some sense given it is breaking. But please go ahead and merge this PR (if only so I can rebase a local branch!)

joshmoore:
Ok. Leaving you guys to sort args & such in follow-on PRs.

PR #3611, opened 2015-03-12
Store omero.group in training_ice.config file

sbesson:
See https://trello.com/c/q6dC4UYU/450-fix-c-jobs-to-create-a-new-user

The training_user user and content created by this script should be testable multiple times by the CI job. If the event context is modified, the example suite fails early. Adding `omero.group` to the ice.config file ensures the training example scripts use the correct context and should increase the stability of the training job (and prevent race conditions).

To test this PR, re-run [OMERO-5.1-merge-training](http://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/) jobs AFTER the OMERO-5.1-merge-cpp-cmake job. Errors due to incorrect context like [this build](https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/254/console) should not appear anymore.

--no-rebase


joshmoore:
Certainly makes sense to me. Based on https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/257/, merging.

PR #3612, opened 2015-03-12
build: Use long and short git revisions

rleigh-dundee:
The git version used by various git commands in `components/antlib/resources/version.xml` was a short 6-digit form of the commit hash.  This is too short to use without risk of clashes.  This PR adds a long form of the hash and uses that instead; a short revision copy of the hash is retained for user display where needed.

I also noticed we weren't saving the UNIX release date for cmake, so that's also fixed here.

See also: https://github.com/openmicroscopy/bioformats/pull/1664/files

--no-rebase

/cc @sbesson 

--------

Testing:

Check the content of `components/antlib/resources/gitversion.xml` `components/tools/OmeroCpp/cmake/GitVersion.cmake`, `components/bioformats/ant/gitversion.xml` and `components/bioformats/cpp/cmake/GitVersion.cmake`; these should have short and long git versions.

Check the content of `components/tools/OmeroCpp/cmake/GitVersion.cmake`; this should have both a string and numeric release date.

Check `ant version-info` in the source tree and unpacked zip.

Check the `cmake` output for the source tree and unpacked zip.  This should have two lines in the first few lines of the output ("Configuring openmicroscopy...") containing the version, commit hash and release date.  It should be identical for both cases.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -465,10 +465,13 @@ To get started using Eclipse, execute "./build.py build-dev" and import the top-
             <arg value="openmicroscopy"/>
             <arg value="${omero.shortversion}"/>
             <arg value="${omero.plainversion}"/>
+            <arg value="${omero.vcs.shortrevision}"/>
             <arg value="${omero.vcs.revision}"/>
             <arg value="${omero.vcs.date}"/>
+            <arg value="${omero.vcs.date_unix}"/>
             <arg value="${release.shortversion}"/>
             <arg value="${release.version}"/>
+            <arg value="${vcs.shortrevision}"/>
             <arg value="${vcs.revision}"/>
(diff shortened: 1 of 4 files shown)


components/bioformats
```diff
@@ -1 +1 @@
-Subproject commit b47a1fb154552609183b42d8e2f7724f99ebe3f3
+Subproject commit bf7865d3ff4679a9522819b3d480e94319fe7b7c
```
sbesson: Can you revert this submodule bump and mark a dependency in the description instead?

rleigh-dundee: Definitely, this was unintentional and I'll revert it.

rleigh-dundee: Done.

PR #3615, opened 2015-03-13
Remove version and patch prompts from `omero db script`

manics:
- `omero db script`: only prompts for the root password
- `omero db script --password ome`: Shouldn't prompt for anything
- `omero db script --version DBVERSION --patch DBPATCH`: Set db.version and db.patch for developers
- `omero db script DBVERSION DBPATCH`: Old deprecated behaviour (positional args)
- All other existing flags should continue to work

https://trello.com/c/8PPlaBmP/456-remove-compulsory-version-patch-args-from-db-script
--no-rebase

### DIFF
--- a/components/tools/OmeroPy/test/unit/clitest/test_db.py
+++ b/components/tools/OmeroPy/test/unit/clitest/test_db.py
@@ -123,12 +123,12 @@ class TestDatabase(object):
 
     @pytest.mark.parametrize('file_arg', ['', '-f', '--file'])
     @pytest.mark.parametrize('no_salt', ['', '--no-salt'])
-    @pytest.mark.parametrize(
-        'pos_input', ["", "%(version)s", "%(version)s %(patch)s",
-                      "%(version)s %(patch)s ome"])
-    def testScript(self, pos_input, no_salt, file_arg):
-        args = "db script "
-        args += pos_input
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/test/unit/clitest/test_db.py
```diff
@@ -136,14 +136,19 @@ def testScript(self, pos_input, no_salt, file_arg):
             output = self.file
         else:
             output = self.script_file
-        if "version" not in pos_input or "patch" not in pos_input:
-            self.expectVersion(self.data["version"])
-            self.expectPatch(self.data["patch"])
```
sbesson: I suppose the two methods above are now obsolete and can be removed from the test class.

sbesson:
Apart from my nitpicky suggestion, code changes look good and all unit tests are passing.

components/tools/OmeroPy/test/unit/clitest/test_db.py
manics: Removed

sbesson:
Tested with ``OMERO.server-5.1.0-m5-503-5bc33e5-ice35-b444``. All worked as expected. Unit tests are green and expand the coverage. Thanks @manics .

PR #3624, opened 2015-03-16
Preserve map ann table layout with long strings

will-moore:
From https://trac.openmicroscopy.org/ome/ticket/12783

This fixes the layout of the map annotation table with long Key Value strings.

To test, enter long text strings into the Key and Value fields.
Text that is too big for the table will simply be hidden, and the Key / Value table cells will remain at 50% width of the table.

screen shot 2015-03-16 at 15 50 29

--no-rebase


pwalczysko:
Works as expected.

PR #3625, opened 2015-03-16
BinaryAccessPolicy: expand AssertionFailure check

joshmoore:
populate_roi.py was also triggering a LazyInitializationException,
so the Hibernate initialization (which throws an AssertionFailure)
has been refactored out into the method `has()`.

cc: @sbesson


joshmoore:
Note: this is still logging an ERROR in Blitz-0.log, but this is happening in Hibernate *before* we get a chance to catch the exception. Happy to have this merged into m5, and I'll open a separate cleanup PR with logging changes.

PR #3626, opened 2015-03-17
Log null statsInfo at 'debug' instead of 'error'

will-moore:
Since it's no longer an error to have no stats info (see https://github.com/openmicroscopy/openmicroscopy/pull/3580) we should log this merely at 'debug' level instead of 'error'.

To test:
 - Import image with no stats
 - Open in web - observe Web log for lack of errors.

--no-rebase.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -6626,7 +6626,7 @@ class _ChannelWrapper (BlitzObjectWrapper):
         """
         si = self._obj.getStatsInfo()
         if si is None:
-            logger.error("getStatsInfo() is null. See #9695")
+            logger.debug("getStatsInfo() is null. See #9695")
             try:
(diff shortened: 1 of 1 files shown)


sbesson:
Setting `omero.web.debug` to `True`, I can confirm the `StatsInfo` message is displayed at DEBUG level instead of ERROR level with this PR included.
With regard to the impact of this change, I would consider this logging message important enough to promote it to the `INFO` level instead. With the exposition of `--skip` parameters, users may run into more of these issues and we may want to have this information logged by default in production instances. /cc @jburel

jburel:
probably better to have it at ```INFO``` level as @sbesson suggested. not really a DEBUG

sbesson:
```
2015-03-23 08:32:24,506  INFO [                           omero.gateway] (proc.30074) getWindowMin:6629 getStatsInfo() is null. See #9695
2015-03-23 08:32:24,507  INFO [                           omero.gateway] (proc.30074) getWindowMax:6650 getStatsInfo() is null. See #9695
````
Thanks

PR #3602, opened 2015-03-10
Upgrade settings toplinks

will-moore:
This was previously opened as #3579.

To test:
 - BEFORE building with this PR, set top_links to have a single link, E.g.
```
$ bin/omero config set omero.web.ui.top_links '[["Test", "www.ome.org"]]'
```
 - Webclient should only show a single 'Test' link in header.
 - Now upgrade / build with this PR and start OMERO.
 - Should see "Data", "History", "Help", "Test" links at top of webclient.
 - restart the server again and check that more links are not added!
 - Open dist/etc/grid/config.xml and check for 2 version blocks that state "5.1.0" (under 2 different ```<properties>``` blocks)

```
<property name="omero.config.version" value="5.1.0" />
```

Tests should also pass:
```
$ cd OmeroPy
$ ./setup.py test -t test/unit/test_config.py
```

--no-rebase

PR #3627, opened 2015-03-17
Config big image zoom

dominikl:
This PR renames the property `omero.web.viewer.initial_zoom_level` to the more general `omero.client.viewer.initial_zoom_level` and modifies Insight to make use of it.

Test with local server, parallel with Insight and Web:
- Import a big (pyramid) image
- View image in fullviewer. Expected behaviour: Zoom level is set automatically to a reasonable value.
- Set the zoom level property to zero with `./omero config set omero.client.viewer.initial_zoom_level 0` and restart the server
- Open the image in fullviewer; the initial zoom level should be zoomed out as far as possible (make sure it's the same for Web and Insight)
- Change the property with `./omero config set omero.client.viewer.initial_zoom_level 5` and restart the server
- Open the image again in fullviewer; now the initial zoom level should be 100% (again make sure Web and Insight show the same). - This should be also checked on Linux (see @mtbc 's comment about insanely huge full image viewer window)
- Optionally: Check a few other zoom levels in-between.

See [Ticket 11859](https://trac.openmicroscopy.org/ome/ticket/11859)

I'm not quite sure what other implications the change of the property name `omero.web.viewer.initial_zoom_level` will have, e. g. if someone already set this property deliberately and updates the server and then `omero.client.viewer.initial_zoom_level` (with default = 0) is used instead. How do we handle this, just a note in the documentation? 
/cc @will-moore @aleksandra-tarkowska 

--no-rebase

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerModel.java
@@ -88,6 +88,7 @@ import org.openmicroscopy.shoola.env.rnd.RndProxyDef;
 import org.openmicroscopy.shoola.env.rnd.data.Region;
 import org.openmicroscopy.shoola.env.rnd.data.ResolutionLevel;
 import org.openmicroscopy.shoola.env.rnd.data.Tile;
+import org.openmicroscopy.shoola.util.CommonsLangUtils;
 import org.openmicroscopy.shoola.util.file.modulo.ModuloInfo;
 import org.openmicroscopy.shoola.util.image.geom.Factory;
 import org.openmicroscopy.shoola.util.ui.UIUtilities;
@@ -333,19 +334,26 @@ class ImViewerModel
      */
     private int getDefaultResolutionLevel()
     {
-    	//Determine the level according to the window size.
-    	Dimension screenSize = Toolkit.getDefaultToolkit().getScreenSize();
-    	int w = 9*(screenSize.width/10);
-    	int h = 8*(screenSize.height/10);
-    	ResolutionLevel level;
-    	Dimension d;
-    	for (int i = resolutions.size() - 1; i >= 0; i--) {
-    		level = resolutions.get(i);
-    		d = level.getImageSize();
-    		if (d.width < w || d.height < h)
-    			return level.getLevel();
-    	}
-    	return 0;
+        String zoomLevel = (String) ImViewerAgent.getRegistry().lookup(
+                LookupNames.BIGIMAGE_INITIAL_ZOOM);
+        if (isBigImage() && CommonsLangUtils.isNotBlank(zoomLevel)) {
+            //Use the default zoom level from the properties
+            return Integer.parseInt(zoomLevel);
+        }
+        
+        // Determine the level according to the window size.
+        Dimension screenSize = Toolkit.getDefaultToolkit().getScreenSize();
+        int w = 9 * (screenSize.width / 10);
+        int h = 8 * (screenSize.height / 10);
+        ResolutionLevel level;
+        Dimension d;
+        for (int i = resolutions.size() - 1; i >= 0; i--) {
+            level = resolutions.get(i);
+            d = level.getImageSize();
+            if (d.width < w || d.height < h)
+                return level.getLevel();
+        }
+        return 0;
     }
     
 	/**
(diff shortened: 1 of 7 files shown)


components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -579,7 +579,7 @@ def leave_none_unset_int(s):
           " 'webtest/webclient_plugins/center_plugin.overlay.js.html',"
           " 'channel_overlay_panel']``. "
           "The javascript loads data into ``$('#div_id')``.")],
-    "omero.web.viewer.initial_zoom_level":
+    "omero.client.viewer.initial_zoom_level":
```
aleksandra-tarkowska: Please deprecate the old settings

joshmoore: These brings us back to @manics question about auto-upgrading...

will-moore: Does it help to upgrade etc/grid/config.xml as I've done in https://github.com/openmicroscopy/openmicroscopy/pull/3602 ?

jburel:
removing exclude label

dominikl:
I'm still not really sure about this deprecated config mapping, respectively the implications for this setting on a server upgrade. Unfortunately couldn't find a clue in #3602 either.

components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerModel.java
```diff
@@ -333,19 +334,26 @@
...
+                LookupNames.BIGIMAGE_INITIAL_ZOOM);
+        if (isBigImage() && CommonsLangUtils.isNotBlank(zoomLevel)) {
+            //Use the default zoom level from the properties
+            return Integer.parseInt(zoomLevel);
+        }
+        
+        // Determine the level according to the window size.
+        Dimension screenSize = Toolkit.getDefaultToolkit().getScreenSize();
+        int w = 9 * (screenSize.width / 10);
+        int h = 8 * (screenSize.height / 10);
```
mtbc: I am guessing these 9, 8 make sense.

mtbc:
Without having changed any settings or resized anything, initial view of image looks a bit strange:
full-view
The bird's eye view is as large as the initial view of the image!

mtbc:
Woah, with initial_zoom_level 5 it opened a full viewer window for which `xprop` tells me,
```
program specified size: 8111 by 1512
```
That's insanely large. Perhaps more work needed here?

dominikl:
Interesting, this insanely huge window just appears on Linux; but although on Windows it's size is OK, it's slightly offset on screen, just Mac looks OK. 
With respect to bird's eye view. This can happen, if the image is zoomed out as far as possible. Same happens in OMERO.web, but it doesn't look as drastic because the bird's eye view is smaller than in Insight. What should happen in that case? Ignore the set default zoom level and calculate on which fit to screen/image size?

mtbc:
(I think I'll have to defer to the UX team on that one, mostly I just hope for a default view that doesn't look surprisingly weird!)

dominikl:
Yes, I'll probably change to default zoom level if nothing is specified to something in between instead of 0 ( = completely zoomed out).

dominikl:
FYI: The windows size for big images was always set insanely huge, i. e. OSX and Windows seem to have some built-in protection/limit for that, so it just became obvious on Linux. But it's fixed now.

The problem "Birds eye view can be same size or larger than actual image when completely zoomed out (level 0)" still persists. This is true for Insight and Web. I think this should be tackled in another PR; with this PR Insight now behaves exactly like Web with respect to the initial default zoom level.

rleigh-dundee:
Retested on Linux.  The behaviour WRT the zoom levels is OK.  However, I do see some UI glitches which are probably unrelated to the PR specficially, but probably need fixing:

With `initial_zoom_level 0`, initial view has the image aligned on the top left:
big-insight-z0-initial-view

But after zooming in and back out again, it's centred:
big-insight-z0-after-zoom-in-out

With `initial_zoom_level 5` (or 2), the initial view has the bird eye view offset from the bottom right:
big-insight-z5-initial-view

But after zooming out and back in again, it gets placed in the bottom right corner:
big-insight-z5-after-zoom-out-in

components/tools/OmeroWeb/omeroweb/settings.py
joshmoore: Probably should have a quick discussion regarding this before merging.

aleksandra-tarkowska: @will-moore you cannot read from ```confix.xml```. It should go though service

aleksandra-tarkowska: @dominikl the best if you remove that change from here and I will open another PR once that is merge

joshmoore:
Closing in favor of 3711

PR #3630, opened 2015-03-17
CLI import improvements

sbesson:
A set of RFEs related to https://trello.com/c/l5vppgeI/325-cli-import-features.

Most important changes:
- another Java low-level flag `-no-upgrade-check` is added to disable the Bio-Formats upgrade check. The corresponding Python argument mapping is `--skip=upgrade` or `--skip=all`
- the options introduced are modified to use preferentially hyphens rather than underscore /cc @manics @ximenesuk. Python arguments are modified to support backwards compatibility where possible. New arguments introduced on the `develop` line are converted to use hyphens. This include `--no-stats-info`, `--no-pixels-checksum`, `--qa-baseurl` /cc @joshmoore @jburel @pwalczysko 
- the debug level set in the command is propagated to other loggers which were displayed at `DEBUG` level, e.g. `UpgradeCheck` or `CurrentPlatform`. This should fix failing `testDebugArguments` tests on OS X /cc @ximenesuk 
- add support for multiple `--skip` arguments, e.g.

 ```
bin/omero import test.fake --skip all
bin/omero import test.fake --skip checksum --skip minmax --skip thumbnails --skip minmax
bin/omero import test.fake --skip thumbnails --skip minmax
```

With regard to hyphens, the remaining question are 1- whether these changes are acceptable and do not modify existing workflow, 2- whether this strategy should be applied more extensively at the Java level? This could be done by moving underscore arguments under the deprecated options like `plate_name`.
--no-rebase

### CODE REVIEW

File: components/blitz/src/ome/formats/importer/ImportConfig.java [hunk 1]
```diff
@@ -329,7 +331,9 @@ public void configureDebug(String levelString) {
              level = Level.toLevel(levelString);
          }
          ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("ome.formats")).setLevel(level);
-         ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("loci")).setLevel(level);
+         ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("ome.services.blitz")).setLevel(level);
+         ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("ome.system")).setLevel(level);
+		          ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("loci")).setLevel(level);
```
mtbc: odd indent change

components/tools/OmeroPy/test/integration/clitest/test_import.py
```diff
@@ -287,7 +286,7 @@ def testDebugArgument(self, tmpdir, capfd, level, prefix):
         self.cli.invoke(self.args, strict=True)
         o, e = capfd.readouterr()
         levels = self.parse_debug_levels(o)
-        assert set(levels) <= set(debug_levels[debug_levels.index(level):])
+        assert set(levels) <= set(debug_levels[debug_levels.index(level):]), o
```
manics: Is the `, o` intentional?

sbesson: Yes. In case the assert fails this displays the full failing lines rather than the debug level only.

sbesson:
Last commit should fix  https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/398/testReport/test.integration.test_reimport/TestReimportArchivedFiles/testConvertSynthetic/ and https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-web/280/testReport/test.integration.test_marshal/TestImgDetail/test_image_detail/

components/tools/OmeroPy/test/integration/clitest/test_import.py
ximenesuk: It does look odd though! maybe if `o, e` were `out, err` or `stdout, stderr` it might be clearer?

manics:
As brought up elsewhere, the test files are now executable.

sbesson:
Last commits should improve readability of assert and fix test files permissions

sbesson:
@ximenesuk: the PR overall adds support for arguments of type `--skip OPT1 --skip OPT2`

ximenesuk:
Existing comments aside this looks good.

joshmoore:
My few cents on this was that I'd be for keeping support for underscores but SUPPRESS them in all help. I'll ping @chris-allan about any breaking impact for his usage.

sbesson:
Agreed for keeping support for 5.0.x options like `--checksum_algorithim` via suppress/deprecated options. This will come in a follow-up PR. My main question here is about options introduced as part of `develop` and never released like `--qa_baseurl` or `--no_stats_info`. Should we support both forms or move straight to hyphen-based forms?

ximenesuk:
So both `--skip minmax` and `--skip checksum` result in black thumbnails and images with the min/max, not surprisingly for the former, set to 0..65535 This was for a DV image. I didn't realise the checksum option had the same effect but that makes sense if the file isn't being parsed. Is there any way to force the min/max calculation post-import?

sbesson:
Ah thanks @ximenesuk. Yes there seems to be a collision between the pixels checksum disabling and the `store.populateMinMax()` call. I would classify this as a bug and investigate in a follow-up PR (and write proper integration tests for the `skip` semantics.

ximenesuk:
@will-moore 
```
cd dist
sh ../examples/Training/training_setup.sh
ICE_CONFIG=/path/to/training_ice.config python /path/to/__main__.py
```
worked well for me. Obviously to use real training users and images you would have to edit the ice config file by hand.

sbesson:
Thanks Colin, while investigating the `checksum/minmax` issue I realized the `--no-stats-info` option does not completely work as expected since the min/max values were still calculated (.via `openBytes` in `parseDataByPlane`). Commit f0eb514 should explain the last modification, namely:
- `--no-pixels-checksum` is fully dropped - there is no rationale for maintaining it outside `--no-stats-info`especially since both operations use the same code block and are correlated
- `--no-stats-info` now skips `ManageRepositoryRequestI.parseData()` which means both the min/max calculation and the pixels checksum calculation.
Corresponding options are propagated to the top-level CLI import. So the skip options should work as follows:

```
bin/omero import --skip checksum  # No file checksumming
bin/omero import --skip minmax  # No min/max calculation (incl. no pixels checksum calculation)
bin/omero import --skip thumbnails  # Not thumbnail calcultion
```

Integration tests to follow to cover this semantics.

sbesson:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/ has 4 failures for these new tests due to the `UpgradeCheck` test which seems too stringent. However, the previous test checking the logic skipping the statsinfo is passing. This means the last set of modifications should be functional and ready to test. I will remove the `UpgradeCheck` assert for future investigations.

ximenesuk:
See https://github.com/openmicroscopy/ome-documentation/pull/1152#discussion-diff-26831257 If we are going to expose `--skip minmax` at a user level then we must provide a way for min/max to be calculated post-import either manually or via PixelData.

sbesson:
Thanks. This point definitely needs to be clarified both in the code but most importantly in the docs. I am trying to clarify my thoughts on the issue so that we all agree on the semantics before pushing anything:
- for some large datasets (HCS or SPIM typically), calculating minmax is so computationally expensive that it needs to be explicitly skipped at import. This is not the default option and (power) users using this option will need to be aware of all its implications, hence my related doc PR will need to be completely explicit about this point /cc @hflynn 
- for planes imported with no min/max calculation, the default will be to use the full pixels range
- for now, users need to adjust the rendering settings manually if they skip min/max calculation
- with regard to min/max  re-calculation, at the moment, we do not have any API in place for doing this. This may happen for large images since pixel values are parsed during pyramid calculation - we should be able to assess that
- however, I think exposing min/max calculation for users needs to be scoped. For the same exact reasons that min/max calculation impacted (or prevented) import, triggering such operations server-side can easily destabilize a running server, e.g. a user importing ~100 SPIM with `--skip minmax` and then launching a min/max recalculation on all of them.

If we want to provide solutions, I think we need to define clearly the policy under which such re-calculation can happen and work an API addition for re-calculating min/max  values if appropriate. This could make it into a patch release of 5.1.x.  A conservative approach would be to make this functionality admin-only. /cc @joshmoore @jburel @chris-allan

ximenesuk:
Agree all round. Just need to decide what we do now for 5.1 and how it can be improved post 5.1.

pwalczysko:
``--no-stats-info`` was tested now on trout merge. Imported okay, image okay, viewing and thumb okay.
Rendering manipulation okay.

sbesson:
Assuming everyone is happy. LGTM for 5.1.0-m5 /cc @joshmoore?

ximenesuk:
Yep from me.

joshmoore:
Happy for m5. Merging. Tidbits for later:

 * Whitespace in https://github.com/openmicroscopy/openmicroscopy/pull/3630/files#diff-7fc1dbb8d465f135057e39c51a65e72cR412 is off :smile:
 * CLI/QA submissions probably need a round of testing after these changes.
 * Flex which also has special INI handling might also need looking at.

sbesson:
Whitespace will be fixed as part of a follow-up hyphenated options PR.

joshmoore:
Other two points added to the weekly testing card.

PR #3631, opened 2015-03-18
Disable graph and intensity pane for bigimages without stats

dominikl:
Workaround for [Ticket 12790](https://trac.openmicroscopy.org/ome/ticket/12790)

Problem: Graph and Intensity view pane should actually be disabled for big images, but float images aren't recognized as big images (PixelService.requiresPixelPyramid() always returns false for float images -> which on the client side means, it's not a big image). And with #3525 float big images don't have a statsinfo anymore, i. e. global min/max of channels can't be determined, which raised the exception in the ticket above, because this is needed for the graph and intensity view pane. This PR disables the graph and intensity view pane in these cases.
It's more a workaround, because the underlining cause (float images aren't recognized as 'big images') is not tackled here. 

@jburel Your thoughts, use this workaround or add some kind of isBigImage() method to Pixelservice, to determine if an image exceeds max plane size, regardless of pixel pyramid is required or not?

Actually, further thinking about it, with #3580 it's possible to also import small images without stats generation with `--no_stats_info` (even this doesn't make sense, I guess), which also would raise the exception in Ticket 12790, so regardless of the big image problem, this PR would make sense to merge anyway.

**Test:**
- Import a big float image, open the image in full viewer, open roi/measurement tool. No crash should happen, there shouldn't be a graph and intensity view pane.
- Repeat the same for other kinds of non-big images and make sure the Graph and Intensity pane are there in these cases.

Sample image for testing: mrc/AMI-public-data/GroEL/05may19a_00013gr_00008sq_00002hl_00002ef.mrc in  team/will/float-data.zip

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerUI.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.measurement.view.MeasurementViewerUI 
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


pwalczysko:
Although the PR IS included, the crash still happens on user-4 image ID 25557

dominikl:
Sorry @pwalczysko , looks like I pushed a wrong commit before. Now it's the right one and it should work.

joshmoore:
To do this "right", do we need another method like `requiresPyramids` which is `pyramidValid` or whatever? Should it throw an exception? Just thinking of what state we've left other users in.

dominikl:
I don't really know, what to do on the server side, but I guess we shouldn't/can't do anything about that for 5.1.0-m5. The `requiresPixelPyramid` method is still valid, it has just been slightly "misused" in Insight to also determine if an image is a big image. The only implications I've seen so far was the problem with the graphs/intensity pane which should be fixed with this PR, and that the preview tab is available for big float images (which shouldn't be the case for big images). Can't see any other problems arising from that (yet).

pwalczysko:
Tested for 
  - float big image ID 25557
  - an imported dv with --no-stats-info
  - images with stats which should have the Graph pane


All works as expected.

joshmoore:
Thanks for the explanation, @dominikl. If you guys do need another boolean method to differentiate, say the word.

jburel:
@dominikl: for float we should not use ```requiredPyramid`` but we could the ```isLarge``` method.
The only limitation is that the ```isLarge``` is arbitrary and not configured server side.

dominikl:
Ah, I didn't know that there is an `isLarge` method (guess it's `ImViewerModel.isLargePlane()`). Ok, I'll make use of this method in a follow-up PR, so that big float images behave like all big images (e. g. with respect to the preview panel).

PR #3370, opened 2015-01-21
Store last CLI session port

sbesson:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12609

This PR mostly adds a session file storing the last port used while connecting with the CLI (without modifying the file storage logic). The `SessionStore.set_current()/SessionStore.get_current()` logic is modified to store/return this value and the corresponding unit tests are modified to document these new inputs.

To test this PR, try to connect to the same server on different ports, e.g.
```
$ bin/omero login user@localhost:14064
$ bin/omero logout
$ bin/omero login
```

The last call should output `Previously logged in to localhost:14064 as user-1` before prompting the user for connection information.

PR #3632, opened 2015-03-18
Fix CLI sessions file

sbesson:
Reported by @manics. The addition of last port in https://github.com/openmicroscopy/openmicroscopy/pull/3370 broke `bin/omero sessions file`.

This PR should fix this bug and add an integration test to extend our coverage.

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/script.py
+++ b/components/tools/OmeroPy/src/omero/plugins/script.py
@@ -717,7 +717,7 @@ http://stackoverflow.com/questions/3471461/raw-input-and-timeout/3911560
         else:
             client = self.ctx.conn(args)
             store = SessionsStore()
-            srv, usr, uuid = store.get_current()
+            srv, usr, uuid, port = store.get_current()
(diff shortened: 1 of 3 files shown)


manics:
`$ git grep get_current`
Looks like this will need changing too:
`components/tools/OmeroPy/src/omero/plugins/script.py:            srv, usr, uuid = store.get_current()`

sbesson:
Doh. Thanks again Simon.

manics:
Good to merge

joshmoore:
Thanks, guys. Merging. Have a vague feeling that moving forward we might want to move away from multiple returns in favor a single return object.

sbesson:
Definitely. all other places in the CLI use a safer strategy.

```
current =  store.get_current()
host = current[0]
```
I will make a card for this and try to implement it (and extend the stub integration test) for 5.1.0

joshmoore:
Thanks, @sbesson. Not too particularly worried about `get_current()` specifically, but just our API-design strategy. (This one is clearly my fault, regardless)

PR #3629, opened 2015-03-17
static files versioning

aleksandra-tarkowska:
Ticket https://trac.openmicroscopy.org.uk/ome/ticket/10437. This PR adds:

 - version numbers to all 3rd party JavaScripts and CSS stylesheets, for example: ```http://localhost:4080/static/3rdparty/panojs-2.0.0/PanoJS.js```

 - suffix those omero specific with omero version, for example: ```<script type="text/javascript" src="{% static "webgateway/js/ome.popup.js"|add:url_suffix %}"></script>``` creates ```http://localhost:4080/static/webgateway/js/ome.popup.js?_5.1.0-m6-ice35-SNAPSHOT```

To test:
 - search for all keywords = names of the 3rd party libraries had name changed that all links are updated. For list of keywords refer to name of the commits below *adding version number to 3rdparty/***KEYWORD**
 - test if there are no HTTP404 when loading css and js
   - image viewer
   - data manager
   - metadata panel (including: edit name, annotation forms, batch annotation)
   - tags
   - public/shares
   - basket
   - activity
   - webadmin (including: sort users, piechart, creation)
 - test examples https://github.com/aleksandra-tarkowska/openmicroscopy/tree/js_versioning/examples/Web

@will-moore @pwalczysko 

--no-rebase

PR #3633, opened 2015-03-18
session cookie domain

aleksandra-tarkowska:
1. This should help with side effect reported in https://trello.com/c/aRDoP7dC/355-omero-web-startup-review-default-ip. To test:
 - add
 ```
bin/omero config set omero.web.session_cookie_domain "localhost"
 ```
 - open private windows in your webbrowser
 - go to localhost:4080, and log in. All shoudl be as normal
 - if you have another domain configured on you machine use it, otherwise go to 127.0.0.1:4080 and log in. you should be back to your login page without any error message. that means you cookie wasn't save.

 if you remove ```omero.web.session_cookie_domain``` you should be able to log in from all possible domains assigned to the same machine

1.  This should also resolve issue reported in https://github.com/openmicroscopy/openmicroscopy/pull/3629#issuecomment-82874726. To test:
 - remove all cookies created for your local domain in your webbrowser
 - add
 ```
bin/omero config set omero.web.session_cookie_name "myowncookiename"
 ```
 - check if you log in to web cookie name has changed

cc @sbesson 

--no-rebase

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -396,6 +396,11 @@ CUSTOM_SETTINGS_MAPPINGS = {
          86400,
          int,
          "The age of session cookies, in seconds."],
+    "omero.web.session_cookie_domain":
+        ["SESSION_COOKIE_DOMAIN",
+         None,
+         leave_none_unset,
+         "The domain to use for session cookies"],
     "omero.web.logdir":
         ["LOGDIR", LOGDIR, str, "A path to the custom log directory."],
 
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -401,6 +401,11 @@ CUSTOM_SETTINGS_MAPPINGS = {
          None,
          leave_none_unset,
          "The domain to use for session cookies"],
+    "omero.web.session_cookie_name":
+        ["SESSION_COOKIE_NAME",
+         None,
+         leave_none_unset,
+         "The name to use for session cookies"],
     "omero.web.logdir":
         ["LOGDIR", LOGDIR, str, "A path to the custom log directory."],
 


sbesson:
Talking to @aleksandra-tarkowska, the proposal is to have unique cookie_name per deployment on the CI server using something like:

```
bin/omero config set omero.web.session_cookie_name config_$JOB_NAME
```

Is that correct

manics:
https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-SESSION_COOKIE_DOMAIN
Sounds like it's still tied to the top-level domain?

manics:
Also W.R.T. the cross-domain proxy scenario, the MITM proxy could still rewrite the session cookie.

aleksandra-tarkowska:
@manics @sbesson refered to 2. Unique cookie name should resolve issue @pwalczysko faced today. We do not have any better environment to test it then CI. It doesn't heart to have that configured anyway.

pwalczysko:
Not sure about all the implications here, but the issue with the multiple windows/tabs logging into merge and latest is gone (thanks).

manics:
`SESSION_COOKIE_NAME` works as expected on trout (simultaneous logins on latest and merge).

`SESSION_COOKIE_DOMAIN`: I guess there's no harm having this, but I don't think it's particularly useful if the your main aim is to guard against MITM attacks (since it's easy to bypass).

pwalczysko:
@manics This PR is mainly useful for me I guess (possibly I am wrong) - in testing situation, it was a real pain not to be able to login simultaneously to merge and latest.

manics:
@pwalczysko Yes, there's two changes in here though, `SESSION_COOKIE_NAME` fixes your problem.

pwalczysko:
@manics : Ah, sorry, missed this, thanks for clarifying.

sbesson:
@manics: so should we merge this or should we have the `SESSION_COOKIE_DOMAIN` fixed here or somewhere else?

aleksandra-tarkowska:
@manics but that is particular useful if you would allow a cookie to be shared across

manics:
There's probably no harm in having it.

PR #3331, opened 2015-01-12
Prefix application_server.port when calling `omero admin ports`

manics:
https://trello.com/c/ROD6R0L7/180-omero-web-custom-prefix

`omero admin ports --prefix N` now prefixes `omero.web.application_server.port` (`N4080`) in addition to the standard server ports by modifying `config.xml`. It also sets the port in `omero.web.server_list`.

Note: Updated with unit tests from @sbesson 

--no-rebase

PR #3634, opened 2015-03-18
Include webserver in `admin ports --help` example

manics:
The help message should've been updated as part of https://github.com/openmicroscopy/openmicroscopy/pull/3331

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -327,12 +327,12 @@ will be prevented. Use --skipcheck to change the ports anyway.
 
 Examples:
 
-    %(prog)s --prefix=1                             # sets ports to: 14061, \
-14063, 14064
-    %(prog)s --prefix=1 --revert                    # sets ports back to: \
-4061, 4063, 4064
-    %(prog)s --registry=4444 --tcp=5555 --ssl=6666  # sets ports to: 4444 \
-5555  6666
+    %(prog)s --prefix=1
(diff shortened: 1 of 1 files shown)


components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -327,12 +327,12 @@ def __init__(this, name, help, wait=False):
...
 Examples:
 
-    %(prog)s --prefix=1                             # sets ports to: 14061, \
-14063, 14064
-    %(prog)s --prefix=1 --revert                    # sets ports back to: \
-4061, 4063, 4064
-    %(prog)s --registry=4444 --tcp=5555 --ssl=6666  # sets ports to: 4444 \
-5555  6666
+    %(prog)s --prefix=1
+        # sets ports to: 14061, 14063, 14064, 14080
```
sbesson: If we move these comments to separate lines, I would be in favor of unifying our policy. My suggestion would be to:
- align them with the command 
- have the comment before the command cf https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/plugins/import.py#L49.

Thoughts? /cc @ximenesuk

sbesson:
One policy comment else :+1: for the content change.

manics:
Rewritten. Do you also want ` %(prog)s` -> `bin/omero admin ports`?

sbesson:
No this is fine. Thanks.

manics:
--no-rebase

PR #3639, opened 2015-03-19
Download_as closes RE

will-moore:
This fixes http://trac.openmicroscopy.org/ome/ticket/12794 which is due to a couple of bugs in webgateway/download_as for batch exporting images as PNG, JPG etc. See https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7766

We were using ``` @login_required(doConnectionCleanup=False) ``` incorrectly here, since we are not streaming data from OMERO directly (don't need to keep services open once the request returns and streaming starts). This meant that open Rendering Engines were not getting closed.

We were also failing to call image._re.close() for each image after rendering and saving into the zip folder. For each image, a new ProxyObjectWrapper for each rendering engine was getting created but the underlying rendering engines were not getting closed each time.

Both these issues are fixed now.
Need to test that all "Export As" functionality is still working.

Potential backport https://trello.com/c/tgJwB05l/188-5-0-9-potential-backports

--no-rebase

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -2009,6 +2009,8 @@ def download_as(request, iid=None, conn=None, **kwargs):
                     imgPathName = makeImageName(
                         img.getName(), format, temp_zip_dir)
                     pilImg.save(imgPathName)
+                    # Close RenderingEngine
+                    img._re.close()
                 # create zip
                 zip_file = zipfile.ZipFile(temp, 'w', zipfile.ZIP_DEFLATED)
                 try:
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks for this, @will-moore. The only thing I would add for these close statements would be that having them in a try/finally would be double-plus-good.

will-moore:
@joshmoore Is that what you meant?

gusferguson:
@will-moore 

Tested with https://trout.openmicroscopy.org/merge/webclient/ user-3 read-only-1 and user-12 on user-3 data.

Behaves as expected.
Good to merge.

joshmoore:
@will-moore : exactly. Thanks. If we could slowly but consistently move to having that everywhere, it'd be great.

will-moore:
@joshmoore OK. For most web uses we only access an image at a time, so we don't get a ton of open REs, but certainly Batch_Image_Export script and OMERO.figure export script need this.

will-moore:
PR for Batch_Image_Export: https://github.com/ome/scripts/pull/100

joshmoore:
Great, thanks, @will-moore!

PR #3640, opened 2015-03-19
Import ome tiff

jburel:
Do not save the roi in the ome-tiff.
To test:
 * Open a local image in ImageJ
 * Crop the image
 * Draw rois on the cropped image. Add them as overlays or add them to the roi manager
 * Save the cropped image to omero
 * The cropped image will be imported as ome-tiff
 * Check that the ROIs are imported but they are not included into the ome-tiff
 * To check that the rois are not included in the ome-tiff. Download the original file
 * Open the ome-tiff and check the XML part of it does not content any roi.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
@@ -23,7 +23,6 @@
 package org.openmicroscopy.shoola.agents.fsimporter.chooser;
 
 //Java imports
-import ij.IJ;
 import ij.WindowManager;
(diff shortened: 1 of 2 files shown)


joshmoore:
Two quick questions:
 * why shouldn't the ROIs be saved?
 * is this intended for 5.1.0?

jburel:
The rois will be saved but not in the ome-tiff, otherwise we cannot modify them afterwards. This follows the workflow of import+roi import for image when in its original format
This is a bug noticed by @bramalingam 
Discussed with @pwalczysko, @bramalingam and @gusferguson

bramalingam:
Everything works as expected. Please merge.

jburel:
--no-rebase

PR #3641, opened 2015-03-19
Add m5 version history

hflynn:
OMERO 5.1.0-m5 version history to be auto-generated in the docs

--no-rebase

cc @sbesson @jburel @joshmoore

### DIFF
--- a/history.txt
+++ b/history.txt
@@ -1,6 +1,21 @@
 OMERO version history
 =====================
 
+5.1.0-m5 (March 2015)
+---------------------
+
+Developer preview release - **only intended as a developer preview for
+updating code before the full public release of 5.1.0. Use at your own risk**.
+
+Changes include:
+
+-  implementation of OMERO.mail for emailing users via the server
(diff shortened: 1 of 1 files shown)


history.txt
```diff
@@ -1,6 +1,21 @@
...
 
+5.1.0-m5 (March 2015)
+---------------------
+
+Developer preview release - **only intended as a developer preview for
+updating code before the full public release of 5.1.0. Use at your own risk**.
+
+Changes include:
+
+-  implementation of OMERO.mail for emailing users via the server
```
joshmoore: Is there anything I should do in the docs PR to make this linkable? or is the formatting for that just too much for the history page?

hflynn: I was going to put the cross-links in the What's New pages rather than here - if we add links here now they will have broken links until your docs are merged. Also, the only links we have on the history page at present are to the security vulnerabilities.

joshmoore: Works for me.

PR #3645, opened 2015-03-20
resurrect DeleteITest

mtbc:
#3142 removed https://github.com/openmicroscopy/openmicroscopy/blob/352c6d1f7d23b163c8c2f4925de274beb61ef990/components/blitz/test/ome/services/blitz/test/DeleteITest.java. This code resurrects and somewhat fixes up many of those test methods in a new form: the class is now in `OmeroJava` extending `AbstractServerTest`, deleting using `Delete2`. Check that https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-java/lastSuccessfulBuild/testngreports/integration.delete/AdditionalDeleteTest/ is passing.

In the longer term, QA permissions review will determine what tests should exist and clean up the coverage from the legacy tests that already exist. For the meantime, more tests are probably better than fewer.

--no-rebase

### DIFF
--- a/components/blitz/resources/ome/services/blitz-graph-rules.xml
+++ b/components/blitz/resources/ome/services/blitz-graph-rules.xml
@@ -212,6 +212,7 @@
         <bean parent="graphPolicyRule" p:matches="I:Image[E]{!a}.rois = [I]" p:changes="I:{a}"/>
         <bean parent="graphPolicyRule" p:matches="I:Image[E]{!a}.stageLabel = [I]" p:changes="I:{a}"/>
         <bean parent="graphPolicyRule" p:matches="Image[I].wellSamples = WS:[E]" p:changes="WS:[O]"/>
+        <bean parent="graphPolicyRule" p:matches="P:Pixels[E]{!a}.channels = [I]" p:changes="P:{a}"/>
         <bean parent="graphPolicyRule" p:matches="P:Pixels[E]{!a}.settings = [I]" p:changes="P:{a}"/>
         <bean parent="graphPolicyRule" p:matches="P:Pixels[E]{!a}.relatedTo = [I]" p:changes="P:{a}"/>
         <bean parent="graphPolicyRule" p:matches="Pixels[I].relatedTo = P:[E]{!a}" p:changes="P:{a}"/>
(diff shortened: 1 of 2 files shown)


joshmoore:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-java/399/testngreports/integration.delete/AdditionalDeleteTest/ is certainly all green.

joshmoore:
Code looks good. The fact that the rules XML file was changed probably needs a minimal nod in @pwalczysko's direction, but conceptually it makes complete sense. (I do wonder what other cases like that might be lurking).

mtbc:
Unfortunately some probably do still lurk, partly caused by how general some of the rules are. https://trello.com/c/Ryvi7Cra/363-review-graph-traversal-rules is mostly about making all the rules more specific so that it's less likely that this kind of thing may crop up; that I'd like to do sooner rather than later. Deletion is probably one of the operations where it is reasonable to specify almost any OMERO model object as the target. (Naturally, @pwalczysko's testing focuses on what can be done from the graphical clients.) For some x in 5.1.x, probably with you and QA, I'd also much like to thoroughly talk through the model graphs we ought to be able to deal with and what should happen in each case, as envisaged in https://trello.com/c/PZeCDbjf/186-graphs-review-chgrp-chown-delete, so that we can make sure that the rules clearly match that and that integration test coverage matches.

pwalczysko:
As for this particular PR:
  - quickly checked Delete behaviour on trout merge, it works as expected
  - the deletion of some (not all) channel(s) should definitely not cause the deletion of an image -> maybe though the client will get into problems if such an image would be opened ?
  - the deletion of all channels would seem to me shoiuld result in the deletion of the image though

As for the card https://trello.com/c/PZeCDbjf/186-graphs-review-chgrp-chown-delete - the rules which need a "decision by UX" - almost each of them needs a further discussion, so that it is more clear what implications such actions on server-side would have. The general rules from the UX side are outlined in https://docs.google.com/a/openmicroscopy.org/presentation/d/18kc75LHiItkHr_dvG1R6jW1qmTNayan8x4ZQC6s7XmI/edit#slide=id.p7 (we are going for "Gus's suggestion"). 
Note that the initial positions of the server and UX team are quite far away, and we are meeting in a compromise ground. The Deep Copy question plays a big role.

mtbc:
One of my challenges this year might be to ask my server-side weird questions in a way that the UX team can make sense of. :smiley:

(I don't know how much sense the use case of deleting channels makes at all; some of the integration tests may not really have plausible acts in mind, in their specifics. But, perhaps we'd not want to automatically delete the image anyway in case it had some important annotation on it or something.)

mtbc:
I don't know how likely it is that it will make the shortlist for 5.1.x work, but my guess would be that we'll do this graph rules review before deep copy, because because the cost of bugs might be high, our delete doesn't come with an undo. But, if TPTB deemed deep copy sufficiently high priority, it could be something that follows not many months after.

joshmoore:
`graph rules` < `deep copy` would seem to make good sense. We'll keep that in mind during scoping. Otherwise, in chatting with @pwalczysko, I can't see undue impact from this. Merging, and we can keep adding tests before/after the re-working.

PR #3637, opened 2015-03-19
update some delete examples

mtbc:
Updates some examples to use `Delete2`, now that `Delete` is deprecated. Examples should still compile and run.

(I am happy to accept corresponding `examples/Training/matlab/DeleteData.m` fixes against this branch.)

--no-rebase

PR #3649, opened 2015-03-22
Remove duplicate MATLAB examples

sbesson:
All these examples are someway (and tested) under `examples/Training/matlab`. The duplicates can be removed.

--no-rebase


sbesson:
Note this is breaking https://ci.openmicroscopy.org/job/OMERO-5.1-merge-docs/451/ which is pointing at the `TreeList` examples.

hflynn:
https://github.com/openmicroscopy/ome-documentation/blob/develop/omero/developers/GettingStarted/AdvancedClientDevelopment.txt looks like it needs an overhaul to me but that's highly unlikely to happen before the release now I think so I don't know whether you want to just do a quick interim fix at this stage?

hflynn:
You could just update the links to point at the matlab training for now.

sbesson:
True although the MATLAB training examples are definitely linked from the MATLAB documentation page. From my standpoint,:
- either we delete these examples and the corresponding doc links in the `AdvancedClientDevelopment` page
- or I just close this PR or revert it to only removing the isolated `Annotations/retrievetags.m`  file. Then post- 5.1.0, we can do some proper scoping on how we organize/maintain/test/document our example suite (see https://github.com/openmicroscopy/openmicroscopy/pull/3637#issuecomment-84906276)

hflynn:
What about a variation of option 1, where you replace the doc links with a link to the MATLAB documentation?
We should then do the post-5.1.0 scoping/clean-up of the rest of that page in addition.

sbesson:
After careful re-reading, the `TreeList` examples are not exactly covered by the training example. Force pushed to clean the annotation M-file only for now.

hflynn:
Ok, works for me. I already had a card about the page in question, have added a link here to it and bumped it up the backlog list.

sbesson:
@bramalingam ping

bramalingam:
Can only see the current state of the PR, Agree with @sbesson's comment https://github.com/openmicroscopy/openmicroscopy/pull/3649#issuecomment-84926184, if this fix is not breaking anything for the time being : Its good to me merged.

PR #3651, opened 2015-03-23
Fixed NPE in importer only startup

dominikl:
This fixes a NPE which is thrown on startup of OMERO.Importer client

Test: Just start OMERO.Importer client


mtbc:
Code change looks sensible, standalone importer client artifact starts and imports just fine. Good to merge.

PR #3657, opened 2015-03-24
Import as thumbnail

jburel:
Problem noticed when retesting ticket https://trac.openmicroscopy.org.uk/ome/ticket/12676
A black thumbnail could be displayed when importing for other
to test:
 * log as admin or group owner
 * Select a folder
 * Import as a user from the group.
 * Check that thumbnails are not black.


dominikl:
Works like expected.

joshmoore:
There are seemingly some whitespace issues, @jburel. But happy to have merged if you don't want to deal with that here.

jburel:
mainly to the mix of tabs and whitespaces in the file. I have stopped removing them for large files otherwise it is tricky to follow changes

jburel:
--no-rebase

PR #3658, opened 2015-03-24
Handle failure if target container is in different group.

ximenesuk:
This PR mitigates the effect of bug described in https://trac.openmicroscopy.org.uk/ome/ticket/12781 It catches the error early and thus avoids the creation of a Fileset and the upload of files for what is a doomed import.

Test via a user who is a member of more than one group. From the CLI try importing into a Dataset in a group other than the one the user is logged in to. It should fail with a meaningful error message. Try a Dataset in the current group, that should work okay. Switching group using `omero sessions` should allow the failed import to succeed. Importing without a target Dataset should succeed. The same applies to Screens. Import via Insight is unaffected.

For example:
```
$ omero import -d 2 ~/Work/Images/dv/CFPNEAT01_R3D.dv
...
Created session 492c2565-dcdb-4d51-95cd-1c6601679506 (user-0@localhost:4064). Idle timeout: 10.0 min. Current group: pg-0
...
2015-03-24 17:43:38,600 6143       [      main] ERROR  formats.importer.cli.CommandLineImporter - Error during import process.
java.lang.RuntimeException: Target container in group: 4, not current group: 3
...

$ omero import -d 1 ~/Work/Images/dv/CFPNEAT01_R3D.dv
Using session 492c2565-dcdb-4d51-95cd-1c6601679506 (user-0@localhost:4064). Idle timeout: 10.0 min. Current group: pg-0
...
2 files uploaded, 1 fileset created, 1 image imported, 0 errors in 0:00:08.362

$ omero sessions group 4
Using session 492c2565-dcdb-4d51-95cd-1c6601679506 (user-0@localhost:4064). Idle timeout: 10.0 min. Current group: pg-0
Group 'pg-0' (id=3) switched to 'rog-0' (id=4)
$ omero import -d 2 ~/Work/Images/dv/CFPNEAT01_R3D.dv
Using session 492c2565-dcdb-4d51-95cd-1c6601679506 (user-0@localhost:4064). Idle timeout: 10.0 min. Current group: rog-0
...
2 files uploaded, 1 fileset created, 1 image imported, 0 errors in 0:00:04.680
```

--no-rebase

### DIFF
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -1903,6 +1903,10 @@ public class OMEROMetadataStoreClient
                             klass.getName(), id));
             }
             long grpID = obj.getDetails().getGroup().getId().getValue();
+            if (grpID != eventContext.groupId) {
+                throw new RuntimeException(String.format("Target container in group: %s, not current group: %s",
+                            grpID, eventContext.groupId));
+            }
             setCurrentGroup(grpID);
             return obj;
         }
--- a/components/tools/OmeroPy/test/integration/clitest/test_import.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_import.py
@@ -26,6 +26,7 @@ import pytest
 import stat
 import re
 import omero
+from omero.cli import NonZeroReturnCode
 from omero.rtypes import rstring
 
 
@@ -560,8 +561,13 @@ class TestImport(CLITest):
 
         assert obj
 
-    @pytest.mark.broken(ticket="11539")
     def testTargetInDifferentGroup(self, tmpdir, capfd):
+        """
+        The workflow this test exercises is currently broken. Until
+        it is investigated and fixed the error is trapped early and
+        an exception is raised. The test is modified to test for this
+        fail case. See ticket 12781 (and 11539).
+        """
         new_group = self.new_group(experimenters=[self.user])
         self.sf.getAdminService().getEventContext()  # Refresh
         dataset = omero.model.DatasetI()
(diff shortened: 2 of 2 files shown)


sbesson:
Conceptually works fine. I had thought of restricting the query to use the context group instead of using `-1` in `OMEROMetadataStore.getTarget()`. However, this implementation produces a more informative error message as it makes a distinction between not existing/found targets and targets in different groups.

Couple of suggestions open for discussion:
- https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/test.integration.clitest.test_import/TestImport/testTargetInDifferentGroup/ works fine but we may want to hide the stack trace. Should we encapsulate the `ic.setXXX` block in `ImportLibrary.importCandidates()` in a `try/catch` block and output a `log.error()` instead?
- other integration tests I can think of related to `getTarget/setTarget`: `testTargetInDifferentGroup` could test both datasets and screens (or 2 tests), additionally we could add a test for non-existing container.

ximenesuk:
I wanted to impact the code as little as possible, hence not changing the `-1` which is the search we really do want to do. This has the benefit of the error message as you observer which would then allow the user to switch groups and retry the import. The 'not found' would not reveal that option.

I agree on both points, output and testing could be improved. Though the stacktrace for the wrong group is positively minimal compared to that for target not found! I'm open to looking at either issue within this PR or later (given neither affects core functionality) depending on timescales for 5.1.0

joshmoore:
Certainly don't mind seeing more tests, but I think the exception is a good reminder of the fact that this is actually just a workaround that we need to fix. In other words, happy with the state of this, but leave you guys to decide on next actions, let's just have this in RC1.

PR #3669, opened 2015-03-27
5.1.0 version history

hflynn:
Tries to combine various lists of new features etc - focusing on user facing things and pointing developers to read the milestone release details.

Won't be built on staging until the autogen job runs.

### DIFF
--- a/history.txt
+++ b/history.txt
@@ -1,6 +1,32 @@
 OMERO version history
 =====================
 
+5.1.0 (March 2015)
+------------------
+
+A full, production-ready release of OMERO 5.1.0; updating the Data Model to
+the January 2015 schema, including adding support for units and new more
+flexible user-added metadata; and introducing new user features, new supported
+formats and many fixes and performance improvements:
+
+-  support for units throughout the Data Model allowing for example, image
+   sizes for electron microscopy to be stored in nanometers rather than all
+   image sizes being set as micrometers
+-  new key-value pairs annotations for adding experimental metadata
+-  copy and paste rendering settings between images and choose different
+   users' settings in the UI
+-  import images to OMERO from ImageJ and save ROIs and overlays from ImageJ
+   to OMERO
+-  improved performance for moving and deleting data
+-  import times are much improved for large datasets such as HCS and SPIM data
+-  improved performance for many file formats and new supported formats via
+   Bio-Formats (now over 140)
+-  new OMERO.mail feature lets admins configure the server to email users
+-  support for limiting the download of original files
+-  many developer updates such as removal of deprecated methods, and updates
+   to OMERO.web and the C++ implementation (see the 5.1.0-m1 to 5.1.0-m5
+   developer preview release details below)
+
 5.1.0-m5 (March 2015)
 ---------------------
 
--- a/history.txt
+++ b/history.txt
@@ -5,18 +5,20 @@ OMERO version history
 ------------------
 
 A full, production-ready release of OMERO 5.1.0; updating the Data Model to
-the January 2015 schema, including adding support for units and new more
+the January 2015 schema, including support for units and new more
 flexible user-added metadata; and introducing new user features, new supported
 formats and many fixes and performance improvements:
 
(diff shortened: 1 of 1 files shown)


jburel:
improved performance for moving and deleting data => improve performance

jburel:
improved performance for many file formats => same

hflynn:
Pushed, thanks Will, your ping got a bit lost in a barrage of comments from Petr and I forgot it.

hflynn:
@jburel I'm not sure 'improve' reads better personally, don't know what anyone else thinks.

jburel:
'Import As' for group owners as well as administrators. => "Administrators" this is not a new feature. It has been in since 5.0

jburel:
I will probably also add that OMERO.editor has been removed.

history.txt
```diff
@@ -1,6 +1,34 @@
...
+-  support for units throughout the Data Model allowing for example, pixel
+   sizes for electron microscopy to be stored in nanometers rather than being
+   set as micrometers
+-  new, searchable key-value pairs annotations for adding experimental
+   metadata
+-  improved workflow for rendering settings in the UI and parity between the
+   clients
+-  import images to OMERO from ImageJ and save ROIs and overlays from ImageJ
+   to OMERO
+-  'Import As' for group owners as well as administrators
```
will-moore: My first impression when I read 'Import As' is that it must be like 'Save As', E.g. 'Import As OME-TIFF' or 'Import As SPW'. For users not familiar with 'Import As' (probably most, since only Admins have used this before), it might be clearer to say 'Import as another user' or just say "Group owners can now import data on behalf of other group members"?

history.txt
```diff
@@ -1,6 +1,34 @@
...
+-  improved performance for moving and deleting data
+-  import times are much improved for large datasets such as HCS and SPIM data
+-  improved performance for many file formats and new supported formats via
+   Bio-Formats (now over 140)
+-  new OMERO.mail feature lets admins configure the server to email users
+-  support for limiting the download of original files
```
will-moore: Why would you want to do that? Maybe add "...for public facing OMERO.web deployments"?

jburel:
Missing:
* removed calculation of PGI (pretty good image). The default rendering settings are now set to global min/max if calculated or to full pixels range.

history.txt
```diff
@@ -1,6 +1,41 @@
...
+-  importing as another user, previously only available for administrators, is
+   now usable by group owners as well, allowing you to import data that will
+   then be owned by the user you import it for
+-  improved performance for moving and deleting data
+-  import times are much improved for large datasets such as HCS and SPIM data
+-  improved performance for many file formats and new supported formats via
+   Bio-Formats (now over 140)
+-  improved rendering by removing calculation of PGI (Pretty Good Image) so
+   that default rendering settings are now set to global min/max if it is
+   calculated, or to full pixel range
```
pwalczysko: This is a bit confusing. So at the beginning of the sentence it says that ``...by removing calculation....``, suggesting that the calculation was removed. Later in the sentence it says ``...if it is calculated....`` ???

will-moore: - removed the auto-levels calculation for initial rendering settings, simply using the min/max pixel intensities. (maybe omit the "if not calculated" bit here, since this is kinda an edge case and should probably be mentioned under performance improvements rather than rendering improvements.

jburel: @pwalczysko: this is 2 different things. ```if calculated``` reference to the min/max value. so probably better to use the term if available

jburel: @will-moore: this is not always true. It is not available for large float/double images.

hflynn:
Ok, hopefully both clear and true now!

will-moore:
Looks fine to me

pwalczysko:
Looks good to me.

jburel:
@hflynn: read well now. Thanks

joshmoore:
Merging after all the LGTMs. Last minute adjustments might still be made.

hflynn:
--no-rebase

PR #3671, opened 2015-03-30
C++: add conversions.h to CMakeLists.txt

joshmoore:
Without conversions.h in lib/include/omero, compiling
a client is all but impossible.

cc: @rleigh-dundee @emilroz

### DIFF
--- a/components/tools/OmeroCpp/CMakeLists.txt
+++ b/components/tools/OmeroCpp/CMakeLists.txt
@@ -347,6 +347,7 @@ set(OMERO_CLIENT_STATIC_HEADERS
     src/omero/callbacks.h
     src/omero/client.h
     src/omero/clientF.h
+    src/omero/conversions.h
     src/omero/min.h
     src/omero/ObjectFactory.h)
 


rleigh-dundee:
The change looks fine and good to merge if everything is green.

Were any other headers added since the cmake conversion which might also be missing?  There weren't last time I checked, but I haven't looked in detail over the last few months at what might have been added in the interim.

On the CI side, we might want to develop some OmeroCpp clients independent of openmicroscopy.git to be built against the installed library/headers.  Or have an alternative CMakeLists.txt for the examples so they can be built independently against external headers/libs.  Either would allow validation of the install which we don't currently have.

joshmoore:
> Were any other headers added since the cmake conversion which might also be missing?

Was just checking this. This seems to say, "no":
```
$ git diff --name-status v5.1.0-m1..HEAD . | grep -E ^A | grep -E h\$ | xargs -IX echo X | cut -b33- | xargs -IY ls /opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac\ OS\ X-10.10.2-x86_64/include/Y
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/IceNoWarnPop.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/IceNoWarnPush.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/conversions.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/model/ElectricPotentialI.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/model/FrequencyI.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/model/LengthI.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/model/PowerI.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/model/PressureI.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/model/TemperatureI.h
/opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include/omero/model/TimeI.h
ls: /opt/ome8/target/OMERO.cpp-5.1.0-m5-ice35-SNAPSHOT-Mac OS X-10.10.2-x86_64/include//safegtest.h: No such file or directory
```

joshmoore:
> Or have an alternative CMakeLists.txt for the examples

Agreed. In fact, I thought that was still happening.

rleigh-dundee:
@joshmoore The examples are a standard part of the cmake build, so they build in-tree every time you do a build.  The reason was so that they wouldn't bitrot (they hadn't been built for well over a year or so with scons as far as I could tell and needed fixing up).  We can easily add a second alternatively-named file here to set up the include/library paths and then source the regular CMakeLists.txt.  Could probably be added to the same build job, even.

rleigh-dundee:
I've done a test build and install and I can confirm the header is now installed correctly.  Please merge.

PR #3675, opened 2015-03-31
Final import mode

ximenesuk:
Hitherto the import modes have lacked a certain symmetry. This PR fixes that unfortunate shortcoming by adding the final import mode, thus completing the symmetric set and restoring necessary beauty to import. This final mode is `ln_s_rm`, a symlink followed by a deletion of the original files.

Testing
-------

To test this PR use the command line import with the transfer option `ln_s_rm`, ie

```
bin/omero import -- --transfer=ln_s_rm path/to/image/file
```

The console output should confirm that the correct mode is used, that the image is transferred and imported and that the original image files are successfully deleted.

A further useful sanity check would be to view the imported image in one of the clients.

Caution: given that this is a `rm` mode in-place import you should fully understand the side effects of this import mode and not test it using the only copies of valuable images.

--no-rebase

### DIFF
--- a/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
+++ b/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
@@ -371,6 +371,7 @@ public class CommandLineImporter {
             + "          ln              \t# Use hard-link.\n"
             + "          ln_s            \t# Use soft-link.\n"
             + "          ln_rm           \t# Caution! Hard-link followed by source deletion.\n"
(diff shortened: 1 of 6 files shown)


dominikl:
I'm just wondering what this mode really does... from a filesystem's view, creating a soft link to a file and then removing the file, will result in a 'broken' soft link pointing to nowhere.

dominikl:
:-) Got me... wasn't aware of today's date, argh....

mtbc:
Thank you @ximenesuk for the entertainment. :+1: You had me unsure until I saw the copyright headers!

PR #3678, opened 2015-04-02
adjust gateway and CLI away from deprecated delete

mtbc:
Affects unlinkAnnotations in Python gateway and `bin/omero fs rename`. Integration tests should still pass and functions depending on them should still work.

--no-rebase


jburel:
cc @ximenesuk

PR #3679, opened 2015-04-02
Remove deprecated search method from Insight

dominikl:
Test:
Check that the search functionality in Insight is still working

--no-rebase


gusferguson:
@dominikl 

Tested with OMERO.insight-5.1.0-116-9075c2b-ice34-b463-mac_Java7+ user-3 read-only 1 and read-annotate-1.

Behaves as expected.
Good to merge.

jburel:
Ideally the method from the I/F should be marked as deprecated if insight is used headless but this is probably not happening.

PR #3680, opened 2015-04-02
Matlab : Update deleteObjects

bramalingam:
To Test,

Download the omeroMatlab toolbox,
Follow the delete work flow as follows,

%Params
hostname= 'localhost';  %Host address
username = 'root';  %Username for Insight
imageId= 1;
password = 'ome'; %Password for Insight

%--------------------------------------------------------------------------

%load client 
client = loadOmero(hostname);
session = client.createSession(username, password);

%delete Image
deleteImages(session, imageId);

%Similarly try:
Projects,Datasets,PlateAcquisitions,Plates and Screens

### DIFF
--- a/components/tools/OmeroM/src/createObjectDictionary.m
+++ b/components/tools/OmeroM/src/createObjectDictionary.m
@@ -37,5 +37,5 @@ for  i = 1 : nObjects
     d(i).name = names{i};
     d(i).class = ['omero.model.' classnames{i}];
     d(i).Iobject = str2func([d(i).class 'I']);
-    d(i).delete = ['/' classnames{i}];
+    d(i).delete = [classnames{i}];
 end
--- a/components/tools/OmeroM/src/createObjectDictionary.m
+++ b/components/tools/OmeroM/src/createObjectDictionary.m
@@ -32,11 +32,11 @@ ip.addRequired('classnames', @(x) iscellstr(x) && numel(x) == nObjects);
 ip.parse(names, classnames);
 
 % Initialize structure array
-d(1 : nObjects) = struct('name', '', 'class', '', 'Iobject', '', 'delete', '', 'targetObject', '');
+d(1 : nObjects) = struct('name', '', 'class', '', 'Iobject', '', 'delete', '', 'delete2', '');
 for  i = 1 : nObjects
     d(i).name = names{i};
     d(i).class = ['omero.model.' classnames{i}];
     d(i).Iobject = str2func([d(i).class 'I']);
     d(i).delete = ['/' classnames{i}];%@deprecated
-    d(i).targetObject = java.util.Hashtable;
+    d(i).delete2 = classnames{i};
 end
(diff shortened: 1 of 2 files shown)


components/tools/OmeroM/src/delete/deleteObjects.m
```diff
@@ -42,14 +42,13 @@ function deleteObjects(session, ids, type)
...
-    deleteCommands(i) = omero.cmd.Delete(objectType.delete, ids(i), []);
+    idlist.add(java.lang.Long(ids(i)));
 end
-
-% Create DoAll command object and add delete commands to the requests
-doAll = omero.cmd.DoAll();
-doAll.requests = toJavaList(deleteCommands);
+targetObj = java.util.Hashtable;
+targetObj.put(objectType.delete(2:end),idlist);
+deleteCommands = omero.cmd.Delete2(targetObj, ChildList, true);
```
mtbc: Did it turn out you can pass a null childlist, or does it make you create an empty list?

mtbc: do you need `false` instead of `true`?

bramalingam: Oh yes. It needs to be false.

bramalingam: and yes the ChildList has been set to null. This works as well.

jburel:
* Testing all the various delete methods All work fine.
* The ```createObjectDictionary``` should be updated, we no longer need delete field i.e. ```/```

sbesson:
Regarding `createObjectDictionary`, I would stick to the same strategy as the rest of the code and keep the `delete` field but mark it as deprecated (and remove it when the old delete command is removed). Looking at the code, a possibility would be to store the object `classname` as another field in the returned objects structure and use `object(i).classname` in the `targetObj.put()`

bramalingam:
If am right, JM's suggestion was to remove the '/' in the delete field. I am anyways stripping that away, when am creating the hash map.

jburel:
@bramalingam: no, not to remove the "/", @sbesson is more suitable.

jburel:
You have made the changes!!

components/tools/OmeroM/src/createObjectDictionary.m
```diff
@@ -32,10 +32,11 @@
...
 % Initialize structure array
-d(1 : nObjects) = struct('name', '', 'class', '', 'Iobject', '', 'delete', '');
+d(1 : nObjects) = struct('name', '', 'class', '', 'Iobject', '', 'delete', '', 'targetObject', '');
 for  i = 1 : nObjects
     d(i).name = names{i};
     d(i).class = ['omero.model.' classnames{i}];
     d(i).Iobject = str2func([d(i).class 'I']);
-    d(i).delete = ['/' classnames{i}];
+    d(i).delete = ['/' classnames{i}];%@deprecated
+    d(i).targetObject = java.util.Hashtable;
```
sbesson: I am not sure this is adding much since the value is independent of the nature of the object. My initial proposal was to store the class name i.e:

```
  d(i).classname = classnames{i}
```

and then use this new field as an argument in replacement of the deprecated `delete`:

```
targetObj.put(objectType.classname, idlist);
```

sbesson:
:+1:  Good to merge tomorrow assuming the training job is green.

sbesson:
See https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/289/console

PR #3682, opened 2015-04-03
Omero-IJ Plugin : Roi reader update Z T

bramalingam:
Testing instructions:

1) Download and add the OMERO-IJ plugin to the Plugins folder of ImageJ.
2) Open an OMERO image on ImageJ using the Omero-IJ plugin.
3) Add ROIs to image (using the ROIManager/overlays). 
4) Now : Plugins --> OMERO --> Save ROIs to OMERO.
5) This should save all the ROI's to the OMERO image.

To Check:
1) Open the image on insight, using the view option within the plugin.
2) Check if the newly created ROI's have the correct Z and T information associated with them.

### DIFF
--- a/components/blitz/src/loci/ome/io/OmeroReader.java
+++ b/components/blitz/src/loci/ome/io/OmeroReader.java
@@ -588,7 +588,7 @@ public class OmeroReader extends FormatReader {
         if (r == null) {
             return null;
         }
-        return new NonNegativeInteger(r.getValue());
+        return new NonNegativeInteger(r.getValue()+1);
     }
 
     /** Converts omero.model.Shape (omero.model.RectI in this case) to ome.xml.model.* and updates the MetadataStore */
--- a/components/blitz/src/loci/ome/io/OmeroReader.java
+++ b/components/blitz/src/loci/ome/io/OmeroReader.java
@@ -53,6 +53,7 @@ import ome.units.UNITS;
 import ome.xml.model.enums.FontStyle;
 import ome.xml.model.primitives.NonNegativeInteger;
 import ome.xml.model.primitives.Timestamp;
+import omero.RInt;
 import omero.RString;
 import omero.RTime;
 import omero.ServerError;
@@ -279,11 +280,11 @@ public class OmeroReader extends FormatReader {
                 client = client.createClient(false);
                 serviceFactory = client.getSession();
             }
-            
+
             IAdminPrx iAdmin = serviceFactory.getAdminService();
             IQueryPrx iQuery = serviceFactory.getQueryService();
             EventContext eventContext = iAdmin.getEventContext();
-            
+
             if (group != null || groupID != null) {
 
                 ExperimenterGroup defaultGroup =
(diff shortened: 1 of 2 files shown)


jburel:
why the +1 when adding z/t or c?

jburel:
```int c = roi.getCPosition();shape.setC(c-1); ``` review formatting

bramalingam:
There was a indexing difference between Omero and ImageJ. One started from 0, and the other started from 1. This was specific for C,Z and T alone, was not the case for ROI's and etc.

jburel:
index start at 0 for c,z, t when saved in OMERO. UI is +1

bramalingam:
ImageJ indexing seemed to start from 1 for C,Z and T

bramalingam:
The ROI's got propagated to all planes/Channels/Time-points in ImageJ, if the value was set to 0.

pwalczysko:
Overlay workflow from IJ to OMERO does not work for T.
This defaults always to 1 in OMERO, whatever the value was in IJ.

jburel:
* roi.getZPosition() is an int so you will only need to check if !=0 (same for C and T)
* ```if(z!=0){ ``` formatting ```if (z != 0) {```
* remove unused ij.IJ import

jburel:
I have added the image at a given Z but it is propagated to all z

bramalingam:
@jburel : Did u try this : 
If you want ImageJ to display the ROI's according to the slice information, please select the following option in the ROIManger : 
ROIManger--> More >> --> Options --> Associate "Show All" ROIs with slices

This will associate the information while viewing the Image as well.

Sorry I had added this to the other Bioformats PR. Should have added it here as well.

pwalczysko:
Retest in Fiji with updated insight-ij plugin and bioformat-plugins.jar.

ROI Manager workflows: All worked as expected, including both single and multiple regions, choosing non-default channels in Fiji and propagation of regions (from OMERO to Fiji).
The only confusion (not a bug though) comes from the fact that Fiji/IIJ is particular about association of ROIs with channels, whereas Insight (or maybe whole OMERO) is not. This means when you draw a region in OMERO and open in Fiji, it is propagated by default over all channels. If you draw a region in Fiji, in OMERO it is also propagated over all channels, but the info about the channel on which you drawn the region in Fiji is preserved. The next time you reopen the same image in Fiji, you will see the particular region on the associated channel only. Nevertheless, if you propagated the region in OMERO in between, the new, propagated regions do not have the channel info anymore, and will show after reopen in Fiji on all channels. Quite edge case though.

pwalczysko:
Overlay workflows: Multiple problems here.
   - All is fine in case of multi-z & single T image
   - T information is not being preserved form Fiji to OMERO in case of multi-z & multi-T image
      - Insight shows no info in the T column of Manager tab of Roi Manager
      - elipse is being displayed without lines, just contours are visible in this case
      - when the shape is clicked on in Manager, the shape shows up on the image, but when T slider is dragged, the shape vanishes from image
   - T information is partially preserved from Fiji to OMERO in case of single-z & multi-T image
     - insight shows correct T in the T column of Roi Manager
     - when the shape is clicked in Manager, the right T is selected in the image and the shape is displayed
     - ellipse is displayed okay in this case
     - when the T slider in image is scrolled, the shape still shows on all T planes (it shoiuld show only on the associated plane, as indicated in the Manager T column)

pwalczysko:
In summary,
   - ROI Manager workflows are okay
   - Overlay workflows are not okay for T

As I understand it, this PR is not attempting anymore to sort out the Overlay problems, as this is not OMERO-side problem. If this assumption is correct, then this is ready to merge.

@jburel ?

jburel:
Overlays in general are attached to an image (major difference with ROI manager)
so we might have to check the provenance of the shapes i.e. overlay or roi manager. We should investigate further if the information is stored in Fiji/IJ. 
This can be done in another PR.

jburel:
overlays seem to be missing the T information. Not under our controls
Minor formatting:
```if(c!=0){ ``` => ```if (c != 0) {```
same for z, t

bramalingam:
@jburel, @sbesson , @joshmoore  : Done.

jburel:
@bramalingam: Thanks waiting on travis and build before merging

jburel:
Travis is happy, thanks merging

PR #3693, opened 2015-04-06
Replaced deprecated Delete/Chgrp under OmeroPy

ximenesuk:
This PR moves most of the Python integration tests to using Delete2 and Chgrp2. At the moment it doesn't alter any tests that use the gateway methods `deleteObjects` and `chgrpObjects`. Those methods themselves need to be modified first. Further commits may be added to this PR, or done in a second PR along with other changes to code such as plugins if necessary.

The first two commits from @mtbc modify the `fs` plugin and a `gateway` method.

The test https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/test_delete.py#L272 is also currently not updated  since it relies on the response object which has changed. I'm not sure this is related to ticket 3639 so I'm not too sure what it is meant to test and whether it is effectively tested elsewhere without the parsing of the response.

Testing
----------
The modified integration tests should pass.

`fs rename` should work as expected.


ximenesuk:
I want to slightly reshape this PR, adding permission test changes and removing non-test code to a separate PR. Closing temporarily.

PR #3694, opened 2015-04-06
ROI example

jburel:
Add support for Point and Line in ROIs example.


jburel:
@ximenesuk: good point. It usually prefers setter. I followed the structure already in place.
Something to discuss with @will-moore when he gets back, he might be able to remember the reason why it was done that way in the first place.

will-moore:
I have no strong opinion. Setters is fine with me.

jburel:
I have opted: for no setters when creating the shapes and getters when reading the shapes. So it is clear that both can be used.

ximenesuk:
Changes look good to merge.

PR #3695, opened 2015-04-07
Optimize and fix original file paths

chris-allan:
This PR includes various performance improvements to the stack required to extract original file paths from the server via the `/webgateway/original_file_paths/...` OMERO.web JSON endpoint. Without these or similar changes there is the very real scenario where requesting the original file paths could **crash** your server due to the sheer amount and complexity of data being processed.

The "real world" example dataset which spurred the need for the improvements has 4608 Images (384 Wells, 12 Fields) and 4608 OriginalFiles (one per field) associated with the same Fileset. Half that number is needed to run a server out of memory with a 1GB heap. Call tree is as follows:

```
original_file_paths()
\___ _ImageWrapper.getImportedImageFiles()
\___ _ImageWrapper.getFileset()
\___ _ImageWrapper.countFilesetFiles()
\___ BlitzGateway.getFilesetFilesInfo()
```

The first set of improvements are to the fileset and archive file count methods, `getFilesetFileInfo()` and `getArchivedFilesInfo()`, on the gateway. Before:

```
...
In [25]: print datetime.now()
2015-04-03 08:45:46.464612

In [26]: gw.getFilesetFilesInfo([66101])

Out[26]: {'count': 4608, 'size': 6564240074L}

In [27]: print datetime.now()
2015-04-03 08:45:49.193313
...
```

After (14ee8844e62c434d8c4b827735a3d10424abc1b4):

```
In [12]: print datetime.now()
2015-04-03 14:38:57.178159

In [13]: gw.getFilesetFilesInfo([66101])
Out[13]: {'count': 4608L, 'size': 6564240074L}

In [14]: print datetime.now()
2015-04-03 14:38:57.371813
```

~2.7seconds vs. ~190ms (~14x speedup)

The second set of improvements are to the original file load method `getImportedImageFiles()` on the image wrapper. Before, even with a 2GB heap the aforementioned dataset causes out of memory and GC overhead limit exceptions. This is likely due to the multiplicity of the queries performed and the complexity of the object graph. The number of objects loaded would be >20,000.

After (1356aff4c2f3f18728cce5a0cc67973248285777):

```
In [11]: print datetime.now()
2015-04-03 19:04:43.403131

In [12]: files = list(image.getImportedImageFiles())
print datetime.now()

In [13]: print datetime.now()
2015-04-03 19:04:49.463482
```

~6seconds.

These projection changes complete the speedup.

After (27ce963e1a7e8a97ac346c3d68982f44855a2bf4), run on `localhost`:

```
$ time curl "https://localhost/webgateway/original_file_paths/66101/?server=1&username=...&password=..." &> /dev/null

real	0m0.356s
user	0m0.024s
sys	0m0.037s
```

~350ms

Integration tests cover the affected code paths. /cc @ximenesuk 

To be reviewed by @will-moore and/or @aleksandra-tarkowska.

/cc @jballanc, @knabar, @emilroz, @jburel, @sbesson

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2829,7 +2829,9 @@ class _BlitzGateway (object):
                 '   where i_image.id in (:ids)'\
                 ')'
         queryService = self.getQueryService()
-        count, size = queryService.projection(query, params, self.SERVICE_OPTS)[0]
+        count, size = queryService.projection(
+            query, params, self.SERVICE_OPTS
+        )[0]
         if size is None:
             size = 0
 
@@ -2876,7 +2878,9 @@ class _BlitzGateway (object):
                 '    where i_link.child.image.id in (:ids)'\
                 ')'
         queryService = self.getQueryService()
-        count, size = queryService.projection(query, params, self.SERVICE_OPTS)[0]
+        count, size = queryService.projection(
+            query, params, self.SERVICE_OPTS
+        )[0]
         if size is None:
             size = 0
         return {'fileset': False, 'count': unwrap(count), 'size': unwrap(size)}
@@ -8804,7 +8808,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         for original_file in original_files:
             yield OriginalFileWrapper(self._conn, original_file)
 
-    def getImportedImageFilePaths (self):
+    def getImportedImageFilePaths(self):
         """
         Returns a generator of path strings corresponding to the Imported
         image files that created this image, if available.
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2824,7 +2824,6 @@ class _BlitzGateway (object):
                 'from FilesetEntry as fse where fse.id in ('\
                 '   select distinct(i_fse.id) from FilesetEntry as i_fse '\
                 '   join i_fse.fileset as i_fileset'\
-                '   join i_fse.originalFile '\
                 '   join i_fileset.images as i_image '\
                 '   where i_image.id in (:ids)'\
                 ')'
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2830,6 +2830,8 @@ class _BlitzGateway (object):
                 ')'
         queryService = self.getQueryService()
         count, size = queryService.projection(query, params, self.SERVICE_OPTS)[0]
+        if size is None:
+            size = 0
 
         query = 'select ann.id, ann.ns, ann.textValue '\
                 'from Fileset as fileset '\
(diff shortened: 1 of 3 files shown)


sbesson:
Quick fire test performed on a large  SPIM dataset (~3600 files) by asking the Web client to display the file paths:
- without this PR, the query is started and the server hangs as described (7GB heap), requiring a restart of the server
- with this PR, the same query on the same dataset with a server with similar memory settings returns within 100ms.

aleksandra-tarkowska:
@chris-allan I am struggling with a large dataset. I managed to test:

I managed to test ```getFilesetFilesInfo```:
 - with Plate 
```
Merge user-3
conn.getFilesetFilesInfo([24001])
{'count': 123L, 'annotations': [{'ns': None, 'id': None}], 'fileset': True, 'size': 1008799953L}
```

Paths on server: ```user-3_9/2015-03/12/09-07-54.472/plate1_G11.pnl ...```

```
$ du /repositories/OMERO-5.1-merge-deploy/ManagedRepository/user-3_9/2015-03/12/09-07-54.472/
985408	/repositories/OMERO-5.1-merge-deploy/ManagedRepository/user-3_9/2015-03/12/09-07-54.472/
```
looks like values are different. I tested with:

```
size = 0
for f in img.getFileset().listFiles():
    size +=f.size
print size
1008799953
```
might be worth to look

 - with SVS:
```
Merge: user-1
conn.getFilesetFilesInfo([230854])
{'count': 1L, 'annotations': [{'ns': None, 'id': None}], 'fileset': True, 'size': 340130041L}
statinfo.st_size: 340130041
```

mtbc:
(As an aside: I don't know if `DiskUsage` or `UsedFilesRequest` from http://downloads.openmicroscopy.org/omero/5.1.0/api/slice2html/omero/cmd.html are of any use for this kind of thing, or if their performance is adequate under comparable stresses, but we may want to review their design if they could make the gateway's life easier.)

chris-allan:
I would guess that difference is because `du` is dividing by 1024, @aleksandra-tarkowska:

```
In [2]: 1008799953 / 1024
Out[2]: 985156
```

The additional size (`985408 - 985156 = 252`) is the on disk overhead of the directories in that hierarchy.

Can't comment on the user of those commands, @mtbc, other than to say that as I recall they were fundamentally operating on IDs. Sending 1000's, at worst tens of thousands, of IDs back to the server then for an `IN` query so you can get a list of paths is certainly not ideal.

/cc @joshmoore

mtbc:
@chris-allan: Makes sense, not ideal at all.

aleksandra-tarkowska:
```
$ du --bytes /repositories/OMERO-5.1-merge-deploy/ManagedRepository/user-3_9/2015-03/12/09-07-54.472/
1008804049    /repositories/OMERO-5.1-merge-deploy/ManagedRepository/user-3_9/2015-03/12/09-07-54.472/

1008804049 - 1008799953 = 4096
```
thank @chris-allan for point that out

will-moore:
I haven't repeated the performance testing above, but code changes seem fine and web functionality and UI appears unchanged. Good to merge.

will-moore:
Batch download of a variety of files (with and without filesets) is also still working fine (uses the methods above).

chris-allan:
Just realised there might be an issue if there are no annotations present based on the output of @aleksandra-tarkowska. I'll look at this today.

joshmoore:
Closing to re-open with an additional patch.

PR #3696, opened 2015-04-07
Replaced deprecated methods OmeroPy integration tests

ximenesuk:
This PR moves most of the Python integration tests to using Delete2, Chgrp2 and Chmod. At the moment it doesn't alter any tests that use the gateway methods deleteObjects and chgrpObjects. Those methods themselves need to be modified first.

Testing
-------
The modified integration tests should pass.


sbesson:
Sorry for the delay. All Python integration tests are definitely passing and this PR should be good to merge maybe after a second run of the integration tests.
Only minor thought while reading through the test library changes is that a docstring could be added to the `doAllSubmit()` method to explain its usage scope vs `doSubmit()`. Otherwise, would it make sense to merge both methods into a single one and wrap the requests input as part of a `DoAll` if it is a list?

ximenesuk:
Thanks @sbesson, that sounds like a useful change, I'll take a look tomorrow and push a commit.

sbesson:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/417/ has 1 transient failure unrelated to these changes. Thanks @ximenesuk, merging.

PR #3697, opened 2015-04-08
psql: Correct version error message to pretty-print version

rleigh-dundee:
Make the error message for a too-old db server version a bit more friendly:
- Include the name "PostgreSQL" rather than the generic "database"
- Pretty-print version numbers rather than using a bare integer
- Use a simple PL/pgSQL function to format the integer version

--------

Testing: Run the db script on an old server, you'll see a message like this:

```
psql:OMERO5.1__1.sql:53: ERROR:  PostgreSQL database server version 9.1.13 is less than OMERO prerequisite 9.2.0
```

### DIFF
--- a/components/dsl/resources/ome/dsl/psql-header.vm
+++ b/components/dsl/resources/ome/dsl/psql-header.vm
@@ -20,6 +20,16 @@ BEGIN;
 -- check PostgreSQL server version and database encoding
 --
 
+CREATE OR REPLACE FUNCTION db_pretty_version(version INTEGER) RETURNS text AS $$
+DECLARE
+    textver TEXT;
+
+BEGIN
+    textver = (version/10000)::text || '.' || ((version/100)%%100)::text || '.' || (version%%100)::text;
+    RETURN textver;
+
+END;$$ LANGUAGE plpgsql;
+
 CREATE FUNCTION assert_db_server_prerequisites(version_prereq INTEGER) RETURNS void AS $$
 
 DECLARE
@@ -31,7 +41,7 @@ BEGIN
     SELECT pg_encoding_to_char(encoding) INTO STRICT char_encoding FROM pg_database WHERE datname = current_database();
 
     IF version_num < version_prereq THEN
-        RAISE EXCEPTION 'database server version %% is less than OMERO prerequisite %%', version_num, version_prereq;
+        RAISE EXCEPTION 'PostgreSQL database server version %% is less than OMERO prerequisite %%', db_pretty_version(version_num), db_pretty_version(version_prereq);
     END IF;
 
     IF char_encoding != 'UTF8' THEN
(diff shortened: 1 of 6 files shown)


manics:
New DB:
```
BEGIN
CREATE FUNCTION
CREATE FUNCTION
ERROR:  PostgreSQL database server version 8.4.20 is less than OMERO prerequisit
e 9.2.0
...
```
5.0 -> 5.1 upgrade (`sql/psql/OMERO5.1__1/OMERO5.0__0.sql`):
```
BEGIN
CREATE FUNCTION
 omero_assert_db_version
-------------------------

(1 row)

DROP FUNCTION
CREATE FUNCTION
CREATE FUNCTION
ERROR:  PostgreSQL database server version 8.4.20 is less than OMERO prerequisite 9.2.0
...
```

Good to merge

joshmoore:
We'll need to address the fact that this will leave new users with the method and old users without. Either this will have to wait for the next DB bump, or should be rolled into any auto-upgrades that happen for 5.1.1. cc @mtbc

rleigh-dundee:
Presumably it will get copied into / sourced by any future upgrade scripts, so having the older scripts updated won't be a problem.

joshmoore:
Changing the old scripts isn't per se a problem, but is more or less pointless since they will have already been executed and can't be executed again. In general, for any DB change [the steps](https://www.openmicroscopy.org/site/support/contributing/schema-changes.html) are:

 * modify the templates under `dsl/resources`
 * bump the dbpatch
 * generate the schema with `build-schema`
 * prepare the upgrade which can be either an offline upgrade which we do on major versions or if possible an auto-upgrade.

mtbc:
I'd be leery of rolling this into auto-upgrade because it'll be rather PostgreSQL-specific but certainly happy to see this for 5.2 even if there is awkwardness within 5.1. But, maybe it's okay for people to have different versions: on http://www.openmicroscopy.org/site/support/omero5.1/sysadmins/server-upgrade.html I cunningly wrote rather vaguely only of, "a message saying that your database server version is less than the OMERO prerequisite".

joshmoore:
Recorded under https://trello.com/c/cur5hd3e/44-5-2-db-changes; closing for the moment.

mtbc:
@joshmoore: The DB patch bump makes sense for SQL that leaves a lasting change in the database, but I don't see the point for transient checks before the upgrade even starts, it just means that users who didn't upgrade yet don't see the benefit. Maybe for discussion elsewhere though.

mtbc:
I'll use this new version for https://trello.com/c/A4nhbaE8/33-have-sql-upgrade-scripts-require-pg-9-3.

mtbc:
See `db_pretty_version` in https://github.com/openmicroscopy/openmicroscopy/commit/22069e51f085d9a710e2871bf02b93d34649f6cd and used elsewhere since.

PR #3698, opened 2015-04-09
Factor out doAllSubmit from test library.

ximenesuk:
This PR follows from a comment at https://github.com/openmicroscopy/openmicroscopy/pull/3696#issuecomment-90985790

The only tests affected are the ones modified in this PR, they should pass.

/cc @sbesson 

--no-rebase

### DIFF
--- a/components/tests/python/library/__init__.py
+++ b/components/tests/python/library/__init__.py
@@ -741,9 +741,14 @@ class ITest(object):
     def doSubmit(self, request, client, test_should_pass=True,
                  omero_group=None):
         """
-        Performs the request waits on completion and checks that the
-        result is not an error.
+        Performs the request(s), waits on completion and checks that the
+        result is not an error. The request can either be a single command
+        or a list of commands. If the latter then the request list will be
+        wrapped in a DoAll.
         """
+        if isinstance(request, list):
+            request = DoAll(request)
+
         sf = client.sf
         if omero_group is not None:
             prx = sf.submit(request, {'omero.group': str(omero_group)})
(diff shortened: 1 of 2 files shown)


sbesson:
Getting a "We can’t automatically merge this pull request." (which is surprising). Could you try merging `origin/develop`?

sbesson:
Despite GH issues requiring the additional merge commit, this PR was included in this morning's build and https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/418/ is green apart from an intermittent search failure. Changes are sensible and ready to merge although we can afford to wait for another run on Monday to stay on the safe side.
A semi-related discussion we had with @ximenesuk earlier this morning: do/should we have a policy with regard to breaking the test API (which is mostly consumed internally) on a stable line? /cc @joshmoore @jburel

components/tests/python/library/__init__.py
```diff
@@ -747,9 +747,14 @@ def loginAttempt(self, name, t, pw="BAD", less=False):
...
                  omero_group=None):
         """
-        Performs the request waits on completion and checks that the
-        result is not an error.
+        Performs the request(s), waits on completion and checks that the
+        result is not an error. The request can either be a single command
+        or a list of commands. If the latter then the request list will be
+        wrapped in a DoAll.
         """
+        if isinstance(request, list):
```
joshmoore: and tuples a la [iter()](http://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable)?

ximenesuk: Do we need to check for a tuple if we are specifying a list as the parameter and as this is a helper in our test environment? Happy to add that, just wondering whether it is needed.

joshmoore: Probably not. I was just reading through all commits and assumed `gateway` rather than `library`. Sorry for the noise.

joshmoore:
I don't think we have to worry overly about breaking the testing API on a stable line (though that doesn't mean we should get crazy). The most likely result is that we'll need to back-/forward-port this to other lines, but so be it. Merging.

PR #3415, opened 2015-01-29
CLI web cleanup

sbesson:
This PR does some basic cleanup to the `WebControl` class:

- adds internal `_get_python_dir()` and `_get_templates_dir()` method
- uses monkeypatching instead of defining hidden arguments for testing the templates generation in unit tests
- cleans up the deprecated selenium library/selenium tests and selenium subcommand
- runs `self.clearsessions()` at every start of OMERO.web

--no-rebase

PR #3699, opened 2015-04-09
Clearsessions message

sbesson:
Reported by @kennethgillen during a production server upgrade to 5.1.0. With https://github.com/openmicroscopy/openmicroscopy/pull/3415, the start of OMERO.web now automatically performs a cleanup of the expired Django sessions. This step can be time consuming especially if the number of stale sessions is large. Additionally, there is no logging information for system administrators which may be misinterpreted as the inability for OMERO.web to start.

This PR:
- adds some basic logging information for the sessions clean-up step. Because it is fully delegated to Django there does not seem to be an easy way to have either a time estimate or a progression status /cc @aleksandra-tarkowska ?
- adds an optional `--keep-sessions` flag to skip the expired sessions clean-up in `start`, `restart` and `iis` subcommands.
- add an optional `--no-wait` flag to use `popen` rather than `call` when calling `clearsessions` in `start`, `restart` and `iis` subcommands to prevent hanging the Web restart

/cc @chris-allan

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/web.py
+++ b/components/tools/OmeroPy/src/omero/plugins/web.py
@@ -106,12 +106,15 @@ class WebControl(BaseControl):
             "Advanced use: Creates needed symlinks for static"
             " media files (Performed automatically by 'start')")
 
-        parser.add(
+        clearsessions = parser.add(
             sub, self.clearsessions,
             "Advanced use: Can be run as a cron job or directly to clean "
             "out expired sessions.\n See "
             "https://docs.djangoproject.com/en/1.6/topics/http/sessions/"
             "#clearing-the-session-store for more information.")
+        clearsessions.add_argument(
+            "--no-wait", action="store_true",
+            help="Do not wait on expired sessions clean-up")
 
         #
         # Developer
@@ -315,23 +318,23 @@ class WebControl(BaseControl):
         if rv != 0:
             self.ctx.die(607, "Failed to collect static content.\n")
 
-    def clearsessions(self, args, wait=True):
+    def clearsessions(self, args):
         """Clean out expired sessions."""
         self.ctx.out("Clearing expired sessions. This may take some time... ")
         location = self._get_python_dir() / "omeroweb"
-        args = [sys.executable, "manage.py", "clearsessions"]
-        if wait:
-            rv = self.ctx.call(args, cwd=location)
+        cmd = [sys.executable, "manage.py", "clearsessions"]
+        if not args.no_wait:
+            rv = self.ctx.call(cmd, cwd=location)
             if rv != 0:
                 self.ctx.die(607, "Failed to clear sessions.\n")
             self.ctx.out("[OK]")
         else:
-            self.ctx.popen(args, cwd=location)
+            self.ctx.popen(cmd, cwd=location)
 
     def start(self, args):
         self.collectstatic()
         if not args.keep_sessions:
-            self.clearsessions(args, wait=args.no_wait)
+            self.clearsessions(args)
         import omeroweb.settings as settings
         link = ("%s:%s" % (settings.APPLICATION_SERVER_HOST,
                            settings.APPLICATION_SERVER_PORT))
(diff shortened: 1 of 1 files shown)


manics:
Is it possible to run `clearsessions` in the background if it's called from `web start`?

sbesson:
@manics: that would mean using `self.ctx.popen` as opposed to `self.ctx.call` in the context of the server start, right? I cannot comment on the potential implications of such an change, but it sounds sensible a priori.

aleksandra-tarkowska:
I presume you want both seb. If you run ```omero web clearsessions``` it should be ```call```, but if you run ```omero web start``` it should be ```popen```, does it make sense?

sbesson:
@aleksandra-tarkowska: agreed with regard to the double usage, that's certainly what I planned to do. What I am still unclear about is whether it is okay to trigger concurrently `manage.py runfcgi` and `manage.py clearsessions` in production with virtually no control over the termination of the second thread? Can that lead to issues or corruptions for existing sessions?

manics:
No-ones ever said you can't run `clearsessions` whislt OMERO.web is running, and the conclusion for 5.1 was that we should recommend running `clearsessions` in a cron job. If there's potential for corruption it'll need revising.

sbesson:
Pushed.

components/tools/OmeroPy/src/omero/plugins/web.py
```diff
@@ -57,14 +57,21 @@ def _configure(self, parser):
...
+        restart = parser.add(
+            sub, self.restart, "Restart the OMERO.web server")
         parser.add(sub, self.status, "Status for the OMERO.web server")
 
         iis = parser.add(sub, self.iis, "IIS (un-)install of OMERO.web ")
         iis.add_argument("--remove", action="store_true", default=False)
 
+        for x in (start, restart, iis):
+            x.add_argument(
+                "--skip-clear-sessions", action="store_true",
```
joshmoore: "keep-sessions"?

sbesson: Also works. That being said with the improvements made following the suggestions of @aleksandra-tarkowska/@manics (i.e. clearing sessions using `popen` rather than `call`), maybe we do not need this option at all?

joshmoore:
For testing purposes, could you push a flag to allow "no-wait" for `web clearsessions`?

components/tools/OmeroPy/src/omero/plugins/web.py
joshmoore: Likely better to keep this, in case web start/stop starts to cause problems.

sbesson:
So `bin/omero web restart [--keep-sessions | --no-wait]` ?

sbesson:
Pushed and PR description updated

joshmoore:
Makes sense but I'd still add a no-wait to `clearsessions`.

sbesson:
With the last versions, on UNIX, the following subcommand should be tested

```
bin/omero web clearsessions
bin/omero web clearsessions --no-wait
bin/omero web restart
bin/omero web restart --no-wait
bin/omero web restart --keep-sessions
```

joshmoore:
Tested on merge-deploy and merge-integration, e.g.:
```
[hudson@ome-c6100-3 tmp]$ ls -ltra sessionid_* | wc
    163    1467   15911
[hudson@ome-c6100-3 tmp]$ ls -ltra sessionid_integration*|wc
    115    1035   11385
[hudson@ome-c6100-3 tmp]$ time /opt/hudson/workspace/OMERO-5.1-merge-integration/src/dist/bin/omero web clearsessions
Clearing expired sessions. This may take some time...
[OK]

real    0m3.032s
user    0m1.216s
sys     0m0.226s
[hudson@ome-c6100-3 tmp]$ ls -ltra sessionid_integration*|wc
     79     711    7821
[hudson@ome-c6100-3 tmp]$ time /opt/hudson/workspace/OMERO-5.1-merge-integration/src/dist/bin/omero web clearsessions --no-wait
Clearing expired sessions. This may take some time...

real    0m0.665s
user    0m0.596s
sys     0m0.065s
[hudson@ome-c6100-3 tmp]$ ls -ltra sessionid_integration*|wc
     79     711    7821
```
LGTM (despite the multiple calls to `add_mutually_exclusive`.... :smile:)

sbesson:
I can push another commit for `add_mutually_exclusive` if you prefer.

PR #3702, opened 2015-04-10
CLI session unit tests

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-April/005225.html

This PR does not modify the existing behavior/limitations of the CLI sessions rejoining logic but extends the unit test coverage for session key-based connection in preparation of future improvements/bug fixes. 

Summary of main changes:
- unit tests are added to cover multiple scenarios of session re-attachments (with/without port/group properties). The stored/passed properties are stored as tuples of the test class for clarity.
- the error message in the case of conflicting properties is reviewed and promoted to `self.ctx.err()` level so that it appears more systematically for debugging purposes /cc @PaulVanSchayck

To test this PR:
- check all the unit tests are passing
- try a conflicting property scenario for session rejoining and check the conflicting message is displayed by default and is sensible

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/sessions.py
+++ b/components/tools/OmeroPy/src/omero/util/sessions.py
@@ -127,7 +127,9 @@ class SessionsStore(object):
                     new == str(omero.constants.GLACIER2PORT)):
                 continue
             elif old != new:
-                conflicts += (key + (":%s!=%s;" % (old, new)))
+                if conflicts != "":
+                    conflicts += "; "
+                conflicts += (key + (": %s!=%s" % (old, new)))
         return conflicts
 
     def remove(self, host, name, uuid):
--- a/components/tools/OmeroPy/src/omero/util/sessions.py
+++ b/components/tools/OmeroPy/src/omero/util/sessions.py
@@ -129,7 +129,7 @@ class SessionsStore(object):
             elif old != new:
                 if conflicts != "":
                     conflicts += "; "
-                conflicts += (key + (": %s!=%s" % (old, new)))
+                conflicts += "%s: %s!=%s" % (key, old, new)
         return conflicts
 
     def remove(self, host, name, uuid):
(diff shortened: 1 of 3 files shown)


components/tools/OmeroPy/src/omero/util/sessions.py
```diff
@@ -127,7 +127,9 @@ def conflicts(self, host, name, id, new_props, ignore_nulls=False):
                     new == str(omero.constants.GLACIER2PORT)):
                 continue
             elif old != new:
-                conflicts += (key + (":%s!=%s;" % (old, new)))
+                if conflicts != "":
+                    conflicts += "; "
+                conflicts += (key + (": %s!=%s" % (old, new)))
```
manics: Very nit-picky, but since you're modifying this line `("%s: %s!=%s" % (key, old, new))` would be clearer

sbesson: Can do. I have been considering moving to a multi-line instead of semicolon separation. Thoughts?

ximenesuk:
This looks good. As a side-issue the pattern:
`@pytest.mark.parametrize('port', [None, 4064, 14064])`
occurs quite a few times, is it worth defining a `PORTS` or `DEFAULT_PORTS` list constant?

ximenesuk:
This looks good to merge.

The last commit did get me to muse on getting individual elements from tuples and wondered if,
```
name = cli.STORE.get_current()[1]
assert name != key
```
is any more transparent than
```
current = cli.STORE.get_current()
assert current[1] != key
```
but it's probably six and two threes.

sbesson:
Happy to replace the last commit by the first suggestion as part of my next CLI sessions PR.

joshmoore:
This goes back to my comment about not returning tuples but always objects. Perhaps named tuples which should be backwards compatible? In any event, merging.

PR #3703, opened 2015-04-10
Error NPE

jburel:
Fix NPE when message is null.

To test:
 * try to import flex file from QA : 9456, It will not import (this is expected see https://trac.openmicroscopy.org/ome/ticket/12529)
 * Check that no error dialog pops up.

If you can test with build from Friday to compare

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/ImportException.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/ImportException.java
@@ -177,19 +177,20 @@ public class ImportException
             return COMPRESSION;
         } else if (cause instanceof FormatException) {
             String message = cause.getMessage();
-            if (message.contains(MISSING_LIBRARY_TEXT.toLowerCase()))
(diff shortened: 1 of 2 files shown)


dominikl:
Could confirm, Friday's build is crashing with a NPE. Import of `imagebad.flex` doesn't crash today's build, but I also can't view the exception or submit the exception (clicking on view or submit has no effect).

jburel:
You cannot view any exception today b/c it has been fixed in https://github.com/openmicroscopy/bioformats/pull/1722

jburel:
one more commit to come on that PR
when imagebad is selected, no container detected. error should be handled earlier.
i.e. if you are in the P/D view a dataset is created (if new from folder)

jburel:
@dominikl: to test my last commit. 
 * Select imagebad.flex. 
 * Click import. Make sure you are in the P/D view and "New from folder is selected"
 * Check that no dataset is created and a file not valid message is displayed.

dominikl:

PR #3704, opened 2015-04-10
OMERO Java Gateway

dominikl:
First step for creating an Insight independent Java Gateway. Moves all relevant classes from Insight to omero.gateway package within the blitz.jar library. Also modifies Insight in the way, that it already uses the omero.gateway.Gateway for all connection related stuff, i. e. Insight won't deal anymore with creating  sessions, initiating/closing services, etc. that's now done within the omero.gateway.Gateway.

This PR should be tested (and merged if approved) quickly, because with ongoing development it will soon be again impossible to rebase it. Moving functionality from Insight's Gateway to the general Java Gateway can be done in follow up PRs, which will be far less invasive, i. e. can easily be done parallel to develop branch.

Should this go into breaking build first?

/cc @jburel 

Note: The code is still quite 'messy'. Will do some refactoring, cleanup and proper documentation, after it's decided to go ahead with this approach at all.

Example code for using the Gateway see [GatewayExample.java](https://github.com/dominikl/openmicroscopy/blob/yet_another_gateway_attempt/components/blitz/src/omero/gateway/GatewayExample.java)


jburel:
Let's put it in breaking so we do not have any surprise

mtbc:
Should `Requests.java` also move?

dominikl:
@mtbc You're right. ALso moved `Requests.java` to blitz.

mtbc:
Once we get breaking jobs fixed again next week, maybe we can again rebase and pull this out of breaking and see how it goes in the general merge build.

mtbc:
Breaking failures today are just a couple of lock timeouts for tests that passed on Friday.

jburel:
Are you planning to add to ```Browse Facility``` methods like getProjects etc.

dominikl:
Yes, I'll try to add the same DataObject retrieval methods like the matlab interface provides.

dominikl:
Looking at my last commit, I wonder if `RenderingFacility` is actually a proper name/place for the `getPlane` method; as it provides access to the raw pixels value, there is actually no rendering involved. What do you think about the naming here @jburel ?

jburel:
@dominikl: the term rendering is a bit confusing. RawDataFacility might be more suitable
We could potentially add there method to access pixels data and general files.

hflynn:
Note that there's a link to Requests.java in the OMERO developer docs (developers/Server/GraphsMigration.txt) which will need updating once this is sorted. Will other developer docs changes be needed too?

dominikl:
@hflynn That's likely. This PR moves a bunch of classes from Insight to Blitz package. Is there an easy way how to find out which classes are directly referenced from the documentation?

hflynn:
Any broken links in the documentation show up in https://ci.openmicroscopy.org/view/Docs/job/OMERO-5.1-merge-docs/

jburel:
This PR is to be rebased on decoupling when the decoupling branch is stable.

PR #3705, opened 2015-04-13
Metadata panel followup

dominikl:
Fixes some issues with the metadata panel:
- Use same font style for "creation date" label as for "acquisition date", "creation date", etc. **Test: Check font style of the "creation date" label**
- Show "creation date" also for plate runs (PlateAcquisitions) **Test: Make sure "creation date" label is shown for plate runs**
- FilesetInfoDialog (accessed via the "../../" icon) now shows the real inplace import option used (ln or ln_s) **Test: Import an image with `--transfer=ln_s` and one with `transfer=ln` and check that these options are shown in the FilesetInfoDialog**
- Reduced the linebreak threshold for channel names; long channel names and/or images with many channels made the channel names label quite big, so you heavily had to use the horizontal scrollbar. **Test: Open an image with lots of channels or one with very long channel names; make sure this does not spoil the metadata panel layout**
- Wrap long comments regardless of whitespaces; if a comment had a very long text string without whitespace, it was not broken into several lines and like the previous point, caused the appearance of the very long horizontal scrollbar. **Test: Add a very long text string (without white spaces) as comment; make sure this does not spoil the metadata panel layout**
- Avoid showing acquisition date "1970-01-01" if timestamp is zero. **Test: Open an image which has an acquisition date, but with value zero; make sure acquisition date label is not shown; unfortunately I can't remember which file I used for testing. Is it possible to create an artificial one?**

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
@@ -4471,13 +4471,25 @@ class EditorModel
     }
     
     /**
-     * Returns <code>true</code> if the file is an inplace import <code>false</code> otherwise.
+     * Returns the {@link ImportType}
      * @return See above.
      */
-    boolean isInplaceImport() {
-    	StructuredDataResults data = parent.getStructuredData();
-		if (data == null) return false;
-    	return CollectionUtils.isNotEmpty(data.getTransferLinks());
+    ImportType getImportType() {
+        StructuredDataResults data = parent.getStructuredData();
+        if (data != null) {
+            Collection<AnnotationData> tfl = data.getTransferLinks();
+            if (tfl != null) {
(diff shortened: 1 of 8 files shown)


mtbc:
I could try to zero an existing acquisition date for some image on trout if that helps.

dominikl:
I remember the way I produced such an image: Export an image with acquisition date to ome.tiff with version 2010-04 this will result in loosing the acquisition date, respectively set it to zero. Reimport the image, and make sure the acquisition is not shown.

bramalingam:
Except for one small bug : The creation Date having a different font stilt (All in Bold) to that of Acquisition data and etc, everything else works as expected.

dominikl:
Fixed the font style issue. Review: Just last PR, i. e. select a project or dataset, check that the font style for "Creation Date" matches the font style of "Acquisition Date" etc. if an image is selected.

pwalczysko:
Works as expected.

PR #3706, opened 2015-04-13
Edit root user

will-moore:
See: http://trac.openmicroscopy.org/ome/ticket/12821

When another admin is editing 'root' user, the "Active" checkbox is disabled (since you can't disabled 'root'). But in the form handling, the disabled checkbox is ignored, so this was trying to disable 'root' and gives a security violation.
Now we manually check if we're editing 'root' and don't try to disable if we are.
To test:
 - Log in as an admin (not 'root') and make any change to root user and submit the form. Should see no exception. Check save has worked.


aleksandra-tarkowska:
Works as expected, good to merge

PR #3708, opened 2015-04-13
remove concurrent hashmap

mtbc:
ConcurrentHashMap undergoes API changes among Java 6, 7, 8, so to keep code nicely portable it's generally easier to simply avoid using it given that we already have Guava available and that offers a suitable substitute.

Testing: No tests or builds should break, and the OMERO source should no longer contain direct usage of ConcurrentHashMap. (It has already been removed from Bio-Formats.) Import should still work, and management of Insight sessions / connectors should still be fine; @jburel may have specific testing ideas there. (I guess, try logging in and out and switching user and whatnot.)

To avoid regressions, the less-trivial code changes should be reviewed by eye.

--no-rebase

### DIFF
--- a/components/blitz/src/ome/services/blitz/fire/SessionManagerI.java
+++ b/components/blitz/src/ome/services/blitz/fire/SessionManagerI.java
@@ -9,7 +9,6 @@ package ome.services.blitz.fire;
 
 import java.util.List;
 import java.util.Set;
-import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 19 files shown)


pwalczysko:
Tested import via insight (switching users and Import As) and command line import. All works fine.

dominikl:
Looks fine to me. My only worry is, by using `Cache` instead of `Map`, is it not possible that entries are automatically removed by the caching/eviction algorithm without the proper 'termination' methods being called? E. g. the `derived` connector cache in `Connector.java`, if derived connectors are removed from the cache, their `close()` method never gets called, leaving sessions open.

joshmoore:
I'd assume http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/cache/CacheBuilder.html#newBuilder() holds:

```
Constructs a new CacheBuilder instance with default settings, including strong keys, strong values, and no automatic eviction of any kind.
```

dominikl:
Alright, missed that part of the documentation. Should be fine then!

joshmoore:
This has been in for a while with no issues as far as I can tell. We may eventually want to include the discussions here in a "policy doc" so that it's clear where we should use what, but that may take some more discussing/implementing. Thanks, @mtbc

PR #3710, opened 2015-04-13
share_preview_panel_12805

will-moore:
Fixes exception when browsing share image preview panel for incomplete images, see: https://trac.openmicroscopy.org/ome/ticket/12805

To test:
 - log in to trout as 'root' and browse to public shares.
 - Browse share '1401', select image and open Preview panel.
 - Image will be incomplete (viewer won't work) but you shouldn't see any exception/crash


aleksandra-tarkowska:
works as expected. no crash. good to merge

PR #3712, opened 2015-04-14
Mouse wheel Z-frame scrolling

dominikl:
Fixes [Ticket 12829](http://trac.openmicroscopy.org.uk/ome/ticket/12829)

Test: Open an image (not a *big* image) with multiple Z frames. Make sure the mouse wheel scrolling through the Z frames works.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BirdEyeViewComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BirdEyeViewComponent.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.imviewer.browser.BirdEyeView 
  *
  *------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mtbc:
Good to merge.

will-moore:
@dominikl Certainly the mouse-wheel scrolling works, but it sometimes also zooms the image. I'm not sure exactly what triggers this, but if I scroll up and down repeatedly, I usually end up at a higher zoom level. This happens most reliably at max Z when all scrolling results in zoom, but also can happen at anyt other Z-index too. Also, clicking and dragging of the image with the mouse still changes the z-index instead of panning, but maybe you meant to fix that in a different PR?

mtbc:
Gah, sorry I missed that misfeature.

dominikl:
@will-moore I can't reproduce the mixed scroll/zoom behaviour. There's a difference between small and big images. Small images: Scroll and drag change Z frame, no zoom; big images: scroll zooms in/out, drag moves the image around. I think that's intended.

PR #3714, opened 2015-04-14
Fix image simple marshal

chris-allan:
Fix a typo bug found in `ImageWrapper.simpleMarshal()` during performance testing and improvement. Unit tests included.

/cc @jballanc

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -6982,7 +6982,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                     rv['tiled'] = ((rv['size']['height'] * rv['size']['width'])
                                    > (maxplanesize[0] * maxplanesize[1]))
                 else:
-                    rv['tiles'] = False
+                    rv['tiled'] = False
 
         return rv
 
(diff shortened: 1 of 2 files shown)


joshmoore:
Kicked travis (HDF5 error)

will-moore:
I wonder if we should align ```ImageWrapper.simpleMarshal()``` which uses ```tiled``` and ```webgateway.marshal.py imageMarshal()``` which uses ```tiles```? Seems a little confusing to have the different keys.
They also use different methods to determine if they are tiled images, using image size in Blitz gateway but using ```re.requiresPixelsPyramid()``` in webgateway, but there may be good reasons for this and it's not so related to this PR.

joshmoore:
@will-moore: good to merge, then? If so, can you create a card to track any investigation/refactoring that you want to see happen re: `imageMarshal`?

will-moore:
Yes, good to merge.

will-moore:
@chris-allan Where is ```ImageWrapper.simpleMarshal()``` used? What would be the effect of changing ```tiled``` to ```tiles``` for this?

chris-allan:
Its main function as far as I'm aware is as part of the marshaling core that is used as part of various JSON calls. In working on #3695 we never investigated this more thoroughly, since we were basically removing singular usage of wrapper sub-class marshaling, but that could certainly be done.

PR #3688, opened 2015-04-06
metadata panel and share

aleksandra-tarkowska:
Ticket: https://trac.openmicroscopy.org/ome/ticket/12787
This should resolve issues related to all 3 general, acquisition and preview panels as well as viewer.

As share offers limited functionality that is most what we can offer in that specific context.
To test check if you are satisfying with:
 - metadata in general pane + download as JPEG, PNG, TIFF
 - metadata in acquisition pane.
 - rendering in preview
   - make sure save buttons never became active
   - test with image that has multiple user settings, try too apply them (you won't be able to save)
 - rendering in full viewer

cc: @pwalczysko

PR #3715, opened 2015-04-14
Use allow=0 for share post-processing (Fix #12562)

joshmoore:
The no-op method in SharingACLVoter was incorrect in
that by not calling copyRestrictions, there were *no*
restrictions on the object as opposed to nothing being
allowed. Now, all should return False in the context
of a share.

Testing: This should keep the same behavior as gh-3688
but without @aleksandra-tarkowska's workaround.

### DIFF
--- a/components/server/src/ome/security/sharing/SharingACLVoter.java
+++ b/components/server/src/ome/security/sharing/SharingACLVoter.java
@@ -13,15 +13,16 @@ import ome.api.IShare;
 import ome.conditions.SecurityViolation;
 import ome.model.IObject;
 import ome.model.internal.Details;
+import ome.model.internal.Permissions;
 import ome.security.ACLVoter;
 import ome.security.SystemTypes;
 import ome.security.basic.CurrentDetails;
 import ome.security.basic.TokenHolder;
 import ome.services.sharing.ShareStore;
 
+import org.hibernate.Session;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.hibernate.Session;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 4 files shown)


joshmoore:
Force pushed and cherry-picked at @aleksandra-tarkowska's request.

aleksandra-tarkowska:
**To test** log in as user who has data in *read-ann* group. Choose image(s) from *read-ann* group and share with user who is a member of private gorup only (for example *user-1*). Log in as *user-1* and check if there is no edit button on any field like name,description, etc in metadata - general pane.

sbesson:
Is there a corresponding `test_ishare` integration test for the scenario you just described?

aleksandra-tarkowska:
@sbesson thanks, perhaps good idea to add

pwalczysko:
Tested taking the lead from https://github.com/openmicroscopy/openmicroscopy/pull/3715#issuecomment-93419681 
   - the behaviour in Share is as described in Ola's comment https://github.com/openmicroscopy/openmicroscopy/pull/3715#issuecomment-93419681
   - the behaviour when not in share, but in p, ro, ra, rw groups is a bit different thatn described by Ola's comment, but this is because the comment is not precise, not because the behaviour is unexpected
    - when the behaviour of CanEdit was tested according to https://docs.google.com/a/openmicroscopy.org/spreadsheets/d/1gx7KX-putW9_HZ3XVpFFYRVB2v22B6WTcauRCHrlmKE/edit#gid=87647062 and compared with trout latest, the behavour was as expected, and in accordance with trout  latest

The tests described in this comment above all passed.

joshmoore:
Thanks, @pwalczysko! I'll leave this for tomorrow so we can see the automated test results.

joshmoore:
Based on https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-web/306/testReport/test.integration.test_share/TestShare/, merging.

sbesson:
Makes sense to me. My only nitpicky comment/question is whether the new share integration test truly belongs to `OmeroWeb` as it is only using/testing the Python gateway (no django_client). If not, we can move it to `OmeroPy` in a cleanup PR.

joshmoore:
Ah, sorry, @sbesson. Are there any other tidy ups planned, or shall I open a PR?

sbesson:
Assuming https://github.com/openmicroscopy/openmicroscopy/pull/3728 is accepted, `test_ishare` will need to be refactored` to use the new helper methods systematically and the tidy-up can happen there. If the helper objects PR is not worth it, we can decide to open a tidy-up PR.

PR #3717, opened 2015-04-15
Disable scripts menu in share page

will-moore:
See https://trac.openmicroscopy.org/ome/ticket/12832

Disables script menu for share page. Although the image owner *may* be able to run scripts on their own images in share, the majority of times it will be users who only have access to the images via share permissions, so the script won't be able to access the images.

To test:
 - Go to public share page.
 - The script menu button should look disabled and clicking on it won't launch script menu. Tooltip will tell you it's disabled in shares.


pwalczysko:
Works as expected.

PR #3718, opened 2015-04-15
Preserve Z/T changes

dominikl:
Fixes [Ticket 12709](http://trac.openmicroscopy.org.uk/ome/ticket/12709) and also some occasionally appearing problems that thumbnails are not in sync with the rendering settings (unfortunately mixed that altogether in one commit).

Note: With this PR **Z/T changes are silently stored**, i. e. the user's last selected Z/T frame will be stored in the rendering settings automatically as soon as he moves away from the image (selects a different image, close the image viewer, or just switch from the preview pane to the general pane). Whereas 'real' rendering setting changes (to the channels) are discarded, when moving away from the preview panel.
Is this an acceptable behaviour? Can't treat Z/T changes as 'real' rendering setting changes, because this leads to whole lot of inconsistencies with respect to undo/redo, save, copy/paste, etc.

**Test:**
For the scope of the ticket:
- Select an image with multiple Z frames. Change the currently viewed Z frame in the preview pane, switch to the general pane (at this point the current Z frame should be stored and also the thumbnail updating), open the SplitViewFigure dialog (make sure your selected Z frame is shown), create the figure (make sure the figure also shows your selected Z frame).
- Do the same again but now modifiy the Z frame in the full image viewer.

General:
- Ideally some more rendering settings workflows (copy/paste, undo/redo, etc.) should be tested as well.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
@@ -2718,7 +2718,7 @@ class EditorModel
                     }
                 } catch (Exception e) {
                     MetadataViewerAgent.getRegistry().getLogger()
-                            .warn(this, "Could not save renderign settings");
+                            .warn(this, "Could not save rendering settings");
                 }
                 renderer.discard();
                 renderer = null;
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
@@ -96,6 +96,7 @@ import org.openmicroscopy.shoola.env.data.model.SaveAsParam;
 import org.openmicroscopy.shoola.env.data.model.ScriptObject;
 import org.openmicroscopy.shoola.env.data.util.SecurityContext;
 import org.openmicroscopy.shoola.env.data.util.StructuredDataResults;
+import org.openmicroscopy.shoola.env.event.ReloadThumbsEvent;
 import org.openmicroscopy.shoola.env.log.LogMessage;
 import org.openmicroscopy.shoola.env.rnd.RenderingControl;
 import org.openmicroscopy.shoola.env.ui.UserNotifier;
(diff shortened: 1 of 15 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
```diff
@@ -2701,10 +2702,27 @@ void setRootObject(Object refObject)
...
+                        renderer.saveCurrentSettings();
+                        MetadataViewerAgent
+                                .getRegistry()
+                                .getEventBus()
+                                .post(new ReloadThumbsEvent(renderer
+                                        .getRefImage().getId()));
+                    }
+                } catch (Exception e) {
+                    MetadataViewerAgent.getRegistry().getLogger()
+                            .warn(this, "Could not save renderign settings");
```
mtbc: rendering

dominikl:
Forgot to /cc @gusferguson and @jburel on this, as this would be a noticeable change in user experience.

will-moore:
Seems to be working fine with the Split-view figure always respecting the latest Z-section.
Also, the general rendering settings behaviour is unchanged although I haven't done a full test of every scenario.
I think I'm OK with the auto-save of Z-section, but agree that @gusferguson and @jburel should probably  check that this doesn't break any other workflows, since it is a significant change.

gusferguson:
@dominikl @will-moore @jburel @pwalczysko 

Tested using OMERO.insight-5.1.0-452-936e5ef-ice35-b481-mac_Java7+ trout user-3 read-annotate-1.

Behaves as expected and split-view figure uses currently viewed z-section.

This is quite a substantial change in behaviour as it effectively makes the Z-section selection a "super-rendering setting" i.e. the only rendering setting modality that is saved automatically when changed.

I would break it down as follows:

- Split View Figure uses selected Z section - essential
- thumbnail updates to show selected Z-section - expected
- viewer opens with Z-section selected in Preview displayed - intuitive
- auto save of Z-section section selection to rendering settings i.e. persists beyond that view/session - not so intuitive
- no ability to actively Save Z-section selection - as exists for other rendering settings changes - dissonant
- reset to Imported resets Z-section to default - good
- what do we do with T-points - the same thing or leave them as they are?

I can't see any other workflows that are affected by this change. but it will cause a Insight-Web discrepancy that would need to be addressed before release in my opinion.

What are y'all's thoughts?

jburel:
Thanks @gusferguson for the report.
Major changes in behaviour and we do not want to introduce discrepancy between Web/insight.
Further discussions required. 
moving it to breaking

jburel:
Added card. To be discussed during next client meeting
https://trello.com/c/3H67wdIJ/337-viewer-mouse-wheel-and-save-rendering

jburel:
Comment added to card. @dominikl to look at it on his return.

dominikl:
Closing. Will open a new PR for just enabling the save button on z/t change.

PR #3720, opened 2015-04-15
CLI sessions: do not check group while rejoining session with key

sbesson:
Follow up of https://github.com/openmicroscopy/openmicroscopy/pull/3702 and http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-April/005245.html.

This PR fixes a first bug where rejoining a stored session file using the session key was restricted to using the same connection arguments (including the initial connection group). Main changes of this PR are:
- fixing the port assymmetry (i.e. session file stored with the default port or without any omero.port can be rejoined without passing the port/by passing `-p 4064`)
- remove the group check restriction when joining an existing session stored on disk using the session key
- unit tests are modified to cover these bug fixes
- the unit tests when reusing session (i.e. without key) are also extended to cover all the group/port scenarios. In this case, the property check restriction (i.e. `omero.group` needs to match the content of the stored session file) is maintained for now.

/cc @PaulVanSchayck

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -260,10 +260,12 @@ class SessionsControl(BaseControl):
         pasw = args.password
         if args.key:
             if name and not self.ctx.isquiet:
-                self.ctx.err("Overriding name since session set")
+                self.ctx.err("Overriding name since session key set")
             name = args.key
+            if args.group and not self.ctx.isquiet:
+                self.ctx.err("Ignoring group since session key set")
             if args.password and not self.ctx.isquiet:
-                self.ctx.err("Ignoring password since key set")
+                self.ctx.err("Ignoring password since session key set")
             pasw = args.key
         #
         # If no key provided, then we check the last used connection
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -344,7 +344,7 @@ class SessionsControl(BaseControl):
                                  False, set_current=False)
             else:
                 rv = self.check_and_attach(store, server, stored_name,
-                                           args.key, props)
+                                           args.key, props, check_group=False)
             action = "Joined"
             if not rv:
                 if port:
@@ -355,7 +355,8 @@ class SessionsControl(BaseControl):
         elif not create:
             available = store.available(server, name)
             for uuid in available:
-                rv = self.check_and_attach(store, server, name, uuid, props)
+                rv = self.check_and_attach(store, server, name, uuid, props,
+                                           check_group=True)
                 action = "Reconnected to"
 
         if not rv:
@@ -401,7 +402,8 @@ class SessionsControl(BaseControl):
 
         return self.handle(rv, action)
 
-    def check_and_attach(self, store, server, name, uuid, props):
+    def check_and_attach(self, store, server, name, uuid, props,
+                         check_group=False):
         """
         Checks for conflicts in the settings for this session,
         and if there are none, then attempts an "attach()". If
@@ -411,7 +413,8 @@ class SessionsControl(BaseControl):
         exists = store.exists(server, name, uuid)
 
         if exists:
-            conflicts = store.conflicts(server, name, uuid, props)
+            conflicts = store.conflicts(server, name, uuid, props,
+                                        check_group=check_group)
             if conflicts:
                 self.ctx.err(
                     "Failed to join session %s due to property conflicts: %s"
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -418,9 +418,12 @@ class SessionsControl(BaseControl):
             conflicts = store.conflicts(server, name, uuid, props,
                                         check_group=check_group)
             if conflicts:
-                self.ctx.err(
-                    "Failed to join session %s due to property conflicts: %s"
-                    % (uuid, conflicts))
+                if "omero.port" in conflicts:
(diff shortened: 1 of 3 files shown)


sbesson:
Note: the current state of this PR may not cover (yet) the scenario described in http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-April/005245.html as the CLI may be rejoining a non-locally stored session via the https://github.com/openmicroscopy/openmicroscopy/blob/v5.1.0/components/tools/OmeroPy/src/omero/plugins/sessions.py#L343 mechanism.

joshmoore:
* [x] Port re-use
 * [x] Group check restrictions __*if*__ the assumption is that passing `-g` will be ignored when using `-k` 

One minor issue: If I *try* to login to 24064, then I don't care if we don't use a file with port 4064:
```
$ omero login root@localhost -p 24064 -w ome
Failed to join session d0ed7ffc-c951-449e-89e0-948b4cc89c58 due to property conflicts: omero.port: 4064!=24064
```

sbesson:
Interesting. Does that mean the change in https://github.com/openmicroscopy/openmicroscopy/commit/d8af55285d5eaca03aa54906d97be51eaec59b65 should be context sensitive?

joshmoore:
Perhaps it's just the psychological difference between "skipping" and "failing"? Not sure. Guess we should consider what will the user actually wants to deal with. Based on the output of `sessions list`, probably only a mismatch in group will be surprising to the user since a localhost/port pair is clearly "something else".

sbesson:
Yes I certainly agree `omero.port` and `omero.group` should be someway treated differently (this PR is actually separating them). Indeed, `omero.port` should be classified together with `hostname`. Will try to push an extra commit or two.

sbesson:
Last 2 commits should:
- add early warnings if group is passed together with session key while logging
- different mismatching ports conflicts (logged at debug level and invisible most of the time) from other properties (group) conflicts

joshmoore:
On the port front, :thumbsup: :
```
$ omero sessions clear
2 session(s) cleared
$ omero login root@localhost -w omero
Previous session expired for root on localhost:4064
Created session 5f6915f1-34ee-4fe3-927e-7d045fae95fc (root@localhost:4064). Idle timeout: 10.0 min. Current group: system
$ omero login root@localhost:24064 -w omero
Created session 19d63971-fc1d-43ea-abbb-cea874ea7ba9 (root@localhost:24064). Idle timeout: 10.0 min. Current group: system
```

On the groups side, certainly better but:
```
$ omero login root@localhost:24064 -w omero
Created session 19d63971-fc1d-43ea-abbb-cea874ea7ba9 (root@localhost:24064). Idle timeout: 10.0 min. Current group: system
$ omero login root@localhost:24064 -w omero -g system
Previously logged in to localhost:24064 as root
Skipped session 19d63971-fc1d-43ea-abbb-cea874ea7ba9 due to property conflicts: omero.group: None!=system   # <========= Surprised me.
Created session 8730f9f4-fb81-4fa7-bf68-78bdb5d5dc06 (root@localhost:24064). Idle timeout: 10.0 min. Current group: system
$ omero login root@localhost:24064 -w omero -g system
Using session 8730f9f4-fb81-4fa7-bf68-78bdb5d5dc06 (root@localhost:24064). Idle timeout: 10.0 min. Current group: system
$ omero login root@localhost:24064 -w omero -g testAnnotationPermissions
Previously logged in to localhost:24064 as root
Skipped session 8730f9f4-fb81-4fa7-bf68-78bdb5d5dc06 due to property conflicts: omero.group: system!=testAnnotationPermissions
Skipped session 19d63971-fc1d-43ea-abbb-cea874ea7ba9 due to property conflicts: omero.group: None!=testAnnotationPermissions
Created session c1d03f44-58be-43ef-9690-82bb28c0b05c (root@localhost:24064). Idle timeout: 10.0 min. Current group: testAnnotationPermissions
```
There was one surprise the first time I passed a group, since effectively they *did* match, but I hadn't requested one the first time. So the port logic has it easy since it uses `DEFAULTGLACIERPORT`. With group, that's a runtime default. Not sure if we can do anything here.

sbesson:
Yes this is up for debate. 
The first scenario (no group stored locally, group specified in arguments) is very hard to handle as the expected outcome is ambiguous, i.e.:
- either join an existing session using its current group (like session key), i.e. silently drop the group option
- or join an existing session and modify the group context after joining. This can be extremely dangerous as we do not know what else is consuming the session.
- or create a new session with the group passed as an argument, i.e. preserve the current conservative statu quo

Did you try the reverse case (group existing, rejoining without group argument)? In that case I would expect a fresh session to be recreated in the current state. Arguably we could also add flexibility here by making user of the `ignore_nulls` option of `conflicts`. Thoughts?

joshmoore:
> Did you try the reverse case...

Looks like it doesn't create a new session:
```
$ omero login -g testAnnotationPermissions root@localhost:24064 -w omero
Previous session expired for root on localhost:24064
Created session 67ff6931-c082-4fca-8dbb-6b9ae947c050 (root@localhost:24064). Idle timeout: 10.0 min. Current group: testAnnotationPermissions
$ omero login root@localhost:24064 -w omero
Using session 67ff6931-c082-4fca-8dbb-6b9ae947c050 (root@localhost:24064). Idle timeout: 10.0 min. Current group: testAnnotationPermissions
```

sbesson:
* shrugs * the log message "Using" vs "Reconnected to" actually indicates we are testing another part of the sessions logic here, namely when the login arguments match the last stored host/user. Will look at it to see if we can unify.

sbesson:
Unless you want the latter case to be covered as part of this PR or you find any blocker, I propose we get this merged. I will review/fix the store reuse case in a separate PR.

joshmoore:
Sounds good. There's certainly nothing *horrible* here at the moment, so definitely just fine tuning. Thanks.

PR #3721, opened 2015-04-15
URLs cleanup

sbesson:
See https://trello.com/c/5l9TG9PJ

This PR cleans some of the most problematic URLs in our code base, including:
- CI test servers links (trout, gretzky)
- deprecated websites (cvs.openmicroscopy.org)
- source code links for files migrated from Bio-Formats which were not cleaned as part of https://github.com/openmicroscopy/bioformats/pull/1455
- all of `components/server/sql` unless there is a good reason to preserve it /cc @joshmoore

Only item I could not figure out how to resolve is the `users.oo.org` link from `components/romio/test/ome/io/nio/utests/PlanarDataTest.java`. Does this file still exist and/or is the test still valid? /cc chris-allan


joshmoore:
Barring the one comment, all the individual commits at least make sense to me.

sbesson:
Pushed

ximenesuk:
This all looks good to merge.

PR #3722, opened 2015-04-16
Filepath info and Location dialog

dominikl:
Tiny fix for [Ticket 12824](http://trac.openmicroscopy.org.uk/ome/ticket/12824)

Test: Make sure the Location dialog ("Show parent projects and datasets") behaves like the "Show file paths" dialog; both should be automatically closed if they loose the focus.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/util/FilesetInfoDialog.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/util/FilesetInfoDialog.java
@@ -23,8 +23,6 @@ import java.awt.Dimension;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Point;
-import java.awt.event.WindowEvent;
-import java.awt.event.WindowFocusListener;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


mtbc:
The location dialog works far better for me under XFCE4 than the file paths dialog. Could they be made to behave like each other, but in the opposite way? :smiley:

dominikl:
Just checked with XFCE, should be fine.

pwalczysko:
I can see absolutely no problem with this behaviour on neither Mac OS 10.8 neither on Win 7 laptop.

mtbc:
For me, this *does* extend the problem with the file paths dialog to the location dialog also, so my plea in https://github.com/openmicroscopy/openmicroscopy/pull/3722#issuecomment-93711106 is repeated. :smiley: The problem is that sometimes when I click the icon the popup window does not appear (or immediately vanishes, hard to tell).

mtbc:
One option that might be interesting to try is not to auto-close if focus is lost within, say, 200ms, of the window opening? Perhaps then we can have the best of both worlds.

dominikl:
Delayed auto-close a bit and check again if out-of-focus before closing. Maybe with @mtbc 's XFCE version the dialog looses the focus for a very short bit during it's initialization, which immediately triggers the auto-close - just guessing. Unfortunately I can't test, don't see this bug with my XFCE installation.

mtbc:
Gah, thank you for trying, but now it just means I see the window for a third of a second! I guess we have a choice between make things better for some users but worse for others, or making all the users have to dismiss the dialog manually.

dominikl:
Thanks @mtbc . I removed the last commit again, as it doesn't fix the problem and therefore just adds unnecessary code. Probably just have to live with it (also seems to be very specific to your XCFE version).

pwalczysko:
Please do not opt for ANY compromise on the Windows / Mac front. These are the vast majority of client users. Linux - definite small minority (for client-side).

mtbc:
@pwalczysko: Bit of a chicken-and-egg issue there, though, right? Luckily Insight already has other UI code which is platform-specific. But I am happy for this issue not to block this PR.

For what it's worth, I have Xfce 4.10 installed.

dominikl:
Interesting, so it's not the XFCE version either; also used 4.10. 
Anyway... good to merge @pwalczysko ? (I reverted the PR back to the state you tested 3 days ago, so no re-testing should be necessary).

pwalczysko:
Still wroks fine. Ready to merge.

mtbc:
For what it's worth, I may have guessed what the problem is: focus-follows-mouse as the focus policy. (Default in older window managers, of course, less so in newer desktop environments.)

dominikl:
You're right @mtbc , changed to focus-follows-mouse policy and could reproduce the weird behaviour. But as you said, most common policy is clock-to-focus (especially on Windows and Mac, I suppose you can't even change it there), so won't take it into account.

PR #3723, opened 2015-04-16
Show tags of the selected objects only

dominikl:
Tiny fix for [Ticket 12679](http://trac.openmicroscopy.org.uk/ome/ticket/12679)

I couldn't reproduce the missing tag issue, but I fixed the problem, that the dialog showed all tags available. Now just the ones which are attached to the selected images are shown.

Test:
Select a few images with different tags attached to them. Open the ThumbnailFigure dialog and make sure only the tags which are attached to the selected images are shown.


pwalczysko:
To be tested on nightshade, Will do tomorrow.

pwalczysko:
Works as expected - sorts out both the original problem described by @will-moore on nightshade but also works fine on trout merge.

Ready to merge.

PR #3726, opened 2015-04-16
Only show useful search errors

will-moore:
Better handling of search errors, particularly those caused by parsing of invalid queries with a colon.

To test, try searching with valid and invalid key:value queries.
Invalid searches should give a useful error message.

Screenshots below are BEFORE and AFTER this change:

screen shot 2015-04-16 at 14 37 53

screen shot 2015-04-16 at 15 00 18

### DIFF
--- a/components/tests/ui/testcases/web/search.txt
+++ b/components/tests/ui/testcases/web/search.txt
@@ -48,9 +48,9 @@ ${FailureMessage}       Search should Not Contain an Alert Window.
 
 ${SEARCH URL}           ${WELCOME URL}search/
 
-@{SearchKeyWords}       *  \  *:  :
+@{SearchKeyWords}       *  \  *:  :  :abc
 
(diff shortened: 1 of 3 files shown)


pwalczysko:
In the output text I would stick to the nomenclature outlined in help http://help.openmicroscopy.org/search.html : ``Key-value Pairs`` instead of ``Key Value`` to describe the entity.  At present, I got wondering whether there is any significance of calling things ``Key Value`` and having syntax ``key: value`` (capitals and non-capitals) - I know that there is no such significance, but it would be good to disperse such doubts in a user as well.

pwalczysko:
Other than the minor reformulation, works as expected. I think it improves the user experience.

will-moore:
Fixed the error message to use lower case 'key-value' and 'key:value'.
Also fixed robot tests: https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-robotframework/215/console
```
05:52:05 Test All Keywords                                                     | FAIL |
05:52:05 'False' should be true.
```
Fixed the display to show "0 results" instead of "NaN result" when nothing found and added a test for this to the robot tests.

joshmoore:
Did this get relisted?

will-moore:
@joshmoore Yes. It had @sbesson's name listed to test this today.

joshmoore:
Ah, sorry. The change in title threw me: "Colon search error"

sbesson:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-robotframework/216/console only reports a transient error unrelated to this PR. Merging.

PR #3727, opened 2015-04-16
Viewer "Changing display settings" text

will-moore:
This updates the help message available from the (?) button in the viewer 'Edit Channels' dialog to reflect changes in updating the settings (don't need to click "Apply").

screen shot 2015-04-16 at 16 49 14

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
@@ -660,11 +660,12 @@
   /* ]]> */
 </script>
     <div id="channel-window-help" style="display: none;">
-      <h2>Changing the Image Display Settings:</h2>
-      <p>You can use the "Rendering Details" dialog box to change the settings for viewing this image.
-        The changes will not be saved until you click the "Save" button.</p>
-      <p>To change the color displayed for a channel, click on the color-picker icon at the right, which opens the "Choose Color" dialog box. You can choose the color using the interactive features at the top. Hex color codes can be input in the text box.  Preset colors can be selected at the bottom. Changes take effect when you click "OK".</p>
-      <p>To change the color intensity, drag the slider handles for the individual channels to the left or right. This changes the values that are used for the minimum and maximum displays settings.</p>
+      <h2>Rendering Details:</h2>
+      <p>Use the "Rendering Details" dialog box to change the viewer settings for this image.
+        The changes will be saved when you click the "Save" button.</p>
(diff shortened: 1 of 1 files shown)


components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
```diff
@@ -660,18 +660,11 @@
...
-      <p>To change the color intensity, drag the slider buttons for the individual channels to the left or right. This changes the values that are used for the minimum and maximum displays settings. Changes take effect when you click "Apply".</p>
+      <p>You can use the "Rendering Details" dialog box to change the settings for viewing this image.
+        The changes will not be saved until you click the "Save" button.</p>
+      <p>To change the color displayed for a channel, click on the color-picker icon at the right, which opens the "Choose Color" dialog box. You can choose the color using the interactive features at the top. Hex color codes can be input in the text box.  Preset colors can be selected at the bottom. Changes take effect when you click "OK".</p>
+      <p>To change the color intensity, drag the slider handles for the individual channels to the left or right. This changes the values that are used for the minimum and maximum displays settings.</p>
```
mtbc: "displays settings" seems odd (not this PR's fault)

rleigh-dundee: Maybe make it shorter, and more imperative by removing some of the "you can" and double negatives from the text and clearing up some of the wording to be simpler.  For example:

```
Use the "Rendering Details" dialog box to change the viewer settings for this image.  The changes will be saved when you click the "Save" button.

To change the channel color, click on the color-picker icon at the right to open the "Choose Color" dialog box. Using the color using the selector at the top, select the desired hue using the outer ring, and saturation and value with the inner square. Alternatively, hexadecimal color codes may be input in the text box, or preset colors may be selected at the bottom. The changes will take effect when you click "OK".

To change the color intensity, drag the slider handles for the individual channels to the left or right to change the the minimum and maximum display values.</p>
```

will-moore:
@rleigh-dundee You wanna review my use of your suggested text? I took most of your changes, but not all.

rleigh-dundee:
@will-moore Certainly improved, thanks.  The thing I'm not so sure about is the "interactive features" in the middle paragraph; this is kind of nebulous since it doesn't say *which* interactive features, *how* to use them and the *effect* they have.  The entire UI is "interactive features", so the current text isn't particularly enlightening for the reader (IMO).

will-moore:
Yeah, but I don't think we need to tell users how to use a color-picker. "Click on color to pick color" is a bit dumb. I've tweaked it a bit, but I don't think this is too critical. I'd never even opened this info box until this PR!

rleigh-dundee:
Probably OK to merge then.

PR #3728, opened 2015-04-16
Python test library object helpers

sbesson:
See https://trello.com/c/OTdzTuMZ

This PR is only modifying the Python test API and some of the integration tests. The first commit is split  library methods by functionality for better readability and expands/reviews the helper methods used for object creation and manipulation. The following set of commits is using these object helpers in a subset of test classes.

All tests should still be passing with this PR. Additionally, what should be assessed is whether this proposal improve the readability of individual tests while reducing the amount of logic for each test setup.

/cc @ximenesuk

### DIFF
--- a/components/tests/python/library/__init__.py
+++ b/components/tests/python/library/__init__.py
@@ -106,6 +106,12 @@ class ITest(object):
         cls.query = cls.sf.getQueryService()
 
     @classmethod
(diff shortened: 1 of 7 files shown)


ximenesuk:
Changes look good both the reorganisation and the use of helpers. Cuts out a lot of cruft.

PR #3735, opened 2015-04-20
Bugfix: Orphaned images selection

dominikl:
Tiny fix for [Ticket 12814](http://trac.openmicroscopy.org.uk/ome/ticket/12814)

Test: Log in as a user who has some images in 'Orphaned Images' (e. g. user-3), select an orphaned image -> you should see it's metadata in the right panel; click on an empty space in the central panel -> metadata panel should be empty; select the **same** image again -> you should again see the metadata of the image.


mtbc:
This addresses the ticket and seems to me to be **good to merge**.

A relevant issue which could be addressed here or separately: When I click in the empty space away from the orphaned image thumbnails, the thumbnail is deselected, but remains selected in the tree view.

dominikl:
Yes, good point, I try to fix that in this PR, too.

mtbc:
:+1: Good to merge.

PR #3737, opened 2015-04-20
Deprecate omero_ext.uuid

sbesson:
See http://trac.openmicroscopy.org.uk/ome/ticket/3774.
With OMERO 5.1.x, Python 2.6 is the minimal version so this workaround is not
needed anymore.


joshmoore:
Thinking about this, @sbesson, I think we need to deprecate the `*uuid.py` modules instead. Existing/modified scripts could break with this.

sbesson:
Force pushed to preserve the history of `omero_ext/*uuid.py` and deprecation warning added.

ximenesuk:
This looks fine. Tests and DropBox passing/working as expected. Good to merge.

PR #3738, opened 2015-04-20
Update swingx library

dominikl:
This PR updates the swingx library from 0.9.4 to the most recent version 1.6.4. To make Insight work with the new version, I had to adjust the `TaskPaneBrowser` ( = the left hand side component, hosting the "Projects", "Screens", "Attachments", ... task panes). I also removed the usage of `TableLayout` within the `TaskPaneBrowser`, this should fix the tickets [11403](http://trac.openmicroscopy.org.uk/ome/ticket/11403) and [12655](http://trac.openmicroscopy.org.uk/ome/ticket/12655).

**Note:** The PR also **removes the JTabbedPane** layout. The layout of the left hand side component could be changed from the TaskPane style to TabbedPane style by modifying the treeviewer.xml. As the TabbedPane style actually isn't working properly (e. g. the search panel is missing), the user can't change the style via Insight anyway (this option is only available via editing the xml file), and it's quite a hassle to always take this (practically never used) style into account coding wise, I think it's better to remove it.
Any objections @jburel @gusferguson ?

**Test:** The swingx update especially affected task panes (used in *left hand side panel* and *right side (meta data viewer) panel*) as well as tables (*search results*, *ROI manager*, etc.), so it would be a good idea, to check if these components are still working.


jburel:
The configuration option was mainly introduced when we did the transition of style. Nothing in doc or help indicated that the layout can be modified.

dominikl:
Another point to consider: As the 5.1.1 is quite close, maybe I should exclude this PR and include it again after release, aiming for 5.1.2, so that it could stay a bit longer in the merge build for internal testing. There might be other parts in Insight affected by the swingx library update, I haven't considered yet. Could hardly find documention about the caveats for updating this library.

jburel:
Due to the nature of the changes, I think we should not target it for 5.1.1.
We could have some unexpected side effects.

mtbc:
After "switch user" I don't have a tree view any more.

joshmoore:
> Due to the nature of the changes, I think we should not target it for 5.1.1.

Marking as `breaking` then.

pwalczysko:
In ROI Manager I cannot adjust column width in this table, all the other tables are okay (can adjust column width in them)
screen shot 2015-04-21 at 11 14 51

pwalczysko:
In Admin tab, logged in as user-4 to trout merge, I am trying to change the default group of a user in the right-hand pane, selecting the user first on the left-hand pane.

After I change the default group and navigate away without save, I am asked whether I want to save the changes befrore navigating away.
I say ``Yes``.
But when I navigate back, the change was not saved, even after closing and reopening Insight the change is not saved. 
Note that I was trying to change the default group of other users (should not be allowed, but why am I still getting the non-greyed out menu and the question about saving ?).
I also tried to chage my own default group, which is allowed, and I failed to do it (see above) by navigating away and saying Yes in the dialog.

screen shot 2015-04-21 at 11 39 15

pwalczysko:
Otherwise detected no untoward issues in right-hand, left-hand and ROI manager tables.

dominikl:
Thanks @pwalczysko for the initial review. Will look into these issues after 5.1.1 release.

dominikl:
- Can we remove the breaking label again, so that this is tested within the usual merge build PR reviews, @joshmoore , @jburel ?
- The problem spotted by @pwalczysko (Admin tab) is not related to this PR, will be tackled in a separate PR. 
- Added a workaround for the 'disappearing tree view when switching user' problem noticed by @mtbc

jburel:
@dominikl: done

mtbc:
Does Travis need a kick?

joshmoore:
Leaving this in the merge build for some time to kick the tires.

mtbc:
Rarely lately I start up Insight and there is no metadata pane on the right. This time I tried selecting things in the tree view and the metadata pane appeared. However, the tree view insists on being very wide -- it can't be resized down to a more normal width -- and as I do other things, like try to resize the whole window vertically, the tree view grows wider and wider. It's funny and scary at the same time!

jburel:
@dominikl 
excluding for now so #3770 can be tested.
This PR introduced some strange behaviours.

dominikl:
I'm closing this PR, creates more hassle than it solves. Just have to stick to the old swingx library for now.

PR #3725, opened 2015-04-16
Refactor plugins using Delete2 & Chgrp2 requests

ximenesuk:
This PR refactors the CLI plugins `delete` and `chgrp` to use the new `Delete2` and `Chgrp2` requests respectively. In addition `fs rename` has been modified to use `Delete2`.

The `--edit` option has been removed as it was broken by the earlier changes

The `--opt` option have been removed and its usage replaced by two new options `--include` and `--exclude`. How the old option maps to the new ones is detailed in the testing section below.

A new option `--ordered` has been introduced to force the preservation of the order of the arguments.

Integration tests have been added but, as ever, more would be useful to test more complex functionality.

Testing
-------
Relevant cli unit and integration tests should pass.

A number of simple delete and chgrp should be checked, see `--help` for examples. Also see below for how the new usage maps to the old usage. In addition more complex "skiphead" requests should be tried, for examples see below.

Prior to this PR objects had to be listed separately. So to delete three images and two datasets.
```
$ omero delete /Image:51 /Image:52 /Image:53 /Dataset:101 /Dataset:201
```
While this form can still be used the following more concise version is now possible.
```
$ omero delete Image:51,52,53 Dataset:101,201
```
Note also that the preceding `/` is no longer used though the command should still work if it is.

Prior to this PR including or excluding objects from a delete used the `KEEP` and `HARD` syntax with `--opt`. So to delete and image but keep its annotations:
```
$ omero delete --opt /Annotation:KEEP /Image:51 
```
This form is no longer possible. Instead the `--exclude` option should be used.
```
$ omero delete Image:51 --exclude Annotation
```
ie the annotations are *excluded* from the delete. Multiple classes of object can be excluded:
```
$ omero delete Image:51,52,53 --exclude TagAnnotation,FileAnnotation
```
Similarly uses of `/Object:HARD` would be replaced by `--include Object`. 

All of this also applies to `chgrp` so to change the group of dataset but not its images.
```
$ omero chgrp 101 Dataset:201 --exclude Image
```
This may orphan the images if they are not also in another dataset.

To delete or change the group of all the objects under some other object a `SkipHead` command is used behind the scenes. CLI users don't need to worry about the details this but the CLI syntax is important.

Delete all images under a project an object hierarchy is used. 
```
$ omero delete Project/Dataset/Image:53
```
As only the first and last objects are important this hierarchy can be shortened
```
$ omero delete Project/Image:53
```
Both of these commands will delete **all** of the images under Project 53.

[...]

PR #3739, opened 2015-04-20
Replace deprecated Delete/Chgrp methods in gateway

ximenesuk:
This PR replaces deprecated delete and chgrp methods in the gateway and the gateway tests.

With the PR as it stands the tests pass. However, not all functionality may be fully tested and so this PR is probably best tested on a breaking build. In particular methods such as `deleteSettings` and calls to `deleteObjects` with extended graph specs that will exploit `SkipHead` need through checking.

The one method of concern is is `deleteObjectDirect`:

https://github.com/ximenesuk/openmicroscopy/blob/deprecated-gateway/components/tools/OmeroPy/src/omero/gateway/__init__.py#L3803

This is used extensively by the web but a couple of basic patterns, using `Delete2` commands or using `deleteObjects` did cause at least one test to fail. The tests now pass through re-ordering the deletes. It may be that `deleteObjectDirect` should now be deprecated in favour of `deleteObjects` but I'll leave that as a point for discussion.

I have not yet dealt with the `ObjectOne+ObjectTwo` deprecated graphspec. It may be that the webclient simply doesn't use this form and the comment in the docstrings has just been copied in. If it is used then it would need to be re-written and the deprecated form trapped.

Testing
-------
To test this as many aspects as possible of delete and change group with in the web need to be tested.

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3841,7 +3841,6 @@ class _BlitzGateway (object):
                                 * 'Well'
                                 * 'Annotation'
                                 * 'OriginalFile'
-                                * 'Image+Only'
                                 * 'Image/Pixels/Channel'
 
                                 As of OMERO 4.4.0 the correct case is now
@@ -3856,6 +3855,10 @@ class _BlitzGateway (object):
         :rtype:                 :class:`omero.api.delete.DeleteHandle`
         """
 
+        if '+' in graph_spec:
+            raise AttributeError(
+                "Graph specs containing '+'' no longer supported: '%s'"
+                % graph_spec)
         if not isinstance(obj_ids, list) and len(obj_ids) < 1:
             raise AttributeError('Must be a list of object IDs')
 
@@ -3940,6 +3943,11 @@ class _BlitzGateway (object):
         :param obj_ids:         IDs for the objects to move.
         :param group_id:        The group to move the data to.
         """
+
+        if '+' in graph_spec:
+            raise AttributeError(
+                "Graph specs containing '+'' no longer supported: '%s'"
+                % graph_spec)
 
         graph = graph_spec.lstrip('/').split('/')
         obj_ids = map(long, obj_ids)
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3931,40 +3931,45 @@ class _BlitzGateway (object):
         :param graph_spec:      String to indicate the object type or graph
                                 specification. Examples include:
 
-                                * '/Image'
-                                * '/Project'   # will move contents too.
-                                * NB: Also supports 'Image' etc for convenience
+                                * 'Image'
+                                * 'Project'   # will move contents too.
+                                * NB: Also supports '/Image' etc for backwards
+                                  compatibility.
         :param obj_ids:         IDs for the objects to move.
         :param group_id:        The group to move the data to.
         """
 
-        if not graph_spec.startswith('/'):
-            graph_spec = '/%s' % graph_spec
-            logger.debug('chgrp Received object type, using "%s"' % graph_spec)
+        graph_spec = graph_spec.lstrip('/')
+        graph = graph_spec.split("/")
+
+        chgrp = Chgrp2(targetObjects={graph[0]: obj_ids}, groupId=group_id)
+
+        if len(graph) > 1:
+            skiphead = SkipHead()
+            skiphead.request = chgrp
+            skiphead.targetObjects = chgrp.targetObjects
+            skiphead.startFrom = [graph[-1]]
+            chgrp = skiphead
+
+        requests = [chgrp]
 
         # (link, child, parent)
         parentLinkClasses = {
-            "/Image": (omero.model.DatasetImageLinkI,
-                       omero.model.ImageI,
-                       omero.model.DatasetI),
-            "/Dataset": (omero.model.ProjectDatasetLinkI,
-                         omero.model.DatasetI,
-                         omero.model.ProjectI),
-            "/Plate": (omero.model.ScreenPlateLinkI,
-                       omero.model.PlateI,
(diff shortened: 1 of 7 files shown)


ximenesuk:
@will-moore you may want to take a look at this (excluded) PR as it may break web. I also want to remove the use of the update service in `deleteObjectDirect` as per the description but may need your help!

will-moore:
Sorry - I'm gonna need a bit of an update on what ```Delete2``` is and what is ```SkipHead```?

mtbc:
The replacement for `Delete` (`SkipHead` being for longer graphspec paths), as in http://www.openmicroscopy.org/site/support/omero5.1/developers/Server/GraphsMigration.html.

ximenesuk:
Okay, this now is working locally for basic deletes and group changes. It would now be good to get it tested more widely without impacting on any other testing work.

/cc @joshmoore @sbesson @will-moore @pwalczysko

joshmoore:
Swapped `exclude` for `breaking`

pwalczysko:
Passed the ``non-permission`` part of the  [delete scenario](https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/DeleteWithPerm.html)

pwalczysko:
Passed the ``non-permission`` part of [Move scenario](https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/MoveDataBetwGroups.html)

pwalczysko:
@will-moore @ximenesuk @mtbc @jburel : @ximenesuk seems to suggest we need a full permissions testing as well. It should be not too difficult to organize, as we have our DB from December. Should we go for it next week ? (will need the permissions DB with the MapAnnotations and other Annotations on it deployed) @sbesson

mtbc:
It's certainly a good way to increase my desire to automate the scenario using Robot Framework!

will-moore:
Things initially seem to be OK. Deleting Datasets and Images and chgrp images, except Deleting datasets with projected images (Mark created https://trac.openmicroscopy.org.uk/ome/ticket/12856). Haven't been looking at the DB after deletes (need to run locally or get hql access to breaking DB).

joshmoore:
Time to drop `breaking`?

pwalczysko:
@joshmoore Yes, this is the deal. We have to have breaking dropped especially for tomorrow's permissions test. Due to the server/DB setup, we need to have it on trout merge.

ximenesuk:
This last commit, which can be reverted before merging, echoes the Delete2/Chgrp2 command to the log file. This should help conform what the command actually is and allow us to try to produce via the CLI. The web debugging needs to be at DEBUG for further testing. /cc @pwalczysko @joshmoore

mtbc:
One of the problems discovered in yesterday's permissions testing is outwith this PR. Filed https://trac.openmicroscopy.org/ome/ticket/12868 accordingly.

joshmoore:
Kicked travis due to `testCreateJPEG` (becoming a new favorite). That build was 16 hours ago, so likely not present in the build today?

https://github.com/snoopycrimecop/openmicroscopy/commit/1abc4a9d22891af85f04123c333f5a62cdcb1b77
```
  # PR 3739 ximenesuk 'Replace deprecated Delete/Chgrp methods in gateway' (status: failure)
```

cc: @pwalczysko !

pwalczysko:
@joshmoore yes, it was not present today, but this is precisely a good thing. I take it so: yesterday, all the bugs were accumulated and described (with this PR in). Today, without this PR in, the bugs were repeated. This means what was repeated today, was not a bug due to this PR.

pwalczysko:
The only thing to try to repeat would be line 8 from https://docs.google.com/spreadsheets/d/1YAxDXfgrtHHtzZUxhRSUsYEYnW34ecGgwXV835FpuHE/edit#gid=1485338836 tomorrow, when this PR is in .

mtbc:
Today it's in.

pwalczysko:
Absolutely cannot repeat the line 8 of https://docs.google.com/spreadsheets/d/1YAxDXfgrtHHtzZUxhRSUsYEYnW34ecGgwXV835FpuHE/edit#gid=1485338836 - this means this one does not repeat with this PR and does not repeat without this PR. I think this means that this PR has nothing to do with this . All good here from the UI perspective.

ximenesuk:
Thanks @pwalczysko.
I'll push some commits today that should require no further functional testing, just a little code review and integration tests.

pwalczysko:
Did now also tests on Rnd settings behaviour:
   - basic workflows like Set Rnd settings, save and then use ``Imported`` and ``Min Max`` or ``Full range`` works fine on your own and not-own images
   - moves of images preserve the Rnd settings as expected, and in the target groups the workflow from the abovementioned bullet point work as well, both on own and not-own images

ximenesuk:
@pwalczysko thanks! I can also confirm that there are integration tests for resetting the rendering settings so the method `_deleteSettings()` is being tested.

ximenesuk:
These final three commits add and fix argument checking for the `deleteObjects()` and `chgrpObjects()` and provide some unit tests for that argument checking.

mtbc:
Aha, I see the conflicts now: you updated gateway tests' `admin.changePermissions` to `Chmod` in https://github.com/ximenesuk/openmicroscopy/commit/b3cf8fbebdf6e94aa89aaa021b95bfff45cf315b and I updated them to `Chmod2` in https://github.com/mtbc/openmicroscopy/commit/e18bf9e20160f1e61fdd9a17a90be30d79aee63e. But, I have a `_waitOnCmd` pattern going on that you don't; am I making more work for myself than necessary?

ximenesuk:
@mtbc Yes, see https://github.com/openmicroscopy/openmicroscopy/pull/3725#discussion_r28578888 for comment on `c.submit` vs `c.sf.submit`, the former does the waiting for you.

mtbc:
Aha, thank you for the tip, @ximenesuk. (It'll yell about a not-OK response, right?) I'll adjust accordingly.

pwalczysko:
The publication scripts work as expected.

joshmoore:
:thumbsup:

PR #3740, opened 2015-04-20
5.1.1 misc size fixes

joshmoore:
#### [ticket:12820](https://trac.openmicroscopy.org.uk/ome/ticket/12820#comment:7) - ServiceHandler logging ####

https://trac.openmicroscopy.org.uk/ome/ticket/12820 describes long figure json strings being printed in Blitz-0.log. This should limit those strings to a max. length of 1000 chars. cc @will-moore 

#### Large (50k+) directories  ####

`findRepoFiles` previously passed all the files (`basenames`)
in an entire directory into one query. This leads to strange
errors with PostgreSQL of the form:

```
IOException: Tried to send an out-of-range integer as a 2-byte value: 52671
```

Now, the list of basenames are grouped into batches, filling
up a single `Map<String, Long>` which is returned to the caller.

To test, generate a large directory and attempt to import:
```
/tmp/dont-look-here$ cat make.sh
echo "test_Z<00000-60000>.fake" > fake.pattern
for x in $(seq 0 60000); do touch $(printf "test_Z%05d.fake\n" $x); done
```
If the individual files start uploading, then the bug has been fixed. It is likely worth adding an import job, but the upload itself takes a significant amount of time. Until that can be fixed, this may need to be manual. Note: this will need https://github.com/melissalinkert/bioformats/tree/filestitcher-mem to be tested properly. @melissalinkert to open a PR.

### DIFF
--- a/components/model/src/ome/util/SqlAction.java
+++ b/components/model/src/ome/util/SqlAction.java
@@ -39,6 +39,7 @@ import org.springframework.jdbc.core.simple.SimpleJdbcOperations;
 import org.springframework.jdbc.core.simple.SimpleJdbcTemplate;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
 
 /**
  * Single wrapper for all JDBC activities.
@@ -566,6 +567,8 @@ public interface SqlAction {
      */
     public static abstract class Impl implements SqlAction {
 
+        protected final static int MAX_IN_SIZE = 1000;
+
         protected final Logger log = LoggerFactory.getLogger(this.getClass());
 
         protected abstract SimpleJdbcOperations _jdbc();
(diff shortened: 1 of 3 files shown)


joshmoore:
f33ccda23df9cf028277512c70aa9ae27f04bd8d: I've validated via a 60K fake file. Without this commit, I get an early `PSQLException`; with, well, I've uploaded 12K of the 60K... :smile:

sbesson:
Tested the second part of the PR following the 60K fake pattern scenario in the description (with https://github.com/openmicroscopy/bioformats/pull/1752 included). 
Fake pattern import completed in:

```
60002 files uploaded, 1 fileset created, 1 image imported, 0 errors in 3:43:47.344
```

and the resulting image has the correct number of sections:

screen shot 2015-04-24 at 08 53 29

Assuming the first part of this PR has been tested this is good to go (and we can make performance improvements in the due course of 5.1.x).

sbesson:
Discussed with @joshmoore, unit test coverage for the json bit are in place and passing. Merging this for 5.1.1

PR #3734, opened 2015-04-18
CLI sessions list

sbesson:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12392 and https://trac.openmicroscopy.org.uk/ome/ticket/12834

To test this PR, check Travis is green and reproduce the scenarios described in both tickets.

For 12392, the main change is that inactive sessions purged by default by the `sessions list` subcommand are not listed anymore by the command output, i.e. the command lists all local non-expired sessions. To list all local sessions independently of their status, `bin/omero sessions list --no-purge` should be used. Listing differences between the output of two consecutive calls to `bin/omero sessions list` should be only caused by either newly created sessions or sessions which expired in-between the two calls.

/cc @kennethgillen @pwalczysko

PR #3742, opened 2015-04-20
Fix usage of OMERO_SESSION_DIR

sbesson:
Initial implementation of OMERO_SESSION_DIR led to the creation of a omero/sessions subfolders. This behavior is counter-intuitive and does not match the CLI documentation. This commit fixes this behavior by deprecating `OMERO_SESSION_DIR` in favor of `OMERO_SESSIONDIR` which allows to set the equivalent of `$HOME/omero/sessions` by environment variable.

To test this PR, check the sessions storage mechanism under various environments:

```
# Should store under ~/omero/sessions
bin/omero login -s localhost -u user -w pass -C
bin/omero sessions file
# Should store under ~/xxx/omero/sessions
OMERO_SESSION_DIR=~/xxx bin/omero login -s localhost -u user -w pass -C
OMERO_SESSION_DIR=~/xxx bin/omero sessions file
# Should store under ~/xxx2
OMERO_SESSIONDIR=~/xxx2 bin/omero login -s localhost -u user -w pass -C
OMERO_SESSIONDIR=~/xxx2 bin/omero sessions file
# Should store under ~/xxx2
OMERO_SESSIONDIR=~/xxx2 OMERO_SESSION_DIR=~/xxx bin/omero login -s localhost -u user -w pass -C
OMERO_SESSIONDIR=~/xxx2 OMERO_SESSION_DIR=~/xxx bin/omero sessions file
```

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -97,12 +97,26 @@ Options for logging in:
     $ bin/omero sessions login --sudo=root example@localhost
     Password for root:
 
-Other commands:
+Other sessions commands:
 
-    $ bin/omero sessions list
-    $ OMERO_SESSIONDIR=/tmp bin/omero sessions list
+    # Logging out of the currently active sessions
     $ bin/omero sessions logout
-    $ bin/omero sessions clear
+
+    # List all locally available sessions (purging the expired ones)
+    $ bin/omero sessions list
+
+    # List all local sessions
+    $ bin/omero sessions list --no-purge
+
+
+Custom sessions directory:
+
+    # Specify a custom session directory using OMERO_SESSIONDIR
+    $ export OMERO_SESSIONDIR=/tmp/my_sessions
+    # Create a new session stored under OMERO_SESSIONDIR
+    $ bin/omero sessions login
+    $ bin/omero sessions file
+    $ bin/omero sessions list
 """
 
 
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -207,8 +207,7 @@ class SessionsControl(BaseControl):
         self._configure_dir(login)
 
     def _configure_dir(self, parser):
-        parser.add_argument("--session-dir", help=SUPPRESS,
-                            default=os.environ.get('OMERO_SESSIONDIR', None))
+        parser.add_argument("--session-dir", help=SUPPRESS)
 
     def help(self, args):
         self.ctx.err(LONGHELP % {"prog": args.prog})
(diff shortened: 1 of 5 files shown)


joshmoore:
So you assume that anyone who has used it to date, is actually using `$X/omero/sessions/omero/sessions` and therefore the change to just `$X/omero/sessions` should be minimally breaking?

joshmoore:
Independent of the "breakingness", the testing of gh-3734 was much more sensible with this change in place.

sbesson:
On the breakingness, this PR is certainly open to have this discussion before code freeze.

My major issue is that the CLI help, the [Sphinx documentation](https://www.openmicroscopy.org/site/support/omero5.1/users/cli/sessions.html?highlight=omero_session_dir#sessions-directory) and the implementation are not in-line with each other. If we estimate this code change is too breaking for 5.1.1, we can certainly close this PR and I will fix the CLI help elsewhere. /cc @ximenesuk @ehrenfeu

ehrenfeu:
I can't say much about this right now as I didn't have a chance of testing gh-3734 yet, although it is pretty high on my priority list - hopefully I'll find the time there in the next two weeks (in my case this was linked to the HRM-OMERO-connector).

components/tools/OmeroPy/src/omero/plugins/sessions.py
```diff
@@ -100,7 +100,7 @@
 Other commands:
 
     $ bin/omero sessions list
-    $ OMERO_SESSION_DIR=/tmp bin/omero sessions list
+    $ OMERO_SESSIONDIR=/tmp bin/omero sessions list
```
joshmoore: I'd used something uniquer here, like `/tmp/$$`

sbesson: Or like https://github.com/openmicroscopy/ome-documentation/pull/1194/files ? :)

manics:
Would it be reasonable to keep the variable as `OMERO_SESSION_DIR` and just change its behaviour?

sbesson:
@manics: this was my initial implementation (1938859). I guess the debate here is how breaking this would be. We can have a call this morning.

manics:
If `OMERO_SESSION_DIR` is set the created session is still stored in `~/omero/sessions`:
```
$ rm -r ~/xxx/rm: cannot remove `/homes/spli/xxx/': No such file or directory

$ OMERO_SESSION_DIR=~/xxx bin/omero login -s localhost -u user-1 -w ome
<string>:144: DeprecationWarning: OMERO_SESSION_DIR is deprecated. Use OMERO_SESSIONDIR instead.
Using session 81926c0b-4eab-4eb2-b3f1-a350e6efc2de (user-1@localhost:4064). Idle timeout: 10.0 min. Current group: private-1

$ ls ~/xxx/
ls: cannot access /homes/spli/xxx/: No such file or directory

$ ls ~/omero/sessions/localhost/user-1/
81926c0b-4eab-4eb2-b3f1-a350e6efc2de
```

sbesson:
Just realized the deprecation warning for the `--session-dir` argument needs to be wrapped in another `if base_dir` statement (else it is now printed by default on every regular `bin/omero sessions` call on Python 2.6). I will push another fix shortly but the environment logic is unmodified and can be retested following the PR description.

manics:
good to merge

PR #3745, opened 2015-04-21
Fixed DnD mif move

dominikl:
While checking [Ticket 12171](https://trac.openmicroscopy.org.uk/ome/ticket/12171) I noticed that moving a single image which is part of a MIF via drag&drop into another group crashed Insight with an IllegalStateException being thrown. This PR fixes the issue.

**Test:** 
Select a subset of images of a MIF and move them via Drag&Drop into another group; expected: All images of the MIF should be moved (but you should get asked for confirmation beforehand).


pwalczysko:
Works as expected, but note that I updated https://trac.openmicroscopy.org/ome/ticket/11785#comment:3 - the DnD feature itself is in need of some fix. If this is for another PR, then this is ready to merge.
@dominikl ?

dominikl:
Thanks @pwalczysko . Yes, the general DnD problem in Insight is a matter for a separate PR.

PR #3746, opened 2015-04-21
chgrp dialog clearer

will-moore:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12798

Now the chgrp dialog asks you to choose a group and gives you a better warning if you hit 'OK' without choosing a group.

screen shot 2015-04-21 at 16 10 08


dominikl:
Note: During testing it happened once, that I could move an empty dataset into another group's new dataset (for some reason the "New" button in the move dialog, let me create and choose a dataset instead of a project), which of course resulted in an exception. But I wasn't able to reproduce it, take it as a one-time incident.

PR #3747, opened 2015-04-22
Always drag image, never scroll through Z/T or zoom

dominikl:
Follow-up PR to #3712 , respectively [Ticket 12837](https://trac.openmicroscopy.org.uk/ome/ticket/12837)

As @will-moore mentioned, mouse dragging should always drag the image (and not scroll through Z/T or zoom the image). Previously the behaviour of mouse dragging in Insight was different for 'small' and 'big' images. With this PR mouse dragging will always drag the image regardless of the image size.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/ImageCanvasListener.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/ImageCanvasListener.java
@@ -166,50 +166,15 @@ class ImageCanvasListener
     }
 
     /**
-	 * Zooms in and out the image if the <code>Shift</code> key is down,
-	 * pans if the <code>Alt</code> key is down.
-	 * Selects a new z-section and time-point otherwise.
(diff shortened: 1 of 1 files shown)


pwalczysko:

will-moore:
This is definitely working. One thing I noticed is that when the ROI tool is open, click & dragging actions on the image only affect ROI selection/movement (don't drag the image) which is expected. But, the mouse-wheel behaviour has also changed to move the image within the viewport instead of increment Z-index. I don't think this is expected? 
NB: I still occasionally see zoom changing during rapid mouse-wheel scrolling through Z-index.
Both of these are smaller issues and don't affect this PR.

jburel:
Z-increment using the mouse-wheel has alway been like that when the MT is up
We can decide if we want to change this behaviour later on.
Note that there is still a major difference between web and insight on the mouse wheel front. 
This is outside the scope of this PR

Added card. To be discussed during next client meeting
https://trello.com/c/3H67wdIJ/337-viewer-mouse-wheel-and-save-rendering

jburel:
Card added, merging

PR #3749, opened 2015-04-22
Save All for Plates

will-moore:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12799

To test, browse SPW to open Preview panel for image in well.
 - Edit rendering settings
 - Click "Save To All"
 - Should see settings applied to plate and thumbnails should update.


pwalczysko:
All works as expected. Ready to merge.

PR #3751, opened 2015-04-22
fix OriginalFile ordering for deletion

mtbc:
As suggested by https://trello.com/c/gglTLzkS/480-have-delete2-order-file-deletions, improves the new graph requests such that original file deletions may be combined within the same request regardless of if they must be deleted in a certain order.

Integration tests ought still pass, including the new `DeleteServiceFilesTest.testRecursiveDelete` at https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-java/lastCompletedBuild/testngreports/integration/DeleteServiceFilesTest/testRecursiveDelete/.

Manual testing may be most easily effected with the help of @ximenesuk's CLI PRs as you need a hierarchy of original files that are not linked to anything else, and there is the extra complication that regular users appear not to be able to delete directories. Still, to give it a try from a simple Python client:

1. Arrange a file hierarchy in a repository. What I do is import an image and then, say, for image 12345, submit a `Delete2(targetObjects={'Image' : [12345]}, childOptions=[ChildOption(excludeType=['OriginalFile'])])` to delete all of it except for its files.
1. Find the IDs of the files and directories. You can't just use a dry-run `Delete2` on the image to find them as that doesn't mention directories, so for a fresh import I just look for the latest entries in the `originalfile` table.
1. Submit a single `Delete2` request to delete all those files and directories, listing their IDs in any crazy order you like regardless of which contain which. Do this as the root user, so that directories may be deleted.
1. Observe that they are all listed in the `Delete2Response` and deleted from the database and the filesystem. (They are probably listed in deletion order, from "deep" to "shallow".)

Alternatively, instead of unlinking the original files from everything else before deleting them, you may of course simply include whatever links them (e.g., an image) in the actual delete request you submit; just note that specifying the image will imply the files but not the directories (which are also original files) and you need to include the directories in the same delete request to properly test this PR.

--no-rebase

### DIFF
--- a/components/blitz/src/omero/cmd/graphs/Chgrp2I.java
+++ b/components/blitz/src/omero/cmd/graphs/Chgrp2I.java
@@ -34,6 +34,7 @@ import com.google.common.base.Function;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
 import com.google.common.collect.SetMultimap;
 
 import ome.model.internal.Permissions;
@@ -165,7 +166,9 @@ public class Chgrp2I extends Chgrp2 implements IRequest, WrappableRequest<Chgrp2
                         targetObjectCount++;
                     }
                 }
-                return graphTraversal.planOperation(helper.getSession(), targetMultimap, true);
+                final Entry<SetMultimap<String, Long>, SetMultimap<String, Long>> plan =
+                        graphTraversal.planOperation(helper.getSession(), targetMultimap, true);
+                return Maps.immutableEntry(plan.getKey(), GraphUtil.arrangeDeletionTargets(helper.getSession(), plan.getValue()));
             case 1:
                 graphTraversal.unlinkTargets();
                 return null;
--- a/components/blitz/src/omero/cmd/graphs/Chown2I.java
+++ b/components/blitz/src/omero/cmd/graphs/Chown2I.java
@@ -34,6 +34,7 @@ import com.google.common.base.Function;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
 import com.google.common.collect.SetMultimap;
 
 import ome.api.IAdmin;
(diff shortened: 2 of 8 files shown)


ximenesuk:
Once this is in a build I can test via the revised plugin.

mtbc:
Thank you! Travis is passed so hopefully in today's next build. Do feel free to ask me stuff if I can usefully expedite your efforts.

mtbc:
(Note that the latest commit is just affecting the test, so this morning's build should already be valid server-wise.)

mtbc:
```
===============================================
OmeroJava.integration
Total tests run: 1196, Failures: 0, Skips: 0
===============================================
```

ximenesuk:
When tested as root this works just great. A complex nest of folders from an import and some leftover imports (by a non-root user):
```
$ omero delete OriginalFile:165,163 OriginalFile:162 OriginalFile:171,170,169 OriginalFile:168,163
Created session 0a4426d7-f866-4962-a782-22d945e1695e (root@localhost:4064). Idle timeout: 10.0 min. Current group: system
omero.cmd.Delete2 OriginalFile 165,163,162,171,170,169,168,163... ok
```

On first testing I failed to read @mtbc's instructions properly and tried a delete as the user:
```
$ omero delete OriginalFile:165,163 OriginalFile:162 OriginalFile:171,170,169 OriginalFile:168,163
Using session b08eb058-865f-4cc0-b9f3-fb26bb130fc8 (user-0@localhost:4064). Idle timeout: 10.0 min. Current group: pg-0
omero.cmd.Delete2 OriginalFile 165,163,162,171,170,169,168,163... failed: 'graph-fail'
	message=Permissions are null! Security system failure -- refusing to continue. The Permissions should be set to a default value.
...
```
This is reproducible and it is the nature of the exception that is curious. What's also notable is that if the ids are broken down into files and then directories the delete is possible by the user, this is from an earlier import:
```
$ omero delete OriginalFile:159,160,161 OriginalFile:156,157,158 --ordered
Using session 8bc9accd-3da8-47a7-a23d-48a75f9852eb (user-0@localhost:4064). Idle timeout: 10.0 min. Current group: pg-0
omero.cmd.Delete2 OriginalFile 159,160,161, OriginalFile 156,157,158... ok
```
In all cases the underlying files and folders were deleted.

mtbc:
Great, thank you: that's an interesting discovery. At least this PR improves things, but maybe there is yet scope for further improvement in a later version if ordinary users *can* sometimes delete directories (I hadn't run into a case where I could). I'll create a card about that.

mtbc:
After further thought while asleep: I bet there is a separate issue outwith this PR where a normal user cannot combine into the same request target objects from different groups. (Directories are in the `user` group.) I think this is true, and can easily be fixed, but, at this point, while this PR is good for 5.1.1, given the code freeze that fix would be in a separate PR for 5.1.2.

mtbc:
> a normal user cannot combine into the same request target objects from different groups

I have now confirmed the above suspicion. The issue's outwith this PR and I will open a PR to fix it in 5.1.2. @ximenesuk: This PR is good to merge?

joshmoore:
Ok, closing before @mtbc adds more to his monologue. Thanks, guys.

mtbc:
I've been adding more on trello instead. :smiley:

PR #3752, opened 2015-04-22
Import ij

jburel:
Fix possible NPE
see https://trac.openmicroscopy.org.uk/ome/ticket/12833

to test: 
 * Connect to OMERO using the ij plugin
 * Open an image
 * Crop the image,
 * Select Save and Upload
 * Select a project and "new from folder"
 * Check that the cropped image is imported as ome.tiff

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/ImporterUIElement.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/ImporterUIElement.java
@@ -444,7 +444,7 @@ class ImporterUIElement
 				}
 			} else {
 				if (importable.isFolderAsContainer()) {
-					String name = f.getTrueFile().getParentFile().getName();
+					String name = f.getParentName();
 					//first check if the name is already there.
 					Entry<JLabel, Object> entry;
 					Iterator<Entry<JLabel, Object>>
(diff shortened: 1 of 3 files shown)


gusferguson:
@jburel 

Tested using OMERO.insight-ij-5.1.0-452-936e5ef-ice35-b481 and bioformats_package.jar from merge build, user-3, read-annotate-1.

- Opened DV in ImageJ - cropped - Upload and Save to OMERO - get import dialog
- Select existing project - dataset from folder - screenshot 1
- Click Add to Queue - screenshot 2
- Click Import - Import greys out but no Import tab opens and no import takes place - screenshot 3

Not working.

import data

import data

import data

gusferguson:
@jburel 

Tested using OMERO.insight-ij-5.1.0-511-eee50ab-ice34-b483 and merge build bioformats package from today. Trout user-3.

Cropped images are saved back to OMERO as OME-TIFF as expected.
Good to merge.

It is interesting to note that this whole workflow gives a way to circumvent OMERO permissions by opening a read-only image belonging to another user and then saving it back to the server as your own image. Something to bear in mind down the track,

joshmoore:
Thanks, @gusferguson. Merging.

(At the risk of misusing this PR for a discussion, is the workflow not something a la deep copy? So I don't know if I would really consider it circumventing the permission system, but it does allow copying, just when data duplication.)

PR #3753, opened 2015-04-23
Optimize and fix original file paths

joshmoore:
Replaces https://github.com/openmicroscopy/openmicroscopy/pull/3695
----
**Original PR from @chris-allan**:
This PR includes various performance improvements to the stack required to extract original file paths from the server via the `/webgateway/original_file_paths/...` OMERO.web JSON endpoint. Without these or similar changes there is the very real scenario where requesting the original file paths could **crash** your server due to the sheer amount and complexity of data being processed.

The "real world" example dataset which spurred the need for the improvements has 4608 Images (384 Wells, 12 Fields) and 4608 OriginalFiles (one per field) associated with the same Fileset. Half that number is needed to run a server out of memory with a 1GB heap. Call tree is as follows:

```
original_file_paths()
\___ _ImageWrapper.getImportedImageFiles()
\___ _ImageWrapper.getFileset()
\___ _ImageWrapper.countFilesetFiles()
\___ BlitzGateway.getFilesetFilesInfo()
```

The first set of improvements are to the fileset and archive file count methods, `getFilesetFileInfo()` and `getArchivedFilesInfo()`, on the gateway. Before:

```
...
In [25]: print datetime.now()
2015-04-03 08:45:46.464612

In [26]: gw.getFilesetFilesInfo([66101])

Out[26]: {'count': 4608, 'size': 6564240074L}

In [27]: print datetime.now()
2015-04-03 08:45:49.193313
...
```

After (14ee8844e62c434d8c4b827735a3d10424abc1b4):

```
In [12]: print datetime.now()
2015-04-03 14:38:57.178159

In [13]: gw.getFilesetFilesInfo([66101])
Out[13]: {'count': 4608L, 'size': 6564240074L}

In [14]: print datetime.now()
2015-04-03 14:38:57.371813
```

~2.7seconds vs. ~190ms (~14x speedup)

The second set of improvements are to the original file load method `getImportedImageFiles()` on the image wrapper. Before, even with a 2GB heap the aforementioned dataset causes out of memory and GC overhead limit exceptions. This is likely due to the multiplicity of the queries performed and the complexity of the object graph. The number of objects loaded would be >20,000.

After (1356aff4c2f3f18728cce5a0cc67973248285777):

```
In [11]: print datetime.now()
2015-04-03 19:04:43.403131

In [12]: files = list(image.getImportedImageFiles())
print datetime.now()

In [13]: print datetime.now()
2015-04-03 19:04:49.463482
```

~6seconds.

These projection changes complete the speedup.

After (27ce963e1a7e8a97ac346c3d68982f44855a2bf4), run on `localhost`:

```
$ time curl "https://localhost/webgateway/original_file_paths/66101/?server=1&username=...&password=..." &> /dev/null

real	0m0.356s
user	0m0.024s
sys	0m0.037s
```

~350ms

Integration tests cover the affected code paths. /cc @ximenesuk 

To be reviewed by @will-moore and/or @aleksandra-tarkowska.

/cc @jballanc, @knabar, @emilroz, @jburel, @sbesson 

**From @sbesson**:
Quick fire test performed on a large  SPIM dataset (~3600 files) by asking the Web client to display the file paths:
- without this PR, the query is started and the server hangs as described (7GB heap), requiring a restart of the server
- with this PR, the same query on the same dataset with a server with similar memory settings returns within 100ms.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2829,7 +2829,9 @@ class _BlitzGateway (object):
                 '   where i_image.id in (:ids)'\
                 ')'
         queryService = self.getQueryService()
-        count, size = queryService.projection(query, params, self.SERVICE_OPTS)[0]
+        count, size = queryService.projection(
+            query, params, self.SERVICE_OPTS
+        )[0]
         if size is None:
             size = 0
 
@@ -2876,7 +2878,9 @@ class _BlitzGateway (object):
                 '    where i_link.child.image.id in (:ids)'\
                 ')'
         queryService = self.getQueryService()
-        count, size = queryService.projection(query, params, self.SERVICE_OPTS)[0]
+        count, size = queryService.projection(
+            query, params, self.SERVICE_OPTS
+        )[0]
         if size is None:
             size = 0
         return {'fileset': False, 'count': unwrap(count), 'size': unwrap(size)}
@@ -8804,7 +8808,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         for original_file in original_files:
             yield OriginalFileWrapper(self._conn, original_file)
 
-    def getImportedImageFilePaths (self):
+    def getImportedImageFilePaths(self):
         """
         Returns a generator of path strings corresponding to the Imported
         image files that created this image, if available.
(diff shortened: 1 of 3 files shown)


chris-allan:
Integration tests to cover the cases fixed by 230a2be473c92379a72c871af2a445dcf5e8846e now included on `chris-allan/openmicroscopy:optimize-ofp-develop`. Should pull in cleanly and address the scenarios that were noticed by @aleksandra-tarkowska on #3695.

joshmoore:
Pushed.

sbesson:
Retested with a 3K files dataset. Works as expected. Given the state of the integration tests https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/test.integration.gatewaytest.test_fs/TestFileset/, merging this PR for 5.1.1. Thanks all.

PR #3754, opened 2015-04-23
avi

jburel:
Add avi to the list of format to check.

Check that the min/max for avi are set to 0-255
This will only be true if the pixels type is "uint8"

### DIFF
--- a/components/server/src/ome/formats/OMEROMetadataStore.java
+++ b/components/server/src/ome/formats/OMEROMetadataStore.java
@@ -85,7 +85,7 @@ import org.perf4j.StopWatch;
 public class OMEROMetadataStore
 {
     /** List of graphics domains we are checking.*/
-    private static String[] DOMAINS = {"jpeg", "png", "bmp", "gif", "tiff"};
+    private static String[] DOMAINS = {"jpeg", "png", "bmp", "gif", "tiff", "avi"};
 
     /** Logger for this class. */
     private static Logger log = LoggerFactory.getLogger(OMEROMetadataStore.class);


joshmoore:
Kicked travis:
```
[ivy:resolve] 	Server access Error: Connection timed out url=http://artifacts.openmicroscopy.org/artifactory/maven/ome/appbundler/1.0.0/appbundler-1.0.0.jar
```

joshmoore:
@jburel : is there a ticket along with this with testing instructions?

mtbc:
Would be good to know where to find or how to make a suitable AVI and where to check for min/max (rendering settings slider bars?).

will-moore:
Works fine with avi from test_images_good and one generated by Make_Movie export. All 3 channels are 0 - 255.

joshmoore:
@will-moore : and the behavior on latest as a comparison?

will-moore:
On latest, after import (or clicking Min/Max) the channels are NOT all set to 0 - 255, depending on pixel intensity.
Whereas on merge they are set to 0 - 255 for all channels regardless of pixel intensity.

jburel:
sorry forgot to add testing files Thanks @will-moore 
ref card https://trello.com/c/ByPxvCd9/59-avi-msrle-importability

joshmoore:
Maybe as we start refactoring, we try to compute based on https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/AVIReader.java#L157 cc: @melissalinkert 

Merging; thanks guys.

PR #3755, opened 2015-04-23
ResetDefaults cant annotate

will-moore:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12796

Even if you can't annotate an image, you should be able to call resetDefaults() to get the "Imported" rendering settings.

To test:
 - view an image that you can't annotate
 - Try to reset to "Imported settings"
 - Check you get the same as the owner of the image.

Also added a gateway test.


ximenesuk:
This works as described and the test is passing. Good to merge.

PR #3758, opened 2015-04-23
Adding version 5.1.1 version history

hflynn:
Based on draft text from https://trello.com/c/oac6QylW/318-5-1-1-omero-history-txt

PR #3760, opened 2015-04-27
Remove GraphModify2 from history

hflynn:
Update to history as discussed on #3758 cc @joshmoore


sbesson:
@joshmoore: good to go?

joshmoore:
In general, yes, but perhaps time for a general once over so we don't have *too* many history PRs?

hflynn:
--no-rebase

PR #3761, opened 2015-04-27
Fix sync_on_login

joshmoore:
Previously, non-ldap groups were stripped from a user if
sync_on_login was set to true since the newly added `ldap`
flagged was not being checked. By only listing LDAP groups
for potential removal, manually-added, non-LDAP groups can
be maintained.

#### Testing ####

 * Connect your server to a test LDAP instance.
 * Login as a known user and check that they are present in the expected LDAP groups.
 * Create a local, non-LDAP group via `bin/omero group add`
 * Add the user via `bin/omero group adduser --name X --user-name Y`
 * Check in `bin/omero user list` that the user is now in one extra group
 * **Remove the user from one of the expected LDAP groups in the LDAP DB** (tricky)
 * Create a new session as the user `bin/omero login -C Y@localhost`
 * List the groups that the user is a member of `bin/omero user info` and check that the non-LDAP group is still listed.

### DIFF
--- a/components/server/src/ome/logic/LdapImpl.java
+++ b/components/server/src/ome/logic/LdapImpl.java
@@ -296,21 +296,25 @@ public class LdapImpl extends AbstractLevel2Service implements ILdap,
         String ldapDN = getPersonContextMapper().getDn(ldapExp);
         DistinguishedName dn = new DistinguishedName(ldapDN);
         List<Long> ldapGroups = loadLdapGroups(username, dn);
-        List<Object[]> omeGroups = iQuery
+        List<Object[]> currentGroups = iQuery
                 .projection(
-                        "select g.id from ExperimenterGroup g "
+                        "select g.id, g.ldap from ExperimenterGroup g "
                                 + "join g.groupExperimenterMap m join m.child e where e.id = :id",
                         new Parameters().addId(omeExp.getId()));
 
-        Set<Long> omeGroupIds = new HashSet<Long>();
-        for (Object[] objs : omeGroups) {
-            omeGroupIds.add((Long) objs[0]);
+        final Set<Long> currentLdapGroups = new HashSet<Long>();
+        for (Object[] objs : currentGroups) {
+            Long id = (Long) objs[0];
+            Boolean isLdap = (Boolean) objs[1];
+            if (isLdap) {
+                currentLdapGroups.add(id);
+            }
         }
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Depending on our testing, we may decide to remove the following for 5.1.2:
```
# .. warning::
#    Currently setting this to true the user will be
#    removed from any groups to which they have been
#    added outside of LDAP! Please use carefully.
```

cc: @hflynn @pwalczysko

hflynn:
That is in the autogen docs built from the codebase so would need https://github.com/openmicroscopy/ome-documentation/pull/1196 updating.

hflynn:
Ah actually, this is on the known limitations page, I'll make a card.

pwalczysko:
Used ``jsmith2`` belonging to 2 LDAP groups (``omedev`` and ``haxed``) for the test. Created a new non-LDAP group ``test-josh`` and went through the scenario. The behaviour was as expected. The "removal" from the LDAP group(s) (checked both) was approximated by changing the name in JExplorer to ``jsmith2x`` under the particular group.
  - after login was done, the changes in JExplorer (=LDAP) were reflected in OMERO
  - the membership of the new non-LDAP group was respected every time
  - after reversal of the name change (i.e. ``jsmith2x`` -> ``jsmith2``) and thus return to the LDAP groups the user was again listed in the LDAP groups and was able even to view his data there 
  - tested on the cmd line but also using Insight for the logins and group/user checks

All works here as expected.
(interesting points were made during a discussion with @joshmoore as listed on https://trello.com/c/NdKek7Ag/303-ldap - probably for another PR (?))

joshmoore:
Thanks, @pwalczysko. At this point, there's nothing that I would consider a **bug**, i.e. "loss of data", but there will certainly be surprises for the user unless there are subsequent client changes. E.g.:
 * group owner users OMERO to add a user to their lab group which is defined in LDAP
 * that user logs in and is *not* a member of the group

This would require a check for `group.ldap is True` (and possibly `sync_on_login is True`) in administration panels. We'll need to decide between:
 * add this fix but leave the warning
 * add this fix and remove the warning, but include a lesser warning (elsewhere?)
 * add this fix and fix the clients as well

cc: @jburel

joshmoore:
General thought is to leave the warning for the moment until we can better prepare the LDAP testing (~5.1.3)

PR #3771, opened 2015-04-29
Disable delete key in jsTree

will-moore:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12831 and http://trac.openmicroscopy.org.uk/ome/ticket/12840 (also see https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7785).

To test, use a keyboard with a "Delete" key:
 - Check that Delete doesn't do anything to the tree main webclient page, shares or tags page.


mtbc:
Works fine, good to merge.

(RFE: Would be nice someday if it does the same as Delete from the context menu.)

joshmoore:
@will-moore : leaving you to file the RFE as/if appropriate. Merging.

PR #3772, opened 2015-04-29
Viewer Interpolate checkbox

will-moore:
Fixes http://trac.openmicroscopy.org/ome/ticket/12828 (see https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7786).

 To test:
 - Zoom in a lot in web image viewer
 - Toggle the "Interpolate" checkbox (checked by default).
 - When unchecked, image should appear pixelated.
 - Compare with same image in Insight, to check this pixelation is valid.

screen shot 2015-04-29 at 14 08 45

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/css/ome.viewport.css
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/css/ome.viewport.css
@@ -157,3 +157,15 @@ div.weblitz-viewport-tiles img {
   color: white;
 }
 
+/* from http://phrogz.net/tmp/canvas_image_zoom.html */
+.pixelated {
+  image-rendering:optimizeSpeed;             /* Legal fallback */
+  image-rendering:-moz-crisp-edges;          /* Firefox        */
+  image-rendering:-o-crisp-edges;            /* Opera          */
+  image-rendering:-webkit-optimize-contrast; /* Safari         */
+  image-rendering:optimize-contrast;         /* CSS3 Proposed  */
+  image-rendering:crisp-edges;               /* CSS4 Proposed  */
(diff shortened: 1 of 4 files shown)


mtbc:
Works nicely in Firefox 31.6 and Chrome 42.0.

if the pixelation is valid is an interesting judgment call; unfortunately, Insight zoom seems to max out at 300%. At that level:

Chromium: 
chromium
Firefox: 
firefox
Insight: 
insight

Doesn't look obviously wrong to me. At least, it gets smoothed over (excellently in web) when the interpolation box is checked!

gusferguson:
@joshmoore - seems to be a relatively simple step towards aligning the clients. I would have to look at the whole thing in more depth to see what to do longer term. Interpolation as an issue seems to run through a number of workflows now. I will update Help as needed if this goes into 5.1.2.

joshmoore:
Thanks, @gusferguson.

aleksandra-tarkowska:
@will-moore Can I ask how it was tested with big images? I can't see a difference.

mtbc:
I am sorry, I don't know what I was thinking that day! `worm.tiff` is not big enough to count as a big image. On a *real* big image, I don't see the difference either. Maybe reopen ticket?

aleksandra-tarkowska:
I would vote after hiding this option as you cannot zoom in to the level where you have pixelated tiles. PanoJS viewer upscale tiles up to 400% that never been reported as a poor quality. I think we should hide that option for tiled pyramids thats all.

will-moore:
Sorry - completely forgot to apply this to PanoJS viewer.
But I think it is applicable, although maybe not widely requested.
If PanoJS allows you to zoom to 400% then we should allow this to be non-interpolated.

will-moore:
Fixed for big images in https://github.com/openmicroscopy/openmicroscopy/pull/3796.

PR #3777, opened 2015-04-30
Apache stanza fcgi

sbesson:
See https://github.com/openmicroscopy/openmicroscopy/pull/3769#issuecomment-97706639

This PR users `/var/run/omero.fcgi` as the default filename for the `FastCGIExternalServer` directive in the Apache stanza. This change is expected to reduce the number of  `Permission Denied` issues encountered when upgrading a server and starting OMERO.web in production as highlighted byhttps://github.com/openmicroscopy/ome-documentation/pull/1203. Since `/var/run` exists on all UNIX-like platforms and should be readable/executable by the Apache agents, this filename was chosen as the lowest common denominator for all platforms. 

This PR should be tested both under Apache 2.2 and 2.4 by downloading the server, generating the Apache stanza via `bin/omero web config` and starting OMERO.server and OMERO.web. With this PR included, the expectation is that the Web client will be able to start without any need to modify permissions/ownerships under the server directory.

### DIFF
--- a/components/tools/OmeroPy/test/unit/clitest/test_web.py
+++ b/components/tools/OmeroPy/test/unit/clitest/test_web.py
@@ -174,22 +174,20 @@ class TestWeb(object):
 
         if prefix:
             missing = self.required_lines_in([
-                ('FastCGIExternalServer ',
-                 '/var/run/omero.fcgi" -host %s -idle-timeout 60'
-                 % expected_cgi),
+                ('FastCGIExternalServer "/var/run/omero.fcgi" -host %s'
+                 ' -idle-timeout 60' % expected_cgi),
                 ('Alias %s/error ' % prefix, 'etc/templates/error'),
                 ('Alias %s ' % static_prefix[:-1],
                  'lib/python/omeroweb/static'),
-                ('Alias %s "' % prefix, '/var/run/omero.fcgi/"'),
+                ('Alias %s "/var/run/omero.fcgi/"' % prefix),
                 ], lines)
         else:
             missing = self.required_lines_in([
-                ('FastCGIExternalServer ',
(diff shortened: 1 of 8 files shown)


components/tools/OmeroPy/test/unit/clitest/test_web.py
```diff
@@ -175,19 +175,21 @@ def testApacheConfig(self, prefix, cgihost, cgiport, capsys, monkeypatch):
...
             missing = self.required_lines_in([
                 ('FastCGIExternalServer ',
-                 'var/omero.fcgi" -host %s -idle-timeout 60' % expected_cgi),
+                 '/var/run/omero.fcgi" -host %s -idle-timeout 60'
+                 % expected_cgi),
                 ('Alias %s/error ' % prefix, 'etc/templates/error'),
                 ('Alias %s ' % static_prefix[:-1],
                  'lib/python/omeroweb/static'),
-                ('Alias %s "' % prefix, 'var/omero.fcgi/"'),
+                ('Alias %s "' % prefix, '/var/run/omero.fcgi/"'),
```
manics: Since there's no longer an unknown prefix on `var` you could change this to the non-tuple form:
`'Alias %s "/var/run/omero.fcgi/"' % prefix`

(https://github.com/sbesson/openmicroscopy/blob/apache_stanza_fcgi/components/tools/OmeroPy/test/unit/clitest/test_web.py#L80)

And similarly for the next two changes.

sbesson: Pushed

manics:
Good to merge

joshmoore:
Merging. Anything needed in the way of doc updates? cc: @hflynn

PR #3780, opened 2015-04-30
assist #12856: add deletion integration test for projections

mtbc:
Review the code to check that the scenario that the test recreates corresponds with the bug described in http://trac.openmicroscopy.org/ome/ticket/12856. Try running it to see that it fails on the line you would expect.

Cc: @will-moore 
--no-rebase

### DIFF
--- a/components/tools/OmeroJava/test/integration/delete/HierarchyDeleteTest.java
+++ b/components/tools/OmeroJava/test/integration/delete/HierarchyDeleteTest.java
@@ -36,12 +36,16 @@ import omero.model.Annotation;
 import omero.model.CommentAnnotationI;
 import omero.model.Dataset;
 import omero.model.DatasetI;
+import omero.model.DatasetImageLink;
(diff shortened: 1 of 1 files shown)


will-moore:
Test looks fine. Just one question:
Is it sufficient to link 2 images by Instrument to simulate projected image, since they will also share channels etc?
Is it worth trying to use pixelsService.copyAndResizeImage() to create the second image, after linking Instrument to the first?

mtbc:
It at least seems enough to trigger the wrong behavior. I wanted to avoid using `pixelsService.copyAndResizeImage()` in case we go ahead and change it so that new images don't have those shared subgraphs, but still have legacy images to live with. (The bug is actually caused by the `relatedTo` among the pixels objects, even the shared instrument isn't making a difference.)

joshmoore:
Assuming this is sufficient for the moment. Merging.

PR #3781, opened 2015-04-30
Map annotation of Wells

will-moore:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12857

To test:
 - Add map annotation to well (saves to Image in well)
 - Reload well to check it has been saved
 - Try choosing different Field index to check annotation is added to correct index.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/mapannotations.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/mapannotations.html
@@ -96,6 +96,11 @@
                 $mapAnnotationSpinner = $(".mapAnnotationSpinner", $table.parent()),
                 tableData = [];
 
+            {% if index %}
+            // for wells, we need to know index
+            data['index'] = "{{ index }}";
+            {% endif %}
+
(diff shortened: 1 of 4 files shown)


gusferguson:
@will-moore 

Tested using https://trout.openmicroscopy.org/merge/webclient/ user-3 and user-12 data read-annotate-1

Able to add Key-value Pairs to wells of own data and other user's in read-annotate group. Copy and paste works as expected. When added to single field of multi-field well is only shown on the added field as expected. 

This was confusing to test because of the reversion to the "Run" view when changing fields, so it gave the impression the K-VPs kept disappearing. Is it worth my putting in an RFE for focus to be retained on the selected well when changing fields, or is this technically not possible?

Good to merge.

safari

will-moore:
@gusferguson It's not perfect, but that should fix well selection on changing of index.
You'll notice that the Acquisition panel loads to the right first (since this is still selected in the tree), then the selected well re-loads.

pwalczysko:
Sorry, did not manage today. Please relist for tomorrow with my name.

pwalczysko:
Works as expected including the re-load of the well and keeping the selection of well when going to different field.

Ready to merge.

PR #3785, opened 2015-05-02
Download single files

will-moore:
This allows the downloading of individual files from a set of files selected for download.

To test:
 - Select 1 or more images, check the tooltip on "Download..." menu item for number of files that will download.
 - If more than 1, then clicking "Download" will open download placeholder dialog
 - You will see the option to download all files as zip (same as before)
 - Also should be a list of files, clicking on one will download it directly.
 - If multiple filesets are selected, files will be grouped in filesets and fileset indicated
   on hover (see screenshot).
 - Clicking on Zip, should show a "creating zip" message and hide everything else.

screen shot 2015-05-02 at 23 22 53

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/download_placeholder.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/download_placeholder.html
@@ -58,6 +58,20 @@ Download files
 <h1 class="downloading">Please <a href="#" onClick="window.close()">close this window</a> once the download is complete.</h1>
 </div>
 
+{% if fileLists %}
+<table>
+    {% for fs in fileLists %}
+    <tbody>
+        {% for f in fs %}
+        <tr>
+            <td><a href="{% url 'get_original_file' f.id %}">{{ f.name }}</a></td>
+        </tr>
+        {% endfor %}
+    </tbody>
+    {% endfor %}
+</table>
+{% endif %}
+
 <script>
 
 
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/download_placeholder.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/download_placeholder.html
@@ -53,7 +53,7 @@ Download files
             padding: 5px;
         }
         td:hover {
-            background-color: #ccccff;
+            background-color: hsl(215,20%,75%);
         }
         td .download {
             visibility: hidden;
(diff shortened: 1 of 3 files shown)


joshmoore:
* travis didn't like `./components/tools/OmeroWeb/omeroweb/webclient/views.py:2720:31: W291 trailing whitespace`
 * Perhaps include the size of each file in the listing?

pwalczysko:
Sorry, did not manage today. Please relist tomorrow with my name.

pwalczysko:
The first tooltip (the one popping up when the ``Download...`` menu item is selected) is

  - not fully correct (saying ....as .zip, suggesting the .zip is the only option, which is not true)
  - running out of the screen a bit because Firefox tooltip is like that 

Maybe former point is one to fix, latter is something we can live with ?

pwalczysko:
It would do no harm to reiterate the info about the Download being a download of ``original format files`` also in the dialog, not just in the tooltip.

pwalczysko:
In summary, 2 RFEs of dialog texts https://github.com/openmicroscopy/openmicroscopy/pull/3785#issuecomment-99417323 and https://github.com/openmicroscopy/openmicroscopy/pull/3785#issuecomment-99419442

will-moore:
Updated tooltip as suggested and changed title to dialog to "Download Original Files" except when downloading Jpeg etc.
To test, select several images and "Download as JPEG".
Compare this with selecting a single image with multi-file fileset and "Download..." original file.

pwalczysko:
This works as expected. The ``Download oritinal files`` case is consistent and it says the info in tooltip, dialog and suggested filenames.

But the ``Export as...`` options have a bit of  a "quarrel" between the ``Export`` word used in the menu, ``Download`` word used in tooltip and ``Download`` words used everywhere in the dialog. Additionally, the suggested file name in the dialog goes ``Save_as.jpeg`` which adds a third confusing term to the bargain.
Would it be possitle to unify the Export option so that everything there says "Export" ?

will-moore:
@pwalczysko Used 'Export' in dialog when exporting jpeg/png etc.

screen shot 2015-05-11 at 12 04 11

aleksandra-tarkowska:
All looks good and works as expected. Just a small comment. Could we not introduce yet another color to the UI. Could you use grey-blue from .jstree-default .jstree-clicked?
firefoxscreensnapz039

will-moore:
Using jsTree selected colour now. 

screen shot 2015-05-12 at 12 30 08

pwalczysko:
Color is now matching. The dialog messages and names of files are matching.

Ready to merge.

PR #3788, opened 2015-05-05
updating icon

aleksandra-tarkowska:
The icon for Remove Content in Public tab in Web (share) was the same as the Delete icon under Data tab. Now it uses *scissors* to make clear the Delete action from Share is not possible
firefoxscreensnapz035


mtbc:
Good idea.
pr-3788
Good to merge.

mtbc:
button leads sensibly to,
pr-3788

joshmoore:
@gusferguson : anything worrisome screenshot-wise?

gusferguson:
@joshmoore - thanks - I think I will have to update one or two - I will put it on the list.

joshmoore:
:thumbsup:

PR #3743, opened 2015-04-21
CLI: review debug help

sbesson:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/1293

This PR reviews the help for the debug topic of the CLI, bringing it in line with the subcommands help. It also removes the reference to the deprecated `DEBUG` environment variable.

To test this PR, review the output of

```
bin/omero help debug
```

PR #3789, opened 2015-05-05
CLI: add OMERO_USERDIR and fix profile debugging

sbesson:
See https://trello.com/c/IT01phB4.

This commit introduces more flexibility over the control of the OMERO user directory in the CLI by introducing the `OMERO_USERDIR` environment variable. Usages of `path(get_user_dir)/"omero"` are replaced by a new `get_omero_userdir()` utility function which implements the logic for reading the variable and returns a `path.path` object.

Additionally, this PR fixes the bug reported in https://github.com/openmicroscopy/openmicroscopy/pull/3743#issuecomment-95067341 when running the profile debugging mode within a directory not owned by the current user by storing the debugging profile under the OMERO user directory.

Unit tests are extended to cover the semantics of `OMERO_USERDIR` in the CLI sessions plugin and make sure the sessions directory variables (`OMERO_SESSIONDIR`, `OMERO_SESSIONS_DIR`) are still taking precedence if set.


joshmoore:
Running help and login with and without `OMERO_USERDIR`, with and without `-dp`:
```
[jamoore@ome-c6100-3 ~]$ ls myomero
hotshot_edi_stats  sessions
[jamoore@ome-c6100-3 ~]$ ls omero
hotshot_edi_stats  sessions  tmp
```

joshmoore:
Questions:
 * would basing `OMERO_TEMPDIR` on `OMERO_USERDIR` be in scope?
 * if so, would a `omero help env` make sense?

NB:
```
src/omero/util/temp_files.py:158:                omero_dir = path(target) / "omero"
src/omero/util/temp_files.py:161:                        """"omero" is not a directory: %s""" % omero_dir)
src/omero/util/temp_files.py:210:        return path(choice) / "omero" / "tmp"
```
Testable via `OMERO_TEMPDIR=/tmp/foo bin/omero admin diagnostics`

sbesson:
`OMERO_TEMPDIR` is certainly worth modifying to derive from `OMERO_USERDIR` if the commits above make sense. I will push a commit to that effect.

`bin/omero help env` as a way to list all environment variables that can be configured for the CLI and their usage scope is certainly doable as well unless @ximenesuk or @hflynn think of a better placeholder for this information,

sbesson:
Last commits should:
- modify `omero.util.temp_files` to use `get_omero_userdir` as the default OMERO user's directory
- deprecate the `OMERO_TEMPDIR` variable in favor of `OMERO_TMPDIR` which sets the temporary directory directly (instead of creating the `OMERO_TEMPDIR/omero/tmp` hierarchy)
- add unit tests to cover the selection logic for the temporary folder with the various environment variables.

While pushing the last commit, I realize that https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/util/TempFileManager.java should likely also be modified in a follow up PR to follow the same semantics if we accept the last changeset. /cc @ximenesuk @mtbc 

Assuming we agree on the last changes, I can push another commit which adds an `env` (`environment`?) topic to the CLI help with the new `OMERO_TMPDIR`.

joshmoore:
Pre-testing looks good. Definitely will need a second round of Java changes to include insight and the server. cc: @jburel

joshmoore:
Likely need to update the `diagnostics` command to include `USERDIR` and `TMPDIR`:
```
[hudson@ome-c6100-3 dist]$ bin/omero admin diagnostics | grep DIR
[Environment:OMERO_TEMPDIR=(unset)
[hudson@ome-c6100-3 dist]$ OMERO_USERDIR=/tmp/foo bin/omero admin diagnostics | grep DIR
Environment:OMERO_TEMPDIR=(unset)
[hudson@ome-c6100-3 dist]$ OMERO_TEMPDIR=/tmp/foo bin/omero admin diagnostics | grep DIR
/opt/hudson/workspace/OMERO-5.1-merge-build/src/dist/lib/python/omero/util/temp_files.py:141: DeprecationWarning: OMERO_TEMPDIR is deprecated. Use OMERO_TMPDIR instead.
  DeprecationWarning)
Environment:OMERO_TEMPDIR=/tmp/foo
[hudson@ome-c6100-3 dist]$ OMERO_TMPDIR=/tmp/foo bin/omero admin diagnostics | grep DIR
Environment:OMERO_TEMPDIR=(unset)

```

joshmoore:
Use of `TMPDIR` itself seems fine though:
```
[hudson@ome-c6100-3 dist]$ export PYTHONPATH=lib/python

[hudson@ome-c6100-3 dist]$ export OMERO_TEMPDIR=t1
[hudson@ome-c6100-3 dist]$ python lib/python/omero/util/temp_files.py dir
lib/python/omero/util/temp_files.py:141: DeprecationWarning: OMERO_TEMPDIR is deprecated. Use OMERO_TMPDIR instead.
  DeprecationWarning)
t1/omero/tmp/omero_hudson/30200

[hudson@ome-c6100-3 dist]$ export OMERO_USERDIR=t1
[hudson@ome-c6100-3 dist]$ python lib/python/omero/util/temp_files.py dir
t1/tmp/omero_hudson/31563

[hudson@ome-c6100-3 dist]$ export OMERO_TMPDIR=t1
[hudson@ome-c6100-3 dist]$ python lib/python/omero/util/temp_files.py dir
t1/omero_hudson/30550
```

sbesson:
Last 2 commits should:
- log the environment variables in `diagnostics` (should we add `OMERO_SESSIONDIR`)
- add a new `env` help topic, see `bin/omero help env`

joshmoore:
@jburel : any worries from your side about being able to make use of the same variables in Java?

jburel:
The one we can take advantage of is the ```OMEROUSER_DIR```
This will require few adjustments in the init sequence since the log service is initialised before any connection.

If we also decide to use TMP DIR, this will require further adjustments

components/tools/OmeroPy/test/unit/clitest/test_sessions.py
```diff
@@ -50,8 +50,12 @@ def testDefaultSessionsDir(self):
...
 
     @pytest.mark.parametrize('environment', (
-        [], ["OMERO_SESSIONS_DIR"], ["OMERO_SESSIONDIR"],
-        ["OMERO_SESSIONS_DIR", "OMERO_SESSIONDIR"]))
+        [],
+        ["OMERO_SESSIONS_DIR"], ["OMERO_SESSIONDIR"], ["OMERO_USERDIR"],
+        ["OMERO_SESSIONS_DIR", "OMERO_SESSIONDIR"],
+        ["OMERO_SESSIONS_DIR", "OMERO_USERDIR"],
+        ["OMERO_SESSIONDIR", "OMERO_USERDIR"],
+        ["OMERO_SESSIONS_DIR", "OMERO_SESSIONDIR", "OMERO_USERDIR"]))
```
manics: What are you testing with `OMERO_SESSIONS_DIR`? It doesn't seem to be in the main code.

manics:
```
OMERO_TEMPDIR=x/z OMERO-51/bin/omero -d3 admin status
Returning popen: ['icegridadmin', '--Ice.Config=/Users/simon/omero-test/OMERO-51/etc/internal.cfg', '-e', 'node ping master']
```
creates the temp dir but doesn't output the deprecated warning, should it?

```
OMERO_TEMPDIR=x/z OMERO_TMPDIR=x/y OMERO-51/bin/omero ...
```
Uses `OMERO_TEMPDIR`, I think if they're both defined `OMERO_TMPDIR` should have priority (and output a warning).

manics:
```
OMERO_USERDIR=x/a OMERO_SESSIONDIR=x/b OMERO_TMPDIR=x/c OMERO_TEMPDIR=x/d OMERO-51/bin/omero -dp admin status
```
fails with `IOError: [Errno 2] No such file or directory: 'x/a/hotshot_edi_stats'` (since it hasn't been created as part of the the session or tmp dirs). `WONTFIX` is an acceptable response since this is a unlikely to occur in practice.

sbesson:
On the `OMERO_TMPDIR` vs `OMERO_TEMPDIR` precedence order, you are absolutely correct. I will invert the logic similarly to what is done for `OMERO_SESSIONDIR` vs `OMERO_SESSION_DIR`.

For warnings, are you using Python 2.7? If so, I realized that 2.7 does not output deprecated warning [by default](https://docs.python.org/2/whatsnew/2.7.html#changes-to-the-handling-of-deprecation-warnings). This raises the question of whether using `DeprecationWarning` is the most approriate call as opposed to a `self.ctx.err()` /cc @ximenesuk @joshmoore .

Once we have made a decision on the warnings, I will push commits to fix the precedence and the tests.

manics:
> For warnings, are you using Python 2.7? If so, I realized that 2.7 does not output deprecated warning by default. This raises the question of whether using DeprecationWarning ...

Yes, 2.7. My interpretation of the docs is that `DeprecationWarning` is for deprecated API features used by developers, whereas these env-vars are user features.

sbesson:
Pushed commits fixing the envvar precedence. I kept the usage of `DeprecationWarning` but happy to fill a card to discuss our deprecation policy across Python components (although I feel this is outside the scope of this PR).
Also, I propose we  handle the various permission issues in a follow up PR if that makes sense. It is not clear to me whether settings invalid `OMERO_USERDIR/OMERO_SESSIONDIR...` locations should fail the command or rather there should be an attempt to fallback on default directories.

components/tools/OmeroPy/src/omero/util/temp_files.py
```diff
@@ -131,14 +131,33 @@ def tmpdir(self):
...
+        if custom_tmpdir_deprecated:
+            default_tmpdir = path(custom_tmpdir_deprecated) / "omero" / "tmp"
+        custom_tmpdir = os.environ.get('OMERO_TMPDIR', default_tmpdir)
+
+        # List target base directories by order of precedence
+        targets = []
+        if custom_tmpdir:
+            targets.append(path(custom_tmpdir))
+        targets.append(get_omero_userdir() / "tmp")
+        targets.append(path(tempfile.gettempdir()) / "omero" / "tmp")
```
manics: RFE for discussion: Use `tempfile.mkdtmp()` to create a random temp directory instead of `tempfile.gettempdir()) / "omero" / "tmp"`?

manics:
RFE: Use `get_omero_userdir` in tests instead of `get_user_dir` (and deprecate method?).

I'm happy for this to be merged with RFEs/discussion handled elsewhere.

sbesson:
I would say let's handle the RFEs in follow-up PRs.

joshmoore:
Agreed & merging. This will need to be carefully detailed in the upgrade announcement. cc: @hflynn

PR #3790, opened 2015-05-05
10661 dynamic params

joshmoore:
See: http://trac.openmicroscopy.org.uk/ome/ticket/10661#comment:7

#### Background ####

An interesting recent bug lead to the review of the `JobParams` creation. The user-4 on merge/trout was receiving an exception when trying to run `Make_Movie.py` since the chosen color was not a member of a list of the form `["White", "Red", None, "Black", None, None, None]`. What seems to have happened is that some exception (`OutOfMemory`?) caused the params themselves to be corrupted. The reason that this was only happening for user-4 (and his/her group) was that `JobParams` were being created **per group**. This unnecessary overhead was a secondary bug.

#### Implementation ####

Rather than trying to fix the 2 underlying bugs caused by caching params for each group (and trying to clean up old params, etc.), ticket 10661 suggested reloading all scripts on server start-up. This is now possible along with 2 configuration options under the `omero.scripts.cache.` namespace:

 * `spec` determines the settings for the Guava cache, i.e. how quickly will the objects be freed from memory
 * `cron` how often will a full reloading of all the scripts be attempted. One run is always performed on startup so that the initial access by any user is not delayed.

#### Functional differences ####

 * The general user should notice no differences in behavior.
 * Developers will notice that `parsejob` instances are no longer being persisted. This could conceivably be viewed as a breaking change, but considering that the job itself consists of a byte array containing the serialized form of an Ice object, it's unlikely this was of use.
 * Any script with the namespace "dynamic" (TBD) will **not** be cached, and instead will be reprocessed on each invocation. This will include sizeable overhead.
 * (**Planned**) Disabling the new configuration options will rollback to using the previous `ParamsHelper` method of storing `parsejobs` in the database.

cc: @manics @pwalczysko @rleigh-dundee @ctrueden

### DIFF
--- a/components/blitz/build.xml
+++ b/components/blitz/build.xml
@@ -15,54 +15,25 @@
             <delete file="${basedir}/.sconsign.dblite"/>
         </target>
 
-	<target name="compile" depends="generate,_compile" unless="skip.compile"/>
-	<target name="_compile" description="Internal develop target which doesn't run codegen">
-            <if>
-                <uptodate targetfile="${target.dir}/blitz.jar">
-                    <srcfiles dir="${basedir}" includes="**/*.java,**/*.ice,**/*.vm"/>
-                </uptodate>
-            <then>
-                <echo>Up to date</echo>
-            </then>
-            <else>
-                <antcall target="icegen" inheritRefs="true" inheritAll="true"/>
-		<hard-wire file="ome/services/blitz/fire/SessionManagerI.java"
-		    token="&quot;ome.security.basic.BasicSecurityWiring&quot;"/>
-
-                <!-- WORKAROUND: Attempting to compile mutually dependent directories
-                     via javac is quite complicated. Moving the difficult file out of the way -->
-                <property name="file.real.sessionmanager" value="${src.dest}/ome/services/blitz/fire/SessionManagerI.java"/>
-                <property name="file.fake.sessionmanager" value="${src.dest}/ome/services/blitz/fire/SessionManagerI.fake"/>
-                <move file="${file.real.sessionmanager}" tofile="${file.fake.sessionmanager}"/>
-		<myjavac>
-		    <src path="${src.dir}"/>
-		    <src path="${src.dest}"/>
-		    <src path="${basedir}/generated"/>
-		    <include name="pojos/**"/>
-		    <include name="omero/**"/>
-		    <include name="omero/model/**"/>
-		    <include name="ome/services/blitz/impl/CloseableServant.java"/>
-		    <include name="ome/services/blitz/util/ConvertToBlitzExceptionMessage.java"/>
-		    <include name="ome/services/blitz/util/ObjectFactoryRegistry.java"/>
-		    <include name="ome/formats/model/UnitsFactory.java"/>
-		</myjavac>
-		<myjavac>
-		    <src path="${src.dir}"/>
-		    <src path="${src.dest}"/>
-		    <src path="${basedir}/generated"/>
-                    <exclude name="omero/**"/>
-                    <exclude name="omero/model/**"/>
-		</myjavac>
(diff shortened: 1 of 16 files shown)


joshmoore:
Fix pushed for user scripts. Removed breaking.

joshmoore:
Dynamic scripts support pushed. See https://github.com/openmicroscopy/openmicroscopy/commit/9946644bc41334e5b29c5ff35136c5b404c0702f

joshmoore:
All `(not broken)` integration tests passing as of https://github.com/joshmoore/openmicroscopy/commit/3287242035f4da16e9cb8f8ac8f357a1682edefc

manics:
Tested with https://gist.github.com/manics/832c5ebcdf8c10f0c87b (a single parameter which is a list of datasets).

Dynamic parameters are recalculated and `namespaces` can already be passed as a keyword argument to scripts.client which makes it really easy to use. However if I run the above script, then create a new dataset, the parameters list is recalculated but if I choose the new one and run the script I get an error:
```
Dynamic Test Invalid parameters: VALUE LIST for "Object": Dataset:252 test 2 not in ['Dataset:251 test', 'Dataset:2 test-0256', 'Dataset:1 test-0128'] 
```
which suggests it's still validating the new parameter against an older cached list.

joshmoore:
I saw the expected behavior from the command-line, but logged in as root. Writing an integration test with your gist (to be pushed later), I realized the problem: the `getParams` method is happening in the context of the root/system user for official scripts.

joshmoore:
Pushed a fix, @manics.

manics:
It's still not working consistently for me- it worked once after renaming a dataset, but not on subsequent additions/renames.

joshmoore:
Can you help me to modify the test so that it fails?

manics:
Can you cherry-pick https://github.com/manics/openmicroscopy/commit/58198743f4b9aa7420376c19f0511e95236da729 https://github.com/manics/openmicroscopy/commit/5e6740a857352fb945b7e300d322288e6c9cbc5e

```
OmeroPy$ ./setup.py test -v -t test/integration/test_scripts.py -k testDynamicUpdate
...

        contains_uuids(False, False)
        with pytest.raises(omero.ValidationException):
            run_with_param(uuid1)

        update.saveObject(self.new_dataset(name=uuid1))
        contains_uuids(True, False)
>       run_with_param(uuid1)

...
    def runScript(self, scriptID, inputs, waitSecs, _ctx=None):
>       return _M_omero.api.IScript._op_runScript.invoke(self, ((scriptID, inputs, waitSecs), _ctx))
E       ValidationException: exception ::omero::ValidationException
E       {
E           serverStackTrace =
E           serverExceptionClass =
E           message = Invalid parameters:
E       VALUE LIST for "Object": 7c88c770-a6ea-4f45-9ebe-d12504f65fb3 not in ['<No objects found>']
E
E       }
```

joshmoore:
Done & locally testing. Thanks.

joshmoore:
With this again not in today's build, I'd propose we push this to 5.1.3.

joshmoore:
Build fix in `blitz/build.xml` should allow this to be included by travis. cc: @sbesson @manics

joshmoore:
@dominikl / @jburel : https://github.com/openmicroscopy/openmicroscopy/pull/3899 will likely conflict with https://github.com/joshmoore/openmicroscopy/commit/93f68cbe899cf37af64e57c84953dce1566a5761 -- happy to discuss.

dominikl:
@joshmoore Shouldn't be a huge conflict, I just have single line change in build.xml, the other changes won't conflict imho.

manics:
Looks good to me

pwalczysko:
Retested Make Movie just now. Works fine.

sbesson:
Given all comments above, merging in order to fix the conflict with the decoupling branch. Thanks all.

PR #3791, opened 2015-05-05
Comment Delete2 dryRun

will-moore:
This is a slight improvement to the way we handle Comment deletion.

When a Comment is removed from an object, we need to check whether it is linked to any other objects. If it isn't then we delete the Comment itself.

Previously, we simply looked up links to Projects, Datasets, Images etc in turn.
Now, we use a Delete2 with ```dryRun = True``` to see if the Comment delete would have deleted any links to P/D/I etc.

To test, select several objects, add a Comment.
Then remove the comment from each in turn, checking the DB to see that it is deleted only after being removed from last object.

Also added integration test for this behaviour.

### DIFF
--- a/components/tools/OmeroPy/test/integration/test_delete.py
+++ b/components/tools/OmeroPy/test/integration/test_delete.py
@@ -231,6 +231,55 @@ class TestDelete(lib.ITest):
         callback = self.waitOnCmd(self.client, handle)
         callback.close(True)
 
+    def testDeleteComment(self):
+        comment = omero.model.CommentAnnotationI()
+        comment = self.update.saveAndReturnObject(comment)
+        images = list()
+        # Single Comment linked to 3 Images and 3 Datasets
+        for i in range(0, 3):
+            img = self.new_image(name="testDeleteComment")
+            img.linkAnnotation(comment)
+            images.append(self.update.saveAndReturnObject(img))
+            ds = self.make_dataset("testDeleteComment")
+            ds.linkAnnotation(comment)
+            self.update.saveAndReturnObject(ds)
+
+        cid = comment.id.val
+        assert self.query.find("CommentAnnotation", cid)
+
+        # Remove comment from first image
+        linkIds = []
+        for l in images[0].copyAnnotationLinks():
+            linkIds.append(l.id.val)
+        command = Delete2(targetObjects={"ImageAnnotationLink": linkIds})
+        handle = self.client.sf.submit(command)
+        self.waitOnCmd(self.client, handle)
+
+        # Delete Dry Run...
+        command = Delete2(targetObjects={"CommentAnnotation": [cid]},
+                          dryRun=True)
+        handle = self.client.sf.submit(command)
+        self.waitOnCmd(self.client, handle)
+
+        # ...Should tell us that remaining links will be deleted
+        rsp = handle.getResponse()
+        assert ('ome.model.annotations.ImageAnnotationLink'
+                in rsp.deletedObjects)
+        links = rsp.deletedObjects['ome.model.annotations.ImageAnnotationLink']
+        assert len(links) == 2
+        dlnks = rsp.deletedObjects[
+            'ome.model.annotations.DatasetAnnotationLink']
+        assert len(dlnks) == 3
+
+        # Comment should not yet be deleted
+        assert self.query.find("CommentAnnotation", cid)
+
+        # Finally, delete Comment
+        command = Delete2(targetObjects={"CommentAnnotation": [cid]})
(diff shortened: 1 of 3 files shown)


mtbc:
The approach looks right-headed.

joshmoore:
Tested via:
```
omero -q login pr3791@localhost
I="Image:2563" # Imported fake
D=$(omero -q obj new Dataset name=cli-test)
C=$(omero -q obj new CommentAnnotation textValue=cli-test)
L1=$(omero -q obj new DatasetAnnotationLink parent=$D child=$C)
L2=$(omero -q obj new ImageAnnotationLink parent=$I child=$C)
```

Both deleting from the dataset and from the image works fine and leaves the other instance(s) alone.

In the case of a comment annotation only one one object (including after a previous deletion on a second object), the comment is actually removed.

:+1:

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -1257,11 +1258,19 @@ def remove(self, parents, index, tag_owner_id=None):
                         self.conn.deleteObjectDirect(al._obj)
                 # if comment is orphan, delete it directly
                 orphan = True
+
+                # Use delete Dry Run...
+                cid = self.comment.getId()
+                command = Delete2(targetObjects={"CommentAnnotation": [cid]},
+                                  dryRun=True)
+                handle = self.conn.c.sf.submit(command)
```
joshmoore: NB: `sf.submit() && conn._waitOnCmd()` can be shortened to `cb = self.conn.c.submit()`

joshmoore: Bug: This handle is never closed.

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -1257,11 +1258,19 @@ def remove(self, parents, index, tag_owner_id=None):
...
+                self.conn._waitOnCmd(handle)
+                # ...to check for any remaining links
+                rsp = handle.getResponse()
```
joshmoore: This makes another remote call to the server, whereas `cb.getResponse` uses a locally cached copy.

components/tools/OmeroPy/test/integration/test_delete.py
```diff
@@ -231,6 +231,55 @@ def testOddMessage(self):
...
+
+        cid = comment.id.val
+        assert self.query.find("CommentAnnotation", cid)
+
+        # Remove comment from first image
+        linkIds = []
+        for l in images[0].copyAnnotationLinks():
+            linkIds.append(l.id.val)
+        command = Delete2(targetObjects={"ImageAnnotationLink": linkIds})
+        handle = self.client.sf.submit(command)
```
joshmoore: This handle is not closed.

will-moore:
@joshmoore If I do this, do I have to worry about closing the handle, or is that done for me?
```
command = Delete2(targetObjects={"CommentAnnotation": [cid]}, dryRun=True)
cb = self.conn.c.submit()
rsp = cb.getResponse()
```

will-moore:
@joshmoore ...if so, then those commits should address your points.

joshmoore:
> @joshmoore If I do this, do I have to worry about closing the handle, or is that done for me?

Unfortunately, not at the moment. But from your, @ximenesuk's, and @mtbc's recent work, I wonder if it's worth the semi-breaking change of going ahead and closing the handle:
```
        try:
            callback.loop(loops, ms)  # Throw LockTimeout
        except LockTimeout:
            if failontimeout:
                callback.close(closehandle)
                raise
            else:
                return callback

        rsp = callback.getResponse()
        if isinstance(rsp, omero.cmd.ERR):
            if failonerror:
                callback.close(closehandle)
                raise omero.CmdError(rsp)

        # ==> Should a finally here close the handle if failontimeout is False?

        return callback
```

will-moore:
@joshmoore OK, so I still need to close that handle in container.py

will-moore:
@joshmoore Added .close() after dryRun delete.

joshmoore:
All looks good, @will-moore. Thanks for tackling those. Just waiting on green.

(To the above discussion https://github.com/openmicroscopy/openmicroscopy/pull/3791#issuecomment-100892988 it was decided to not implement anything for 5.1.2)

mtbc:
For what it's worth, it would be quite possible to have the server handle this for you: if the web user deletes a comment or rating or whatever, web could just delete the link and the server could then delete that annotation only if orphaned.

PR #3784, opened 2015-05-01
Delete Plate Run

will-moore:
See http://trac.openmicroscopy.org.uk/ome/ticket/12860

This is a small fix to the delete dialog message, to use 'Plate Run' instead of 'Acquisition' and remove '-locked' from the message (if you were not the owner). Also fixes refresh of Plate during Run deletion:

To test:
 - Login as Admin and try to delete data belonging to another user.
 - Project etc should simply be "Project" in the delete confirm dialog (not 'Project-locked').
 - Acquisition should be "Plate Run" (see first screenshot).
 - On delete, centre panel should be cleared and nothing should be selected in tree (fixes bug shown in second screenshot).

screen shot 2015-05-01 at 14 25 07

Before this PR, centre panel reloaded Plate while Run was deleting.
screen shot 2015-05-01 at 14 25 28

NB: Empty plate is not well handled in Web:
screen shot 2015-05-01 at 14 25 59

And worse in Insight:
screen shot 2015-05-01 at 14 03 03

PR #3792, opened 2015-05-05
Delete refresh centre

will-moore:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12866

To test:
 - Delete an image in a Dataset
 - Dataset should become selected but centre panel should NOT reload (thumbnail should simply disappear).

NB: this bug was a result of recent PR https://github.com/openmicroscopy/openmicroscopy/pull/3784

Also added Robot tests which includes changes to the setup script to import a number of images for deletion.


pwalczysko:
Works as expected.

PR #3794, opened 2015-05-06
fix #12868: prevent error when deleting container with undeletable content

mtbc:
Fixes error in workflow from http://trac.openmicroscopy.org/ome/ticket/12868. Now a group owner Alice may delete Bob's dataset that includes images that Bob tagged.


pwalczysko:
On today's trout merge:
  - the case mentioned in the http://trac.openmicroscopy.org/ome/timeline?from=2015-05-06T15%3A13%3A14%2B01%3A00&precision=second is passing now both in r-a as well as r-w groups (all yours, just the annotations not yours)
  - Also the ``Bob and Alice`` case from the header of the PR works fine, r-a and r-w groups, also with other annotations (like file anttachments, comments, ROIs) present

joshmoore:
LGTY?

pwalczysko:
All the other cases mentioned in the testing sheet which failed on delete are passing now.
This seems all okay here.

pwalczysko:
All looks good.

PR #3795, opened 2015-05-07
Csv ij

jburel:
Allow user to save the ROI measurement as a cvs file attached to the image.
cc @pwalczysko @bramalingam 

Initial review.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/MainIJPlugin.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/MainIJPlugin.java
@@ -237,7 +237,7 @@ implements PlugIn
         String home = "";
         String configFile = null;
         int index = LookupNames.IMAGE_J;
-        boolean save = false;
+        int save = -1;
         if (args != null) {
             String[] values = args.split(" ");
             List<String> l = new ArrayList<String>();
@@ -248,8 +248,10 @@ implements PlugIn
                     if (k.length == 2) {
                         if (k[1].equals("import")) {
                             index = LookupNames.IMAGE_J_IMPORT;
-                        } else if (k[1].equals("save")) {
-                            save = true;
+                        } else if (k[1].equals("saveRoi")) {
+                            save = SaveEvent.ROIS;
+                        } else if (k[1].equals("saveResult")) {
+                            save = SaveEvent.RESULTS;
                         }
                     }
                 } else l.add(v);
@@ -268,9 +270,9 @@ implements PlugIn
         }
         try {
             container = Container.startupInPluginMode(home, configFile, index);
-            if (save) {
+            if (save >=0) {
                 container.getRegistry().getEventBus().post(
-                        new SaveEvent(LookupNames.IMAGE_J));
+                        new SaveEvent(LookupNames.IMAGE_J, save));
             }
             attachListeners();
         } catch (StartupException e) {
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/MeasurementsSaver.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/MeasurementsSaver.java
@@ -1,0 +1,107 @@
+/*
+ *------------------------------------------------------------------------------
+ *  Copyright (C) 2015 University of Dundee. All rights reserved.
+ *
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ *------------------------------------------------------------------------------
+ */
+package org.openmicroscopy.shoola.agents.fsimporter;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.openmicroscopy.shoola.agents.fsimporter.view.Importer;
+import org.openmicroscopy.shoola.env.data.util.SecurityContext;
+import org.openmicroscopy.shoola.env.data.views.CallHandle;
+
+import pojos.AnnotationData;
+import pojos.DataObject;
+import pojos.FileAnnotationData;
+import pojos.ImageData;
+
+
+/**
+ * Save the measurements.
+ *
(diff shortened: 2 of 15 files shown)


joshmoore:
What's the basis for the CSV format? Is there any chance that this could subsequently be piped to `populate_roi.py`?

jburel:
no particular reason, mainly what ImageJ users are expecting from discussion with graeme and others.
Ideally we could move to tables.

joshmoore:
Makes sense for it to meet IJ-users expectations; mostly just wondering if there is anything we should do to make the two interoperate.

jburel:
agree, will have to investigate that i.e. adjustment to populate or review formatting of the csv file.

bramalingam:
Works as expected. Please merge.

RFE: As discussed with @jburel : Can we update the menu item option in imagej to have "Save to OMERO" instead of "Save ROI's to OMERO"
Is it worth having an option : "Update ROI/Results in Omero", just in case if the user wants to update the existing ROI's in OMERO instead of adding to the existing list.!

jburel:
@bramalingam: we won't be able to "update" the results, a new attachment will be created.

bramalingam:
Sorry not the results. I was thinking about results in py tables, when I was typing that.

jburel:
@gusferguson: menu entry modified this will require some screenshots adjustments when the PR is merged

gusferguson:
@jburel - Thanks - have put it on list.

pwalczysko:
Bug: The ``from OMERO -> to OMERO`` workflow does not work.

   - exchange the bioformats_package.jar for today's BIOFORMATS-5.1-merge-build #845 one
   - exchange the insight-ij plugin using the one from today's OMERO-5.1-merge-build
   - open an image from trout merge, user-4 ID 9425 from OMERO using the plugin in ImageJ
   - delete the two regions which are already present on the image (in ROI manager of IJ)
   - draw two new regions (elipse), each time adding to ROI Manager
   - click "Measure" in the ROI Manager
   - observe the Results window poping up with 2 lines in it
   - go Plugins > Save to OMERO
   - selecting the default options, save the results to omero
   - refresh the insight-ij UI to see the new regions and attachments -> nothing new there, neithre regions nor attachments 


Note that the above workflow works fine when new image is opened from my Desktop.

pwalczysko:
I am not sure about the menu now - this is again in the situation where Graeme was protesting:
Yes, the user will get the info after he clicked on ``Save to OMERO`` about the regions and/or results, but before he clicks there, he will be confused about the ``Save to OMERO`` and ``Upload and Save to OMERO`` difference (no way to tell anything about these menu items without trying) - could we not give more hints somehow ?

jburel:
Will a "Save Results to OMERO" be enough?
cc @graemeball @gusferguson

gusferguson:
Would "Save Changes to OMERO" not be better.

jburel:
We save  ROI and measurements so I am not sure the term "Changes" will be understood

pwalczysko:
My vote would be for "Save Results to OMERO" - changes might mean "all you did to the image in ImageJ" which we do not handle in this command (you have to reimport the image).

pwalczysko:
All works fine here, the bug is sorted out.

Do we want to do the strict reliance on ROI Manager for Results though ? I mean you can pop up the Results table also via Analyze > Measure command in ImageJ. In such case, the ``Save to OMERO`` will bring just an empty csv sheet as an attachment (because nothing was added to ROI Manager).
Although we might argue that Results make no sense without ROIs, I am not sure about this and also if you have the ROIs in Overlay, you still do not get the Results, just because you did not use ROI Manager.

jburel:
I think we should add the other possibilities.
@bramalingam could you list them so we are sure to cover them all.

jburel:
Review saving options as suggested by @graemeball.
@bramalingam is looking at the imageJ to identify where to read the table from

graemeball:
I think the results table actually returns tsv (tab-separated, rather than csv) -- but it should be trivial to replace s/\t/,/. Here is a Gist for a little IJ plugin that gets the active Results window and creates a string from the contents:
    https://gist.github.com/graemeball/e11cd66523e7d08bf807
Presumably you would want to write a temp file containing text like this and upload to OMERO / attach as a FileAnnotation.

jburel:
@graemeball thanks for the gist
I will have a look
we already write the "table" into a tmp file and save it as FileAnnotation linked to the image.
So it will be a matter of adjusting the source of the table.

jburel:
Modified following feedback and discussions

bramalingam:
Works as expected. Please merge.

Just noticed that the server seems to be slower, with regards to creating ROI's from the ome-tiff. The performance seems to have dropped, might need a look in.! In my test case, I was push 1450 ROI's with the image, and I had to wait for a minute or so to get all the ROI's.

jburel:
This is probably due to the fact that you will have one roi-one shape.
Not related to this PR.

jburel:
The only thing we could improve in another PR is the feedback given when saving rois/results post import

jburel:
Actually I am going to adjust the notification when rois are saved post import in this PR

jburel:
Few more commits related to saving rois/results post import
 * Display message indicating rois saving 
 * Save results only if results table not empty

To test 1:
 * Open an image in IJ, 
 * draw a roi and add it to ROI manager. 
 * Import.
 * Check that the message is displayed after thumbnail creation. ROI Saving, ROI saved
 * Check that the roi is created

To test 2:
 * Open an image in IJ, 
 * draw a roi and add it to ROI manager. 
 * Click Analyze>Measure
 * Import.
 * Check that the message is displayed after thumbnail creation. ROI Saving, ROI saved
 * Check that the roi is created
 * Check that a csv file is attached to the image.

pwalczysko:
Bug: When the Importer is already open, and another image in ImageJ is opened and ROI drawn, then the ``Save ROIs to OMERO`` does open the Import Location Window again as expected. But, after Location is selected, and the Add to the Queue clicked, nothing happens. Strangely, only after closing the ``Import Location Window`` manually, the import proceeds as expected.

pwalczysko:
test1 passed.

pwalczysko:
test2 passed. 

One small RFE there though: Although I have successfully uploaded the csv as Attachment when importing the image, I have no message about it (no information whatsoever). Maybe it does not belong to the importer progress bar (the ROI saving is reported as expected there). But I have no idea that the results were saved like this, so maybe add another message telling me that ?

pwalczysko:
I had just 2 problems, and noticing that the RFE I reported will be dealt with another PR now, it all boils down to https://github.com/openmicroscopy/openmicroscopy/pull/3795#issuecomment-105250946 - but as this is a bit of an edge case, I would think we shoiuld merge it and deal with it post-rc1.

jburel:
@pwalczysko 
I re-use the same message label and start the roi saving and results saving at the same time
So one could overwrite the other. I opted to display message for ROI saving only since the operation could take a long time.

jburel:
Regarding the bug. The Import location dialog should not pop up

jburel:
@pwalczysko 
Looking closely at the code. 
* The dialog pops up
*  you select the destination
* click add to queue.
* The import starts

Did you see something different?

pwalczysko:
Sorry, cannot repeat the Bug now. Please merge.

jburel:
Ok Thanks @pwalczysko

PR #3796, opened 2015-05-07
Viewer interpolate for big images

will-moore:
This extends the Interpolate / pixelate effect to the big image viewer which was missed in #3772.

To test, zoom in to max (400%) in the big image viewer, then toggle the ```Interpolate``` checkbox and see if the image becomes pixelated when Interpolate is off.
This is kinda subtle since the max zoom is only 400%.


mtbc:
:+1: Works just fine, **good to merge**.

Noticed separately: for big images the "Zoom (%)" on the left of the image appears to be an impotent lie. Outwith this PR though.

joshmoore:
Leaving @will-moore to file an RFE if needed.

PR #3805, opened 2015-05-12
Group summery isLeader

will-moore:
Fixes http://trac.openmicroscopy.org.uk/ome/ticket/12879

To test:
 - A user needs to be owner of their Default group and a regular member of another group.
 - Check that they are listed correctly under the Group / User menu as Owner and Member of the correct groups (not owner of both).


mtbc:
Works perfectly, good to merge.

PR #3806, opened 2015-05-12
Use chmod2 for group updatePermissions

will-moore:
@mtbc said this was a good idea?

To test:
 - Edit the permissions of a group as Admin.
 - And as a group owner.

 Check that groups are updated as expected.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -39,7 +39,7 @@ import omero.scripts
 
 from omero.rtypes import rbool, rint, rstring, rlong, rlist, rtime, unwrap
 from omero.model import ExperimenterI, ExperimenterGroupI
-from omero.cmd import Chmod2
+from omero.cmd import Chmod
 
 from omero.gateway import TagAnnotationWrapper, AnnotationWrapper
 from omero.gateway import OmeroGatewaySafeCallWrapper
@@ -1452,9 +1452,10 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
         """
 
         perms = str(permissions)
-        logger.debug("Chmod2 of group ID: %s to %s" % (group.id, perms))
-        command = Chmod2(targetObjects={'ExperimenterGroup': [group.id]},
-                         permissions=perms)
+        logger.debug("Chmod of group ID: %s to %s" % (group.id, perms))
+        command = Chmod(type="/ExperimenterGroup",
+                        id=group.id,
+                        permissions=perms)
         cb = self.c.submit(command)
         cb.close(True)
 
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -39,6 +39,7 @@ import omero.scripts
 
 from omero.rtypes import rbool, rint, rstring, rlong, rlist, rtime, unwrap
 from omero.model import ExperimenterI, ExperimenterGroupI
+from omero.cmd import Chmod2
 
 from omero.gateway import TagAnnotationWrapper, AnnotationWrapper
 from omero.gateway import OmeroGatewaySafeCallWrapper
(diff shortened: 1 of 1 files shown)


mtbc:
One can't use `Chmod2` in 5.1 clients without fallback as the 5.1.1 server doesn't have it. `Chmod` is safe and amounts to the same thing though (`ChmodFacadeI` translates it to `Chmod2` on servers that have it).

gusferguson:
@will-moore 

Tested with trout merge user-6 user-5 user-7 user-3 and user-4 on specifically created groups

Changing group permissions behaved as expected and permissions all appeared correct with superficial testing.
As long as Mark's comment above does not preclude it - Good to merge.

joshmoore:
@mtbc comment should definitely be addresses in 5.1.2 if this goes in. @will-moore : in this PR?

will-moore:
OK, will use ```Chmod2``` in this PR...

will-moore:
Changing ```Chmod2``` to ```Chmod``` in the above code gave me an error:
```
__init__() got an unexpected keyword argument 'targetObjects'
```
at 
```
        command = Chmod(targetObjects={'ExperimenterGroup': [group.id]},
                        permissions=perms)
```

mtbc:
`Chmod` works like `Chgrp`; see https://github.com/mtbc/openmicroscopy/commit/ba97aa6dc59ced610f0f6f9d41c9cc2d31673a66 for an example of how `Chmod` differs from `Chmod2`.

gusferguson:
@will-moore 

Tested using trout merge user-6 user-3 and user-4 all permission levels checked and all behave as expected.
Good to merge from this perspective.

Probably unrelated issue:
Although probably unrelated to this PR - this is the first time I have observed that the "Warning: some objects you selected are owned by other users" text is coming up when just deleting a user's own data - see screenshot. 

webclient

will-moore:
Thanks Gus, I'll fix that in #3819.

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -1382,8 +1383,7 @@ def updateGroup(self, group, name, permissions, owners=list(),
         # Should we update updateGroup so this would be atomic?
         admin_serv.updateGroup(up_gr)
         if permissions is not None:
-            logger.warning("WARNING: changePermissions was called!!!")
```
joshmoore: Just as a reminder: this WARNING was here because the form was unintentionally calling `updatePermissions` on every save.

will-moore: This is still happening and needs fixing, right?

joshmoore: I'd guess that it's *not* happening, but the warning was essentially documentation about what can go wrong. Not sure if we should just remove this branch entirely. @aleksandra-tarkowska may know more.

mtbc:
Reversion now pushed to my branch https://github.com/mtbc/openmicroscopy/commit/db7406c5; I will open a PR once we branch OMERO 5.2.

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -1440,21 +1440,24 @@ def setDefaultGroup(self, group_id, exp_id=None):
...
-        admin_serv = self.getAdminService()
-        if permissions is not None:
-            logger.warning("WARNING: changePermissions was called!!!")
-            admin_serv.changePermissions(obj._obj, permissions)
+        perms = str(permissions)
+        logger.debug("Chmod of group ID: %s to %s" % (group.id, perms))
+        command = Chmod(type="/ExperimenterGroup",
+                        id=group.id,
+                        permissions=perms)
+        cb = self.c.submit(command)
```
aleksandra-tarkowska: That is not sufficient. you should always use self._waitOnCmd(handle)

joshmoore: No. This is precisely what `c.submit` does. It's very different from `c.sf.submit`. (Sorry for the poor naming choice. My fault!)

PR #3807, opened 2015-05-12
Imagej import

jburel:
Fix the issue when a new image was created in imagej from for examples the combination of 2 images
To test this PR:
 * Open 2 images
 * Combine the 2 images
 * Import the result to OMERO

Also fix the "no datase/no project" option in the location dialog
If select, make sure that "--" is displayed in the column

Try various algorithms
cc @bramalingam


pwalczysko:
I tried to open and concatenate 2 dvs. It worked in general, although the concatenated image gets different color of channels afer the import that the original images, but I think this is because I have to be more precise when working with channels in ImageJ.
Will leaave @bramalingam to do his workflow.

bramalingam:
Tried concatenation and orthogonal views. Both the imports work as expected. Good to merge.

With the orthogonal views though, we might have to fix a bug in bioformats_exporter, on its handling of Channels, while exporting a processed image from ImageJ.

PR #3808, opened 2015-05-12
Min/Max roi

jburel:
To test 
 * Use one of the image listed in https://trac.openmicroscopy.org/ome/ticket/12877
You should NOT need to have to regenerate roi in imageJ

cc @pwalczysko

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/GraphPane.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/GraphPane.java
@@ -190,7 +190,7 @@ public class GraphPane
 		Map channels = model.getActiveChannels();
 		Entry entry;
 		Iterator i = channels.entrySet().iterator();
-		double value = 0;
+		double value = Double.MAX_VALUE;
 		int channel;
 		while (i.hasNext())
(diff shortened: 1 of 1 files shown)


pwalczysko:
For images, see team/pwalczysko/QA 11065/*.* or imported to trout merge, user-4, read-only-1, Project QA 11065

jburel:
I have created rois on the few images w/o noticing any issues
The problem is not coming from rois generated in imageJ but to the min/max values of a given channel
I have not been able to reproduce the issue with the images imported by @pwalczysko.

pwalczysko:
I could not reproduce the problem either. Even with exactly the same setup as the QA user was using, Win7, 5.1.1 clients and ij-plugin, the same image which I reimported, drawing polygon regions in IJ and saving them back. All works as expected.

pwalczysko:
For the polygon drawing problem I have created ticket https://trac.openmicroscopy.org/ome/ticket/12883#ticket. (not related to this PR, just reported by the same user at the same time.)

We could not see any problems even reopening the original nightshade image which caused the crashes first using the merge insight. 

Ready to merge.

pwalczysko:
cc @kennethgillen

PR #3783, opened 2015-05-01
Webadmin 'Read-Write' perms

will-moore:
This allows Admins to create 'Read-Write' groups in webadmin. See https://trac.openmicroscopy.org/ome/ticket/10751

To test:
 - Log in as Admin, go to webadmin, create a Group...
 - See that there is now a 4th "Read-Write" option for group perms.
 - Create a group with these perms.
 - Test new group behaves / appears as expected in web (or insight)
 - Edit the new group (check that perms are "Read-Write" before editing)
 - Edit permissions and check updated OK.

PR #3811, opened 2015-05-13
Read write

dominikl:
Enables ReadWrite group permission in Insight, [Trello](https://trello.com/c/RpdAOmKp/50-permissions-system-problems-chown)
Test:
- Create a new group with ReadWrite permission and add two users to it, make one owner of the group.
- Log in as one of the users and import some images.
- Switch to the other user and delete an image of the previous user. Make sure the confirmation dialog states that you're going to delete another user's data.
- Log in as the group owner. Change permission level of the group. Make sure you cannot change the permission back to ReadWrite.
- Log in as admin. Change permission level back to ReadWrite

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/GroupProfile.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/GroupProfile.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.metadata.editor.GroupProfile 
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
@@ -114,7 +114,7 @@ class GroupProfile
     	//permission level
 
     	permissionsPane = new PermissionsPane(ref.getPermissions(),
-    			UIUtilities.BACKGROUND_COLOR);
+    			UIUtilities.BACKGROUND_COLOR, model.isAdministrator());
     	level = permissionsPane.getPermissions();
     	permissionsPane.allowDowngrade(!mayChangePermissions);
     	permissionsPane.setBorder(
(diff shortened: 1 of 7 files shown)


mtbc:
What happens if one deletes one's own dataset containing another user's data? I'm just thinking that, if there is sometimes a warning dialog, it would be good if users didn't take undue heart from its absence.

dominikl:
You mean always show a warning like "you may delete other users' data"? What's Web going to display? That should be the same in Insight anyway, current text is just more like a placeholder.

dominikl:
/cc @will-moore 
corresponding Web PR: #3783

mtbc:
I don't know what should be done about it, I'm just raising it as a UX issue. :smiley:

gusferguson:
@dominikl 

Tested with OMERO.insight-5.1.1-315-5775dd7-ice35-b501-mac_Java7+ and OMERO.insight-5.1.1-315-5775dd7-ice35-b501-win Win 8 user-3, user-4 and user-6.

Behaves as expected with above test.
Note that changes in permission to a group only show for other users when they log out and back in again, but I don't think this is an issue.
Warning dialogues all look good - screenshots 1 and 2.
Good to merge.

confirm delete

parallels desktop

dominikl:
I added a flag that the warning dialog for changing a group to RW pops up only once, thought it would be a bit annoying otherwise. Maybe we should have a quick chat about that - also about @mtbc 's comment - in the client strategy meeting later.

gusferguson:
@dominikl - OK - can talk about it at client meeting - I have also just noticed some minor layout inconsistencies that it may be an idea to look at at the same time - I will show you at the client meeting.

joshmoore:
> What happens if one deletes one's own dataset containing another user's data? I'm just thinking that, if there is sometimes a warning dialog, it would be good if users didn't take undue heart from its absence.

@mtbc, I think my point in bringing this up, was *if* we thought it sufficiently computable what should be consider a warning. i.e. if with `dryRun=True` we could detect all such cases server side. What we *do* with that detected fact, is a larger discussion.

mtbc:
Probably fairly computable. Have to think about the best approach there, though. Asking `DiskUsage` may include, say, others' attached files that probably wouldn't be deleted. Would probably prefer clients not to have to take each class from the `dryRun` response map and do `SELECT DISTINCT details.group.id FROM :class WHERE id IN (:ids)`, though that might be the best workaround within 5.1.

jburel:
@gusferguson: we ran out of time last week. Do you want to discuss it? @dominikl is now away.
so I will have to finish this PR.

gusferguson:
@jburel - okay - but they are minor, could easily be done as RFEs for addressing later.

jburel:
@gusferguson: in that case, will merge this one with the corresponding web one.

joshmoore:
Merging for 5.1.2.

PR #3812, opened 2015-05-13
check for Unicode units on startup

mtbc:
See https://trello.com/c/CVUflk8g/490-encoding-for-db-init-upgrade-script. Adds a startup check to the server to prevent the server being used if the wrong character encoding was used to set up its database. Also tweaks SQL scripts to ensure that those with Unicode symbol literals are read with the correct encoding.

--no-rebase

***

To test, start out with a local OMERO.server and SQL scripts that do not have this PR merged. Partway through testing you will switch to a local OMERO.server and SQL scripts that do have this PR merged.

1. Start out without this PR.
1. Use `export PGCLIENTENCODING=WIN1252` in your shell to mess up the reading of Unicode SQL scripts.
1. Empty your binary repository and use `bin/omero db script` and `psql` to create a 5.0 or 5.1 database. If you used a 5.0 database, use `psql` to upgrade it to 5.1.
1. Start the server and try to import some metadata-rich images. You will quickly run into problems.
1. Stop the server, include this PR, then start it again.
1. Observe that the server doesn't quite manage to start up usably. Stop the server and observe the helpful errors in `Blitz-0.log`.
1. Empty your binary repository and use `bin/omero db script` and `psql` to create a 5.0 or 5.1 database. If you used a 5.0 database, use `psql` to upgrade it to 5.1.
1. Observe that the server starts up just fine and all manner of metadata-rich images may be imported with units preserved.

### DIFF
--- a/components/dsl/resources/ome/dsl/psql-header.vm
+++ b/components/dsl/resources/ome/dsl/psql-header.vm
@@ -35,7 +35,9 @@ BEGIN
     END IF;
 
     IF char_encoding != 'UTF8' THEN
-       RAISE EXCEPTION 'OMERO database character encoding must be UTF8, not %%', char_encoding;
+        RAISE EXCEPTION 'OMERO database character encoding must be UTF8, not %%', char_encoding;
+    ELSE
+        SET client_encoding = 'UTF8';
     END IF;
 
 END;$$ LANGUAGE plpgsql;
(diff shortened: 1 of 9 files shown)


joshmoore:
Some of these changes will suffer from the same issue as https://github.com/openmicroscopy/openmicroscopy/pull/3697

mtbc:
True, it doesn't fix the DB for people who already upgraded, in that case it just detects the problem.

dominikl:
Works like expected. A server with faulty unit tables won't start and writes a reasonable error message in Blitz.log. New installation and upgrade sql scripts (tested upgrade from 5.0.8) now also work with non-UTF8 encoding set as default. Good to merge.

PR #3810, opened 2015-05-13
adding login params for autologin

aleksandra-tarkowska:
This PR should allow passing login parameters to auto login into Insight.

JNLP file now gets host port and session ID that is minimum to log in

```
<property name="jnlp.omero.host" value="trout...org"/>
<property name="jnlp.omero.port" value="4064"/>
<property name="jnlp.omero.sessionid" value="79e4596a-bb8c-470f-8e6a-47b91ed0c5ab"/>
<property name="jnlp.omero.web.host" value="http://omero.testdomain.com:4080/"/>
```

Note: Webstat page is also available when user is **not logged in**. In this case there will be default server host and port set and no sessionid provided. These parameters should be used as a helper to choose the right server from the list.

cc: @jburel @dominikl

PR #3814, opened 2015-05-13
SessionID from webstart

dominikl:
Pick up the session id to auto login into insight from webstart, see #3810


jburel:
@dominikl: 
i have a branch with some more complex code since we do not want to pop the splashscreen
and we also need to handle the case w/o session id.

jburel:
see https://github.com/jburel/openmicroscopy/tree/jnlp-parameters

dominikl:
@jburel  Alright, didn't see you already had a branch for this. Closing this branch then. Was just curious if it's possible to get the jnlp properties via `System.getProperty()`.

PR #3815, opened 2015-05-14
Revive foreground_poll logic for callbacks.py

joshmoore:
In 5.1.0 and before, the CmdCallbackI constructor
in Python (unlike Java and C++) was performing its
initial polling invocation in the calling thread.
Due to its statelessness, OMERO.web was making use
of this fact. Now by passing foreground_poll=True,
the previous, deprecated behavior can be revived.

More complicated behavior should be achieved by
subclassing CmdCallbackI and overriding initialPoll.


will-moore:
Delete callback seems to be fixed on trout. Good to merge from me.

PR #3816, opened 2015-05-14
Add a check for .lock files on startup

joshmoore:
Though not an entire `fs fsck` solution this at least
alerts users to the fact that a .lock file is present.
The result of these files *could* be that imports are
not possible.

See: https://trello.com/c/WiG4Kd78/510-fs-fsck

#### Testing ####

 * Start your server and find the lock file, e.g. /OMERO/.omero/repository/a9aadade-139e-426c-9ff8-788b0ca90c23/.lock
 * Stop the server
 * The lock file should be gone. `touch` the lock file location.
 * Run `bin/omero admin start` and check for the following warning:

```
$ dist/bin/omero admin start
WARNING: lock files in /OMERO
----------------------------------------
/OMERO/.omero/repository/a9aadade-139e-426c-9ff8-788b0ca90c23/.lock
----------------------------------------

You may want to stop all server processes and remove
these files manually. Lock files can remain after an
abrupt server outage and are especially frequent on
remotely mounted filesystems like NFS.

No descriptor given. Using etc/grid/default.xml
```

More advanced testing would including trying to reproduce the issue described in https://trello.com/c/WiG4Kd78/510-fs-fsck on an omero.data.dir located on a NFS mount.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -701,7 +701,7 @@ present, the user will enter a console""")
         if 0 == self.status(args, node_only=True):
             self.ctx.die(876, "Server already running")
 
-        self.check_lock(config=config)
+        self.check_lock(config)
 
         self._initDir()
         # Do a check to see if we've started before.
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -20,6 +20,7 @@ import stat
 import platform
 import datetime
 
+from glob import glob
 from path import path
 
 import omero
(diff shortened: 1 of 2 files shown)


mtbc:
The lockfile in `/OMERO/.omero/repository/` seems to be the important one, it ignores the one in `/OMERO/ManagedRepository/.omero/repository/`.

mtbc:
Certainly the code looks plausible and the testing goes as described. Seems at least a general improvement so **good to merge**.

joshmoore:
Considering `ManagedRepository`could be on another server, I punted on detecting its lock file. However, the warning message points the sysadmin to the same situation, so minimally it will be a reminder for me when this shows up on the forums, as opposed to having to remote debug with Chris! 

However, I did realize that the following is likely confusing: `bin/omero admin restart && bin/omero admin start` will print the warning, too.

components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -700,6 +701,8 @@ def startasync(self, args, config):
         if 0 == self.status(args, node_only=True):
             self.ctx.die(876, "Server already running")
 
+        self.check_lock(config=config)
```
sbesson: Given the signature below, should we simply have `self.check_lock(config)`?

joshmoore:
Pushed.

sbesson:
From a merge server went through several cycles of `start/stop | restart` with lock addition with default/non-default `omero.data.dir`. Code is sensible and ready to merge. 

Additionally, tried to reproduce the issue by working from a binary repository on a smb mount. Interestingly, this caused a failure at import


````
2015-05-18 15:43:41,061 32014      [      main] ERROR        ome.formats.importer.ImportLibrary - Error on import
java.lang.RuntimeException: Cannot exclusively use the managed repository.

Likely no ManagedRepositoryPrx is being returned from the server.
This could point to a recent server crash. Ask your server administrator
to check for stale .lock files under the OMERO data directory. This
is particularly likely on a server using NFS.

   at ome.formats.importer.ImportLibrary.checkManagedRepo(ImportLibrary.java:735) ~[blitz.jar:na]
   at ome.formats.importer.ImportLibrary.createImport(ImportLibrary.java:332) ~[blitz.jar:na]
   at ome.formats.importer.ImportLibrary.importImage(ImportLibrary.java:491) ~[blitz.jar:na]
   at ome.formats.importer.ImportLibrary.importCandidates(ImportLibrary.java:287) ~[blitz.jar:na]
   at ome.formats.importer.cli.CommandLineImporter.start(CommandLineImporter.java:245) [blitz.jar:na]
   at ome.formats.importer.cli.CommandLineImporter.main(CommandLineImporter.java:858) [blitz.jar:na]
2015-05-18 15:43:41,061 32014      [      main] INFO         ome.formats.importer.ImportLibrary - Exiting on error
```

As discussed with @joshmoore, a more stringent test would be to wait until a `ManagedRepositoryPrx` is returned in `omero admin start`. While this may slightly increase the start time, this would ensure both the DB and the binary repo are accessible at startup which is exactly what sysadmins are expecting /cc @kennethgillen.

PR #3821, opened 2015-05-15
Tweak the CLI delete report

ximenesuk:
This PR alters the output of the `--report` option of the `omero delete` command in line with https://github.com/openmicroscopy/openmicroscopy/pull/3725#issuecomment-96042925

The deleted objects list should use comma-separated ID lists rather than one object per line as before. Deleted `ome.model.display.*` objects should be excluded from the output:

```
$ omero delete Image:101 --dry-run --report
omero.cmd.Delete2 Image 101... ok
Steps: 1
Elapsed time: 0.451 secs.
Flags: []
Deleted objects
ome.model.fs.FilesetEntry:101
ome.model.core.Image:101
ome.model.core.OriginalFile:156,155
ome.model.core.LogicalChannel:51
ome.model.core.Channel:51
ome.model.jobs.PixelDataJob:53
ome.model.jobs.IndexingJob:55
ome.model.fs.Fileset:51
ome.model.jobs.ThumbnailGenerationJob:54
ome.model.jobs.MetadataImportJob:52
ome.model.jobs.UploadJob:51
ome.model.core.Pixels:101
ome.model.fs.FilesetJobLink:51,55,54,53,52
ome.model.stats.StatsInfo:65
ome.model.jobs.JobOriginalFileLink:51
```

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/delete.py
+++ b/components/tools/OmeroPy/src/omero/plugins/delete.py
@@ -49,18 +49,23 @@ class DeleteControl(GraphControl):
     def print_delete_response(self, rsp):
         self.ctx.out("Deleted objects")
         objIds = self._get_object_ids(rsp)
-        for k in sorted(objIds):
+        for k in objIds:
             self.ctx.out("%s:%s" % (k, objIds[k]))
 
     def _get_object_ids(self, rsp):
+        import collections
         objIds = {}
         for k in rsp.deletedObjects.keys():
             if rsp.deletedObjects[k]:
                 for excl in EXCLUDED_PACKAGES:
                     if not k.startswith(excl):
                         ids = ','.join(map(str, rsp.deletedObjects[k]))
-                        k = k[k.rfind('.')+1:]
                         objIds[k] = ids
+        newIds = collections.OrderedDict(sorted(objIds.items()))
+        objIds = collections.OrderedDict()
+        for k in newIds:
+            key = k[k.rfind('.')+1:]
+            objIds[key] = newIds[k]
         return objIds
 
 try:
--- a/components/tools/OmeroPy/src/omero/plugins/delete.py
+++ b/components/tools/OmeroPy/src/omero/plugins/delete.py
@@ -48,12 +48,20 @@ class DeleteControl(GraphControl):
 
     def print_delete_response(self, rsp):
         self.ctx.out("Deleted objects")
+        objIds = self._get_object_ids(rsp)
+        for k in sorted(objIds):
+            self.ctx.out("%s:%s" % (k, objIds[k]))
+
+    def _get_object_ids(self, rsp):
+        objIds = {}
         for k in rsp.deletedObjects.keys():
             if rsp.deletedObjects[k]:
                 for excl in EXCLUDED_PACKAGES:
                     if not k.startswith(excl):
                         ids = ','.join(map(str, rsp.deletedObjects[k]))
-                        self.ctx.out("%s:%s" % (k, ids))
+                        k = k[k.rfind('.')+1:]
+                        objIds[k] = ids
+        return objIds
 
 try:
     register("delete", DeleteControl, HELP)
--- a/components/tools/OmeroPy/src/omero/plugins/delete.py
+++ b/components/tools/OmeroPy/src/omero/plugins/delete.py
@@ -27,6 +27,8 @@ Examples:
 
 """
 
+EXCLUDED_PACKAGES = ["ome.model.display"]
+
 
 class DeleteControl(GraphControl):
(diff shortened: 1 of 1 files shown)


joshmoore:
`EXCLUDED_PACKAGES` is an interesting idea. I wonder if this should 1) be configurable and/or 2) be kept in sync with what's output by import in addition to the Pixels IDs.

ximenesuk:
@joshmoore that was one of my questions on this PR: putting this somewhere configurable/available to other functions.

joshmoore:
Output from testing gh-3824:
```
[jamoore@ome-c6100-3 pr3824]$ omero -q delete --report --dry-run Image:23409 --include=Annotation
omero.cmd.Delete2 Image 23409... ok
Steps: 1
Elapsed time: 0.332 secs.
Flags: []
ome.model.fs.Fileset:809
ome.model.annotations.FileAnnotation:12926
ome.model.core.Image:23409
ome.model.core.Pixels:21209
ome.model.stats.StatsInfo:44028
ome.model.fs.FilesetEntry:7059
ome.model.core.Channel:51062
ome.model.fs.FilesetJobLink:3141,3142,3143,3144,3145
ome.model.core.LogicalChannel:21962
ome.model.jobs.PixelDataJob:4196
ome.model.core.OriginalFile:19747,19746,19748
ome.model.jobs.ThumbnailGenerationJob:4197
ome.model.jobs.MetadataImportJob:4195
ome.model.jobs.UploadJob:4194
ome.model.jobs.JobOriginalFileLink:2215
ome.model.annotations.CommentAnnotation:12925
ome.model.jobs.IndexingJob:4198
ome.model.annotations.ImageAnnotationLink:9023,9022,9021
ome.model.annotations.TagAnnotation:12924
```

I noticed a couple of other things:
 * The package prefixes likely should be stripped for use in `omero obj`, but that means if we expand exclusion based on package, things could get tricky.
 * Annotations are not sorted together. Perhaps ok, but a bit surprising.

ximenesuk:
Thanks @joshmoore 

The package prefixes could be stripped post-exclusion as long as the server response object always contains full object names.

Sorting would be trivial, and would make sense compared to the apparent randomness coming from the server - I assum this order is based on the actual graph traversal?

mtbc:
I expect the server to keep returning full object names. It might be good to sort them before stripping them down to the class name so that related ones tend to appear together. The order coming from the server is probably determined by the order of the strings' hash codes!

ximenesuk:
This last commit removes the package prefix. It then sorts simply. That doesn't group related objects but it does make the output a little less haphazard. Some other approach could eventually be tried but hopefully this will help readability a little.

ximenesuk:
This final commit does group by container, core, fs, etc. Maybe that is more useful than simply alphabetic?

mtbc:
```
$ omero delete Thumbnail:109 --report
Using session c53c9685-e9c6-46c1-91fd-b890f5463858 (user-2@localhost:4064). Idle timeout: 10 min. Current group: private-1
omero.cmd.Delete2 Thumbnail 109... ok
Steps: 3
Elapsed time: 2.599 secs.
Flags: []
Deleted objects
$ 
```
Pretty funny. (Why do we always omit the display stuff?)

The classes are in a nice order, thank you. Still, after that effort, for things like,
```
FilesetJobLink:754,755,752,753,751
```
why not sort the IDs too?

Despite the above, this PR is already **good to merge** as it is, especially with a code freeze looming.

mtbc:
After further conversation: to keep the report manageably small, list all classes but list >2 contiguous IDs just by range endpoints?

joshmoore:
> list all classes but list >2 contiguous IDs by just range endpoints?

For delete, *grudgingly* ok. But for `chgrp` that wouldn't then be paste-able to a delete command.

ximenesuk:
It would if we taught delete how to handle ranges!

ximenesuk:
I'm happy for this to be merged now unless the id ordering is required here? I'll add other RFEs to a card if so.

joshmoore:
Ok. So reading through the RFEs are:
 * handle delete of excluded-only somehow (e.g. Thumbnail)
 * rationalize exclusions (make configurable?)
 * order IDs
 * move to ID ranges

But since this is all for the details report only, we'll assume this can survive for the moment.

mtbc:
With the last two, are the first two still needed?

joshmoore:
Well, the first should have sensible defaults at least so that `delete Thumbnail` doesn't show nothin'. (so to speak)

mtbc:
(Well, I mean, with the last two, do we need any exclusions at all?)

joshmoore:
@mtbc: quite possibly!

PR #3822, opened 2015-05-15
Save enabled on z/t change

will-moore:
This fixes http://trac.openmicroscopy.org.uk/ome/ticket/12709 in web.

To test:
 - Open Preview panel
 - Change Z or T index
 - Save button should become enabled and saving should update Thumbnail etc.
 - Test same behaviour in Image Viewer.

NB: we don't have a whole undo/redo queue for Z/T changes, so after Z/T change and "Save" is enabled, if you change rendering settings then click "Undo" the Save button will become disabled again.

### DIFF
--- a/components/tests/ui/robot_setup.sh
+++ b/components/tests/ui/robot_setup.sh
@@ -14,6 +14,8 @@ GROUP_NAME_2=${GROUP_NAME_2:-robot_group_2}
 USER_NAME=${USER_NAME:-robot_user}
 USER_PASSWORD=${USER_PASSWORD:-ome}
 CONFIG_FILENAME=${CONFIG_FILENAME:-robot_ice.config}
+IMAGE_NAME=${IMAGE_NAME:-test&sizeZ=3&sizeT=10.fake}
+TINY_IMAGE_NAME=${TINY_IMAGE_NAME:-test.fake}
 
 # Create robot user and group
 bin/omero login root@$HOSTNAME:$PORT -w $ROOT_PASSWORD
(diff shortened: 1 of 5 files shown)


dominikl:
Works as described. I just noticed a problem with the multichannel-4D-series.ome.tiff example file (see user-3/ome-tiff). Once you change and save the T plane, the save button for this image is always enabled (it gets disabled directly after being saved, but if you move to a different image and back again, it's enabled again without any changes). I'm not sure if this related to this PR.

will-moore:
Fixed a couple of things there.
Now I'm checking Z/T changes "properly" so the "Save" button should be disabled if Z and T are the same as saved values and there are no unsaved rendering settings.
E.g. change Z - "Save" is enabled, then change it back, "Save" is disabled etc.

gusferguson:
@will-moore 

Tested using trout merge user-4 read-only-1 and read-write-1.

Behaves as expected on own and other user's data in Preview and full viewer. 
Dominik's issue appears to have been resolved successfully and "proper" change is working as expected.
Good to merge.

jburel:
@will-moore: do you have some robot tests, checking if the button is active or not after a slider move e.g. z-slider?

will-moore:
@jburel I don't have any Robot tests for that yet, but could add it. I just need to figure out how to create multi-Z/T test images...

gusferguson:
@will-moore - one of Iain Porter's .dvs has multiple of both - can you use a normal image?

joshmoore:
I'd assume just a `robot&sizeZ=10&sizeT=100.fake` or similar would suffice for a robot test.

will-moore:
Robot test added to check that 'Save' becomes enabled/disabled when Z-index changes.

gusferguson:
@will-moore 

Retested with trout merge user-4 read-only-1 - behaves as expected.
Good to merge.

PR #3823, opened 2015-05-15
roi_thumb shapes with no Z/T

will-moore:
Noticed a very minor issue with roi thumbnails when shapes don't have Z or T set.

To test:
 - I don't think you need to find a shape that has Z/T unset, just check that the web image viewer ROI table displays ROI "preview" thumbs is still working OK.
 - Tested locally that this does work when Z is None.

screen shot 2015-05-15 at 16 16 43


gusferguson:
@will-moore 

Tested using trout merge user-3 read-only-1

Thumbnails displayed - see screenshot.
Good to merge.

-home-hudson-for_import_to_howe-non-archived-imoprt-case-user-3-read-only-1-no_projects-non-archivedv-cfpneat01_r3d dv log

PR #3824, opened 2015-05-15
Exclude annotations in CLI delete

ximenesuk:
This PR mirrors the default action in the other clients and does not delete linked annotations (File, Tag and Term) without confirmation.
```
omero delete Image:51
``` 
will delete the image but not its annotations. Whereas
```
omero delete Image:51 --include Annotation
``` 
will also delete the images annotations. Individual annotation types can be selected,
```
omero delete Image:51 --include FileAnnotation
``` 
will delete any linked FileAnnotation objects but not other types. Other annotations can be deleted in the same command,
```
omero delete Image:51 Annotation:101
``` 
will not delete linked annotations but it will delete the separate specified annotation. Individual linked annotations can be deleted by explicitly using their IDs,
```
omero delete Image:51 Annotation:51
```
will delete the linked annotation with the given id but exclude other linked annotations.

--no-rebase

### DIFF
--- a/components/tests/python/library/__init__.py
+++ b/components/tests/python/library/__init__.py
@@ -39,7 +39,8 @@ import omero.gateway
 from omero.cmd import DoAll, State, ERR, OK, Chmod2, Chgrp2, Delete2
 from omero.callbacks import CmdCallbackI
 from omero.model import DatasetI, DatasetImageLinkI, ImageI, ProjectI
-from omero.model import Annotation, FileAnnotationI, OriginalFileI
+from omero.model import Annotation, FileAnnotationI, TagAnnotationI
+from omero.model import OriginalFileI
 from omero.model import DimensionOrderI, PixelsI, PixelsTypeI
 from omero.model import Experimenter, ExperimenterI
 from omero.model import ExperimenterGroup, ExperimenterGroupI
@@ -798,6 +799,13 @@ class ITest(object):
         """
         return self.new_object(DatasetI, name=name, description=description)
 
+    def new_tag(self, name=None):
+        """
+        Creates a new tag object.
+        :param name: The tag name. If None, a UUID string will be used
+        """
+        return self.new_object(TagAnnotationI, name=name)
+
     def make_image(self, name=None, description=None, date=0, client=None):
         """
         Creates a new image instance and returns the persisted object.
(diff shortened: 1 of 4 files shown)


joshmoore:
Will the excluded objects be printing in a detailed report? I.e. so one could manually clean them up if so desired.

ximenesuk:
As it stands, no. The report is just that which the response provides. I'm open to suggestions here. The easiest option is to do a dry-run (without the default exclusion) and then a diff on the reports after the delete. That would add an overhead to every CLI delete command but would certainly be possible.

joshmoore:
If we do that, it would also be an option for the `rwrw--` WARNING.

joshmoore:
* default arguments seem sensible to me.
 * `--report --dry-run` shows me the difference in the 2 runs: :+1:
 * `--include=Annotation` and `--include=FileAnnotation` (for example) do different (correct) things: :thumbsup:

Barring any further changes you want to make to the report here, I'd say *good to merge*. Since there's another report PR anyway, perhaps hold off?

ximenesuk:
I'd be happy for this to be merged and have any extended report showing diffs as a later PR once gh-3821 is in as well.

joshmoore:
Agreed. Merging. I'll leave you to file a RFE along these lines:

I'd suggest printing a WARNING similar to https://github.com/openmicroscopy/openmicroscopy/pull/3817 when an object is being deleted *without* passing `--include Annotation`. This should optimally be done server-side, but since insight & web are currently doing it locally, it's likely ok to start with a client-side implementation.

cc: @mtbc @jburel @will-moore

mtbc:
Surely one doesn't usually want ``-include Annotation``? For instance, one probably wouldn't want to strip the annotations off other images that aren't being deleted. I may be coming to this late and misunderstanding.

joshmoore:
@mtbc : agreed, but the UI has the benefit of the confirmation dialog. And until we add `--yes`-style functionality to the CLI, the best we can manage is something informational with the hope that people will use `--dry-run`.

mtbc:
Oh! You mean, warn them about annotations that would be deleted *unless* they passed `--include=Annotation`? (Rather than warning them in general anytime they don't pass  `--include=Annotation`.)

joshmoore:
Yes. Just a general re-education program now that `delete Image:$X` will be doing *less* than before.

PR #3826, opened 2015-05-18
Modify CreateImage training example to create multi-channel image

sbesson:
As pointed out by @ctrueden in http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2015-May/003302.html, we do not have a clear example making use of `RawPixelsStore.setPlane()` in the case of multi-channel images.

This PR modifies the existing `CreateImage` training example to create a new image of the same dimensionality as the original one instead of a single channel image.

/cc @imunro @jburel

### DIFF
--- a/examples/Training/java/src/training/CreateImage.java
+++ b/examples/Training/java/src/training/CreateImage.java
@@ -26,6 +26,7 @@ package training;
 
 //Java imports
 import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


examples/Training/java/src/training/CreateImage.java
```diff
@@ -115,12 +116,15 @@ private ImageData loadImage(long imageID)
...
 	{
 		if (z < 0 || sizeZ <= z) 
 			throw new IllegalArgumentException(
-					"z out of range [0, "+sizeZ+"): "+z+".");
-		return Integer.valueOf(sizeZ*t + z);
+					"z out of range [0, "+sizeZ+"]: "+z+".");
+		if (c < 0 || sizeC <= c)
+			throw new IllegalArgumentException(
+					"c out of range [0, "+sizeC+"]: "+c+".");
+		return Integer.valueOf(sizeZ*sizeC*t + sizeZ*c + z);
```
mtbc: Is this `Integer.valueOf` required at all?

jburel:
Changes look fine. We can bring back the combined image examples under advance when we work on the training material over the summer.

@mtbc: ```Integer.valueOf``` is not really 'required', but in theory we should have better performance due to caching.

mtbc:
With or without `Integer.valueOf`, this compiles to the same bytecode for me with both Oracle and OpenJDK: they both use `Integer.valueOf` under the hood.

jburel:
JVMs have improved since that code was written. not a blocker

mtbc:
Agreed, not a blocker; can clean up in later review of training examples.

jburel:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/329/ is green.
Merging

PR #3827, opened 2015-05-18
Delete Tags

will-moore:
This fixes tag deletion and permissions issues with all deletes.

To test:
 - Log in to Read-Write group.
 - Show Tag tree for "All Members" - Check that root of tree shows "All Members"
 - Select a mixture of Tags that you own or don't own.
 - Delete button should be enabled - click it.
 - Check that delete message makes sense.
 - Go ahead with delete and check that everything works as expected.
 - Check that in a group with other permissions that that Delete is not enabled for other users' tags unless you are Admin or Group Owner.
 - Confirm that deletion of other objects (P/D/I/S/P) is still OK, since code has been refactored to avoid duplication.


components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -2504,6 +2504,15 @@ def __prepare__(self, **kwargs):
 omero.gateway.WellWrapper = WellWrapper
 
 
+class TagWrapper(OmeroWebObjectWrapper, omero.gateway.TagAnnotationWrapper):
+    """
+    omero_model_TannotationI class wrapper overwrite omero.gateway.ImageWrapper
```
joshmoore: Minor s/Image/Tag/ cut-n-paste issue.

gusferguson:
@will-moore 

Tested using trout merge user-3 user-4 Test 15-05-14 07:58:50

Select a mixture of Tags that you own or don't own.
Delete button should be enabled - click it.
Check that delete message makes sense.
Go ahead with delete and check that everything works as expected.
- all work as expected

Check that in a group with other permissions that that Delete is not enabled for other users' tags unless you are Admin or Group Owner.
- not working - delete on tags not owned by ordinary member (user-4) remains available after downgrading to read-annotate and read-only - screenshot 1

Confirm that deletion of other objects (P/D/I/S/P) is still OK, since code has been refactored to avoid duplication.
- works as expected.

webclient

will-moore:
@gusferguson Thanks, that should fix the enabling of Delete button according to correct permissions.

gusferguson:
@will-moore 

Using trout merge, Safari, Firefox, user-3, user-4 and user-6.
Retest of previously good actions still good.
Issue with being able to delete other user's tags when not Group Owner or admin is resolved.
Behaves as expected.

Good to merge if error below not significant to this PR.

As part of testing - using Firefox - logged in as user-6 in Web Admin - tried to change Test 15-05-14 07:58:50 group to Read-Write. On clicking Save got following error message.
Could not replicate.
Filed as ticket: https://trac.openmicroscopy.org.uk/ome/ticket/12896

joshmoore:
Merging for 5.1.2.

PR #3829, opened 2015-05-19
Rating

jburel:
Display rating value. see https://trac.openmicroscopy.org/ome/ticket/12855
To test:
 * Import an image and rate it. Make sure the same value is displayed in insight and web


mtbc:
Much better, thank you. Already **good to merge**.

One odd thing: After the ticket's workflow, if I then as `user-5` also rate the image, and select above the rating "Show added by others", it makes no difference to the rating or average at all, even showing my own rating so I might think it's a rating added by others. Should the rating be affected by that setting, or should the pane be rearranged so it doesn't look like it would be? Can be for a different PR though.

jburel:
The filtering "show All, mine" etc has no effect on rating.
Not planning to change that now, if we modify it, this will be done in another PR

mtbc:
Okay. :+1:

PR #3733, opened 2015-04-17
Get channels no re

will-moore:
Resolves a long-standing issue of unnecessarily loading the Rendering Engine when displaying the right "General" panel.

Blitz Gateway image.getChannels() loads the RE by default, allowing channel.getColor() etc. but this is not needed for the General panel.
A new method, image.getChannelsNoRE() does not attempt to load the rendering engine and this is now used for the General pane.

To test:
 - Import a big image (jpeg/png) and try loading General panel during Pyramid generation (when rendering engine throws MissingPyramidException. Panel should load without exception and show channel names.
 - Test that editing channel names works.
 - Also try viewing channel names for SPW images.

PR #3830, opened 2015-05-19
Improve rendering of Image metadata details

chris-allan:
This is a fairly straight-forward change that should help improve the rendering
performance of the right-hand metadata panel for Images. Previously, in order
to get the list of Channel labels to display, the template would have to load
the image's pixels, the pixel's channels, and then each channel's logical
channel. With this change, the relevant labels are retrieved directly via HQL
query.

### Benchmarking

The following are times taken to render the complete template for the right-hand side panel. The first number in each set represents a plate with 3 wells, the last two a fully-populated plate.

* Before changes:
- 0:00:01.550193
- 0:00:02.114671
- 0:00:02.787666

* After switching to `getChannelLabels`:
- 0:00:00.118799
- 0:00:02.053346
- 0:00:02.155348

* `getChannelLabels` combined with `original_file_paths` optimizations from #3695 
- 0:00:00.085634
- 0:00:00.080752
- 0:00:00.080852

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7512,7 +7512,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         q = self._conn.getQueryService()
         params = omero.sys.ParametersI()
         params.addId(self.getId())
-        query = "select lc.name, lc.emissionWave, lc.id "\
+        query = "select lc.name, lc.emissionWave, chan.id "\
                 "from LogicalChannel lc, Channel chan, Image img "\
                 "where chan.logicalChannel = lc "\
                 "and chan.pixels.image = img "\
@@ -7520,10 +7520,13 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         res = q.projection(query, params, self._conn.SERVICE_OPTS)
         ret = []
         for name, emissionWave, idx in res:
-            label = (name and name.val) or\
-                    (emissionWave and emissionWave.val) or\
-                    (idx and idx.val) or -1
-            ret.append(label)
+            if name is not None and len(name.val.strip()) > 0:
+                ret.append(name.val)
+            elif emissionWave is not None and\
+                    len(unicode(emissionWave.val).strip()) > 0:
+                ret.append(unicode(emissionWave.val))
+            else:
+                ret.append(unicode(idx.val))
         return ret
 
     @assert_re()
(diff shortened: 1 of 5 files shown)


will-moore:
Although you're not using the channel wrappers in the template, you still have ```{% with channels=manager.channel_metadata %}``` in the template, and the Channel wrappers are still being loaded with
```
#views.py
# Load channel metadata if appropriate
manager.channelMetadata(noRE=True)

# call this in container.py
self.channel_metadata = img.getChannels(noRE=noRE)
```

will-moore:
NB: this code changed recently in https://github.com/openmicroscopy/openmicroscopy/pull/3733
Otherwise, code looks fine and displayed channel names are same as before.

chris-allan:
Yep, thanks @will-moore. It's the curse of trying to port this stuff from `5.0.x` to `5.1.x`. I wanted to look at exactly that but I was bogged down by unit compatibility. I'm looking at it now so we can be sure to minimise that overhead as much as possible.

Couple more commits with respect to that to come.

chris-allan:
Additional commits to make use of `{% with ... %}`. This results in actual caching of results and avoids repeated recreation of `WellSampleWrapper` instances.

will-moore:
@chris-allan This line in ```container.py``` 
```
self.channel_metadata = img.getChannels(noRE=noRE)
```
is still getting called from this point in views.py when loading right panel
```
        # Load channel metadata if appropriate
        manager.channelMetadata(noRE=True)
```
even though this isn't being used. If you comment out ``` #manager.channelMetadata(noRE=True) ``` in views.py you'll see that loading right panel is faster!
So I would do this at a minimum.
Maybe the container.py code can also be removed - haven't looked too closely yet.

chris-allan:
Yeah, I was really cautious about removing that as other people might be relying on it being in place. It's certainly not used by any of our templates.  `channel_metadata` was added to the container for metadata details in 5.1.1 as part of #3733, no? I can certainly remove it but if other people are relying on it being in place then stuff is going to break. Quite spectacularly.

Thoughts?

chris-allan:
After further discussion with @will-moore, removed the load that is part of `webclient/views.py` in 4d541e49fe1d094f6f4453dd19ab0caf6dcb9861.

will-moore:
@chris-allan I can't find any use of ```channel_metadata``` except in ```views.py``` when loading the Metadata panel:
```
        manager.channelMetadata()
        for theC, ch in enumerate(manager.channel_metadata):
            ...
```
You're effectively replacing #3733 which added that call.

chris-allan:
Yep, which was added to the container interface as part of the 5.1.1 milestone. We know what's in use in our code. I'm only concerned about what someone else, out with the project, may have done while relying on the interface established where `channel_metadata` is populated and/or `channelMetadata()` has been called.

If you're happy accepting those possibilities and don't think they are a problem, then so am I.

aleksandra-tarkowska:
Noticed about 40% better response time. Looks good to me

sbesson:
Thanks guys.

PR #3832, opened 2015-05-20
Add IgnoreUnrecognizedVMOptions flag

joshmoore:
Our use of MaxPermSize has led to a number of warnings
in Java8 about the removal of that option. To prevent
these warnings, this adds another flag to ignore unknown
flags....

#### Testing ####

We're looking to remove this everywhere:
```
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
```
 * `./build.py` should no longer print it.
 * It should no longer show up in `var/log/master.err`

At the same time, this should not break any Java version from 6-8, and the jvmcfg unit tests should remain green.

### DIFF
--- a/components/tools/OmeroPy/test/unit/old_templates.xml
+++ b/components/tools/OmeroPy/test/unit/old_templates.xml
@@ -199,6 +199,7 @@
              See #4670
         -->
         <option>-XX:MaxPermSize=128m</option>
+        <option>-XX:+IgnoreUnrecognizedVMOptions</option>
         <target name="memcfg">
             <option>${omero.blitz.maxmemory}</option>
             <option>${omero.blitz.permgen}</option>
--- a/components/tools/OmeroPy/test/unit/old_templates.xml
+++ b/components/tools/OmeroPy/test/unit/old_templates.xml
@@ -199,7 +199,6 @@
              See #4670
         -->
         <option>-XX:MaxPermSize=128m</option>
-        <option>-XX:+IgnoreUnrecognizedVMOptions</option>
         <target name="memcfg">
             <option>${omero.blitz.maxmemory}</option>
             <option>${omero.blitz.permgen}</option>
--- a/build.py
+++ b/build.py
@@ -70,7 +70,11 @@ def find_java():
 
 
 def calculate_memory_args():
-    return "-Xmx600M -XX:MaxPermSize=256m".split(" ")
+    return (
+        "-Xmx600M",
+        "-XX:MaxPermSize=256m",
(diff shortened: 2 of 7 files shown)


mtbc:
I can test this on Oracle 6 or OpenJDK 7 if CI doesn't already cover them?

joshmoore:
JDK6 would be much appreciated being the most likely to have any issues. (I don't suspect though, due to http://mail.openjdk.java.net/pipermail/jdk6-dev/2009-October/000851.html)

components/tools/OmeroPy/test/unit/old_templates.xml
```diff
@@ -199,6 +199,7 @@
              See #4670
         -->
         <option>-XX:MaxPermSize=128m</option>
+        <option>-XX:+IgnoreUnrecognizedVMOptions</option>
```
manics: Shouldn't `old_templates.xml` be left unchanged?

mtbc:
@jburel turns out to have JDK6,7,8. :smiley:

jburel:
sorry I only have 6 and 8. I will build with those 2

jburel:
tested with

java version "1.6.0_65"
Java(TM) SE Runtime Environment (build 1.6.0_65-b14-466.1-11M4716)
Java HotSpot(TM) 64-Bit Server VM (build 20.65-b04-466.1, mixed mode)

No warning

java version "1.8.0_31"
Java(TM) SE Runtime Environment (build 1.8.0_31-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.31-b07, mixed mode)

no warning

jburel:
Do you want to update the doc in ```omero.properties``` indicating that the value will be ignored?

mtbc:
Should I do 7 or is it already adequately covered in CI?

joshmoore:
@jburel : pushed.

jburel:
@joshmoore: thanks

components/tools/OmeroPy/test/unit/old_templates.xml
joshmoore: If you say so. Pushing now.

joshmoore:
Pushed. The only 2 tests accessing `old_template.xml` pass:
```
  508  ./setup.py test -t test/unit/clitest/test_admin.py
  510  ./setup.py test -t test/unit/test_jvmcfg.py
```

sbesson:
Unit tests are passing (Cf Travis status), merging.

PR #3786, opened 2015-05-04
5.1.2: session api additions

joshmoore:
In order to complete the CLI session workflow changes, a few API changes were needed.

 * `omero.cmd.UpdateSessionTimeoutRequest` can now be used to modify the `timeToLive` and `timeToIdle` values. cc: @sbesson 
 * `omero.cmd.CurrentSessionsRequest` can now be used to query the in-memory state of all sessions as opposed to the in-database state. Inactive sessions will be timed-out in the process of querying and will *not* be included in the list. cc: @kennethgillen 

#### Testing ####

Check the various CLI commands, their various arguments, and their associated help:
 * `bin/omero sessions group`
 * `bin/omero sessions timeout`
 * `bin/omero sessions who`

#### Caveat ####

Clients should take care when talking to pre-5.1.2 servers to handle the `null` return value from `session.submit`:

https://github.com/openmicroscopy/openmicroscopy/pull/3786/files#diff-e44a2f62042695e699c5c5a9725de6d4R655
```
+        except omero.ClientError, ce:
+            if ce.err == "Null handle":
+                v = client.getConfigService().getVersion()
+                self.ctx.die(561,
+                             "Operation unsupported. Server version: %s" % v)
```

PR #3834, opened 2015-05-20
CLI: implement timeout logic 

sbesson:
Following https://github.com/openmicroscopy/openmicroscopy/pull/3786, this PR:

- implements the logic for the `--timeout` argument using a new property `omero.timeout` to minimize the API modification under `omero.util.sessions`
- add an integration test covering this implementation
- sets the default timeout for `--sudo` sessions using the `omero.sessions.timeout` configuration value

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/sessions.py
+++ b/components/tools/OmeroPy/src/omero/util/sessions.py
@@ -347,8 +347,13 @@ class SessionsStore(object):
             principal.name = name
             principal.group = props.get("omero.group", None)
             principal.eventType = "User"
+            try:
+                cs = sf.getConfigService()
+                idleTimeout = long(cs.getConfigValue('omero.sessions.timeout'))
+            except:
+                idleTimeout = 0L
             sess = sf.getSessionService().createSessionWithTimeouts(
-                principal, 0, 0)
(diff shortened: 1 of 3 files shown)


components/tools/OmeroPy/src/omero/util/sessions.py
```diff
@@ -347,8 +348,13 @@ def create(self, name, pasw, props, new=True, set_current=True, sudo=None):
             principal.name = name
             principal.group = props.get("omero.group", None)
             principal.eventType = "User"
+            try:
+                cs = sf.getConfigService()
+                idleTimeout = long(cs.getConfigValue('omero.sessions.timeout'))
+            except:
+                idleTimeout = 0L
```
joshmoore: Perhaps default to 1 hour and warn? This in a loop (with no logout) would crash the server.

sbesson: Improved by retrieving the `timeToIdle` from the initial (sudoer) session instead of querying the server again.

joshmoore:
Manual testing on trout and https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration-python/456/ all say, :thumbsup:

PR #3835, opened 2015-05-20
Re float

jburel:
Remove usage of Range and loop to determine the interval
Add units test for image with large range
To test
 * Import the image attached to https://trello.com/c/OfYVeeTS/234-nightshade-dsid-66704-not-thumbnailling.
 * Check that the image can be viewed.

### DIFF
--- a/components/rendering/src/omeis/providers/re/quantum/Quantization_32_bit.java
+++ b/components/rendering/src/omeis/providers/re/quantum/Quantization_32_bit.java
@@ -212,7 +212,6 @@ public class Quantization_32_bit extends QuantumStrategy {
     public Quantization_32_bit(QuantumDef qd, Pixels pixels) {
         super(qd, pixels);
         values = CacheBuilder.newBuilder()
-                .maximumSize(MAX-MIN+1)
                 .expireAfterWrite(10, TimeUnit.MINUTES)
                 .build(new CacheLoader<Double, Integer>() {
                     public Integer load(Double key) throws Exception {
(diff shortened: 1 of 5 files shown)


gusferguson:
@jburel 

Tested with OMERO.insight-5.1.1-485-4f9ce01-ice35-b511-mac_Java7+ user-9 and image 4025756.ome.tiff

Image imports normally, thumbnail created and can be viewed.
Not sure is the data is displaying correctly as image has no recognisable appearance - see screenshot.
If this is expected appearance of image then good to merge.

main

jburel:
@gusferguson: that's how it looks when you open it in ImageJ

gusferguson:
@jburel  - Good to merge then.

components/rendering/src/omeis/providers/re/quantum/Quantization_32_bit.java
```diff
@@ -213,7 +212,6 @@ private int _quantize(double value)
     public Quantization_32_bit(QuantumDef qd, Pixels pixels) {
         super(qd, pixels);
         values = CacheBuilder.newBuilder()
-                .maximumSize(MAX-MIN+1)
```
joshmoore: Not for this PR, but we might want to move all these specifications out to etc/omero.properties so that if they start to affect memory, a sysadmin could lower them. cc: @mtbc cf:
 * https://github.com/joshmoore/openmicroscopy/blob/10661-dynamic-params/components/blitz/src/ome/services/blitz/util/ParamsCache.java#L105
 * https://github.com/joshmoore/openmicroscopy/blob/10661-dynamic-params/etc/omero.properties#L196

PR #3836, opened 2015-05-20
Original file download

will-moore:
Fixes the download vv viewing of files in browser.

To test:
 - Run a script, E.g. split view figure and inspect the resulting image and stdout (i) from the Activities dialog. These should open in the browser instead of downloading.
 - Check that downloading still works - Select image(s) with more than 1 original file > Download. In the downloader dialog, click on individual files to download. These should download directly.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/download_placeholder.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/download_placeholder.html
@@ -124,7 +124,7 @@ Download files
         {% for f in fs %}
         <tr>
             <td>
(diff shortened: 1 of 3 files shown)


gusferguson:
@will-moore 

Tested using trout merge user-3 Test 15-05-14 07:58:50

Script results 
- works as expected.

Download 
- works as expected with Leica .lei files
- when downloading 2 file big-scn ID 22470 - they download straight off without opening dialog - this may be expected.

Good to merge if SCN not an issue.

will-moore:
@gusferguson That SCN image only has a single file in it's fileset, so yes - it is expected to download directly. Cheers.

PR #3839, opened 2015-05-21
OMERO.matlab: add group support

sbesson:
Following separate discussions with @dpwrussell and @chris-allan, this PR starts adding cross-group support to OMERO.matlab. While `AdminService.setSecurityContext()` can be used to achieve this, it has the downside of potentially changing the default group of the current user. The proper way to do cross-group operations is to pass a map to ice with `omero.group` set.

Most of the relevant functions in OMERO.matlab now support a `group` parameter/key value allowing to specify the `id` of the group (or `-1` for cross-group queries).

The signature of the getter methods with an explicit identifer is modified to be cross-group and cross-owner by default (BREAKING change). To restore the previous current behavior (current session owner, current group), the `owner` and `group` parameter/key values can be passed explicitly

```
getProjects(session, projectIds); % returns all projects of identifier ids
getProjects(session, projectIds, 'owner', ownerId, 'group', 'groupId`); % returns all projects of identifier ids owned by user ownerId in group groupId
```

Most writing functions are also modified to support a group argument.

### DIFF
--- a/components/tools/OmeroM/src/annotations/writeFileAnnotation.m
+++ b/components/tools/OmeroM/src/annotations/writeFileAnnotation.m
@@ -13,6 +13,9 @@ function fa = writeFileAnnotation(session, filePath, varargin)
 %    fa = writeFileAnnotation(session, filePath, 'description',
 %    description) also sets the description of the file annotation.
 %
+%    fa = writeFileAnnotation(session, filePath, 'group', groupid)
+%    sets the group.
+%
 %    Examples:
 %
 %        fa = writeFileAnnotation(session, filePath)
@@ -48,6 +51,7 @@ ip.addRequired('filePath', @(x) exist(x, 'file') == 2);
 ip.addParamValue('mimetype', '', @ischar);
 ip.addParamValue('namespace', '', @ischar);
 ip.addParamValue('description', '', @ischar);
+ip.addParamValue('group', [], @(x) isscalar(x) && isnumeric(x));
 ip.parse(session, filePath, varargin{:});
 
 % Create original file
@@ -59,7 +63,8 @@ if ~isempty(ip.Results.mimetype),
     originalFile.setMimetype(rstring(ip.Results.mimetype));
 end
 
-originalFile = updateOriginalFile(session, originalFile, filePath);
+originalFile = updateOriginalFile(...
+    session, originalFile, filePath, varargin{:});
 
 % Create a file annotation
 fa = omero.model.FileAnnotationI;
@@ -74,4 +79,9 @@ if ~isempty(ip.Results.namespace),
 end
 
 % Save the file annotation
-fa = session.getUpdateService().saveAndReturnObject(fa);
+context = java.util.HashMap;
+if ~isempty(ip.Results.group)
+    context.put(...
+        'omero.group', java.lang.String(num2str(ip.Results.group)));
+end
+fa = session.getUpdateService().saveAndReturnObject(fa, context);
--- a/components/tools/OmeroM/src/annotations/getAnnotations.m
+++ b/components/tools/OmeroM/src/annotations/getAnnotations.m
@@ -1,4 +1,4 @@
-function annotations = getAnnotations(session, ids, type)
+function annotations = getAnnotations(session, ids, type, varargin)
 % GETANNOTATIONS Retrieve annotations from a given type from the OMERO server
 %
 %   annotations = getAnnotations(session, ids, type) returns all the 
@@ -7,6 +7,8 @@ function annotations = getAnnotations(session, ids, type)
 %   Examples:
 %
 %      annotations = getAnnotations(session, ids, type);
+%      annotations = getAnnotations(session, ids, type, 'group', groupId);
+%        returns annoations from the input group specfied by groupId.
 %
 % See also: GETANNOTATIONTYPES, GETDOUBLEANNOTATIONS, GETCOMMENTANNOTATIONS,
 % GETFILEANNOTATIONS, GETLONGANNOTATIONS, GETTAGANNOTATIONS,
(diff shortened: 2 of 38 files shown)


jburel:
user should probably also be considered, especially when functionalities "do something for someone" are pushed e.g. analyse image from user-1 when logged as user-3

sbesson:
@jburel: if by user you mean object's owner, the `getXXX` methods already supportx either `'owner', userId` or more generically `'parameters', parameters' accepting a `ParametersI()` object. So you can do:

```
all_images_userid = getImages(s, 'owner',userId, 'group', -1);
```

Discussing with @joshmoore, I could go one step further and also implement a very generic `map` parameter in a similar way as the `parameters` option described above. This would allow an advanced user to have the most granularity while retrieving objects using the following workflow:

```
parameters = ParametersI();
parameters.set(xxx);
map = java.util.HashMap;
map.put(key, value);
objects = getObjects(s, type, ids, 'parameters',parameters, 'map', map);
```

jburel:
@sbesson: I agree that we should go a step further. 
since owner might one think (method), but I was thinking of cases like "annotator" where user is the owner of the links not the owner of the data.
so map will help covering this extension point.

emilroz:
Image retrieval does not work both though getImages and getObjects.

sbesson:
@emilroz: which commit did you test. My last set of commits esp. 9ba7395 should include the `group` support for `getImages`.

emilroz:
@sbesson only the first two commits. Will retest with the latest additions.

sbesson:
See also https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/333 for the amended training example.

emilroz:
@sbesson I tested get - Project, Dataset, Screens, Plates and Images with the 'group' parameter all works as expected.

sbesson:
Putting on hold. Let us take integrate this properly for 5.1.3.

sbesson:
Speaking to @emilroz earlier this morning, additional commits coming to:
- extend this multi-group workflow to other methods (esp. RawPixelsStore, RawFileStore)
- rename the generic `map` key/value parameter as `context`

components/tools/OmeroM/src/annotations/writeFileAnnotation.m
```diff
@@ -59,7 +63,8 @@
     originalFile.setMimetype(rstring(ip.Results.mimetype));
 end
 
-originalFile = updateOriginalFile(session, originalFile, filePath);
+originalFile = updateOriginalFile(...
+    session, originalFile, filePath, varargin{:});
```
joshmoore: Copying https://github.com/snoopycrimecop/openmicroscopy/commit/3a07788f721d13ed3bbce96a31f06b7d19040c53#commitcomment-11597771

> See https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-tracking/384/console, if description (or likely namespace) is passed, it will not pass the updateOriginalFile input check

emilroz: @joshmoore commit 15b29dcaec4f170e50f7bb42e5c920dde7382b8a should have fixed this issue.

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-training/353/console
```
Creating local file with content: file annotation content modified
Expected a string for the parameter name, instead the input type was 'cell'.
```

emilroz:
@sbesson training example should pass now.

sbesson:
With the split between `dev_5_1` and develop, removing `ON_HOLD` marker for consideration.

jburel:
https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-training/51/ is green

jburel:
in ```getObjects```:

 * % If no input id, return the objects owned by the session user in the    % current context => Missing return

in ```examples/Training/matlab/ReadData.m```
 * ```disp('Retrieving projects across owned by any user in the current group')``` => "across" to be removed
 * ```fprintf(1, 'Retrieving images contained in project %g\n', projectId);
   images3 = getImages(session, 'dataset', projectId);``` =>should be project
 * ```disp('Retrieving all screens and orphaned plates owner by the session user')``` => owned not owner
 * ```fprintf(1, '  Found %g datasets\n', numel(allPlatesAllGroups));``` => should be plates

Also a corresponding doc PR to expose the group option will be required i.e. add more examples to the matlab page (advanced section??)

sbesson:
@jburel: pushed. Will open a first doc PR tonight (and drive other doc improvements via the Trello card).

jburel:
--no-rebase

jburel:
@sbesson: will it be possible to rebase that PR?

sbesson:
@jburel: as marked in the description, this PR is breaking some of the OMERO.matlab API so I don't think it is suitable for a rebase.

chris-allan:
This breaks some of those semantics but does not break the API, correct? The reason why I brought this up with @jburel earlier today was that OMERO.matlab in 5.1.x is essentially unusable, and will remain so until 5.2.0, in a multi-group context without this or similar PR.

jburel:
Plan to discuss with @sbesson later today (other meetings before!)

sbesson:
@chris-allan/@jburel: agreed that OMERO.matlab in multi-group context does not work at all on 5.1.x at the moment and this PR is adding support for this. Also agreed this PR is introducing a semantic breaking  change rather than a syntatic breaking change. My core feeling is that for any consumer of the API, this will and should be considered as a breaking change.
Assuming we rebase this PR on 5.1.x, we will want to be extremely clear in terms of documentation of its implications and how consumer code should be modified. Happy to discuss more.

jburel:
Discussed with @sbesson:
We are going to review existing tools mtools, flimfit and u-track to get usage and example.
Need to add warning so people know exactly the group context they are working on

sbesson:
--rebased-to #4111

PR #3841, opened 2015-05-21
ImportConfig: be defensive when casting

ctrueden:
In case the SLF4J logging implementation in play is not logback classic,
let's be defensive before casting.

### DIFF
--- a/components/blitz/src/ome/formats/importer/ImportConfig.java
+++ b/components/blitz/src/ome/formats/importer/ImportConfig.java
@@ -330,10 +330,16 @@ public class ImportConfig {
          } else {
              level = Level.toLevel(levelString);
          }
-         ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("ome.formats")).setLevel(level);
-         ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("ome.services.blitz")).setLevel(level);
-         ((ch.qos.logback.classic.Logger)LoggerFactory.getLogger("ome.system")).setLevel(level);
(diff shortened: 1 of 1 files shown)


ctrueden:
There are still other uses of ch.qos.logging of course:
```
$ git grep -l ch.qos.logback | grep \.java$
components/blitz/src/ome/formats/importer/ImportConfig.java
components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
components/blitz/test/ome/formats/utests/IniFileLoaderTest.java
components/blitz/test/ome/formats/utests/MetadataValidatorTest.java
components/common/src/ome/system/metrics/DefaultMetrics.java
components/common/src/ome/util/logback/SiftingWhileRejectingDefaultAppender.java
components/insight/SRC/org/openmicroscopy/shoola/env/log/LoggerImpl.java
```
But this might help reduce @imunro's woes.

joshmoore:
I'd think for the purpose of matlab usage, only:
```
components/blitz/src/ome/formats/importer/ImportConfig.java
```
and potentially:
```
components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
```
would be of interest. The rest is server-side and that is not-interface-able.

joshmoore:
Merging. I'm not completely sure that we aren't better suited to warn the user (via exception) that we don't actually support SL4J implementations other than logback at the moment, but if we're going to knowingly be that strict, we should at least make it configurable. **So** merging under the understanding that this doesn't imply overall support but just a short-term workaround for an annoying behavior.

PR #3844, opened 2015-05-26
prevent group owner from self removal

aleksandra-tarkowska:
Group owner shouldn't be able to remove himself from the group he can manager users.

To test log in as group owner. Go to User Settings -> My Group and check if current user is highlited and there is no X allowing to delete from current group.

firefoxscreensnapz051

cc: @will-moore

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form_owner.html
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form_owner.html
@@ -43,11 +43,10 @@
                 var highlightCurrent = function() {
                     var userId = {{ userId }};
                     var selected = $.grep($('#id_members').data('chosen').results_data, function(item){
-                        if (parseInt(item.value) === userId) {
-                            $("#id_owners_chzn_c_"+item.array_index).addClass('search-choice-current').find("a").first().unbind("click").remove();                            
-                            $("#id_members_chzn_c_"+item.array_index).addClass('search-choice-current').find("a").first().unbind("click").remove();                            
-                        }
+                        return item.value == userId;
                     });
+                    $("#id_owners_chosen a[data-option-array-index='"+selected[0].options_index+"']").parent().addClass('search-choice-current').find("a").first().unbind("click").remove();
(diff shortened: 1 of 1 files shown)


pwalczysko:
Note : Filed a blocker testing this PR https://trac.openmicroscopy.org/ome/ticket/12902#ticket - probably not caused by this PR, but the fix for it might conflict with this one. The workflow mentioned here is working fine when tested as prescribed by @aleksandra-tarkowska but according to @joshmoore some communication and conflict solving might be needed here to sort out the blocker.

sbesson:
Retested on eel.openmicroscopy.org using the `owner` user on multiple groups with various permissions. Group owner form correctly prevents the current `owner` to remove himself from any of the groups as described in the PR.

mtbc:
Seems to be working fine, e.g.,

group-owner

and,

group-owner

I am not actually sure that it's a feature that group owners can't remove themselves but I assume that's already been determined in client/UX meetings as how things should be. :+1:

will-moore:
@mtbc Yes, if you removed yourself as group leader, this would be "undoable" since you would no longer have permissions to edit (same a server Admin can't remove themselves from system group).

pwalczysko:
All works here as described in the PR. The Lock Timeout is a separate problem. Ready to merge.

joshmoore:
Merging for 5.1.2

mtbc:
@will-moore: We allow users to do all manner of undo-able things; this one's unusually fixable in that you can get another admin to put you back. Happy to accept judgment that it's still too severe though!

pwalczysko:
@mtbc : The reasoning is rather in the point ``...you can get another admin...`` - the group owners are not necessarily IT-savvy - their removal of themselves out of the group ownership is usually highly undesirable, and any mistake is costly, because admin is far away and busy (typical situation in many if not all institutions). It was also the admin who made them a group owner in the first place, so it makes sense also from the workflow viewpoint.

mtbc:
Could use a dialog box to confirm it's what they meant to do, though? Let's hope these group owners don't accidentally delete a bunch of data, which even finding an admin can't fix. With the data loss that can come from something like an accidental downgrade to private, they have better be somewhat savvy! Happy to accept your judgment on what it is okay to wholly prohibit them from doing.

PR #3846, opened 2015-05-26
Roi table ij

jburel:
Problem reported during the rc1 testing
To test:
 * Open an image stored in OMERO in imageJ
 * Draw rois. Do not open the results table in ImageJ
 * Select "Plugins>OMERO>Save ROIs to OMERO"
 * A Saving dialog will pop up. Keep the default option i.e. save rois and table
 * Check that the rois are saved and no table are saved.

cc @pwalczysko @bramalingam

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ResultsSaver.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ResultsSaver.java
@@ -177,7 +177,7 @@ public class ResultsSaver
                     //create a tmp file.
                     File f = createFile(img);
                     if (f != null) {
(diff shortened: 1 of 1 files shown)


pwalczysko:
After preliminary test this works as expected (used 5.1.2 rc1 bioformats_package.jar and ImageJ).
ROIs are saved in case no Results table is present.
ROIs and Results table are saved in case Results table is present.
Results table only is saved in case ``Save Results to OMERO`` is executed and in the dialog the ROIs are unticked.

Leave @bramalingam to have one more check in the morning, but this passes for me.

bramalingam:
Tested both FIJI and ImageJ. Works as expected. Please merge.

joshmoore:
Merging for 5.1.2

PR #3852, opened 2015-05-27
use _waitOnCmd to handle async request

aleksandra-tarkowska:
This should resolve https://trac.openmicroscopy.org/ome/ticket/12902

cc: @joshmoore

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -39,7 +39,6 @@ import omero.scripts
 
 from omero.rtypes import rbool, rint, rstring, rlong, rlist, rtime, unwrap
 from omero.model import ExperimenterI, ExperimenterGroupI
-from omero.cmd import Chmod
 
 from omero.gateway import AnnotationWrapper
 from omero.gateway import OmeroGatewaySafeCallWrapper
@@ -1453,11 +1452,14 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
 
         perms = str(permissions)
         logger.debug("Chmod of group ID: %s to %s" % (group.id, perms))
-        command = Chmod(type="/ExperimenterGroup",
-                        id=group.id,
-                        permissions=perms)
-        cb = self.c.submit(command)
-        cb.close(True)
+        command = omero.cmd.Chmod(type="/ExperimenterGroup",
+                                  id=group.id,
+                                  permissions=perms)
+        handle = self.c.sf.submit(command, self.SERVICE_OPTS)
+        try:
+            self._waitOnCmd(handle)
+        finally:
+            handle.close()
 
     def saveObject(self, obj):
         """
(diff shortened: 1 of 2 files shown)


joshmoore:
Other locations for discussion:
```
OmeroPy/src/omero/gateway/__init__.py:            handle = self._conn.c.sf.submit(doall, self._conn.SERVICE_OPTS)
OmeroPy/src/omero/gateway/__init__.py:        handle = self.c.sf.submit(doall, self.SERVICE_OPTS)
OmeroPy/src/omero/gateway/__init__.py:        prx = self.c.sf.submit(chmod)
OmeroPy/src/omero/gateway/__init__.py:        prx = self.c.sf.submit(da, ctx)
OmeroPy/src/omero/gateway/__init__.py:        handle = self._conn.c.sf.submit(req)
OmeroWeb/omeroweb/webadmin/views.py:                    handle = conn.c.sf.submit(req)
OmeroWeb/omeroweb/webadmin/views.py:            handle = conn.c.sf.submit(req)
```

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -1453,11 +1452,14 @@ def updatePermissions(self, group, permissions):
...
         logger.debug("Chmod of group ID: %s to %s" % (group.id, perms))
-        command = Chmod(type="/ExperimenterGroup",
-                        id=group.id,
-                        permissions=perms)
-        cb = self.c.submit(command)
-        cb.close(True)
+        command = omero.cmd.Chmod(type="/ExperimenterGroup",
+                                  id=group.id,
+                                  permissions=perms)
+        handle = self.c.sf.submit(command, self.SERVICE_OPTS)
```
joshmoore: Why are we passing SERVICE_OPTS  and not before? Can we go for a minimal patch at this point? Really the only thing that needs to happen in this whole PR is the addition of `loops=50`.

joshmoore:
cc: @mtbc 
All the other uses of `submit()` are assumed to be generally fast enough and since this is the only one which has shown problems during testing, it will be extended (x5 to 25 seconds) with the hope that that will suffice but we could still see `LockTimeout` exceptions in the field which may well trigger a refactoring of activities for 5.1.x so that we can more accurately display the state of the asynchronous operations.

aleksandra-tarkowska:
@joshmoore I added failontimeout to other places where we were silently failing. I hope that is ok

joshmoore:
Closing this PR for a review in 5.1.3. There was some discussion about more generally how the sync-v-async methods work. To not try too much at this date, @aleksandra-tarkowska will open a PR just with `..., loops=$N,...` on changePermissions. The other locations that are discussed here (along with https://github.com/openmicroscopy/openmicroscopy/blob/5f0e355b07782c8c6e745f53ff34ce41b52f94fd/components/tools/OmeroPy/src/omero/gateway/__init__.py#L3794 cc: @ximenesuk) will still need a review and some form of correction since there's a race condition on whether or not the server call will actually be cancelled.

See: https://trello.com/c/dLV0ykkO/528-asynchronous-calls

PR #3855, opened 2015-05-28
Save z changes

dominikl:
Enables the save button after Z or T has been changed. There are still some minor issues with the z/t sliders if both views (preview and full viewer) are open at the same time (z/t slider then shows the last selected z/t state, even if the z/t changes haven't been stored), not going to fix that in this PR.


pwalczysko:
2 issues:
  1. RFE: Even when the video player button in the Full Viewer is used for z/t change, enable the Save button
  2. Conceptual issue: I think we agreed that the Save to All action should NOT propagate the same z/t selection over all the images in the Dataset -> but this is now the case unfortunately

pwalczysko:
cc @gusferguson @jburel @will-moore

dominikl:
RE 1) Will implement this within this PR;
RE 2) Not related to this PR, PR just enables the save button, does not modify any logic about saving rendering settings.

pwalczysko:
Works as expected even with video button. Ready to merge.

PR #3856, opened 2015-05-28
Image J

jburel:
Fix workflow issue noticed while reviewing 5.1.2
To test:
 * Open an image from disk in imageJ
 * Import the image in omero
 * Draw roi on the image.
 * Click Save ROIs.
 * The dialog indicating that the image is not imported should not pop up. The roi should be added to the image previously imported

cc @pwalczysko

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
@@ -928,7 +928,13 @@ public class FileImportComponent
 		} else if (image instanceof List) {
 			List<ThumbnailData> list = new ArrayList<ThumbnailData>((List) image);
 			int m = list.size();
-			imageLabel.setData(list.get(0));
+			ThumbnailData data = list.get(0);
+			long iid = data.getImageID();
+			if (data.getImage() != null) {
+			    iid = data.getImage().getId();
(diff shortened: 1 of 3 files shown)


pwalczysko:
Bug: Although the ``dialog taht the image is not imported`` does not pop up as expected, and the activities window shows the ``Results saved`` message, the ROI is not being saved.

jburel:
@pwalczysko: sorry. It looks like I was looking at a different image. The group context is -1 that's why it failed

jburel:
@pwalczysko: problem solved

pwalczysko:
ImageJ works as expected.

I know this is not related to this PR, but the Console in Fiji keeps popping up after the start of ``insight-ij`` plugin saying

```
no update needed
New update(s) found: 2.4.7 [http://www.terracotta.org/confluence/display/release/Release+Notes+Ehcache+Core+2.4]. Please check http://ehcache.org for the latest version.
[Fatal Error] Info.plist:1:1: Content is not allowed in prolog.
```

Nothing we can do about it ? (looks quite annoying during demos)

pwalczysko:
Regarding https://github.com/openmicroscopy/openmicroscopy/pull/3856#issuecomment-109930178 just as a question, this is ready to merge, works fine both in IJ and Fiji.

jburel:
@pwalczysko: not related to this PR, due to one of the dependencies we currently use.

PR #3825, opened 2015-05-16
Updates to omero.properties 

joshmoore:
See: https://trello.com/c/cATe4yap/306-5-1-omero-properties

PR #3857, opened 2015-05-28
Server configuration: autogenerated docs improvement

sbesson:
This PR follows up on https://github.com/openmicroscopy/openmicroscopy/pull/3825 which was adding DB properties parsing to the documentation page. Main changes are described as  follows:

- a new configuration section is introduced listing the `Database` server properties
- the `FS` section is renamed as `Binary repository` in the docs to unify with the existing semantics
- a new top-level `Core` section is created with hyperlinks to the fundamental server properties (any better name?)
- in terms of underlying logic, a `Header` object is added allowing to define for each configuration section a display name, a reference (to prevent breakage of `fs_configuration` refs) and a stub for a description (which could be displayed under the title).

To test this PR, check the output of the rST file generated by

```
bin/omero config parse --rst
```

Also check

```
bin/omero config parse --keys
bin/omero config parse --defaults
bin/omero config parse --headers
```
work as expected.

/cc @joshmoore @hflynn @mtbc @ximenesuk

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -67,8 +67,8 @@ omero.db.prepared_statement_cache_size=10
 # you can modify the defaults.
 #
 # Note: some other properties are defined in
-# the file etc/profiles/${omero.db.profile}
-# Especially of importance is omero.db.port
+# the file :file:`etc/profiles/${omero.db.profile}`
+# Especially of importance is :property:`omero.db.port`
 omero.db.profile=psql
 
 # Whether JMX statistics are collected
@@ -531,7 +531,7 @@ omero.search.ram_buffer_size=64
 # At least one user-owned directory must be included in the path.
 #
 # The template path is created below :property:`omero.managed.dir`,
-# e.g. /OMERO/ManagedRepository/<omero.fs.repo.path>/
+# e.g. :file:`/OMERO/ManagedRepository/${omero.fs.repo.path}/`
 omero.fs.repo.path=%user%_%userId%//%year%-%month%/%day%/%time%
 
 # Rules to apply to judge the acceptability of FS paths for writing into
--- a/components/tools/OmeroPy/src/omero/install/config_parser.py
+++ b/components/tools/OmeroPy/src/omero/install/config_parser.py
@@ -143,8 +143,11 @@ and :doc:`Windows <windows/server-installation>` pages, as well as the
 
 .. _core_configuration:
 
-Core
-----
+Mandatory properties
+--------------------
+
+The following properties need to be correctly set for all installations of the
+OMERO.server. Depending on your set-up, default values may be sufficient.
 
 - :property:`omero.data.dir`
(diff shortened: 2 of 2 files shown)


sbesson:
Deployed at https://www.openmicroscopy.org/site/support/omero5.1-staging/sysadmins/config.html

hflynn:
Staging page looks fine to me. I can't think of a better term than 'core'.

joshmoore:
```
Core
----
  omero.data.dir
  omero.db.host
  omero.db.name
  omero.db.pass
```
doesn't really work for me. Either I think we need some text explaining how we mean "core" or we need something to explain why this section is so short. "Mandatory properties"? "Critical properties (see below)"?

sbesson:
As said, above, adding a short text below a given configuration section should be easy enough with the new `Header` object. Happy to try that if we have a template.

etc/omero.properties
```diff
@@ -67,8 +67,8 @@ omero.db.prepared_statement_cache_size=10
 # you can modify the defaults.
 #
 # Note: some other properties are defined in
-# the file etc/profiles/${omero.db.profile}
-# Especially of importance is omero.db.port
+# the file :file:`etc/profiles/${omero.db.profile}`
```
manics: The '`{` `}` are missing from the rendered text.

etc/omero.properties
```diff
@@ -531,7 +531,7 @@ omero.search.ram_buffer_size=64
 # At least one user-owned directory must be included in the path.
 #
 # The template path is created below :property:`omero.managed.dir`,
-# e.g. /OMERO/ManagedRepository/<omero.fs.repo.path>/
+# e.g. :file:`/OMERO/ManagedRepository/${omero.fs.repo.path}/`
```
manics: Missing `{` `}`

manics:
Since you've started modifying `etc/omero.properties` I've noticed a few other entries which could do with some sphinx formatting (e.g. filenames, java classnames....)... but I guess you'd rather not do that here.

sbesson:
Feel free to either report them or open a PR against this branch. More commits will be needed to handle this top-level "core" properties section.

etc/omero.properties
sbesson: Understood, this means Sphinx must escape those then. Should I remove the `file` markup and default to regular verbatim?

manics: That's probably best. Someone could go through the whole file and add the relevant markup in a future PR?

joshmoore: cF. https://trello.com/c/cATe4yap/306-5-1-omero-properties

manics:
Also `bin/omero config parse --values` doesn't work (did you meant `--headers`?

sbesson:
Yes I meant the different options of `bin/omero config parse` i.e. `--defaults` and `--headers` notably.

joshmoore:
Reads much better to me. Thanks.

PR #3860, opened 2015-06-05
Tag update

jburel:
Fix possible NPE when updating a tag. reported via QA system

To test:
 * Select a tag in the tree (left-hand panel)
 * Update the tag (right-hand panel)
 * Save


gusferguson:
@jburel 

Tested using OMERO.insight-5.1.2-42-601658d-ice35-b525-mac_Java7+ trout user-3 read-only-1.

No error message on changing description or name, rating or adding tag.
Good to merge.

PR #3861, opened 2015-06-08
Five failing tests that break junitxml.

ximenesuk:
This is a temporary PR to capture the junitxml output of several failing tests. Fails such as these have in the past generated 404s in the derived html. This PR will be closed once it has been included once.


ximenesuk:
Thanks @sbesson the plugin seems to have fixed all the non-standard failures:

https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-breaking-integration-python/285/testReport/

which means I can close an issue over at pytest HQ!

PR #3863, opened 2015-06-09
bulk annotations right panel

will-moore:
See https://trello.com/c/MGeMSVXY/239-hdf-table-data-display-for-selected-wells-in-web

Load and display bulk annotations in metadata General tab.
Uses the same code as in the full image viewer.
Bulk annotations can be added as described: https://github.com/openmicroscopy/ome-internal/blob/master/testing_scenarios/BulkAnnotations.txt

NB: Bulk annotations are not filtered by owner as other annotations are. Is this needed?

screen shot 2015-06-09 at 14 15 18

### DIFF
--- a/components/tests/ui/robot_setup.sh
+++ b/components/tests/ui/robot_setup.sh
@@ -16,6 +16,8 @@ USER_PASSWORD=${USER_PASSWORD:-ome}
 CONFIG_FILENAME=${CONFIG_FILENAME:-robot_ice.config}
 IMAGE_NAME=${IMAGE_NAME:-test&sizeZ=3&sizeT=10.fake}
 TINY_IMAGE_NAME=${TINY_IMAGE_NAME:-test.fake}
+PLATE_NAME=${PLATE_NAME:-SPW&plates=1&plateRows=1&plateCols=2&fields=1&plateAcqs=1.fake}
+BULK_ANNOTATION_CSV=${BULK_ANNOTATION_CSV:-bulk_annotation.csv}
 
 # Create robot user and group
 bin/omero login root@$HOSTNAME:$PORT -w $ROOT_PASSWORD
@@ -25,12 +27,21 @@ bin/omero user add $USER_NAME $USER_NAME $USER_NAME $GROUP_NAME $GROUP_NAME_2 --
 bin/omero user joingroup --name $USER_NAME --group-name $GROUP_NAME --as-owner
 bin/omero logout
 
-# Create fake file
+# Create fake files
 touch $IMAGE_NAME
 touch $TINY_IMAGE_NAME
+touch $PLATE_NAME
+
+# Create batch annotation csv
+echo "Well,Well Type,Concentration" > "$BULK_ANNOTATION_CSV"
+echo "A1,Control,0" >> "$BULK_ANNOTATION_CSV"
+echo "A2,Treatment,10" >> "$BULK_ANNOTATION_CSV"
 
 # Create robot setup
 bin/omero login $USER_NAME@$HOSTNAME:$PORT -w $USER_PASSWORD
+# Parse the sessions file to get session key
+key=$(grep omero.sess $(bin/omero sessions file) | cut -d= -f2)
+echo "Session key: $key"
 nProjects=1
 nDatasets=1
 nImages=2
@@ -57,6 +68,12 @@ do
   bin/omero import -d $delDs $TINY_IMAGE_NAME --debug ERROR
 done
 
+# Import Plate
+bin/omero import $PLATE_NAME --debug ERROR > plate_import.log 2>&1
+plateid=$(sed -n -e 's/^Plate://p' plate_import.log)
+# Use populate_metadata to upload and attach bulk annotation csv
+python lib/python/omero/util/populate_metadata.py -k $key Plate:$plateid $BULK_ANNOTATION_CSV
+
 # Logout
 bin/omero logout
 
@@ -70,3 +87,4 @@ echo "omero.datasetid=${dataset##*:}" >> "$CONFIG_FILENAME"
 
 # Remove fake file
 rm *.fake
+rm $PLATE_NAME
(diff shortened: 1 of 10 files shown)


pwalczysko:
On the data mentioned in https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/BulkAnnotations.html
Nevertheless, I have a slight formatting issue, the BULK ANNOTATIONS header is creeping into the first line of the annotation when done on trout merge.
https://trout.openmicroscopy.org/merge/webclient/?show=well-313305 login as user-4


screen shot 2015-06-10 at 10 38 42

pwalczysko:
I think the question about the filtering (Show all etc.) is for @eleanorwilliams

eleanorwilliams:
From the IDR point of view, I guess no it doesn't matter if you can't filter by owner as everything will have one owner, but as a general part of Omero, yes it would be good to be able to only search in your own bulk annotations.

jburel:
Probably not required to have the term "Bulk Annotations".

joshmoore:
What will be the alternative? I'd suggest we synchronize including what shows up via `populate_metadata` and other CLI components.

jburel:
The source does not really matter, it is displayed as non editable (at the moment) key-value list. Since it is under the "annotations" section I will probably not add a header 
@gusferguson what do you think?

joshmoore:
Some actions I can think of it (eventually) mattering for are:
 * download the related HDF5
 * find other objects which are related in the same way (i.e. go up to plate/screen)
 * switch between similar sources (i.e. a second bulk annotation)

will-moore:
@pwalczysko Fixed the layout issue (not sure why I didn't see this before).

Also added test for populate_metadata.py cc @joshmoore.

pwalczysko:
Note that in case the plate is inside a screen, I am still experiencing the bug https://trac.openmicroscopy.org/ome/ticket/12211.  In Web, the Keywords are actually populated, but the values are not. See screenshot.

screen shot 2015-06-10 at 16 27 42

joshmoore:
```
./components/tools/OmeroWeb/omeroweb/webgateway/views.py:2305:80: E501 line too long (84 > 79 characters)
```

jburel:
@pwalczysko: The error is due to populate metadata itself (we already have a ticket).

gusferguson:
@jburel @will-moore - I agree that the Bulk Annotation header adds nothing as it is a meaningless term. And it appears to be getting in the way of the layout too.

jburel:
The new feature will have to be tested with large tables since it will probably lead to the application to be frozen.
Problem was seen in insight when reading roi and analysis results.

will-moore:
Integration tests not run because "the training script failed"

https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration/
```
09:29 05:48:15 + python lib/python/omero/util/populate_metadata.py -k 1b23204c-788a-42ba-9c96-4a815dc114b6 Plate:219 bulk_annotation.csv
05:48:15 Traceback (most recent call last):
05:48:15   File "lib/python/omero/util/populate_metadata.py", line 33, in <module>
05:48:15     import omero.clients
05:48:15 ImportError: No module named omero.clients
```

sbesson:
@will-moore: the call above it likely failing because the PYTHONPATH is not set by the script

will-moore:
I will not try to fix https://trac.openmicroscopy.org/ome/ticket/12211 here since that is not related to this PR. The layout issue reported above is fixed.
I think we need some name for his section of the annotations, even if we don't use "Bulk Annotations". Maybe "Table Data"? Probably a tooltip should mention the name of the table file, ID and where it's attached?

joshmoore:
@will-moore : something like that, yes, please.

will-moore:
Renamed the title to "Tables" and added a tooltip on it (see screenshot)

screen shot 2015-06-15 at 12 27 12

joshmoore:
--rebased-to #3875

joshmoore:
```
./components/tools/OmeroWeb/omeroweb/webgateway/views.py:2320:80: E501 line too long (91 > 79 characters)
./components/tools/OmeroWeb/omeroweb/webgateway/views.py:2321:80: E501 line too long (97 > 79 characters)
```

components/tests/ui/robot_setup.sh
```diff
@@ -57,6 +68,13 @@ do
   bin/omero import -d $delDs $TINY_IMAGE_NAME --debug ERROR
 done
 
+# Import Plate
+bin/omero import $PLATE_NAME --debug ERROR > plate_import.log 2>&1
+plateid=$(sed -n -e 's/^Plate://p' plate_import.log)
+# Use populate_metadata to upload and attach bulk annotation csv
+PYTHONPATH=$PYTHONPATH:lib/python
+python lib/python/omero/util/populate_metadata.py -k $key Plate:$plateid $BULK_ANNOTATION_CSV
```
joshmoore: Without something like https://github.com/joshmoore/openmicroscopy/commit/f12a950990d0866bf0ce9393710a25377bcdc567, I'm thinking that this is causing the recent failures of merge-integration (like https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration/592/). If there are no objections, I'd suggest with close this in favor of gh-3875.

will-moore: If this is merged in to the Metadata branch instead of develop, we'll still have the same problem later when we try to merge Metadata back into develop. Does https://github.com/joshmoore/openmicroscopy/commit/f12a950990d0866bf0ce9393710a25377bcdc567 help? Can we get that into develop?

joshmoore: f12a950 should help, but that would require pulling in the rest of that branch. The point of the spaces is to prevent us from needing to prematurely merge everything. If all the test runs are transferred to the other build, then the theory is when we open metadata --> develop, things should remain green.

will-moore: OK, cool.

PR #3865, opened 2015-06-10
OmeroCpp: Decouple build from bioformats submodule

rleigh-dundee:
Primarily out of convenience, the OmeroCpp cmake logic was sourcing some files from the bioformats submodule.  To decouple, cease searching the bioformats cmake directory and copy the needed files into the OmeroCpp cmake directory.  The vast majority of this is third-party code (gtest and cmake modules)

--------

Testing:

- The OmeroCpp build will work as usual; there are no user-visible changes
- Remove `components/bioformats/cpp/cmake` and try a build; it should continue to work


joshmoore:
My primary concern with this strategy is that as we decouple more and more repositories, we'll need to reproduce this copy to any repo that intends to build C++. I'd *seriously* prefer we find another solution, though I can see that perhaps we need to do it one more time in the interest of getting decoupling rolling.

sbesson:
Also are the failing configurations of https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-cpp-cmake/ related to this change?

rleigh-dundee:
@sbesson Possibly, I'll check on this.

sbesson:
I globally share @joshmoore's concern. Reading http://www.kaizou.org/2014/11/gtest-cmake/, have you tested `cmake` `ExternalProject` as a way to resolve the `gtest` dependency without including it in the source code?

rleigh-dundee:
ExternalProject is an absolutely disgusting hack of last resort.  Embedding gtest in the source tree is bad enough; that's even worse.

rleigh-dundee:
I've pushed a fix to add some missing files.  Note these were missing:

```
components/tools/OmeroCpp/ext/gtest-1.7.0/include/gtest/internal/gtest-param-util-generated.h
components/tools/OmeroCpp/ext/gtest-1.7.0/include/gtest/internal/gtest-param-util-generated.h.pump
components/tools/OmeroCpp/ext/gtest-1.7.0/scripts/gtest-config.in
components/tools/OmeroCpp/ext/gtest-1.7.0/scripts/test/Makefile
```

Whatever rules we have in `.gitignore`, they are being a bit too overzealous here--none of these filenames should have been ignored.

rleigh-dundee:
Green (with develop/merge/cpp):

https://ci.openmicroscopy.org/job/OMERO-5.1-merge-cpp/441/console
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-cpp-cmake/342/console

joshmoore:
In discussing with @rleigh-dundee & @sbesson:

 * agreed that gtest shouldn't be embedded in each repo (this was likely a miscommunication when it was first added)
 * plan on the moving superbuild to a `cmake-ome` repository that will maintain the entire dependency graph for our C++ builds (like any upcoming `pom-ome`)
 * but for now, since these files were already in the repository (i.e. we're already bloated), ok to merge to keep decoupling moving forward.

PR #3867, opened 2015-06-11
Imagej fix

jburel:
Fix problem reported while presenting to local users
Check the following:

 * tick as default only the appropriate box in the Save ROIs... and Save Results... workflows (this means Save ROIs... would have the ROIs box ticked and Measurements unticked, and the Save Results... would have it the other way round
 * reformulate the ```image from all image windows``` to ```images from all image windows```


pwalczysko:

PR #3868, opened 2015-06-11
Include ome.model as part of the Javadocs

sbesson:
The ome.model package was referenced from the top-level Javadocs page but not
included as a fileset. This should allow the ome.model classes to be picked up
and published with the rest of the API. /cc @mtbc

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -650,6 +650,9 @@ To get started using Eclipse, execute "./build.py build-dev" and import the top-
                     </fileset>
                 </classpath>
 
+                <fileset dir="components/model/" defaultexcludes="yes">
+                    <patternset refid="all.java.files"/>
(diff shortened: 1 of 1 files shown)


sbesson:
See https://ci.openmicroscopy.org/job/OMERO-5.1-merge-build/ICE=3.5,jdk=8_LATEST,label=trout/javadoc/

mtbc:
Thank you, this is great, and includes all the classes I'd wanted. The build fix looks sensible. **Good to merge.**

Do we have to wait for 5.1.3 before they appear on http://downloads.openmicroscopy.org/ ?

sbesson:
Correct, the new version of the API Javadocs will have to be released with 5.1.3.

PR #3871, opened 2015-06-13
Create report of the dependencies bundled into the server

sbesson:
While working on the decoupling effort, we are reconsidering the use of transitive dependencies. This PR should allow to create an HTML report of the JARs bundled with the server (`lib/client` and `lib/server`) together with their provenance under `dist/lib`.

To test this PR, download the zip bundle of OMERO.server and check that `lib` contains two HTML reports: one for the server dependencies and one for the client dependencies.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -786,6 +786,7 @@ To get started using Eclipse, execute "./build.py build-dev" and import the top-
             <!-- sync="true" deletes any other files like services.jar or extensions.jar which may be under lib -->
             <ivy:resolve file="ivy.xml" type="jar,egg" conf="client" settingsRef="ivy.toplevel" log="quiet"/>
             <ivy:retrieve conf="client" pattern="${dist.dir}/lib/client/[artifact].[ext]" sync="false" log="quiet" settingsRef="ivy.toplevel"/>
+            <ivy:report conf="client" todir="${target.dir}" settingsRef="ivy.toplevel"/>
     </target>
 
     <target name="copy-server" depends="init">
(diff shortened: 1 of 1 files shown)


sbesson:
Discussing with @joshmoore @jburel @bramalingam, a proposal is to include the HTML versions of this Ivy report into the server itself (for reference/documentation). Any thought/veto? Is `lib` the right place?

bramalingam:
Just a suggestion : Is it worth organising these html files under a folder called "ivy-reports or dependency-tree-reports". As of now it looks orphaned?

bramalingam:
And the file names are misleading as well, without this organisation.

sbesson:
@bramalingam: Finding a proper location for these reports is definitely important. Arguably, they should not even live under `lib`. Is there a better placeholder e.g. a `share/` or `doc/` subfolder of the server? 

As for the name, we could either fix it via the build target or we could re-evaluate the name of the Ivy module name defined in https://github.com/openmicroscopy/openmicroscopy/blob/v5.1.2/ivy.xml#L2. 

/cc @hflynn @joshmoore

joshmoore:
I'm fine with any of the `lib` directories or a new one. Do we have anything else that would live under a `doc/` directory?

sbesson:
The server is currently shipping a couple of text files (`history.txt` and licenses) which could be part of this re-structuration.

joshmoore:
Leaving this open for a renaming then?
```
[jamoore@ome-c6100-3 lib]$ ls -1
client
fallback
insight
ivy-report.css
omero-main-client.html
omero-main-server.html
prefs.class
python
scripts
server
```

sbesson:
Happy to rename if someone has a concrete proposal.

joshmoore:
`dist/licenses/{server,client}/` ?

sbesson:
Leaving this open until the decoupling branch (which also has these commits) gets merged into the mainline. Then I'll close this (if it's not automatically done) and open a separate PR for the reorganization.

An hybrid proposal taking some of the suggestions above and following UNIX architecture used in other projects would be:

```
share/
share/licenses/*.txt
share/server/dependencies.txt
share/client/dependencies.txt
```

Thoughts?

hflynn:
Makes sense to me

bramalingam:
Definitely looks more organised.

sbesson:
Closing in favor of a new PR.

PR #3872, opened 2015-06-15
Map annotation tooltips

will-moore:
Fixes https://trac.openmicroscopy.org/ome/ticket/12774

Tooltip for map annotations is now displayed for the header.
Test that tooltip is displayed for map annotations, and still works for Tags & Files in right panel.

screen shot 2015-06-15 at 15 07 03


pwalczysko:
All works fine, except one issue:

Before the MAnnotation is created/populated, the tooltip still shows, but with blanks in the ``ID``, ``Owner``, ``Linked by`` etc. lines. This is a bit confusing. Probably if the MAnnot does not exist yet, do not show the tooltip at all.

screen shot 2015-06-16 at 12 03 03

pwalczysko:

PR #3873, opened 2015-06-16
rewrite the policy rules for Delete2 and Chgrp2

mtbc:
This PR rewrites how the `Delete2` and `Chgrp2` requests work to make them easier to adjust in the future. The rule interpretation is expanded to accommodate a further kind of rule multi-match. Additionally, the legal target object classes for each request are limited to a given set so, for instance, a detector may no longer be deleted directly, but probably nobody was doing that anyway. Work on `Chown2` and `Chmod2` will follow in later PRs. As a side effect, this fixes http://trac.openmicroscopy.org.uk/ome/ticket/12856.

Testing: This PR may have broken some corner case in deleting data and moving it to other groups. There should be no regressions introduced.


ximenesuk:
Could this now form the basis of revising the `--list` and `--list-details` options of `delete`, `chgrp`, etc.?

mtbc:
It certainly could. See the first list in each file in `components/blitz/resources/ome/services/graph-rules/`. Someday we could even add an API method allowing the client to ask the server directly.

mtbc:
In testing for regressions try to include data with rich object graphs (ROIs, annotations, etc.) from both data owner and other users. To generate different object graphs use functions like projection, or the dataset-to-plate script.

pwalczysko:
Problem, created https://trac.openmicroscopy.org/ome/ticket/12931#ticket - not there on latest, but present on merge. Please have a look.

joshmoore:
My two cents from a partial reading: the XML is **much** more comprehensible, but very not possible to compare before & after.

pwalczysko:
Tested:
 - bruker, Move and Delete fine
 - Dataset to Plate, Move and Delete fine
 - Projections, Move and Delete fine
 - Cut and Paste of MIF members in tree - problem which looks like regression observed https://github.com/openmicroscopy/openmicroscopy/pull/3873#issuecomment-112857226

Does this PR  warrant a full permissions testing round ? @mtbc @joshmoore

mtbc:
Shall we hold off on merging for a couple of days to give you more chance to figure out how to cause https://trac.openmicroscopy.org/ome/ticket/12931? I'll also have a try.

I think, this PR doesn't warrant a full permissions testing round before merging, but there will be following permissions PRs too, this is just the first. I would advise that we do plan a full permissions testing round for 5.1.3, but I would expect that the problems that we do find will be easy for me to fix.

mtbc:
This PR doesn't affect cut and paste so it seems to be working well. :smiley:

ximenesuk:
Re: https://github.com/openmicroscopy/openmicroscopy/pull/3873#issuecomment-112705734

FYI: in the current plugins the list commands use the server command: `omero.cmd.GraphSpecList()` Maybe a version 2 of this could be provided?

mtbc:
I've never known what that class did. Are there docs? Is this about listing the legal target objects?

ximenesuk:
It lists legal target objects with `--list` and the graphs for each object with `--list-details`. I don't know if there are docs but at the cli command parses the "spec" object in the response to get a list of target objects and to build a map of resulting modified objects. There may be something in the slice files.

mtbc:
I didn't see anything in the slice files, but good to know. `--list` should be quite feasible but I am not sure what still would make sense for `--list-details`.

ximenesuk:
I think having `--list` would be useful. At present both are effectively broken as they draw on an xml file that is no longer valid - in fact it may be one you have modified? The details option could be then removed.

mtbc:
A wholly different XML file now applies! https://github.com/openmicroscopy/openmicroscopy/pull/3873#issuecomment-112705734

mtbc:
@pwalczysko: Good to merge?

pwalczysko:
Good to merge.

PR #3866, opened 2015-06-10
making check_version configurable

aleksandra-tarkowska:
this allows disabling check_version

test locally

cc @will-moore

PR #3874, opened 2015-06-16
making check_version configurable (rebased onto metadata)

joshmoore:
This is the same as gh-3866 but rebased onto metadata.

----

this allows disabling check_version

test locally

cc @will-moore


joshmoore:
--rebased-from #3866

joshmoore:
This likely won't be merged into `metadata` since gh-3866 is going into the mainline, but leaving open for the moment, while we find a "mainline bump" strategy.

joshmoore:
Now available on the metadata branch with 62ee88c95e90eb94cb70440b486a5cdf044bb0a7

PR #3875, opened 2015-06-16
bulk annotations right panel (rebased onto metadata)

joshmoore:
This is the same as gh-3863 but rebased onto metadata.

----

See https://trello.com/c/MGeMSVXY/239-hdf-table-data-display-for-selected-wells-in-web

Load and display bulk annotations in metadata General tab.
Uses the same code as in the full image viewer.
Bulk annotations can be added as described: https://github.com/openmicroscopy/ome-internal/blob/master/testing_scenarios/BulkAnnotations.txt

NB: Bulk annotations are not filtered by owner as other annotations are. Is this needed?

screen shot 2015-06-09 at 14 15 18

### DIFF
--- a/components/tests/ui/resources/web/tree.txt
+++ b/components/tests/ui/resources/web/tree.txt
@@ -64,6 +64,14 @@ Select And Expand Dataset
     ${firstNodeId}    Select First Node By Type    dataset
     [Return]          ${firstNodeId}
 
+Select And Expand Plate
+    ${firstNodeId}    Select First Node By Type    plate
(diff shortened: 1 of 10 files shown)


joshmoore:
--rebased-from #3863

jburel:
flake8 error

joshmoore:
Testing against dogfish all shows this to be working well. The only issue brought up by @eleanorwilliams is that the number of empty columns can be overwhelming, which @will-moore will look at next. Merging to let that happen in a fresh PR.

PR #3876, opened 2015-06-16
Populate metadata

joshmoore:
This backports a number of PRs from @glencoesoftware (78, 89, 95) which fix minor issues in the populate_metadata.py script. To do so, it was necessary to revert @ximenesuk's flake8 changes, since those three PRs were made against dev_5_0.

Primary changes include:

 * Handle variable Image names.
 * Allow resolving the bulk annotations at image level.
 * Expand the robustness of populate metadata so that it is more inclined to complete successfully, warn more accurately and fail more gracefully when the source CSV is less than ideal. Also fixed issues with the loading of the metadata under cross-group conditions.


joshmoore:
Merged on the basis of the sysgro review (screen=151)

PR #3877, opened 2015-06-16
Postit panels scrolling

will-moore:
Fixes https://trac.openmicroscopy.org/ome/ticket/12894

To test, using **Chrome** browser, open an image that has a lot of ROIs (E.g. 30) in the web image viewer and "Show ROIs" so that the ROIs table is bigger than the 'postit' popup panel.
Use the mouse-wheel to scroll to the bottom of the table.
A scroll bar may or may not be shown initially but should appear while scrolling?


jburel:
Travis restarted

gusferguson:
@will-moore 

Tested with trout merge user-3 read-only-1 no image with 45 ROIs using Chrome, Firefox and Safari on Mac and Chrome, Firefox and IE on Windows.

Behaves as expected.
Good to merge.

PR #3880, opened 2015-06-18
Tagging fixes

dominikl:
- Enables Tagset selection on import stage, see [Ticket 12920](http://trac.openmicroscopy.org.uk/ome/ticket/12920); *Test: Create a Tagset, open the import dialog, and make sure you can see and select the tagset there.*
- Fixes a problem that tagged plates disappeared from the tree view after switching and returning to the Tag browser. *Test: Tag a plate; switch to the Tags browser; expand the tag the plate was tagged with and make sure the plate is displayed as a sub node of the tag node. Switch to another browser, Search, Screens, whatever. Return to the Tags browser and make sure it still looks like before (tag node expanded showing the plate).*
See [Trello - Tagging problems](https://trello.com/c/eo1Svfvp/30-tagging-problems)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/util/browser/TreeViewerTranslator.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/util/browser/TreeViewerTranslator.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.treemng.TreeViewerTranslator
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


joshmoore:
Kicked travis. See gh-3882.

pwalczysko:
- Tagset problem: Works fine.

- Plate in Tagtree problem : Principally all works fine, except ordering:

  - Before I navigate to another tab (like Search etc.), the Plates are ordered alphabetically with images, datasets etc. in between
  - After I navigate to another tab, the ordering is changed -> now the Plates are listed separately at the  end of the tree

Suggest: Unify the ordering, preferably the same way as it is in other tabs (==Plates separately, images separately, datasets separately.... etc.)

dominikl:
Apparently a small change, but it looks like nearly impossible to me at the moment. I'd suggest to postpone the ordering problem to another, separate PR.

pwalczysko:
In that case, happy to merge.

jburel:
Sorting the nodes can be done in another PR. Merging

PR #3881, opened 2015-06-18
Script ui tabbing

will-moore:
Fixes https://trac.openmicroscopy.org/ome/ticket/12080

To test:
 - Open a script from the scripts menu
 - On opening, the "focus" should be on the first input.
 - Tab key should move you through all the fields, going to "Run Script" *before* "Cancel".

### DIFF
--- a/components/blitz/src/omero/cmd/RequestObjectFactoryRegistry.java
+++ b/components/blitz/src/omero/cmd/RequestObjectFactoryRegistry.java
@@ -176,9 +176,10 @@ public class RequestObjectFactoryRegistry extends
                 new ObjectFactory(ChmodI.ice_staticId()) {
                     @Override
                     public Ice.Object create(String name) {
-                        if (graphRequestFactory.isGraphsWrap()) {
+                        // TODO: Chmod2 too slow to be automatically substituted for Chmod
+                        /* if (graphRequestFactory.isGraphsWrap()) {
                             return new ChmodFacadeI(graphRequestFactory);
-                        } else {
+                        } else */ {
                             return new ChmodI(ic,
                                     ctx.getBean("chmodStrategy", ChmodStrategy.class));
                         }
--- a/components/blitz/src/omero/model/ElectricPotentialI.java
+++ b/components/blitz/src/omero/model/ElectricPotentialI.java
@@ -777,7 +777,7 @@ public class ElectricPotentialI extends ElectricPotential implements ModelBased 
            setUnit(value.getUnit());
         } else {
             UnitsElectricPotential targetUnit = UnitsElectricPotential.valueOf(target);
-            Conversion conversion = conversions.get(value.getUnit()).get(target);
+            Conversion conversion = conversions.get(value.getUnit()).get(targetUnit);
             if (conversion == null) {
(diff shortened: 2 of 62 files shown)


gusferguson:
@will-moore 

Tested on trout merge using Mac Safari user-3 read-only-1

First field does not have focus on opening - need to tab once to get to it.
At end of all fields tab causes focus to disappear.
If tab again focus goes back to first field.

It looks as if this PR did not get into the build after all.

will-moore:
Yeah, it seems that the PR isn't in today. However, even if it IS working, I would still expect these to happen:
```
At end of all fields tab causes focus to disappear.
If tab again focus goes back to first field.
```

gusferguson:
@will-moore - the focus never moves on to the buttons.

gusferguson:
@will-moore 

Tested using trout merge user-3 read-only.

When script UI opens the focus is on the first text field - bypassing any drop-down controls above the text field. Tabbing through then works fine on the remaining text fields and drop-downs. At the end of the “in window” controls focus moves somewhere (does not seem to be “Cancel” then to the “Run Script” button. If tabbing continues then the focus moves to the first control in the window even if this is a drop-down.
I am struggling to work out what should be happening here.

will-moore:
Currently it's picking the first ```<input>``` element (not drop-down ```<select>```), which is practise is possibly more convenient since the "DataType" field is the first parameter in most scripts and should already be auto-populated. But, I guess I should fix it so that focus goes to the first input/select field in the form, since this is more understandable and consistent.
As long at focus goes to the "Run Script" before "Cancel" then this is working. That's what @joshmoore wanted I think?

gusferguson:
@will-moore - I also noticed that theRun Script button is not highlighting when it receives focus - which is a bit disconcerting as there is a gap when you are left feeling nothing is happening. Certainly when yu reach the end of the in-window controls, tab and hit return - nothing happens. I don't know how much of this is limited by the UI framework. If Josh's RTE is fulfilled then it may well just be time to move on.

will-moore:
@gusferguson For me the "Run Script" button does appear focussed after the last field (see screenshots from Batch Image Export).

screen shot 2015-06-22 at 21 53 12

screen shot 2015-06-22 at 21 52 56

will-moore:
Fixed so that the first field (including "select") is selected initially.

will-moore:
oops - Just did a push -f to the wrong branch.

joshmoore:
You should be able to force push back whatever was here before. (would also be good to *know* if that works... :smile:)

will-moore:
I did force push after closing this PR, but this just disabled the "Reopen" button.
Now re-opened at #3894.

joshmoore:
Good to know. Thanks.

gusferguson:
@will-moore 

Browser fun!

Tested using:
Mac:
Safari - does not tab to "Run Script" button or "Cancel" - just never get focus just a gap before focus returns to first element
Chrome - tabs to Cancel but focus skips Return
Firefox - works as expected

Windows:
IE 10, Firefox and Chrome all work as expected.

Windows all working - Firefox on Mac working - time to cut the losses and merge me thinks.

PR #3883, opened 2015-06-19
Versions cleanup

sbesson:
Following up the Ivy files cleanup while decoupling BF, this PR removes obsolete library versions from `etc/omero.properties` as well as corresponding JARs under `lib/repository` if applicable.

The only library which may cause problems is the `xalan` one which is used by the build system. However our dependency management should be able to retrieve it (from Central Maven).

/cc @bramalingam @jburel @joshmoore


joshmoore:
From https://ci.openmicroscopy.org/job/OMERO-DEV-merge-build/ICE=3.5,jdk=8_LATEST,label=octopus/ws/src/dist/lib/omero-main-server.html, where's the joda 2.0/2.2 conflict coming from? (Is that something for this PR or elsewhere?)

sbesson:
See https://trello.com/c/8ehtr2Df/28-review-server-client-dependencies for the discussion on the joda-time conflict.

joshmoore:
Building today worked locally once `components/bioformats` was moved out of the way (and **not** just renamed to, say, `components/bioformats-tmp`!) Code changes all make sense to me.

Further comments on gh-3913

sbesson:
Thanks. Note that https://github.com/openmicroscopy/openmicroscopy/pull/3883 should fix the Bio-Formats removal step by actively excluding it from the build list.

joshmoore:
:thumbsup:

PR #3885, opened 2015-06-19
fix #12932: improve error message for moving plates of projections

mtbc:
Fixes http://trac.openmicroscopy.org/ome/ticket/12932 by improving the error message to reference the culprit image and wellsample. This isn't for end users, but will be rather more helpful for developers. One of the workflows causing the error is http://trac.openmicroscopy.org/ome/ticket/12932#comment:4.


ximenesuk:
The error message using the workflow above is now sensible:

move-error

good to merge if @pwalczysko is okay with this message.

mtbc:
That's the kind of message I was hoping for at least; it's akin to the various other messages.

pwalczysko:
Sounds good to me

PR #3888, opened 2015-06-19
Script from search

jburel:
In this PR, few problems when running script from the search view (see http://trac.openmicroscopy.org/ome/ticket/12934)
to test this PR:

Test 1:
  * Log as user-2. Check that you are in private-1
  * Go to the search tab
  * Search for dv (few results are returned)
  * Select one image result from read-only group.
  * Click on the run script button in the toolbar. Select for example, channel offsets script
  * Check that the id field is populated with the id of the selected image.
  * Run the script.
  * Check that the script returns without error

Test 2:
   * Log as user-2. Check that you are in private-1
  * Go to the search tab
  * Search for dv (few results are returned)
  * Select one image result. The group context does not matter
  * Go to the right-hand pane and select one of the publishing script e.g. Thumbnail figure
  * Check that the id field is populated with the id of the selected image.
  * Check that the script is run
  * Check that the script returns without error

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
@@ -32,8 +32,6 @@ import java.awt.event.ActionListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseListener;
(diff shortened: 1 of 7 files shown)


jburel:
Restarted Travis

pwalczysko:
Works as expected.
test1 and 2 passed. Also independent test following the ticket description passed.

Ready to merge.

PR #3893, opened 2015-06-23
OMEROIJ_Bioformats_dependency_check

bramalingam:
ref : https://trac.openmicroscopy.org/ome/ticket/12925

Testing isntructions:

1) Add "OMEROIJ" plugin and bioformats_package.jar or loci-tools.jar to the plugins folder of imageJ.
2) open imageJ , "Plugins-->OMERO-->Connect to OMERO". This Should throw up the Insight login screen.

Now, rename the bioformats_package.jar (within the plugins folder) with something like,
bioformats_package-1.jar or biformats_package 1.jar or etc (concatenate any string to the name of the jar post bioformats_package or loci-tools with *.jar) and repeat the workflow specified above.

You should still get the Insight login screen.

Note : (when any string extension is added to bioformats_package or loci-tools) Without this PR the above workflow will throw a warning window  saying: can't find bioformats_package or loci-tools within the plugins folder and the login screen will not appear .

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/init/ContainerConfigInit.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/init/ContainerConfigInit.java
@@ -93,8 +93,13 @@ public final class ContainerConfigInit
 				value = values[j];
 				if (value != null) value = value.trim();
 				if (l == null) continue;
+				int idx = value.indexOf(".jar");
+				if (idx==-1){
+				    idx = value.length();
+				}
+				value = value.substring(0, idx);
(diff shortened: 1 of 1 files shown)


jburel:
It works.
It is probably better to use the method ```removeExtension``` from ```FileNameUtils```.

mtbc:
Does this need relisting?

jburel:
@mtbc: my bad, I forgot to review it while @bramalingam was away. No need to relist it.

jburel:
Retested. All working as expected after the suggested change. Merging

PR #3895, opened 2015-06-23
cmake: FindIce: Update for Ice 3.6.0 release

rleigh-dundee:
We already had support for the beta release in place; this removes it, leaving support for the final release only.  It also adds support for finding the various programs provided with Ice in addition to the slice generators; we don't use these at present.

See also: https://github.com/Kitware/CMake/pull/161

--------

Testing: Builds should continue to be green.  We don't use 3.6 at present, so this won't have any functional effect until Ice 3.6.0 is installed and detected.  If you wish to verify the logic is working, you can install the ZeroC rc2 candidate, but since it's just removing the beta version from the list, this is obviously correct.  Tested by me on Windows with Visual Studio 2012 and 2013 x86 and x64 (the supported platforms for 3.6.0).

### DIFF
--- a/components/tools/OmeroCpp/cmake/FindIce.cmake
+++ b/components/tools/OmeroCpp/cmake/FindIce.cmake
@@ -20,7 +20,7 @@
 #   Ice_SLICE_DIRS - the directories containing the Ice slice interface
 #                    definitions
 #
-# Ice programs are reported in::
+# Ice slice programs are reported in::
 #
 #   Ice_SLICE2CPP_EXECUTABLE - path to slice2cpp executable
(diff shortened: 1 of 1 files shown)


will-moore:
These builds are green:
 - https://ci.openmicroscopy.org/job/OMERO-5.1-merge-cpp/
 - https://ci.openmicroscopy.org/job/OMERO-5.1-merge-cpp-cmake/
 - https://ci.openmicroscopy.org/job/OMERO-5.1-merge-cpp-win/

Any others need checking or is this enough for "good to merge"?

rleigh-dundee:
@will-moore No, that's all of them, thanks.

PR #3898, opened 2015-06-24
PixelSize in px fix

dominikl:
If an image uses pixel as unit for the pixel size, Insight crashes because the transformation into microns is not possible. With this PR pixel sizes in pixel will be ignored and treated like no pixel size is provided.
`PixelsData.getPixelSizeX/Y/Z()' is the only place where I can think of this could be an issue.  Are there other places? @jburel

See [QA 11140](https://www.openmicroscopy.org/qa2/qa/feedback/11140)

### DIFF
--- a/components/blitz/src/pojos/PixelsData.java
+++ b/components/blitz/src/pojos/PixelsData.java
@@ -268,7 +268,7 @@ public class PixelsData extends DataObject {
      */
     public Length getPixelSizeX(UnitsLength unit) throws BigResult {
     	Length l = asPixels().getPhysicalSizeX();
-    	if(l == null)
+        if (l == null || l.getUnit().equals(UnitsLength.PIXEL))
     		return null;
     	return unit == null ? l : new LengthI(l, unit);
     }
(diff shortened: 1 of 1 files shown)


jburel:
Could ```l.getUnit()``` be ```null```?

dominikl:
I'm not really sure, but can't harm to check for null. If it's null it also has to be treated like PixelSize not available (i. e. return null).

mtbc:
I don't know if this is relevant but in the database the column pairs for dimensioned quantities like pixels.physicalsize{x,y,z} are constrained such that they have a unit if and only if they also have a value.

dominikl:
Ah thanks @mtbc ... Well, then the null check is actually superfluous.

dominikl:
Tested against the umassmed server, fixed the problem. The bug apparently could only be triggered with images imported previously to an older server version which was updated to 5.1.2, did not affect images directly imported to a 5.1.2 server.

jburel:
Thanks @dominikl. Merging

PR #3900, opened 2015-06-24
Fix for QA 11139

dominikl:
Fixes a potential NPE, see [QA 11139](https://www.openmicroscopy.org/qa2/qa/feedback/11139/)
I don't know how to trigger the bug, because navigation via arrow keys shouldn't be possible if nothing is selected in the central panel, so I think for testing it's sufficient to check if the arrow key navigation in central panel still works.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/browser/BrowserControl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/browser/BrowserControl.java
@@ -447,6 +447,9 @@ class BrowserControl
      */
     public void navigate(int direction, boolean multiSel) {
         ImageDisplay current = model.getLastSelectedDisplay();
+        if (current == null)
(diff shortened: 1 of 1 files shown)


mtbc:
Works fine, good to merge.

jburel:
@dominikl: as mentioned in chat, might be good to ask Alex to test against his server.

dominikl:
@jburel Didn't you mean test against their server because of  #3898 , not this PR?

jburel:
@dominikl: I mean. Point him to the build (indicating that the build can only be used for testing) so he can try against some images. Or he creates a tmp user for us with some images that failed.

jburel:
@will-moore @mtbc: the problem only occurs after an upgrade. 
The same image imported to a 5.1.2 server, no problem noticed.

jburel:
This has been tested against Alex's server with the original data post upgrade.
Merging

PR #3903, opened 2015-06-24
Slider move

jburel:
see https://trac.openmicroscopy.org.uk/ome/ticket/11632

To test:
* Select an image
* Go to preview
* Move the right-hand knob of the channel slider
* Move it next to the left knob. Keep dragging.
* Check that the right knob does not move back to the right


dominikl:

PR #3909, opened 2015-06-25
Fix for ticket 12816

dominikl:
Tiny fix to prevent a potential NPE, see [Ticket 12816](https://trac.openmicroscopy.org/ome/ticket/12816)
Unfortunately I couldn't reproduce the NPE, so code review must be sufficient and maybe check if you can still load big images (which triggers the affected loadTiles()` method).

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerComponent.java
@@ -3211,7 +3211,8 @@ class ImViewerComponent
 	 */
 	public void loadTiles(Rectangle region)
 	{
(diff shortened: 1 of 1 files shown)


mtbc:
Works fine, good to merge.

Though, one thought, belatedly also for #3900: if the null-handling occurs, would it be worth a log message?

joshmoore:
@dominikl : thoughts?

dominikl:
Sorry, this PR slipped out of my attention. Yes, good idea, worth writing something to the log, to investigate where this issue originates.

dominikl:
@joshmoore Thought about it again. This error just happened one time (just one QA report about this), not reproducible, probably some concurrency issue. I don't think it's worth spending more time on it. I'd suggest merge it as is.

joshmoore:
Fair enough.

PR #3912, opened 2015-06-29
Pixels to microns

will-moore:
See https://www.openmicroscopy.org/qa2/qa/feedback/11146/

This fixes display of pixel sizes for images that have pixel size units that are incompatible with microns.
E.g. in the example above, exception is thrown trying to convert from PIXEL to MICROMETER.

To test, set pixel sizes to E.g. PIXEL (complete code sample below) and try viewing in web:

screen shot 2015-06-29 at 11 39 00

```
$ omero shell --login

In [1]: from omero.gateway import BlitzGateway
In [2]: conn = BlitzGateway(client_obj=client)
In [3]: i = conn.getObject("Image", 8551)
In [4]: from omero.model.enums import UnitsLength
In [5]: u = omero.model.LengthI(9.8, UnitsLength.PIXEL)
In [6]: p = i.getPrimaryPixels()._obj
In [7]: p.setPhysicalSizeX(u)
In [8]: p.setPhysicalSizeY(u)
In [9]: conn.getUpdateService().saveObject(p)
```

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -2397,16 +2397,36 @@ class ImageWrapper (OmeroWebObjectWrapper,
         if 'link' in kwargs:
             self.link = 'link' in kwargs and kwargs['link'] or None
 
-    """
-    This override standard omero.gateway.ImageWrapper.getChannels
-    and catch exceptions.
-    """
+    def getPixelSizeUnits(self):
+        """
+        We use the pixelSizeX to find the appropriate units.
+        By default, we get the actual size in Microns, then use the
+        lengthunit filter to pick the most suitable size (same logic
+        is used for doing the length conversions).
+        However, if unit can't be converted, then we just return the
+        current unit's symbol.
+        """
+        from webgateway.templatetags.common_filters import lengthunit
+        try:
+            size = self.getPixelSizeX(units="MICROMETER")
+        except:
+            size = self.getPixelSizeX(True)
+            if size is not None:
+                return size.getSymbol()
+        if size is None:
+                size = 0
+        else:
+            size = size.getValue()
+        return lengthunit(size)
 
     def getPixelSizeXMicrons(self):
         """
         Helper for calling getPixelSizeX(units="MICROMETER") in templates
         """
-        size = self.getPixelSizeX(units="MICROMETER")
+        try:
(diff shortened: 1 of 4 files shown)


dominikl:
Works like expected, *good to merge*.
Thanks for providing an example on how to create this kind of images!

jburel:
The example should be moved to a test so we catch regression later on.

will-moore:
NB: Insight ticket for this issue: http://trac.openmicroscopy.org/ome/ticket/12942

will-moore:
@jburel Added test that runs above code and checks display of ``` Pixels Size (XYZ) (pixel): ``` in web.

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -2397,16 +2397,36 @@ def __prepare__(self, **kwargs):
...
+        We use the pixelSizeX to find the appropriate units.
+        By default, we get the actual size in Microns, then use the
+        lengthunit filter to pick the most suitable size (same logic
+        is used for doing the length conversions).
+        However, if unit can't be converted, then we just return the
+        current unit's symbol.
+        """
+        from webgateway.templatetags.common_filters import lengthunit
+        try:
+            size = self.getPixelSizeX(units="MICROMETER")
```
rleigh-dundee: Why are we defaulting to micrometres?  If the data is using another unit, why are we not defaulting to that and then finding a suitable size only if it's too far our for the value being used.  What happens for non-metric units e.g. inch?

Basically, I'm asking why we are imposing this policy in the client?  If the image was using microns, or was converted to microns for historical reasons, that will already be microns in the database, making the default here moot.  But it does mean for new files using non-micron units, we are overriding what should be the default display unit.

jburel:
Discussed with @rleigh-dundee, the reason behind of the conversion code.
The problem reported only occurred for images imported in a 4.4.x server and post upgrade.
Basically we have "units"= pixels.

will-moore:
@jburel You happy with this PR now that we have tests etc? If there's anything else needs doing, can do today but not tomorrow.

joshmoore:
There may be another round of changes, but this seems to be inline with what was done on the Java side. Merging and we can review when you're back, @will-moore.

jburel:
@will-moore: sorry i did not get notified of your comment. Changes match what has been done on Java side.
We will probably need to discuss UI cleaning later on so we do not need to convert.

PR #3913, opened 2015-06-29
Ivy resolvers cleanup

sbesson:
This PR implements the necessart logic to build the latest OMERO.server with the latest SNAPSHOT of Bio-Formats. Changes include:

- clean up of the Ivy resolvers to better highlight the default `omero-resolver` chain
- cleanup of the deprecated loci.resolver (which defaults to the OME artifactory)
- bump of the Bio-Formats version to 5.1.3-SNAPSHOT

To test this PR, check:
- the CI infrastructure `OMERO-DEV-merge-build` still builds and the server is correctly deployed
- the Bio-Formats JARs bundled with the server uses a SNAPSHOT jar instead of the release 5.1.2 JAR (e.g. test a bug which has been fixed in `develop` but is not in `v5.1.2`)


sbesson:
After cleaning stale artifacts from the Maven repository, the latest state of this PR should correctly deploy a server and build Insight using the current 5.1.3-SNAPSHOT of Bio-Formats, i.e. https://github.com/openmicroscopy/bioformats/commit/04e79dbc866e1b3c5e0af3d8fa6abc0ad17e316d.

While working on the last commits, I realized the strategy used in https://github.com/openmicroscopy/bioformats/pull/1838 might not work with the decoupling. Discussing with @joshmoore earlier today also in the wider context of decoupling/spaces/epics,
- artifacts are fully identified by their `groupId/name/version`
- the Maven version number scheme does not allow us to express our development semantics (typically `space` + `merge/latest/release` + major.minor version)
- once retrieved the artifact is cached under `.m2/repository/group/name/version` and the repository name is effectively lost. This can lead to stale artifacts if 2 repositories use the same group/name/version.
Assuming we want to preserver artifacts names, we may want to explore the usage of the `group` as a way to differentiate our builds e.g. for Bio-Formats

group  | snapshot | branch | type
---------|--------------|-----------|--------
 `ome` | no | core | release
 `ome`  | yes | core | latest
 `ome.merge`  | yes | core | merge
 `ome.space`  | no | `space` | release
 `ome.space`  | yes | `space` | latest
 `ome.space.merge`  | yes | `space` | merge

joshmoore:
As per https://github.com/openmicroscopy/openmicroscopy/pull/3883#issuecomment-117149979, this worked for me in a local build:
```
$ CLASSPATH=... jython
>>> import loci
>>> FT=loci.formats.FormatTools
>>> FT.VERSION
u'5.1.3-SNAPSHOT'
```

The only slight worry is that I wouldn't bump to SNAPSHOT in the mainline, but I assume that's not your intent anyway.

sbesson:
Happy to revert the `SNAPSHOT` commit. Generally, this forces us to think on the strategy for OMERO/Bio-Formats `release/latest/merge` builds. Bumping version numbers to SNAPSHOTs without modifying 'etc/omero.properties` should be doable:
- either via `local.properties` - however this is not trackable by Git
- or via `scc merge` e.g. by including a auto-generated branch bumping the BF version number. This would imply the snapshots would only be build in the merge OMERO build while latest OMERO builds would always have a release Bio-Formats version.

joshmoore:
:+1: for a scripted bump that devs could commit for opening a PR.

sbesson:
See https://github.com/openmicroscopy/openmicroscopy/pull/4024

PR #3914, opened 2015-06-29
plate run index bug

will-moore:
Fixes https://trello.com/c/JcQw9nOY/35-python-plate-acquisitions

It turns out that we were simply failing to pass the correct field index to the plate.html template.

To test, find a plate with multiple Acquisitions. For each, check that the right images are loaded into each field (compare with Insight).


manics:
Tested with a local build (connecting to dogfish), looks good. Note that clicking an image and choosing `Link to this well` gives you a well ID e.g. http://localhost:8080/toast/webclient/?show=well-4484, the plate run information is lost.

manics:
(I don't think it's a blocker for this PR since it's an improvement on the current state.)

joshmoore:
--rebased-to #3915

will-moore:
Is this good to merge? Anything else need doing? (I'm away after today).

manics:
Fine with me, I've added https://trello.com/c/w162WtbA/52-plate-acquisition-url-links

PR #3915, opened 2015-06-30
plate run index bug (rebased onto metadata)

joshmoore:
This is the same as gh-3914 but rebased onto metadata.

----

Fixes https://trello.com/c/JcQw9nOY/35-python-plate-acquisitions

It turns out that we were simply failing to pass the correct field index to the plate.html template.

To test, find a plate with multiple Acquisitions. For each, check that the right images are loaded into each field (compare with Insight).


joshmoore:
--rebased-from #3914

joshmoore:
Now available on the metadata branch with 62ee88c95e90eb94cb70440b486a5cdf044bb0a7

PR #3916, opened 2015-06-30
Exclude Bio-Formats from Ivy buildlist targets

sbesson:
If the Bio-Formats submodule is not removed from the main repository, this
should prevent the build system from entering components/bioformats.

/cc @rleigh-dundee @joshmoore


rleigh-dundee:
Works for me with and without components/bioformats being present.

joshmoore:
:thumbsup:

PR #3917, opened 2015-06-30
Shape Alpha

will-moore:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12394

Previously, the whole shape (stroke and fill) were getting set opacity set according to the line "alpha" in Insight. Now, we set the opacity of the fill and stroke independently.

To test:
 - In Insight, create a shape (E.g. rectangle) that has different opacity for fill and line.
 - Check display in web to see that opacity is same for stroke and line.

screen shot 2015-06-30 at 16 39 54


gusferguson:
@will-moore 

Tested using OMERO.insight-5.1.2-239-5398c69-ice35-b547-mac_Java7+ trout and trout-merge for web - user-3 - read-only-1

Set both line and fill to various opacities.
Behaves as expected - look the same in both viewers.
Good to merge.

safari

PR #3919, opened 2015-06-30
Add doc block for IQuery.projection

joshmoore:
IProjection is the only location which is currently
using the JSON-like serialization of Perms and Units.
It seems useful then to place the documentation for
these directly on the method.

See: https://trello.com/c/vdyTStEq/40-unit-documentation

### DIFF
--- a/components/blitz/resources/omero/api/IQuery.ice
+++ b/components/blitz/resources/omero/api/IQuery.ice
@@ -34,7 +34,34 @@ module omero {
                 idempotent omero::model::IObject findByQuery(string query, omero::sys::Parameters params) throws ServerError;
                 idempotent IObjectList           findAllByQuery(string query, omero::sys::Parameters params) throws ServerError;
                 idempotent IObjectList           findAllByFullText(string klass, string query, omero::sys::Parameters params) throws ServerError;
+
+                /**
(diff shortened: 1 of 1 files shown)


mtbc:
The text looks good and documentation is to be encouraged. :+1: One niggle is that I don't see the nice formatting in `dist/docs/api/slice2html/omero/api/IQuery.html`, it is instead all collapsed into a single paragraph and the separate HQL examples all run together.

joshmoore:
Pushed. Open to suggestions; this is readable in the HTML now, but less so in the .ice. :/

mtbc:
Looks good in the HTML indeed; this PR is **good to merge**.

If you prefer, could revert back to something more like your original, with: add "(a)", "(b)", bullets, punctuation to turn that first `<ul>` list into something that reads okay when concatenated, and have the only actual markup be the `<pre>` block; or, some other tradeoff among the formats. Up to you.

joshmoore:
Hmmm....perhaps that means we're either making a decision about how we all want to add slice doc blocks, or we're punting and we'll clean this up later. I'm open either way. cc: @hflynn @sbesson

sbesson:
Got thumbs up from @hflynn. No strong feeling from my side on the formatting policy. Merging. We can always unify later.

PR #3920, opened 2015-06-30
Export to Excel

jburel:
See https://trac.openmicroscopy.org/ome/ticket/12952

To test:
 * Open an image e.g. dv and draw a roi. Select the shape
 * Go to the Results tab
 * Export the results to excel
 * Check that the values are correctly displayed in the excel sheet.

cc @pwalczysko

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
@@ -258,7 +258,8 @@ public class MeasurementTableModel extends AbstractTableModel
                     }
                     
                     String s = l.getSymbol();
-                    if (!symbols.contains(s)) {
+                    if (!symbols.contains(s) &&
+                            !l.getUnit().equals(UnitsLength.PIXEL)) {
                         symbols.add(s);
                     }
                 }
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
@@ -154,9 +154,7 @@ public class MeasurementTableModel extends AbstractTableModel
             }
             return buffer.toString();
         }
-        System.err.println(value);
         if (showUnits) {
-            //System.err.println(value);
             if (value instanceof Length) {
                 MeasurementUnits units = getUnitsType();
                 Length n = (Length) value;
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
@@ -263,6 +263,13 @@ public class MeasurementTableModel extends AbstractTableModel
                             !l.getUnit().equals(UnitsLength.PIXEL)) {
                         symbols.add(s);
                     }
+                } else if (v instanceof Number) {
+                    if (k.equals(AnnotationKeys.ANGLE.getKey())) {
+                        String s = "("+UIUtilities.DEGREE_SYMBOL+")";
+                        if (!symbols.contains(s)) {
+                            symbols.add(s);
+                        }
+                    }
                 }
             }
             if (symbols.size() == 1) {
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
@@ -154,14 +154,16 @@ public class MeasurementTableModel extends AbstractTableModel
             }
             return buffer.toString();
         }
+        System.err.println(value);
         if (showUnits) {
+            //System.err.println(value);
             if (value instanceof Length) {
                 MeasurementUnits units = getUnitsType();
                 Length n = (Length) value;
                 String s;
+                KeyDescription key = getColumnNames().get(col);
+                String k = key.getKey();
                 if (!units.getUnit().equals(UnitsLength.PIXEL)) {
-                    KeyDescription key = getColumnNames().get(col);
-                    String k = key.getKey();
                     if (unitsDisplay.size() > col && unitsDisplay.get(col)) {
                         s = UIUtilities.formatValue(n,
                                 AnnotationKeys.AREA.getKey().equals(k));
@@ -172,10 +174,9 @@ public class MeasurementTableModel extends AbstractTableModel
                     if (CommonsLangUtils.isNotBlank(s))
                        return s;
                 } else {
-                    s = UIUtilities.twoDecimalPlaces(n.getValue());
-                    if (CommonsLangUtils.isNotBlank(s)) {
-                        return s;
-                    }
+                    s = UIUtilities.formatValueNoUnit(n,
+                            AnnotationKeys.AREA.getKey().equals(k));
+                    return s;
                 }
             }
         }
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
@@ -131,7 +131,7 @@ public class MeasurementTableModel extends AbstractTableModel
         if (row < 0 || row > values.size()) return null;
         MeasurementObject rowData = values.get(row);
         Object value = rowData.getElement(col);
-        if (value instanceof List) {
+        if (value instanceof List && !showUnits) {
             List<Object> l = (List<Object>) value;
             
             if (l.size() == 1) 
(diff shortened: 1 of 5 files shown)


pwalczysko:
Works as designed, but:



1. The usecase in question (and I suppose many other usecases as well) need the value in the Excel-sheet cell to be recognized by Excel as number. This is because the immediate follow-up evaluation of those numbers in Excel is usually asked for. Suggestion as discussed with @jburel : scan the column and if all the values in the column are the same unit (which would be in the majority of cases I would guess), display the unit in the header of the column. Only when the units in the particular rows are mixed, display the units inside the cells.

2. The Z and T columns cells have after the number of the plane a period (a dot). This seems like an unwanted artifact ? (not a big deal as the numbers are still recognized as numbers by Excel)

screen shot 2015-07-01 at 11 52 37

jburel:
@pwalczysko:
Fixed problem 1, I will not touch the second point in this PR since it has been like that for a long time
This can maybe be done in another PR

components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
```diff
@@ -177,4 +222,58 @@ public boolean isCellEditable(int row, int col)
...
+                this.columnNames.size());
+        Iterator<KeyDescription> kk = this.columnNames.iterator();
+        while (kk.hasNext()) {
+            key = kk.next();
+            list.add(new KeyDescription(key.getKey(), key.getDescription()));
+        }
+        MeasurementTableModel model = new MeasurementTableModel(
+                list, this.unitsType);
+        model.values.addAll(this.values);
+        //Check column names de
```
mtbc: truncated?

gusferguson:
@jburel 

Tested using Tested using OMERO.insight-5.1.2-308-a751719-ice35-b550-mac_Java7+ user-3 read-only-1

Works as expected with just rectangular ROIs on image, but when other ROIs are added, it breaks down and all units revert to being displayed in the cells - see three exports in screenshots - 1 - just rectangles, 2 - line added, 3 - eclipse added.

If just an ellipse is drawn, the units are moved to column headers as expected - screenshot 4 - but if just a line is drawn, they are not and you get the (0.0 PIXEL) screwiness - screenshot 5.

A couple of other points 
- the units in the columns for line are really screwed up with (0.0 PIXEL)
- there are no units for "Angle" - just inconsistent 

microsoft excel

microsoft excel

microsoft excel

microsoft excel

microsoft excel

jburel:
@gusferguson: thanks. I will investigate the problem

jburel:
@gusferguson: problems should now been fixed.

gusferguson:
@jburel 

Tested using OMERO.insight-5.1.2-444-5cfcf30-ice34-b555-mac_Java6 trout user-3 read-only-1

When adding ROIs one by one, all units appearing as expected in column headers for all types of ROIs in all orders of adding - see screenshot 1.

When exporting results for an existing test image that has a large number of ROIs - 45 - ID 26654, one of the columns - Centre (Y) persists with units in each cell rather than the header - screenshot 2. 

Despite drawing up to 60 ROIs on an image and saving and closing and re-opening, I could not replicate this problem with newly drawn ROIs - screenshot 3.
Could it be that the test image ROIs that showed this were drawn by script and this somehow upset it?

The only other really minor issue is that if an angle of a line is a whole number of degrees, the value in the column is showing with the decimal point but no 0 following it e.g. "45." - see screenshot 3. When I look at  the format of the cell in excel it is showing as custom "#.##" - not as a number - so that is probably why. This is minor.

Anyway - from the testing using newly drawn ROIs I think this is good to merge with or without a fix to the degree decimal point zero issue.

microsoft excel

15-07-0910 xls

15-07-09g xls

pwalczysko:
I do think that this PR is improving things a lot and is almost there now.  But, really sorry to interfere, I know that the workflow of the user in question is asking for the numbers in the excel file produced by omero to be 
    - recognised as Numbers in the Excel program.  

This is usually indicated in Excel program by having the strings in the cells so that they bind to the right (plain text binds to the left). When viewing the screenshots of Gus, we have the numbers bound to the left (all of them with exception of angles). 

I retested with a line and found that indeed the cells with line lengths cannot be for example summed up using an excel function, so they are not recognized as numbers, except again, weirdly, the angles (see below). Is there a possibility to alleviate this ? If not, I would agree to merge this and go ahead as it is, definitely a massive improvement for the user.

screen shot 2015-07-09 at 10 15 50

gusferguson:
@pwalczysko - thanks for picking that up Petr. If the RFE was specific about needing numbers then it would be expected to get the formatting of the columns in the spreadsheet correct. I am not sure how this is achieved, but hopefully there is a way.

pwalczysko:
After discussion with @gusferguson we concluded that if the number-recognising-problem in excel cannot be alleviated, maybe we should be exporting a csv file from omero (which can then be easily imported to excel in 1 step, and the numbers will be recognised as numbers) ?

jburel:
@pwalczysko, @gusferguson: Thanks for report. I would rather stick to what we have i.e. save to Excel and add (in another PR) save as cvs
since saving to Excel has been there for a very long time

jburel:
@pwalczysko, @gusferguson: The problem should now been fixed.
There is still a very minor issue when the number is an int and there is a dot after it e.g. 2.
This has been since the beginning and does not block anything (noticed by @pwalczysko earlier)

Could you review it tomorrow?

gusferguson:
@jburel @pwalczysko 

Tested using OMERO.insight-5.1.3-m1-193-bd2d3a0-ice35-b556-mac_Java6 trout user-3 read-only-1

Using image ID 23401 and others

- have lost units in the results pane of the ROI tool for all except angle of line in newly drawn ROIs
- ditto for units in spreadsheet - screenshot 1
- units retained and displayed as before in previously drawn ROIs e.g. image ID 3891 - both ROI tool and spreadsheets (exported fresh today) - screenshots 3 and 4
- the numbers in the Excell spreadsheet are now behaving as numbers and can be added etc.
- the cell format is still reading custom #.## and the toolbar formatting a numbers - screenshot 2
- the non-display of 0 after decimal point in integer values is present in all columns, not just angles, but that is not an issue as J-M describes above.

All shown in screenshots below:

Newly drawn ROIs:

microsoft excel

Cell formatting:

microsoft excel

Previously drawn ROIs:

test 15-07-10 08-12-28 xls

test 15-07-10 08-08-58 xls

gusferguson:
@jburel - note that this test was done on the original merge build of the morning - will recheck after second later build, but thoughts are this testing should be valid.

gusferguson:
@jburel @pwalczysko 

Retested with OMERO.insight-5.1.3-m1-193-12fd548-ice35-b557-mac_Java6 after trout rebuild.

Working fine now for freshly drawn ROIs - ID 9233:
- units showing in both ROI tool and in column header in exported spreadsheet as expected
- numbers adding up fine in spreadsheet
- formatting and no trailing 0 after decimal point as before

Previously drawn ROIs remain as before.

So above effects were the result of some build mess-up.

From my POV this is now good to merge.

Petr - will you double check.

microsoft excel

gusferguson:
Tested with OMERO.insight-5.1.3-m1-193-12fd548-ice35-b557-win

Works as above with Windows 8.

jburel:
@gusferguson: If the unit is not the same for each cell of a given column, the symbol is displayed in the cell itself. That's the outcome of a discussion I had with @pwalczysko.

pwalczysko:
Tested on Mac 10.10 and Win7. Works as expected. The summing is fine, numbers are recognized as numbers. Ready to merge.

@jburel : (definitely do not want to hinder merging this, just a clarifying question). Understand that Gus got the different units in the same column in https://github.com/openmicroscopy/openmicroscopy/pull/3920#issuecomment-120268854 (in one case he has nm, all other cases microns). Not disputing the tactics and agree with @jburel that this is what we agreed upon, i.e. to display the units in cells in such case. Question : what are the cases in which such a situation occurs ? ( = how did Gus get the differing units display in the same column ?)

gusferguson:
@pwalczysko @jburel - the differing units displayed in the same column were from previously drawn ROIs - image ID 26654 - don't know if that makes a difference.

jburel:
@gusferguson: which user?

gusferguson:
@jburel - user-3 read-only-1

jburel:
@gusferguson, @pwalczysko: fixed the "." issue when exporting results. 
This will be included in 15/07 build

jburel:
@gusferguson: the case nm displayed in one of the cells is b/c of the conversion logic in place when displaying the most suitable units for a given value. Code prior to the introduction of units.
At this stage, I really do not want to change it. 
Another option will be to use the "most used unit" in a given column (in your case microns) but that will imply changing all the logic in place to move the units in the header.
If you are okay with that, we can revisit post 5.1.3

gusferguson:
@jburel - I won't be able to test the build tomorrow, but I am sure @pwalczysko will be able to.
On the other issue with the units - I think it would be fine to wait.

pwalczysko:
@jburel @gusferguson Agree, no big changes needed, I will just re-test tomorrow and we can merge as it is, I think it is fully fit for purpose as it stays now. Thanks for the clarification.

jburel:
@pwalczysko: are you happy with the latest change?

pwalczysko:
@jburel sorry, forgot about this , retesting now.

pwalczysko:
Test passed. The dot in the B and C columns is gone as expected. Shown also the overall behaviour to @mporter-gre and he okayed it.

PR #3921, opened 2015-07-01
Allow to retrieve bundle in Ivy (eclipse project classpath)

dominikl:
Equivalent to 210b26f290f0 just for the tryEclipse task (which generates the class path for the eclipse projects).

Test: Run `./build.py clean build-dev`, import the projects in Eclipse (or check `.classpath` within the project directories), make sure the class paths are set up correctly (in particular the presence of guava.jar, kryo.jar and metrics-logback.jar for Insight and Blitz project).


joshmoore:
Makes sense looking at it. Guess the question is: are there any others of these looming?

dominikl:
That's probably a question for @sbesson , the build system is still a bit of a mystery for me.

mtbc:
In the new Eclipse project classes like `EventLogQueue`, `OMEXMLModelComparator`, `BasicSecuritySystem`, `HierarchyDeleteTest`, etc. (basically, classes from various modules that use the third-party JARs at issue) all look good, and the build worked fine from a clean tree. :+1: from my point of view!

sbesson:
Thanks @mtbc. Merging this. Future problems arising for Eclipse users can be fixed in dedicated PRs.

PR #3924, opened 2015-07-02
Move filter condition static string to function

jballanc:
This removes the static strings in `AllGroupsSecurityFilter` and
`OneGroupSecurityFilter` that define the filter condition and replaces
them with `static` methods of the same name. The only place this string
was being used in either class was in `getDefaultFilterCondition`, so
that reference has been changed to a call to the new method.

### DIFF
--- a/components/server/src/ome/security/basic/AllGroupsSecurityFilter.java
+++ b/components/server/src/ome/security/basic/AllGroupsSecurityFilter.java
@@ -62,18 +62,6 @@ public class AllGroupsSecurityFilter extends AbstractSecurityFilter {
 
     static public final String filterName = "securityFilter";
 
-    private static String myFilterCondition = String.format(
-                  "\n( "
-                + "\n  1 = :is_share OR "
-                + "\n  1 = :is_admin OR "
-                + "\n  (group_id in (:leader_of_groups)) OR "
-                + "\n  (owner_id = :current_user AND %s) OR " // 1st arg U
-                + "\n  (group_id in (:member_of_groups) AND %s) OR " // 2nd arg G
-                + "\n  (%s) " // 3rd arg W
-                + "\n)"
-                + "\n", isGranted(USER, READ), isGranted(GROUP, READ),
-                isGranted(WORLD, READ));
-
     final SqlAction sql;
 
     /**
@@ -96,8 +84,22 @@ public class AllGroupsSecurityFilter extends AbstractSecurityFilter {
         this.sql = sql;
     }
 
+    private static String myFilterCondition() {
+        return String.format(
+                "\n( "
+                + "\n  1 = :is_share OR "
+                + "\n  1 = :is_admin OR "
+                + "\n  (group_id in (:leader_of_groups)) OR "
+                + "\n  (owner_id = :current_user AND %s) OR " // 1st arg U
+                + "\n  (group_id in (:member_of_groups) AND %s) OR " // 2nd arg G
+                + "\n  (%s) " // 3rd arg W
+                + "\n)"
+                + "\n", isGranted(USER, READ), isGranted(GROUP, READ),
+                isGranted(WORLD, READ));
+    }
+
     public String getDefaultCondition() {
-        return String.format(myFilterCondition, roles.getUserGroupId());
+        return String.format(myFilterCondition(), roles.getUserGroupId());
     }
 
     public Map<String, String> getParameterTypes() {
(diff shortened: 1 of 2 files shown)


mtbc:
I'm curious about the motivation; some classfile optimization issue that affects even private statics?

jballanc:
Because of optimizations, the `static String` is inlined into the body of `defaultFilterCondition`, such that providing a different version of the filter string in a subclass has no affect unless the subclass also overrides `defaultFilterCondition`.

mtbc:
How can the subclass override `defaultFilterCondition` when it is private? (Ah, this is some inner class thing?)

mtbc:
I still don't understand; the static string is private so the subclass shouldn't be able to provide a different version of it. Even if inlined into `getDefaultCondition()`, when you override that method you provide a different method body anyway that can't reference `myFilterCondition`? If you happen to have some code on some branch that depends on this change and makes everything clear I'd love to see it.

(Sorry for prior version of this comment; I'm just feeling confused!)

mtbc:
Ah! Does the subclass' class file contain a copy of the bytecode of the superclass' method (with the inlined string) even if the subclass' code oughtn't technically be able to see that string? So that's the problem this fixes?

jballanc:
The issue is that if the subclass doesn't re-implement `getDefaultCondition()`, then calling that method bubbles up to the original class where the method *does* still contain the `private static String`. As a simple example of what I mean:

```java
// foo/Bar.java
package foo;

public class Bar {
    private static String msg = "Secret Bar message.";

    public void announce() {
        System.out.println(msg);
    }
}
```
```java
// foo/Baz.java
package foo;

public class Baz extends Bar {
    private static String msg = "Secret Baz message.";
}
```
```java
// Foo.java
import foo.Bar;
import foo.Baz;

public class Foo {
    static final public void main(String[] args) {
        System.out.println("From Bar...");
        Bar bar = new Bar();
        bar.announce();
        System.out.println("From Baz...");
        Baz baz = new Baz();
        baz.announce();
    }
}
```
```bash
> javac foo/Bar.java foo/Baz.java Foo.java
> java Foo
From Bar...
Secret Bar message.
From Baz...
Secret Bar message.
```

mtbc:
Thank you for the minimal example! But, if I do what your PR seems to,
```java
// foo/Bar.java
package foo;

public class Bar {
    private static String msg() { return "Secret Bar message."; }

    public void announce() {
        System.out.println(msg());
    }
}
```
```java
// foo/Baz.java
package foo;

public class Baz extends Bar {
    private static String msg() { return "Secret Baz message."; }
}
```
```java
// Foo.java
import foo.Bar;
import foo.Baz;

public class Foo {
    static final public void main(String[] args) {
        System.out.println("From Bar...");
        Bar bar = new Bar();
        bar.announce();
        System.out.println("From Baz...");
        Baz baz = new Baz();
        baz.announce();
    }
}
```
```
> javac foo/Bar.java foo/Baz.java Foo.java
> java Foo
From Bar...
Secret Bar message.
From Baz...
Secret Bar message.
```
... it doesn't seem to help. I appreciate your continued patience with me!

jballanc:
Heh, I actually just noticed that...on second inspection, I think you're right @mtbc . Just changing the string to a method seems insufficient. Back to the drawing board I suppose...

mtbc:
`protected` is probably your friend. :smiley:

jballanc:
Just tried `protected` and even that's insufficient. I think there's an issue with calling `static` from non-`static` scope.

mtbc:
Yeah, has to be non-static for any useful overriding.

mtbc:
Looking far more plausible, great.

mtbc:
How frequently will the method be called? With the current change, we go from constructing the string once to constructing it every time.

jballanc:
The call to `getDefaultCondition()` only happens at instantiation, which I think shouldn't be that frequent. Besides, if my understanding of JVM optimizations is correct the string should be cached at the call-site, making it functionally equivalent to the static string case (once optimizations are applied).

mtbc:
I'd be surprised if `AllGroupsSecurityFilter.getDefaultCondition()` had any useful string caching for the result of the `String.format` in `myFilterCondition()` but if infrequent then it's not a big deal.

dominikl:
From the code perspective this looks fine to me. Unfortunately I don't know how to test it. Is it covered by the integration tests? If it is, good to merge from my point of view.

jballanc:
I suppose so long as the integration tests still work, that's confirmation of the default case. Using the same patch we've adjusted the test timeout on a private branch and I can confirm that the tests do, indeed, run for longer with it set. Would that be sufficient?

jballanc:
If the server is still properly restricting group-level access, then that would imply this is working as expected. We've got a private branch with a different filter string (for a different DB) in a subclass, and it is working as expected there as well.

dominikl:
Ok, then I think it's good to merge.

PR #3925, opened 2015-07-02
Make test timeout configurable

jballanc:
This adds the ability to specify `omero.test.timeout` in
`local.properties`. This value will be used in `AbstractServerTest` for
the `scalingFactor` that is, in turn, used to determine the timeout for
`CallbackCmdI`-based tests.

### DIFF
--- a/components/tools/OmeroJava/test/integration/AbstractServerTest.java
+++ b/components/tools/OmeroJava/test/integration/AbstractServerTest.java
@@ -12,6 +12,7 @@ import static org.testng.AssertJUnit.assertNull;
 import static org.testng.AssertJUnit.assertTrue;
 import static org.testng.AssertJUnit.fail;
 
+import java.lang.Long;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -162,7 +163,7 @@ public class AbstractServerTest extends AbstractTest {
     public String GUEST_GROUP = "guest";
 
     /** Scaling factor used for CmdCallbackI loop timings. */
-    protected long scalingFactor = 500;
+    protected long scalingFactor;
 
     /** The client object, this is the entry point to the Server. */
     protected omero.client client;
@@ -246,6 +247,7 @@ public class AbstractServerTest extends AbstractTest {
         root = newRootOmeroClient();
         tmp.__del__();
 
+        scalingFactor = new Long(System.getProperty("omero.test.timeout"));
         final EventContext ctx = newUserAndGroup("rw----");
         this.userFsDir = ctx.userName + "_" + ctx.userId + FsFile.separatorChar;
         SimpleBackOff backOff = new SimpleBackOff();
--- a/components/tools/OmeroJava/test/integration/AbstractServerTest.java
+++ b/components/tools/OmeroJava/test/integration/AbstractServerTest.java
@@ -246,7 +246,14 @@ public class AbstractServerTest extends AbstractTest {
         root = newRootOmeroClient();
         tmp.__del__();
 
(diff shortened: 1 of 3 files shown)


components/tools/OmeroJava/test/integration/AbstractServerTest.java
```diff
@@ -12,6 +12,7 @@
 import static org.testng.AssertJUnit.assertTrue;
 import static org.testng.AssertJUnit.fail;
 
+import java.lang.Long;
```
mtbc: Surely unnecessary?

etc/local.properties.example
```diff
@@ -72,6 +72,10 @@ testng.useDefaultListeners=false
 # etc/ivyconf.xml for more information.
 loci.resolver=omero-resolver
 
+# Controls how long CmdCallbackI based tests will wait
+# to finish.
```
mtbc: Mention the units. (Seconds?)

dominikl:
Looks good to me. Apart from the point that the scalingFactor actually isn't really the timeout (timeout = nLoops * scalingFactor, where nLoops differs), but I also don't have a better term than *timeout* for this parameter.
Will check again on Monday when this PR will be included in the merge build.

etc/local.properties.example
jballanc: Milliseconds would be closest, though since this really only sets the scaling factor, the exact timeout is dependent on each test (though in the most common case it would be milliseconds).

components/tools/OmeroJava/test/integration/AbstractServerTest.java
```diff
@@ -246,6 +246,7 @@ protected void setUp() throws Exception {
         root = newRootOmeroClient();
         tmp.__del__();
 
+        scalingFactor = new Long(System.getProperty("omero.test.timeout"));
```
dominikl: Is it worth catching the case that omero.test.timeout property is not set (would throw a NPE at this point)?

dominikl:
Apart from my comment regarding a potential NPE this looks fine to me. CI Integration tests also passed. So if there aren't thoughts about the potential NPE, this is good to merge.

jballanc:
Since this code only affects tests, and any NPE is going to result in a very quick test run (if you know what I mean :wink:), I'm not *too* concerned. Still, I've added a basic catch-all that should help if someone mis-configures things.

dominikl:
Yes, you're right, probably wouldn't have been necessary. Anyway, good to merge, when your travis build passes. Can you restart it or @joshmoore ? Apparently I can do it.

joshmoore:
Kicked travis due to "exited with 137"

joshmoore:
Kicked again.

joshmoore:
(Discussing with @sbesson, this will conflict with decoupling. Best to get it in ASAP and we can adjust later. For example, when @jballanc looks into adding the same timeout to the Python stack)

components/tools/OmeroJava/test/integration/AbstractServerTest.java
```diff
@@ -246,6 +246,14 @@ protected void setUp() throws Exception {
         root = newRootOmeroClient();
         tmp.__del__();
 
+        try {
+            scalingFactor = new Long(System.getProperty("omero.test.timeout"));
+        } catch {
```
sbesson: Note this is failing Travis with

```
[nomemoryjava] warning: [options] bootstrap class path not set in conjunction with -source 1.6
[nomemoryjava] /home/travis/build/openmicroscopy/openmicroscopy/components/tools/OmeroJava/test/integration/AbstractServerTest.java:255: error: '(' expected
[nomemoryjava]         } catch {
```

PR #3926, opened 2015-07-02
View well

jburel:
Keep origin of the image i.e. if the image is going from Well sample
This will allow the loading of ROI measurements when the mt is opened. cc @joshmoore 
To test;
 * Make sure you can open a standard image (right-click ,toolbar, double click, right-hand panel: general/preview)
* Make sure you can open an image from well sample (right-click , double click, right-hand panel)

To test the roi measurements section, the annotation will either have to be linked to plate cc @manics  or @joshmoore will open a PR to adjust the query in ROIService

@manics has linked the file to the plate
Looking closely at the query, the table also needs to be linked to the roi in order to work. This could potentially imply a lot of links
@manics is going to add the links to the roi so we can test it. This will not be part of the IDR import script

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/ViewAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/ViewAction.java
@@ -25,9 +25,11 @@ package org.openmicroscopy.shoola.agents.dataBrowser.actions;
 
 //Java imports
 import java.awt.event.ActionEvent;
+
 import javax.swing.Action;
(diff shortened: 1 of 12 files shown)


manics:
All ROIs in http://dogfish.openmicroscopy.org/webclient/?show=plate-151 should now be linked to the table

jburel:
Still no luck, no measurement returns when using ```getRoiMeasurements```

joshmoore:
--rebased-to #3938

sbesson:
Excluding in order to test the Java gateway work.

PR #3929, opened 2015-07-02
Temporary exclude ExporterTest from the OmeroJava integration suite

sbesson:
With the new build system, ExporterTest are failing due to a deployment bug for the  5.1.2 specificationjJar leading to skipped tests for the full integration  suite. This commit temporarily disables the ExporterTest to review/merge the decoupling effort (BF, Java) into the mainline.

In the mid-term, the specification is expected to re-include the XML transform and these tests should be reenabled for the 5.1.3 release of OMERO.

See also https://github.com/openmicroscopy/bioformats/pull/1836

/cc @jburel @dominikl @melissalinkert


sbesson:
With this commit https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/ is now passing except for `integration.gateway` tests.

jburel:
I will look at the tests post 5.1.3 to see what we need to move to B-F and what we need to keep.
Looks good

PR #3930, opened 2015-07-02
Load format

jburel:
Load the format when loading the image


dominikl:
Integration tests failed, image.getFormat() is null.

jburel:
I think I will need to add a new test since this image generated in the test probably does not have the format set.

mtbc:
It might be worth adding a setFormat to ModelMockFactory.simpleImage.

jburel:
This image now has the format set.
The tests now pass
```
===============================================
Ant suite
Total tests run: 41, Failures: 0, Skips: 0
===============================================
```

dominikl:
Looks fine, tests passed, too. Good to merge.

PR #3932, opened 2015-07-03
Handle new ROI Points string format

manics:
There are (at least) 2 formats for storing Points in ROI Polygons. The webgateway only had one of them.

https://trello.com/c/hJLlf1Xr/47-error-on-preview-rois

Are there any other formats?

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/ROI_utils.py
+++ b/components/tools/OmeroPy/src/omero/util/ROI_utils.py
@@ -51,6 +51,54 @@ from omero.model import PolygonI
 from omero.model import MaskI
 from omero.rtypes import rdouble, rint, rstring
 
+#
+# HELPERS
+#
+
+
+def pointsStringToXYlist(string):
+    """
+    Method for converting the string returned from
+    omero.model.ShapeI.getPoints() into list of (x,y) points.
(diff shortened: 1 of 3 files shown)


manics:
--rebased-to #3933

joshmoore:
Implementation looks good. Up for cherry-picking https://github.com/joshmoore/openmicroscopy/commit/ac9fea252e9ad5abed60408d6f79781b0fc786e1 ?

As for whether or not there are other legacy formats, I don't know.

manics:
Done

jburel:
There is also some legacy "mask" points see https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/pojos/PolygonData.java

manics:
Looking at the original https://github.com/openmicroscopy/openmicroscopy/blob/673d1c2f11e1cdfc2c28936a0376682b94bdb7c8/components/tools/OmeroWeb/omeroweb/webgateway/views.py#L460

Only the first set of points is taken, so mask would never be used. Is that correct?

joshmoore:
Talking to @jburel, the `masks[]` is legacy. Merging.

manics:
--no-rebase

manics:
--rebased-from #3933

PR #3937, opened 2015-07-06
allow group owners to chown

mtbc:
This PR expands `Chown2` such that it allows a non-admin to chown somebody else's data if they are an owner of the group in which the data is. Perhaps most easily tested from the CLI.

https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-java/lastCompletedBuild/testngreports/integration.chown/ should continue to pass.


mtbc:
8 more passing integration tests this morning. :smiley:

dominikl:
Tested with CLI, works like expected, good to merge.

PR #3938, opened 2015-07-07
View well (rebased onto metadata)

joshmoore:
This is the same as gh-3926 but rebased onto metadata.

----

Keep origin of the image i.e. if the image is going from Well sample
This will allow the loading of ROI measurements when the mt is opened. cc @joshmoore 
To test;
 * Make sure you can open a standard image (right-click ,toolbar, double click, right-hand panel: general/preview)
* Make sure you can open an image from well sample (right-click , double click, right-hand panel)

To test the roi measurements section, the annotation will either have to be linked to plate cc @manics  or @joshmoore will open a PR to adjust the query in ROIService

@manics has linked the file to the plate
Looking closely at the query, the table also needs to be linked to the roi in order to work. This could potentially imply a lot of links
@manics is going to add the links to the roi so we can test it. This will not be part of the IDR import script

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/ViewAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/ViewAction.java
@@ -25,9 +25,11 @@ package org.openmicroscopy.shoola.agents.dataBrowser.actions;
 
 //Java imports
 import java.awt.event.ActionEvent;
+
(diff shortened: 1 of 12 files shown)


joshmoore:
--rebased-from #3926

joshmoore:
Talking to @jburel, closing since the upcoming demos (m2) will be done via web. Merge builds with #3926 can still be used against `idr-demo`.

PR #3939, opened 2015-07-07
Plate acquisition date

manics:
https://trello.com/c/oPlGJKqQ/60-plateacquisitions-in-web-should-show-starttime
Show `PlateAcquisition.startDate` if it's set, otherwise fallback to the previous behaviour.

To test: find a PlateAcquisition with a startTime set. OMERO.web should show this time as the Creation date instead of the import time.

PR #3940, opened 2015-07-07
Plate acquisition date (rebased onto develop)

manics:
This is the same as gh-3939 but rebased onto develop.

----
https://trello.com/c/oPlGJKqQ/60-plateacquisitions-in-web-should-show-starttime
Show `PlateAcquisition.startDate` if it's set, otherwise fallback to the previous behaviour.

To test: find a PlateAcquisition with a startTime set. OMERO.web should show this time as the Creation date instead of the import time.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5706,7 +5706,8 @@ class _PlateAcquisitionWrapper (BlitzObjectWrapper):
             if t > 0:
                 return datetime.fromtimestamp(t / 1000)
         except:
-            return super(_PlateAcquisitionWrapper, self).getDate()
+            pass
+        return super(_PlateAcquisitionWrapper, self).getDate()
 
     def getName(self):
         name = super(_PlateAcquisitionWrapper, self).getName()
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5700,6 +5700,14 @@ class _PlateAcquisitionWrapper (BlitzObjectWrapper):
     def __bstrap__(self):
         self.OMERO_CLASS = 'PlateAcquisition'
 
+    def getDate(self):
+        try:
+            t = self._obj.startTime.val
+            if t > 0:
+                return datetime.fromtimestamp(t / 1000)
+        except:
+            return super(_PlateAcquisitionWrapper, self).getDate()
+
     def getName(self):
         name = super(_PlateAcquisitionWrapper, self).getName()
         if name is None:
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_plate_wrapper.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_plate_wrapper.py
@@ -87,7 +87,8 @@ class TestPlateAcquistionWrapper(object):
 
         plateacq = update_service.saveAndReturnObject(plateacq)
         plate_id = plateacq.plate.id.val
-        return gw.getObject('Plate', plate_id)
+        plate = gw.getObject('Plate', plate_id)
 
         acq = list(plate.listPlateAcquisitions())[0]
-        assert acq.getStartTime() > 0
+        assert acq.getStartTime() is None
+        assert acq.getDate() > datetime.fromtimestamp(0)
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_plate_wrapper.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_plate_wrapper.py
@@ -6,16 +6,17 @@
 # Use is subject to license terms supplied in LICENSE.txt
 
 """
-   gateway tests - Testing the gateway plate wrapper
+   gateway tests - Testing the gateway plate and plate acquisition wrappers
 
    pytest fixtures used as defined in conftest.py:
    - gatewaywrapper
 
 """
 
+from datetime import datetime
 import pytest
 
-from omero.model import PlateI, WellI, WellSampleI, ImageI
+from omero.model import PlateAcquisitionI, PlateI, WellI, WellSampleI, ImageI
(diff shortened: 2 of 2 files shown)


manics:
--rebased-from #3939

manics:
Just realised there's a mistake in the integration test, now fixed. The UI testing should be unaffected though.

pwalczysko:
The PR does what is described in the header.
Two questions though:

  - why is the Plate still with the Creation time which matches the Import time ? This is a bit confusing, and suggests that the Plate, unlike the Run, was created at Import time.
  - it seems that the Acquisition time from the wells is gone now in the UI (trout merge web client connected to dogfish) -> it used to be there though, compare the same plate on trout latest viewed by webclient trout latest

With this PR on a selected well:

screen shot 2015-07-08 at 14 01 19

Without this PR the same well selected:

screen shot 2015-07-08 at 14 00 41

pwalczysko:
I seem to remember that the wording ``acquistion date``, ``creation date`` and ``import date`` had some meaning in the web UI, which is now a bit compromised -> ``creation date`` being actually reserved for items which were created at import - am I right @will-moore ?

manics:
> why is the Plate still with the Creation time which matches the Import time ? This is a bit confusing, and suggests that the Plate, unlike the Run, was created at Import time.

I haven't actually touched the UI, just the underlying PythonGateway wrapper object. I agree the Plate time not matching the PlateAcquisition time is confusing, but I don't know what a good alternative is. Plates don't have an explicit datetime field.

> it seems that the Acquisition time from the wells is gone now in the UI (trout merge web client connected to dogfish) -> it used to be there though, compare the same plate on trout latest viewed by webclient trout latest

Could that be due to differences in the plate or importer? If you connect to trout:latest from the trout/merge webclient it shows the same?

> I seem to remember that the wording acquistion date, creation date and import date had some meaning in the web UI, which is now a bit compromised -> creation date being actually reserved for items which were created at import

No idea I'm afraid.

pwalczysko:
After a discussion with @manics it seems that we need some advice, mainly about:

1. What is the role of the Creation time ? What should it actually represent (creation time of the object, but pre-import or post-import) ? What is its place next to Import time ? Is it something only for Proj/Datasets/Screens (Plates ?) How should it be read from the back end ? How to unify things here ?
2. How to solve the confusion between the times of Plate Run and Plate ? 


Opinions please ?
@joshmoore @will-moore @aleksandra-tarkowska @gusferguson @jburel

jburel:
Please check the strategy in ```https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/pojos/PlateAcquisitionData.java``` to display "Label" when no name set (```getLabel``` method)

jburel:
The "plate object" will only an internal creation time. The information will be held at the plate acquisition if I remember correctly.
For historical reason we have "enforced" a plate run even if there is only one.

manics:
It looks like Insight doesn't display the `PlateAcquisition.startTime` anywhere:
screen shot 2015-07-13 at 19 46 18
Should I just close this PR?

pwalczysko:
It seems to me that htis PR rather exposes a weakness in the clients regarding the times than doing something wrong. On the other hand, with obvious discrepancy between Plate Run and Plate creation time and possibly not being able to fix the bigger issue quickly, we would confuse the user a lot.

manics:
Well, I'm confused :). Do you want to do some UX scoping in which case I can either close or put this on-hold.

pwalczysko:
@manics : with no JM and no Will atm we probably want to put it on hold

will-moore:
Creation Date is the same as "Import Date" in fact I think we might refer to it as "Import Date" in some places in the clients (E.g. search by "Import Date"). Acquisition Date is a time-stamp on the Image. So, for a Run, this could display the startTime as the Acquisition Date (not Creation Date).

manics:
What should `_PlateAcquisitionWrapper.getDate()` return? `startDate`? Or keep the current `creation date` behaviour and add a new method?

joshmoore:
Closing as per the card: https://trello.com/c/oPlGJKqQ/60-plateacquisitions-in-web-should-show-starttime

joshmoore:
@jburel : is this something you would consider for 5.2.2 or should I rebase to `metadata52`?

jburel:
@joshmoore: I will go over the old notes about "dates"

jburel:
going back to this PR:
 * ```Acquisition date```: information usually stored in the file. i.e. microscope acquisition
 * ```Creation Date``` is used for non-image object. i.e. when the object was created in the DB. The term ```Import Date``` is used when dealing with images since it made more sense for users
 * Plate Run label: the name should be used if set, otherwise the start-end value i.e. when the run started and when it ended are used as a label. Web only have a method called "name"/label might be preferable.
Web only currently checked if start not null and end not null
cc @will-moore

manics:
There's still the question of which of those dates should `_PlateAcquisitionWrapper.getDate()` return?

jburel:
I would rather have 2 methods;
* One getAcquisitionDate method (currently getDate returns that for image):
  For image it will return the value, and for plate run the Start value (if set)
* One getDate or to be more explicit getCreationDate

will-moore:
@jburel - Sorry, I don't understand your ```Plate Run label``` paragraph. What label is this? The name used in the tree & top of right panel? This PR doesn't concern that label, does it?

@manics ```ImageWrapper.getDate()``` tries to return Acquisition Date and if that fails, then returns Creation Date. We also have ```creationEventDate() and getAcquisitionDate()``` which are a bit clearer. So I guess you should try to mimmic the same with PlateAcquisitionWrapper.

jburel:
@will-moore: 
I was indicating the method used in the pojos i.e. we have a getLabel method that uses the name if set otherwise set a "label" using the start-end values if set otherwise we use the id of the run.
This method could be used as a ref to improve support in the wrapper

jburel:
note that the getAcquisition in the pojo is not supported for Plate Run, so it will have to be added

will-moore:
I'll look at simply adding Start and End fields to the right-hand pane for Run

PR #3948, opened 2015-07-09
Add gitversion.xml to all build.xml import blocks

joshmoore:
Without gitversion.xml imported, any component-level
build like ant -f components/tools/OmeroJava/build.xml
will fail with a version of UNKNOWN. Top-level builds
do not show this behavior since gitversion.xml is imported.

Testing:
 * download the appropriate src build (openmicroscopy*.zip)
 * run `./build.py build-dev`
 * run `./build.py -f components/tools/OmeroJava/build.xml -Dtestng.useDefaultListeners=true -Dtestreports.dir=target/reports/unit test`

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -52,9 +52,7 @@ To get started using Eclipse, execute "./build.py build-dev" and import the top-
     </description>
 
     <property name="import.dir" value="${basedir}/components/antlib/resources"/>
-    <import file="${import.dir}/gitversion.xml" optional="true"/>
     <import file="${import.dir}/global.xml"/>
-    <import file="${import.dir}/version.xml"/>
 
     <target name="init" depends="check-ivy,check-scons,check-ice">
         <mkdir dir="${blitz.comp}/generated"/>
--- a/components/antlib/resources/global.xml
+++ b/components/antlib/resources/global.xml
@@ -25,6 +25,9 @@
 
 -->
 
+        <!-- Before global.xml -->
+        <import file="gitversion.xml" optional="true"/>
+
         <tstamp>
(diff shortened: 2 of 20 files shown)


joshmoore:
--rebased-to #3949

rleigh-dundee:
Looks fine to me.  But I wonder if it would make sense to include gitversion.xml and version.xml in global.xml to save the need to manually include everywhere.  Unless there's a reason why this doesn't make sense I can't recall offhand.

joshmoore:
I don't know/remember one, @rleigh-dundee. If you don't either, than perhaps it was just an oversight on our part. Happy to give it a try.

rleigh-dundee:
I do recall there might be some reason why we didn't originally do that, but I can't remember any specifics unfortunately.  I think you would have to try and see.

sbesson:
One thing that comes to my mind is that importing `version.xml` globally may have been conflicting with the Bio-Formats versioning logic. As we decoupled the Bio-Formats build, we may have removed this limitation as well.

joshmoore:
Pushed. Easier to check with `w=1`: https://github.com/openmicroscopy/openmicroscopy/pull/3948/files?diff=unified&w=1

cc: @rleigh-dundee @sbesson

joshmoore:
Restarted job due to HDF5 unit test. Not in today's build.

sbesson:
Testedfrom the merge source zip by running the following commands:

- `./build.py build-dev` -> pass
- `./build.py -f components/<component>/build.xml` for each component of the server -> pass
- `./build.py -f components/tools/OmeroJava/build.xml -Dtestng.useDefaultListeners=true -Dtestreports.dir=target/reports/unit test` -> pass

PR #3949, opened 2015-07-09
Add gitversion.xml to all build.xml import blocks (rebased onto metadata)

joshmoore:
This is the same as gh-3948 but rebased onto metadata.

----

Without gitversion.xml imported, any component-level
build like ant -f components/tools/OmeroJava/build.xml
will fail with a version of UNKNOWN. Top-level builds
do not show this behavior since gitversion.xml is imported.

Testing:
 * download the appropriate src build (openmicroscopy*.zip)
 * run `./build.py build-dev`
 * run `./build.py -f components/tools/OmeroJava/build.xml -Dtestng.useDefaultListeners=true -Dtestreports.dir=target/reports/unit test`

### DIFF
--- a/components/blitz/build.xml
+++ b/components/blitz/build.xml
@@ -2,6 +2,7 @@
 <project name="blitz" default="install" basedir=".">
 
     <property name="import.dir" value="${basedir}/../antlib/resources"/>
+    <import file="${import.dir}/gitversion.xml" optional="true"/>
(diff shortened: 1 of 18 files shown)


joshmoore:
--rebased-from #3948

joshmoore:
No longer needed with 5.1.3 merged into `metadata`

PR #3950, opened 2015-07-09
Make rstring more tolerant of non-`str`

joshmoore:
Rather than have consumers start using the essentially
hidden and more lenient RStringI implementation, this
makes the `omero.rtypes.rstring` wrapper as lenient of
non-`str` types, including `unicode`.

See: https://trello.com/c/hjCA90az/18-rfe-rtype-methods-should-auto-convert

Testing:

 * unit tests should suffice to show whether or no this will cause any problems since the method is now **more** lenient.
 * However, discussion is needed on whether the change in behavior (i.e. no longer throwing a `ValueError` on non-`str` arguments) is acceptable for 5.1.x.

cc: @will-moore @sbesson @chris-allan @aleksandra-tarkowska @ximenesuk

### DIFF
--- a/components/tools/OmeroPy/src/omero/rtypes.py
+++ b/components/tools/OmeroPy/src/omero/rtypes.py
@@ -56,7 +56,7 @@ def rtype(val):
         return rlong(val)
     elif isinstance(val, float):
         return rfloat(val)
-    elif isinstance(val, str):
+    elif isinstance(val, StringTypes):
         return rstring(val)
     elif isinstance(val, omero.model.IObject):
         return robject(val)
--- a/components/tools/OmeroPy/src/omero/rtypes.py
+++ b/components/tools/OmeroPy/src/omero/rtypes.py
@@ -28,6 +28,8 @@ IceImport.load("omero_RTypes_ice")
 IceImport.load("omero_Scripts_ice")
 IceImport.load("omero_model_RTypes_ice")
 
+from types import StringTypes
+
 
 def rtype(val):
     """
@@ -294,12 +296,13 @@ def rstring(val):
         return remptystr
     elif isinstance(val, omero.RString):
         return val
-    elif isinstance(val, str):
+    elif isinstance(val, StringTypes):
         if len(val) == 0:
             return remptystr
         else:
             return RStringI(val)
-    raise ValueError("Not string type: %s" % type(val))
+    else:
+        return RStringI(str(val))
 
 # Static factory methods (collections)
 # =========================================================================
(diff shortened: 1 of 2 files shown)


chris-allan:
This is a good start.

However, I think the problem from a usability perspective is that in Python you often do not know whether you have an `str` or `unicode` instance. This PR does not address that problem and hides potential encoding errors from a user of `rstring()`.

As an encoded string is really the only form of string that Ice supports I feel like `rstring(u'some unicode string')` should perform `.encode('utf-8')` for the user of that convenience function. This support should also be extended to the `rtype()` convenience function. Otherwise every API consumer is going to end up implementing an `rstring()` or `rtype()` wrapper that handles the encoding.

joshmoore:
Pushed

chris-allan:
Looks great.

components/tools/OmeroPy/src/omero/rtypes.py
```diff
@@ -294,12 +296,15 @@ def rstring(val):
         return remptystr
     elif isinstance(val, omero.RString):
         return val
-    elif isinstance(val, str):
+    elif isinstance(val, StringTypes):
```
aleksandra-tarkowska: I am afraid this will never let you to test ```unicode```. I think you want ```basestring``` that is a base class for ```str``` and ```unicode```.```StringType``` is ```str```.

components/tools/OmeroPy/src/omero/rtypes.py
```diff
@@ -294,12 +296,15 @@ def rstring(val):
...
         return val
-    elif isinstance(val, str):
+    elif isinstance(val, StringTypes):
         if len(val) == 0:
             return remptystr
         else:
+            if isinstance(val, unicode):
+                val = val.encode("utf-8")
             return RStringI(val)
-    raise ValueError("Not string type: %s" % type(val))
```
aleksandra-tarkowska: I do not see any test for unicode, then you would get ValueError

components/tools/OmeroPy/src/omero/rtypes.py
joshmoore:
```
>>> import types
>>> types.StringTypes
(<type 'str'>, <type 'unicode'>)
>>> isinstance(u'test', types.StringTypes)
True
```
?

aleksandra-tarkowska: Which python?

aleksandra-tarkowska: Sorry is StringTypes missed the s in the end. All fine

sbesson:
After discussing with @joshmoore, merging for 5.1.3-m2

PR #3954, opened 2015-07-10
OmeroIJ: Switch dimensions and save image

bramalingam:
ref: https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7843

Testing Instructions:

1) Open OMERO Image (Multi-dimensional image) in ImageJ, and Click on Image-->Properties (in the ImageJ Menu).
2) Switch the Z and T values and click OK. This should switch the image properties in the ImageJ Image Window.
3) Plugins->OMERO->Save Image(s) to OMERO

This should save the modified image back to OMERO as an ome-tiff.

Without this PR, "add to queue"  in the import dialog will be active but not functional. (i.e you won't be allowed to add the modified image back to the import dialog).

For more details please read the thread (above).

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/FileObject.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/FileObject.java
@@ -307,7 +307,7 @@ public class FileObject
                 try {
                     sw.close();
                 } catch (IOException e) {
-                    IJ.log("I/O Excception:" + e.getMessage());
+                    IJ.log("I/O Exception:" + e.getMessage());
                 }
                 pw.close();
             }
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/FileObject.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/FileObject.java
@@ -25,16 +25,29 @@ import ij.IJ;
 import ij.ImagePlus;
 import ij.io.FileInfo;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringReader;
+import java.io.StringWriter;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
 
 import loci.formats.codec.CompressionType;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.openmicroscopy.shoola.util.CommonsLangUtils;
+import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
 
 /**
  * Object hosting the information about the "file" to import.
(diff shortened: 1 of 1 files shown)


jburel:
@bramalingam: you need to close the input stream in a try/catch/finally block

jburel:
you can initiliaze a variable with ```nodeList.item(x).getAttributes() ``` instead of 3 calls

pwalczysko:
The PR from the UI perspective does what it set out to do. The new image is uploaded with no problem.

There is an ImageJ workflow issue : If the image is true mult-t and multi-z (not just 1-t multi-z or mulit-t 1-z), then the swapping of the numbers in the Image > Properties in ImageJ does not deliver the expected transposition of z and t dimensions, instead IJ apparently just takes the images on "first come first served" basis and puts them into the stack. @graemeball was contacted - this is purely an IJ issue and thus this PR from the UI perspective is fine.
I suppose the code comments of @jburel should be worked upon and then after quick retest we can go ahead here.

pwalczysko:
@bramalingam is working on a Macro which should solve the IJ workflow issue. Good data to test this new macro would be user-4 on trout merge ID 31611

bramalingam:
Macro to resolve the imagej issue for multidimensional images (in case of dimension switching) : 

```
title = getTitle();
getDimensions(width, height, ch, z, t);
options = "  title=[Concatenated Stacks]";
for (i=1; i<=z; i++) {
slice = d2s(i,0);
run("Make Substack...", "channels=1-" + ch + " slices=&slice frames=1-" + t);
run("Properties...", "channels=&ch slices=&t frames=1");
selectWindow(title);
options = options + " image" + i + "=[" + title + "-" + i + "]";
}
close();
run("Concatenate...", options);
run("Properties...", "channels=&ch slices=&t frames=&z");
```

pwalczysko:
Tested with the macro above. Works as expected. We could maybe comment on the forum post to recommend this macro for mulit-t and multi-z images dimension swaps ? Or store it somewhere ? @hflynn @jburel @joshmoore https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7843

jburel:
We have few code snippets like this one to test PR or on forum.
we discussed having a "prototype"/example repo.
@pwalczysko: I think @hflynn has already an example card (sorry for not listing it here) where we should start listing those PRs so when the time comes we can easily find the code sample

hflynn:
https://trello.com/c/XullyCzo/75-examples and the linked gdoc - still in a work in progress, will finish it off this week.

pwalczysko:
@hflynn : reading https://docs.google.com/document/d/161D9_JpI8FPd6rjiRzJxWZrBBYJ-QHgWbmrBugwb9qw/edit not sure where this fits, it is a bit "external" but useful. In the meantime, could we bookmark this PR somewhere so that Balaji's work does not get lost ?

hflynn:
Yeah, i don't know either! There are a bunch of example cards but https://trello.com/c/pd5ckwuZ/14-examples-files-followup seems to be the right place for this, I'll add it.

pwalczysko:
Even after the last code cleaning, this works all as expected.

I guess there is one more code remark of @jburel to address @bramalingam https://github.com/openmicroscopy/openmicroscopy/pull/3954#commitcomment-12138500

bramalingam:
@jburel : I did try to log all kinds of errors. If it is not very helpful I could replace that error with : IJ.log("I/O exception");

mtbc:
Maybe still at least include something from the error itself, even just its message, so we know what kind of I/O error: very handy when people paste the log output back at us in QA!

jburel:
typo ``` Excception```

bramalingam:
@jburel , @pwalczysko : Does this PR need another Re-test?

pwalczysko:
@bramalingam : I think so, please relist.

pwalczysko:
Simple swapping of z and t when only 1 z or 1 t and using Image > Properties work as expected. Retested also the macro. Still all as expected. Ready to merge.

PR #3956, opened 2015-07-10
DSL build dependencies cleanup

sbesson:
This PR:

- adds some option to generate Ivy reports for the various configuration under `target/build.deps`
- reviews the build dependencies for the `dsl` component, using the Maven dependencies rather than `lib/repository` and cleaning duplicate/conflicting dependencies


joshmoore:
Looks fine. What's the logic for which jars are still under lib/repository, e.g., `lib/repository/checkstyle-4.3.jar`? Is it just "we'll get them all eventually"?

sbesson:
Correct. Goal is to reduce the number of JARs to the bare minimum. I am going through them as we go but I do not have the knowledge of when/what all of them were introduced for.

PR #3961, opened 2015-07-13
do not WARN when omero.test.timeout default is not overridden

mtbc:
Prevent the WARN message when Java integration tests are run with `omero.test.timeout` taking its default value. In testing, check that setting or not setting `omero.test.timeout` still works as one would expect and still WARNs of silly values like `foo`.

cc: @jballanc


dominikl:
Looks fine. Good to merge.

PR #3964, opened 2015-07-14
Reconnection fix (decoupling)

dominikl:
This fixes the outstanding reconnection issue, see [Trello Test Checklist](https://trello.com/c/2WY5iyS7/8-java-gateway)


dominikl:
The last commit should fix the Treeviewer freeze after connection loss/reconnect. Although I couldn't figure out, why the Treeviewer on develop (without the decoupling) PR usually does not freeze; sometimes it also freezes especially in debug mode (which indicates some weird timing/threading issue).

PR #3967, opened 2015-07-15
Change signature to `protected` for helper methods

jballanc:
The helper methods for `EventListenersFactoryBean` are called from
`protected` methods, but are themselves declared `private`. This limits
the ability of any subclass to override the `protected` methods in any
meaningful way. Changing the scope of these methods to `protected` will
fix this.

### DIFF
--- a/components/server/src/ome/security/basic/EventListenersFactoryBean.java
+++ b/components/server/src/ome/security/basic/EventListenersFactoryBean.java
@@ -221,14 +221,14 @@ public class EventListenersFactoryBean extends AbstractFactoryBean {
     // =========================================================================
 
     /** calls override for each key */
-    private void override(String[] keys, Object object) {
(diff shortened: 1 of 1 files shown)


jballanc:
Nothing specific to test other than to make sure everything still builds and runs. I have a private branch where I've confirmed that these changes are sufficient to allow for useful subclassing of the `EventListenersFactoryBean`.

mtbc:
This is fine and is already **good to merge**.

Since elevation to `protected` is offering a kind of API to other classes, even just subclasses, this might be a good opportunity to fix/complete the Javadoc for those methods, if convenient.

joshmoore:
Considering how horribly under-documented this section of the code-base is, I'd say we're probably ok as is. Merging for now, with strong hopes of one day doing a proper javadoc cleanup.

PR #3968, opened 2015-07-16
Fixed 'Empty' and 'Loading...' nodes

dominikl:
Fixes a bug introduced by #3927 ; instead showing the nodes "Loading..." and "Empty" a node with the same name like the dataset was displayed.

Test: Expand a dataset in treeviewer, make sure it shows a childnode named "Loading..." while loading the dataset content. Expand an empty dataset, make sure it shows one childnode named "Empty"


aleksandra-tarkowska:
Works fine, although it looks like that if you open few datasets all EMPTY labels are greyed out before content is loaded. Is that expected?

dominikl:
Thanks @aleksandra-tarkowska . I just checked again with an older build: This was also the case before PR #3927 , just never noticed it. So I won't fix this here.

PR #3970, opened 2015-07-16
Modify IceSSL Ciphers to ALL with exceptions

ximenesuk:
See https://trello.com/c/vvivb0o7/567-java-8u51-connection-problems for full discussion.

This PR changes two of the instances of `IceSSL.Ciphers` relevant to Java to `ALL`. This may have a security impact on other versions of Java.

Tesing: Insight running on machines with Java 8_51 should be able to connect to OMERO servers. The CLI importer on machines with Java 8_51 should work.

--rebased-to #3971

Added: only the single client change may be required but it seemed sensible to change both uses. It might also be worth reviewing the C++ and Python config, should that mirror the Java settings?

### DIFF
--- a/components/blitz/src/omero/client.java
+++ b/components/blitz/src/omero/client.java
@@ -319,7 +319,7 @@ public class client {
         id.properties.setProperty("Ice.Default.PreferSecure", "1");
         id.properties.setProperty("Ice.Plugin.IceSSL", "IceSSL.PluginFactory");
         id.properties.setProperty("IceSSL.Protocols", "tls1");
-        id.properties.setProperty("IceSSL.Ciphers", "NONE (DH_anon)");
+        id.properties.setProperty("IceSSL.Ciphers", "ALL");
         id.properties.setProperty("IceSSL.VerifyPeer", "0");
 
         // Setting default block size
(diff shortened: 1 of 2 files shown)


mtbc:
Also discussion at https://trello.com/c/tQ5dbzp1/73-java-8u51.

kennethgillen:
`ALL` will allow the low-strength ciphers to be used, though, right? Should we be using something like `HIGH:!DSS:!aNULL@STRENGTH` a la http://stackoverflow.com/questions/3775836/disable-weak-ciphers-in-ssl-connection or `ALL:!ADH:!LOW:!EXP:!MD5:@STRENGTH` as deemed 'reasonable' by Ice ref https://doc.zeroc.com/display/Ice35/IceSSL+Properties#IceSSLProperties-IceSSL.Ciphers

mtbc:
Testing: Probably good to try with the "secure connection" lock icon on the server login dialog both on and off.

ximenesuk:
@kennethgillen  `ALL` will allow low-strength but as noted in the discussion Java won't. I would have been happy to try `HIGH` but that isn't documented for Ice! I am trying the more reasonable option locally before considering pushing.

ximenesuk:
This second commit uses the Java version of the 'reasonable' setting suggested above.

joshmoore:
Closing to let #3982 through

PR #3973, opened 2015-07-20
Autofocus on web login username field

manics:
Load OMERO.web. The keyboard focus should default to the username field so instead of clicking on it first you should be able to enter your username straightaway, hit tab, enter your password, and hit enter to login.


mtbc:
Works on Firefox and Chrome. Good to merge.

joshmoore:
Merging after confirming with @will-moore

PR #3974, opened 2015-07-20
Order PlateAcquisitions by name in web

manics:
https://trello.com/c/lvyNfeoJ/42-run-ordering

To test, find a Plate with multiple acquisitions- these should be ordered by name instead of by ID.


manics:
--rebased-to #3975

joshmoore:
screen shot 2015-07-21 at 12 33 46

This looks good to me; it also follows a clean pattern of `order by NAME, ID...` for each of the elements.

joshmoore:
Merging after confirming with @will-moore

PR #3975, opened 2015-07-20
Order PlateAcquisitions by name in web (rebased onto metadata)

manics:
This is the same as gh-3974 but rebased onto metadata.

----

https://trello.com/c/lvyNfeoJ/42-run-ordering

To test, find a Plate with multiple acquisitions- these should be ordered by name instead of by ID.


manics:
--rebased-from #3974

joshmoore:
No longer needed since 5.1.3 has been merged into `metadata`

PR #3976, opened 2015-07-21
Disable SSL ciphers unsupported by Java

joshmoore:
A recent change to Java dropped support for several
SSL ciphers which were being used by Glacier2. This
removes those ciphers from the list so that newer
versions of Java can still connect.


joshmoore:
Closing to open against `develop`.

PR #3977, opened 2015-07-21
Fix DnD autoscroll behaviour at least for Windows

dominikl:
Unfortunately still broken for OSX and Unix; that's a JDK bug,
first reported 2001, apparently never been fixed:
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4407536

Test: Log in as a user with lots of projects/datasets (or datasets with lots of images), expand some project/datatsets so that the tree exceeds the visible area and you get a vertical scrollbar. Select an image, try to drag and drop it to a dataset which lies outside the current visible area by either moving it to the top edge of the visible area (should activate automatic upwards scrolling) or the bottom edge (downwards scrolling).
**Will only properly work on Windows**, on OSX/Unix you have to continuously move the mouse a little bit within the "autoscroll sensible" area on the top or bottom edge of the visible area to trigger the scrolling. But still it is also an improvement for OSX/Unix, because without the PR DnD autoscrolling doesn't work at all.

See [Ticket 12959](https://trac.openmicroscopy.org/ome/ticket/12959)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/util/dnd/DnDTree.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/util/dnd/DnDTree.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.util.dnd.DnDTree 
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2011 University of Dundee & Open Microscopy Environment.
+ *  Copyright (C) 2006-2015 University of Dundee & Open Microscopy Environment.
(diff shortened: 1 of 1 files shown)


will-moore:
Tested on Windows 7 (modern.ie on VirtualBox). Strange behaviour where the scroll jumps to the top during drag, making it impossible to drag/drop very far down the panel.

dominikl:
Indeed, also could confirm this now on Win8, unfortunately haven't noticed that already yesterday. Needs some more investigation, will close this PR and open a new one after I'm back from holidays.

PR #3981, opened 2015-07-22
fix SkipHead hang

mtbc:
Fixes that `SkipHead` requests would hang for clients if an invalid class name was given. Noticed by @ximenesuk in trying something like `bin/omero delete Image/RenderingDefs:1234` (it's actually "RenderingDef" singular). The error isn't yet pretty, but at least it's now promptly delivered.

This PR also reworks and extends integration tests: https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-java/lastCompletedBuild/testngreports/integration.delete/MultiImageFilesetDeleteTest/ should still pass.


ximenesuk:
The tests pass and locally the command returns promptly with an NPE. The CLI shows the full stacktrace which is something that may need fixing in the CLI. Good to merge.

mtbc:
Thank you! In the longer term we should be handling the NPE server-side but this close to release I definitely want to keep changes minimal.

PR #3971, opened 2015-07-16
Modify IceSSL Ciphers to ALL with exceptions (rebased)

ximenesuk:
See gh-3970 for full details.

--rebased-from #3970

PR #3982, opened 2015-07-22
Make all Ice properties configurable

joshmoore:
The OMERO.client implementations: omero/client.java,
omero/clients.py, and omero/client.{h,cpp} were all
setting a number of Ice properties regardless of what
the client had passed in. This seems overly draconian.
The primary reason for allowing it, however, is that
if there are issues with upstream libraries like Java
or OpenSSL, it may be necessary to work around issues
as seen in:

http://blog.openmicroscopy.org/tech-issues/2015/07/21/java-issue/

cc: @manics @ximenesuk -- closing gh-3970 / gh-3971 for the moment.

### DIFF
--- a/components/blitz/src/omero/client.java
+++ b/components/blitz/src/omero/client.java
@@ -120,7 +120,7 @@ public class client {
      * See {@link #setAgent(String)}
      */
     private volatile String __agent = "OMERO.java";
-    
+
     /**
      * See {@link #setIP(String)}
      */
@@ -294,6 +294,14 @@ public class client {
         init(id);
     }
 
+    private void optionallySetProperty(Ice.InitializationData id, String key, String def) {
+        String val = id.properties.getProperty(key);
+        if (val == null || val.length() == 0) {
+            val = def;
+        }
+        id.properties.setProperty(key, val);
+    }
+
     /**
      * Initializes the current client via an {@link Ice.InitializationData}
      * instance. This is called by all of the constructors, but may also be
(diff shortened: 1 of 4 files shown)


manics:
Looks good on the Python side:
```
# ice.config
omero.user=xxx
omero.pass=xxx
omero.host=xxx
IceSSL.Ciphers=ADH:This doesn't exist
```
```
In [64]: cli=omero.client(args=['--Ice.Config=%s' % 'ice.config'])

In [65]: cli.getProperties().getPropertiesForPrefix('IceSSL')
Out[65]:
{'IceSSL.Ciphers': "ADH:This doesn't exist",
 'IceSSL.Protocols': 'tls1',
 'IceSSL.VerifyPeer': '0'}
```

manics:
Java too:
```
# ice.config
omero.user=xxx
omero.pass=xxx
omero.host=xxx
IceSSL.Ciphers=ALL !ADH !LOW !EXP !MD5 !xxxxxx
```
```
package com.example;

public class SimpleConnection {
    public static void main(String[] args) throws Exception {
        omero.client client = new omero.client(args);
        client.createSession();
        try {
            System.out.println("IceSSL: " +
                client.getProperties().getPropertiesForPrefix("IceSSL"));
        } finally {
            client.__del__();
        }
    }
}
```
```
$ ICE_CONFIG=ice.config java -ctarget/classes:OMERO.java-5.1.3-m2-166-e1d00a6-ice35-b570/libs/\* com.example.SimpleConnection
IceSSL: {IceSSL.VerifyPeer=0, IceSSL.Protocols=tls1, IceSSL.Ciphers=ALL !ADH !LOW !EXP !MD5 !xxxxxx}
```

manics:
And c++
```
omero.user=xxx
omero.pass=xxx
omero.host=xxx
IceSSL.Ciphers=xxxxxx
```
```
#include <omero/client.h>
#include <iostream>

using namespace std;

int main(int argc, char* argv[])
{
    omero::client_ptr omero = new omero::client(argc, argv);
    omero::api::ServiceFactoryPrx sf = omero->createSession();
    sf->closeOnDestroy();

    auto map = omero->getProperties()->getPropertiesForPrefix("IceSSL");
    for (auto it = map.begin(); it != map.end(); ++it)
    {
        cout << it->first << ": " << it->second << endl;
    }

    return 0;
}
```
```
$ ICE_CONFIG=ice.config LD_LIBRARY_PATH=target/OMERO.cpp-5.1.3-m3-ice35-SNAPSHOT-Mac\ OS\ X-10.10.4-x86_64/lib ./a.out
libc++abi.dylib: terminating with uncaught exception of type Ice::PluginInitializationException: Instance.cpp:713: Ice::PluginInitializationException:
plug-in initialization failed: IceSSL: unable to set ciphers using `xxxxxx':
no cipher match
Abort trap: 6
```

manics:
Good to merge, unless anyone else has comments?

joshmoore:
Taking that as a no. I was hoping to also have a configuration option (`bin/omero config set`) for the settings in `etc/grid/templates.xml` but that requires a reworking of `jvmcfg.py` (also no longer correctly named). Since it's XML, making changes to that file don't really require a new release (i.e. configurable by nature). Rather than try to rush something like that in, I'm merging this as is after discussing with @sbesson . Thanks, @manics.

PR #3983, opened 2015-07-22
enable more downgrade to private groups

mtbc:
Allow downgrade to private groups even when a root-owned script has been run by a user in the group. To test: import `test_images_good/bd-pathway/2009-05-01_000/` into a read-only group, then downgrade the group to private.

### DIFF
--- a/components/blitz/resources/ome/services/blitz-servantDefinitions.xml
+++ b/components/blitz/resources/ome/services/blitz-servantDefinitions.xml
@@ -390,6 +390,11 @@
       <constructor-arg value="${omero.graphs.wrap}"/>
   </bean>
 
+  <bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
+    <property name="staticMethod" value="ome.services.graphs.GroupPredicate.setSecuritySystem"/>
+    <property name="arguments"><list><ref bean="securitySystem"/></list></property>
+  </bean>
+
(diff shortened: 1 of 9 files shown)


components/blitz/resources/ome/services/blitz-servantDefinitions.xml
```diff
@@ -390,6 +390,11 @@
       <constructor-arg value="${omero.graphs.wrap}"/>
   </bean>
 
+  <bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
+    <property name="staticMethod" value="ome.services.graphs.GroupPredicate.setSecuritySystem"/>
+    <property name="arguments"><list><ref bean="securitySystem"/></list></property>
```
joshmoore: As a bug fix, I can live with this, but in general, static state will get us into trouble and I actively try to remove it where possible. More optimally (eventually) would be to inject either a GroupPredicate or worst case a GroupFactoryFactory where you need it.

ximenesuk:
The suggested testing works fine. The downgrading succeeds and the plate is visible in the newly changed group. Is anything further needed here to test?

pwalczysko:
Further test on the previously failing downgrades passed see cells K9 and K8 of https://docs.google.com/spreadsheets/d/1GMuEmneZmlp-S7w0gf57hXqUYJxARs9iBw_mH4lX_J4/edit#gid=1731867412.
This looks good to merge for me.

PR #3984, opened 2015-07-22
do not chown images' annotations and ROIs in read-annotate groups

mtbc:
Arises from experimentation by @ximenesuk and discussion with @pwalczysko. If chowning an image, in read-annotate and read-write groups the annotations and ROIs should not be chowned. Integration tests should continue to pass.

### DIFF
--- a/components/blitz/resources/ome/services/graph-rules/blitz-chown-rules.xml
+++ b/components/blitz/resources/ome/services/graph-rules/blitz-chown-rules.xml
@@ -269,6 +269,11 @@
         <bean parent="graphPolicyRule" p:matches="L:IAnnotationLink.parent = X:[DI], L.child = A:TagAnnotation[E]{!a}, X =/!o A"
                                        p:changes="A:{a}"/>
 
+        <!-- Do not give annotations in a read-annotate group. -->
+
+        <bean parent="graphPolicyRule" p:matches="A:Annotation[E]{r};perms=???a??" p:changes="A:{a}"/>
+        <bean parent="graphPolicyRule" p:matches="A:Annotation[E]{r};perms=???w??" p:changes="A:{a}"/>
+
         <!-- Delete or give orphaned annotations that are owned by the deleted or given object's owner. -->
 
         <bean parent="graphPolicyRule" p:matches="L:IAnnotationLink.parent = X:[I], L.child = A:Annotation[E]{o}, X =/o A"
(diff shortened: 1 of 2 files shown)


ximenesuk:
This looks to work okay. I tested locally moving data from with all four group types. The changed behaviour with read-annotate was as expected, annotations remained with the original owner.

One observation on what is seen afterwards (via Insight and Web at least). When an Image is `chown`ed to a read-write group the ownership of the image changes but it remains in its original dataset. In discussion with @mtbc this is expected and makes sense. However, this means that the image is not visible to the new owner in the tree view. They would need to specifically look at the other user or look at the Images view in Insight. This is, presumably, a client issue.

mtbc:
That's interesting. Perhaps there's a similar client issue where I can make your orphaned images disappear by putting them in my datasets? :smiley:

One issue that we may want to address in the longer term is that some attachments should possibly still be chowned, like those in the `openmicroscopy.org/omero/measurement` namespace? I can imagine a workflow where someone chowns the image to someone else and then it's moved into a private group and loses its companion files, which may not be terrible but I'll note it on a Trello card anyway.

ximenesuk:
Otherwise this now looks good to merge unless @pwalczysko requires further testing at this stage?

pwalczysko:
@mtbc : Definitely to be able chown the attachments separately is desirable

pwalczysko:
I am confident about @ximenesuk 's tests. Ready to merge.

mtbc:
Yes, attachments can still be chowned, they just now don't go automatically/implicitly with an image in read-annotate.

ximenesuk:
From my first comment I wonder if there should be a mechanism in the clients to somehow view images that are owned but not present in the normal tree view? Almost another type of Orphaned Images structure? /cc @jburel @will-moore

will-moore:
@ximenesuk Handling of orphaned / crosslinked data is contentious and different in the different clients - see https://trello.com/c/6IztFNAq/18-orphaned-dataset-in-another-s-project. The same discussion could apply to images (E.g. is an image an orphan if it's in another users's Dataset).

mtbc:
See also http://trac.openmicroscopy.org/ome/ticket/12143: "Bug: users can make other users' data seem to vanish".

ximenesuk:
Yes, it seems there is a more general problem with invisible data created by various means but still owned by somebody!

joshmoore:
Just clarifying, @will-moore / @mtbc / @ximenesuk -- the above is all for future work, otherwise you're LGTM like @pwalczysko ?

will-moore:
Yes, this can wait.

ximenesuk:
Yep, goode to merge.

mtbc:
Yes, no further commits planned for this PR.

PR #3989, opened 2015-07-24
chmod to private now does unlinking, so warn accordingly

mtbc:
Fixes https://trac.openmicroscopy.org/ome/ticket/12974 for Insight. Try downgrading a group to Private and see if the warning makes sense for what actually happens.


joshmoore:
cc: @ximenesuk @will-moore

pwalczysko:

PR #3993, opened 2015-07-27
Add ability to run scripts against fa's from OMERO.web

chris-allan:
This PR adds the ability to run scripts against file annotations from the user interface which should pave the way for more interesting scripting workflows that run against uploaded and attached files.

Various small pieces of refactoring to allow the script UI to be popped up and interacted with by other parts of the user interface are also included.

Both single object selection,

screen shot 2015-07-27 at 13 49 09

and multiple object selection,

screen shot 2015-07-27 at 13 49 18

are supported. The checkboxes allow for the selection of the file annotation and for convenience a script UI activation button has been added to the attachment section of the righthand panel.

I don't doubt that some robot framework tests will break with this work. A discussion with @will-moore about how to best test the functionality will be forthcoming.

Populate metadata workflows and testing as worked on by @joshmoore and @pwalczysko can also be updated with scripts from glencoesoftware/omero-user-scripts to functionally test the work.


will-moore:
As discussed with @chris-allan:
 - Suggest that we don't show checkboxes on File Annotations initially but that these are shown when user clicks the "Select Files for Scripts" button (and hidden on next click)
 - Scripts are then always run using the regular "Run Script" button at top of page
 - Possibly move check-boxes to be beside the ```- and X``` buttons instead of at the left of FAs
 - Possibly show FAs with a different background 'selection' colour if their checkboxes are checked
 - If we use the convention in the url of "File_Annotation" instead of "File_Annotations" then we are consistent with the use of "Project", "Dataset" etc in the "Data_Type" and "IDs" parameters (Although it does also work with a separate Parameter named "File_Annotations").

chris-allan:
Changes made as outlined by @will-moore above. Screenshots updated.

/cc @emilroz

joshmoore:
Holding off for 5.1.3

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.popup.js
```diff
@@ -292,6 +292,17 @@ OME.openScriptWindow = function(event, width, height) {
                 args.push(key.capitalize() + "=" + sel_types[key].join(","));
             }
         }
+        var fileAnnotationCheckboxes =
+            $("#fileanns_container input[type=checkbox]").filter(":checked");
+        if (fileAnnotationCheckboxes.length > 0) {
+            var fileAnnotationIds = [];
+            fileAnnotationCheckboxes.each(function() {
+                var li = $(this).parent();
```
will-moore: The checkbox parent is now the ```<div>```. You'll need another ```.parent()``` to get the ```<li>``` for the id.

will-moore:
Apart from needing the tiny fix of ``` var li = $(this).parent().parent(); ``` everything is working nicely. 
One very minor thought is that you might want to disable the "Select Files for Script" button if there are no  files on the object. But then you'd need to enable it when some file got attached, so maybe more work than it's worth?

jburel:
cc @gusferguson since we will need to update the help. 
Now that the functionality is available in web. I will check the requested change to be done in insight to support it.

jburel:
Discussed with @dominikl so we can add the same functionality in insight so both clients are in synch

dominikl:
Is there an example script for testing somewhere?

jburel:
@dominikl https://github.com/glencoesoftware/omero-user-scripts see description

emilroz:
@dominikl PR for the populate_metadata.py is here: https://github.com/glencoesoftware/omero-user-scripts/pull/13

And the script here: https://github.com/emilroz/omero-user-scripts/blob/populate_metadata_fa_additions/util_scripts/Populate_Metadata.py

dominikl:
Thanks. What's actually expected, I guess the 'File ID' field should be populated with the selected FileAnnotations, right? Doesn't happen for me.
screen shot 2015-08-13 at 10 01 40

emilroz:
@dominikl it's the wrong version of the script. It should have ```File Annotation``` instead of ```File ID```. The correct version of the script is 0.3. Copying the code from here should work: https://github.com/emilroz/omero-user-scripts/blob/populate_metadata_fa_additions/util_scripts/Populate_Metadata.py

dominikl:
Yes, sorry @emilroz  . Everything's fine!

jburel:
@gusferguson: PR to be added to your list of "screenshots" required.

jburel:
Merging since the insight one has been merged.
This will have to be rebased to 5.1

sbesson:
--rebased-to #4126

PR #3997, opened 2015-07-28
Quick fix to Python client ctor test

ximenesuk:
This fixes a broken test by setting the port.

### DIFF
--- a/components/tools/OmeroPy/test/integration/test_client_ctors.py
+++ b/components/tools/OmeroPy/test/integration/test_client_ctors.py
@@ -57,6 +57,7 @@ class TestClientConstructors(lib.ITest):
         id = Ice.InitializationData()
         id.properties = Ice.createProperties()
         id.properties.setProperty("omero.host", self.host)
+        id.properties.setProperty("omero.port", str(self.port))
         id.properties.setProperty("omero.user", "root")
         id.properties.setProperty("omero.pass", self.rootpasswd)
         c = omero.client(id=id)


mtbc:
The diff looks reasonable.

sbesson:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/514/testReport/test.integration.test_client_ctors/TestClientConstructors/ is green with this PR included. Merging.

components/tools/OmeroPy/test/integration/test_client_ctors.py
```diff
@@ -57,6 +57,7 @@ def testInitializationDataConstructor(self):
         id = Ice.InitializationData()
         id.properties = Ice.createProperties()
```
ximenesuk: So now this is merged I have worked out what is going on here. This line returns an empty `Ice.PropertiesI` object, which is then only partially populated because the port isn't set by default. This means that when the client constructor gets to here https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/clients.py#L132 the hard-coded `4064` is used.

In uses of the constructor when `id` is not used the call to `createProperties` here https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/clients.py#L127 is made with an explicit `args` list which may be empty. Even with an empty list the properies are created correctly and hence this is why the other tests, using other constructor arguments, pass even though port is not set explicitly.

Thus a better fix may be:
```id.properties = Ice.createProperties([])```
which properly creates the Ice properties object.

PR #3999, opened 2015-07-28
Add an additional Python client ctor test

ximenesuk:
See gh-3997 for discussion. The integration tests should pass.

(Minor typo fixed by @mtbc included here so that it doesn't feel lonely in its own PR)

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/export.py
+++ b/components/tools/OmeroPy/src/omero/plugins/export.py
@@ -20,7 +20,7 @@ Example Usage:
   omero export --file new.ome.xml --type XML Image:1
   omero export --file output-dir  --iterate  Dataset:2
 
-If --iterate is used used, the only supported obj is Dataset:<id>
+If --iterate is used, the only supported obj is Dataset:<id>
 """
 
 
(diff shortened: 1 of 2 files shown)


sbesson:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration-python/515/testReport/test.integration.test_client_ctors/TestClientConstructors/ certainly looks good with this change in. :+1: from me. @joshmoore: does that test addition make sense to you?

ximenesuk:
Incidentally, in the code base we have a mix of calls to `createProperties` that have no specific args, some with no args and some with empty lists. Testing both uses makes sense but it may be worth reviewing the other language tests.

joshmoore:
Seeing the same on Ice 3.4:

```
>>> Ice.createProperties().getProperty("foo")
''
>>> Ice.createProperties([]).getProperty("foo")
'bar'
>>> Ice.initialize().getProperties().getProperty("foo")
''
>>> Ice.initialize([]).getProperties().getProperty("foo")
'bar'
>>> Ice.__file__
'/opt/ice/ice-3.4.2/python/Ice.py'
```

I find this surprising, but certainly better to have it recorded. Merging.

PR #4000, opened 2015-07-29
MIF and overlay

jburel:
Problem reported by @pwalczysko: see https://trello.com/c/oeLB1cl3/28-mif-roi-proliferation
To test this PR:

test 1:
 * Open a mif in imageJ from local file system
 * Select only one series
 * Draw a roi and add it as overlay. Do not use the roi manager.
 * Save the image to OMERO
 * Check that only the series opened has the overlay.

test 2:
 * Repeat the same steps but this time open few series.

test 3:
 * Repeat the same step but this time open a non mif image e.g. dv

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/ImporterModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/ImporterModel.java
@@ -22,7 +22,6 @@
  */
 package org.openmicroscopy.shoola.agents.fsimporter.view;
 
-import ij.IJ;
 import ij.ImagePlus;
 
 import java.io.File;
@@ -631,7 +630,6 @@ class ImporterModel
                     o = j.next();
                     if (o.isImagePlus()) {
                         index = o.getIndex();
-                        IJ.log("index: "+index);
                         rois = reader.readImageJROI(-1, (ImagePlus) o.getFile());
                         indexes.put(index, rois);
                         if (index < 0) {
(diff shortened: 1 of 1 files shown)


bramalingam:
Overlay-MIF specific RFE's/Bug's:

test 1 and test 3 work without any bugs.

test 2: 
--Open multiple series, add overlay's in all of them. 
--Now save images back to OMERO (with "Add Images from all image windows") option checked

Bug: Saves the MIF properly (only once, no duplications seen) But the ROI's are saved from the currently active window/series from ImageJ and all the ROI's from the other Window's/Series's are not preserved.

Some Bug's which are not specific to this PR:
--IJ.log triggered somewhere post import of ROI’s
--Reimporting an image from disk is allowed, whereas saving ROI’s check's if the Image is already present within OMERO.
--Cannot open any image in “view mode” (in insight viewer). Shows dialog: Loading Render data: "The image could not be opened.The image is not a valied image".

pwalczysko:
test1 : Passes as expected.

pwalczysko:
test2: - can confirm the bug reported by Balaji partially. He opened 2 images out of the series and drawn 1 ROI on each of them, and 1 of these ROIs got lost. I opened 3 images out of the series and drawn also 1 ROI on each of them, and 1 of these ROIs got lost as well (2 were preserved though). It loosk like 1 ROI on one image gets always lost...

pwalczysko:
> Reimporting an image from disk is allowed, whereas saving ROI’s check's if the Image is already present within OMERO.

Yes, I can confirm this. Is this a bug though ? I think if I deliberately want to reimport an image, I should be allowed to do so ? With the ROIs, I think we all agree that this is expected behaviour (was explicitly asked by the users)

pwalczysko:
Cannot confirm the "view mode" bug, for me this worked at all stages (in Fiji).

pwalczysko:
test3 passes as expected.

jburel:
To simplify review and the fix I am going to combine the 2 PRs

jburel:
merged with gh-4001.
Closing

PR #4002, opened 2015-07-30
replacing javadeploy webstart button with direct link

aleksandra-tarkowska:
Regarding what @rleigh-dundee requested:
> If I enter the URL directly, chrome downloads it without prompting, but if I then click on the download link it fires up webstart.  Not sure why it doesn't do that automatically, but certainly exposing it would help rather than hinder the user even if it's not 100% automated.

test all web browsers if webstart starts up nicely

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webstart/templates/webstart/includes/webstart_insight_script.html
+++ b/components/tools/OmeroWeb/omeroweb/webstart/templates/webstart/includes/webstart_insight_script.html
@@ -21,12 +21,7 @@
 {% endcomment %}
 
 {% if insight_url %}
-<script src="//www.java.com/js/deployJava.js"></script>
-<script>
-    deployJava.launchButtonPNG='{% static "webstart/img/icon-omero-insight.png" %}';
-    var url = "{{ insight_url }}";
-    deployJava.createWebStartLaunchButton(url, '1.6.0');
-</script>
+    <a href="{{ insight_url }}"><img src="{% static "webstart/img/icon-omero-insight.png" %}"/></a>
 {% else %}
     <h1 class="error">Webstart is not available</h1>
 {% endif %}


components/tools/OmeroWeb/omeroweb/webstart/templates/webstart/includes/webstart_insight_script.html
```diff
@@ -21,12 +21,7 @@
 {% endcomment %}
 
 {% if insight_url %}
-<script src="//www.java.com/js/deployJava.js"></script>
-<script>
-    deployJava.launchButtonPNG='{% static "webstart/img/icon-omero-insight.png" %}';
-    var url = "{{ insight_url }}";
-    deployJava.createWebStartLaunchButton(url, '1.6.0');
-</script>
+    <a href="{{ insight_url }}"><img src="{% static "webstart/img/icon-omero-insight.png" %}"</a>
```
mtbc: missing a `>` to close `<img`?

rleigh-dundee:
This isn't working I'm afraid, and is more complicated than it needs to be.

At the bottom, there's a link "To go to the webclient, click here".  All this needs is an equivalent "To go to the webstart JNLP file, click here", so allow the user to download and/or run the file directly *under all circumstances*.  The change in this PR still uses overly complex conditional logic, when we really want it to be unconditional.  The logic in this PR isn't *wrong*, it's just that it doesn't provide a direct link for when it doesn't work--it doesn't allow the user to always get a direct link.

jnlp-firefox-linux
jnlp-chromium-linux

Firefox - Linux: Javascript link, no direct link
Chromium - Linux: Javascript link, no direct link
Firefox - MacOS X: Javascript link, no direct link
Firefox - Windows: Broken javascript and additional direct link
IE - Windows: Javascript link, no direct link

rleigh-dundee:
Testing against, eel, works correctly on all platforms and browsers.  The only exception was Safari due to lack of codesigning, so should work there for releases or if configured properly.

Good to merge.

aleksandra-tarkowska:
--rebased-to #4016

jburel:
Thanks merging

PR #4005, opened 2015-07-31
CLI: add sessions key

sbesson:
See https://trello.com/c/dyRbaNiM/19-key-omero-sessions-key

Implement a simple RFE to return the current session key via the CLI. To test this PR:
- check the new integration test is green
- from the command line check the output of `bin/omero sessions key` is null if logged out and the current session key if logged in
- test a simple workflow reusing the session key:

  ```
   bin/omero login USER@HOSTNAME
   bin/omero -C -s HOSTNAME -k $(bin/omero sessions key) COMMAND
```

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -241,7 +241,10 @@ class SessionsControl(BaseControl):
         file = parser.add(
             sub, self.file, "Print the path to the current session file")
 
-        for x in (file, logout, keepalive, list, clear, group):
+        key = parser.add(
+            sub, self.key, "Print the key of the current active session")
+
+        for x in (file, key, logout, keepalive, list, clear, group):
             self._configure_dir(x)
 
     def _configure_login(self, login):
@@ -820,10 +823,18 @@ class SessionsControl(BaseControl):
             t.event.set()
 
     def file(self, args):
+        """Return the file associated to the current active sessions"""
         store = self.store(args)
         srv, usr, uuid, port = store.get_current()
         if srv and usr and uuid:
             self.ctx.out(str(store.dir / srv / usr / uuid))
+
+    def key(self, args):
+        """Return the key associated to the current active sessions"""
+        store = self.store(args)
+        srv, usr, uuid, port = store.get_current()
+        if uuid:
+            self.ctx.out(uuid)
 
     def conn(self, properties=None, profile=None, args=None):
         """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: components/tools/OmeroPy/src/omero/plugins/sessions.py [hunk 1]
```diff
@@ -820,11 +823,19 @@ def run(self):
...
         store = self.store(args)
         srv, usr, uuid, port = store.get_current()
         if srv and usr and uuid:
             self.ctx.out(str(store.dir / srv / usr / uuid))
 
+    def key(self, args):
+        """Return the key associated to the current active sessions"""
+        store = self.store(args)
+        srv, usr, uuid, port = store.get_current()
+        if uuid:
```
joshmoore: `else: self.ctx.die(321, "No key found")` ?

sbesson: Can do. In that case, I was mirroring the behavior of `session file` which does not throw a CLI exception is no local session file is found. Either of them is fine.

sbesson: Can do. In that case, I was mirroring the behavior of `session file` which does not throw a CLI exception is no local session file is found. Either of them is fine.

components/tools/OmeroPy/src/omero/plugins/sessions.py
```diff
@@ -820,11 +823,19 @@ def run(self):
             t.event.set()
 
     def file(self, args):
+        """Return the file associated to the current active sessions"""
```
mtbc: "associated /with/" is more idiomatic (but "related to")

mtbc:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-python/lastCompletedBuild/testReport/test.integration.clitest.test_sessions/TestSessions/ looks good.

mtbc:
Functionally seems to work excellently, it's a good addition to the CLI that I've been wanting.

sbesson:
@mtbc: pushed

mtbc:
:+1: Looks good to me!

sbesson:
@jburel: re https://github.com/openmicroscopy/openmicroscopy/pull/4005#discussion_r36017509, I suggest we get this merged as it is and rebase onto develop. I will open a `develop`-only PR changing the return code of `sessions file/key` to return an error if empty and we can decide whether we want to backport it into `dev_5_1`.

jburel:
merging. @joshmoore's comment will be addressed in another PR

sbesson:
--rebased-to #4022

PR #4013, opened 2015-08-03
remove workaround for namespace bug in old gcc

mtbc:
C++ integration tests should pass even with gcc 4.6. Cc: @rleigh-dundee.

--no-rebase because a bug in gcc 4.4 requires the namespace specifications

### DIFF
--- a/components/tools/OmeroCpp/test/integration/chgrp.cpp
+++ b/components/tools/OmeroCpp/test/integration/chgrp.cpp
@@ -42,8 +42,8 @@ TEST(ChgrpTest, testSimpleChgrp ) {
     image->setName( rstring("testSimpleChgrp") );
     image = ImagePtr::dynamicCast( iupdate->saveAndReturnObject( image ) );
 
-    omero::api::LongList imageIds;
-    omero::api::StringLongListMap objects;
+    LongList imageIds;
+    StringLongListMap objects;
     ChildOptions options;
     Chgrp2Ptr chgrpCmd = new Chgrp2();
     imageIds.push_back( image->getId()->getValue() );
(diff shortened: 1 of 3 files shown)


rleigh-dundee:
The changes looks fine, though the practice of lots of "using namespace ..." is not recommended in general (not a criticism of this PR, just the original way this was written).

However, does this mean we are dropping support for GCC 4.4 in CentOS6?

(I'd quite like to, and we have 4.9.3 available in the SCL devtoolset-3 package, but just wanted to double-check first since it means losing binary compatibility with the base system.  That said, the length of time we can reasonably support LTS releases is something which we might want to discuss.)

mtbc:
Added to https://www.openmicroscopy.org/site/community/minutes/conference-calls/2015/2015-08-04-tuesday-team-meeting for discussion.

rleigh-dundee:
Note this will fail tomorrow--see my note in the failing builds section of the standup notes.  Not hard to resolve, I just didn't want to interfere with any ongoing work I was unaware of.  /cc @sbesson

rleigh-dundee:
Green in https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-cpp/310/

From the point of view of correctness, this PR looks fine, and the GCC 4.9 provided by devtoolset-3 works, so could be enabled more widely on our CentOS 6 CI nodes.  We should probably wait to see if there's consensus on removing GCC 4.4 support before merging though.

mtbc:
Are non-breaking CI nodes now okay to require gcc 4.6 so label can be removed?

rleigh-dundee:
@mtbc Looks fine to remove the breaking label; we'll just need to tweak the merge job to use a later GCC as well which is a simple copy-paste of the logic.

mtbc:
Thank you, much appreciated.

rleigh-dundee:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-cpp/7/ and https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-cpp-cmake/ are green.  Please merge.

PR #4016, opened 2015-08-04
replacing javadeploy webstart button with direct link (rebased onto dev_5_1)

aleksandra-tarkowska:
This is the same as gh-4002 but rebased onto dev_5_1.

----

Regarding what @rleigh-dundee requested:
> If I enter the URL directly, chrome downloads it without prompting, but if I then click on the download link it fires up webstart.  Not sure why it doesn't do that automatically, but certainly exposing it would help rather than hinder the user even if it's not 100% automated.

test all web browsers if webstart starts up nicely


aleksandra-tarkowska:
--rebased-from #4002

rleigh-dundee:
Works the same as the original PR and is correct.  Please merge.

aleksandra-tarkowska:
Thanks you @rleigh-dundee

PR #4009, opened 2015-08-03
Insight cleanup

jburel:
while looking at QA https://www.openmicroscopy.org/qa2/qa/feedback/11334/
I noticed that we still have few references to the editor and a build target.
Removing the remaining bits

I have also removed the out-of-date change logs. It has not been updated since 2010.
Check that the jobs are green

PR #4018, opened 2015-08-04
Insight cleanup dev 5 1

jburel:
gh-4009 rebased on dev_5_1

while looking at QA https://www.openmicroscopy.org/qa2/qa/feedback/11334/
I noticed that we still have few references to the editor and a build target.
Removing the remaining bits

I have also removed the out-of-date change logs. It has not been updated since 2010.
Check that the jobs are green


sbesson:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-build/584/ being green, merging.

PR #4015, opened 2015-08-03
Make Java 1.7 the minimum supported version

rleigh-dundee:
Note this does not change jarsigner, which can be upgraded to use 1.7 features separately.

- Bump minimum version numbers from 1.6 to 1.7
- Drop special cased building of Java6 MacOS X app bundles

--------

Testing:

- Check travis and CI builds are green.  This shouldn't cause any breakage, but it's possible if we have any CI nodes using Java6 they will fail the build and need upgrading
- Check client and server are still fully functional; we're bumping the JVM minimum version, so there might be potential for issues with the 1.7 bytecode (unlikely, but still needs testing)
- Check all clients, particularly Mac client in case I broke the packaging.

PR #4020, opened 2015-08-04
use narrower reflective operation exception

mtbc:
Should be no change in behavior, largely tested by if the code compiles at all.

--no-rebase as Java 6 does not offer `ReflectiveOperationException`


jburel:
All files modified.
This is Java 1.7 specific.
see gh-4015 for 1.7 being the minimum

mtbc:
So basically, good once #4015 is good?

jburel:
@mtbc: yes.

mtbc:
:+1: Makes sense.

PR #4004, opened 2015-07-31
Ij measurement

jburel:
Allow the user to specify the name of the file
To Test:
 * Open an image in imageJ. Image saved in OMERO
 * draw a region. 
 * Go to Analyze> Measure
 
test 1:
 * Enter a name for the measurement e.g. test
 * Check that the name is preserved and the saved file is test.csv

test 2:
 * Enter a name for the measurement e.g. test.foo
 * Check that the name is preserved, the extension is remove and the saved file is test.csv

test 3:
 * Keep the field blank
 * Check that the default name is set i.e. ImageJ-NameOfImage-Date.csv

PR #4025, opened 2015-08-06
Ij measurement (rebased onto dev_5_1)

jburel:
This is the same as gh-4004 but rebased onto dev_5_1.

----

Allow the user to specify the name of the file
To Test:
 * Open an image in imageJ. Image saved in OMERO
 * draw a region. 
 * Go to Analyze> Measure
 
test 1:
 * Enter a name for the measurement e.g. test
 * Check that the name is preserved and the saved file is test.csv

test 2:
 * Enter a name for the measurement e.g. test.foo
 * Check that the name is preserved, the extension is remove and the saved file is test.csv

test 3:
 * Keep the field blank
 * Check that the default name is set i.e. ImageJ-NameOfImage-Date.csv

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ResultsSaver.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ResultsSaver.java
@@ -38,6 +38,7 @@ import org.openmicroscopy.shoola.env.data.model.ResultsObject;
 import org.openmicroscopy.shoola.env.data.util.SecurityContext;
 import org.openmicroscopy.shoola.env.data.views.BatchCall;
 import org.openmicroscopy.shoola.env.data.views.BatchCallTree;
+import org.openmicroscopy.shoola.util.CommonsLangUtils;
 import org.openmicroscopy.shoola.util.roi.io.ROIReader;
 
 import com.google.common.io.Files;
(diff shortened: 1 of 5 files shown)


jburel:
--rebased-from #4004

jburel:
@bramalingam is focusing on fixing readers. This PR can be reviewed later on.

pwalczysko:
Works as expected with 
ImageJ 1.50 Java 1.8.0_45 (64 bit) downloaded as ``experimental version of ImageJ`` in the Mac OS X paragraph of http://imagej.nih.gov/ij/download.html

pwalczysko:
Works as expected with ImageJ 1.49o Java 1.6.0. 
Ready to merge.

pwalczysko:
Please note that if you open an image from local disk and go through the workflow (create ROI, save Analyse > Measure forllowed by Save Results to OMERO) there is a bug:
  - the first time (when the image is NOT YET in OMERO), the naming of the results csv file falls to default, irrespective of what you are putting into the dialog

I guess this bug is worth fixing. Leave it to JM if he wants to do it in another PR (the PR here does what it says it do in the header and is a rebase, nevertheless the bug above is quite an important case.)

jburel:
@pwalczysko: Thanks I will fix the problem in this PR. It will have to be done in develop too.
Another case to add to the scenario ;-)

pwalczysko:
The bug https://github.com/openmicroscopy/openmicroscopy/pull/4025#issuecomment-129973748 is fixed now. All good here.

jburel:
Thanks @pwalczysko I will cherry-pick the last commit (not in develop)

components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ResultsSaver.java
```diff
@@ -77,14 +78,20 @@
      * Create a temporary file
      *
      * @param img The image object to handle.
+     * @param fileName The name to use if any.
      * @return See above.
      */
-    private File createFile(ImagePlus img)
+    private File createFile(ImagePlus img, String fileName)
```
sbesson: Only question on this PR: is this okay to do on a stable line? Or should create a `private File createFile(ImagePlus img)` method for backwards-compatibilty?

jburel: private method so only used within the class. So it won't break anything.

sbesson: Understood. Assuming all usages within the class have been updated, merging.

PR #4010, opened 2015-08-03
have model object glossary link unit properties to best Javadoc

mtbc:
Generates https://github.com/openmicroscopy/ome-documentation/pull/1251 which is staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/developers/Model/EveryObject.html. There should now be links from dimensioned quantities to reasonable units Javadoc.

Also shortens anchor names.

--rebased-to #4027

PR #4027, opened 2015-08-06
have model object glossary link unit properties to best Javadoc (rebased from develop)

mtbc:
Generates openmicroscopy/ome-documentation#1260 which is staged at http://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Model/EveryObject.html. There should now be links from dimensioned quantities to reasonable units Javadoc.

Also shortens anchor names.

--rebased-from #4010


sbesson:
All relevant builds being green, merging.

Issue #4033, opened 2015-08-11
Importing images from ImageJ fails if they are loaded from non-OME TIFFs


stelfrich:
It is not possible to import images that have been loaded from a TIFF file to OMERO with OMERO.insight-ij.5.1.3. The result is an exception:

```
[Fatal Error] :1:1: Content is not allowed in prolog.
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at org.openmicroscopy.shoola.env.data.model.FileObject.isNewImage(FileObject.java:264)
	at org.openmicroscopy.shoola.agents.fsimporter.chooser.FileSelectionTable.allowAddToQueue(FileSelectionTable.java:399)
	at org.openmicroscopy.shoola.agents.fsimporter.chooser.FileSelectionTable.addFiles(FileSelectionTable.java:579)
	at org.openmicroscopy.shoola.agents.fsimporter.chooser.ImportDialog.addImageJFiles(ImportDialog.java:1608)
	at org.openmicroscopy.shoola.agents.fsimporter.chooser.ImportDialog.propertyChange(ImportDialog.java:1671)
	at java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:339)
	at java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:276)
	at java.awt.Component.firePropertyChange(Component.java:8154)
	at org.openmicroscopy.shoola.agents.fsimporter.chooser.LocationDialog.actionPerformed(LocationDialog.java:950)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1995)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2318)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:387)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:242)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:236)
	at java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:272)
	at java.awt.Component.processMouseEvent(Component.java:6289)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3267)
	at java.awt.Component.processEvent(Component.java:6054)
	at java.awt.Container.processEvent(Container.java:2041)
[...]

PR #4034, opened 2015-08-11
Newimage

jburel:
Fix issue https://github.com/openmicroscopy/openmicroscopy/issues/4033

To Test:
 * Open any image an save it to TIFF. 
 * Open the newly created TIff image
 * Save that image to OMERO (Plugins > OMERO > Save Image(s) to OMERO).

cc @stelfrich, @bramalingam 

Also test the parsing of the string with @bramalingam macro

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/FileObject.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/FileObject.java
@@ -328,21 +328,48 @@ public class FileObject
             }
             String xmlStr = info.description;
             if (CommonsLangUtils.isBlank(xmlStr)) return false;
-            //Get Current Dimensions
+          //Get Current Dimensions
             int sizeC_cur = img.getNChannels();
             int sizeT_cur = img.getNFrames();
             int sizeZ_cur = img.getNSlices();
-
-            Document doc = xmlParser(xmlStr);
-            Node node = getPixelsNode(doc);
-            if (node == null) return false;
-            NamedNodeMap nnm = node.getAttributes();
             int sizeC_org = sizeC_cur;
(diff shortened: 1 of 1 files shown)


bramalingam:
Both the macro and the 'File-->Open" of tiff images, work as expected. Good to merge.

stelfrich:
I find the handling of the issue a bit misleading, tbh.

With this PR, the SAXParseException is logged and everything continues as usual. As an unexperienced user, if you haven't had the log window open before, it will popup and show you an error but the import can continue. On first sight of the Exception, I killed Fiji because I thought the import wasn't possible..

jburel:
@stelfrich: we can certainly check before building the Document if the string "looks like" XML so we won't mislead the user.

stelfrich:
@jburel I have made some slight changes based on this PR. But I do I proceed now? I can't directly push to the PR, can I?

jburel:
@stelfrich: You will have to open a PR against this PR. Then I will merge your changes and they will be included in the build

stelfrich:
@jburel Do have an idea where this error is logged during the parsing process:
```[Fatal Error] :1:1: Content is not allowed in prolog.```

I am a bit lost but I think we should get rid of the too, to not confuse users..

jburel:
@stelfrich: This is due to the creation of the Document, it might be better to actually check before creating the Document if the description "looks like " some XML.
I will look at that today

bramalingam:
No errors found in the log. Both workflows : opening tiff from local disk (using file-->open) and change of image properties using the macro, work as expected. Good to merge, according to me.!

sbesson:
@jburel: no more commit coming on this PR?

jburel:
That one is done.

jburel:
--rebased-to #4138

PR #4035, opened 2015-08-11
Add null check if no table.

jburel:
Add null check if no table found

Select an image w/o map annotations
and click on the green + button above table annotation

cc @imunro

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/MapAnnotationsComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/MapAnnotationsComponent.java
@@ -665,6 +665,7 @@ public class MapAnnotationsComponent extends JPanel implements
 		MapTable t = getSelectedTable();
 		if (t == null)
 			t = getUserTable();
+		if (t == null) return;
 		MapTableModel m = (MapTableModel) t.getModel();
 		// if nothing's selected add to end of table, otherwise below selected
 		// row


gusferguson:
@jburel 

Tested using OMERO.insight-5.1.3-65-e6ccebb-ice35-b589-mac_Java7+ trout user-3 read-only-1.

Behaves as expected.
Good to merge.

components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/MapAnnotationsComponent.java
```diff
@@ -665,6 +665,7 @@ private void insertRow() {
 		MapTable t = getSelectedTable();
 		if (t == null)
 			t = getUserTable();
+		if (t == null) return;
```
sbesson: Since the same check is performed one line above, should we combine both blocks?

jburel: we could check that they are both null i.e. ```getSelectedTable() == null && getUserTable() == null``` return;
but we will still need to assign it if  ```getSelectedTable()``` is ``` null```

sbesson: Ah misread the behavior. This makes sense

jburel:
--rebased-from #4043

PR #4040, opened 2015-08-12
Skip testTiles if Python is not 2.7 or above

ximenesuk:
With the Ice3.5 and Python 2.6 the following test fails (due to changes in Ice effectively), see:

https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/50/testReport/OmeroPy.test.integration.test_rawpixelsstore/TestTiles/testTiles/

This PR skips that test unless Python is 2.7 or higher. This commit can be reverted once support of Python 2.6 is removed. The awkward mport is due to `flake8`.

The above test should be skipped on that ci node, it should pass locally with Python 2.7 and Ice 3.5.

--no-rebase

### DIFF
--- a/components/tools/OmeroPy/test/integration/test_rawpixelsstore.py
+++ b/components/tools/OmeroPy/test/integration/test_rawpixelsstore.py
@@ -4,7 +4,7 @@
 """
    Tests for the stateful RawPixelsStore service.
 
-   Copyright 2011-2014 Glencoe Software, Inc. All rights reserved.
+   Copyright 2011-2015 Glencoe Software, Inc. All rights reserved.
    Use is subject to license terms supplied in LICENSE.txt
 
 """
--- a/components/tools/OmeroPy/test/integration/test_rawpixelsstore.py
+++ b/components/tools/OmeroPy/test/integration/test_rawpixelsstore.py
@@ -12,6 +12,8 @@
 import omero
 import threading
 import library as lib
+import pytest
+__import__("sys")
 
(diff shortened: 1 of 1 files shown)


jburel:
@ximenesuk: maybe indicates only ice version = 3.5 unless you have tried with ice 3.6

ximenesuk:
See https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-python/51/testReport/

Fails -1, skips +1

As this is not in DEV breaking that branch may still have the fail but if this test is chnage is merged that should resolve too.

ximenesuk:
@jburel I haven't tested with Ice 3.6 though the breaking change was between Ice 3.4 and Ice 3.5. I don't know what the status of 3.6 is with Python 2.6 - are they compatible. /cc @rleigh-dundee 

However, this fix resolves the fail on the current CI node but the skip will be removed once we drop support for Python 2.6

jburel:
What I mean: you wrote ```Ice >= 3.5```. it should probably be Ice = 3.5

ximenesuk:
Thanks for the clarification @jburel, I won't change the skip message yet as it may be possible for someone to confirm Ice3.6 behaviour on different Python systems.

jburel:
@ximenesuk: okay, as long as we review the test when we drop python 2.7.
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/51/ is green.
Merging

PR #4043, opened 2015-08-12
Add null check if no table.

jburel:
Add null check if no table found

Select an image w/o map annotations
and click on the green + button above table annotation

--rebased-from #4035


gusferguson:
@jburel 

Tested with OMERO.insight-5.1.3-69-3fabcc8-ice34-b590-mac_Java7+ on eel member-all-1 red-annotate-1

Behaves as expected.
Good to merge.

PR #4045, opened 2015-08-13
Fixed storing of ROIs from non-hyperstack images from ImageJ

stelfrich:
ROIs aren't imported with their correct positions to OMERO because of the weird hyperstack handling of ImageJ. This PR fixes the issue for images with the following dimensions:
* C=1, Z=1, T=5
* C=3, Z=1, T=5
* C=1, Z=3, T=5
* C=3, Z=1, T=1
* C=1, Z=3, T=1

PR #4046, opened 2015-08-13
Fix composite ROI import from ImageJ

stelfrich:
The subROIs of a composite ROI were never actually read because the type of the parent was checked instead of the subROI's type.

Furthermore, the position of the subROIs is not properly set if they are acquired via ```ShapeRoi.getRois()```. If have fixed that for now, but it might also be a good idea to propose a change to Wayne to address that issue on the IJ side. This fix is crucial for PR #4045 to work properly.


jburel:
restarted travis

jburel:
closing replaced by https://github.com/openmicroscopy/openmicroscopy/pull/4170

PR #4049, opened 2015-08-13
Webstart

jburel:
Use Number format to read port.
To test:
 * Open the web client
 * Start insight from web
 * Check that insight correctly re-connect to the session on the correct port.


jburel:
Closing this PR, probably going to deprecate webstart. Major browsers will stop supporting it.

PR #4051, opened 2015-08-14
Add ability to run scripts against fa's from OMERO.Insight

dominikl:
Same as #3993 just for Insight

Test:
- Install @emilroz  's script which uses file annotations: https://github.com/emilroz/omero-user-scripts/blob/populate_metadata_fa_additions/util_scripts/Populate_Metadata.py 
- Add some file annotations to a plate
- Select the plate and some file annotations and run the script, which should open the script dialog
- Make sure the Plate ID and the the file annotations IDs are set in the dialog
- Run the script. Will fail, but in 'Info' in the activities popup you should see, that the parameters were passed correctly to the script.
- Additional test: Run some other scripts, to make sure the changes didn't break the script dialog for other scripts.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AcquisitionDataUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AcquisitionDataUI.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.metadata.editor.AcquisitionDataUI 
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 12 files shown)


will-moore:
@dominikl Sorry I missed this yesterday. All works fine.
Minor issue: If I click the button to show the checkboxes, then I browse to a different Object, the checkboxes are already shown. I would expect them to be returned to the 'hidden' state whenever a new object is loaded.
One trivial suggestion: 
I wonder if you could reduce the prominence of the "Select Files For Scripts" button by removing the blue & red window from the icon, and just go for the plain script icon. With the current icon, the button is about the most eye-catching button in the right panel which is inconsistent with it's niche use-case.

screen shot 2015-08-18 at 11 11 46

dominikl:
Thanks @will-moore , added your suggestions.

will-moore:
Looks good @dominikl, thanks. Good to merge.

jburel:
@gusferguson: this is the PR discussed yesterday during group meeting,
The corresponding web one is gh-3993

dominikl:
This should also be rebased to dev_5_1, right?

jburel:
@dominikl: to avoid major rebasing issues, both PRs will be rebased to 5.1 (cc @emilroz @chris-allan) as discussed yesterday

dominikl:

jburel:
More testing required @dominikl?

dominikl:
I don't think so. I just had to solve a merge conflict after #4092 was merged.

jburel:
@gusferguson: PR to be added to your list of "screenshots" required for help

gusferguson:
@jburel - thanks

jburel:
--rebased-to #4129

PR #4036, opened 2015-08-11
Javadoc cleanup: model

sbesson:
Follow-up of https://github.com/openmicroscopy/openmicroscopy/pull/4031

As part of the build system cleanup, this PR reviews the Javadoc warnings for the model component of the server and removes unnecessary import headers.

To test this PR, check no more warning is reported for the `model` component when running `./build.py build-default`.

PR #4057, opened 2015-08-17
 Javadoc cleanup: common

sbesson:
Follow-up of #4036

As part of the build system cleanup, this PR reviews the Javadoc warnings for the common component of the server and removes unnecessary import headers and internal versions.

To test this PR, check no more warning is reported for the common component when running `./build.py build-default`.

--no-rebase


components/common/src/ome/api/JobHandle.java
```diff
@@ -116,7 +116,7 @@ long submit(@NotNull
      * Sets the job's message string, and returns the previous value.
 
      * @param message
-     * @return
+     * @return thee previous message value
```
rleigh-dundee: the

components/common/src/ome/api/Search.java
```diff
@@ -158,7 +158,7 @@
 
     /**
      * Returns the current case sensitivity setting. If
-     * {@link #setCaseSentivice()} has not been called, the
+     * {@link #setCaseSentivice(boolean)} has not been called, the
```
rleigh-dundee: Sensitive ?

components/common/src/ome/api/Search.java
```diff
@@ -411,10 +410,11 @@ void addOrderByDesc(@NotNull
     // =========================================================================
 
     /**
-     * Returns transient (without ID) {@link TextAnnotation} instances which
-     * represent terms which are similar to the give terms. For example, if
-     * the argument is "cell", one return value might have as its textValue:
-     * "cellular" while another has "cellularize".
+     * Returns transient (without ID)
+     * {@link ome.model.annotations.TextAnnotation} instances which represent
+     * terms which are similar to the give terms. For example, if the argument
```
rleigh-dundee: given

components/common/src/ome/api/Search.java
sbesson: That's what I also assumed first but this is the current name of the method. We can definitely rename it (follow-up PR?) but I have no view of the impact of this change on the rest of the code.

rleigh-dundee: A followup PR would be fine.  Could be deferred to 5.2 if it's going to break stuff, but if it's already broken then it's probably safe to fix now.

sbesson: Agreed. This work is meant to be develop only indeed. Will open a follow-up PR once this is in and have @joshmoore check it is not breaking anything.

jburel: you will need to indicate that the existing method will be deprecated before you can remove it
in that case a PR in 5.1 to mention it

sbesson:
Pushed typo fixes.

jburel:
typo: ```objectd```

sbesson:
Pushed as well

rleigh-dundee:
Thanks for the fixes.  Diffing the build log before and after shows a lots of fixed issues in the common component.  The only exception is there are some unfixed ones in LuceneQueryBuilder, though they can be fixed in another PR if that makes more sense of course.

Looks fine to merge.

sbesson:
@rleigh-dundee: thanks. Which command did you use to generate the `LuceneQueryBuilder` javadoc warnings? I spotted there are still some `@return` without strings in this class but they are in private methods which we are currently not considering in the Javadoc (https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-build/65/). I would first clean Javadocs for the public method and clean the build output and then we can come back to private methods.

rleigh-dundee:
@sbesson Nothing fancy.  I ran `script before` then `ant build-default` and then again after the merge with `script after`, and the `diff -u before after` to get a full diff of the entire build.  This showed up LuceneQueryBuilder warnings unchanged in the common component.

jburel:
Thanks for report. 
We can do the rest of the cleaning in a follow-up PR if required.
Merging so can move forward with the cleaning.

PR #4059, opened 2015-08-18
Draw empty histogram if there's no data

dominikl:
Tiny fix for Ticket [12994](http://trac.openmicroscopy.org/ome/ticket/12994).

- Draw a ROI
- Open Intensity view
- Deselect all channels and make sure an empty histogram plot is shown when no channel is selected (before: Insight crashed)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/GraphPane.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/GraphPane.java
@@ -48,6 +48,7 @@ import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 
 //Third-party libraries
+import org.apache.commons.collections.CollectionUtils;
 
 //Application-internal dependencies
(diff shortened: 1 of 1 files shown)


gusferguson:
@dominikl 

Tested with OMERO.insight-5.2.0-m0-443-bf9d1d3-ice35-b64-mac eel user-3 read-only-1

Behaves as expected with redrawing after toggling off channels.
Empty histogram when no channel selected.
No crashes.
Good to merge.

Note: Refresh of histogram after moving or adjusting line can be a bit laggy or not happen at all, but that is entirely another issue - not for this PR.

jburel:
Rebase candidate since it is a bug fix.

jburel:
@dominikl: Can ```data``` be ```null```, if yes, better to use ```CollectionUtils.isEmpty```

dominikl:
Thanks for the hint @jburel . Replaced isEmpty() with the appropriate CollectionUtils method, don't think a re-review is necessary for this.

dominikl:
--rebased-to #4069

PR #4060, opened 2015-08-18
Don't zip and preserve original paths when downloading images

dominikl:
See Trello card [Download multiple files - preserve directories](https://trello.com/c/asMv2nvz/27-download-multiple-files-preserve-directories)

With this PR original files will be downloaded as they are (not zipped like before) and their original folder structure will be preserved. Test: Download some original files and make sure the folder structure is correct (see File Path info dialog, folder structure should match the path in the managed repository).

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
@@ -3077,6 +3077,56 @@ class EditorModel
 	        downloadFiles(file, override);
 	    }
 	}
+	
+    /**
+     * Starts an asynchronous loading; preserving the original folder structure
+     * 
+     * @param path
+     *            The folder where to download the content.
+     * @param override
+     *            Flag indicating to override the existing file if it exists,
+     *            <code>false</code> otherwise.
+     */
+    void downloadOriginal(String path, boolean override) {
+        if (!(refObject instanceof ImageData))
+            return;
+
+        List<ImageData> images = new ArrayList<ImageData>();
+        List<DataObject> l = getSelectedObjects();
+        if (!CollectionUtils.isEmpty(l)) {
+            Iterator<DataObject> i = l.iterator();
+            DataObject o;
+            List<Long> filesetIds = new ArrayList<Long>();
+            long id;
+            ImageData image;
+            while (i.hasNext()) {
+                o = i.next();
+                if (isArchived(o)) {
+                    image = (ImageData) o;
+                    id = image.getFilesetId();
+                    if (id < 0)
+                        images.add(image);
+                    else if (!filesetIds.contains(id)) {
+                        images.add(image);
+                        filesetIds.add(id);
+                    }
+                }
+            }
+        }
+        if (!CollectionUtils.isEmpty(images)) {
+            DownloadArchivedActivityParam p;
+            UserNotifier un = MetadataViewerAgent.getRegistry()
+                    .getUserNotifier();
+            IconManager icons = IconManager.getInstance();
+            Icon icon = icons.getIcon(IconManager.DOWNLOAD_22);
+            SecurityContext ctx = getSecurityContext();
+            p = new DownloadArchivedActivityParam(new File(path), images, icon);
+            p.setOverride(override);
+            p.setZip(false);
+            p.setKeepOriginalPaths(true);
+            un.notifyActivity(ctx, p);
+        }
+    }
 
 	/** 
 	 * Starts an asynchronous call to retrieve disk space information. 
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/DownloadAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/DownloadAction.java
@@ -24,30 +24,22 @@ import java.awt.event.ActionEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.File;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-
 import javax.swing.Action;
 import javax.swing.JFrame;
-import javax.swing.filechooser.FileFilter;
 
 import org.apache.commons.collections.CollectionUtils;
-import org.apache.commons.io.FilenameUtils;
+
 import org.openmicroscopy.shoola.agents.dataBrowser.DataBrowserAgent;
 import org.openmicroscopy.shoola.agents.dataBrowser.IconManager;
 import org.openmicroscopy.shoola.agents.dataBrowser.browser.ImageDisplay;
 import org.openmicroscopy.shoola.agents.dataBrowser.view.DataBrowser;
 import org.openmicroscopy.shoola.agents.events.hiviewer.DownloadEvent;
 import org.openmicroscopy.shoola.env.event.EventBus;
-import org.openmicroscopy.shoola.util.filter.file.ZipFilter;
 import org.openmicroscopy.shoola.util.ui.UIUtilities;
 import org.openmicroscopy.shoola.util.ui.filechooser.FileChooser;
-
 import pojos.DataObject;
-import pojos.FileAnnotationData;
 import pojos.ImageData;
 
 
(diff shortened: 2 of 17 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
```diff
@@ -3077,6 +3077,56 @@ void download(File file, boolean override)
...
+                    if (id < 0)
+                        images.add(image);
+                    else if (!filesetIds.contains(id)) {
+                        images.add(image);
+                        filesetIds.add(id);
+                    }
+                }
+            }
+        }
+        if (!CollectionUtils.isEmpty(images)) {
```
mtbc: just FYI there's a `isNotEmpty`

mtbc:
So does no zipping mean that the transfer of data over the network is uncompressed? (If so, is that okay?)

dominikl:
The files actually weren't zipped before the transfer. Files were downloaded into a temporary folder, then zipped, then the tmp folder deleted. Didn't make much sense anyway.

mtbc:
Goodness! This PR definitely sounds like a good idea then.

gusferguson:
@dominikl 

Tested with OMERO.insight-5.2.0-m0-443-bf9d1d3-ice35-b64-mac eel user-3 read-only-1

File downloads as:
user-3_9/2015-08/18/09-46-01.606/P-TRE_10_R3D_D3D.dv_vol
Paths on server is:
user-3_9/2015-08/18/09-46-01.606/P-TRE_10_R3D_D3D.dv_vol 

Tested using "archived" .dv that had no server path visible, but original file present (example of Version 4.x archived file.
File downloads as:
/home/hudson/for_import_to_howe/archived-imoprt-case/user-3/read-only-1/no_projects/archivedv/very_small.d3d.dv
Path on server is:
"No information"

So this case is covered.

This PR works as expected and as such is good to merge.

However cc. @jburel @pwalczysko - I find this workflow **very** confusing.
Trying to find the files after download is a real mission - particularly when you subsequently download other images from the same dataset as previously downloaded images - or from same project - different dataset - is very difficult to identify and find the image I last downloaded.
I didn't appreciate until I tested this what a nightmare this change will be for existing users to try and work out what has happened to their downloaded files.
Do you think we need a dialog box with this behaviour as an option, but also a "Just dump them all in an 'OMERO Downloads' folder option. i.e. option to revert to previous behaviour minus the zipping.

Screenshot shows what the file hierarchies looks like in the Finder. Quite where the "-9" on the user name comes from I am not sure.

For images with path name:

finder

For image without path name:

finder

dominikl:
Yes, I noticed that as well, one probably expects just the images in the selected download folder and not some strange folder hierarchy, except one specifically chose so. The problem is, where to put this option? A submenu below "Download..." menu item (I think we already rejected this in one of the client meetings)? Another dialog before the folder selection dialog (not very nice imho)? A quite massive dialog with all the possible download options, like @gusferguson already scoped one (which would be quite a big change/lot of work)?

pwalczysko:
@dominikl : Will is going to give some insights how this is done in Web. I just think to follow what he is doing there is reasonable. Basically, no unnecessary creation of any folders ever, when possible then the file is straight under the Downloads folder. If the file structure itself has some subfolders, then these will be preserved, but definitely not ``/user-3_9/read-only-1/.....`` from the original path (or is it path on server ?) type of complicated and not very useful subdirs.

pwalczysko:
cc @will-moore

will-moore:
@dominikl The logic we are using in the web involves ```fileset.getTemplatePrefix()``` to remove this from the full path to the file and just use the relative path to preserve directory structure within a fileset. You can see the code here: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/util.py#L40

gusferguson:
But if you look at the original trello card Douglas said that the requirement was for "the relative paths the same as at upload (in ManagedRepository)."

dominikl:
Ah thanks, of course, I have to remove the part of the internal managed repository folder structure `/user-3/...` from the original file path.

So for example:
```
user-3_9/2015-08/18/10-14-11.704/Fig 4.mvd2 
user-3_9/2015-08/18/10-14-11.704/Data/102.aiix
user-3_9/2015-08/18/10-14-11.704/Data/...
```
Will result in
```
Fig 4.mvd2 
Data/102.aiix
Data/...
```

pwalczysko:
@gusferguson : ``As at upload....`` is what he means I guess. Yes, these are preserved in ManagedRepository, but under all the crud like ``/user-3_9/2015-08/`` type of subfolders which do not bring anything once outside of OMERO. ``As at upload...`` means when the file was uploaded it had some structure BEFORE it came to OMERO. This must be preserved.

dominikl:
But continueing with this example, there's a problem... if you have another image, also with a `Data/` sub-structure, e. .g 
```
Fig 5.mvd2 
Data/397.aiix
Data/...
```

you can't download it to the same destination folder.

dominikl:
* Now really just the image file specific directories are replicated in the download folder, not the whole managed repository structure.
* Multi file images will be stored within a directory with the same name as the main image file, e. g. in above mentioned example this would result in 
```
Fig 5.mvd2/Fig 5.mvd2 
Fig 5.mvd2/Data/397.aiix
Fig 5.mvd2/Data/...
```
OMERO.web uses a serial numbering schema as folder name for this purpose. Is there a specific reason for not using the image name @will-moore ?

will-moore:
@dominikl Image names are not guaranteed to be unique. The extra level of directory with the index name is only used if we have naming conflicts. Up until we have a duplicate file, we don't have the extra level.

dominikl:
Duplicate file/path names are now handled by adding a consecutive number in brackets behind the name, but still as base folder name for a multi file image the name of the image is used. Like @pwalczysko also mentioned, a folder name which is a number only isn't very informative for the user. 

Due to the amount of changes, the PR has to be reviewed again. Would be good to trigger the Download action from all possible locations (Treeviewer, Databrowser and Metadata panel) and make sure they all behave equally.

dominikl:
Unfortunately I can't get rid off this empty file format drop down... hope, this isn't a big problem. I guess it's because the dialog extends from javax.swing.JDialog, so it probably can't be removed anyway.

screen shot 2015-08-20 at 12 28 53

jburel:
It is possible to remove it but that will be tricky
and platform specific

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.0-m0-488-c5c9403-ice35-b66-mac eel user-3 read-only-1

Download works from TreeViewer and General panel, but there is no download of original file available from metadata panel - only original metadata - which works fine.
Images with the same name are saved with (x) appended e.g. Test Image 01.png and Test Image 01(1).png as expected.
Downloading the same image a second time works with an incremented (x) appended.
File .xxx suffixes are preserved.
Empty drop-down not an issue as far as I am concerned.
Good to merge as such.

However, when downloading an image which has had it’s name changed in OMERO - the download is named the original name, not the new name.
Although this is true for both Insight and Web, I think this also going to be pretty confusing for the user.
Can see why it happens, but not sure it should happen.
What are other opinions? @pwalczysko  @will-moore  @jburel  @bramalingam 
Happy to deal with this as a separate RFE so as not to delay this PR.

pwalczysko:
As far as I understand it, the behaviour concerning the changed name in OMERO is actually as expected here. This is in lines with the philosophy: ``We give you back EXACTLY what you gave us, i.e. OMERO never changes a single hair on your original  data.`` I think this philosophy is still in place, the question is of course is it totally wanted by the user like this ? I would imagine that this merges into the areas of "export", like ``export for me whole dataset with images, tags and annotations for the purpose A, B or C...`` which is worked on by @mtbc ? I can imagine that the users might want both (an alternative I mean), i.e. the original name OR the OMERO name. I am afraid the OMERO name option is not available/worked out atm. afaik.

will-moore:
I think it's generally best NOT to rename the original files that users download, since this may break the recognition of that file by native software. Even adding an index to the image name could cause a problem. Same goes for directories that are part of the fileset structure, although directories that we create purely for the download can be named as we like.

mtbc:
I'd not thought that the image names were necessarily anything to do with the actual file names, it's just what we set them to by default on import for want of anything better. I suppose that's most obvious with imports that aren't one-image-per-file. Interesting to think that users may imagine otherwise.

pwalczysko:
@mtbc : It would be rare for users to imagine the state of things as they are in this case. How many users will even be aware that they are importing multiple files (when they think about them as one image for example ?) ? Users will have a very simplistic view of things, file name = OMERO name will be rather the default of thinking than the other way around I am afraid.

pwalczysko:
After discussion with @dominikl and @will-moore it looks like web and insight do more or less the same thing: if multiple files are downloded and there is a naming conflict, new folder is created in both insight and web (and the naming conflict is sorted in the naming of the folder). If only one file is downloaded and the name conflicts, then Insight will add (x) where x is the order number to the name of the file (which is the same as what the OS is doing for Web anyway.)
This seems to be safe with respect to the readability by the original software.

dominikl:
Just to clarify: 
- If an image consists of multiple files, these files will go into a folder with the same name as the original image file. Their names won't be changed. But if this folder already exists, a consecutive number to the folder name will be added. 
- But if an image consists of one file only, a consecutive number will be added to it's filename if a file with the same name already exists (e. g. downloaded twice). An alternative for this would be, to even create a folder for single file images (which imho would be a bit confusing for the user, to get a number of folders in their download folder instead of files directly).

dominikl:

dominikl:
Anything else to add/modify here? If not I could rebase it to dev_5_1.

dominikl:
Well, I guess, it should be ready to merge then?

pwalczysko:
@dominikl : I think so, ready to merge.

sbesson:
See See https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-build/ICE=3.5,jdk=8_LATEST,label=octopus/77/consoleText

```


Standard Doclet version 1.8.0_51
Building tree for all the packages and classes...
/opt/hudson/workspace/OMERO-DEV-merge-build/src/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/Editor.java:297: warning - @param argument "file" is not a parameter name.
/opt/hudson/workspace/OMERO-DEV-merge-build/src/components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroMetadataService.java:770: warning - @return tag cannot be used in method with void return type.
/opt/hudson/workspace/OMERO-DEV-merge-build/src/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ArchivedImageLoader.java:225: warning - @param argument "imageID" is not a parameter name.
/opt/hudson/workspace/OMERO-DEV-merge-build/src/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ArchivedImageLoader.java:225: warning - @param argument "name" is not a parameter name.
```

jburel:
Discussed with @sbesson: I will fix the javadoc issue in a follow up PR

dominikl:
--rebased-to #4135

PR #4063, opened 2015-08-19
Make sure default "Filter" text is selected

dominikl:
Tiny fix for [Ticket 12300](https://trac.openmicroscopy.org/ome/ticket/12300) , has to be tested on Windows.

Try to add tags to an object, which opens the Tags dialog. The "Filter" text field should have the focus, and when start typing straight away the default text "Filter" should get removed (before: typing started after the text).


pwalczysko:
Tested on Win7. Works as expected.

dominikl:
--rebased-to #4077

PR #4075, opened 2015-08-21
Import as crash

dominikl:
Fix for [Ticket 12929](https://trac.openmicroscopy.org/ome/ticket/12929)

Test:
- Log in as an admin user
- Kick off an "import as" process which takes some time, so you'll have time for the next step. Important: Really start the import process, don't just add it to the queue.
- Go back to the import step, and start another "import as" job for another user. But now create a Project and/or Dataset for the import. Without the PR this step was enough to crash Insight, although the crash could happen delayed by some time, depending on how "heavy" the previous started import jobs are.
- Make sure both import jobs are passing like expected.

Note:
The first commit is the actualy bugfix. The second commit just adds a couple of toString methods to some Gateway/Pojo classes, as I found this makes debugging much easier; can remove this commit again if needed.

### DIFF
--- a/components/blitz/src/omero/gateway/SecurityContext.java
+++ b/components/blitz/src/omero/gateway/SecurityContext.java
@@ -168,4 +168,14 @@ public class SecurityContext {
                 && Objects.equal(((SecurityContext) obj).getExperimenter(),
                         this.getExperimenter());
     }
+
+    @Override
+    public String toString() {
+        return "SecurityContext [groupID=" + groupID + ", experimenter="
+                + experimenter + ", sudo=" + sudo + ", serverInformation="
+                + serverInformation + ", compression=" + compression + "]";
+    }
+
+    
+    
 }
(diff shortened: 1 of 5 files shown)


dominikl:
@mtbc noticed a problem creating datasets in the import dialog, might be related to this PR, have to check that.

jburel:
No problem noticed when I tested it yesterday

dominikl:

dominikl:
I don't understand it, I'm sure I tested it and it worked, but now I can't sudo create projects/dataset because I get 
```
2015-08-26 13:43:12,815 INFO  [       ome.services.sessions.SessionBean] (l.Server-3) Handling session exception: 
ome.conditions.SecurityViolation: Group sudo is not permitted for group group1 (gid=3)
```

dominikl:
I'll keep that one excluded for now, probably will close it due to the problem there.

dominikl:
Closing the PR, there must be some deeper problem within the Gateway code (but it's not introduced by the Gateway decoupling PR, as it also happens in the 5.1 line).

PR #4061, opened 2015-08-18
A few minor CLI fixes

ximenesuk:
This PR is a holder for a few minor CLI fixes.

Partially
--rebased-to #4076

The first is: prevent CLI chmod from timing out when downgrading to private. See https://trello.com/c/GZCqaV5D/35-chmod-on-cli-not-usable  This fix uses the same fix as in https://github.com/openmicroscopy/openmicroscopy/pull/3853 To test a group that has some complexity: users, data, etc. should be downgraded to private on the commandline:
```
omero group perms --id XX --perms "rw----"
```

PR #4076, opened 2015-08-21
A few minor CLI fixes (dev)

ximenesuk:
This is a partial rebase of gh-4061 The one difference is that the `--list` option for `delete`m etc., is handled rather than suppressed.

--rebased-from #4061

Testing
-------
```
omero delete --help
```
should not show the --list-details options. However,
```
omero delete --list
omero delete --list-details
```
should both work and give a list of legal targets for this command (and `chgrp`).

The command
```
omero group perms
```
should give a helpful message when the user does not have permission to chnage the group permissions.

Three commits remove the prompt if no description is provided, this brings the CLI in line with other clients. To test:
```
omero tag create
omero tag createset
```
should both prompt for a name if not provided, via the --name option, but not a description if not provided, via the --desc or --description. Modified integration tests should pass.

Two commits fix a bug in `tag list` and list empty tag sets when using the `listsets` subcommand. To test create a couple of tagsets, with one empty.
```
omero tag listsets
```
should include the empty tag set in the list.

### DIFF
--- a/components/blitz/resources/omero/cmd/Graphs.ice
+++ b/components/blitz/resources/omero/cmd/Graphs.ice
@@ -44,17 +44,6 @@ module omero {
          "java:type:java.util.ArrayList<omero.cmd.GraphModify>:java.util.List<omero.cmd.GraphModify>"]
         sequence<GraphModify> GraphModifyList;
 
-        /**
-         *
-         **/
-        ["deprecated:use LegalGraphTargets"]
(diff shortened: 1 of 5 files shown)


sbesson:
Tested using OMERO.py-5.2.0-m0-505-7996a15-ice35-b67. All works as expected. As a minor RFE, we might want to expand the `tag listsets` integration test to create and test empty tagsets as well.

PR #4077, opened 2015-08-21
Make sure default "Filter" text is selected (rebased onto dev_5_1)

dominikl:
This is the same as gh-4063 but rebased onto dev_5_1.

----

Tiny fix for [Ticket 12300](https://trac.openmicroscopy.org/ome/ticket/12300) , has to be tested on Windows.

Try to add tags to an object, which opens the Tags dialog. The "Filter" text field should have the focus, and when start typing straight away the default text "Filter" should get removed (before: typing started after the text).


dominikl:
--rebased-from #4063

gusferguson:
@dominikl 

Tested using OMERO.insight-5.1.3-110-ad69046-ice35-b598-win Windows 8 trout user-3 read-only-1

Behaves as expected.
Good to merge.

PR #4070, opened 2015-08-19
Fixed wrong file count

dominikl:
See [Ticket 12905](https://trac.openmicroscopy.org/ome/ticket/12905)

Test: Check the File path info dialog shows the correct number of files (compare with OMERO.Web)

PR #4078, opened 2015-08-21
Fixed wrong file count (rebased onto dev_5_1)

dominikl:
This is the same as gh-4070 but rebased onto dev_5_1.

----

See [Ticket 12905](https://trac.openmicroscopy.org/ome/ticket/12905)

Test: Check the File path info dialog shows the correct number of files (compare with OMERO.Web)


dominikl:
--rebased-from #4070

mtbc:
Good to merge.

PR #4055, opened 2015-08-17
Rendering settings fixes

dominikl:
Fixes two minor issues with rendering settings:
- Copy the rendering settings from an image. "Paste and Save" them on to other images. Make sure a warning dialog is shown (like it is for other actions e. g. "Set Imported and Save"
- Modify the rendering settings of an image in the preview panel, *do not* save them but copy them. Select the dataset the image belongs to and trigger a "Paste and Save" action. Make sure that the rendering settings of all compatible images are modified accordingly *including* the image from which you copied the rendering settings.
- Do the same again, but now "Paste and Save" onto some other images excluding the image the rendering settings were copied from. Make sure this image does *not* change.

PR #4079, opened 2015-08-21
Rendering settings fixes (rebased onto dev_5_1)

dominikl:
This is the same as gh-4055 but rebased onto dev_5_1.

----

Fixes two minor issues with rendering settings:
- Copy the rendering settings from an image. "Paste and Save" them on to other images. Make sure a warning dialog is shown (like it is for other actions e. g. "Set Imported and Save"
- Modify the rendering settings of an image in the preview panel, *do not* save them but copy them. Select the dataset the image belongs to and trigger a "Paste and Save" action. Make sure that the rendering settings of all compatible images are modified accordingly *including* the image from which you copied the rendering settings.
- Do the same again, but now "Paste and Save" onto some other images excluding the image the rendering settings were copied from. Make sure this image does *not* change.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/view/TreeViewerComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/view/TreeViewerComponent.java
@@ -2660,8 +2660,13 @@ class TreeViewerComponent
 			"you wish to apply the settings to.");
 			return;
 		}
-		model.firePasteRenderingSettings(ids, klass);
-		fireStateChange();
+		
+        MessageBox box = new MessageBox(getUI(), "Save rendering settings",
+                RENDERINGSETTINGS_WARNING);
+        if (box.centerMsgBox() == MessageBox.YES_OPTION) {
+            model.firePasteRenderingSettings(ids, klass);
+            fireStateChange();
+        }
 	}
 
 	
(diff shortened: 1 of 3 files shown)


dominikl:
--rebased-from #4055

gusferguson:
@dominikl 

Tested using OMERO.insight-5.1.3-110-ad69046-ice35-b598-win Windows 8 Java Version 8 build 1.8.0_45-b15 - trout user-3 read-only-1

Testing behaves as expected.
Good to merge from this PR POV.

Issue with Preview pane unrelated to this PR:

When opening Preview pane the image view is tiny - see screenshot.
Have had this before and returns to normal if resizing pane, but this time not returning normal if pane resized and unable to resize image view section of pane - drag handle just does not work.
Full Viewer is fine.

Java 8 issue??

parallels desktop

dominikl:
Interesting. I know you had this problem before, but I couldn't reproduce it. I'll check that with Win8/Java8 and the same image.

dominikl:
@gusferguson  Couldn't reproduce unfortunately. But we should keep that in mind, maybe find a way how to trigger that deterministically.

dominikl:
Same here, don't merge yet. There's still an issue, have to check this PR again.

dominikl:
Fixed the problem. 
To Test: Modify rendering settings in preview panel; copy settings (don't save); select a bunch of images in the databrowser or treeviewer, including the image the settings were copied from; paste and save; make sure the settings are also applied to the image the settings came from.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.1.3-194-9228c3d-ice35-b604-mac_Java6 trout user-3 read-only-1

Behaves as expected.
Good to merge.

PR #4080, opened 2015-08-21
Fix delete notOwned warning

will-moore:
This fixes the display of the warning that you will delete another user's data.

To test, try to delete your own data.
 - You should not see any warning that you will delete anyone else's data.
 - Delete someone else's data (need read-write permissions) and check that you DO see the warning in the delete confirm dialog.

This is fixed independently in jstree_2015 branch.

--no-rebase


mtbc:
Seems to be working just fine, doing the right thing when I target my images or others', in different groups.

In a read-write group if I put someone else's orphaned image in my dataset, then delete the dataset, I delete their image with no warning; that's expected? If so, good to merge.

will-moore:
@mtbc Yes, that's certainly expected since that's always been the behaviour and this PR is only a bug-fix.

mtbc:
:+1:

PR #4082, opened 2015-08-21
Enable 'Show ROIs' after loaded

will-moore:
See http://trac.openmicroscopy.org/ome/ticket/12983

This ensures that the "Show ROIs" button is not enabled until the image has loaded since it fails if used before then (noted on Big Images in particular).

To test:
 - Open a Big Image with ROIs in the main image viewer.
 - Note that the "Show ROIs" link doesn't become enabled until the viewer has loaded.
 - Try to click it before and after loading. Should only work after loading and ROIs should load OK.

This has been cherry-picked to the jstree_2015 branch:
https://github.com/will-moore/openmicroscopy/commit/a4876000c03f8e2414da5e552072bf5ae73d1816

--no-rebase


gusferguson:
@will-moore 

Tested with trout merge on Mac -Safari and Windows 8 - IE 10 user-3 read-only-1

Behaves as expected.
Good to merge.

PR #4083, opened 2015-08-21
ROI tool OutOfMemory

dominikl:
See [Ticket 12995](https://trac.openmicroscopy.org/ome/ticket/12995)

For analyzing ROI shapes a LinkedHashMap was used, which consumed a lot of memory. This PR replaces it with a double array and a list of points. It also removes some unnecessary duplications of data structures which in total significantly reduces the memory usage.

Test: Analyze some ROIs: Make sure the histogram plots and intensity values are still calculated correctly. Analyze a rectangular ROI same size or bigger as the one mentioned in ticket (250x280px) on a multi channel image, make sure Insight doesn't crash and it doesn't take several minutes for the graphs to appear.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/AnalysisStatsWrapper.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/AnalysisStatsWrapper.java
@@ -23,7 +23,6 @@
 package org.openmicroscopy.shoola.agents.measurement.util.model;
 
 //Java imports
-import java.awt.Point;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -33,7 +32,6 @@ import java.util.TreeMap;
 
 //Application-internal dependencies
 import org.openmicroscopy.shoola.env.rnd.roi.AbstractROIShapeStats;
-import org.openmicroscopy.shoola.env.rnd.roi.ROIShapeStats;
 import org.openmicroscopy.shoola.env.rnd.roi.ROIShapeStatsSimple;
 import org.openmicroscopy.shoola.util.ui.UIUtilities;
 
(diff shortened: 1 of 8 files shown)


jburel:
I doubt methods from ```ROIShapeStats``` are used outside, better to mark the public method deprecated and introduced a new one.

dominikl:
Reverted the changes to `ROIShapeStats` in case it's used elsewhere; added a separate double array implementation `ROIShapeStatsSimple` instead.

dominikl:
--exclude  until #4092 is merged

mtbc:
Despite a fairly large ROI, I think I must have a rather better CLS laptop than @mporter-gre does!

The histograms from latest and merge look pretty similar. Without this PR:
latest

with this PR:
merge

mtbc:
Note http://trac.openmicroscopy.org/ome/ticket/13009 which may be relevant to this PR.

mtbc:
Actually, the exception from that ticket seems to be very reproducible from the merge build today, I can't trigger it from the latest build, so I'm guessing this PR's at fault.

dominikl:
Yes, unfortunately the IntensityView needs the x,y pixel coordinates; have to overhaul this PR, will exclude it for now.
--exclude

joshmoore:
Kicked due to travis/hdf5 error.

jburel:
user-1 DV dataset
I have propagated a large rectangle through z
We try to analyse all the shapes in one go. This leads to exception and/or the application is totally not responding
When I finally have a result, I move the z-slider, the shapes are re-analysed
so it is probably better to only analyze the shape displayed and not the full stack

This was not introduce in this PR, depending on the number of changes required it can be done in a follow-up PR

jburel:
Card created for issue described above see https://trello.com/c/3HccWkMv/40-roi-memory-issue

Merging this PR so we can start the migration of the Pojos.

dominikl:
--rebased-to #4137

PR #4084, opened 2015-08-21
Interpolate config option

will-moore:
See http://trac.openmicroscopy.org/ome/ticket/12924

To test:
 - Check in web full image viewer, that by default, interpolation is ON (checkbox is checked and image is not pixelated when zoomed right in).
```
$ bin/omero config set omero.client.viewer.interpolate_pixels False
$ bin/omero admin restart
```
 - Re-open image viewer. Interpolation should be OFF (checkbox unchecked and pixellated image on zoom).

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
@@ -1065,11 +1065,6 @@
       // disable 'Max Intensity' projection for single-Z images
       var disable_intmax = (viewport.loadedImg.rdefs.invertAxis || viewport.getSizes().z < 2);
       $('[name="wblitz-proj"][value=intmax]').prop('disabled', disable_intmax);
-
-      // Turn off interpolation if disabled
-      {% if not interpolate %}
-        viewport.setPixelated(true);
-      {% endif %}
     });
 
     // 'Color' checkbox to left of image
--- a/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
@@ -934,7 +934,7 @@
           </div>
           <div class="odd row">
             <label for="wblitz-interpolate">Interpolate</label>
-            <input id="wblitz-interpolate" type="checkbox" checked/>
+            <input id="wblitz-interpolate" type="checkbox" {% if interpolate %}checked{% endif %}/>
           </div>
             <h1>Current Image</h1>
           <div class="even row">
@@ -1066,6 +1066,10 @@
       var disable_intmax = (viewport.loadedImg.rdefs.invertAxis || viewport.getSizes().z < 2);
       $('[name="wblitz-proj"][value=intmax]').prop('disabled', disable_intmax);
 
+      // Turn off interpolation if disabled
+      {% if not interpolate %}
+        viewport.setPixelated(true);
+      {% endif %}
     });
 
     // 'Color' checkbox to left of image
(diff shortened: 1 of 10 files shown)


etc/omero.properties
```diff
@@ -826,6 +826,9 @@ omero.client.ui.menu.dropdown.everyone=All Members
 # Flag to show/hide all users.
 omero.client.ui.menu.dropdown.everyone.enabled=true
 
+# Client viewers interpolate pixels by default.
+omero.client.interpolate_pixels=true
+
```
aleksandra-tarkowska: As this is viewer specific, could we keep naming convention and rename it to omero.client.viewer.interpolate_pixels

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -1905,6 +1906,10 @@ def full_viewer(request, iid, conn=None, **kwargs):
     """
 
     rid = getImgDetailsFromReq(request)
+
+    interpolate = toBoolean(conn.getConfigService().getConfigValue(
```
aleksandra-tarkowska: This will call server each time you open a viewer. That is very inefficient as it is not image specific. Could you load it once and add it to server exactly as https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/decorators.py#L254

aleksandra-tarkowska: This will call server each time you open a viewer. That is very inefficient as it is not image specific. Could you load it once on login and add it to server exactly like https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/decorators.py#L249

aleksandra-tarkowska: You could name it             request.session['server_settings']['viewer'] =

dominikl:
I guess this is a property Insight also should take into account, right?

jburel:
@dominikl: yes

will-moore:
Following the convention of ```omero.client.viewer.initial_zoom_level``` which is

```
omero.client.viewer.initial_zoom_level
conn.getInitialZoomLevel()
request.session['server_settings']['initial_zoom_level']
```
I now have:
```
omero.client.viewer.interpolate_pixels
conn.getInterpolateSetting()
request.session['server_settings']['interpolate_pixels']
```
And this is loaded just on login.

will-moore:
@dominikl  As queried on the ticket, I think that we should use the interpolation setting in the Preview tab (where we don't show the Interpolate checkbox).

components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
```diff
@@ -1066,6 +1066,10 @@ <h3 class="can-collapse" id="bulk-annotations" style="display: none;">Bulk Annot
       var disable_intmax = (viewport.loadedImg.rdefs.invertAxis || viewport.getSizes().z < 2);
       $('[name="wblitz-proj"][value=intmax]').prop('disabled', disable_intmax);
 
+      // Turn off interpolation if disabled
+      {% if not interpolate %}
+        viewport.setPixelated(true);
```
aleksandra-tarkowska: instead of adding interpolation to template variable it would be preferred to add it to ``viewport.loadedImg``

aleksandra-tarkowska:
travis failed

dominikl:
The interpolation property is taken into account, but if it is set to true, the image is interpolated, but the checkbox is not ticked.

will-moore:
@dominikl Fixed the checkbox not being checked with interpolate: true.

dominikl:
The problem really was, that I copied the command from the PR description which said 'omero.client.interpolate_pixels', which should actually be 'omero.client.viewer.interpolate_pixels'. With the correct name everything works now as expected, good to merge.

will-moore:
--rebased-to #4155

will-moore:
@jburel OK to merge this?

jburel:
as indicated on gh-4155, the small diff between web and insight is acceptable cc @gusferguson 
Merging

PR #4073, opened 2015-08-20
Don't allow collapsing of all task panes

dominikl:
Fix for [Ticket 12997](https://trac.openmicroscopy.org/ome/ticket/12997)

The bug was introduced with #3927 , it allowed collapsing of all task panes, which then were re-layout in the center of the component. With this PR now, it should not be possible anymore to collapse all task panes (like it was before #3927 ).

PR #4085, opened 2015-08-24
Don't allow collapsing of all task panes (rebased onto dev_5_1)

dominikl:
This is the same as gh-4073 but rebased onto dev_5_1.

----

Fix for [Ticket 12997](https://trac.openmicroscopy.org/ome/ticket/12997)

The bug was introduced with #3927 , it allowed collapsing of all task panes, which then were re-layout in the center of the component. With this PR now, it should not be possible anymore to collapse all task panes (like it was before #3927 ).


dominikl:
--rebased-from #4073

mtbc:
Good to merge.

PR #4052, opened 2015-08-17
sudo sessions timeout

dominikl:
See Ticket [12972](http://trac.openmicroscopy.org/ome/ticket/12972)
Previously if a session for another user was created a zero timeout was set, with this PR now the same timeout as the existing session is used.
Test: Code review, check that the import as other user (will initiate a sudo session) still works.

PR #4086, opened 2015-08-24
Set sudo sessions timeout (rebased to dev_5_1)

dominikl:
Same as #4052 , rebased to dev_5_1

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/Connector.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/Connector.java
@@ -859,7 +859,9 @@ class Connector
                 p.name = userName;
                 p.eventType = "Sessions";
                 ISessionPrx prx = entryEncrypted.getSessionService();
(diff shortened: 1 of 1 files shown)


sbesson:
Tested using OMERO.insight-5.1.3-128-a6ee1d9-ice35-b599-mac_Java7+. When importing as, a sudo session is created with a valid timeout

```
 weblitz_test_author | testAnnotationPermissions | 2015-08-25 08:08:43     | createSession  | 10 min. | 290      | 2015-08-25 08:08:48 
```

and expires after 10min of inactivity.

More generally, does anyone have a feeling for a better name for the `createSession` agent? `OMERO.sudo`? Reuse the agent of the initial client? /cc @joshmoore @jburel

joshmoore:
`OMERO.sudo` doesn't sound too bad. If it's worth it, I can look into propagating from the calling session.

dominikl:
--no-rebase (is already in develop #4052 )

dominikl:
--rebased-from #4052

PR #4087, opened 2015-08-24
Interpolate config option

dominikl:
Same as #4084 but for Insight. 
Test: Best to be tested in conjunction with #4084 .

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BrowserModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BrowserModel.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


dominikl:

jburel:
There is no exception thrown when parsing the string to check if it is true or false
The null check is also performed in the Boolean.class

dominikl:
You're right, thanks @jburel 
I found another problem. The interpolation checkbox in the fullviewer is meant to store the setting in the `ViewerPreferences` but this actually doesn't work. Will fix that in this PR. I suppose the user preference - if it exists - should overwrite the setting retrieved from server, right?

jburel:
@dominikl  the pref should overwrite the settings

will-moore:
@dominikl @jburel Do we want to get this and https://github.com/openmicroscopy/openmicroscopy/pull/4084 into 5.1.4? I think it would be nice if we can. Should I wait for this PR to be ready for testing before listing my PR to test?

dominikl:
Yes, I will try to fix the outstanding problem with this PR on Monday. You can list your PR in the meantime, I don't mind. It just would save a bit of work for the reviewer to test both in conjunction, because the steps are the same.

dominikl:
Ready to be re-reviewed.
Test:
- Set interpolation to False ` bin/omero config set omero.client.interpolate_pixels False` followed by `bin/omero admin restart`
- Open a really small image in full viewer and zoom in, so that you can see the pixelation. Make sure the interpolation checkbox in the toolbar is not ticked. But **don't click** on it, because this will create user preferences which override the server preferences and you won't be able to test the next step.
- Set interpolation to True `bin/omero config set omero.client.interpolate_pixels True` followed by `bin/omero admin restart`
- Open the image in full viewer again and zoom in. Make sure the image now is interpolated and the checkbox is ticked.
- Untick the checkbox. Image should now be pixelated again.
- Restart Insight, and open the image again. Make sure the interpolation is still disabled (i. e. your user preference is preferred over the server setting)

dominikl:
--exclude

dominikl:
--rebased-to #4153

jburel:
final check with @dominikl when he gets back.

jburel:
```private boolean interpolationServerProperty = false;```
The default is set to true in web in case of exception.

PR #4088, opened 2015-08-24
Tidy up some CLI file errors

ximenesuk:
See: https://trac.openmicroscopy.org/ome/ticket/12979 for an example of the unhelpful message.

Changing the exception raised allows `argparse` to interpret the cause of the exception and provide helpful output. For background see discussion at http://bugs.python.org/issue9509#msg127007 This change affects three plugins. 

Testing
-------
```
omero export Image:1 --file foo.ome.tif
omero export Image:1 --file foo.ome.tif
```

The second call should display a file exists error message.
```
omero admin cleanse /nonexistent --dry-run
omero admin cleanse file.name --dry-run
```
Should display directory does not exist and not a directory messages respectively.

```
omero config set --file nonexistent.xml
omero config load nonexistent.xml
omero config parse --file nonexistent.xml
```
Should all display file does not exist messages.

/cc @mtbc 

--rebased-to #4091

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -38,6 +38,7 @@ from threading import Lock
 from path import path
 
 from omero_ext.argparse import ArgumentError
+from omero_ext.argparse import ArgumentTypeError
 from omero_ext.argparse import ArgumentParser
 from omero_ext.argparse import FileType
 from omero_ext.argparse import Namespace
@@ -305,7 +306,7 @@ class NewFileType(FileType):
     """
     def __call__(self, s):
         if s != "-" and os.path.exists(s):
-            raise ValueError("File exists: %s" % s)
+            raise ArgumentTypeError("File exists: %s" % s)
         return FileType.__call__(self, s)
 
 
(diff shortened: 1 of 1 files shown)


pwalczysko:
On the first two suggested  commands, although I am getting ``file exists`` the second time I execute the command, I have on the first execution
```
[petr@LS25165 ~/Downloads/OMERO.server-5.1.3-128-a6ee1d9-ice35-b599]$ bin/omero export Image: 23302 --file foo.ome.tif
usage: bin/omero export [-h] -f FILE [-t {TIFF,XML}] [--iterate] [-C]
                        [-s SERVER] [-p PORT] [-g GROUP] [-u USER]
                        obj
bin/omero export: error: unrecognized arguments: 23302
```
where 23302 is a valid ID of an image. What am I missing here ?

pwalczysko:
``Should display directory does not exist and not a directory messages respectively.``
These 2 tests passed as expected.

pwalczysko:
The third test with the nonexistent xml file passed at all 3 steps.

pwalczysko:
In summary, all works as described in the header of the PR, except the problem https://github.com/openmicroscopy/openmicroscopy/pull/4088#issuecomment-134689079

ximenesuk:
Re https://github.com/openmicroscopy/openmicroscopy/pull/4088#issuecomment-134689079 there should be no space between the colon and the id:
```
Image:1
```
not
```
Image: 1
```
With a space the `Image:` and the `1` are seen as two separate arguments, the second of which is then not recognised - this is a `Unix` thing, not just an OMERO thing.

PR #4093, opened 2015-08-25
Migrate more server configuration files to the templates directory

sbesson:
See https://trello.com/c/lHVF6Olb/36-switch-between-prefixed-ports.

This PR reviews server deployment configuration in the same manner as `bin/omero admin jvmcfg` did. Most generally, this PR attempts at differentiating the staging configuration (`config.xml` + `etc/templates`) from the live/running configuration (under `etc`) and clarify the scope of the various `admin` subcommands.

Summary of changes:
- Migrates Ice configuration files previously modified live by `admin ports` and `templates.xml` under `etc/templates`. Renames `generated.xml` as `templates.xml` as we do not have any more conflict name
- adds the logic to regenerate all the configuration files (including JVM settings and ports) and expose it in a `bin/omero admin rewrite` command. Only regenerates templates if no server is running by default.
- modifies the deployment subcommands `omero admin [start|restart|startasync|deploy]`  to use the new `rewrite` subfunction
- modifies the scope of  `bin/omero admin jvmcfg` to display the JVM settings without regenerated the `etc/grid/templates.xml` and `bin/omero admin diagnostics` to read the live memory settings under `etc/grid/templates.xml` instead of recalculating them from `etc/templates/grid/templates.xml` and `etc/grid/config.xml`
- adds new `omero.ports.xxx` properties controlling the ports on which the server applications are deployed and drops `bin/omero admin ports` in favor of this configuration
- extend `test_admin.py` unit tests to cover the new writing stratedy and fix other unit tests requiring `ICE_CONFIG`

This PR will required coupled deployment PR for the testing on the CI server and should be considered as breaking from a deployment perspective for servers using non-standard ports.

Still to do:
- [x] expose `--force-rewrite` option in deployment command

/cc @joshmoore @manics @aleksandra-tarkowska


manics:
AFAIK `admin ports` has always been a bit of a niche option, so instead of deprecating why not just remove it completely (but keep the subcommand so you can print out a suitable error message)?

manics:
The Web application_port port change was done for convenience, since the command is mostly used by us for running multiple CI servers.

manics:
`copycfg` vs `jvmcfg`: is there any reason for not combining them? IIRC `jvmcfg` is regenerated automatically at server startup. Same for `copycfg`?

sbesson:
Answering your points:
- I am all in favor of dropping `change_ports.py` instead of deprecating them. Especially since as the deprecation message says, it would create a temporary configuration that would be overwritten and we don't want to put anyone in this situtation
- will have a go at combine both `jvmcfg` and `copycfg` into a single command (and keep `jvmcfg` as a shortcut for backward compatibility )

sbesson:
@manics: last commits remove the `admin ports` logic (but preserve the deprecated subcommand with the warnings) and migrate all the template copying logic under `admin jvmcfg`.

components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -1646,65 +1687,11 @@ def reindex(self, args, config):
...
-                weblistkey, weblist_from, config.filename))
-        else:
-            config[weblistkey] = weblist_to
-            self.ctx.out('Converted: %s => %s %s in %s' % (
-                weblist_from, weblist_to, weblistkey, config.filename))
+        self.ctx.err(
+            "WARNING: the admin ports subcommand is deprecated. Changes will"
+            " be overwritten the next time the configuration files are"
+            " regenerated. Use the omero.ports.xxx configuration properties"
+            " instead.")
```
manics: It's no longer deprecated, it's completely removed :). `ctx.die` instead?

sbesson: Fair enough. I'll update the message once we all agree with the functionality.

joshmoore: Perhaps instead of `err`'ing, set the appropriate value?

joshmoore:
> is omero admin copycfg a good option or should we look into merging all generation steps (copycfg and jvmcfg)?

What's the utility to the user? `jvmcfg` is intended to show which memory values have been calculated (i.e. there's logic behind it). I think users can largely be oblivious to this rewriting other than the what is shown by `config get` or `admin diagnostics`

> bin/omero admin ports --prefix PREFIX was also modifying the webserver port. Is there any reason omero.ports.prefix should also affect the Web application_port or should we keep both completely separate?

My feeling is that we want to keep this but I could imagine either wanting something like the `localhost` behavior ("only change if not otherwise set") or perhaps a separate named port which also adheres to prefix, etc.

components/tools/OmeroPy/src/omero/plugins/admin.py
manics: My thoughts are the majority of users are unlikely to need this, so might as well keep things simple.

joshmoore:
Re: https://github.com/openmicroscopy/openmicroscopy/pull/4093#discussion_r38078500 :

Discussion points to dropping `ports` but keeping some of the workflow, specifically `--skipcheck` (to be renamed something like `--force-rewrite`) at the admin.py level.

sbesson:
Excluding for now. As shown by https://ci.openmicroscopy.org/view/Breaking/job/OMERO-DEV-breaking-upgrade/, we are consuming `omero admin ports` in `omego` and thus cannot drop it as part of this PR. Will push another commit to deprecate it instead.

sbesson:
See https://ci.openmicroscopy.org/view/Breaking/job/OMERO-DEV-breaking-upgrade/399/

sbesson:
Last commit should implement a `--force-rewrite` argument for all commands making use of `self.status` (except `bin/omero status` itself, although it we could if it is sensible) allowing to rewrite the generated configuration *before* the server status is checked e.g.

```
sbesson@vpn2-086:dist (cfg_templates) $ bin/omero admin start
No descriptor given. Using etc/grid/default.xml
Waiting on startup. Use CTRL-C to exit
sbesson@vpn2-086:dist (cfg_templates) $ bin/omero config set omero.ports.prefix 1
sbesson@vpn2-086:dist (cfg_templates) $ bin/omero admin stop --force-rewrite
Server not running
sbesson@vpn2-086:dist (cfg_templates) $ bin/omero admin stop 
Server not running
sbesson@vpn2-086:dist (cfg_templates) $ bin/omero config set omero.ports.prefix
sbesson@vpn2-086:dist (cfg_templates) $ bin/omero admin restart --force-rewrite 
Waiting on shutdown. Use CTRL-C to exit
.No descriptor given. Using etc/grid/default.xml
Waiting on startup. Use CTRL-C to exit
sbesson@vpn2-086:dist (cfg_templates) $ 
````

manics:
--rebased-to #4225

sbesson:
@joshmoore: if no veto from your side, I'll move this out of breaking for a regular review tomorrow.

joshmoore:
Holding off on any naming discussions, I'm ok with taking this out of breaking.

sbesson:
Note one more commit will be added to this PR to fix https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-matlab/.

ximenesuk:
A couple of quick comments on (re)moved files.

Since `etc/grid` in the source tree is now effectively empty should this an its outdated `README` be removed? On the subject of the `README`, this is now well defunct, e g. `see http://trac.openmicroscopy.org.uk/ome/wiki/OmeroGrid` but is still copied through to `dist/etc/grid`. Could an updated `README` be added under `etc/templates/grid`?

Should the auto-genenerated `templates.xml` have a comment added at the top inline with that in `config.xml`?

components/tools/OmeroPy/src/omero/plugins/admin.py
ximenesuk: At the moment `omero admin ports --help` returns the full help, no mention of deprecation/removal. So whatever is decided the help also needs to be changed/removed.

ximenesuk:
Well, I've not managed to break anything!

sbesson:
@ximenesuk: pushed.

sbesson:
Barring vetoes from @joshmoore or @manics, I would suggest we get this merged now as I have other pending changes on top of this branch.

manics:
:+1: so I can add my changes too

PR #4094, opened 2015-08-25
Add tag set test

mtbc:
In investigating http://trac.openmicroscopy.org/ome/ticket/12998 @ximenesuk uncovered an issue with how the server processes child options. This PR adjusts the behavior and expands an integration test to cover the case. Testing: both https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.delete/AnnotationDeleteTest/ and @ximenesuk should be happy.

--rebased-to #4098

### DIFF
--- a/components/blitz/src/omero/cmd/graphs/ChildOptionsPolicy.java
+++ b/components/blitz/src/omero/cmd/graphs/ChildOptionsPolicy.java
@@ -105,9 +105,8 @@ public class ChildOptionsPolicy {
 
             @Override
             protected boolean isAdjustedBeforeReview(Details object) {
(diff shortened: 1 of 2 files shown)


mtbc:
Intended child option behaviors are documented at http://downloads.openmicroscopy.org/omero/5.1.3/api/slice2html/omero/cmd/graphs.html.

ximenesuk:
Everything looks good here. Tested locally and with what is now gh-4097 which amends and adds to Python tests.

The one thing this has thrown up for me is the potential complexity of using include and exclude options at the same time. It may be that the CLI docs could be improved in this regard - once I understand the complexity myself!

Good to merge.

PR #4089, opened 2015-08-24
add integration test for tag set deletion

mtbc:
http://trac.openmicroscopy.org/ome/ticket/12998 made me wonder if there is some problem with tag set deletion. Integration test coverage is poor so this PR adds a relevant test whose result one finds at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.delete/AnnotationDeleteTest/.

--no-rebase

PR #4098, opened 2015-08-26
have child options ignore parents (rebased from develop)

mtbc:
In graph operations child options previously had a bug such that they could affect how parents were treated. Intended child option behaviors are documented at http://downloads.openmicroscopy.org/omero/5.1.3/api/slice2html/omero/cmd/graphs.html. This PR fixes the bug and adds integration test coverage accordingly: make sure that the test looks sensible and check that https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-java/lastCompletedBuild/testngreports/integration.delete/AnnotationDeleteTest/ passes.

--rebased-from #4089
--rebased-from #4094

### DIFF
--- a/components/blitz/src/omero/cmd/graphs/ChildOptionsPolicy.java
+++ b/components/blitz/src/omero/cmd/graphs/ChildOptionsPolicy.java
@@ -105,9 +105,8 @@ public class ChildOptionsPolicy {
 
             @Override
             protected boolean isAdjustedBeforeReview(Details object) {
-                if (object.action == GraphPolicy.Action.EXCLUDE &&
-                        object.orphan != GraphPolicy.Orphan.IS_LAST && object.orphan != GraphPolicy.Orphan.IS_NOT_LAST) {
-                    /* the model object is [E]{ir} */
+                if (object.action == GraphPolicy.Action.EXCLUDE && object.orphan == GraphPolicy.Orphan.RELEVANT) {
+                    /* the model object is [E]{r} */
                     for (final ChildOptionI childOption : childOptions) {
                         final Boolean isIncludeVerdict = childOption.isIncludeType(object.subject.getClass());
                         if (isIncludeVerdict == Boolean.TRUE && (requiredPermissions == null ||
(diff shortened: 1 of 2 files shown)


ximenesuk:
A CLI scenario that tests this is to create two tag sets with a shared tag:
```
$ omero tag create --name "t1"
TagAnnotation:1
$ omero tag create --name "t2"
TagAnnotation:2
$ omero tag create --name "t3"
TagAnnotation:3
$ omero tag createset --name "ts1" --tag 1 2
TagAnnotation:4
$ omero tag createset --name "ts2" --tag 2 3
TagAnnotation:5
$ omero tag list
+- 4:'ts1'
|\
| +- 2:'t2'
| +- 1:'t1'

+- 5:'ts2'
|\
| +- 3:'t3'
| +- 2:'t2'
$ omero delete TagAnnotation:4 --include TagAnnotation --report
omero.cmd.Delete2 TagAnnotation 4... ok
Steps: 6
Elapsed time: 0.638 secs.
Flags: []
Deleted objects
  AnnotationAnnotationLink:1-3
  TagAnnotation:1,2,4
```
ie just one tag set should be deleted with its contents. Before this PR both sets would be deleted.

dominikl:
Code looks fine. Integration tests passed. Ran through CLI example, worked as well. **Good to merge.** I wonder what Insight is doing, looks like it does something in between, include tags but not the ones who are also part of other tagsets.

mtbc:
That's probably what you'd get without any child options.

dominikl:
Shouldn't the behavior without option be, that just the tagset gets deleted, but not the tags?

mtbc:
Since OMERO 4.x I think, if you orphan a tag by deleting something, you also delete the tag. (The tags in a tagset are actually annotations on the tagset!) To prevent the orphaned tags from being deleted, you can include a child option with `excludeType={'TagAnnotation'}`.

mtbc:
Admittedly, the implementation of tagsets is presumably what was easiest to do without changing the model. There's also plenty of scope for a larger UX rethink of how all this works, as I am sure that there are plenty of subtle behaviors that aren't ideal; there are many cases to consider.

joshmoore:
@dominikl : anything else?

dominikl:
@joshmoore No, the PR is absolutely fine! Was just wondering about the delete behavior in Insight.

PR #4097, opened 2015-08-26
Add and amend delete tests

ximenesuk:
This PR adds some tag set delete tests to reflect the behavious now expected by gh-4094 following the bug fix there. It also fixes and renames a test which should have failed but was passing due to the same issue, see https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/61/testReport/OmeroPy.test.integration.clitest.test_delete/TestDelete/testIncludeExclude/ Finally, a couple of names and comments have been changed at the same time just to make the tests' intentions clearer.

The integration tests should pass.

--depends-on #4094 

--rebased-to #4100

PR #4100, opened 2015-08-26
Add and amend delete tests (dev 5.1)

ximenesuk:
See gh-4097 the Python delete tests should pass with gh-4098 in place.

--depends-on #4098

--rebased-from #4097


dominikl:
Code looks reasonable, integration tests pass. Good to merge.

PR #4065, opened 2015-08-19
Data loading

jburel:
This PR fixes the loading issues of "orphaned" datasets or plates.
Some datasets (resp. plates) were returned as orphaned even if they were linked to a project (resp. screen)
This was the cases when
 * user 1 has a project p1 and add dataset d2. d2 is owned by user-2
 * user-2 logs in and d2 was returned as "orphaned" since it was not linked to any project owned by p1
See https://trello.com/c/6IztFNAq/29-orphaned-dataset-in-another-s-project

Tests have been added to cover that use case and all integration tests (PojosService) should remain green

cc @will-moore @pwalczysko

PR #4105, opened 2015-08-27
Data loading (rebased onto dev_5_1)

jburel:
This is the same as gh-4065 but rebased onto dev_5_1.

----

This PR fixes the loading issues of "orphaned" datasets or plates.
Some datasets (resp. plates) were returned as orphaned even if they were linked to a project (resp. screen)
This was the cases when
 * user 1 has a project p1 and add dataset d2. d2 is owned by user-2
 * user-2 logs in and d2 was returned as "orphaned" since it was not linked to any project owned by p1
See https://trello.com/c/6IztFNAq/29-orphaned-dataset-in-another-s-project

Tests have been added to cover that use case and all integration tests (PojosService) should remain green

cc @will-moore @pwalczysko


jburel:
--rebased-from #4065

dominikl:

PR #4106, opened 2015-08-28
add ROIs to OME-TIFF export

mtbc:
Try exporting then importing an image with ROIs and make sure that the ROIs are preserved. Of course, this preserves only the properties defined in `ROI.xsd`. Also check that https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ExporterTest/ passes.

Still does not work for mask ROIs, sorry @imunro.

--no-rebase

### DIFF
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -75,7 +75,6 @@ import ome.formats.model.ReferenceProcessor;
 import ome.formats.model.ShapeProcessor;
 import ome.formats.model.TargetProcessor;
 import ome.formats.model.WellProcessor;
-import ome.services.blitz.repo.ManagedImportRequestI;
 import ome.units.quantity.ElectricPotential;
 import ome.units.quantity.Frequency;
 import ome.units.quantity.Length;
@@ -221,7 +220,6 @@ import omero.sys.EventContext;
 import omero.sys.ParametersI;
 import omero.util.IceMapper;
 
-import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -1037,10 +1035,7 @@ public class OMEROMetadataStoreClient
      */
     public RInt toRType(Color value)
     {
-        java.awt.Color javaColor = new java.awt.Color(
-                value.getRed(), value.getGreen(), value.getBlue(),
-                value.getAlpha());
-        return toRType(javaColor.getRGB());
+        return toRType(value.getValue());
     }
 
     //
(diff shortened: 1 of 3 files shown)


jburel:
@mtbc: sorry I didn't have time with the various meetings to review it.
I will do that tomorrow am

jburel:
one think to review, outside the scope of this PR is the stylesheets to downgrade to be sure that the ROIs can be preserved during the process

jburel:
Tested the following:
 * Export image with a roi with shapes propagated across z
 * Export image with several rois
 * Export image with rois and annotation on shape

The tests passed.
Something to consider now, ability to pass a parameter to export or not the rois
we now have images with a large number of images so the user should be able to decide to add them or not.

I export to a previous schema (2013), The rois are exported but ref for annotation to the wrong schema
This is outside the scope of this PR and will need investigation

jburel:
Another parameter that we could consider passing will be the compression used for data

mtbc:
The annotation schema version bug is interesting. I wonder who might know about that. Definitely worth noting.

For this PR, any more testing required?

joshmoore:
> one think to review, outside the scope of this PR is the stylesheets to downgrade to be sure that the ROIs can be preserved during the process

https://trello.com/c/CQMxBS71/21-downgrading-rois

> Something to consider now, ability to pass a parameter to export or not the rois

https://trello.com/c/lKfhVnjW/22-skip-rois-in-export

> The rois are exported but ref for annotation to the wrong schema

https://trello.com/c/83il07Mv/23-annotation-ref-version

> Another parameter that we could consider passing will be the compression used for data

That's probably more along the lines of https://trello.com/c/d2OqpHCq/14-export-import-decide-api

Anything else?

jburel:
@mtbc: I think I have an idea for the schema issue and stylesheet
Nothing I can think of for now
but all the points can be addressed in follow up PR
certainly a good start
I have not yet tested annotation on ROI, only on shape

jburel:
for annotation on shape:
```
<Shape>....
<AnnotationRef xmlns="http://www.openmicroscopy.org/Schemas/SA/2015-01" ID="urn:lsid:export.openmicroscopy.org:Annotation:73e8e6ff-d6fe-4c12-8397-3af0d1d901d2_12801:219647"/></Shape>
```

for annotation on roi:

```
</Shape></Union><AnnotationRef xmlns="http://www.openmicroscopy.org/Schemas/SA/2015-01" ID="urn:lsid:export.openmicroscopy.org:Annotation:73e8e6ff-d6fe-4c12-8397-3af0d1d901d2_12801:219647"/></ROI>

````

Both options are working

jburel:
happy to have it merged when we decide

joshmoore:
:+1: Merging to allow schema changes to be built on top.

PR #4111, opened 2015-08-31
OMERO.matlab: add group support (rebased onto dev_5_1)

sbesson:
This is the same as gh-3839 but rebased onto dev_5_1.

----

Following separate discussions with @dpwrussell and @chris-allan, this PR starts adding cross-group support to OMERO.matlab. While `AdminService.setSecurityContext()` can be used to achieve this, it has the downside of potentially changing the default group of the current user. The proper way to do cross-group operations is to pass a map to ice with `omero.group` set.

Most of the relevant functions in OMERO.matlab now support a `group` parameter/key value allowing to specify the `id` of the group (or `-1` for cross-group queries).

The signature of the getter methods with an explicit identifer is modified to be cross-group and cross-owner by default (BREAKING change). To restore the previous current behavior (current session owner, current group), the `owner` and `group` parameter/key values can be passed explicitly

```
getProjects(session, projectIds); % returns all projects of identifier ids
getProjects(session, projectIds, 'owner', ownerId, 'group', 'groupId`); % returns all projects of identifier ids owned by user ownerId in group groupId
```

Most writing functions are also modified to support a group argument.

### DIFF
--- a/components/tools/OmeroM/src/annotations/getAnnotations.m
+++ b/components/tools/OmeroM/src/annotations/getAnnotations.m
@@ -1,4 +1,4 @@
-function annotations = getAnnotations(session, ids, type)
+function annotations = getAnnotations(session, ids, type, varargin)
 % GETANNOTATIONS Retrieve annotations from a given type from the OMERO server
 %
 %   annotations = getAnnotations(session, ids, type) returns all the 
@@ -7,6 +7,8 @@ function annotations = getAnnotations(session, ids, type)
 %   Examples:
 %
 %      annotations = getAnnotations(session, ids, type);
+%      annotations = getAnnotations(session, ids, type, 'group', groupId);
+%        returns annoations from the input group specfied by groupId.
 %
 % See also: GETANNOTATIONTYPES, GETDOUBLEANNOTATIONS, GETCOMMENTANNOTATIONS,
 % GETFILEANNOTATIONS, GETLONGANNOTATIONS, GETTAGANNOTATIONS,
@@ -34,17 +36,26 @@ annotationTypes = getAnnotationTypes();
 annotationNames = {annotationTypes.name};
 ip = inputParser;
 ip.addRequired('session');
-ip.addRequired('ids', @(x) isvector(x) || isempty(x));
+ip.addRequired('ids', @(x) isvector(x) || ~isempty(x));
 ip.addRequired('type', @(x) ischar(x) && ismember(x, annotationNames));
-ip.parse(session, ids, type);
+ip.addParameter('group', [], @(x) isscalar(x) && isnumeric(x));
+ip.parse(session, ids, type, varargin{:});
+
 annotationType = annotationTypes(strcmp(type, annotationNames));
 
 % Create list of annotations identifiers to load
 ids = toJavaList(ip.Results.ids, 'java.lang.Long');
 
 % Create container service to load annotations
+context = java.util.HashMap;
+if ~isempty(ip.Results.group)
+    context.put(...
+        'omero.group', java.lang.String(num2str(ip.Results.group)));
+else
+    context.put('omero.group', '-1');
+end
 service = session.getMetadataService();
-annotationList = service.loadAnnotation(ids);
+annotationList = service.loadAnnotation(ids, context);
 
 % Filter annotation list by annotation class
 for i = annotationList.size - 1 : -1 : 0
--- a/components/tools/OmeroM/src/annotations/getCommentAnnotations.m
+++ b/components/tools/OmeroM/src/annotations/getCommentAnnotations.m
@@ -1,4 +1,4 @@
-function cas = getCommentAnnotations(session, ids)
+function cas = getCommentAnnotations(session, ids, varargin)
 % GETCOMMENTANNOTATIONS Retrieve comment annotations from the OMERO server
 %
 %   cas = getCommentAnnotations(session, ids) returns all the comment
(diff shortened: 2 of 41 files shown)


sbesson:
--rebased-from #3839

sbesson:
Commit fef05f8 should revert the semantic breaking change when retrieving objects by identifiers on the 5.1.x line keeping query limited to the current session group by default (cross-group queries can still be performed by passing `group, -1` to the function).

imunro:
I may have missed something, there's a lot of code here - but there doesn't seem anything in the Matlab interface to get a list of the current users groups ? Separate PR ?

sbesson:
@imunro: the AdminService should provide methods for retrieving the groups which the user is member or owner of - see the changes to `ConnectToOmero` in https://github.com/sbesson/openmicroscopy/commit/a52ff492400210a61025bbc7c6f664c8be190199#diff-4d09531e048f73ba5a7ef23b40454ce4R62

imunro:
@sbesson I just thought that this should be neatly packaged into OMERO.Matllab as you have with everything else.

sbesson:
Understood. Then I think it is definitely outside the scope of this large PR. Feel free to create a card or even have a first go as implementing such a function.

sbesson:
Have you have time to test the features added by this PR or should I relist for tomorrow?

imunro:
I'm working through them - slowly.

imunro:
@sbesson I can't work out what the changes to the functions in Annotations are all about.
Why do you need to pass  a group if you're already passing a set of IDs.
If I call say getDatasetFileAnnotations with a dataset id that isn't in the group I select then what's supposed to happen?

sbesson:
@imunro: at the moment, I would expect any getter method passing some object ID(s) belonging to a different group than the session group will return nothing because all the queries are done in the context of the session group by default. This should be the behavior prior to this PR as well.

If we did not include fef05f8 in this PR, this behavior would be switched to querying objects across groups if identifiers are passed.

imunro:
That would seem more sensible to me. If you know the ids  passing the group seems to add only uneccessary complexity

imunro:
The ceateProject, createDataset etc functions OTOH do need a group don't they?

sbesson:
Good point. This is likely a follow up. Right now they would be created in the context of the current session group as well but we could extend `createObject` to support (`group`, groupId).

imunro:
ok not in this PR then.

imunro:
Tested so far w/out problems: getDatsets, getProjects, getImages. Lots left to do.

sbesson:
As explained in https://trello.com/c/gSKqCAe2/49-omero-matlab-group-support, force pushed the extra fef05f8 commit away from this rebased PR.

sbesson:
See https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/429/consoleText for the latest version of the training examples

imunro:
Not sure if this is expected behaviour but passing a non-existent group gives a java error:
>> datasets = getDatasets(session,'group',904)
Error using getObjects (line 114)
Java exception occurred:
omero.ValidationException
    serverStackTrace = "ome.conditions.ValidationException: No row with the given identifier exists:

imunro:
It is also possible to pass any negative number as a group and get the -1 behaviour.
Should we check for this ?

sbesson:
@imunro the `ValidationException` is certainly expected and could be wrapped in a better error message in `getObjects`. We can add a item for doing that after this PR. As for negative numbers lower than -1, assuming this API allows this, I think doing an additional check is overkill at this point.

sbesson:
Note RFEs added to https://trello.com/c/gSKqCAe2/49-omero-matlab-group-support

imunro:
Currently passing a group explicitly overrides the default "all groups" behaviour when IDs are passed. ie
images = getImages(session,182048,'group',905) where this image is not in this group returns [].
This makes sense but strikes me as a recipe for a user becoming confused.
Needs careful explanation in the docs.

sbesson:
@imunro is this not covered as part of https://github.com/openmicroscopy/ome-documentation/pull/1273/files

```
% Returns all images owned by the session user in the specified group
images = getImages(session, 'group', groupId);
% Returns all images with the input identifiers in the specified group
images = getImages(session, ids, 'group', groupId);
% Returns all images owned by the session user across groups
images = getImages(session, 'group', -1);
```

How would you change the comment to make it clearer?

imunro:
@sbesson I can't improve on that but it isn't what I'm seeing in the header of the Matlab file is that not merged as yet?

sbesson:
No I was referring to the Sphinx documenation as the docs. If we are talking of the MATLAB header, they
contain the code examples but not the comments above although the `'group'` usage should be described above.

imunro:
I have re-tested getProjects, getDatasets, getImages, get DatasetFileAnnotations and getImageFileAnnotations and all seem to work as required. 
 I would suggest, however that for those functions  where ids are required rather than optional e'g getDatasetFileAnnotations, the extra 'group'  options adds no extra functionality that I can see and considerable potential for confusion.   Unless I'm missing something then I would suggest this be removed at some future point.

jburel:
The group follows the same approach taken in other gateway i.e. the group context has to be always specified in the java gateway for example.
It adds potentially a bit of complexity, we can certainly improve when we have more usage/feedback

Merging

sbesson:
--rebased-to https://github.com/openmicroscopy/openmicroscopy/pull/4136 last 2 commits

PR #4115, opened 2015-09-01
a batch of tidying up headers and imports in Java files

mtbc:
This is the first in what is intended to be a series of PRs that minorly tidy the comments and imports at the start of Java source files. Nothing important should be lost and the code should still build fine.

--no-rebase

### DIFF
--- a/components/blitz/src/ome/services/blitz/impl/PixelsI.java
+++ b/components/blitz/src/ome/services/blitz/impl/PixelsI.java
@@ -1,13 +1,10 @@
 /*
- *   $Id$
- *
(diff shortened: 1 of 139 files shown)


sbesson:
Looks sensible to me and all jobs are green (unit tests in Travis as well as https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-java/). Happy to merge right now as this may conflict with other PRs. @joshmoore @jburel ?

joshmoore:
:thumbsup: though I do wonder if we have any screenshots which need "occurrences" updated :smile:

PR #4126, opened 2015-09-03
Add ability to run scripts against fa's from OMERO.web (rebased onto dev_5_1)

sbesson:
This is the same as gh-3993 but rebased onto dev_5_1.

----

This PR adds the ability to run scripts against file annotations from the user interface which should pave the way for more interesting scripting workflows that run against uploaded and attached files.

Various small pieces of refactoring to allow the script UI to be popped up and interacted with by other parts of the user interface are also included.

Both single object selection,

screen shot 2015-07-27 at 13 49 09

and multiple object selection,

screen shot 2015-07-27 at 13 49 18

are supported. The checkboxes allow for the selection of the file annotation and for convenience a script UI activation button has been added to the attachment section of the righthand panel.

I don't doubt that some robot framework tests will break with this work. A discussion with @will-moore about how to best test the functionality will be forthcoming.

Populate metadata workflows and testing as worked on by @joshmoore and @pwalczysko can also be updated with scripts from glencoesoftware/omero-user-scripts to functionally test the work.


sbesson:
--rebased-from #3993

joshmoore:
:thumbsup:

PR #4129, opened 2015-09-03
Add ability to run scripts against fa's from OMERO.Insight (rebased onto dev_5_1)

jburel:
This is the same as gh-4051 but rebased onto dev_5_1.

----

Same as #3993 just for Insight

Test:
- Install @emilroz  's script which uses file annotations: https://github.com/emilroz/omero-user-scripts/blob/populate_metadata_fa_additions/util_scripts/Populate_Metadata.py 
- Add some file annotations to a plate
- Select the plate and some file annotations and run the script, which should open the script dialog
- Make sure the Plate ID and the the file annotations IDs are set in the dialog
- Run the script. Will fail, but in 'Info' in the activities popup you should see, that the parameters were passed correctly to the script.
- Additional test: Run some other scripts, to make sure the changes didn't break the script dialog for other scripts.


jburel:
--rebased-from #4051

jburel:
looks good.
Merging

PR #4131, opened 2015-09-04
Formatting timestamps

will-moore:
This fixes a bug in python 2.6 when comparing Decimal with floats and also other bugs in the calculation of times for formatting.
https://trac.openmicroscopy.org/ome/ticket/13016

To test, user trout (python 2.6).
 - Find Image 24551 (user-3), which has timepoints every 5 minutes.
 - Look at Delta T times, Under POL channel on Acquisition tab.
 - Should see human readable time values.
 - Check that with timestamps over an hour, that minutes are shown correctly (don't go over 59!)
 - We also round() the numbers before truncating to integers for display, but the only way to see the original values is to download the log file.

Before:
screen shot 2015-09-04 at 10 51 10

After:
screen shot 2015-09-04 at 11 11 06

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/templatetags/common_filters.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/templatetags/common_filters.py
@@ -263,11 +263,11 @@ def timeformat(value):
         return u'%d\u00A0\u00B5s' % (round(value * 1000 * 1000))
     elif value < 1:
         return u'%d\u00A0ms' % (round(value * 1000))
-    elif value < 60:
+    elif round(value) < 60:
         # Round and format seconds to one decimal place
         value = round(value * 10) / 10
         return u'%0.1f\u00A0s' % value
-    elif value < 60 * 60:
+    elif round(value) < 60 * 60:
         value = round(value)        # Avoids '1min 60s'
         return u'%d\u00A0min\u00A0%d\u00A0s' % (value / 60, value % 60)
     else:
--- a/components/tools/OmeroWeb/omeroweb/webgateway/templatetags/common_filters.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/templatetags/common_filters.py
@@ -256,7 +256,7 @@ def timeformat(value):
             value = Decimal(force_unicode(float(value)))
         except (ValueError, InvalidOperation, TypeError, UnicodeEncodeError):
             return u'%s s' % str(value)
-    # String formatting shows integer values for all, so we round() for accuracy
+    # Formatting shows integer values for all, so we round() for accuracy
     if value == 0:
         return u'%d\u00A0s' % value
     if value < Decimal("0.001"):
@@ -270,4 +270,5 @@ def timeformat(value):
         return u'%d\u00A0min\u00A0%d\u00A0s' % (value / 60, value % 60)
     else:
         value = round(value)        # Avoids '1h 60min'
-        return u'%d\u00A0h\u00A0%d\u00A0min' % (value / 3600, round((value % 3600)/60))
+        return u'%d\u00A0h\u00A0%d\u00A0min' % (value / 3600,
+                                                round((value % 3600)/60))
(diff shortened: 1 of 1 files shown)


mporter-gre:
That looks great! Splitting the timestamps into "x min y s" is really helpful

will-moore:
@dominikl @jburel Maybe formatting in this style should also be used in Insight, since seconds only is less human readable:

screen shot 2015-09-07 at 14 11 19

jburel:
@dominikl: do you want to have a look at insight?

dominikl:
Yes, I'll go for it.

ximenesuk:
See above, should times such as `40 min 0 s` be displayed or should it be truncated to `45 min`?

On the flipside while testing the Insight PRs I noticed this difference, for `Image:9236` (user-3),

image

The Delta T here is shown as 16s, in Insight it is 16s 208ms which is a significant difference in information. See: https://github.com/openmicroscopy/openmicroscopy/pull/4149#issuecomment-139133425

will-moore:
@ximenesuk I was about to go ahead with truncating e.g. ``` 40 min 0 s``` to ``` 40 min``` but then I wonder if the ```0s``` gives users clarity that we are not just truncating to the nearest minute but are actually rounding to the nearest second?

will-moore:
Format seconds to 1 decimal place:

screen shot 2015-09-15 at 22 14 11

ximenesuk:
This looks good. For seconds, I think one decimal place is sufficient.

As for truncating minutes to show `0s` I'm not sure. I think not having the seconds showing when they are zero (or round to zero) is sufficient. This display is only a human-readable representation of the underlying data and as such I guess readers should understand the limitations?

@pwalczysko @gusferguson any thoughts?

/cc @dominikl

will-moore:
@mporter-gre Do you have any suggestions here? My concern is that if we show times such as 0s, 5 min, 10 min, etc that users might think that we are rounding to nearest minute, e.g. 5 min 20 secs -> 5 min.

gusferguson:
In my opinion any 0s should be shown - even to the decimal point of seconds - 5 min 0.0 secs. Scientists would find this normal in my view.

mporter-gre:
As a personal preference, I would like to see a consistent format across time points even if there are 0 values. Not because I would assume rounding, it's just easier to read if the pattern of digits is always the same.

pwalczysko:
Agree with @mporter-gre - Re : 
> ...readers should understand the limitations?
No, they do not.

will-moore:
Examples of what we have now are:
```
0 s
200 ms
2.9 s
1 min 0 s
55 min 0 s
60 min 0 s (oops - should be 1 h 0 min)
1 h 5 min
12 h 45 min
```
(using images 9253 and 24551, user-3 trout as examples).
I think this is a reasonable level of rounding for each time (don't need fraction of a second if your time is 1 minute or over). Just need to fix ```60 min 0 s```

ximenesuk:
@pwalczysko I wrote **should** - not that they do.

We cannot display every number to its complete accuracy in this display - that was the impetus for these PRs in the first place. Once any rounding or truncating takes place there is an onus on the reader to interpret the numbers accordingly - we could document this somehere but then users don't always read documentation either!

Supplying a value such as `5m 0s` does not give any more confidence that the value is exactly `5m` any more than it might be `5m 0.4s`. But I would agree with @mporter-gre that  consistency is important and that means accuracy is inevitably lost.

manics:
`m` or `min`?

will-moore:
@manics I don't know - No one's complained about ```min``` so far!

ximenesuk:
This looks fine to me, I even found an example of `1min 0s` to check!

Any other comments from the gallery or should this get merged and then @dominikl can make similar changes in his two Insight PRs.

jburel:
@ximenesuk: could you indicate the example you used so we can validate using insight

ximenesuk:
@jburel trout user-2 Image:9532 t=13

will-moore:
--rebased-to #4220

PR #4135, opened 2015-09-04
Don't zip and preserve original paths when downloading images (rebased onto dev_5_1)

dominikl:
This is the same as gh-4060 but rebased onto dev_5_1.

----

See Trello card [Download multiple files - preserve directories](https://trello.com/c/asMv2nvz/27-download-multiple-files-preserve-directories)

With this PR original files will be downloaded as they are (not zipped like before) and their original folder structure will be preserved. Test: Download some original files and make sure the folder structure is correct (see File Path info dialog, folder structure should match the path in the managed repository).

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/Editor.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/Editor.java
@@ -288,6 +288,17 @@ public interface Editor
 	 */
 	public void download(File file, boolean override);
 
+    /**
+     * Downloads the archived files, preserving the original folder structure
+     * 
+     * @param file
+     *            The folder where to download the content.
+     * @param override
+     *            Flag indicating to override the existing file if it exists,
+     *            <code>false</code> otherwise.
+     */
+    public void downloadOriginal(String path, boolean override);
+	
 	/**
 	 * Sets the parent of the root object. This will be taken into account
 	 * only if the root is a well sample.
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/DownloadAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/DownloadAction.java
@@ -24,30 +24,22 @@ import java.awt.event.ActionEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.File;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-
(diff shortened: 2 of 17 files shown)


dominikl:
--rebased-from #4060

components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/Editor.java
```diff
@@ -288,6 +288,17 @@ public void setChannelAcquisitionData(int index,
 	 */
 	public void download(File file, boolean override);
 
+    /**
+     * Downloads the archived files, preserving the original folder structure
+     * 
+     * @param file
```
mtbc: Should be path, okay for rebase but try not to introduce more Javadoc warnings :smiley: Can enable Javadoc "yellow warnings" in Eclipse.

mtbc:
Has this or its rebase been tested on Windows?

mtbc:
Seems to be completely ignoring the folder name I enter.

mtbc:
Download seems to be working just fine for me though on Linux. :+1:

dominikl:
The folder problem is worrying... Was that only on Windows?

mtbc:
That was on Linux. My Windows isn't in a good state right now but would be good if someone could test that if not already.

dominikl:
Could reproduce it. The crucial point is, to type the name of the directory instead of navigating to it via mouse clicks; it doesn't pick up the directory you typed but takes the directory which is currently active in the dialog. Affects Windows and Linux (not OSX, as it doesn't offer the text box to type the name of the directory). I strongly suppose it is an issue of the underling Java file dialog (i. e. hard to impossible to fix), anyway, has to be tackled - if at all - in a separate PR.

dominikl:
By the way, thanks @mtbc for the tip, enabled javadoc warnings in Eclipse now.

jburel:
@mtbc @dominikl: maybe something to add to the "contributing page" for eclipse users?

dominikl:
Good idea. But I'm not really sure, where to put it.

mtbc:
Perhaps http://www.openmicroscopy.org/site/support/contributing/code-contributions.html?

jburel:
https://github.com/openmicroscopy/openmicroscopy/pull/4135#issuecomment-138288123 
not related to this PR but to the FileChooser itself.
Better to be tackled in another PR

sbesson:
@jburel: so should we merge this PR?

jburel:
Was this tested on windows?
If not I will do it tomorrow
cc @pwalczysko @mtbc

mtbc:
I didn't test it on Windows.

jburel:
Tested on windows using volocity files. Everything works as expected. (excluding folder name not related to this PR)
Merging

PR #4137, opened 2015-09-07
ROI tool OutOfMemory (rebased onto dev_5_1)

dominikl:
This is the same as gh-4083 but rebased onto dev_5_1.

----

See [Ticket 12995](https://trac.openmicroscopy.org/ome/ticket/12995)

For analyzing ROI shapes a LinkedHashMap was used, which consumed a lot of memory. This PR replaces it with a double array and a list of points. It also removes some unnecessary duplications of data structures which in total significantly reduces the memory usage.

Test: Analyze some ROIs: Make sure the histogram plots and intensity values are still calculated correctly. Analyze a rectangular ROI same size or bigger as the one mentioned in ticket (250x280px) on a multi channel image, make sure Insight doesn't crash and it doesn't take several minutes for the graphs to appear.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/AnalysisStatsWrapper.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/AnalysisStatsWrapper.java
@@ -23,7 +23,6 @@
 package org.openmicroscopy.shoola.agents.measurement.util.model;
 
 //Java imports
-import java.awt.Point;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


dominikl:
--rebased-from #4083

jburel:
Travis error
```
/home/travis/build/openmicroscopy/openmicroscopy/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/GraphPane.java:260: error: cannot find symbol
```

mtbc:
The statistics look the same with and without this PR:

table-1

table-2

mtbc:
The graphs look the same with and without this PR:

graph-1

graph-2

mtbc:
With a bigger ROI I don't actually notice any performance issues without this PR, but that's fine. Good to merge.

PR #4140, opened 2015-09-07
bump version to use Bio-Formats 5.1.4 (rebased from develop)

mtbc:
OMERO now uses Bio-Formats 5.1.4. In a local build you should find that everything still works fine and that recently merged Bio-Formats fixes -- e.g., from http://www.openmicroscopy.org/community/viewtopic.php?f=11&t=7895 -- are effective.

--rebased-from #4139


sbesson:
Tested on trout/integration. Imported a few formats that have been fixed in Bio-Formats 5.1.4. Both the displayed version and the import worked fine. Merging.

PR #4141, opened 2015-09-07
Add get subcommand to CLI obj command (rebased)

ximenesuk:
See gh-4071

--rebased-from #4071

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -340,8 +340,8 @@ class ObjGetTxAction(NonFieldTxAction):
 
     def on_go(self, ctx, args):
 
-        if len(self.tx_cmd.arg_list) not in (3, 4):
-            ctx.die(335, "usage: get OBJ FIELD [INDEX]")
+        if len(self.tx_cmd.arg_list) != 3:
+            ctx.die(335, "usage: get OBJ FIELD")
 
         field = self.tx_cmd.arg_list[2]
         try:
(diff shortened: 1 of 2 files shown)


mtbc:
https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_obj/TestObj/ passes nicely.

mtbc:
(Testing instructions seem a little light but ...) I'm puzzled:

```
$ omero obj get UploadJob:17 finished
Using session 9923d6ef-7b44-4fea-9898-82ebc5469770 (user-2@localhost:4064). Idle timeout: 10 min. Current group: private-1
No object found: UploadJob:17
$ omero hql 'select finished from UploadJob where id = 17'
Using session 9923d6ef-7b44-4fea-9898-82ebc5469770 (user-2@localhost:4064). Idle timeout: 10 min. Current group: private-1
 # | Col1                     
---+--------------------------
 0 | Tue Sep  8 10:15:53 2015 
(1 row)
```

Are timestamps too complex a type? The "No object found" suggests the problem isn't with the actual field.

mtbc:
(Some experimentation suggests that dimensioned properties aren't yet in scope for this, that's okay.)

mtbc:
For really basic properties this PR does seem to be working fine.

ximenesuk:
Thanks @mtbc the limit of `get` in this PR is basic properties, anything more complex is for discussion and scoping later.

The `obj` command uses the API rather than hql, that may account the `No object found` message.

mtbc:
HQL uses the API too ... ?  But if all this is expected then, especially as this is a rebase, then **good to merge**. (Not sure if some RFE is needed re. "no object found" wrongly suggesting there isn't such an object available in the current context?)

ximenesuk:
Sorry, I meant that it uses the query service get method rather than forming its own hql. Might that have a different effect at the server end? /cc @joshmoore 

The `No object found` message is certainly consistent with the same exception being raised by other commands, e.g. `update` - though it might not be completely helpful :-)

joshmoore:
I don't know why `get` would show nothing. It seems like a problem to dig into elsewhere. :thumbsup:

PR #4134, opened 2015-09-04
List empty tagsets when using omero tag list

ximenesuk:
This PR adds to the fix in https://github.com/ximenesuk/openmicroscopy/commit/32e3434a7b8d46b2d446005fe71e5ed1669e17af which included empty tagsets in the `listsets` command to include empty tagsets in the `list` command as a separate section below the tree view.

To test create some tags and tagsets and have some of those tagsets empty. To create empty tagsets from the CLI you need to create populated tagsets and then remove the contents, e.g.
```
$ omero tag create --name "tag1"
TagAnnotation:437
$ omero tag create --name "tag2"
TagAnnotation:438
$ omero tag create --name "tag3"
TagAnnotation:439
$ omero tag createset --name "tagset1" --tag 437
TagAnnotation:440
$ omero tag createset --name "tagset-empty" --tag 438
TagAnnotation:441
$ omero delete TagAnnotation:438
omero.cmd.Delete2 TagAnnotation 438... ok
$ omero tag list
+- 440:'tagset1'
|\
| +- 437:'tag1'

Orphaned tags:
> 439:'tag3'

Empty tagsets:
> 441:'tagset-empty'
```
This should also work when using the `--tagset` and `--uid` options.

Two tests have been expanded to test for both orphaned tags and empty tagsets.

--rebased-to #4142

PR #4142, opened 2015-09-07
List empty tagsets when using omero tag list (rebased)

ximenesuk:
See gh-4134 for testing instructions.

--rebased-from #4134


sbesson:
Tested the scenario and the semantics on eel/integration. Worked as expected. The[Python  integration tests] are also green. Merging.

PR #4143, opened 2015-09-07
Bugfix: MapAnnotation Image link

dominikl:
Found a bug will trying to reproduce [Ticket 12744](https://trac.openmicroscopy.org/ome/ticket/12744)

Test:
- Select an image.
- Add a MapAnnotation by entering a key, press tab, enter a value, but **don't** hit enter, instead navigate away by selecting another image (before: an exception was thrown at this point because Insight tried to save the ImageAnnotationLink twice)
- Go back to the image and make sure the key/value pair has been saved


dominikl:
--exclude

ximenesuk:
No exception thrown when moving away, annotation there (and saved) when moving back. Tested on all affected objects with no problems. Good to merge.

dominikl:
--rebased-to #4152

PR #4104, opened 2015-08-27
Rendering issue dev_5_1

jburel:
Fix problem related to floating point rendering (and 32-bit)
See ticket https://trac.openmicroscopy.org/ome/ticket/12912

To test:
 * import image from QA 11111 and check https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7825  for details of problem
 * also image ​https://trout.openmicroscopy.org/merge/webclient/img_detail/24054/ under user-1 should be fixed.

cc @sbesson

PR #4144, opened 2015-09-07
Rendering issue dev_5_1 (rebased onto develop)

jburel:
This is the same as gh-4104 but rebased onto develop.

----

Fix problem related to floating point rendering (and 32-bit)
See ticket https://trac.openmicroscopy.org/ome/ticket/12912

To test:
 * import image from QA 11111 and check https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7825  for details of problem
 * also image ​https://trout.openmicroscopy.org/merge/webclient/img_detail/24054/ under user-1 should be fixed.

cc @sbesson


jburel:
--rebased-from #4104

dominikl:
Min/Max settings are now working like expected. The bit depth is still reported as 255 (and can be set from 1 to 255), if that's expected too, good to merge.

jburel:
I have not modified the bit depth section.

PR #4024, opened 2015-08-05
Add simple utility script to bump Bio-Formats version

sbesson:
This script should simplify the bump of the Bio-Formats version number to either a release (x.y.z) or a snapshot (x.y.z-SNAPSHOT) version. Both the version number under `etc/omero.properties` and the resolver under `etc/local.properties.example` should be updated.

To test this PR, try:
- `python bump_bf_version.py x.y.z` (should set `version.bioformats=x.y.z` under `etc/omero.properties` and `ome-resolver` under `etc/local.properties.example`)
- `python bump_bf_version.py x.y.z-SNAPSHOT` (should set `version.bioformats=x.y.z-SNAPSHOT`  under `etc/omero.properties` and `ome-simple-artifactory` under `etc/local.properties.example`)
- `python bump_bf_version.py x.y` should fail with an informative message

PR #4147, opened 2015-09-08
Add simple utility script to bump Bio-Formats version (rebased onto dev_5_1)

sbesson:
This is the same as gh-4024 but rebased onto dev_5_1.

----

This script should simplify the bump of the Bio-Formats version number to either a release (x.y.z) or a snapshot (x.y.z-SNAPSHOT) version. Both the version number under `etc/omero.properties` and the resolver under `etc/local.properties.example` should be updated.

To test this PR, try:
- `python bump_bf_version.py x.y.z` (should set `version.bioformats=x.y.z` under `etc/omero.properties` and `ome-resolver` under `etc/local.properties.example`)
- `python bump_bf_version.py x.y.z-SNAPSHOT` (should set `version.bioformats=x.y.z-SNAPSHOT`  under `etc/omero.properties` and `ome-simple-artifactory` under `etc/local.properties.example`)
- `python bump_bf_version.py x.y` should fail with an informative message

### DIFF
--- a/components/tools/bump_bf_version.py
+++ b/components/tools/bump_bf_version.py
@@ -1,0 +1,54 @@
+#! /usr/bin/env python
+# Script to bump Bio-Formats version
+
+import os
+import re
+import argparse
+
+etc_dir = os.path.join(os.path.dirname(__file__), "..", "..", "etc")
+properties_file = os.path.join(etc_dir, "omero.properties")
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #4024

mtbc:
Good to merge.

PR #4145, opened 2015-09-07
Matlab: add group support to create objects function

sbesson:
See https://trello.com/c/gSKqCAe2/49-omero-matlab-group-support

This PR:
- addresses some comments from https://github.com/openmicroscopy/openmicroscopy/pull/4136
- reviews the object creation functions `createXXX` and add group support to them
- improves the performance of the `createDataset/createPlate` function when linking to projects/screens by performing a single transaction instead of multiple
- extends the `WriteData` training example to cover the semantics of the `createXXX` functions and cover the annotation creation function with a group argument.

/cc @emilroz

PR #4148, opened 2015-09-08
switch source file line breaks to LF

mtbc:
We generally use Unix-style line endings but some files differ. This PR fixes (now, some of) those outliers. No change at all should be observed with OMERO's behavior. See https://github.com/openmicroscopy/openmicroscopy/pull/4148/files?w=1.

--no-rebase


joshmoore:
"Showing  0 changed files  with 0 additions and 0 deletions." when viewing with `?w=1`

sbesson:
As a heads-up, assuming this is going to conflict with the rebase of https://github.com/openmicroscopy/openmicroscopy/pull/4145, can we postpone this cleanup once the MATLAB changes are in?

mtbc:
Rather than check for specific conflicts, have force-pushed away commits that touch any MATLAB file, or any server/blitz file (for @jburel's Javadoc work on `ome.services.*`); we can finish them up once those PRs are in.

sbesson:
:+1: works for me

PR #4149, opened 2015-09-08
Display times in more readable way

dominikl:
Same as #4131 just for Insight.

Test: See test instructions on the above mentioned PR, compare with OMERO.Web.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ChannelAcquisitionComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ChannelAcquisitionComponent.java
@@ -28,6 +28,7 @@ import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
+import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Collection;
@@ -89,6 +90,9 @@ class ChannelAcquisitionComponent
 
 	/** Action ID to show or hide the unset general data. */
 	private static final int	GENERAL = 0;
+	
+	/** Format used for displaying time in ms */
+	private static final DecimalFormat msFormat = new DecimalFormat("#ms");
 	
 	/** Reference to the parent of this component. */
 	private AcquisitionDataUI					parent;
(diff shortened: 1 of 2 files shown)


ximenesuk:
Just to be clear, this PR is on develop while the equivalent web one is on 5.1  Will both be rebased?

dominikl:
Yes, was intending to rebase the PR. I just open PRs by default against develop and then rebase, or should it be the other way round?

dominikl:
Sorry, should have added, the example image for testing is also on eel, user-3, image id: 23156

ximenesuk:
@dominikl no problem! I just wanted to check before testing so I knew which server to look at.

ximenesuk:
The changes look good. The one thing I did notice is that scientific notation is sometimes used and sometimes not for numbers that could be represented normally without any problems. See this for `Image:9235`

image

and this for `Image:9241` where scientific and standard notation are both used:

image

This may be outside this PR and I have not double-checked with web.

dominikl:
Thanks, the switch to the scientific notation is not nice, shouldn't happen. I'll fix that within this PR.

dominikl:
--rebased-to #4154

ximenesuk:
The scientific notation display has been fixed and so the PR looks good.

However, and maybe this is just me, I think displaying hours, minutes and seconds separately with units makes sense as they are in non-decimal number bases and that reading is natural. And, when a time is less than one second showing it in milliseconds or microseconds makes sense too. But when there are seconds and milliseconds I think that should be displayed as a decimal as we are not dealing with decimal-based units. So:

image

`500ms` looks fine but `16s 208ms` does not read as clearly as `16.208s`

That said, and given @dominikl is away and that this is an improvement, this PR should be merged and maybe the time display could be discussed at a usability level? /cc @pwalczysko @gusferguson

gusferguson:
@ximenesuk - I agree, it should be displayed as you suggest. Would be happy to merge this and adjust with another PR when practical.

ximenesuk:
See also: https://github.com/openmicroscopy/openmicroscopy/pull/4131#issuecomment-139135488

Which is the correct time `16s` or `16s 208ms` ?

gusferguson:
@ximenesuk
That discrepancy certainly needs looking into before merging.

ximenesuk:
@gusferguson I'm guessing that web is truncating and losing precision but it would be good to get @will-moore to check that.

jburel:
@will-moore: did you check?

will-moore:
@jburel Yes - web is rounding to nearest second in the case of ```16s```. We have to round the data somewhere. The question is always how much accuracy is needed. Probably ```16.2s``` is what we want, instead of ```16s``` or ```16s 208ms```?

jburel:
```16.2``` is probably right

ximenesuk:
I'd agree with rounding to a single decimal place.

sbesson:
Same comment as gh-4131. Merging.

PR #4151, opened 2015-09-09
CLI: add token creation command

sbesson:
See https://trac.openmicroscopy.org/ome/ticket/12324

This PR adds support for token creation in the CLI, i.e. sessions with no `timeToIdle`. The CLI `sessions` subcommand now has an extra `createtoken` subcommand which will create such sessions for the current logged in user. The default expiration is set to 30 days but can be configured using the `--expiration` option:

```
bin/omero sessions createtoken
bin/omero sessions createtoken --expiration TIME
```

The `sessions who` subcommand has been extended to also report the session expiration time next to the timeout and improve the time formatting of long-lived sessions to support hours and days (including the corresponding unit tests). An integration test is also added to cover the semantics of the token creation

In its current state, the current implementation this PR uses `SessionBean.createSessionWithTimeouts` and thus suffers from the same limitations as the `--sudo` option, i.e. it can only be run by group owners or admins. Suggestions are welcome here on whether a new `SessionBean.createToken` should be created (or any method), or the check logic of `SessionBean.createSessionWithTimeouts` should be relaxed in the case wehere the `Principal` user is the same as the current context user (or any alternate solution)

UPDATE: now this PR adds a `createToken` method subjected to  to `ISession` with its implementation in `SessionBean` requiring password to be passed. Pytthon integration tests are added to cover the semantics of the new command and check token creation is not possible from a session key.
/cc @joshmoore @ximenesuk


ximenesuk:
Everything looks to work okay - creating, listing, expiring, etc. - and the tests look good. Is my understanding of the last paragraph in the description correct in that `--sudo` is not required?

Is there a concrete use case that this can be tested with at the moment?

sbesson:
@ximenesuk: correct, the latest version of this PR should allow any user to create a token as long as the session was created with a password. While implementing, I realized one of the limitations of this approach is that sessions (including tokens) die when the server is restarted. Is there a way/ticket to make sessions with `timeToIdle` set to zero persistent up to `timeToLive` @joshmoore/@mtbc?

The concrete use case proposed in the ticket was the usage of tokens in scripts - presumably for importing? would that be useful for Dropbox?

joshmoore:
Thoughts:

 * I'm not a huge fan of "createtoken" as a name. Will we end up with other token commands? If so `token [create | list | remove | ...]`?
 * `omero sessions -h` is a bit unintuitive now with the description of "Sessions are stored locally on disk. Several can be active simultaneously, but only one will be used for a single invocation of bin/omero."
 *  `omero sessions createtoken -h` description: what's a "token"?
 * Still working through the workflow, but I did get "Bad session key. Cannot join 887252ae-e1b2-4a28-b86c-040edc4f6cd5 ..."
 * Also, while calling `createtoken` after a `--sudo` I was asked for the user's (i.e. not root's) password.

sbesson:
I will leave the documentation comments for the end since they are the easiest to address once we have made the difficult design decisions. 

Separating the `token` command was my first commit. The main issue is that the `sessions` control already mixes local and remote sessions which tokens are a special case of and there are lots of duplicates. For instance, `token list` could also be covered by `sessions who` with an extra `--token-only` filter. Assuming we split `token` from `sessions`, should we also move `sessions who`?

As for the `--sudo` case, well found, this not something that is covered. The question is: should we simply disable sudo when creating tokens or accept the `sudoer` password?

manics:
`sessions who` always seemed an odd one out to me, since the other subcommands are to do with local session management.

sbesson:
Understood. Trying to decide of the best way forward here. If we want to use this opportunity to review the local vs remote sessions discrepancy and reorganize  the CLI sessions plugin, my feeling is that we need more scoping and this can be properly scheduled post 5.1.4.

The remaining question is whether we should still have a go at providing a first implementation of a token creation API for 5.1.4 and if so if we should implement it as a `token` subcommand (or whatever qualifies remote sessions)? I can have a go at this proposal for tomorrow.

manics:
Is there a pressing requirement for it to be in 5.1.4? If not open against develop and we're free to make breaking changes?

sbesson:
Nothing more than the fact the ticket was on the 5.1.4 milestone. Assuming there is no pressing need, I am happy to move it to the 5.1.0 milestone.

sbesson:
Card created https://trello.com/c/yQ0Jw9gf with a summary of the major points to be addressed for 5.2.0

PR #4152, opened 2015-09-09
Bugfix: MapAnnotation Image link (rebased onto dev_5_1)

dominikl:
This is the same as gh-4143 but rebased onto dev_5_1.

----

Found a bug will trying to reproduce [Ticket 12744](https://trac.openmicroscopy.org/ome/ticket/12744)

Test:
- Select an image.
- Add a MapAnnotation by entering a key, press tab, enter a value, but **don't** hit enter, instead navigate away by selecting another image (before: an exception was thrown at this point because Insight tried to save the ImageAnnotationLink twice)
- Go back to the image and make sure the key/value pair has been saved


dominikl:
--rebased-from #4143

ximenesuk:
This looks good to merge.

PR #4153, opened 2015-09-09
Interpolate config option (rebased onto dev_5_1)

dominikl:
This is the same as gh-4087 but rebased onto dev_5_1.

----

Same as #4084 but for Insight. 
Test: Best to be tested in conjunction with #4084 .


dominikl:
--rebased-from #4087

jburel:
see https://github.com/openmicroscopy/openmicroscopy/pull/4155#issuecomment-139375127
for difference between the clients

jburel:
```
     boolean isInterpolation() {
-        return interpolation;
+        if (ImViewerFactory.isInterpolation() != null)
+            return ImViewerFactory.isInterpolation();
+        else
+            return interpolationServerProperty;
     }
```
else not required

jburel:
works as expected

jburel:
Code comment can be handled in a follow-up PR since @dominikl is away and the matching web PR has been merged.

PR #4154, opened 2015-09-09
Display times in more readable way (rebased onto dev_5_1)

dominikl:
This is the same as gh-4149 but rebased onto dev_5_1.

----

Same as #4131 just for Insight.

Test: See test instructions on the above mentioned PR, compare with OMERO.Web.


dominikl:
--rebased-from #4149

ximenesuk:
This looks good an is consistent with gh-4149

My comment at https://github.com/openmicroscopy/openmicroscopy/pull/4149#issuecomment-139133425 applies here too.

jburel:
Merged as it is, a follow-up PR will be required cf. last comments on gh-4131

PR #4155, opened 2015-09-09
Interpolate config option (rebased onto dev_5_1)

will-moore:
This is the same as gh-4084 but rebased onto dev_5_1.

----

See http://trac.openmicroscopy.org/ome/ticket/12924

To test:
 - Check in web full image viewer, that by default, interpolation is ON (checkbox is checked and image is not pixelated when zoomed right in).
 - Also Preview viewer should be interpolated by default.
```
$ bin/omero config set omero.client.viewer.interpolate_pixels False
$ bin/omero admin restart
```
 - Log out of webclient. Log In again.
 - Re-open image viewer. Interpolation should be OFF (checkbox unchecked and pixellated image on zoom).
 - Preview viewer should be pixelated too.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1543,7 +1543,7 @@ class _BlitzGateway (object):
         :return:    String
         """
         return (self.getConfigService().getConfigValue(
-                "omero.client.interpolate_pixels") or 'true')
+                "omero.client.viewer.interpolate_pixels") or 'true')
 
     def getWebclientHost(self):
         """
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1535,6 +1535,16 @@ class _BlitzGateway (object):
         return (self.getConfigService().getConfigValue(
                 "omero.client.viewer.initial_zoom_level") or 0)
 
+    def getInterpolateSetting(self):
+        """
+        Returns default interpolation setting on the server.
+        This is a string but represents a boolean, E.g. 'true'
+
+        :return:    String
+        """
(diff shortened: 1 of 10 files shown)


will-moore:
--rebased-from #4084

jburel:
Checkbox is selected by default
I have tried on various images but selecting/unselecting does not seem to have any effect

Any image I should look at

will-moore:
@jburel Checkbox is supposed to be checked by default.
It's only supposed to be unchecked after interpolate is set to false.

jburel:
Ok I found some images. Toggling button on/off is working.
@will-moore: clarification: The checkbox was selected as expected (forgot part of the sentence)

jburel:
The difference between web and insight:
 * If the user unselects the interpolate box in insight and open another image. The option is not reset.
This is not the case in web
cc @pwalczysko @gusferguson

jburel:
Discussed with @gusferguson today, having the option preserved during the session is fine in the Desktop client, 
The difference is minor so no action required.

will-moore:
@jburel - so, we're good to merge?

jburel:
work as expected. Merging

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1535,6 +1535,16 @@ def getInitialZoomLevel(self):
...
                 "omero.client.viewer.initial_zoom_level") or 0)
 
+    def getInterpolateSetting(self):
+        """
+        Returns default interpolation setting on the server.
+        This is a string but represents a boolean, E.g. 'true'
+
+        :return:    String
+        """
+        return (self.getConfigService().getConfigValue(
```
aleksandra-tarkowska: This is breaking change for 5.1 series. It should go in try catch block to guarantee compatibility with < 5.1.4 server.

PR #4108, opened 2015-08-28
web logging (rebased onto develop)

aleksandra-tarkowska:
This is the same as gh-4090 but rebased onto develop.

----

This PR improves logging by logging full request and status code 

To test go to:
  - http://host/weberror/error_404/
  - http://host/weberror/error_500/

and check if OMEROweb_brokenrequest.log contains entire REQUEST OBJECT like:

```
2015-08-24 16:42:39,349 ERROR [                          django.request] (proc.34016) handle_uncaught_exception:226 HTTP 500 <WSGIRequest
path:/weberror/error_500/,
GET:<QueryDict: {}>,
POST:<QueryDict: {}>,
...
```

note that OMEROweb.log will only contains python stacktrace and HTTP 404 as a WARNING

cc @manics @sbesson

PR #4090, opened 2015-08-24
web logging

aleksandra-tarkowska:
This PR improves logging by logging full request and status code 

To test go to:
  - http://host/weberror/error_404/
  - http://host/weberror/error_500/

and check if OMEROweb_brokenrequest.log contains entire REQUEST OBJECT like:

```
2015-08-24 16:42:39,349 ERROR [                          django.request] (proc.34016) handle_uncaught_exception:226 HTTP 500 <WSGIRequest
path:/weberror/error_500/,
GET:<QueryDict: {}>,
POST:<QueryDict: {}>,
...
```

note that OMEROweb.log will only contains python stacktrace and HTTP 404 as a WARNING

cc @manics @sbesson

PR #4156, opened 2015-09-09
Custom server error deprecation

sbesson:
See https://trello.com/c/ra7UKuEJ/50-resurrect-5-1-web-method

This PR restores a handler removed in https://github.com/openmicroscopy/openmicroscopy/pull/4090 and deprecates it (removed in 5.2 via #4108).

--no-rebase


joshmoore:
Code changes makes sense. Since it's not used anywhere, no functional test. :thumbsup:

PR #4136, opened 2015-09-04
Matlab group support training

sbesson:
This PR rebases the additional commits included in in https://github.com/openmicroscopy/openmicroscopy/pull/4111 to test the group support via the training examples.

To test this PR, check https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-training/ is green

--rebased-from https://github.com/openmicroscopy/openmicroscopy/pull/4111

PR #4160, opened 2015-09-10
Matlab: add group support to create objects function (rebased onto develop)

sbesson:
This is the same as gh-4145 but rebased onto develop.

----

See https://trello.com/c/gSKqCAe2/49-omero-matlab-group-support

This PR:
- addresses some comments from https://github.com/openmicroscopy/openmicroscopy/pull/4136
- reviews the object creation functions `createXXX` and add group support to them
- improves the performance of the `createDataset/createPlate` function when linking to projects/screens by performing a single transaction instead of multiple
- extends the `WriteData` training example to cover the semantics of the `createXXX` functions and cover the annotation creation function with a group argument.

/cc @emilroz

### DIFF
--- a/components/tools/OmeroM/src/annotations/linkAnnotation.m
+++ b/components/tools/OmeroM/src/annotations/linkAnnotation.m
@@ -50,7 +50,7 @@ end
 % Create object annotation link
 context = java.util.HashMap;
 group = parent.getDetails().getGroup().getId().getValue();
-context.put('omero.group', num2str(group));
+context.put('omero.group', java.lang.String(num2str(group)));
 
 link = objectType.annotationLink();
 link.setParent(parent)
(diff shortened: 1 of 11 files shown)


sbesson:
--rebased-from #4145

sbesson:
See https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-training/90/consoleText

jburel:
sorry I missed that one yesterday
in ```createDataset.m```, you used to return a loaded dataset, now this will no longer be the case when you do ```link.getChild()```
This will probably require a warning

sbesson:
Thanks. Captured in https://trello.com/c/gSKqCAe2/49-omero-matlab-group-support. I will look at the impact of this change on the objects created by this function and warn accordingly in the docs.

sbesson:
@jburel: investigated `createDataset(s,name)` vs `createDataset(s,name,parent)`. I could not figure out the difference between the objects returned by both methods (both objects are minimally loaded although there might be nuances in the queries).

jburel:
I do not think it is necessary to spend more time on it. Thanks merging

PR #4167, opened 2015-09-15
Parameterize DSL database types

jballanc:
This moves the translation map used for converting OME model types to
database types into a resource file that can be specialized on a
per-database-profile basis.

### DIFF
--- a/components/dsl/src/ome/dsl/Property.java
+++ b/components/dsl/src/ome/dsl/Property.java
@@ -8,15 +8,13 @@
 package ome.dsl;
 
 // Java imports
+import java.io.IOException;
 import java.sql.Timestamp;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
-
-// Third-party libraries
-import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


jballanc:
To test: following a successful build, `./build.py -f components/model/build.xml publish-schema` should leave the `OMERO5.1__1` sql files unchanged. (Since it's sort of hard to test a negative, you can also attempt to alter one of the mappings, but then a full `./build.py clean build` is required before re-publishing the schema.

mtbc:
Only difference is the GENERATED line.

components/dsl/src/ome/dsl/Property.java
```diff
@@ -8,6 +8,7 @@
 package ome.dsl;
 
 // Java imports
```
mtbc: just FYI, whenever you see these import comments you can delete them

mtbc:
No problems with the diff. **Good to merge.**

sbesson:
--rebased-to #4175

components/dsl/src/ome/dsl/Property.java
jballanc: Thanks, noted!

PR #4168, opened 2015-09-15
Use unidata Maven repository for edu.ucar dependencies

sbesson:
This PR registers a new Maven repository for downloading edu.ucar dependencies like netcdf instead of defaulting to the ome.external repository. bufr and grib are still consumed from the OME artifactory since their version is not available via Unidata Maven repository.

To test this PR:
- from a clean Maven cache or minimally after running `rm -rf ~/.m2/repository/edu/ucar/`, run the default build command
- check the NetCDF dependency is properly retrieved from the Unidata artifactory (i.e. the build passes)
- under `dist/share/server/dependencies.html` check `netcdf 4.3.19` is included
- as a bonus check there is no more `joda-time` version evication with joda-time 2.2 being the only dependency (https://trello.com/c/YJJ4OAd4/64-fix-joda-time-version-mismatch)

This PR is a prerequisite for bumping NetCDF version Bio-Formats side (https://github.com/openmicroscopy/bioformats/pull/1980). Once merged, the `edu/ucar` Maven repositories will need to be cleaned CI-side to make sure 5.1.4 has the proper dependency report.

### DIFF
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -67,7 +67,8 @@
            root="http://artifacts.openmicroscopy.org/artifactory/simple/${simple.repository}/"/>
       <ibiblio name="ome-artifactory" cache="maven"
           usepoms="true" useMavenMetadata="true"
-          m2compatible="true" root="http://artifacts.openmicroscopy.org/artifactory/maven/"/>
+          m2compatible="true"
+          root="http://artifacts.openmicroscopy.org/artifactory/maven/"/>
 
       <ibiblio name="unidata.releases" cache="maven"
               usepoms="true" useMavenMetadata="true"
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -69,6 +69,11 @@
           usepoms="true" useMavenMetadata="true"
           m2compatible="true" root="http://artifacts.openmicroscopy.org/artifactory/maven/"/>
 
+      <ibiblio name="unidata.releases" cache="maven"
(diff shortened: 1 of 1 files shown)


mtbc:
With regard to Joda time version, how to get "the ivy report" mentioned in Trello?

sbesson:
The ivy report is actually properly formatted as HTML using the depencies.html file that is mentioned in the description.

mtbc:
Aha, most convenient.

etc/ivysettings.xml
```diff
@@ -69,6 +69,11 @@
           usepoms="true" useMavenMetadata="true"
           m2compatible="true" root="http://artifacts.openmicroscopy.org/artifactory/maven/"/>
 
+      <ibiblio name="unidata.releases" cache="maven"
+              usepoms="true" useMavenMetadata="true"
+              m2compatible="true" root=" http://artifacts.unidata.ucar.edu/content/repositories/unidata-releases/"/>
```
mtbc: leading space before URL

mtbc:
`dependencies.html` is cool and useful. Only Joda 2.2 on that now. :+1:

mtbc:
Yes, looks like netcdf 4.3.19 wanted joda 2.2.

sbesson:
Pushed stylistic commits

mtbc:
Good to merge.

sbesson:
--rebased-to #4173

PR #4169, opened 2015-09-15
Fix import target for omero import

ximenesuk:
This is a partial potential fix for http://trac.openmicroscopy.org/ome/ticket/12781 However, it fixes the problem only in `omero import` and not in `importer-cli`. This is a relatively easy fix though it would mean the two command line importers diverging slightly and thus needing enhanced documentation. It may not be as comprehansive as it needs to be.

Questions:
 - are there any unintended consequences?
 - would it be worth adding this as a partial fix until something better was found?
 - could similar group switching logic be used in `importer-cli`? (It didn't look as promising.)
 - should the group switch be reverted after the import?
 - there is no logging from the plugins but should a `ctx.out` be used at some point to warn of the switch?

Testing:
 - The re-enabled test should pass, all the import tests with targets should pass.
 - For a user in several groups, create Datasets/Screens in those groups and try to import Images/Plates using the `-d`/`-r` options respectively across the different groups. Imports should succeed.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/import.py
+++ b/components/tools/OmeroPy/src/omero/plugins/import.py
@@ -205,10 +205,41 @@ class ImportControl(BaseControl):
                                not args.java_f and
                                not args.java_advanced_help)
         if connection_required:
-            client = self.ctx.conn(args)
+            client = self.get_client(args)
             self.command_args.extend(["-s", client.getProperty("omero.host")])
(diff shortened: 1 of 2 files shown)


joshmoore:
NB: this will work less with the --target branch.

ximenesuk:
@joshmoore is "will work less" a polite way of saying it won't work?

ximenesuk:
Closing for now since this broke stuff!

PR #4170, opened 2015-09-16
Roi position

jburel:
This PR combines commits from gh-4046 and gh-4045
Improve support of rois from ImageJ

Matches changes made in https://github.com/openmicroscopy/bioformats/pull/1974
To test:
 * Follow scenarios described in https://github.com/openmicroscopy/bioformats/pull/1974

Cc @pwalczysko @stelfrich


jburel:
@pwalczysko: I have refactored few things in the reader in order to handle similar cases that the ones supported in B-F

pwalczysko:
Bug: ROIs ignored in OMERO after import.

  - In IJ, New > Image, create a new image with 3 slices, leave the default sizes otherwise
  - Save the image by Save As > Tiff locally
  - Reopen the image using File > Open (the same result is achieved with B-F Importer as well)
  - Draw 3 ellipses, each on one plane of the 3-plane tiff
  - Add each ellipse to Overlay using Cmd B
  - Check that the ellipses are in Overlay (when you start drawing new regions, they do not vanish)
  - File > Import > to OMERO
  - import the image to trout merge
  - use Insight to open the imported image
  - observe that you have no ROIs in OMERO --> Bug, regression against 5.1.x releases
  - observe that when you now click ``Save ROIs to OMERO`` in the Insight-ij, the ROIs will be saved as expected and the ROIs in OMERO will have the Z and T info as expected.

pwalczysko:
Note that the "ROIs not present bug" above https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141087363 is specific for overlays - all works as expected if you use ROI Manager.

pwalczysko:
Edge case: Open image from OMERO with ROI Manager open containing regions from previous (multi-t and multi-z) images. 

  - follow the workflow in https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141087363
  - leave the ROI Manager open with the 3 regions on 3 different T planes (first, second and third plane)
  - Also from OMERO, open a lei image (3 channels, 1 z/t plane only).
  - observe in IJ that the ROIs are visible on the image
  - use ``Save ROIs to OMERO``
  - observe that the activities window reports ROI Saved
  - go to OMERO -> no ROIs are on the lei image

Not sure whether this is good or bad, but it is a discrepancy with the B-F exporter behaviour (the first plane ROI should be saved, the others excluded)

pwalczysko:
Bug: (regression against 5.1.x releases)

  - from OMERO, open ``050118.lei [07-01-a]`` (first image of the 56-lei MIF, on trout under 5.0 project, ID 21907)
  - in IJ, draw a ROI and add it to the ROI Manager
  - use ``Save ROI(s) to OMERO``
  - observe that the Activities window reports ``ROIs Saved``
  - observe no ROI is saved in OMERO (using Insight)

pwalczysko:
Bug: Ellipse is not shown in OMERO

   - from OMERO, open a dv
   - draw an ellipse and add it to the Overlay (ROI Manager workflow does not work here, because of https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141101606)
   - use Save ROIs to OMERO 
   - observe you have the region in OMERO, listed in the OMERO ROI Manager as expected
   - click onto the region in the Full Viewer
   - observe that you have only boundary selection 


screen shot 2015-09-17 at 15 25 19

pwalczysko:
I would suggest to stop testing for today - with the 2 ROI saving bugs, https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141087363 and https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141101606 it would be difficult to test the more rare and complicated cases properly

stelfrich:
@pwalczysko @jburel : something underlying must have change in the meantime.. when I started out with my initial PR, overlays were taken into account even when just using Plugins > OMERO > Save Image to OMERO. With this PR, the methods for reading ROIs are not even executed.

jburel:
I have few commits to push

jburel:
@pwalczysko: problem should now be fixed. I have also found a problem with mif and rois that existed before this PR.
This should also be fixed. 
I now need to look at the ellipse problem

jburel:
The ellipse problem might have been a side effect of the other problems.
It should now work.

components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/ROIReader.java
```diff
@@ -207,10 +220,37 @@ private void formatShape(Roi roi, ShapeData shape)
...
+            shape.setZ(0);
+            shape.setT(0);
+        } else if (imageC == 1 && imageT == 1) {
+            z = pos;
+            t = 0;
+            c = 0;
+            shape.setC(0);
+            shape.setT(0);
+        }
+        if (c >= imageC || z >= imageZ || t >= imageT) {
```
stelfrich: @jburel: This check will leave out ROIs on the last frame, since imageC, Z, and T are 1-based (while c, z, and t are 0-based). I'd suggest to change it to `if (c > imageC || z > imageZ || t > imageT)`.

jburel: good point

pwalczysko:
Bugs https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141087363 (ROIs not saved from Overlay), https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141101606 (ROIs from IJ not saved on images opened from OMERO) and https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141098763 (Edge case when old regions are applied on planes which do not exist).

The only exception is the bug anticipated by @stelfrich - this one is apparent when doing https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141087363 workflow (the ROIs on the last plane are not saved).

To do: after the last commits of @jburel are in, 
   - re-check the workflow https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-141087363 concentrating on the last plane ROI saving
   - check the MIF workflows

jburel:
@pwalczysko: the "MIF" problem mentioned was a code issue. ROI could have been converted twice in some cases.

stelfrich:
@jburel @pwalczysko: is https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-141507203 an issue that should be addressed in this context?

sbesson:

jburel:
see https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-142228754
imageJ limitation

pwalczysko:
@jburel : The last commits work fine, the overlay and the ROI manager work fine. 

Problem: 
  - the MIF workflow does not work (the ROIs are not saved on different MIF images in the ``Beta catenin`` file)
     - open ``Beta catenin`` (all 5 images)
     - draw a ROI on the last three of them (Beta Catenin.lif [Series004]- draw a rectange, Beta Catenin.lif [Image003] - draw a polygon, Beta Catenin.lif [Image002] - draw a line)  using Overlay
     - ``Save Image(s) to OMERO`` and select the default ``Image from current window``
     - in OMERO, the Beta Catenin.lif [Image003] does not have its region, the other two DO have their respective regions

pwalczysko:
@jburel : Additional info to the MIF workflow 
    - it is always the image which is most recently in focus in IJ which will NOT have the ROI saved on it in OMERO after ``Save Image(s) to OMERO`` (whien you use ``Add Image from active window`` option
    - note that when you use ``Add images from all open windows`` option, the workflow works as expected -> all ROIs are on the right place, nothing is missing

pwalczysko:
Bug: Ellipse

   - go through the workflow of MIF on ``Beta catenin`` and use Overlay, drawing an ellipse, then ``Add images from all open windows`` to be sure you get the ellipse into OMERO
   - open the image in OMERO, find the ellipse and 
   - observe that it has not  been really drawn, just the bordering selection markers are show 
 
This bug was mentioned above already, but is specific for some types of images only, that is why it was not repeated.

jburel:
@pwalczysko: 
 * mif issue should now be fixed.
 * Ellipse and color is also fixed. The key point when testing is to select the option ```Add images from all open windows```

pwalczysko:
The MIF issue as reported is fixed now (when ``Add images from current window`` is used, the whole MIF is imported as expected, and the regions are present as drawn on Overlay.

pwalczysko:
The ellipse issue - unfortunately no improvement at all - see screeenshot
screen shot 2015-09-23 at 10 09 28

pwalczysko:
Another MIF problem: when continuing the workflow in the sense

   - with the Beta catenin MIF already saved in OMERO with some ROIs on it
   - leave the MIF open in IJ and draw more ROIs on one image only
   - use ``Save ROI(s) to OMERO`` to save the newly added ROIs
   - the ROIs are saved as expected (no new import is suggested as expected) but
   - the ROIs are saved on a wrong image (drawn on ``...Image002``, saved on ``...Image003``)

pwalczysko:
In summary:
 1. the MIF problem reported yesterday https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-142252782 is fixed
 2. the ellipse problem https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-142262758 is not fixed
 3. new MIF problem observed (not tested this advanced workflow yet)

jburel:
@pwalczysko:
 * I do not think The ellipse issue is a regression
 * new MIF probably exists in the release version.

pwalczysko:
Unfortunately, the ellipse issue is a regression. I have an expected behaviour on OMERO 5.1.3 release insight-ij plugin combined with B-F 5.1.4 release.

pwalczysko:
(interestingly, with the release version of insight-ij, the ellipses will always appear as yellow in OMERO, in here, the ellipses which happen to arrive successfully in OMERO, i.e. not as a ghost, their color is white)

pwalczysko:
Simplified workflow for the ellipse: 
  - open the first image in ``test_images_good/dv`` from disk in IJ using B-F importer
  - draw one ellipse and add it to Overlay (ellipse appears yellow in IJ)
  - use ``Save image(s) to OMERO``
  - use Insight to check the ellipse in OMERO -> it is just a ghost

pwalczysko:
@jburel : Yes, I agree, the additonal MIF problem is not very importeant, when using ``Add images from all open windows``, the behaviour is as expected. Also when you just draw a new ROI and keep the focus and use ``Add images from selected window`` the behaviour is as expected.

jburel:
@pwalczysko: I cannot reproduce the ellipse problem

pwalczysko:
As discussed with @jburel : The ellipse problem is very difficult to reproduce with different IJ setups, sometimes IJ has itself problems with drawing the ellipse.
All the ellipses are visible in Web correctly.
As this is an edge case issue, we suggest to go ahead with this PR in any case even if the ellipse problem persists. We can have follow-up release to fix this edge case problem.

sbesson:
Thanks. Merging.

pwalczysko:
@jburel @bramalingam @stelfrich : Additional info to the ellipse behaviour on Overlays https://github.com/openmicroscopy/openmicroscopy/pull/4170#issuecomment-142550571. After some additional tests with @bramalingam it turns out that:
   1. This is NOT a regression (when hyperstack is chosen, just like a multi-z, multi-c dv, e.g. ID: 8751 user-1 trout merge, and the 5.1.3 insight-ij with 5.1.4 BF is used, and an ellipse is added to Overlay, then the 
        - T position is not set in Insight ROI Manager
        - the ellipse is a ghost
   2. The reason why I thoiught this is a regression is that when you use a non-hyperstack image, like ``test_images_good/tiff/QA images/Artificial_xyt_8bit.tif``, then the ellipse will be visible in Insight, but this is the same for 5.1.3 release insight-ij as well as the 5..1.3 merge with this PR in.
   3. The other regions like rectangle "work" in the Ad 1. situation, but they are being shown as ``blue`` in OMERO (should appear yellow though)
   4. The workflow always works with ROI Manager in IJ used, because this gives 
       - the T position set as X (where is the proper T number where the ROI was drawn)
       - the ellipse will appear as white in IJ (this is a bit confusing, as it should be yellow, but admittedly a very minor issue)

This all brings us back to the discussion with @stelfrich here https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-142242934 where he is aware of the problem -> basically again Overlays handling in IJ issue, which we had from Day1.

bramalingam:
A bit more background information:

1) Everything works (overlays and ROIManager workflows) if the image is a Stack.
2) Overlay workflows fail if the image is a HyperStack.

This is because of the way ImageJ handles the positions of the ROI's between Stacks and Hyperstacks (in both Overlay and ROIManager cases).
All the StrokeColor issues as noted by @pwalczysko are minor issues.

jburel:
--rebased-to #4213

PR #4172, opened 2015-09-16
Size limit for download as jpg

will-moore:
See https://trac.openmicroscopy.org/ome/ticket/12982

This limits the download as jpeg / png / tiff to images of 12k * 12k or smaller.
This cut-off was determined by trying to download images of various sizes.
Above 2147483648 pixels (46k * 46k) we get 
```
java.lang.NegativeArraySizeException: null
    at omeis.providers.re.RGBIntBuffer.<init>(RGBIntBuffer.java:62) ~[common.jar:na]
```
For 15k images we get
```
java.lang.OutOfMemoryError: Java heap space
    at omeis.providers.re.RGBIntBuffer.<init>(RGBIntBuffer.java:62) ~[common.jar:na]
    at omeis.providers.re.RenderingStrategy.getIntBuffer(RenderingStrategy.java:186) ~[rendering.jar:na]
    at omeis.providers.re.HSBStrategy.renderAsPackedInt(HSBStrategy.java:275) ~[rendering.jar:na]
```
but 12k images download OK, but quite slowly.

To test:
 - Select an image smaller than 12k * 12k or equivalent, and try to download as jpg/png/tiff (should work for 'big' images as well as regular images.
 - Select an image bigger than 12k * 12k and check that the download as jpg/png/tiff options are disabled with a suitable tooltip.
 - Select multiple images. If any single image is too big, the download as jpg/png/tiff options should be disabled. Otherwise, download as a zip should work OK.


mtbc:
(Regarding that first puzzling exception message I'll see if I can get https://github.com/mtbc/openmicroscopy/commit/c9612623 merged in.)

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -204,6 +204,23 @@ def getPlateId(self):
         elif self.acquisition:
             return self.acquisition._obj.plate.id.val
 
+    def canExportAsJpg(self, objDict=None):
+        """
+        Can't export as Jpg, Png, Tiff if bigger than approx 12k * 12k.
+        Limit set by OOM error in omeis.providers.re.RGBIntBuffer
+        """
+        can = True
+        k = 12100 * 12100
```
aleksandra-tarkowska: this could go to settings.py

pwalczysko:
The functionality works as described in the PR header.
Interesting edge case (sizewise, not usage-wise) is the ``5.0 data/big-svs/UMD001_ORO.svs`` which is on the trout db in every group in Project called ``5.0 data``, Dataset ``big-svs``.
The size is 17317 x 11041. It did download on trout latest, but very slowly.
I understand there is some discussion still ongoing about where exactly to put the cutoff
@will-moore @aleksandra-tarkowska

will-moore:
As discussed with @aleksandra-tarkowska, since the limit depends on OOM error, this may vary with different set-ups so it's best to make the limit configurable (see last commit)

will-moore:
You can now set cut-off like this:
```
$ bin/omero config set omero.client.download_as_jpg.max_size 1000000
$ bin/omero admin restart
```
Then log out of web & log in again....

jburel:
the same limit will hold for export as png so I will strip the jpg

will-moore:
@jburel but then you are left with ```'download_as.max_size'``` which doesn't have so much meaning. I could go for ```download_as_jpg_tiff_png.max_size``` but that seems like overkill.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1545,6 +1545,20 @@ def getInterpolateSetting(self):
...
 
+    def getDownloadAsJpgMaxSizeSetting(self):
+        """
+        Returns default max size of images that can be downloaded as
+        jpg, png or tiff, expressed as number of pixels.
+        Default is 144000000 (12k * 12k image)
+
+        :return:    Integer
+        """
+        size = self.getConfigService().getConfigValue(
```
aleksandra-tarkowska: The same here, it is breaking change for 5.1 series. It should go in try catch block to guarantee compatibility with < 5.1.4 server.

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -628,6 +628,12 @@ def leave_none_unset_int(s):
           " 'webtest/webclient_plugins/center_plugin.overlay.js.html',"
           " 'channel_overlay_panel']``. "
           "The javascript loads data into ``$('#div_id')``.")],
+    "omero.web.download_jpg.max_size":
```
aleksandra-tarkowska: that can be removed now

sbesson:

dominikl:
@jburel Guess this new property has to be taken into account by Insight as well, right? I think at the moment jpg export is only allowed for non-tiled images.

jburel:
@will-moore: it will have to, This is applicable in the viewer only we will also have to adjust the "export as" script used to export as jpeg/png/tiff

jburel:
I still think we should modify the name to something like size_download_as.
We then have method like ```canExportAsJpg/downloadSsJpeg``` with description "can export as jpeg, png etc. This is even more confusing

aleksandra-tarkowska:
As this PR is not ready yet, could you please exclude it as is blocking https://github.com/openmicroscopy/openmicroscopy/issues/4191 (this is blocker for 5.1.4 and should be reviewed ASAP)

jburel:
This PR will also require changes in insight to be in synch and also in the export script (i.e. 2 more PRs) This, in my opinion, will require further testing and I do not feel comfortable to have it included for the release.
cc @pwalczysko @gusferguson, @dominikl

will-moore:
@jburel From the user's point of view in web, this PR fixes the 404 bug (by preventing the user trying to download as jpg when the image is too big) and this also prevents OOM errors on the server etc. So I think this should be merged as it is, without waiting for Insight. This is not a feature that was previously aligned between the clients, so we are not introducing divergence here. This is simply the first step of convergence!

aleksandra-tarkowska:
@will-moore I agree with you but I think priority is to get #4191 in as it is blocker, this PR could be merged just after that as it introduce new property

jburel:
The first step will be to turn it off if the image is a big image with the actual limit i.e. the one used to build pyramid. That parameter is already in place

will-moore:
Sorry - missed this last night - just fixing now...

will-moore:
@jburel @sbesson Final change is in now - revert to using 'Big Image' size as limit for download as.
You want to do a final build now?

jburel:
The export as "JPEG, PNG, TIFF" options are greyed out for Big image
The OME-TIFF option should probably also be greyed out since it is not possible to perform it but it does not crash so this can be done in another PR
Tested with 8kx8k.jpeg

jburel:
The option is greyed out but if you click on the option e.g. png, you can still download the image as png for example.

will-moore:
@jburel Should be fixed now!

dominikl:
Works like expected. Export as JPG/PNG/TIFF options are greyed out and disabled if a selected image is a 'big' image. Export as OME-TIFF still avaiable, fails on 'big' images but provides an explanation why it failed.

jburel:
Added an item to https://trello.com/c/4mDNygGv/38-5-1-3-follow-up so we can work on the new limit in 5.2

PR #4175, opened 2015-09-16
Parameterize DSL database types (rebased onto develop)

sbesson:
This is the same as gh-4167 but rebased onto develop.

----

This moves the translation map used for converting OME model types to
database types into a resource file that can be specialized on a
per-database-profile basis.


sbesson:
--rebased-from #4167

mtbc:
Good to merge.

PR #4176, opened 2015-09-16
Right panel fileanns slow 13022

will-moore:
See http://trac.openmicroscopy.org/ome/ticket/13022

This fixes the slow loading of right panel "General" tab when the P/D/I has several file annotations.
Two factors were causing the poor performance:
 - Loading annotations didn't also load the files, so that ann.getFile() created a new ```OriginalFileWrapper``` with an unloaded file, causing the file to be loaded each time on the fly. Now, we include files when loading annotations.
 - Further poor performance was caused by not caching the file for the FileAnnotationWrapper, so that repeated calls to getName() or getFileSize() were loading the file each time. We now cache the file in ann.getFile()

To test:
 - Compare loading speed of right panel for P/D/I with lots of file annotations vv none. Should be about the same.

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_annotation.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_annotation.py
@@ -308,6 +308,35 @@ def testFileAnnotation(author_testimg_generated, gatewaywrapper):
     assert gateway.getObject("Annotation", annId) is None
 
 
+def testFileAnnotationSpeed(author_testimg_generated, gatewaywrapper):
+    """ Tests speed of loading file annotations. See PR: 4176 """
+    tempFileName = "tempFile"
+    f = open(tempFileName, 'w')
+    fileText = "testFileAnnotationSpeed text"
+    f.write(fileText)
+    f.close()
+    ns = TESTANN_NS
+    image = author_testimg_generated
+
+    # use the same file to create many file annotations
+    for i in range(20):
+        fileAnn = gatewaywrapper.gateway.createFileAnnfromLocalFile(
+            tempFileName, mimetype='text/plain', ns=ns)
+        image.linkAnnotation(fileAnn)
+    os.remove(tempFileName)
+
+    now = time.time()
+    for ann in image.listAnnotations():
+        if ann._obj.__class__ == omero.model.FileAnnotationI:
+            # mimmic behaviour of templates which call multiple times
+            print ann.getId()
+            print ann.getFileName()
+            print ann.getFileName()
+            print ann.getFileSize()
+            print ann.getFileSize()
+    print time.time() - now
+
+
 def testFileAnnNonDefaultGroup(author_testimg_generated, gatewaywrapper):
     """ Test conn.createFileAnnfromLocalFile() respects SERVICE_OPTS """
     gatewaywrapper.loginAsAuthor()
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_annotation.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_annotation.py
@@ -16,6 +16,7 @@
 import time
 import datetime
 import os
+from tempfile import NamedTemporaryFile
 
 import omero.gateway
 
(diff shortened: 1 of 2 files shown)


gusferguson:
@will-moore @jburel - problem with testing against trout or eel - the delay was seen most markedly on Demo server - other servers not really affected. So need to have some way of testing it on the Demo server. Can then use same images I used before for benchmarking.

will-moore:
@chris-allan You want to look over these code & behaviour tweaks to Blitz Gateway?

chris-allan:
My initial gut feeling after looking at this is that changing the gateway semantics of `_loadAnnotationLinks()` for everyone in order to improve the performance of a selective case is not something I'm comfortable with. At a minimum we should prepare a series of integration tests that clearly address the positives and negatives of doing this. The questions I would like to answer:

 * What is the actual quantifiable fold performance improvement?
 * How are other types of annotations affected? Are we going to have the same problem there?
 * What is the performance degradation vs. not loading files for say 100 file annotations? 1000?

I don't doubt that there are others.

/cc @knabar, @jballanc, @emilroz

will-moore:
With a gateway test that creates 20 file annotations (will commit this in a sec), looping through all annotations and recording the time like this:
```
now = time.time()
    for ann in image.listAnnotations():
        if ann._obj.__class__ == omero.model.FileAnnotationI:
            print ann.getId()
    print time.time() - now
```
I get the following times:
For the above code, without any changes to Blitz Gateway I get this as a baseline:
0.0473 secs

If I add multiple print statements into the loop, to simulate the behaviour of the web templates,
```
            print ann.getId()
            print ann.getFileName()
            print ann.getFileName()
            print ann.getFileSize()
            print ann.getFileSize()
```
0.4107 secs
This is comparable to what I was seeing locally, although nowhere near as bad as reported http://trac.openmicroscopy.org/ome/ticket/13022

If I add the ann.getFile() to cache the loaded file, I get:
0.1534 secs
This is already a big gain, without any change in the initial loading behaviour.
So I think it's a no-brainer to add this.

If I also add the extra loading of files along with annotations by adding ```left outer join fetch a.file``` to the query (instead of loading 'on the fly') I get:
0.0662 secs

If I don't use the file, don't call getFile(), removing all the ```print getFile* ``` above, then we see that the query itself is still a tiny bit slower than when the query didn't load files. This is the penalty for unnecessarily loading files
0.0606 secs

However, adding the ```left outer join fetch a.file``` doesn't have any effect on the speed of loading other annotations e.g. Comments. 

So, in my test with 20 file annotations, you pay about a 0.02 secs penalty for loading files when you don't need them but you save 0.09 secs if you do need the files. It seems to me that there's almost no use case for loading file annotations without the file, since you need the file for everything, even to getFileName(). So it makes sense to include the file in the initial loading query, especially since it has no performance effect on loading other annotations.

will-moore:
@chris-allan Let me know your thoughts on this, since we hope to have a 5.1.4 RC build on Monday.
 Cheers.

components/tools/OmeroPy/test/integration/gatewaytest/test_annotation.py
```diff
@@ -308,6 +308,35 @@ def testFileAnnotation(author_testimg_generated, gatewaywrapper):
     assert gateway.getObject("Annotation", annId) is None
 
 
+def testFileAnnotationSpeed(author_testimg_generated, gatewaywrapper):
+    """ Tests speed of loading file annotations. See PR: 4176 """
+    tempFileName = "tempFile"
```
chris-allan: This should be replaced by `tempfile.NamedTemporaryFile()` which has auto-delete semantics, will not cause file creation race conditions, etc.

chris-allan:
I've run the following Gist to try and isolate the query performance, see what the scaling is for larger annotation counts, etc. on a fairly fast 5.1.3 system with a reasonably loaded database (>1 million images, >150K annotations):

 * https://gist.github.com/chris-allan/e62f5a2d12dda91b3894

The results for 10, 100 and 1000 annotations are as follows:

```
Testing (10) comment annotation speed...
No joined files elapsed: 0.014093875885
Joined files elapsed: 0.0145931243896
Testing (10) file annotation speed...
No joined files elapsed: 0.0126929283142
Joined files elapsed: 0.0210750102997
```

```
Testing (100) comment annotation speed...
No joined files elapsed: 0.0561380386353
Joined files elapsed: 0.0575759410858
Testing (100) file annotation speed...
No joined files elapsed: 0.0881330966949
Joined files elapsed: 0.154210805893
```

```
No handlers could be found for logger "omero.gateway"
Testing (1000) comment annotation speed...
No joined files elapsed: 0.622567892075
Joined files elapsed: 0.518607139587
Testing (1000) file annotation speed...
No joined files elapsed: 0.944436073303
Joined files elapsed: 1.63939094543
```

The drop in performance of ~57% seems to be pretty linear at least up until 1000 annotations. That 0.6 second additional cost is actually pretty dramatic at the 1000 annotation level considering what you actually need are likely at most one or two fields. Definitely not queries I would suggest be used on anything other than comparatively trivial numbers of file annotations. We know from the work on  fileset, fileset entry and original file loading (#3695, #3753) that you just cannot be making these types of queries on those types of object counts and expect reasonable performance. Even 1000 comment annotations takes >0.6s.

Knowing that other annotation types are unaffected by such a query is at least positive. It is however a worrying trend, as it strongly suggests that the penalty is in the object creation and serialisation. I will compare at some point, not for this PR, the Java vs. Python cost per object.

The caching change certainly in certainly a no-brainer. I buy the argument that if you actually **do** have a file annotation it is useless without its original file partner. The effectively zero cost for other annotation types also means that for a large percentage of cases there is no performance degradation to speak of.

I'm happy for these changes to go in once the issues with test cases are resolved. In the future, maybe even for 5.2, I would strongly suggest that the data loading for these templates **not** use the gateway or gateway wrapper objects.

will-moore:
@chris-allan Thanks for the additional testing and feedback. I definitely agree that we should move webclient to use dedicated performant queries instead of gateway objects, especially to avoid lazy loading as well as slow queries & object creation.
I've updated the test to use tempfile.

sbesson:

sbesson:
@will-moore: are there upcoming commits to fix the issues with test cases described in https://github.com/openmicroscopy/openmicroscopy/pull/4176#issuecomment-141463177?

chris-allan:
I think this was 05edb16, @sbesson. Assuming there are no test failures I'm happy with the current changes.

sbesson:
Thanks for the clarification. Given the state of https://ci.openmicroscopy.org/job/OMERO-5.1-merge-integration-web/438/, merging this.

will-moore:
--rebased-to #4224

PR #4178, opened 2015-09-17
switch Java source file line breaks to LF

mtbc:
We generally use Unix-style line endings but some files differ. This PR fixes some of those outliers. No change at all should be observed with OMERO's behavior. See https://github.com/openmicroscopy/openmicroscopy/pull/4178/files?w=1.

--rebased-to #4185


ximenesuk:
Without examining every line, the "changes" look good. Given that the integration tests didn't fail badly I'd say this is good to merge.

sbesson:
Same question as https://github.com/openmicroscopy/openmicroscopy/pull/4179, would this PR actually be a rebase candidate to keep things in sync?

mtbc:
Same for #4148? I could open a similar 5.1 PR if now's a time it's unlikely to conflict.

mtbc:
See #4185.

PR #4179, opened 2015-09-17
switch MATLAB source file line breaks to LF

mtbc:
We generally use Unix-style line endings but some files differ. This PR fixes some of those outliers. No change at all should be observed with OMERO's behavior. See https://github.com/openmicroscopy/openmicroscopy/pull/4179/files?w=1.

--no-rebase


sbesson:
https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-training/97/ and https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-matlab/40/ both passed and the code diff without whitespace does not show any change. Happy to merge this.

Additionally, I would consider this PR as a candidate for rebasing to keep the MATLAB code in sync in between `dev_5_1` and `develop` and facilitate bug fixing on both lines /cc @chris-allan @emilroz

chris-allan:
:+1:

Definitely a good idea.

sbesson:
Thanks @chris-allan. Merging. Can you rebase this onto `dev_5_1` @mtbc ?

mtbc:
See #4185.

PR #4182, opened 2015-09-17
Add CLI children plugin

manics:
Contains methods for moving all children of a parent object, and for managing orphans

I though this might be handy after @eleanorwilliams discovered some problematic screen imports were resulting in a large number of orphaned images, and attempting to delete them in web takes a very long time.

```
    # Move all orphaned images into Dataset:1
    omero children move Image:orphan Dataset:1

   # Move all images owned by any user from Dataset:1 into Dataset:2
    omero children move --all Dataset:1 Dataset:2

    # Delete all orphaned images owned by current user
    omero children delete Image:orphan

    # Delete all orphaned images owned by any user
    omero children delete --all Image:orphan
```
Todo: Add tests

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/children.py
+++ b/components/tools/OmeroPy/src/omero/plugins/children.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+#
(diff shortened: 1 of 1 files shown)


joshmoore:
Merging as is for 0.0.3. @ximenesuk : note this was less used in the general testing of 0.0.3 and so may need more work.

PR #4183, opened 2015-09-18
switch code template line breaks to LF

mtbc:
We generally use Unix-style line endings but some files differ. This PR fixes some of those outliers. No change at all should be observed with OMERO's behavior. See https://github.com/openmicroscopy/openmicroscopy/pull/4183/files?w=1.

--rebased-to #4185


dominikl:
Looks fine, good to merge.

PR #4185, opened 2015-09-18
switch line breaks to LF (rebased from develop)

mtbc:
We generally use Unix-style line endings but some files differ. This PR fixes those outliers. No change at all should be observed with OMERO's behavior. See https://github.com/openmicroscopy/openmicroscopy/pull/4185/files?w=1.

--rebased-from #4148
--rebased-from #4178
--rebased-from #4179
--rebased-from #4183


rleigh-dundee:
The changes all look fine and are line endings only.  All the jobs are green; waiting on the cpp build only.  Please merge once all are green.

PR #4180, opened 2015-09-17
deprecate OMERO.server's DeleteException

mtbc:
With the graphs reimplementation `DeleteException` isn't used any more.

PR #4188, opened 2015-09-18
deprecate OMERO.server's DeleteException (rebased from develop)

mtbc:
With the graphs reimplementation `DeleteException` isn't used any more.

--rebased-from #4180


jburel:
Do we really need to rebase it? since it will be removed in 5.3

mtbc:
It gives people more advance warning of the deprecation, so more time to adjust their code accordingly.

PR #4192, opened 2015-09-18
Removed seemingly unused TagWrapper class that changed interface

dpwrussell:
I'm not sure what this was added for, but I don't think it does anything and it changes the interface. I know it's not exactly a public API, but seems like it would be good not to change it if possible.

I noticed because it broke webtagging.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -2551,15 +2551,6 @@ class WellWrapper(OmeroWebObjectWrapper, omero.gateway.WellWrapper):
 omero.gateway.WellWrapper = WellWrapper
 
 
-class TagWrapper(OmeroWebObjectWrapper, omero.gateway.TagAnnotationWrapper):
-    """
-    omero_model_TagAnnotationI class wrapper overwrite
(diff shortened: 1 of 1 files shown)


will-moore:
Webtagging seems to be working OK (tested as part of https://github.com/MicronOxford/webtagging/pull/31) so I assume this is fixed (unless there's a specific use-case that shows the bug)? Creating and browsing of Tags is working too.
Good to merge.

dpwrussell:
@will-moore I didn't add any code to work around this so webtagging is working with this PR.

joshmoore:
:+1: Worth adding a Deprecated on dev_5_1?

will-moore:
@dpwrussell This was in dev_5_1, but didn't cause problems in webtagging for 5.1? https://github.com/openmicroscopy/openmicroscopy/commit/02986bebc

@joshmoore I'm not really sure where we'd document this deprecation. The Blitz Gateway API would change (return TagWrapper instead of omero.gateway.TagAnnotationWrapper) but only if you'd imported the webclient_gateway.py - so most users of the Blitz Gateway wouldn't be affected by this change in web.

I remember why I added this now - so that the tag tree template could access ```class="{{ t.getPermsCss }}``` since the ```getPermsCss()``` is on the ```OmeroWebObjectWrapper class```. But this is not needed now with new jsTree.

PR #4195, opened 2015-09-21
Time format follow up (dev_5_1)

dominikl:
Follow-up PR for #4154 
Display seconds as fraction with one decimal place; added whitespace between value and unit; always display next lower unit even if the value is 0 (e. g. 1 min 0 s ; or 1 h 0 min )

Test: Check that the Insight timestamp format matches OMERO.Web, for example on trout user-2 Image 9532

PR #4199, opened 2015-09-23
Time format follow up (dev_5_1) (rebased onto develop)

dominikl:
This is the same as gh-4195 but rebased onto develop.

----

Follow-up PR for #4154 
Display seconds as fraction with one decimal place; added whitespace between value and unit; always display next lower unit even if the value is 0 (e. g. 1 min 0 s ; or 1 h 0 min )

Test: Check that the Insight timestamp format matches OMERO.Web, for example on trout user-2 Image 9532

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ChannelAcquisitionComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/ChannelAcquisitionComponent.java
@@ -510,12 +510,11 @@ class ChannelAcquisitionComponent
             date = DateUtils.round(date, Calendar.SECOND);
             m = date.get(Calendar.MINUTE);
             s = date.get(Calendar.SECOND);
-            return m + " m " + s + " s";
-        } else if (tInS > 0.9) {
+            return m + " min " + s + " s";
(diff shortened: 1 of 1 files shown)


dominikl:
--rebased-from #4195

jburel:
not sure it is required to have 5min 0 s but since it is a rebased.
Note that some work required in web to match the changes in insight see screenshot
(user-3 ImageID 23151)
pr-4199-image-23151

jburel:
To avoid confusion with data displayed in the viewer (status bar) we should display t+1  instead of t
and no longer used the term index

cc @gusferguson 
This should be done in a follow-up PR

dominikl:
That's strange. The goal of the PR was actually to match Web's display, according to @will-moore 's PR  #4131

jburel:
@dominikl: @will-moore has not rebased his PR yet.

jburel:
Card for "units" related work. https://trello.com/c/glNCdpXF/70-units-display

dominikl:
Removing the term "index" and starting t with 1 instead of 0 would be a really tiny change; is it Ok to do it as another commit here, although this is a rebased PR @jburel ?

jburel:
cleaner in another PR since it is a rebased one.

PR #4191, opened 2015-09-18
Ticket 13030

aleksandra-tarkowska:
This should prevent python clients using OMERO.py gateway against older server where the following properties are not present.

To test:
 - go to trout/merge choose seabass and log in as non admin user
 - check if you were successfully logged in and there were no SecurityViolation exception thrown.
 - log out and log in to merge
 - check if interpolation is on in webviewer

cc: @sbesson @joshmoore @will-moore

PR #4200, opened 2015-09-23
Ticket 13030 (rebased onto develop)

aleksandra-tarkowska:
This is the same as gh-4191 but rebased onto develop.

----

This should prevent python clients using OMERO.py gateway against older server where the following properties are not present.

To test:
 - go to trout/merge choose seabass and log in as non admin user
 - check if you were successfully logged in and there were no SecurityViolation exception thrown.
 - log out and log in to merge
 - check if interpolation is on in webviewer

cc: @sbesson @joshmoore @will-moore

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -29,7 +29,7 @@ from omero.util.decorators import timeit
 from omero.cmd import Chgrp2, Delete2, DoAll, SkipHead
 from omero.cmd.graphs import ChildOption
 from omero.api import Save
-from omero.gateway.utils import ServiceOptsDict, GatewayConfig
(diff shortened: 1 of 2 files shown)


aleksandra-tarkowska:
--rebased-from #4191

sbesson:
Tested as follows:
- connected from eel/latest (without this PR) to seabass/5.2 (5.2.0-m0) as non root user - 500 error thrown due to SecurityViolation thrown
- connected from eel/merge (with this PR) to seabass/5.2 (5.2.0-m0) as non root user . Webclient throws a 500 error due to the incompatibility between the current state of develop (which included the jstree work) and the 5.2.0-m0 tag (without the jstree work). Accessing an image viewer directly using `/weclient/img_details` displays the image without error and the interpolation box is present and set.

Merging.

PR #4202, opened 2015-09-23
Make sure the same pixelsize unit is used for all dimensions

dominikl:
Bugfix: [Ticket 13034](http://trac.openmicroscopy.org/ome/ticket/13034)

Test: Check the pixelsizes for a few images (e.g. *.dv), especially the image mentioned in the ticket (nightshade image id: 4045578 ) Make sure the pixel size uses the same unit for all dimensions XYZ (it's shown in the tooltip when you hover over the pixelsize label), and make sure the values are correct (especially for Z dimension) by looking at the original meta data (unit can differ from the original meta data that's expected).

Note: The problem (wrong Z dimension unit) occured, when there was a two figure number for X, like in the example 66nm, but a three figure number for Z (200nm).

--no-rebase

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/PropertiesUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/PropertiesUI.java
@@ -710,28 +710,24 @@ public class PropertiesUI
     	String tooltip = "<html><body>";
     	if (dx != null && dx.doubleValue() > 0) {
     		value += nf.format(dx);
-    		tooltip += "X: "+x+"<br>";
+    		tooltip += "X: "+x.getValue()+" "+LengthI.lookupSymbol(x.getUnit())+"<br>";
     		label += "X";
     	}
     	if (dy != null && dy.doubleValue() > 0) {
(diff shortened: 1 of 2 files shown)


jburel:
minor formatting issue: ```if(unit == null)``` => ```if (unit == null)```

jburel:
it works as expected. 
Matching changes to be implemented in web cc @will-moore 
I will add them to a card.

jburel:
Card for "units" related work. https://trello.com/c/glNCdpXF/70-units-display

jburel:
@pwalczysko: noticed the legacy "red" color when viewing the following image: user-4 23455
This can probably be fixed in this PR

dominikl:
Another commit to come (removing the red color), not ready for merge yet.

dominikl:
Test: 
- Select image 23455 (user-4), font showing pixels size should be same color as anything else (was previously red for this image).
- Tooltip when hovering over pixels size label in general tab should now use the unit symbol (not the name).
- Central panel: Tooltip should now use the same format (two decimal places) like the general metadata tab.

pwalczysko:
Works as expected (both the unit is okay from the header of this PR as well as the red color is gone and the tooltip is as expected).
Ready to merge

PR #4205, opened 2015-09-24
Remove all JAR signing items from the build

sbesson:
See https://trello.com/c/sqkNtuBY/58-drop-webstart

With Webstart being dropped for 5.2, we do not need all the logic to sign JARs under lib/insight anymore. This commit is removing all the signing steps in the build system and the local.properties.

--no-rebase


sbesson:
See https://ci.openmicroscopy.org/job/OMERO-DEV-merge-build/ICE=3.5,jdk=8_LATEST,label=octopus/100/consoleText

joshmoore:
WFM

dominikl:
Checked the build.xml, looks fine; also ran a local build, no problems.

PR #4190, opened 2015-09-18
Register imageJ agent.

jburel:
This PR sets the correct agent when insight is run as an imageJ plugin.
cc @aleksandra-tarkowska

PR #4212, opened 2015-09-24
Imagej agent develop

jburel:
same as gh-4190

Register the imagej plugin and importer agent correctly
* Check that the agent is correctly set in session table
* Check that we have one hit in the record

cc @dominikl @aleksandra-tarkowska 

--rebased-from #4190


sbesson:
Tested using Fiji started with jdk8 (following instructions under http://imagej.net/FAQ#How_do_I_setup_a_launcher_app_for_OS_X_for_running_with_a_different_JVM_version.3F). OMERO.ij plugin connects to a 5.2.x server without issues. The agent is correctly set and recorder and each reconnection creates a new hit. Merging.

PR #4215, opened 2015-09-25
Roi text background

jburel:
Remove the "yellow" background when displaying comment on some rois.
 
To test:
 * Draw a polygon or polyline
 * Add a comment
 * Make sure that no "yellow" rectangle is displayed.

cc @bramalingam


bramalingam:
Tested locally, works as expected. Will retest on Monday with the merge build.

bramalingam:
Everything works as expected. But, in the current implementation the default color of the Roi-Comment-Text is Black and cant be changed. This makes it difficult to read, when most of the images have a black background. If the font color can mimic the ROI stroke color, that would be perfect.

gusferguson:
@jburel 

Tested using OMERO.insight-5.2.0-m0-1116-7e3169b-ice35-b102-win on Windows 10 eel user-3.

Behaves as expected - text colour changes with Line Color.
Good to merge.

PR #4217, opened 2015-09-25
Add `omero admin checkupgrade`

manics:
Add `omero admin checkupgrade`

This allows scripts to automatically determine whether an updated server is available. Sets exit code 1 if so. To test.... you'll have to override the current server version somehow, or trust the integration test.

If you want to be thorough check an invalid `omero.upgrades.url` sets an exit code that's not 0 or 1.

~~TODO. Since this is going in to 5.2 should we set a specific agent to distinguish upgrade checks from server startups. e.g. `OMERO.status`, `OMERO.upgrade`, ... ?~~

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -54,6 +54,8 @@ except ImportError:
 
 DEFAULT_WAIT = 300
 
+CHECKUPGRADE_USERAGENT = "test"
+
 HELP = """Administrative tools including starting/stopping OMERO.
 
 Environment variables:
@@ -1794,12 +1796,11 @@ OMERO Diagnostics %s
         """
 
         config = config.as_map()
-        agent = 'server'
         upgrade_url = config.get("omero.upgrades.url", None)
         if upgrade_url:
-            uc = UpgradeCheck(agent, url=upgrade_url)
+            uc = UpgradeCheck(CHECKUPGRADE_USERAGENT, url=upgrade_url)
         else:
-            uc = UpgradeCheck(agent)
+            uc = UpgradeCheck(CHECKUPGRADE_USERAGENT)
         uc.run()
         if uc.isUpgradeNeeded():
             self.ctx.die(1, uc.getUpgradeUrl())
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1794,9 +1794,12 @@ OMERO Diagnostics %s
         """
 
         config = config.as_map()
-        upgrade_url = config.get(
-            "omero.upgrades.url", "http://upgrade.openmicroscopy.org.uk")
-        uc = UpgradeCheck('server', url=upgrade_url)
+        agent = 'server'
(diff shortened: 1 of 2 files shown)


joshmoore:
`diagnostics` exists. Not sure we need/want another one. @aleksandra-tarkowska?

manics:
I'm after an easy way to know whether an upgrade is available from a script, for this PR I'm setting an exit code but if `diagnostics` has some standard parseable output that would do too.

joshmoore:
Sorry, no objections to this change, I just meant use of the existing `diagnostics` upgrade check name.

manics:
You mean making this a sub-command: `diagnostics --check-upgrade` ?

joshmoore:
No, I'm only looking for the appropriate value for "server" in https://github.com/openmicroscopy/openmicroscopy/pull/4217/files#diff-57f0d9cb121726c674f02fd0bc23bd1cR1794

manics:
OK :-). Though now that it's come up including this in another command instead of creating another top level one might be a good idea....

manics:
Actually, I guess another subcommand of `admin` is't too bad.

@aleksandra-tarkowska What user-agent should I use when checking whether an upgrade is available?

aleksandra-tarkowska:
@manics there is OMERO.test

joshmoore:
@aleksandra-tarkowska : but we ignore those, no? This would be used by actual scripts in (more or less) production.

aleksandra-tarkowska:
No we don't ignore. There is OMERO.test agent in production DB. Hit should be saved

jburel:
@manics: are you still thinking of having this PR in 5.2.0?

manics:
@jburel It'd be nice but not essential.

manics:
@sbesson Integration test added (is there a nicer way of setting the ctx dir?)

manics:
Note I'm now importing `UpgradeCheck` at the top level: https://github.com/manics/openmicroscopy/blob/admin_checkupgrade/components/tools/OmeroPy/src/omero/plugins/admin.py#L38

However I've left https://github.com/manics/openmicroscopy/blob/admin_checkupgrade/components/tools/OmeroPy/src/omero/plugins/admin.py#L1080 in because I'm not sure what's going on with the logging config. @aleksandra-tarkowska ?

manics:
For compelteness it'd be nice to test an invalid url (e.g. `example.org/invalid/checkupgrade`) and assert `rv > 1` to be sure a failed connections is not mistaken for an upgrade

sbesson:
Tested the latest state of this PR with a server with a non-default `omero.upgrades.url` set. `bin/omero admin checkugprade` calls to the correct upgrade URL. Tests under https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-python/112/testReport/OmeroPy.test.integration.clitest.test_admin/TestAdmin/ are also passing.
Assuming we all agree with the agent used (`OMERO.test`), this looks ready to merge /cc @joshmoore @aleksandra-tarkowska . We can review the `upgradecheck` usage in the `diagnostics` command in 5.2.1.

joshmoore:
Agreed. Not critical. It will likely have low usage initially.

jburel:
Thanks all merging.

PR #4218, opened 2015-09-25
Webstart web

jburel:
Remove Java webstart from web component.


aleksandra-tarkowska:
I am sorry @jburel but removing webstart siewny mean that removing webstart django application. that PR will break many features! We have other applications using this code. Let me do it right.

jburel:
@aleksandra-tarkowska: ok.Closing this PR and updating the card.

jburel:
in theory nothing else should rely on the "webstart" code. if it does,  it is probably a good time to clean/refactor.

PR #4219, opened 2015-09-28
Selection browser

jburel:
Fix selection problem noticed by @mporter-gre 
Nodes not visible in central panel were still selected in tree
to test this PR:
 * log in as user-3
 * Browse dataset non-archivedv
 * in central panel, filter by name using for example R3D.
 * 6 images out of 9 should be displayed.
 * Select all 6 in the central panel
 * Check that  the image ```/home/hudson/for_import_to_howe/non-archived-imoprt-case/user-3/read-only-1/no_projects/non-archivedv/off_by_1_a.dv``` is not selected in the tree. It is not visible in the central panel.

cc @pwalczysko


pwalczysko:
Works as expected.
Ready to merge.

sbesson:
Looking at https://github.com/openmicroscopy/openmicroscopy/pull/4219/files?diff=split&w=1, the diff is really a one-line in `BrowserModel.java`. Merging.

PR #4220, opened 2015-09-28
Formatting timestamps (rebased onto develop)

will-moore:
This is the same as gh-4131 but rebased onto develop.

----

This fixes a bug in python 2.6 when comparing Decimal with floats and also other bugs in the calculation of times for formatting.
https://trac.openmicroscopy.org/ome/ticket/13016

To test, user trout (python 2.6).
 - Find Image 24551 (user-3), which has timepoints every 5 minutes.
 - Look at Delta T times, Under POL channel on Acquisition tab.
 - Should see human readable time values.
 - Check that with timestamps over an hour, that minutes are shown correctly (don't go over 59!)
 - We also round() the numbers before truncating to integers for display, but the only way to see the original values is to download the log file.

Before:
screen shot 2015-09-04 at 10 51 10

After:
screen shot 2015-09-04 at 11 11 06


will-moore:
--rebased-from #4131

ximenesuk:
This works fine, and I'm seeing the real 'after' of `1 min 0 s` rather than the old 'after' in the description.

Good to merge.

PR #4223, opened 2015-09-28
Big image export jpg 12982 develop

will-moore:
This is #4172 re-based to develop:

---------------------------------

This limits the download as jpeg / png / tiff to images of 12k * 12k or smaller.
This cut-off was determined by trying to download images of various sizes.

To test:
 - Images bigger than 12k * 12k pixels (144000000) should be disabled for Download as Jpg, Png & Tiff.
 - Configure with a different number of pixels. E.g. 3000 x 3000 = 9000000
```
$ bin/omero config set omero.client.download_as.max_size 9000000
$ bin/omero admin restart
```
Now, images larger than this new size should be disabled for download.

NB: The 5.1 PR #4172 doesn't include the last 'revert' commit that adds the config parameter above. Could rebase this if we want to?

--rebased-from #4172

### DIFF
--- a/components/common/resources/ome/config.xml
+++ b/components/common/resources/ome/config.xml
@@ -224,7 +224,7 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
-    <bean class="ome.system.Preference" id="omero.client.download_as_jpg.max_size">
+    <bean class="ome.system.Preference" id="omero.client.download_as.max_size">
        <property name="db" value="false"/>
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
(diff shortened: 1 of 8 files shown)


sbesson:
@will-moore: from https://github.com/openmicroscopy/openmicroscopy/pull/4172#issuecomment-142217909 and https://trello.com/c/4mDNygGv/38-follow-up, my understanding is that ef2e1f75d2758faab68dd60719d778b35c09fb2e requires a join effort on the Insight side. This would make this PR essentially unmergeable until we have everything in place across clients. It is also slightly confusing in terms of the rebase semantics as the PR is adding extra content/feature.

I would suggest to force push ef2e1f75d2758faab68dd60719d778b35c09fb2e away from this PR so that we can review and merge #4223 and then open a dedicated PR for the new size limit configuration property.

will-moore:
Removed last commit as suggested. Now the download as jpg/png/tiff should be disabled for Big images but not others.

sbesson:
Teste on eel/merge under user-1 with a 4096 x 4096 micrograph image. All export options except OME-TIFF are greyed out. All export options are allowed for a 2048 x 2048 micrograph image.

PR #4224, opened 2015-09-28
Right panel fileanns slow 13022 (rebased onto develop)

will-moore:
This is the same as gh-4176 but rebased onto develop.

----

See http://trac.openmicroscopy.org/ome/ticket/13022

This fixes the slow loading of right panel "General" tab when the P/D/I has several file annotations.
Two factors were causing the poor performance:
 - Loading annotations didn't also load the files, so that ann.getFile() created a new ```OriginalFileWrapper``` with an unloaded file, causing the file to be loaded each time on the fly. Now, we include files when loading annotations.
 - Further poor performance was caused by not caching the file for the FileAnnotationWrapper, so that repeated calls to getName() or getFileSize() were loading the file each time. We now cache the file in ann.getFile()

To test:
 - Compare loading speed of right panel for P/D/I with lots of file annotations vv none. Should be about the same.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4581,6 +4581,10 @@ class FileAnnotationWrapper (AnnotationWrapper, OmeroRestrictionWrapper):
 
     OMERO_TYPE = FileAnnotationI
 
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


will-moore:
--rebased-from #4176

gusferguson:
@will-moore 

Tested with eel merge user-5

Behaves as expected - the slight difference in time is in keeping with the increase in drawing required when many attachments present.
Good to merge.

PR #3933, opened 2015-07-03
Handle new ROI Points string format (rebased onto metadata)

manics:
This is the same as gh-3932 but rebased onto metadata.

----

There are (at least) 2 formats for storing Points in ROI Polygons. The webgateway only had one of them.

https://trello.com/c/hJLlf1Xr/47-error-on-preview-rois

Are there any other formats?

PR #4225, opened 2015-09-28
Migrate more server configuration files to the templates directory (rebased onto metadata)

manics:
This is the same as gh-4093 from @sbesson but rebased onto metadata with additional commits.

----

See https://trello.com/c/lHVF6Olb/36-switch-between-prefixed-ports.

This PR reviews server deployment configuration in the same manner as `bin/omero admin jvmcfg` did. Most generally, this PR attempts at differentiating the staging configuration (`config.xml` + `etc/templates`) from the live/running configuration (under `etc`) and clarify the scope of the various `admin` subcommands.

Summary of changes:
- Migrates Ice configuration files previously modified live by `admin ports` and `templates.xml` under `etc/templates`. Renames `generated.xml` as `templates.xml` as we do not have any more conflict name
- adds the logic to regenerate all the configuration files (including JVM settings and ports) and expose it in a `bin/omero admin rewrite` command. Only regenerates templates if no server is running by default.
- modifies the deployment subcommands `omero admin [start|restart|startasync|deploy]`  to use the new `rewrite` subfunction
- modifies the scope of  `bin/omero admin jvmcfg` to display the JVM settings without regenerated the `etc/grid/templates.xml` and `bin/omero admin diagnostics` to read the live memory settings under `etc/grid/templates.xml` instead of recalculating them from `etc/templates/grid/templates.xml` and `etc/grid/config.xml`
- adds new `omero.ports.xxx` properties controlling the ports on which the server applications are deployed and drops `bin/omero admin ports` in favor of this configuration
- extend `test_admin.py` unit tests to cover the new writing stratedy and fix other unit tests requiring `ICE_CONFIG`

This PR will required coupled deployment PR for the testing on the CI server and should be considered as breaking from a deployment perspective for servers using non-standard ports.

Still to do:
- [x] expose `--force-rewrite` option in deployment command

/cc @joshmoore @manics @aleksandra-tarkowska 

----

Additional commits: Make server var dir configurable
Todo:
- [ ] property name (as always)
- [ ] apache/nginx templates
- [ ] Tests 
- [ ] `windefault.xml` contains `<variable name="OMERO_LOGS" value="${OMERO_HOME}\\var\\log\\"/>`, Is there anything we can do about this?


manics:
--rebased-from #4093

manics:
Not ready for merging, but should be sufficient to test an alternative `var` location.

https://api.travis-ci.org/jobs/82624913/log.txt?deansi=true says deprecation warnings are missing. Is something out of sync?

sbesson:
@manics: was https://github.com/openmicroscopy/openmicroscopy/pull/4194 ported to this branch (or `metadata` fast-forwarded to 5.1.4)? /cc @joshmoore

manics:
Thanks, cherry-picked https://github.com/sbesson/openmicroscopy/commit/de9a4072ba2fe8be6fb5d3e7a470133dcd743b78

joshmoore:
As with gh-3933, since this won't make it into 5.1 and metadata will need to track it for a while longer, merging this if there are no objections.

joshmoore:
(waiting on conflict resolution)

manics:
rebased & force-pushed

joshmoore:
Merging for 0.0.3

PR #4237, opened 2015-10-01
Rename local.properties.example as build.properties

sbesson:
See https://trello.com/c/EnGt2XDX/73-rename-local-properties-example

### DIFF
--- a/etc/build.properties
+++ b/etc/build.properties
@@ -2,9 +2,9 @@
 # build.properties
 # ----------------
 # System variables which will be used at build-time.
-# Any of these values can be overwritten by creating
-# a file called etc/local.properties and editing them
-# in this local file.
+# Any of these values can be overwritten by creating a new file called
+# etc/local.properties and then adding the variables to be overwritten to this
+# new file.
 #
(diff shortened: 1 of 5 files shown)


etc/build.properties
```diff
@@ -1,10 +1,10 @@
...
-# system variables which will be used at build-time
-# Any of these values can be overwritten by copying
-# this file to etc/local.properties and manually
-# editing it.
+# build.properties
+# ----------------
+# System variables which will be used at build-time.
+# Any of these values can be overwritten by creating
+# a file called etc/local.properties and editing them
+# in this local file.
```
ximenesuk: Maybe slight rewording here?

Any of these values can be overwritten by creating a new file called etc/local.properties and then adding the variables to be overwritten to this new file.

ximenesuk:
Other than the wording above this looks fine, no problems building and good to have the change!

sbesson:
Description update pushed.

ximenesuk:
Given the second commit is a change to a comment this is good to merge once Travis is happy.

PR #4241, opened 2015-10-04
Show image in well

will-moore:
Fix for https://trello.com/c/XPrUXfTV/49-browse-spw-image-in-search-results-doesn-t-work
cc @manics.
Fixes show=?image-123 when the image is in a well - We browse to the well - Not necessarily the correct well-index (just first index).
This is the first fix I found for this that seems to work (maybe not the 'proper' fix, and no tests etc - in fact it's possible that 'show' tests may fail - haven't tried them yet).


joshmoore:
Merging for 0.0.3. Not sure of what will need to happen for 5.2.x

PR #4243, opened 2015-10-05
Don't reload image after dis/enabling interpolation

dominikl:
Fixes [Ticket 13052](https://trac.openmicroscopy.org/ome/ticket/13052)

Test: 
- Check that toggling the interpolation option still works (open a very small image and zoom in to the maximum in order to see an effect). 
- Check that toggling the interpolation option *does not* request the image from the server again (open a big, tiled image; if the image is reloaded you should see the tiles gradually building up again; this should *not* happen when the interpolation option is changed).


jburel:
Noticed that the ```set/get``` interpolation method in model are public. they only need to be packaged visible.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.0-m1-138-5a57c5e-ice35-b107-mac eel user-3 read-only-1

Behaves as expected.
Good to merge.

jburel:
Thanks Merging

PR #4245, opened 2015-10-06
Configuration templates (part 2)

sbesson:
Follow-up of https://github.com/openmicroscopy/openmicroscopy/pull/4093, this PR:

- implements the substitutions of `@Ice.Default.Host@` for the configuration templates
- suppresses the help of `admin ports` 
- moves the Web templates to a subdirectory for clarity. The `web config` subcommand and the unit tests are updated accordingly

The CI servers will test the deployment of the servers on a non-default `Ice.Default.Host`. This PR should also be tested locally with `Ice.Default.Host` unset. Additionally, the Web configuration generation should be tested to make sure nothing was broken by the templates migration.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1011,6 +1011,7 @@ present, the user will enter a console""")
             '@omero.ports.tcp@': config.get('omero.ports.tcp', '4063'),
             '@omero.ports.registry@': config.get(
                 'omero.ports.registry', '4061'),
+            '@Ice.Default.Host@': config.get('Ice.Default.Host', '127.0.0.1')
             }
 
         def copy_template(input_file, output_dir):
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -37,7 +37,7 @@ from omero.plugins.prefs import \
 
 from omero_ext import portalocker
 from omero_ext.which import whichall
-from omero_ext.argparse import FileType
+from omero_ext.argparse import FileType, SUPPRESS
 from omero_version import ice_compatibility
 
 try:
@@ -320,24 +320,7 @@ dt_socket,address=8787,suspend=y" \\
             "--finish", action="store_true",
             help="Re-enables the background indexer after for indexing")
 
-        ports = Action(
-            "ports",
-            """Allows modifying the ports from a standard OMERO install (deprecated)
-
-To have multiple OMERO servers running on the same machine several ports
-must be modified from their defaults. Changing the ports on a running server
-will be prevented, use --skipcheck to override this.
-
-Examples:
-
-  # Set ports to registry:14061, tcp:14063, ssl:14064, web:14080
-  %(prog)s --prefix=1
-  # Set ports back to defaults: 4061, 4063, 4064, 4080
-  %(prog)s --prefix=1 --revert
-  # Set ports to: 4444, 5555, 6666, 7777
-  %(prog)s --registry=4444 --tcp=5555 --ssl=6666 --webserver=7777
-
-""").parser
+        ports = Action("ports", SUPPRESS).parser
         ports.add_argument(
             "--prefix",
             help="Adds a prefix to each port ON TOP OF any other settings")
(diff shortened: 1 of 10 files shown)


components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -320,24 +320,7 @@ def __init__(this, name, help, wait=False):
...
-
-  # Set ports to registry:14061, tcp:14063, ssl:14064, web:14080
-  %(prog)s --prefix=1
-  # Set ports back to defaults: 4061, 4063, 4064, 4080
-  %(prog)s --prefix=1 --revert
-  # Set ports to: 4444, 5555, 6666, 7777
-  %(prog)s --registry=4444 --tcp=5555 --ssl=6666 --webserver=7777
-
-""").parser
+        ports = Action("ports", SUPPRESS).parser
```
aleksandra-tarkowska: ``bin/omero admin ports`` are deprecated. Why just not add warning to the help as well.

sbesson: This came from [this discussion](https://github.com/openmicroscopy/openmicroscopy/pull/4093#commitcomment-13534609). The main point is that this command is now a no-op and adding `(deprecated)` to the help does not convey exactly that messafe.

aleksandra-tarkowska:
Looks good to me

sbesson:
Last commit should fix the failing unit test in https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-matlab/57/

mtbc:
OMERO-DEV-merge-matlab has been passing lately.

jburel:
Such properly could potentially be used in insight container.

jburel:
Some tests have hard coded, localhost. do you plan to modify them in this PR and take advantage of the variable?

sbesson:
@jburel: which tests do you have in mind specifically? The `Ice.Default.Host` variable is a server-side property. For client/test consumption, generally I would expect to read the hostname directly from `ice.config`.

jburel:
Sorry I should have made it clearer, I was thinking of test that did not read directly from ice.config.

jburel:
All good. tests are passing. I need to check if not set. I will check with @aleksandra-tarkowska

jburel:
Question:
in ```etc/templates/ice.config```, you set the ```omero.host``` to ```@Ice.Default.Host@````
Do we want to do the same for ```omero.fs.host``` in ```etc/templates/grid/templates.xml ````?

sbesson:
It would make sense to me. Is `omero.fs.host` expected to be modified by configuration property @ximenesuk?

sbesson:
As pointed out by @aleksandra-tarkowska, OMERO.py and OMERO.web are currently relying on the presence of a configuration file under `etc/ice.config`. The last commits restore a default file with values `localhost/4064` which:
- is consumed by the build system for packaging OMERO.matlab and OMERO.py
- removes the burden of maintaining the substitution logic in the build system
- re-allows the following workflow  `./build.py build-dev && dist/bin/omero web start`

The other functionalities of the CLI admin plugin should be unaffected, i.e. the `etc/ice.config` file should be rewritten when calling `bin/omero admin rewrite` or `bin/omero admin start`

ximenesuk:
@sbesson I don't fully understand the question. Do you mean modified by `omero config`?

sbesson:
@ximenesuk: yes, is `omero.fs.host` an internal variable only or could it be adjusted via `omero config`. We do not expose it in the documentation.

ximenesuk:
@sbesson Yes, it's not exposed on the DropBox docs - whether that makes it internal or not I'm not sure. Thoughts @joshmoore ?

joshmoore:
The most likely use of that would be if someone wanted to run `DropBox` elsewhere:

 * `wget OMERO.server.zip`
 * ...edit `etc/grid` ...
 * `bin/omero config set omero.fs.host foo`
 * `bin/omero node dropbox`

sbesson:
Unless there is a clear blocker on Dropbox and assuming the last commits make sense,I would think we can get this PR as it is and include Dropbox as part of 5.1.0-rc1 testing.

joshmoore:
WFM

ximenesuk:
...and me.

PR #4246, opened 2015-10-07
IDR web tweaks

ximenesuk:
These are quick fixes that can be done more nicely further down the line.
 + Disable Show Rois (hard-coded)
 + Default to “All users” for search (hard-coded)
 + Map annotation above Table of annotations

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
@@ -963,11 +963,7 @@
 {% block roi_buttons %}
           <div id="roi_controls">
                 <h1>ROI Count: {{ roiCount }}</h1>
-                {% ifnotequal roiCount 0 %}
-                  <!-- 'link is enabled on imageLoad callback -->
-                    <a id="show-rois-a" href="#" style="color:gray">Show ROIs</a> |
-                    <a id="hide-rois-a" style="color:gray">Hide</a>
-                {% endifnotequal %}
+                <!-- removed show/hide rois buttons -->
           </div>
 {% endblock roi_buttons %}
           <div class="odd row">
--- a/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
@@ -967,7 +967,7 @@
                   <!-- 'link is enabled on imageLoad callback -->
                     <a id="show-rois-a" href="#" style="color:gray">Show ROIs</a> |
                     <a id="hide-rois-a" style="color:gray">Hide</a>
-                {% endifnotequal %}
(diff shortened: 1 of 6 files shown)


components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
```diff
@@ -963,11 +963,7 @@ <h3 class="can-collapse" id="bulk-annotations" style="display: none;">Bulk Annot
 {% block roi_buttons %}
           <div id="roi_controls">
                 <h1>ROI Count: {{ roiCount }}</h1>
-                {% ifnotequal roiCount 0 %}
```
aleksandra-tarkowska: Please use operator https://docs.djangoproject.com/en/1.6/ref/templates/builtins/#id3

aleksandra-tarkowska:
@joshmoore searching by ID is https://github.com/ximenesuk/openmicroscopy/blob/idr-minor-web-tweaks/components/tools/OmeroWeb/omeroweb/webclient/views.py#L770

aleksandra-tarkowska:
hold on, but if map annotation contains number both record should be displayed, Image with ID and the other Object that contains that map annotation. If you do not see that, that means map annotation is not found regardless of that code

components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
ximenesuk: I'm happy to fix this if 1000 (or some other number is suitable).

joshmoore:
@aleksandra-tarkowska : agreed. There were other cases though where the ID matches swamped the found map annotation. I'm re-indexing now so we should be able to test.

cc: @ximenesuk for https://github.com/ximenesuk/openmicroscopy/blob/idr-minor-web-tweaks/components/tools/OmeroWeb/omeroweb/webclient/views.py#L770

aleksandra-tarkowska:
Urlizer is your friend https://docs.djangoproject.com/en/1.6/ref/templates/builtins/#urlize

aleksandra-tarkowska:
To add url to map annotation you need to modify  https://github.com/ximenesuk/openmicroscopy/blob/idr-minor-web-tweaks/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/mapannotation.html#L59

```
            <td>{{ row.0|urlize|escape }}</td>
            <td>{{ row.1|urlize|escape }}</td>
```
This will only work when map annotation is loaded from the server and template is rendered. Link will not appear when you add value in web UI.

components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
aleksandra-tarkowska: Fixed in https://github.com/openmicroscopy/openmicroscopy/pull/4249/

joshmoore:
Merging for 0.0.3

PR #4248, opened 2015-10-07
Restrict size for export as jpg/png/tif

dominikl:
Same as #4172 , just for Insight. Limits the "Exports as JPG/PNG/TIF" actions to a certain maximum image size.

Test: 
- Check that the "Export as ..." menus are enabled for images < 12k x 12k (that's the default limit if nothing else has been set).
- Check that they are disabled for larger images.
- Set a custom limit, e. g. 1k x 1k `./omero config set omero.client.download_as_max_size 1000000` (followed by `./omero admin restart` ) and perform above mentioned checks again.


dominikl:
Replaced by #4254

PR #4249, opened 2015-10-07
Roi count limit

aleksandra-tarkowska:
This PR prevents showing ROI button in a viewer if number of ROIs is bigger then limit.

To test:
 - add more then 10000 ROIs and check if `Show ROIs` in web is not visible
 - check image with less then 10000 ROIs if `Show ROIs` in web is visible
 - check image with no ROIs if `Show ROIs` in web is not visible

Test image with ~1500 http://eel/merge/webclient/?show=image-24083 user-1

Insight may do the same? cc @dominikl

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1532,8 +1532,8 @@ class _BlitzGateway (object):
 
     def getRoiLimitSetting(self):
         try:
-            roi_limit = (self.getConfigService().getConfigValue(
-                         "omero.client.viewer.roi_limit") or 100)
+            roi_limit = (int(self.getConfigService().getConfigValue(
+                             "omero.client.viewer.roi_limit")) or 100)
         except:
             roi_limit = 100
         return roi_limit
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1533,9 +1533,9 @@ class _BlitzGateway (object):
     def getRoiLimitSetting(self):
         try:
             roi_limit = (int(self.getConfigService().getConfigValue(
-                             "omero.client.viewer.roi_limit")) or 100)
+                             "omero.client.viewer.roi_limit")) or 2000)
         except:
-            roi_limit = 100
+            roi_limit = 2000
         return roi_limit
 
     def getInitialZoomLevel(self):
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1530,6 +1530,14 @@ class _BlitzGateway (object):
                 int(c.getConfigValue('omero.pixeldata.max_plane_height')))
         return self._maxPlaneSize
 
+    def getRoiLimitSetting(self):
+        try:
+            roi_limit = (self.getConfigService().getConfigValue(
+                         "omero.client.viewer.roi_limit") or 100)
+        except:
+            roi_limit = 100
+        return roi_limit
+
     def getInitialZoomLevel(self):
         """
         Returns default initial zoom level set on the server.
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -151,11 +151,11 @@ def imageMarshal(image, key=None, request=None):
 
     server_settings = request.session.get('server_settings', {})
 
-    init_zoom = server_settings.get('initial_zoom_level' , 0)
+    init_zoom = server_settings.get('initial_zoom_level', 0)
     if init_zoom < 0:
         init_zoom = levels + init_zoom
 
-    interpolate = server_settings.get('interpolate_pixels' , True)
+    interpolate = server_settings.get('interpolate_pixels', True)
 
     try:
         rv.update({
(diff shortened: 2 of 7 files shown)


dominikl:
@aleksandra-tarkowska could you also add the 'download_as_max_size' property to the server config? See @will-moore 's PR #4223 
```
<bean class="ome.system.Preference" id="omero.client.download_as_max_size">
       <property name="db" value="false"/>
    </bean>
```
I wanted to do that in my PR #4248 but I guess that's going to conflict with this PR then.

components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
```diff
@@ -149,17 +149,13 @@ def imageMarshal(image, key=None, request=None):
...
 
-    try:
-        interpolate = request.session['server_settings']['interpolate_pixels']
-    except:
-        interpolate = False
+    init_zoom = server_settings.get('initial_zoom_level', 0)
+    if init_zoom < 0:
+        init_zoom = levels + init_zoom
+
+    interpolate = server_settings.get('interpolate_pixels', True)
```
ximenesuk: Does this mean the default for `interpolate` has been changed to `True`?

ximenesuk:
Everything seems to work fine. Tested around the 100 limit and at zero ROIs,

ximenesuk:
This is good to merge!

aleksandra-tarkowska:
Hold on, just realized that this PR cause error

aleksandra-tarkowska:
To test last commit, please go to figure and try to add image.

will-moore:
Seems that 100 ROIs is not very many. I would have thought we could manage a larger number before we start seeing problems? @manics Any feeling for what would be a good default number?

ximenesuk:
I had no problems adding images to Figure. I'll leave the discussion of what the default config should be to someone else.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1530,6 +1530,14 @@ def getMaxPlaneSize(self):
                 int(c.getConfigValue('omero.pixeldata.max_plane_height')))
         return self._maxPlaneSize
 
+    def getRoiLimitSetting(self):
+        try:
+            roi_limit = (self.getConfigService().getConfigValue(
+                         "omero.client.viewer.roi_limit") or 100)
+        except:
+            roi_limit = 100
+        return roi_limit
```
will-moore: getConfigService().getConfigValue() - Does this return a string? If so, then you'd get a string returned from getRoiLimitSetting() if this worked, or 100 if it didn't. Should probably be a number in both cases?

joshmoore: > Does this return a string? 

Definitely.

ximenesuk: Yes, looking at other usage it is cast to an int/long when necessary. Sorry, I forgot to check actually setting the config value as I was using eel.

manics:
> Any feeling for what would be a good default number?

There could be any number of ROIs, what's the maximum you'd expect a web browser to handle?

will-moore:
@manics That's what I was hoping you'd know. I've never worked with enough ROIs / shapes to cause a problem in the web.

manics:
100-200 seems reasonable. What's the longer term plan- allow display of just the first N ROIs? Paging?

aleksandra-tarkowska:
limit is now 10,000

aleksandra-tarkowska:
I presume pagination will be the solution

manics:
Could you add a warning or tooltip to say ROIs can't be viewed?

manics:
I received a `Warning: Unresponsive script` dialog after clicking `Show ROIs` on https://eel.openmicroscopy.org/merge/webclient/img_detail/24083/

aleksandra-tarkowska:
@bramalingam you mentioned you use web to show this image, did you also get ``Warning: Unresponsive script``
firefoxscreensnapz010

will-moore:
Seemed that 1476 ROIs was handled OK for me (although very slowly).
The server can handle it (which was the major concern).
I guess some browsers can and some can't handle it, so 1500 is probably about the right limit.
We don't want to disable it for users who can display ROIs OK, since the "Unresponsive script" warning for those who can't isn't fatal.

screen shot 2015-10-14 at 10 14 49

joshmoore:
Kicked travis due to unrelated:
```
[nomemorytestng] FAILED: testBasicUsage on null(omero.util.test.TempFileManagerTest)
[nomemorytestng] org.testng.internal.thread.ThreadTimeoutException: Method org.testng.internal.TestNGMethod.testBasicUsage() didn't finish within the time-out 1000
[nomemorytestng] 	at sun.nio.ch.SharedFileLockTable.add(FileLockTable.java:137)
[nomemorytestng] 	at sun.nio.ch.FileChannelImpl.tryLock(FileChannelImpl.java:1057)
```

ximenesuk:
Referring to the other two comments it means setting a roiLimit of 10 will only allow 9 ROIs to be displayed. It would be better if the roiLimit was the maximum displayable ROIs.

I couldn't see the tooltips using either Chrome or FF with a locally running server. That might be a local problem though.

aleksandra-tarkowska:
ohh sorry, maths, thanks for spotting ;-)

ximenesuk:
The fix is fine and works correctly. I have had to test this locally again as it is not in the build, it conflicts with gh-4262 which is in the build. so cannot be merged until that is resolved.

aleksandra-tarkowska:
@ximenesuk looks like it is in the merge build now

ximenesuk:
Tested with a couple of images from `user-1`,

http://eel.openmicroscopy.org/merge/webclient/?show=image-24083 (ROIs < 2000)
http://eel.openmicroscopy.org/merge/webclient/?show=image-23953 (ROIs > 2000)

and everything works fine. With the typo in the tooltip fixed this should be good to merge.

aleksandra-tarkowska:
@ximenesuk spelling mistake fixed

ximenesuk:
Once Travis is happy I'd say this is now good to merge.

aleksandra-tarkowska:
thank you

PR #4250, opened 2015-10-07
Resolve ZeroC artifacts via OME artifactory

sbesson:
See https://trello.com/c/2qw5o7PL/27-clean-lib-repository

This PR removes the need for local ZeroC artifacts and resolves the Maven artifacts from the OME artifactory instead. To test this PR, check Travis and the various CI builds are green.


rleigh-dundee:
The changes look fine and all the jobs are green (the C++ jobs are still running but all the java jobs are green).  Please merge.

PR #4252, opened 2015-10-07
Tweak dusty.css which is misselecting on some rules

dpwrussell:
I think these grouped rules (and by the looks of [dusty.css](https://github.com/dpwrussell/openmicroscopy/blob/dusty-overselection/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/dusty.css), potentially a whole lot more) are meant to be selecting `.classname elementname` for each rule, or similar, but CSS selectors are not distributive. Instead only the first rule in the group has the `.classname` rule selector.

This is evident in a few places where this behaviour does not seem desirable. E.g. Edit the name of an image in the right panel, if you mousedown the `Save` and `Cancel` buttons, they have different behaviours. Also, when mousedown on `Save`, there is an unpleasant layout change which is much nicer to not have.

It is also very annoying to have overreaching CSS selectors from a plugins perspective because it is then necessary to restore the original behaviour in the plugins CSS.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/dusty.css
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/dusty.css
@@ -559,7 +559,9 @@ button::-moz-focus-inner {
 			margin-bottom:5px;
 		}
 
-	.standard_form input[type=text], input[type=password], textarea {
+	.standard_form input[type=text],
+	.standard_form input[type=password],
+	.standard_form textarea {
 		width:70%;
 		border: solid 1px rgba(0,0,0,.2);
 		border-top: solid 1px rgba(0,0,0,.3);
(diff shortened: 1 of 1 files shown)


dpwrussell:
And apparently I can't spell 'misselecting' in my commits!

will-moore:
The css code changes make a lot of sense.
I'm just wondering if there's anywhere where we were "accidentally" using these css selectors and might not look as nice now that they're more restrictive.

@pwalczysko Just keep an eye open for strange-looking form elements (text input, textarea, submit button etc).

dpwrussell:
I had a click-around and couldn't see anything that seemed different after the change other than the things that were actually fixed, but I'm sure @pwalczysko will be much more thorough!

pwalczysko:
Over couple of days of testing and also specific looking for issues now, no problems detected here. 
Ready to merge.

PR #4253, opened 2015-10-08
Roi count limit

dominikl:
Same as #4249 just for Insight. Limits the availability of the measurement tool to a certain number of ROIs (default: disabled if > 100 ROIs).

Testing: See above mentioned PR (except the measurement tool should be enabled in Insight, if there are no ROIs)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/events/metadata/ROICountLoaded.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/events/metadata/ROICountLoaded.java
@@ -1,0 +1,72 @@
+/*
+ * org.openmicroscopy.shoola.agents.events.FocusGainedEvent 
+ *
+ *------------------------------------------------------------------------------
+ *  Copyright (C) 2015 University of Dundee. All rights reserved.
+ *
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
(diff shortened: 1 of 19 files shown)


ximenesuk:
This works but is it really the intention? The web PR prevents viewing large numbers of ROIs. This PR prevents any access to the measurement tool so I can't delete, add or otherwise modify ROIs.

dominikl:
Good question... have to check with @jburel .

jburel:
The limitation in web is mainly due to the current implementation of the ROI support in web i.e. thumbnail generated for each shape.
I do not think this is required in insight.

gusferguson:
@dominikl @jburel @ximenesuk 

As far as I can tell there is absolutely no reason to have this in Insight. It is only Web that has issues with 100s of ROIs - surely Insight can cope with many thousands.

I know Simon crashed it with millions of ROIs, but the measurement tool should not be disabled for any less than millions surely. And there should probably be a warning notice that ROIs are not being displayed due to number rather than disabling the button.

I agree with Colin that this is an area where Insight and Web should agree to differ.

dominikl:
I remember @bramalingam 's default imagej workflow which creates ~1500 ROIs brings Insight's measurement tool quite to a limit (takes a couple of minutes to load). Shall I keep this PR but set a higher limit than 100 or close it anyway?

bramalingam:
Absolutely agree with @jburel , the issue with the web was the thumbnail generation.! Guess this PR might not be required at all for Insight, unless you start entering Simon's land..!

dominikl:
Ok, it's not too bad, takes about 30s to 1min to open ROI tool for that image. Closing this PR then...

PR #4254, opened 2015-10-08
Big image export

dominikl:
Same as #4223 , just for Insight. Limits the "Exports as" actions to a small images.

Test:
- Check that the "Export as ..." menus are enabled for small images < 3k x 3k 
- Check that they are disabled for larger images.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/EditorModel.java
@@ -4278,13 +4278,25 @@ class EditorModel
 	    return objects;
 	}
 
-	/**
-	 * Returns <code>true</code> if the image is a large image,
-	 * <code>false</code> otherwise.
-	 * 
-	 * @return See above.
-	 */
-	boolean isLargeImage() { return largeImage; }
+    /**
+     * Returns <code>true</code> if the image is a large image,
+     * <code>false</code> otherwise.
+     * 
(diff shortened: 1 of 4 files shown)


ximenesuk:
Just checking here, should all of the Export As... options be disabled or just OME-TIFF ?

dominikl:
AFAIK all of them. But it really got very confusing (e. g. the according Web PR says export should be enabled to up to 12k x 12k, but in fact it's actually just 3k x 3k, i. e. "big images"). But maybe @aleksandra-tarkowska  could confirm this again?

ximenesuk:
It looks okay for images - all Export As options are disabled (if that's the intention). I can't find a big image in a plate but the logic looks correct.

dominikl:
Just had a look at Web. It is possible to export big images as ome-tiff there, but it was always disabled for Insight. Sorry, don't get the logic here...

ximenesuk:
On web it looks like the download size is bigger than the default big image size, so some big images can be exported but not huge ones.

dominikl:
No, that's another point for confusion, the PR description mentions this config property, but it's not implemented that way. 
I have to correct myself, it's not possible to export ome-tiff for big images in web, just the option is enabled, but it fails with the message, that it can't export big images as ome-tiff. So this should be fixed in Web by disabling the option.

ximenesuk:
In that case I'll hold off on any further review until these issues are clarified.

dominikl:
Superfluous, will open another PR later to match #4262 .

PR #4255, opened 2015-10-08
Remove Ice 3.4 support

rleigh-dundee:
- Remove support from the build logic and configuration
- Remove C++ #ifdefs which contain unused code (most are actually Ice 3.3 support)


components/tools/travis-build
```diff
@@ -19,8 +19,26 @@ clean()
     ant clean
 }
 
+get_ice()
+{
+  ice=Ice-3.5.1-b1-ubuntu1204-amd64.tar.xz
+  if [ ! -f "download/$ice" ]; then
+    mkdir -p download
+    cd download
+    wget "http://users.openmicroscopy.org.uk/~rleigh/$ice"
```
joshmoore: I'd prefer not to hard-code links to `users.o.org.uk`.

joshmoore:
Thanks for doing this, @rleigh-dundee. Barring the one comment re: `users` vs. `downloads`, this makes good sense. Let's see what happens with the Ice 3.6 work and the current discussions re: the requirements page.

rleigh-dundee:
@joshmoore Updated to download from http://downloads.openmicroscopy.org/ice/experimental/ which is definitely a better place for this stuff.

rleigh-dundee:
@joshmoore WRT the requirements page, this is independent isn't it?  I mean, that's to do with python 2.7 and Ice 3.5; removing Ice 3.4 is independent of that.  Since support has already been removed from the CI side for Ice 3.4, the documentation change is simply reflecting that Ice 3.4 is no longer supported--we kept it solely for travis builds, and this PR is that the last piece of that being replaced.  Or is there some additional stuff to consider?

rleigh-dundee:
@joshmoore Note the adjusted URL is green in https://travis-ci.org/openmicroscopy/openmicroscopy/builds/84891624

joshmoore:
Closing for the moment; card opened https://trello.com/c/L8mmbgiS/56-ice-3-4-pr

PR #4256, opened 2015-10-08
Only analyze shapes of the current plane

dominikl:
Further improvement of the ROI analysis, see [Trello card](https://trello.com/c/3HccWkMv/40-roi-memory-issue)

With this PR only the shape of the current active plane is analyzed, not the full stack, which should speed up the analysis significantly for large ROIs which are populated through several planes.

Test: Create a large ROI (which covers a lot of pixels), populate it through the stack. Perform an analysis by switching to the Graph Pane. Move through the stack via the sliders. The newly selected plane should then get analyzed and the diagram updated automatically. Check that the results make sense. Ideally this should be tested with different kinds of images (single plane, multi-z, multi-t, multi-z/t).
Also have a look at the Intensity View.

--
Second attempt, with less code changes, hope this time I didn't break as much as before.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityView.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityView.java
@@ -277,7 +277,7 @@ class IntensityView
 	/** The slider has changed value and the mouse button released. */
 	private void handleSliderReleased()
 	{
-		if (zSlider == null || tSlider == null || coord == null|| 
+		if (zSlider == null || tSlider == null || coord == null || 
                 state != State.READY)
 			return;
 		
@@ -1231,6 +1231,7 @@ class IntensityView
                 state = State.ANALYSING;
                 controller.analyseFigures(statsMissingFigures);
             }
+            channelSummaryTable.setVisible(false);
             clearMaps();
             return;
         }
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityView.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityView.java
@@ -1355,10 +1355,7 @@ class IntensityView
 		Object src = evt.getSource();
 		if (src == zSlider || src == tSlider) {
 			formatPlane();
-			OneKnobSlider slider = (OneKnobSlider) src;
-			if (!slider.isDragging()) {
-				handleSliderReleased();
-			}
+			handleSliderReleased();
 		}
 	}
 	
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityView.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityView.java
@@ -1227,10 +1227,10 @@ class IntensityView
 		}
 		
 		if (!hasData) {
-            if (!statsMissingFigures.isEmpty()) {
-                state = State.ANALYSING;
+            if (!statsMissingFigures.isEmpty()) 
                 controller.analyseFigures(statsMissingFigures);
-            }
+            else
+                state = State.READY;
             channelSummaryTable.setVisible(false);
             clearMaps();
             return;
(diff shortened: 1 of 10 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityView.java
```diff
@@ -273,36 +277,40 @@ private void formatPlane()
...
-		{
-			state=State.READY;
-			return;
-		}
-		selectedChannelName = string;
-		int channel = nameMap.get(string);
-		if (channel == -1)
-			return;
-		if (!pixelStats.get(thisCoord).containsKey(channel))
+		if (zSlider == null || tSlider == null || coord == null||
```
mtbc: lost a space before last `||`

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.0-m1-148-1585541-ice35-b111-mac eel user-3 read-only-1 on Mac 10.11

ROI drawn on single z plane, single time point
Graph pane shows as expected when ROI selected and viewing correct z and t.
Graph pane clears when moving to z or t that ROI not on.
Intensity pane shows as expected when ROI selected and viewing correct z and t.
Intensity pane does not subsequently update when z and t changed to ones without ROI on it.
Same display persists in Intensity pane even when ROI selected - i.e. not being updated at all.
Same lack of in Intensity pane update when ROI propagated.

I then saved ROIs and closed the viewer.
On opening the viewer again, and repeating the tests, both the Graph pane and the Intensity pane updated and behaved as expected - i.e. Intensity pane updated correctly.

This workflow remains fairly unintuitive with the updating of panes when ROI does not appear to be selected etc.
However given how fragile this code appears to be, I suggest any further changes are considered carefully before any decisions made.

I would therefore suggest an attempt to fix the Intensity pane update on unsaved ROIs, and when that is sorted and retested, I would like to review the whole workflow and interaction with Petr and perhaps Michael and see if anything else should be tackled (more than likely another PR/ticket).

dominikl:
Thanks @gusferguson , yes, I'll fix the IntensityView problem within this PR. I just put it on hold for a while, to focus on the right-hand side metadata panel changes for the moment.

dominikl:
With the last commit the intensity view table is removed when you move to a plane which doesn't have a measurement result (so that the behaviour matches the graph pane behaviour). Couldn't see a difference between rois being saved and not saved.
With the 5.2 code freeze coming closer I'm afraid I can't do much more on this PR. Would be great to check with Michael and @pwalczysko on this soon.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.0-m1-270-5057004-ice35-b115-win eel logged in as user-6 on own read-write-1

Intensity and graph panes both updating as expected with moving through Z and T on propagated ROIs
Both clear if moving onto Z or T without selected ROI on it.
Both re-display as expected if move back onto Z or T with selected ROI.

2 relatively minor issues remain:
in both Z and T:
select ROI - select Intensity pane - using Z or T bar in pane move to end of propagation stack/point
in main viewer move off Z or T to one without ROI - Intensity pane goes blank as expected
in Intensity pane - move back onto Z or T with slider on pane - display does not refresh
If you do this with the Graph pane, moving back using slider on pane does refresh display
This is a very minor inconsistency but would be good to fix if easy.
Probably related is the second one - if you do this the Z or T slider in the main viewer does not update/Z or T change in response to movements in the Intensity pane slider whereas it does to movements of the Graph pane slider.
Note: - when using under other/normal circumstances slider movement in the intensity pane does update the slider and change the Z or T in the main viewer as expected.

This is a minor edge case that is not critical really but would be good to tidy up.
However if time is pressing then I would be happy if you wanted to merge this PR and do a "minor bug" fix at a later date.

jburel:
@dominikl: is the last commit supposed to fix @gusferguson's comment?
If yes this is not the case.
Also the ui is not update correctly when switching between shapes.see Image ID: 9567 user-6 read-write group.

jburel:
Discussed with @dominikl this am, the case I reported has always been there.
This will be addressed during 5.2.1
Ready to merge

PR #4258, opened 2015-10-13
trac 13018: disabling inactive share access

joshmoore:
When using `joinSession` on an inactive share, throw PermissionDenied:

```
        CannotCreateSessionException Raised if the session
E       PermissionDeniedException: exception ::Glacier2::PermissionDeniedException
E       {
E           reason = Share:1489 (497c947c-8e57-4e9a-9886-f3f87e6016a6) is inactive
E       }

```

See: https://trac.openmicroscopy.org/ome/ticket/13018#comment:4

### DIFF
--- a/components/server/src/ome/services/sessions/SessionManagerImpl.java
+++ b/components/server/src/ome/services/sessions/SessionManagerImpl.java
@@ -544,8 +544,10 @@ public class SessionManagerImpl implements SessionManager, SessionCache.StaleCac
 
             if (Boolean.FALSE.equals(active)) {
                throw new SecurityViolation(prefix + " is inactive");
-            } else if (started.getTime() + timeToLive > System.currentTimeMillis()) {
-                throw new SecurityViolation(prefix + " has expired");
+            } else if ((System.currentTimeMillis() - started.getTime()) > timeToLive) {
+                String msg = String.format("%s has expired: %s, timeToLive=%s",
+                        prefix, started, timeToLive);
+                throw new SecurityViolation(msg);
             }
         }
     }
--- a/components/server/src/ome/services/sessions/SessionManagerImpl.java
+++ b/components/server/src/ome/services/sessions/SessionManagerImpl.java
@@ -519,9 +519,36 @@ public class SessionManagerImpl implements SessionManager, SessionCache.StaleCac
 
     public Session find(String uuid) {
         SessionContext sessionContext = cache.getSessionContext(uuid);
+        checkIfShare(sessionContext);
         return (sessionContext == null) ? null : sessionContext.getSession();
     }
 
+    private void checkIfShare(SessionContext sessionContext) {
+        if (sessionContext.getSession() instanceof Share) {
+            final Long id = sessionContext.getSession().getId();
+            final String uuid = sessionContext.getSession().getUuid();
+            final String prefix = String.format("Share:%s (%s)", id, uuid);
+
+            List<Object[]> rv = executeProjection(
+                    "select s.active, s.timeToLive, s.started from Share s where s.id = :id",
+                    new Parameters().addId(sessionContext.getSession().getId()));
+
+            if (rv.size() != 1) {
+                throw new RuntimeException(prefix + " could not be found!");
+            }
+
+            Object[] items = rv.get(0);
+            Boolean active = (Boolean) items[0];
+            Long timeToLive = (Long) items[1];
+            Timestamp started = (Timestamp) items[2];
+
+            if (Boolean.FALSE.equals(active)) {
+               throw new SecurityViolation(prefix + " is inactive");
+            } else if (started.getTime() + timeToLive > System.currentTimeMillis()) {
+                throw new SecurityViolation(prefix + " has expired");
+            }
+        }
+    }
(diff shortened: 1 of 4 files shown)


joshmoore:
@will-moore / @aleksandra-tarkowska : I'd appreciate any additions to this test to reproduce http://trac.openmicroscopy.org/ome/ticket/13018 -- so far, I always get an exception when attempting to access such a share.

will-moore:
@aleksandra-tarkowska Do you want to have a look at this, or should I try to replicate what we were seeing in web?

aleksandra-tarkowska:
@will-moore I will write the full test, no worry

aleksandra-tarkowska:
@joshmoore done https://github.com/joshmoore/openmicroscopy/pull/51

sbesson:
@joshmoore @aleksandra-tarkowska: is the `test_ishare` failure in https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-Python27/ and  https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-python related to this PR?

will-moore:
The test tries to do  ``` self.new_client(session=s.uuid) ``` a couple of times and both times this raises a ```PermissionDeniedException``` but it would be nice to know that this does actually work (before the share is disabled and a second time before the share is expired).

components/tools/OmeroPy/test/integration/test_ishare.py
```diff
@@ -983,6 +985,50 @@ def test8704(self):
...
+        # test inactive share, if member has no access to the image
+        s = o_share.getShare(sid)
+        with pytest.raises(Glacier2.PermissionDeniedException):
+            self.new_client(session=s.uuid)
+
+        # activate again
+        o_share.setActive(sid, True)
+
+        # test expired share, if member has no access to the image
+        expiration = long(time.time() * 1000) + 86400
```
will-moore: Is there any relevance to the 86400? It looks like this is setting an expiration date in the future? But that wouldn't then be expired, so I'm confused.

components/tools/OmeroPy/test/integration/test_ishare.py
```diff
@@ -983,6 +985,50 @@ def test8704(self):
...
+        o_share.setExpiration(sid, rtime(expiration))
+        self.assert_expiration(expiration, o_share.getShare(sid))
+
+        with pytest.raises(Glacier2.PermissionDeniedException):
+            self.new_client(session=s.uuid)
+        # Forced closing
+        o_session = owner.sf.getSessionService()
+        o_session.closeSession(o_share.getShare(sid))
```
aleksandra-tarkowska: @will-moore share is closed here.

aleksandra-tarkowska:
@will-moore there is a need to create new client object and join session, although looks like test was altered loosing its main logic.

components/tools/OmeroPy/test/integration/test_ishare.py
joshmoore: `expiration` is when the share *will* expire rather than it *has* expired. Server-side this is converted into the appropriate setting for `timeToLive`.

components/server/src/ome/services/sessions/SessionManagerImpl.java
```diff
@@ -519,9 +519,36 @@ protected SessionContext createSessionContext(List<?> list, SessionContext previ
...
+            }
+
+            Object[] items = rv.get(0);
+            Boolean active = (Boolean) items[0];
+            Long timeToLive = (Long) items[1];
+            Timestamp started = (Timestamp) items[2];
+
+            if (Boolean.FALSE.equals(active)) {
+               throw new SecurityViolation(prefix + " is inactive");
+            } else if (started.getTime() + timeToLive > System.currentTimeMillis()) {
```
aleksandra-tarkowska: Share may have no expiry date, see failing test_8118

joshmoore: There's never no expiry but it is set to `Long.MAX_VALUE` which is causing an overflow. I'll push a fix for that in just a second.

joshmoore:
> The test tries to do self.new_client(session=s.uuid) a couple of times and both times this raises a PermissionDeniedException but it would be nice to know that this does actually work (before the share is disabled and a second time before the share is expired).

@will-moore, see e42a8ceae5b281cff5e17f6b507f1200578c3e2c

components/tools/OmeroPy/test/integration/test_ishare.py
aleksandra-tarkowska: Shouldn't that be other way around? First close then check if new client cannot join?

aleksandra-tarkowska: Ok my comment was before sleep. Fair enough.

joshmoore:
Job is greeen: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/20/

will-moore:
OK - apart from minor potential for improving readability of the tests, this is good to merge.

jburel:
Thanks all merging.

PR #4260, opened 2015-10-13
Add support for twitter cards (full image viewer only)

manics:

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -481,6 +481,18 @@ CUSTOM_SETTINGS_MAPPINGS = {
     "omero.web.public.cache.timeout":
         ["PUBLIC_CACHE_TIMEOUT", 60 * 60 * 24, int, None],
 
+    # Social media integration
+    "omero.web.twitter.enabled":
+        ["TWITTER_ENABLED",
+         "false",
+         parse_boolean,
+         "Enable Twitter cards."],
+    "omero.web.twitter.siteuser":
+        ["TWITTER_SITE_USER",
+         None,
+         leave_none_unset,
+         "Twitter site username."],
+
     # Application configuration
     "omero.web.server_list":
         ["SERVER_LIST",
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -489,6 +489,13 @@ CUSTOM_SETTINGS_MAPPINGS = {
          ("Dictionary of `server-name: @twitter-site-username`, where "
           "server-name matches a name from `omero.web.server_list`. "
           "For example: ``'{\"omero\": \"@openmicroscopy\"}'``")],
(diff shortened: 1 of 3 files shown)


joshmoore:
Does configuration need to be per `server_list` item?

manics:
Good point. I guess one option is
```
omero.web.twitter = {
    "<key>": {
        "enabled": true,
        "siteuser": "@openmicroscopy"
    }
}
```
with a key from `omero.web.application_server.host`- assuming its possible to obtain which host was selected.

manics:
The `enabled` key seemed redundant, so now it's just a dictionary of server-names to twitter-names, e.g. `omero.web.sharing.twitter={"trout": "@xxx"}`

manics:
Ideally this would be testable without deploying it live to the IDR...

manics:
Also added opengraph (facebook) support

joshmoore:
Interesting (unrelated) travis failure:
```
package:
      [jar] error while reading original manifest in file: /home/travis/build/openmicroscopy/openmicroscopy/components/tools/OmeroJava/target/omero_client.jar due to error in opening zip file
BUILD FAILED
/home/travis/build/openmicroscopy/openmicroscopy/build.xml:179: The following error occurred while executing this line:
/home/travis/build/openmicroscopy/openmicroscopy/components/antlib/resources/global.xml:449: The following error occurred while executing this line:
/home/travis/build/openmicroscopy/openmicroscopy/components/tools/OmeroJava/build.xml:90: Problem reading /home/travis/build/openmicroscopy/openmicroscopy/components/tools/OmeroJava/target/omero_client.jar
```
Relaunched.

manics:
Example config

```
config set omero.web.server_list '[["demo.example.org",4064,"demotwitter"]]'
config set omero.web.sharing.twitter '{"demotwitter": "@openmicroscopy"}'
config set omero.web.sharing.opengraph '{"demotwitter": "@ Open Microscopy"}'
```

manics:
In theory the last line should set the name of the site shown in facebook, though it doesn't seem to.

joshmoore:
This is currently leading to:

```
  File "manage.py", line 38, in <module>
    import settings
  File "/home/omero/OMERO.server-5.1.4-232-1ff6fcd-ice35-b170/lib/python/omeroweb/settings.py", line 781, in <module>
    'DEPRECATED_SETTINGS_MAPPINGS')
  File "/home/omero/OMERO.server-5.1.4-232-1ff6fcd-ice35-b170/lib/python/omeroweb/settings.py", line 772, in process_custom_settings
    setattr(module, global_name, mapping(global_value))
  File "/opt/anaconda/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/opt/anaconda/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
```

manics:
@joshmoore I think the last build you deployed was missing the last commit 7845fd6 which should've fixed that error.

manics:
--rebased-to #4613

PR #4261, opened 2015-10-13
ROI service

jburel:
Deprecate methods in ROIService
I had to mark each method as deprecated in the slice so it is reflected in Java
There is maybe a more suitable way
cc @joshmoore


joshmoore:
Makes sense. If we do decide to salvage anything from `IRoi` we can just put it under `IRoi2` or whatever it will be called. This will give everyone maximal warning.

PR #4263, opened 2015-10-14
OMEROMetadataStoreClient: speed up import for many LSIDs

joshmoore:
The check of List.contains(LSID) was extremely time expensive.
This code could use a further refactoring but that will mean
changing public methods. Instead, here we simply maintain a
map of checks which can be used to speed up the parsing.

cc: @chris-allan @jburel @mtbc

### DIFF
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -249,6 +249,9 @@ public class OMEROMetadataStoreClient
     private Map<LSID, List<LSID>> referenceCache =
         new HashMap<LSID, List<LSID>>();
 
+    private Map<LSID, Set<LSID>> referenceCacheCheck =
+        new HashMap<LSID, Set<LSID>>();
+
     /** Our authoritative LSID container cache. */
     private Map<Class<? extends IObject>, Map<String, IObjectContainer>>
         authoritativeContainerCache =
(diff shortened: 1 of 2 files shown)


jburel:
It will have been nice to use some features of guava but with ```getReferenceCache``` being public we are limited.
I had limited access to squig today (too slow), so will import decent size images tomorrow and compare numbers

joshmoore:
A very large number of PlaneInfos might suffice to show a difference (i.e. `image&exposureTime=1&sizeT=100000.fake`). In my case, it was a mitocheck dataset with 10K+ ROIs *per Image* of a 384-well plate.

mtbc:
The code looks like a good improvement: don't want to majorly refactor right before release! Assuming @jburel has functional testing in hand.

jburel:
running few more tests today

jburel:
Unfortunately, I have a regression with the tests run
35 w/o this PR and I am 43 with it.

joshmoore:
Can I get details of one or more of the regressions?

jburel:
run some tests again today using the fake file
against eel/merge and eel/latest. No regression but no significant improvement with this PR
The "metadata importing phase" in both cases takes around 30mins.
I might have to try the mitocheck data to see

joshmoore:
I'm fairly convinced that the wait I was seeing was based on ROIs coming from CellH5 but don't know why that would be worse than with fake.

joshmoore:
Discussion re: fake -- it might be that a `Ref` is needed in which perhaps annotation is better than planeinfo.

jburel:
usage of annotation in fake did not make a difference. Trying to find something else?

jburel:
I have created an ome-tiff with 40 rois (one roi/one shape)
The post upload phase is divided by 2 using this PR.
Will generate another ome-tiff with more rois but it is looking good.

jburel:
I ran few more tests with OME-TIFF with ROI.
Improvement in speed when dealing with ROI.

PR #4264, opened 2015-10-14
Jstree followup

will-moore:
This PR adds version number to jstree lib directory and allows user to Drag and Drop Dataset & Screen onto experimenter to make then orphaned.

To test:
 - Drag a Dataset from under a Project onto Experimenter.
 - Should become orphaned
 - Same for a Plate under a Screen
 - Cannot drag orphaned Dataset or Screen and drop onto Experimenter
 - Quick check that jstree still works on Tags and Public pages (no missing library, css, images etc from re-naming).


pwalczysko:
All good here. Ready to merge.

jburel:
One thing to do in another PR will be to externalise the version number
so we only have one place to edit
similar to what is done on the java dependencies front.

jburel:
I have noticed a problem when D&D
When I select a plate for example in a screen and drop it on the same screen
it becomes orphaned. Not expected

will-moore:
@jburel As discussed - it seems that jsTree is actually allowing you to drop the Plate either side of the Screen to become a sibling and doesn't even allow you to drag into the Screen itself (and show X). Either we'll have to accept this or disable the dragging into experimenter (unless I can find a nicer way of customising jsTree without hacking the source).

will-moore:
@jburel Turns out that jsTree is quite configurable. That commit should improve things.

jburel:
@will-moore: conflict

pwalczysko:
Tried Firefox and Chrome on Mac and Firefox on Windows. Works as expected (including the last commit workflow). The additional small pointer which jstree uses to indicate where the object will be dropped is very helpful, I do not think there can be any confusion now.

Ready to merge.

jburel:
One small problem: probably to be fixed in a follow-up PR
If I drag a plate contained in a screen over a dataset, the green tick is visible indicating that I can drop it.
but in reality the plate becomes orphaned
Note that I do not have a green tick when I drag an orphaned plate over a dataset

PR #4177, opened 2015-09-17
Javadoc final warnings

sbesson:
See https://trello.com/c/RpJxX7S7/35-javadoc-cleanup

This removes the remaining warnings thrown by `./build.py release-javadoc`. Spring framework jars are removed from the classpath - we might consider revisiting this when upgrading them. The custom `DEV.TODO` tag is removed from the Javadoc processing and broken Hibernate/Lucene links are removed. It seems the API for the versions included in OMERO is not hosted anymore by the official sites.

--no-rebase

PR #4269, opened 2015-10-15
Repository: spring cleanup

sbesson:
See https://trello.com/c/2qw5o7PL/27-clean-lib-repository

This PR removes the Spring JARs from the local repository. To test it, check the build passes and review the dependencies under `share/{server,client}` in the server.


mtbc:
Do we still need the `repository/org.springframework.*.jar` exclusion for release-javadoc?

sbesson:
I would think so as this was introduced in https://github.com/openmicroscopy/openmicroscopy/pull/4177 to remove Javadoc warnings. This PR does not bump any spring framework version so I expect these warnings to be still present. That being said, one could try to remove the exclusion line and check the output of the `release-javadoc` target.

mtbc:
That makes sense, thank you.

mtbc:
The changes seem to make sense and a completely clean build -- no local caches anywhere -- even gets build-dev building fine. CI looks fine too, nice cleanup. Good to merge.

PR #4273, opened 2015-10-16
Large ann index sql

joshmoore:
See: https://trello.com/c/EwmJfe5u/89-critical-indexing-large-annotation-changes

**To test**, generate large numbers (10-100K) of annotations, delete them, re-add them. If you check the indexer thread without this PR, it will be hung on startup waiting on a JDBC call to complete. If you check `pg_stat_activity`, you will see the call to `annotations_update_note_reindex` being processed. It will hang for hours if not days. It will be necessary to actively kill the process.

*With this PR*, the handling will be done in batches of 100,000 which should take around a minute. Then the Indexer call will run normally. Each subsequent Indexer run will handle another batch.

cc: @mtbc 

To be decided:
 * [ ] should the number 100K be configurable?
 * [x] should this be turned into `OMERO5.2__0` (no dev)


sbesson:
See https://ci.openmicroscopy.org/view/Bio-Formats/job/OMERO-DEV-breaking-integration/555/console

```
21:14:53 psql:dbsetup.sql:7697: ERROR:  syntax error at or near "{"
21:14:53 LINE 5:         row _updated_annotations{'SCRIPT': '<filename here>'...
21:14:53                   
````

mtbc:
I assume we need to add rather more than 10-100K annotations if the batch size is 100k? And we don't delete the annotations, just the links?

mtbc:
Even just the initial add step seems to be taking absolutely ages for my server, my eight cores have been been busy for ages now. Or, if I follow the description without any linking at all, then everything works fine without this PR. Not quite sure what I'm doing here, probably best to pass off or get clearer about what to do. The diff looks plausible, anyway.

joshmoore:
I'm not exactly sure where things were hanging for you; perhaps an alternative would be to just insert directly into the `_updated_annotations` table to simulate the addition/deletion?

mtbc:
Still, *should* these be linked annotations, and *should* there actually be rather more than the batch size? I'll experiment with more approaches to try to reproduce behavior approximating the PR description.

mtbc:
If I insert 1,000,000 rows into `_updated_annotations` in one transaction without this PR, the table's cleared in seconds and the server again settles.

mtbc:
The hang was, with this PR if I insert a bunch of linked annotations, then things don't settle afterward; I wasn't sure what was going on, but PostgreSQL looked to be the culprit, maybe post-transaction triggers?

joshmoore:
> Still, should these be linked annotations, and should there actually be rather more than the batch size? I'll experiment with more approaches to try to reproduce behavior approximating the PR description.

They should definitely be linked, likely to more than one image (in my case, it was one annotation per image).

> If I insert 1,000,000 rows into _updated_annotations in one transaction without this PR, the table's cleared in seconds and the server again settles.

With this PR? And without this PR it hangs forever, no?

> ... PostgreSQL looked to be the culprit, maybe post-transaction triggers?

Hmmm.... I don't know how long it took for all my triggers to *fill* `_updated_annotations`. Perhaps overnight.

mtbc:
No, without: table was fine. But that test may have had all the annotations on the same image.

joshmoore:
And which PG version? This was: 9.3.9

mtbc:
9.4.4

joshmoore:
That may be the issue. You could try a simple `insert _updated_annotations
&& annotation_updates_note_reindex()` to see what the performance looks
like. Perhaps the indexes are being applied on 9.4?

mtbc:
Finally managed to test this properly by exploiting a bug in auto-close import! Diff looks good and all works exactly as described. Good to bring out of breaking, then merge if CI stays green.

PR #4274, opened 2015-10-16
Tag multiple selection

aleksandra-tarkowska:
This fix multiple selection in a tag tree reported in http://trac.openmicroscopy.org/ome/ticket/13065

To test, go to tag tab and try to select multiple tags. You should be only allowed to select one.

cc: @jburel


jburel:
No longer possible to select more than one element in tags view.
Problem reported is fixed.
For later: need to review the display of tags to match the "Explore" tab.

PR #4276, opened 2015-10-16
Bump references to 5.2

sbesson:
This PR updates the documentation and latest redirects references to use 5.2 instead of 5.1. The commits were produced  by running:

```
$ python components/tools/bump_version.py 5.2
```

Note that the latest 5.2 OMERO redirects do not exist yet and the documentation has not been re-promoted. The first release candidate might be a good occasion to have this into place if we want this PR to be properly tested /cc @hflynn @jburel

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Also see [CONTRIBUTING.md](./CONTRIBUTING.md)
 
 Full developer documentation for OMERO is available from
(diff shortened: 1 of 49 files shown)


rleigh-dundee:
The changes certainly look fine to me.

joshmoore:
NB: eventually part of snoopy's EOY work?

PR #4279, opened 2015-10-18
Rename agent of SessionsBean.createSessionWithTimeouts()

sbesson:
See https://trello.com/c/azP0udku/37-rename-createsession-agent

Since this method uses a Principal object and is principally consumed as a way to sudo, the agent is renamed as OMERO.sudo.

To test this PR:
- create a connection from the CLI using `bin/omero login --sudo sudoer user@hostname:port`
- import an image for another user in Insight

In both cases, list the server-side sessions using `bin/omero sessions who` and check both sessions listed above are created with the `OMERO.sudo` agent.


mtbc:
The created sessions use the OMERO.sudo agent and both clients allow import-as. Good to merge.

PR #4242, opened 2015-10-04
Center panel sync

will-moore:
This PR uses the jsTree for images in the centre panel (everything apart from SPW data).
This means that loading Thumbnails should be MUCH faster!

NB: Not yet ready for full testing but generally working OK.
 - Some things not yet done:
     - Remove old unused code.

Need to test all functionality that affects the centre panel.

- Browsing P/D/I S/P/W in tree - Multi-select of P/D/I, browse between P/D/I and SPW etc.
- Double - click thumbnail to open image
- Cut, copy paste in tree
- Selection of images in centre panel (click, shift/cmd click & drag-select)
- Filtering of images in centre panel (along with selection - E.g. select all then filter - filtered images de-selected)
- Zooming of thumbnails - should be preserved between Datasets.
- Switch between icon / table layouts - should be preserved between Datasets.
- Pagination
- Browsing of Tags / Tagsets (containing P/DI, S/PW).
- Public browsing of shares - Opening shared images with share ID.
- Refresh tree will update thumbnails if available (E.g. new rdef in Insight or Web client in another browser).
- Web integration tests should be green.

PR #4281, opened 2015-10-19
No thumbs jstree

will-moore:
This fixes a regression introduced in #4242 where images with no thumbnails are not loaded in jsTree.

To test, find images that have no thumbnail and check that they are shown in jsTree and in center panel (with place-holder "No Image" thumbnail).
Also re-test the loading of new thumbnails on jsTree refresh (not browser refresh) when new thumbnails have been created in another client (another web browser or Insight).

This PR is actually just 2 commits on top of Center panel sync: #4242, so that should be merged before this is.


pwalczysko:
@will-moore : Is image in Share which has been deleted ``an image with no thumbnail`` ? What are the other ways to get such images ?

will-moore:
@pwalczysko I'm not sure of the best way to create an image with no thumbnail. I have a couple locally from failed imports.
You can actually delete a thumbnail via CLI,
E.g. delete thumbnail for Image ID 25152:
```
$ bin/omero delete Image/Thumbnail:25152 --report
```
If you do this then you'll see that the image still appears in jsTree - but you won't see the "broken thumbnail" placeholder, since when the browser tries to get a thumbnail a new one is created!

will-moore:
https://eel.openmicroscopy.org/merge/webclient/?show=image-25152 has no thumb and file deleted so it can't make more! (user-3)

pwalczysko:

will-moore:
@sbesson I was hoping that this PR just showed the last 2 commits after #4242 was merged. But maybe I need to merge develop into this branch? Or rebase?

sbesson:
@will-moore so did I but something clearly happened on the GH interface. Discussing with @jburel, let's get this merged as this has been cleared and we can review less PRs tomorrow.

PR #4231, opened 2015-09-30
Another formatting issue

dominikl:
Follow up PR for #4199 

Changes the term "t index" to simply "t", and start count for t from 1 (instead of 0), see 'Exposure Time' table in channel acquisition metadata.

screen shot 2015-10-01 at 08 34 02

(only the last commit is relevant for this PR, the other two are part for the previously mentioned PR)

PR #4282, opened 2015-10-19
t index formatting

will-moore:
This is the equivalent on #4231 for Insight.
Removes the label "t index" to "t" and uses 1-based time indices.

To test, check that time stamps look same in web and Insight.

screen shot 2015-10-19 at 12 40 26


pwalczysko:
Cannot see any difference between Web and Insight on this one. Ready to merge.

PR #4284, opened 2015-10-19
Rnd copy/paste

jburel:
Fix another out of synch issue
to test: 
Select image in centre pane.
Copy settings.
Open another image in full viewer.
Go back to main window - leaving Full Viewer open.
Select open image and click on Preview tab.
Paste settings and save using Preview pane buttons.
Preview pane updates as expected, Full Viewer updates as expected and Save button on both Preview and Full Viewer are disabled as expected.
Repeat but paste settings using right-click Paste and Save in centre pane or data tree.
Preview pane updates as expected and Save button in Preview is disabled as expected.
In Full Viewer image updates as expected, The  Save button should now become disabled (was enabled w/o this PR).

cc @dominikl  @gusferguson

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/ImViewerAgent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/ImViewerAgent.java
@@ -335,9 +335,10 @@ public class ImViewerAgent
             if (view != null ) {
                 if (view.getPixelsID() != id) {
                     view.pasteRenderingSettings();
+                    view.saveRndSettings(true);
                 }
                 view.reloadRenderingThumbs();
-            }     
+            }
         }
     }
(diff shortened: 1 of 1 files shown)


dominikl:
Works like expected, good to merge. All synchronization issues seem to be fixed now, thank you very much @jburel ! I got really stuck with this.

pwalczysko:
Works for me as well except the second workflow (pasting into central pane and tree when full viewer is open). For some reason this is not working as long as the full viewer is open. The pasting pastes the rnd settings to the full viewer properly, but to the thumb in central pane and preview -> there still remain the old setting. Probably something for 5.1.2, but I would emphasise that this is an edge case (user has to first open the Full viewer and then not actually use it, instead pasting the rnd settings into the thumb of the same image).

PR #4285, opened 2015-10-19
Annotation flag

jburel:
Remove the check if the image has a companion file
to test;
 * Import an image
 * Tag the image.
 * Check that the "yellow" flag is displayed in the tree.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/util/EditorUtil.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/util/EditorUtil.java
@@ -963,7 +963,6 @@ public class EditorUtil
             if (counts == null || counts.size() <= 0) {
                 return count > 0;
             }
-            int n = 1;
             Iterator<Entry<Long, Long>> i = counts.entrySet().iterator();
             long value = 0;
             Entry<Long, Long> entry;
@@ -972,7 +971,7 @@ public class EditorUtil
                 value += (Long) entry.getValue();
             }
             value += count;
-            return value > n;
+            return value > 0;
         } else if (object instanceof ScreenData)
             counts = ((ScreenData) object).getAnnotationsCounts();
         else if (object instanceof PlateData)
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/util/EditorUtil.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/util/EditorUtil.java
@@ -974,13 +974,6 @@ public class EditorUtil
                 return count > 0;
             }
             int n = 1;
-            try {
-                String format = image.getFormat();
-                if (format != null &&
-                        FORMATS_WITH_COMPANION.contains(format.toLowerCase()))
-                    n = 2;
-            } catch (Exception e) {
-            }
             Iterator<Entry<Long, Long>> i = counts.entrySet().iterator();
             long value = 0;
             Entry<Long, Long> entry;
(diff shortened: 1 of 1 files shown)


dominikl:
The annotation flag is displayed after the image has been tagged; but it's not displayed anymore after a restart/re-login. I'm not sure if that's supposed to be fixed by this PR as well?

jburel:
@dominikl I have not changed that. Need to double check.Which image?

dominikl:
That's the case with any image (actually dataobject in general), the annotation flag is displayed directly after adding an annotation but not for objects which have been annotated in the past (well, for some it is for some it's not, seems to be a bit random). There was a trac ticket about it, can't find it at the moment. Maybe @pwalczysko still knows what ticket it was (was reported by Michael as far as I can remember).
Well, anyway, if that's not what the PR is supposed to fix, then it's fine (just wasn't sure about that).

dominikl:
It probably was just reported on the "follow up" trello card, but this actually links to this PR.

pwalczysko:
@dominikl : I just cannot confirm the behaviour of any vanishing yellow flags now as I did couple of tries on logging out after tagging and logging back in. On the other hand, I have noticed the opposite problem, this means the yellow dots seem to be inlationary, not clear on many images why they are there (see below just a few examples)

  - 1.  Observe images IDs 170, 30, 3888 and 26 user-2 on eel merge. These images have yellow dots in a tree and in central pane. I cannot see any single annotation on them though.

pwalczysko:
Other than the https://github.com/openmicroscopy/openmicroscopy/pull/4285#issuecomment-149945613 problem, this works as expected.

RFE: Also consider presence/absence of ROIs as Annotation/no-Annotation (atm the ROIs are completely ignored, no yellow flag is shown - I think many users would like to be able to easily identify which images have ROIs and which not)

jburel:
@pwalczysko: For ROI, this will require some adjustments i.e. an roi count call to be added
since we currently have only a count for annotation linked to the object.

jburel:
@pwalczysko: the images listed will have a companion file which counts as annotation
That has always been the issue.

pwalczysko:
@jburel : I cannot see the companion files on the images in question (under Acquisition tab) - are they somewhere else where I cannot see them ?

Also, there is at least one discrepancy: The same image (ID 21044) and ID 180 once has a yellow dot, once it has not - is it because of the way / OMERO version which was used for import ?

mtbc:
I have a hazy memory of the companion files view being removed from the right pane in at least one of the clients, though admittedly I've now no idea which PR or when.

dominikl:
@pwalczysko I think the vanishing annotation flag was fixed by @jburel 's last commit f8a8b08 . I initially thought this PR was just about that, probably misunderstood the trello card, sorry for adding some confusion here.

pwalczysko:
@mtbc : No, it is not as simple. I have in the same Insight version a companion file displayed in the Acquisition tab. In other cases I have it NOT displayed. Both cases have yellow dot. 

Bottom line is that I am confronted with a flurry of yellow dots -> unfortunately thanks to this to identify the more meaningful annotations like Tags and File annot like this is impossible

pwalczysko:
Double-checked today once more: If the image does not have a companion file, the behaviour is as expected. When annot. is added, yellow dot appears, when annot. removed, yellow dot goes away. The yellow dot persists after logout and login. 

My prefered behaviour would be to ignore the companion files whatsoever for the yellow-dot workflow. But after discussion with @jburel  I understand that this is technically difficult. In that case this is ready to merge.

jburel:
Image with ID 180: as a companion file. That's why it is shown as "annotated"

jburel:
@pwalczysko: happy with the status of this PR?

pwalczysko:
Yes, ready to merge.

PR #4286, opened 2015-10-19
OMERO.matlab: always cast the group to a java.lang.String

sbesson:
This PR fixes a bug encountered while working on MultiMOT with the new multi-group functionality included in OMERO.matlab 5.1.4.

When retrieving objects in a different group than the session group, the method call resulted in exceptions of type `java.lang.Character cannot be cast to java.lang.String`. This exception was not thrown when passing `-1` as the group.
This PR reviews all usages of the group context in OMERO.matlab and unconditionally casts the `groupId` as a `java.lang.String`. The training examples are updated to also read object in a specified group. /cc @emilroz

### DIFF
--- a/examples/Training/matlab/ReadData.m
+++ b/examples/Training/matlab/ReadData.m
@@ -31,7 +31,7 @@ try
     datasetId = p.datasetid;
     imageId = p.imageid;
     plateId = p.plateid;
-    group2 = p.group2
+    groupId = eventContext.groupId;
 
     print_object = @(x) fprintf(1, '  %s (id: %d, owner: %d, group: %d)\n',...
         char(x.getName().getValue()), x.getId().getValue(),...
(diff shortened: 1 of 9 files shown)


examples/Training/matlab/ReadData.m
```diff
@@ -245,6 +276,26 @@
...
+    for i = 1 : numel(screens),
+        print_object(screens(i));
+    end
+    fprintf(1, '\n');
+
+    % Retrieve all the screens owned by the session user in the
+    % context of the specified group
+    disp('Retrieving screens owned by the session user in a different group')
+    screens = getScreens(session, 'group', group2);
+    fprintf(1, '  Found %g plates\n', numel(screens));
```
mtbc: really a plate count?

sbesson:
While https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training didn't fail with this commit, I realized I passed the group name instead of the group identifiers. Will push another commit.

sbesson:
See https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-training/131/console

ximenesuk:
Changes look sensible and locally the `ReadData.m` script ran fine once I had set the `ice.config` correctly!

PR #4298, opened 2015-10-26
Rename 'Public' tab to 'Shares'

will-moore:
As suggested in testing, this renames the "Public" tab to "Shares", which makes more sense since the data under that tab is in no way "Public". This is also more consistent with E.g. "Tags" tab containing Tags and "Shares" tab containing Shares.

@gusferguson @aleksandra-tarkowska Does this make sense to you?
@gusferguson If we go ahead with this fix and it requires changes to a lot of screenshots, I'm happy to help update screenshots in Photoshop (should be quite quick to patch the label change).


mtbc:
It makes sense to me!

gusferguson:
@will-moore @aleksandra-tarkowska - makes sense o me - I am happy for you to change

Just redoing a whole lot of screenshots now, so if it was in eel merge for tomorrow I could get the nevessary photoshop patch and it would be very little problem.

pwalczysko:
Works as expected. As @gusferguson seems to be on track with the screenshots, I do not see any objections to merging this.

gusferguson:
I am good.

PR #4299, opened 2015-10-26
Bugfix: 'Open with' external application

dominikl:
See [Ticket 13055](http://trac.openmicroscopy.org/ome/ticket/13055)

There were two issues:
- Open with ImageJ/Fiji always failed on Windows, because of the backslash path separator.
- Open with ImageJ/Fiji also failed on OSX with images which had spaces in the image name

With this PR non-word characters in the name for the temporary file for the 'open with' action are always replaced by underscores (to reduce the risk of complications with any external application) and the windows backslash is replaced by a slash before passing the path on to the ImageJ/Fiji bioformats importer.

**Test:**
- 'Open with' ImageJ and Fiji on OSX and Windows with an image which previously triggered the issue (white space(s) in image name).
- 'Open with' another application, e. g. Paint etc.; just to be sure this has not been broken.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/ui/OpenObjectActivity.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/ui/OpenObjectActivity.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.env.ui.OpenObjectActivity 
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2010 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


pwalczysko:
On Windows:
  - could repeat the previous bug using the 5.2.0 release insight
  - could confirm that the bug is fixed with this PR
  - could confirm that also an image with spaces in its name will open okay in ImageJ
  - could confirm that this image opens also in Adobe Photoshop okay
  - in Paint opens also okay

All good on Windows 7

pwalczysko:
On Mac:

  - could not repeat the previous bug iwth image with spaces in its name
  - could confirm that with this PR, the opening in Fiji and ImageJ is okay
  - could not confirm any other application, as cannot find any suitable on Mac (Preview fails because of the permission issues, but so does it fail without this PR, Photoshop does not open at all with or without this PR, Omnigraffle cannot open this type of files)

Although there are no problems here detected on Mac, it would be good to have the original image which caused the space-in-the-name induced bug to confirm. @dominikl

dominikl:
Thanks @pwalczysko . I'll try to find an example image which previously triggered the 'Open with' bug on OSX as well. Apart from that, I'll have a look at the still outstanding "permissions" failure with respect to 'Open with' on OSX.

dominikl:
@pwalczysko I also can't reproduce the problem with the spaces in the image name anymore, must have been something else. So maybe the replacement of non-word characters isn't necessary. But as it is just for a temporary file, it doesn't harm and I'd keep it as precaution.

The problem with permissions is specific for OSX and the Preview app. With the last commit I added a workaround for this. 
**Test**: Open an image with "Open with" Preview.app

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.0-65-52cd2f8-ice35-b154-mac eel user-3 read-only-1
Images opening in Preview successfully. Spaces in name are not a problem.
Behaves as expected.
Good to merge.

jburel:
The methods ```replaceWindowsPathSeparator``` and ```replaceNonWordCharacters``` should probably be moved to ```UIUtilities```.

dominikl:
Moved the methods to the ```UIUtilities``` class. I guess no re-testing is needed for that.

mtbc:
@jburel: Okay with you, or retest?

PR #4306, opened 2015-10-28
Fix config creation

simleo:
This PR addresses two related issues in `ConfigXml.open_source` (from `omero/config.py`):

* The method assumes that the config file is read-only if it cannot be opened in `a+` mode, while in reality the `open` call can fail for several other reasons. In particular, if `etc/grid` exists but it's not writable, the exception is raised by the subsequent open-for-read attempt, and the final error message suggests that something is missing:
```
$ dist/bin/omero config set omero.data.dir /OMERO
[Errno 2] No such file or directory: '[...]/dist/etc/grid/config.xml'
```
This is addressed by the first commit. **To test:**
```
$ mkdir -p dist/etc/grid
$ rm -rf dist/etc/grid/*
$ chmod 555 dist/etc/grid
$ dist/bin/omero config set omero.data.dir /OMERO
[Errno 13] Permission denied: '[...]/dist/etc/grid/config.xml'
```

* Since both config file and lock file creation are checked for in the same try/except block, if only the latter fails (e.g., because it already exists but it's not writable), the following happens (with logging level set to DEBUG):
```
$ dist/bin/omero config set omero.data.dir /OMERO
[...]
DEBUG:ConfigXml:open('[...]/dist/etc/grid/config.xml', 'a+') failed
'ConfigXml' object has no attribute 'lock'
```
In the log, wrong info is given on which file could not be opened. Moreover, the error message reflects a subsequent unhandled exception. This is addressed in the second commit. **To test:**
```
$ mkdir -p dist/etc/grid
$ chmod 777 dist/etc/grid
$ rm -rf dist/etc/grid/*
$ touch dist/etc/grid/config.xml.lock
$ chmod 444 dist/etc/grid/config.xml.lock
$ dist/bin/omero config set omero.data.dir /OMERO
[Errno 13] Permission denied: '[...]/dist/etc/grid/config.xml.lock'
```

### DIFF
--- a/components/tools/OmeroPy/src/omero/config.py
+++ b/components/tools/OmeroPy/src/omero/config.py
@@ -140,7 +140,6 @@ class ConfigXml(object):
         if not self.read_only:
             try:
                 self.source = open(self.filename, "a+")  # Open file handle
-                self.lock = self._open_lock()  # Open file handle for lock
             except IOError:
                 self.logger.debug("open('%s', 'a+') failed" % self.filename)
                 if not os.path.isfile(self.filename):
@@ -153,6 +152,8 @@ class ConfigXml(object):
                 if val is not None:
                     raise Exception(
                         "Non-default OMERO_CONFIG on read-only: %s" % val)
+            else:
+                self.lock = self._open_lock()  # Open file handle for lock
 
         if self.source is None:
             self.lock = None
--- a/components/tools/OmeroPy/src/omero/config.py
+++ b/components/tools/OmeroPy/src/omero/config.py
@@ -139,13 +139,12 @@ class ConfigXml(object):
         self.source = None
         if not self.read_only:
             try:
-                # Try to open the file for modification
-                # If this fails, then the file is readonly
                 self.source = open(self.filename, "a+")  # Open file handle
                 self.lock = self._open_lock()  # Open file handle for lock
             except IOError:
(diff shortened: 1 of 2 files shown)


sbesson:
Tested both exceptions described in the PR with OMERO.server-5.2.0-52-02b0959-ice35-b150. Both outcome work as expected. Also tested the non-default profile configuration on read-only is unmodified.

```
sbesson@c001236:OMERO.server-5.2.0-52-02b0959-ice35-b150 $ bin/omero config set omero.data.dir test
sbesson@c001236:OMERO.server-5.2.0-52-02b0959-ice35-b150 $ chmod 444 etc/grid/config.xml 
sbesson@c001236:OMERO.server-5.2.0-52-02b0959-ice35-b150 $ bin/omero config get
omero.data.dir=test
sbesson@c001236:OMERO.server-5.2.0-52-02b0959-ice35-b150 $ bin/omero config set omero.a b
Cannot modify /Users/sbesson/Downloads/OMERO.server-5.2.0-52-02b0959-ice35-b150/etc/grid/config.xml
sbesson@c001236:OMERO.server-5.2.0-52-02b0959-ice35-b150 $ OMERO_CONFIG=test bin/omero config get
Non-default OMERO_CONFIG on read-only: test
```

Newly added unit tests also cover the changes and are passing. Merging.

PR #4304, opened 2015-10-27
Improve web deploy

aleksandra-tarkowska:
Test nginx (no webserver needed)
```
$ dist/bin/omero web restart
Stopping OMERO.web... [NOT STARTED]
Post-processed 'omeroweb.viewer.min.css' as 'omeroweb.viewer.min.css'
Post-processed 'omeroweb.viewer.min.js' as 'omeroweb.viewer.min.js'

0 static files copied to '/Users/ola/OMERO/openmicroscopy/dist/lib/python/omeroweb/static', 618 unmodified, 2 post-processed.
Clearing expired sessions. This may take some time... [OK]
Starting OMERO.web... [OK]
```

Open localhost:4080 (you won't see statics)

```
$ dist/bin/omero web status
OMERO.web status... [RUNNING] (PID 7272)

$ cat dist/var/django.pid 
7272

$ dist/bin/omero web start
Post-processed 'omeroweb.viewer.min.css' as 'omeroweb.viewer.min.css'
Post-processed 'omeroweb.viewer.min.js' as 'omeroweb.viewer.min.js'

0 static files copied to '/Users/ola/OMERO/openmicroscopy/dist/lib/python/omeroweb/static', 618 unmodified, 2 post-processed.
Clearing expired sessions. This may take some time... [OK]
Starting OMERO.web... [FAILED] OMERO.web already started. /Users/ola/OMERO/openmicroscopy/dist/var/django.pid exists (PID: 7272)! Use 'web stop or restart' first.

$ dist/bin/omero web status
OMERO.web status... [RUNNING] (PID 7272)

$ cat dist/var/django.pid 
7272

$ dist/bin/omero web stop
Stopping OMERO.web... [OK]
OMERO.web WSGI workers (PID 7272) killed.
Removed stale /Users/ola/OMERO/openmicroscopy/dist/var/django.pid

$ dist/bin/omero web status
OMERO.web status... [NOT STARTED]

$ dist/bin/omero web stop
Stopping OMERO.web... [NOT STARTED]
```

test bad behavior, wait few sec between steps to make s

```
$ dist/bin/omero web start
Post-processed 'omeroweb.viewer.min.css' as 'omeroweb.viewer.min.css'
Post-processed 'omeroweb.viewer.min.js' as 'omeroweb.viewer.min.js'

0 static files copied to '/Users/ola/OMERO/openmicroscopy/dist/lib/python/omeroweb/static', 618 unmodified, 2 post-processed.
[...]

PR #4307, opened 2015-10-28
Add nginx arguments to restart command

sbesson:
Fix a bug reported during 5.2.0-rc2 testing

Without this PR, `bin/omero web restart` should fail with an `argparse` related error. With it, its behavior should be restored and OMERO.web should be restarted.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/web.py
+++ b/components/tools/OmeroPy/src/omero/plugins/web.py
@@ -137,18 +137,19 @@ class WebControl(BaseControl):
                 "--no-wait", action="store_true",
                 help="Do not wait on expired sessions clean-up")
 
-        start.add_argument(
-            "--workers", type=int, default=5,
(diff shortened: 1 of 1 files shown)


manics:
`web restart` is ambiguous- if args were passed to `web start ...` then arguably `web restart` should automatically include them.

aleksandra-tarkowska:
I am afraid restart never works like that in the past

sbesson:
Note this PR is mostly aimed at fixing a 5.2.0 blocker namely that `bin/omero web restart` is currently broken by restoring the 5.1.4 statu quo and re-exposing the nginx arguments to `restart` for now. @aleksandra-tarkowska has an open PR for reviewing the Web deployment/redeployment logic (https://github.com/openmicroscopy/openmicroscopy/pull/4304) which will review post 5.2.0.

manics:
Good to merge

PR #4311, opened 2015-10-29
Fix update of thumbnails on rdef 'Save' and 'Save to All'

will-moore:
Fixes update of thumbnails on rendering settings 'Save' and 'Save to All'.

To test:
 - Save rendering settings on an image in Preview panel - thumbnail should update.
 - "Save All" should update all thumbnails.
 - Also check that right-click Rendering Settings options update thumbnails as expected.


jburel:
Work as expected.

PR #4313, opened 2015-10-29
Also move Windows.cfg file to the templates

sbesson:
The CLI admin logic for Windows uses the content of this file to check and amend the content of `windefault.xml`. Without this modification both files are out-of-sync since `windefault.xml` is regenerated at every rewrite but `Windows.cfg` has been patched by `win_set_path`, hence preventing Windows restart.
This commit should allow Windows servers to be restarted (including on different ports).


sbesson:
For 5.2.1, we'll want to add a follow-up card to migrate away from `win_set_path` and use the template replacement logic instead.

sbesson:
Tested via CI by running the deployment job twice in a row [build 1](https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-deploy-win2012/41/console) and [build 2](https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-deploy-win2012/42/console). The same server was used instead of downloading a new one and effectively running `bin/omero admin stop && bin/omero admin start`.

jburel:
tested also with ```bin/omero admin restart```. Merging

PR #4315, opened 2015-10-30
Have DropBox user log out after import

ximenesuk:
See https://trello.com/c/4dIYzO6H/13-dropbox-doesn-t-logout-user

To test use DropBox to import a file for a user, watch the `DropBox` log for the import completing and then try some (non-destructive!) cli action that requires log in. You should be prompted to log in rather than reconnecting to the session used by DropBox for the import.

### DIFF
--- a/components/tools/OmeroFS/src/fsDropBoxMonitorClient.py
+++ b/components/tools/OmeroFS/src/fsDropBoxMonitorClient.py
@@ -544,7 +544,7 @@ class MonitorClientI(monitors.MonitorClient):
 
     def loginUser(self, exName):
         """
-        Logins in the given user and returns the client
+        Logs in in the given user and returns the session
         """
 
         if not self.ctx.hasSession():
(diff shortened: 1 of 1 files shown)


components/tools/OmeroFS/src/fsDropBoxMonitorClient.py
```diff
@@ -671,6 +693,7 @@ def importFile(self, fileName, exName):
                 else:
                     self.log.error("%s not found !" % t)
                 self.log.error("***** end of output from importer-cli *****")
+                self.logoutUser(sess)
```
joshmoore: possible in the finally?

components/tools/OmeroFS/src/fsDropBoxMonitorClient.py
```diff
@@ -544,7 +544,7 @@ def getExperimenterFromPath(self, fileId=""):
 
     def loginUser(self, exName):
         """
-        Logins in the given user and returns the client
+        Logs in in the given user and returns the session
```
joshmoore: `in in in in`?

ximenesuk: Pah, over correcting!

components/tools/OmeroFS/src/fsDropBoxMonitorClient.py
ximenesuk: Yes, probably. I did wonder about that.

dominikl:
Imported an image via DropBox, then imported another image via CLI (assuming this is a 'non-destructive' CLI action). Had to log in for the CLI import, new session was created. Good to merge.

ximenesuk:
Thanks, the non-destructive bit wasn't essential just didn't want a reviewer having to delete valuable data :-)

joshmoore:
@ximenesuk : thoughts on handling dangling `omero/sessions` files here?

ximenesuk:
@joshmoore I'm not sure exactly what that would involve.

joshmoore:
https://github.com/ximenesuk/openmicroscopy/blob/dropbox-logout/components/tools/OmeroFS/src/fsDropBoxMonitorClient.py#L663 is effectively calling `bin/omero login` (via `bin/omero import`). The finally block would need a `bin/omero logout` unless there's a way to prevent the file from being created in the first place. @sbesson ?

ximenesuk:
Thanks @joshmoore although I could see there was an implicit log in there I hadn't realised that it was creating  the sessions file. I'll take a look at using a log out unless there is another solution.

ximenesuk:
Closing this for now as the switch to using `logout -k` means the existing commits will need reverting so I may as well force push a new version of this branch.

PR #4316, opened 2015-10-30
Bugfix: Import As freezes

dominikl:
Fixes [Ticket 12929](https://trac.openmicroscopy.org/ome/ticket/12929)
Problem: After an object has been saved the connector was shutdown. This could lead to an abortion of a parallel running import job. This PR removes the connector shutdown, because it's actually not necessary at that point anyway.

**Test**:
Follow the workflow which previously triggered this bug, which is:
- Log in as a group owner or administrator
- *Start*  (not just add to the queue) a long running import job for another user
- Now start another import for the user you're logged in with, but import into a particular dataset which you have to create first on the import dialog (this caused the crash)
- Make sure the imports are passing like expected.

Note: The actual bugfix is just commit 979098b . The first commits add PropertyChangeSupport to the Gateway, i. e. one can add a PropertyChangeListener to the Gateway for debugging purposes, which makes it much easier to monitor what's going on in the Gateway. Any objections on this?

/cc @jburel

### DIFF
--- a/components/blitz/src/omero/gateway/Connector.java
+++ b/components/blitz/src/omero/gateway/Connector.java
@@ -204,6 +204,7 @@ class Connector
      * @param secureClient The entry point to server.
      * @param entryEncrypted The entry point to access the various services.
      * @param encrypted The entry point to access the various services.
+     * @param username The username if this is a derived connector
      * @param logger Reference to the logger.
      * @param elapseTime The time between network check.
      * @throws Exception Thrown if entry points cannot be initialized.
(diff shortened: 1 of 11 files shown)


joshmoore:
```
[nomemoryjava] /home/travis/build/openmicroscopy/openmicroscopy/components/blitz/src/omero/gateway/Gateway.java:70: error: cannot find symbol
[nomemoryjava] import omero.gateway.util.GatewayMonitor;
[nomemoryjava]                          ^
[nomemoryjava]   symbol:   class GatewayMonitor
[nomemoryjava]   location: package omero.gateway.util
```

jburel:
making debugging easier makes sense

components/blitz/src/omero/gateway/Connector.java
```diff
@@ -195,6 +193,26 @@
...
+                entryEncrypted, encrypted, null, logger);
+    }
+    
+    /**
+     * Creates a new instance.
+     *
+     * @param context The context hosting information about the user.
+     * @param secureClient The entry point to server.
+     * @param entryEncrypted The entry point to access the various services.
+     * @param encrypted The entry point to access the various services.
```
mtbc: but it's just a boolean

components/insight/SRC/org/openmicroscopy/shoola/env/data/GatewayMonitor.java
```diff
@@ -0,0 +1,25 @@
+package org.openmicroscopy.shoola.env.data;
```
mtbc: Probably good to have some kind of comment header for the file?

mtbc:
Functionally works fine, just a comment nitpick.

mtbc:
Good to merge.

jburel:
Merging

PR #4322, opened 2015-11-08
Refactor upgrade_check functionalities in the CLI admin plugin

sbesson:
This should allow both the diagnostics and the checkupgrade upgrade check call to share the same logic especially when retrieving the upgrade URL from the server configuration.

To test this PR, run `bin/omero admin diagnostics` and `bin/omero admin checkupgrade` from a server configured to use the staging registry URL and check both `OMERO.diagnostics` and `OMERO.test` agents are recorded in the staging registry DB.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1161,7 +1161,7 @@ OMERO Diagnostics %s
 
         import logging
         logging.basicConfig()
-        check = self.run_upgrade_check("diagnostics", config)
+        check = self.run_upgrade_check(config, "diagnostics")
         if check.isUpgradeNeeded():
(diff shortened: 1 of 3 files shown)


sbesson:
Additional commits will need to be pushed to fix the integration tests - see https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/130

manics:
What's the expected behaviour when `omero admin {diagnostics,checkupgrade}` is run by a different system user?

sbesson:
@manics: i expect the call to work in both cases. Last commit should fix the bug you reported.

sbesson:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/ is green today

pwalczysko:
Passes on ``staging`` as expected.

PR #4323, opened 2015-11-09
Don't try to convert to referenceframe

dominikl:
Don't try to convert the unit for image acquisition position to reference frame, just display as it is.
Fixes [Ticket 13097](https://trac.openmicroscopy.org/ome/ticket/13097)

Test:
Display image acquisition metadata for an image which does not use reference frame as unit, e. g. eel, user-5,image id 26061 (or the images from the bug report, eel, user-5/read-write-1, Dataset name 16890)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerUI.java
@@ -1446,38 +1446,18 @@ class ImViewerUI
 	            }
 	            toolTipText = "";
 	            toolTipText += "Stage coordinates: ";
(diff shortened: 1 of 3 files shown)


pwalczysko:
Suspected of causing the https://trac.openmicroscopy.org/ome/ticket/13099#ticket - to be testied tomorrow.

dominikl:
Additional test:
- Make sure **Insight does not crash** when opening a ```*.dv``` image in full viewer

pwalczysko:

jburel:
@dominikl 
```value == null || value.equals("")``` the method ```isBlank``` from ```CommonsLangUtil``` could be used

jburel:
@dominikl: thanks

PR #4324, opened 2015-11-09
SPW deselection in tree

will-moore:
Fixes https://trello.com/c/eyNU6sWg/25-critical-loss-of-hierarchy-selection

To test:
 - Find / create a Plate with a Run and a Plate without Run (E.g. from Dataset to Plate script).
 - Browse in tree, then select a Well by clicking.
 - Parent Run / Plate should remain selected in tree, while right panel should show the selected well.
 - Try selecting multiple wells by dragging or shift-click. Should get same result.
 - In both cases, toolbar buttons in the tree should get disabled (except those for creating top-level P/D/S).
 - Clicking parent in the tree should re-enable toolbar buttons as before, and show parent in right panel again.


gusferguson:
@will-moore 

Tested using eel merge user-3 read-only-1

Behaves as expected.
Good to merge.

PR #4325, opened 2015-11-09
Use warnings.simplefilter("always") to catch deprecation warnings

sbesson:
Also remove cap on pytest version on the develop branch.

This should allow us to track the latest pytest releases. Additionally this PR can be rebased onto `metadata` to fix the OMERO-unit failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
     - if [[ $BUILD == 'build-python' ]]; then pip install --user -r ./components/tools/OmeroWeb/requirements-py27-nginx.txt; fi
     - export PATH=$PATH:$HOME/.local/bin
     - scc travis-merge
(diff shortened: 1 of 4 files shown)


joshmoore:
Re-launched after a HDF5 travis failure.

sbesson:
NB: this change also exposes another deprecation in `config.py` (`props.getchildren()`) so the Travis failure is expected and I need to fix it.

ximenesuk:
The changes look okay but I'm not sure what I should now be seeing that I wasn't before. For instance,
```
./setup.py test -vst test/unit/clitest/test_sessions.py
```
does display `DeprecationWarning`s, is that the expected outcome?

sbesson:
Closing for now, I will be watching the resolution of https://bitbucket.org/pytest-dev/py/issues/72/

PR #4327, opened 2015-11-10
Gateway login

jburel:
Add option to login by passing arguments
This should be tested using the minimal-omero-client

### DIFF
--- a/components/blitz/src/omero/gateway/LoginCredentials.java
+++ b/components/blitz/src/omero/gateway/LoginCredentials.java
@@ -119,11 +119,11 @@ public class LoginCredentials {
     }
 
     /**
-     * Returns the arguments if set.
+     * Returns the arguments if set as a read-only list.
      *
      * @return See above.
      */
-    public ImmutableList<String> getArguments() { return args; }
+    public List<String> getArguments() { return args; }
 
     /**
      * @return If encryption is enabled
--- a/components/blitz/src/omero/gateway/LoginCredentials.java
+++ b/components/blitz/src/omero/gateway/LoginCredentials.java
@@ -51,12 +51,26 @@ public class LoginCredentials {
     /** ID of the group if not the default group of the user should be used */
     private long groupID = -1;
 
+    /** The connection argument.*/
+    private String[] args;
+
     /**
      * Creates a new instance
      */
     public LoginCredentials() {
         user = new UserCredentials();
         server = new ServerInformation();
+        args = null;
+    }
+
+    /**
+     * Creates a new instance.
+     * @param args The connection argument.
+     */
+    public LoginCredentials(String[] args)
+    {
+        this();
+        this.args = args;
     }
 
     /**
@@ -93,6 +107,13 @@ public class LoginCredentials {
         server.setHostname(host);
         server.setPort(port);
     }
+
+    /**
+     * Returns the arguments if set.
+     *
+     * @return See above.
+     */
+    public String[] getArguments() { return args; }
 
     /**
      * @return If encryption is enabled
--- a/components/blitz/src/omero/gateway/LoginCredentials.java
+++ b/components/blitz/src/omero/gateway/LoginCredentials.java
@@ -20,6 +20,11 @@
  */
 package omero.gateway;
 
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
 /**
  * Holds all necessary information needed for connecting to an OMERO server
  * 
@@ -52,7 +57,7 @@ public class LoginCredentials {
     private long groupID = -1;
 
     /** The connection argument.*/
-    private String[] args;
+    private ImmutableList<String> args;
 
     /**
      * Creates a new instance
@@ -70,7 +75,7 @@ public class LoginCredentials {
     public LoginCredentials(String[] args)
     {
         this();
-        this.args = args;
+        this.args = ImmutableList.copyOf(args);
     }
 
     /**
@@ -113,7 +118,7 @@ public class LoginCredentials {
      *
      * @return See above.
      */
-    public String[] getArguments() { return args; }
+    public ImmutableList<String> getArguments() { return args; }
 
     /**
      * @return If encryption is enabled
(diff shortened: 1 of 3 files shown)


components/blitz/src/omero/gateway/LoginCredentials.java
```diff
@@ -51,15 +51,71 @@
...
+        server.setHostname(host);
+        server.setPort(port);
+    }
+
+    /**
+     * Returns the arguments if set.
+     *
+     * @return See above.
+     */
+    public String[] getArguments() { return args; }
```
joshmoore: Is the returned array intended to be mutable?

jburel:
https://github.com/ome/minimal-omero-client/pull/10 is green. 
It uses the new method

dominikl:
Wanted to test the minimal client yesterday, running it with the arguments ```omero.host=localhost omero.port=4064 omero.user=root omero.password=omero```, get a ```omero.ClientError: No host specified. ``` Not really sure, what I'm doing wrong.

jburel:
it is ```omero.pass```

dominikl:
Ah, sorry. Will test again later.

dominikl:
Tested again with "--", works fine. Good to merge!

components/blitz/src/omero/gateway/LoginCredentials.java
```diff
@@ -51,15 +58,74 @@
...
+        server.setHostname(host);
+        server.setPort(port);
+    }
+
+    /**
+     * Returns the arguments if set.
+     *
+     * @return See above.
+     */
+    public ImmutableList<String> getArguments() { return args; }
```
joshmoore: A heads up that this will tie us to the google API. I don't know if we've yet used it in a public API method (that we intend to support). cc: @mtbc

joshmoore:
Is there a place holder for integration tests for gateway in this repo beyond using `minimal-omero-client`?

components/blitz/src/omero/gateway/LoginCredentials.java
jburel: Could generate an array from the immutable list e.g.
args.toArray(new String[args.size()])

jburel:
we currently have some gateway tests are under integration for testing the services.
I was planning to review that in a follow up PR and use the minimal omero-client for this round

components/blitz/src/omero/gateway/LoginCredentials.java
mtbc: Could just declare `List` return value but actually give them an `ImmutableList` and warn in Javadoc it's read-only.

I'm not aware of where we've used Guava in public API. It's certainly worth thinking about, but this probably falls short of a compelling reason to take the plunge at this point?

jburel: Sounds like a good comprise.

jburel:
Added tests for connection.
This can be run using the following command
```
./build.py -f components/tools/OmeroJava/build.xml test -DTEST=integration/gateway/GatewayUsageTest
```

components/tools/OmeroJava/test/integration/gateway/GatewayUsageTest.java
```diff
@@ -0,0 +1,94 @@
...
+            ExperimenterData root = gw.connect(c);
+            Assert.assertNotNull(root);
+        } catch (Exception e) {
+            throw new DSAccessException("Not able to connect using args",
+                    e);
+        } finally {
+            if (gw != null) {
+                gw.disconnect();
+            }
+            client.__del__();
```
joshmoore: For an `omero.client` that has never had `createSession` called, `__del__()` is likely unneeded if that will simplify tests, since then `new Gateway` could be outside the try/finally reducing down to:

```
+        final Gateway gw = gw = new Gateway(new SimpleLogger());
+        try {
+            ExperimenterData root = gw.connect(c);
+            Assert.assertNotNull(root);
+            throw new DSAccessException("Not able to connect using args",
+                    e);
+        } finally {
+                gw.disconnect();
+        }
```

Personally, I'd also vote for dropping the `catch` block as well and just `throws` statements.

joshmoore:
One minor comment that's for elsewhere: I had missed the addition of the `log` package under `omero`. There might be some room for rationalization there, or at least some planning of which packages should use which other packages so that we don't run out of space under "omero".

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-java/137/testngreports/integration.gateway/GatewayUsageTest/ for the passing integration tests.

dominikl:
Looks good, test also green.

joshmoore:
:thumbsup:

components/blitz/src/omero/gateway/LoginCredentials.java
```diff
@@ -51,15 +57,74 @@
...
+     * 
+     * @param username
+     *            The username
+     * @param password
+     *            The password
+     * @param host
+     *            The server hostname
+     */
+    public LoginCredentials(String username, String password, String host) {
+        this(username, password, host, omero.constants.GLACIER2PORT.value);
```
mtbc: Is there some way to check if `omero.ports.ssl` or somesuch has been configured?

joshmoore: I'm not sure I follow, @mtbc. If you're saying is it necessary to know whether or not to pass 4064 or 4063 as the default, I'd say we always pass 4064. The handshake with the server will then tell if dropping down to 4063 is possible.

mtbc: I was wondering about the case where the OMERO ports have been completely changed from their defaults by configuration. (For instance, maybe we can see if `omero.port` was set in ice.config or somesuch?)

joshmoore: Gotcha. And that makes sense. Here we do need to evaluate the gateway for reusability as a library as opposed to part of an application. It's unlikely that someone would set ice.config or a system property before launching Insight, but as a library, that would be quite a normal usage and the underlying `omero.client` object will read `omero.port` if set in the config.

PR #4330, opened 2015-11-13
add initial version of deep copy to server

mtbc:
To copy instruments for http://trac.openmicroscopy.org/ome/ticket/11532 and ROIs for the folder/template work, some ability for deep copy would be useful. This is the initial version sufficient to serve those two needs. It does not yet offer everything: for instance, original files cannot be copied without a breaking database schema change, so if one tries to copy an image, one will still find acquisition data, etc., but no actual pixels. Check that https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/DuplicationTest/ passes and that `./build.py -f components/blitz/build.xml test -DTEST=omero/cmd/graphs/SpecificityClassifierTest` passes.

To test further, you would have to try creating `DuplicateI` requests from a client. If @ximenesuk decides to provide a CLI interface, I'd recommend it be hidden among the advanced options for now, lest users are too optimistic about what this feature can yet do before we can afford more time for it, not least for adding more integration tests.

--no-rebase as this is a new feature

### DIFF
--- a/components/tools/OmeroJava/test/integration/DuplicationTest.java
+++ b/components/tools/OmeroJava/test/integration/DuplicationTest.java
@@ -1,0 +1,1116 @@
+/*
+ * Copyright (C) 2015 University of Dundee & Open Microscopy Environment.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package integration;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import omero.RLong;
+import omero.RType;
+import omero.ServerError;
+import omero.cmd.Delete2;
+import omero.cmd.Duplicate;
+import omero.cmd.DuplicateResponse;
+import omero.cmd.ERR;
+import omero.gateway.util.Requests;
+import omero.model.Annotation;
+import omero.model.AnnotationAnnotationLink;
+import omero.model.AnnotationAnnotationLinkI;
+import omero.model.Channel;
+import omero.model.DoubleAnnotationI;
+import omero.model.Ellipse;
+import omero.model.EllipseI;
+import omero.model.FileAnnotation;
+import omero.model.FileAnnotationI;
+import omero.model.IObject;
+import omero.model.Image;
+import omero.model.ImageAnnotationLink;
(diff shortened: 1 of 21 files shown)


ximenesuk:
I would be happy to test this via a basic CLI client though I would probably hesitate on releasing anything on that front without further discussion.

mtbc:
Sounds sensible to me. As far as I'm concerned, anything user-facing can wait for 5.3.

components/tools/OmeroJava/test/integration/DuplicationTest.java
```diff
@@ -0,0 +1,1112 @@
...
+    }
+
+    /**
+     * Assert that the given instances have the same property values.
+     * (The properties to compare are based on the behavior of {@link ModelMockFactory#createImage(int, int, int, int, int)}.)
+     * @param original the original instance
+     * @param duplicate the duplicate instance that is expected to correspond to the original
+     */
+    private static void assertSameProperties(Pixels original, Pixels duplicate) {
+        Assert.assertEquals(duplicate.getPhysicalSizeX().getUnit(), original.getPhysicalSizeX().getUnit());
```
joshmoore: This makes me think of an `omero.cmd.Equivalent` graph command.... :smile:

components/tools/OmeroJava/test/integration/DuplicationTest.java
```diff
@@ -0,0 +1,1112 @@
...
+    }
+
+    /**
+     * Assert that the given instances have the same property values.
+     * (The properties to compare are based on the behavior of {@link ModelMockFactory#createImage(int, int, int, int, int)}.)
+     * @param original the original instance
+     * @param duplicate the duplicate instance that is expected to correspond to the original
+     */
+    private static void assertSameProperties(Channel original, Channel duplicate) {
+        // nothing
```
joshmoore: Why is there no channel comparison?

components/tools/OmeroJava/test/integration/DuplicationTest.java
```diff
@@ -0,0 +1,1112 @@
...
+                sb.append(followLongAnnotations((LongAnnotation) result, depth - 1));
+            }
+        }
+        sb.append(']');
+        return sb.toString();
+    }
+
+    /**
+     * Create then duplicate an image and test that the properties of the instances in it and its subgraph are correctly duplicated.
+     * @throws Exception unexpected
```
joshmoore: Duplicating the pixels object is an exercise for the reader, right?

components/tools/OmeroJava/test/integration/DuplicationTest.java
mtbc: The model mock factory doesn't add any interesting properties to it.

joshmoore:
Duplicating an image with a file annotation is certainly an interesting one!

components/tools/OmeroJava/test/integration/DuplicationTest.java
mtbc: The pixels object does get duplicated (at least by default) in duplicating the image.

joshmoore: sorry, `s/pixels object/pixels data/`

mtbc: Indeed, anything that somehow lives in an original file will wait for 5.3, it shouldn't present much of a problem though.

components/tools/OmeroJava/test/integration/DuplicationTest.java
mtbc: That could actually be useful for de-duplication on import: e.g., not creating a duplicate of many tags because an image finally returned (via whatever means) to the group in which it was originally tagged.

joshmoore:
:+1: Tests are all passing. I'd be inclined to get this in and wait until we have a CLI command (or other usage) for further testing.

ximenesuk:
That's fine by me, my testing of this got waylaid by docker work.

PR #4266, opened 2015-10-14
Python gateway & webclient delete improvements

ximenesuk:
This PR addresses the various items in https://trello.com/c/WMnOBpC6/31-improve-deleteobjects-deprecate-deleteobjectdirect-inpython-gateway both RFEs and deprecation of `deleteObjectDirect`.

Tests, both `OmeroPy` and `OmeroWeb` should pass.

Web functionality to be tested is that which deletes and unlinks P/D/I, S/P and Annotations. I.e quite wide-ranging, but concentrating on unlinking and deleting all types of annotations would be useful.

PR #4332, opened 2015-11-16
Jstree code cleanup

will-moore:
This removes and cleans up code no-longer used since the loading of Projects, Datasets, Screens & Orphaned images for jsTree and centre panel are handled by /api/ methods.

To test:
 - Browse tree, load Dataset & Plate data into centre panel.
 - Brief code review could be useful

cc @aleksandra-tarkowska

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
@@ -433,24 +433,6 @@ class BaseContainer(BaseController):
         self.c_size = (len(pr_list) + len(ds_list) + len(im_list) +
                        len(sc_list) + len(pl_list) + len(pa_list))
 
-    def listImagesInDataset(self, did, eid=None, page=None,
-                            load_pixels=False):
-        if eid is not None:
-            if eid == -1:       # Load data for all users
-                eid = None
-            # else:
-            #     self.experimenter = self.conn.getObject("Experimenter", eid)
-        im_list = list(self.conn.listImagesInDataset(
-            oid=did, eid=eid, page=page, load_pixels=load_pixels))
-        # List is already sorted by name, id in query
-        # im_list.sort(key=lambda x: (x.getName().lower(), x.getId()))
-        self.containers = {'images': im_list}
-        self.c_size = self.conn.getCollectionCount(
-            "Dataset", "imageLinks", [long(did)])[long(did)]
-
-        if page is not None:
-            self.paging = self.doPaging(page, len(im_list), self.c_size)
-
     def listContainerHierarchy(self, eid=None):
         if eid is not None:
             if eid == -1:
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
@@ -460,27 +460,6 @@ class BaseContainer(BaseController):
             'plates': pl_list}
         self.c_size = len(pr_list)+len(ds_list)+len(sc_list)+len(pl_list)
 
-    def listOrphanedImages(self, eid=None, page=None):
-        if eid is not None:
-            if eid == -1:
-                eid = None
-            else:
-                self.experimenter = self.conn.getObject("Experimenter", eid)
-        else:
-            eid = self.conn.getEventContext().userId
-
-        params = omero.sys.ParametersI()
-        if page is not None:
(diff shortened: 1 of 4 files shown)


joshmoore:
```
./components/tools/OmeroWeb/omeroweb/webclient/urls.py:71:80: E501 line too long (84 > 79 characters)
```

Also, @will-moore, any chance these methods are being consumed and still work correctly elsewhere? If so, a `DeprecationWarning` are likely more appropriate for a point release. cc: @chris-allan

will-moore:
@joshmoore As I understand it, our policy has always been that ```webgateway``` is the public facing part of OMEROweb https://www.openmicroscopy.org/site/support/omero5.2/developers/Web/WebGateway.html and other apps are internal.

sbesson:
@will-moore: this is not what I understood from the discussion in https://github.com/openmicroscopy/openmicroscopy/pull/4266. Where is this policy documented? As far as any external consumer of our Python API can see, the `omeroweb.webclient` module is  exposed in the [public API documentation](http://downloads.openmicroscopy.org/omero/5.2.0/api/python/omeroweb/omeroweb.webclient.html) and can legitimately be considered as public. Reading your [comment](https://github.com/openmicroscopy/openmicroscopy/pull/4266#issuecomment-149610893) on this PR, I understood the objective was to split public from non-public API in a future major release.

will-moore:
@sbesson @joshmoore The discussion in #4266 suggested that the distinction between public and internal parts of OMEROweb be clarified soon, with the details "to be discussed". But never mentions waiting for the next major release.
If we consider everything under ```webclient``` app to be internal, then removing unused code and moving stuff we want to be public to another app wouldn't be "breaking" changes.
The distinction between public webgateway urls / views and internal webclient is made in our web documentation, where OMEROweb gets a page describing the available json methods in webgateway, https://www.openmicroscopy.org/site/support/omero5.2/developers/Web/WebGateway.html but we don't tell anyone to do the same for webclient. 
The webclient views methods are not generally reusable Python API. They return http responses containing fragments of the webclient UI as html and are highly customised to providing what the webclient needs, not designed to be an API.
We also went to some effort to update the public urls.py for webgateway so that the docs generated are in some way usable: http://downloads.openmicroscopy.org/omero/5.2.0/api/python/omeroweb/omeroweb.webgateway.html#module-omeroweb.webgateway.urls
This is not the case for webclient, where you actually see nothing for urls: http://downloads.openmicroscopy.org/omero/5.2.0/api/python/omeroweb/omeroweb.webclient.html#module-omeroweb.webclient.urls

aleksandra-tarkowska:
Sorry, took me too long to get to this PR. I agree with @will-moore, we didn't expose much urls as they just render templates. There is a low chance people were using them. With the current json API there is no need to keep them. We should be moving to ``webclient/api`` as much as we can. @will-moore what is your plan in terms of moving more to new API? Would it be hard to merge that PR to 5.3?

jburel:
For any gateway plan,  please add cards to https://trello.com/b/Tb2WuoBZ/gateway-api so they can be discussed and scheduled

joshmoore:
@will-moore - thanks for the clarification. Can I propose that your explanation be migrated to ome-documentation (maybe https://www.openmicroscopy.org/site/support/omero5.2-staging/developers/Web.html), and perhaps either an email to devel or a blog post with the plan? As long as there's consensus or general knowledge of that plan, I don't think anyone can complain. It would just be unfortunate if releasing 5.2.1 or 5.2.2 forced someone to emergency release their app. If we give warning, that shouldn't be the case.

will-moore:
Card created at https://trello.com/c/Wk1NO85D/12-web-json-api. Need to schedule a chat / meeting. cc @chris-allan interested?

joshmoore:
big :thumbsup: thanks @will-moore

chris-allan:
Happy to participate.

aleksandra-tarkowska:
@will-moore I am fine with the cleanup

will-moore:
From @joshmoore comment above, I understood that we are OK with removing this functionality in 5.2.1 or 5.2.2 if we give sufficient warning (docs and announcement warning in good time). But discussion in today's client meeting with @jburel suggested that we can't change anything until 5.3.0? So what's the consensus on this? Should I just close this PR?

jburel:
To clarify, we discussed during the meeting marking the methods as deprecated in 5.2.x (w/o having the list of the methods) so they can be removed  in 5.3.
There is a limited change they have been used.

PR #4337, opened 2015-11-17
jquery-migrate cleanup

will-moore:
This fixes the last few warnings from the jQuery migrate plugin and removes the plugin.

This plugin was added to ease the transition from jQuery 1.9 -> 1.11 but now we have moved all our code from using 1.9 features, this can be removed.

First, I looked for JQMIGRATE warnings in console.log while using a large range of workflows in web.
A handful of these remaining warnings were fixed by the first few commits (although none of them would have been critical if overlooked).
Then the plugin was removed.

This shouldn't cause any bugs, but if there's anywhere I missed where we're still using critical jQuery 1.9 features then this could cause a problem.

Since it's not feasible to "test everything", we should at least check that this doesn't break a handful of key workflows, then keep an eye out for any other bugs in future.


pwalczysko:
I see some delays in loading of thumbnails in the Search Results tab. For example, login as user-4, search for ``a``. Observe that for couple of seconds (ca 10) you see no Project icons, instead just the word ``Project`` is highlighted. The same happens with some images when searching for ``P*`` and selecting Projects and Images in the search options.

pwalczysko:
Link to an image which is on a second page of large dataset does no work (it expands the dataset allright, but does not load the second page and highlight the image). This will of course happen also when you use the ``Browse`` from search.
Example: https://eel.openmicroscopy.org/merge/webclient/?show=image-24548 (user-4)

pwalczysko:
The link to image on second page https://github.com/openmicroscopy/openmicroscopy/pull/4337#issuecomment-157682906 does not work on eel latest either.
Went through [web lightweight scenario](https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/WebLightTesting.html) and everything seemed okay. Maybe wait till the end of the week for "ripening" (some inadvertent activity might discover bug which are not included in web-lightweight scenario)

will-moore:
@pwalczysko Sorry - should have replied sooner. Yes, the failure of linking to images that are not on the first page of a Dataset is a known issue and has never worked. I was trying to remember if we had a card/ticket for that but haven't tried searching for it yet.

mtbc:
@pwalczysko: So good to merge now? No other problems noticed?

pwalczysko:
No new bugs popped up.
Ready to merge.

will-moore:
This is causing failure to rename Project, as caught by Robot test https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/171/robot/Web%20&%20Web/Web_1/Post%20Test/Test%20Edit%20Project/ cc @bramalingam

will-moore:
Robot tests should be fixed by that last commit.

bramalingam:
Tested on eel.merge and works as expected. The robot tests are passing as well, good to merge.

PR #4340, opened 2015-11-18
Remove unused webgateway/base_site.html template

will-moore:
As suggested in https://trello.com/c/SKLuiD4z/54-jstree-follow-up, this removes the unused webgateway/base_site.html template.

To test:
 - Easiest way to check if this is really unused is to search OMEROweb code.


joshmoore:
@will-moore : I don't mean to beat a dead horse, but "Easiest way to check if this is really unused is to search OMEROweb code" doesn't really work when there are external projects which we can't just grep for. This is in `gateway` so according to the recent discussion, someone could have legitimately consumed it, right?

joshmoore:
:+1: for whatever form of deprecation you guys think is appropriate and a card (or other reminder) to remove the deprecated features for 5.3 etc.

PR #4341, opened 2015-11-18
Incorrect redirect to http instead of https

aleksandra-tarkowska:
This PR should fix incorrect redirection to ``​http://`` instead of ​``https://``

```
curl -i -s -c cookies.txt -b cookies.txt -e https://omero.local/omero/webclient/login/ -d 'csrfmiddlewaretoken=kyuLRduNLY3OQNUMUBiFjuQMDYrcEwt4&username=user-1&password=ome&server=2&url=url=%2Fmerge%2Fwebclient%2F' -X POST https://omero.local/omero/webclient/login/
HTTP/1.1 302 FOUND
Server: nginx/1.8.0
Date: Wed, 18 Nov 2015 13:00:58 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Vary: Cookie
Location: https://omero.local/omero/webclient/
Set-Cookie: sessionid=jix75augbw8i6xdwvydru249sq8vaj9n; httponly; Path=/
```

cc @manics see https://trello.com/c/7eFWt9Bj/8-django-secure-proxy-ssl-header

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -433,7 +433,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
     "omero.web.logdir":
         ["LOGDIR", LOGDIR, str, "A path to the custom log directory."],
     "omero.web.secure_proxy_ssl_header":
-        ["SECURE_PROXY_SSL_HEADERO",
+        ["SECURE_PROXY_SSL_HEADER",
          '[]',
          json.loads,
          ("A tuple representing a HTTP header/value combination that "
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -437,10 +437,10 @@ CUSTOM_SETTINGS_MAPPINGS = {
          '[]',
          json.loads,
          ("A tuple representing a HTTP header/value combination that "
-          "signifies a request is secure. Example ``SECURE_PROXY_SSL_HEADER"
-          " = ('HTTP_X_FORWARDED_PROTO', 'https')``. For more details see "
-          "https://docs.djangoproject.com/en/1.8/ref/settings/"
-          "#secure-proxy-ssl-header.")],
+          "signifies a request is secure. Example "
+          "``'[\"HTTP_X_FORWARDED_PROTO_OMERO_WEB\", \"https\"]'``. "
+          "For more details see :djangodoc:`secure proxy ssl header <ref/"
+          "settings/#secure-proxy-ssl-header>`.")],
 
     # Public user
     "omero.web.public.enabled":
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -437,7 +437,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
          '[]',
          json.loads,
          ("A tuple representing a HTTP header/value combination that "
-          "signifies a request is secure.Example ``SECURE_PROXY_SSL_HEADER"
+          "signifies a request is secure. Example ``SECURE_PROXY_SSL_HEADER"
           " = ('HTTP_X_FORWARDED_PROTO', 'https')``. For more details see "
           "https://docs.djangoproject.com/en/1.8/ref/settings/"
           "#secure-proxy-ssl-header.")],
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -432,6 +432,15 @@ CUSTOM_SETTINGS_MAPPINGS = {
          "The name to use for session cookies"],
     "omero.web.logdir":
         ["LOGDIR", LOGDIR, str, "A path to the custom log directory."],
+    "omero.web.secure_proxy_ssl_header":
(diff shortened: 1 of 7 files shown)


components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -432,6 +432,15 @@ def leave_none_unset_int(s):
          "The name to use for session cookies"],
     "omero.web.logdir":
         ["LOGDIR", LOGDIR, str, "A path to the custom log directory."],
+    "omero.web.secure_proxy_ssl_header":
+        ["SECURE_PROXY_SSL_HEADERO",
+         '[]',
+         json.loads,
+         ("A tuple representing a HTTP header/value combination that "
+          "signifies a request is secure.Example ``SECURE_PROXY_SSL_HEADER"
```
joshmoore: missing a space.

manics:
Irrespective of whether this fixes the http/https bug or not it should still be useful for https://trello.com/c/7eFWt9Bj/8-django-secure-proxy-ssl-header
I'll test just this particular scenario.

manics:
After a lot of testing...

## No config options set:

http and https works as expected:
- https is redirected to https
- http is redirected to http
- The URL shown in `Link to this Image` uses the correct protocol

## omero.web.application_server.host set to non-localhost:
https is always redirected to http

This could be because gunicorn ignores the header for security unless it's specifically enabled as a gunicorn arg http://gunicorn-docs.readthedocs.org/en/19.3/deploy.html#nginx-configuration 

## omero.web.application_server.host set to non-localhost, `--wsgi-args=--forwarded-allow-ips=*`:
http and https works as expected:
- https is redirected to https
- http is redirected to http
- The URL shown in `Link to this Image` uses the correct protocol

Note `*` should also work with just a single IP

manics:
FYI, If you double-quote the IP it doesn't work: `omero web restart --wsgi-args='--forwarded-allow-ips="10.2.1.66"'`

aleksandra-tarkowska:
@manics and I were investigatin that issue and my comments are in https://github.com/openmicroscopy/management_tools/pull/77#issuecomment-158083311

aleksandra-tarkowska:
To confirm I added docker IP ```+ bin/omero web start --wsgi-args=--forwarded-allow-ips=...,DOCKER_IP```

Then added reverse proxy on docker as follow:

```
upstream omeroweb_merge{
        server cowfish...:4080 fail_timeout=0;
   }
```

that gives:

```
$ curl -i -k https://ola-test-1.docker.../merge/
HTTP/1.1 301 MOVED PERMANENTLY
Server: nginx/1.8.0
Date: Thu, 19 Nov 2015 16:06:37 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Last-Modified: Thu, 19 Nov 2015 16:06:37 GMT
Expires: Thu, 19 Nov 2015 16:06:37 GMT
Location: https://ola-test-1.docker.../merge/webclient/
Cache-Control: max-age=0
```

aleksandra-tarkowska:
I will also open a PR to allow ``bin/omero config set omero.web.gunicorn.wsgiargs string`` that should simplify the entire process

manics:
Tested a multiple proxy configuration:

https://nginx-public-proxy/ -> http://omero-nginx-proxy/

Using the default settings this redirects to http since the internal communication with `omero-nginx-proxy` is over `http`, which means Django sees `X-Forwarded-Proto=http`. This can be overriden by setting a custom header in the front-end `nginx-public-proxy`:
```
proxy_set_header X-Forwarded-Proto-Omero-Web $scheme;
```
and setting the new property defined in this PR:
```
omero config set omero.web.secure_proxy_ssl_header '["HTTP_X_FORWARDED_PROTO_OMERO_WEB", "https"]'
```
Note the header name gets mangled as described in https://docs.djangoproject.com/en/1.8/ref/settings/#secure-proxy-ssl-header which can be confusing at first. Also json doesn't support `(`tuples`)` so you have to use `[`square brackets`]`.

Once that's in place, everything works with the two proxies:
- https is redirected to https
- http is redirected to http
- The URL shown in Link to this Image uses the correct protocol

So apart from the minor comment about json brackets, this is good to merge.

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -432,6 +432,15 @@ def leave_none_unset_int(s):
          "The name to use for session cookies"],
     "omero.web.logdir":
         ["LOGDIR", LOGDIR, str, "A path to the custom log directory."],
+    "omero.web.secure_proxy_ssl_header":
+        ["SECURE_PROXY_SSL_HEADER",
+         '[]',
+         json.loads,
+         ("A tuple representing a HTTP header/value combination that "
+          "signifies a request is secure. Example ``SECURE_PROXY_SSL_HEADER"
+          " = ('HTTP_X_FORWARDED_PROTO', 'https')``. For more details see "
```
manics: You can't use `()` in json, `[]` seems to work

aleksandra-tarkowska:
@manics thanks for testing, comment updated

manics:
Text changes look fine, though I haven't checked it renders correctly. Merge and fix later if necessary?

joshmoore:
Filed a card for the follow up: https://trello.com/c/BJnz66v7/89-regenerate-https-docs

PR #4343, opened 2015-11-19
Handle display of pixel size units for each dimension

will-moore:
See https://trello.com/c/7IzMpre0/15-unit-display-in-web

Previously we got the suitable units for sizeX and only displayed these, alongside the most suitable value for sizeX, sizeY and sizeZ. So, if the sizeZ for example was using different units this would appear incorrect:

screen shot 2015-11-19 at 10 12 06

Now, if units are different for any dimension, we show them:

screen shot 2015-11-19 at 10 16 26

This is slightly different from Insight's strategy, but to get it exactly the same will require a bigger re-working of the code here (worth it)?

screen shot 2015-11-19 at 10 23 41

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/core_metadata.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/core_metadata.html
@@ -40,15 +40,27 @@
         <td id='pixels_type'>{{ image.getPixelsType }}</td>
     </tr>
     <tr>
-        <th>Pixels Size (XYZ) ({{ image.getPixelSizeUnits }}):</th>
+        {% with psX=image.getPixelSizeXMicrons psY=image.getPixelSizeYMicrons psZ=image.getPixelSizeZMicrons %}
+            {% with uX=psX|lengthunit uY=psY|lengthunit uZ=psZ|lengthunit %}
+        <th>Pixels Size (XYZ) ({{ uX }}):</th>
         <td id='pixels_size'>
-            <div class='tooltip'>{{ image.getPixelSizeXMicrons|lengthformat|floatformat:2 }} x {{ image.getPixelSizeYMicrons|lengthformat|floatformat:2 }} 
-                {% if image.getPixelSizeZMicrons %} x {{ image.getPixelSizeZMicrons|lengthformat|floatformat:2 }} {% endif %}
+            <div class='tooltip'>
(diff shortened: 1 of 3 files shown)


pwalczysko:
Sorry, did not manage to have a look today because of the Figure. Maybe over the weekend.

mtbc:
Needs relisting?

pwalczysko:
Tooltip issues:
screen shot 2015-12-01 at 16 31 05
screen shot 2015-12-01 at 16 35 55

pwalczysko:
Although there is a discrepancy between Insight and Web, I think the solution proposed here is acceptable.
Unlike in the header of the PR, the current behaviour if the units in X, Y, Z are not the same is actually:
  - do not show any units behind the ``Pixel Size (XYZ)``
  - show the units in brackets for each of the dimansions (see screenshot)
screen shot 2015-12-01 at 16 52 38

I think it can be argued that such tactics is even more flexible than what Insight does, especially in cases where the dimensions wildly differ between X and Z.

The only issues here are the tooltip problems https://github.com/openmicroscopy/openmicroscopy/pull/4343#issuecomment-161024684

pwalczysko:
Tooltip is okay in both cases highlighted above.
All the other features work as expected. 
Ready to merge.

jburel:
@will-moore: could you make one method only with size as parameter?
We may want to unify the UI in insight cc @dominikl

dominikl:
This has been changed so many times. Is the style shown in @pwalczysko 's screenshot what we're finally going for, for the metadata panel as well as tooltip? ```Pixel Size (XYZ): 2.00 (mm) x 2.00 (mm) x 2.00 (km)``` I. e. no unit in the "Pixel Size" label, therefore the unit after each value in brackets, and value rounded to two decimal places. And what is the value, the original one or converted to a "reasonable" value?

will-moore:
@jburel Not sure what you're asking for. I think we're close enough between web and Insight now - Getting any closer just doesn't seem to be worth the effort?

jburel:
sorry @will-moore I missed your comment
you have 3 methods doing the same thing ```def getPixelSizeXWithUnits(self):``` one with sizeX, one with sizeY and one with sizeZ
one method with parameters doing the conversion will make more sense
i.e. pseudo-code
``getPixelSizeXWithUnits(self): `` uses ``convertPixelsSize(sizeX)`` to do the work

will-moore:
@jburel Last commit cleans up the code duplication

jburel:
@will-moore: Thanks, retested with your last commit. 
Merging

PR #4344, opened 2015-11-19
Gateway followup critical

dominikl:
Contains critical bugfixes which have to go into 5.2.1

- Removes the leaves() parameter where it can be dangerous (accidentally pull *everything* from the server)

### DIFF
--- a/components/blitz/src/omero/gateway/facility/BrowseFacility.java
+++ b/components/blitz/src/omero/gateway/facility/BrowseFacility.java
@@ -454,12 +454,9 @@ public class BrowseFacility extends Facility {
      */
     public Collection<DatasetData> getDatasets(SecurityContext ctx) {
         try {
-            ParametersI param = new ParametersI();
-            param.leaves();
-
             IContainerPrx service = gateway.getPojosService(ctx);
             List<IObject> datasets = service.loadContainerHierarchy(PojoMapper
-                    .getModelType(DatasetData.class).getName(), null, param);
+                    .getModelType(DatasetData.class).getName(), null, null);
 
             Collection<DatasetData> result = new ArrayList<DatasetData>(
                     datasets.size());
@@ -527,8 +524,7 @@ public class BrowseFacility extends Facility {
         try {
             ParametersI param = new ParametersI();
             param.exp(omero.rtypes.rlong(ownerId));
-            param.leaves();
-
+            
             IContainerPrx service = gateway.getPojosService(ctx);
             List<IObject> datasets = service.loadContainerHierarchy(PojoMapper
                     .getModelType(DatasetData.class).getName(), null, param);
--- a/components/blitz/src/omero/gateway/facility/BrowseFacility.java
+++ b/components/blitz/src/omero/gateway/facility/BrowseFacility.java
@@ -84,7 +84,7 @@ public class BrowseFacility extends Facility {
      * @return See above.
      * @throws DSOutOfServiceException  If the connection is broken, or logged in.
      */
-    public Set<DataObject> loadHierarchy(SecurityContext ctx, Class rootType,
+    public Collection<DataObject> loadHierarchy(SecurityContext ctx, Class rootType,
             long userId) throws DSOutOfServiceException {
         ParametersI param = new ParametersI();
         if (userId >= 0) {
@@ -105,7 +105,7 @@ public class BrowseFacility extends Facility {
      * @return See above.
      * @throws DSOutOfServiceException If the connection is broken, or logged in.
      */
-    public Set<DataObject> loadHierarchy(SecurityContext ctx, Class rootType,
+    public Collection<DataObject> loadHierarchy(SecurityContext ctx, Class rootType,
             List<Long> rootIDs, Parameters options)
             throws DSOutOfServiceException {
         try {
--- a/components/blitz/src/omero/gateway/facility/BrowseFacility.java
+++ b/components/blitz/src/omero/gateway/facility/BrowseFacility.java
@@ -83,9 +83,10 @@ public class BrowseFacility extends Facility {
      * @param userId The user's to retrieve the data to handle.
      * @return See above.
      * @throws DSOutOfServiceException
+     * @throws DSAccessException 
      */
     public Collection<DataObject> getHierarchy(SecurityContext ctx, Class rootType,
-            long userId) throws DSOutOfServiceException {
+            long userId) throws DSOutOfServiceException, DSAccessException {
         ParametersI param = new ParametersI();
         if (userId >= 0) {
             param.exp(omero.rtypes.rlong(userId));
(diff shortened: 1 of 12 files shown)


components/blitz/src/omero/gateway/LoginCredentials.java
```diff
@@ -30,6 +30,9 @@
 
 public class LoginCredentials {
 
+    /** The default port of the OMERO server */
+    private static final int DEFAULT_PORT = 4064;
+
```
dominikl: This is fixed anyway in  #4327 (using the GLACIER2PORT constant).

components/blitz/src/omero/gateway/LoginCredentials.java
```diff
@@ -30,6 +30,9 @@
 
 public class LoginCredentials {
 
+    /** The default port of the OMERO server */
+    private static final int DEFAULT_PORT = 4064;
```
joshmoore: `omero.constants.GLACIER2PORT` ?

jburel:
@dominikl: could you add a test to ```BrowserFacilitytest```?

dominikl:
Ok, I'll add a test. And I remove the first commit again, you cherry-picked (and fixed) it anyway.

jburel:
you don't need ```e2c0795``` I cherry-picked your commit see gh-4327

jburel:
Will review it today since i went through the install's one first

jburel:
The same should apply to ```getDatasets(SecurityContext ctx, long ownerId)```

jburel:
I will also add a check if ```ids == null``` to methods like ```getDatasets(SecurityContext ctx, long ownerId, Collection<Long> ids)```

dominikl:
Added @jburel 's suggestions and also changed the loadHierarchy methods to return a Collection instead of a Set in order to have consistent method signatures.

components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -1632,7 +1632,13 @@ Set loadContainerHierarchy(SecurityContext ctx, Class rootType,
 	{
 	    try {
             BrowseFacility f = gw.getFacility(BrowseFacility.class);
-            return f.loadHierarchy(ctx, rootType, rootIDs, options);
+            // tmp solution, should be changed to Collection<DataObject> throughout
```
mtbc: might want TODO

components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -1632,7 +1632,13 @@ Set loadContainerHierarchy(SecurityContext ctx, Class rootType,
...
+            Set result = new HashSet();
+            Collection<DataObject> col = f.loadHierarchy(ctx, rootType,
+                    rootIDs, options);
+            for (DataObject obj : col)
```
mtbc: just use the copying constructor of `HashSet` rather than a loop?

dominikl: Thanks... of course, using the constructor directly makes more sense.

dominikl:
You're right. I missed that earlier for 5.2.0 unfortunately. As the Gateway still is in an early stage, do you think we already have to use the deprecation process now?

jburel:
We did not clearly state in the announcement that method signatures could be changed in 5.2.x
cf. https://www.openmicroscopy.org/community/viewtopic.php?f=11&t=7922
so better to go for the safest option

components/blitz/src/omero/gateway/facility/BrowseFacility.java
```diff
@@ -84,6 +84,55 @@
...
+    /**
+     * Retrieves hierarchy trees rooted by a given node.
+     * i.e. the requested node as root and all of its descendants.
+     *
+     * @param ctx The security context.
+     * @param rootType The type of node to handle.
+     * @param rootIDs The node's id.
+     * @param options The retrieval options.
+     * @return See above.
+     * @throws DSOutOfServiceException If the connection is broken, or logged in.
```
mtbc: That's interesting. This method won't work with logged-in connections?

jburel: What do you mean?

mtbc: it throws an exception if the connection is logged in, according to the comment (also on other methods)

jburel: I see what you mean
That's an error in the doc that was not fixed when migrating methods.
Copy and paste ;-)

dominikl: Yes, I should check the comments more carefully instead of just copy/pasting them.

dominikl:
Removed the wrong comments. The description of DSOutOfServiceException and DSAccessException is actually enough, no need to repeat that on every method which throws them.
Unfortunately I noticed another problem. Some methods catch the exceptions and just log them. This should not be the case, as a client of the Gateway doesn't know if the requested data just doesn't exist or if an error has happened. But changing this would change the method signatures thus the API again.

mtbc:
Should this still be listed and reviewed though despite the other problem?

jburel:
@mtbc: yes. ideally we want the fixes in 5.2.1

dominikl:
@jburel Are you happy with the PR so far? What do you think about changing the methods which now catch the exceptions to throw them instead, like it should have been done in first place, e. g. ```getProjects(...)``` etc.

jburel:
throwing exceptions will allow client using to take a decision e.g. shutdown client since the server is down.
Problem is that we will change the method in a point release.
this is a dilemma: either we fix now and we must make it very clear in the announcement that gateway is likely to be modified/adjusted during 5.2.x, or we wait until 5.3 
@joshmoore strong feeling?

joshmoore:
I'd personally warn against something like `} catch (Throwable t) { log(); }` (as in [line 116](https://github.com/openmicroscopy/openmicroscopy/pull/4344/files#diff-5a49d72a6a2b6f1ac153b4bdef591916R116)) as part of a public API and I *would* consider changes to exception behavior to be part of the public API. Since those two points are in conflict, I'd assume choose based on which is going to be the least burden to users overall.

dominikl:
I don't think many people are using the gateway methods yet, and it was announced as still being in an early stage, so in my opinion it shouldn't be a too big burden to add the throw statements.

Instead of catching and logging, exceptions are processed by the ```handleException``` method, https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/facility/Facility.java#L211 which determines if a ```DSOutOfServiceException``` or a ```DSAccessException``` should be thrown. That's a method I copied over from the old gateway. But is that really a correct implementation. If the problem is connection related ```ConnectionStatus != ConnectionStatus.OK``` the method just returns, no Exception is thrown. That doesn't appear correct to me. Shouldn't a ```DSOutOfServiceException``` been thrown in that case?

joshmoore:
In the more general case, I wonder if an `ExceptionFacility` (`ExceptionHandler`?) wouldn't also be eventually appropriate to allow setting the mapping from one exception to another.

dominikl:
Hm... do you mean that the user of the gateway can provide an implementation of https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/facility/Facility.java#L211 and decide how the underling exception should be handled?

joshmoore:
@dominikl : I was thinking of something more orthogonal, i.e. a "service", "handler", etc., which all facilities use.

dominikl:
As we are too close to the 5.2.1 code freeze, I'm going to implement that for 5.2.2 (added to [Trello card: Gateway improvements](https://trello.com/c/0hvmrXSL/10-gateway-improvements) for now)

joshmoore:
@dominikl : makes sense. As long as all instances of `logError` on a `catch (Throwable t)` are gone, then this is probably good for the moment. Next step is probably to review exceptions thrown and have a clear specification.

jburel:
@dominikl 
The gateway tests like getProjects need to throw exception cf. BrowserFacilityTest

dominikl:
Hm, which methods do still miss the throws clause?

jburel:
@dominikl: sorry I was on the wrong branch when I was checking.
Too many opened terminals between omero/omero-install/doc etc.
Merging

PR #4345, opened 2015-11-19
Scroll to selected thumbnail

will-moore:
This addresses comment at https://github.com/openmicroscopy/openmicroscopy/pull/4242#issuecomment-150147878 which is listed on https://trello.com/c/6ugBp0HG/5-web-center-panel-follow-up.

To test, find a dataset with a large number of images (so that the centre panel needs scrolling).
 - In the tree, select the dataset, then select an image near the bottom. Centre panel should scroll to that thumbnail.
 - Hotkeys should continue to work in jsTree as usual, with centre panel scrolling as needed to show selected thumbnail.
 - This should also work when in "table" view in centre panel
 - Switching between table and icon views should maintain the selected thumbnail in view (scroll to show it if needed)


mtbc:
Needs listing?

gusferguson:
@will-moore 

Tested with eel user-3 read-only-1.

Behaves as expected.
Good to merge.

PR #4346, opened 2015-11-19
Libraries cleanup

sbesson:
See https://trello.com/c/2qw5o7PL/55-clean-lib-repository

This PR cleans up the remaining server and client JARs stored under `lib/repository` available either via Maven Central or via the OME artifactory. As this PR is for a patch release, it should not involve any client or server dependency bump. A couple of `ivy.xml` files have been simplified as libraries are pulled as dependencies via the POMs.

All remaining items under `lib/repository` should be categorized as:
- JARs used for the build system (`ant*`)
- bundles/eggs bundled in the Python build (`ipython`)
- libraries defined upstream with a set of dependencies/versions that differs from ours (`hibernate`). These libraries should be reviewed and cleaned separately, most likely when they are upgraded.

To test this PR compare the dependency trees under `share/{server,client}/dependencies.html` with and without this PR e.g. by using the server artifacts from OMERO-DEV-latest and OMERO-DEV-merge-build and check all tests remain green.


sbesson:
Removing `breaking` label as per https://ci.openmicroscopy.org/view/Breaking/job/OMERO-DEV-breaking-trigger/543/

joshmoore:
Looking at the dependencies.html files, I assume the evictions are matter to handle in a follow-up?

sbesson:
In terms of evictions, most of the dependencies should be identical. Pulling dependencies from Maven Central together with their own transitive dependencies simply exposes more of the version conflicts across our dependency stack. I would vote for reviewing them later although this might have to be done on a case-to-case basis.

I realized this PR actually has a version bump for `xml-apis` since `commons-pool 1.2` requires version `2.0.2`. I can exclude this transitive dependency to keep versions consistent if this is sensible.

joshmoore:
> I realized this PR actually has a version bump for xml-apis since commons-pool 1.2 requires version 2.0.2. I can exclude this transitive dependency to keep versions consistent if this is sensible.

Probably best. (I now sense a "dependency diff" job coming on...)

sbesson:
Pushed.

joshmoore:
Certainly no mention of xml-apis 2.x in https://ci.openmicroscopy.org/job/OMERO-DEV-merge-build/ICE=3.5,jdk=8_LATEST,label=octopus/ws/src/dist/share/server/dependencies.html#commons-pool-commons-pool

Also, https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration/167/ is generally green. Assuming this is good to go.

PR #4347, opened 2015-11-20
Gateway followup

dominikl:
PR for adding new functionality to the Gateway

- Added easier methods for creating Datasets
- Added easier methods for adding Annotations
- Moved the 'get orphaned images' methods from Insight to the Gateway
- Added a Facility for handling OMERO.tables
- Added methods for uploading images plane-by-plane

**Test**:
- Check that the orphaned images folder in Insight still works.
- The other new methods aren't used by Insight, yet (but covered by integration tests). So just code review and check the integration tests don't fail.


dominikl:
Comment about the ```TablesFacility``` commit  af1c662 : I added a class ```TableData``` for holding the data, just saw afterwards that there already is a similar class ```TableResult```, but that seems to be quite measurement table specific. So I'm not sure, keep both, try to use ```TableResult``` for the ```TablesFacility```, or the other way round and use ```TableData``` for the measurement data?

dominikl:
--exclude until #4344 is merged

jburel:
@dominikl: status? 
Looking to have the PR (or some commits) for 5.2.2

dominikl:
Yes, I think some of it should definitely go into 5.2.2 (methods moved to the gateway, etc.). But some commits I added, are just ideas I had (also with respect to the R gateway, making it easier to access data from there). Not sure if these are methods which will be generally useful (they work and are tested by integration tests, however). I'd like to have them in 5.2.2. What are your thoughts? 
Especially the TablesFacility commit a0cfab27af70 duplicates some things which are already partially present in Insight, but I found it really hard to work with. It also provides methods with will be usuful for R. A crucial point from the R discussion was the ability to work with table structures in Omero.

sbesson:
Heads up part of this PR is conflicting with the `regions` work on `ShapeData` /cc @mtbc. Is there a concrete need for having the new ROI Pojos method as of 5.2.2 or is this something  we should consider driving as part of the breaking ROI work?

jburel:
ROI should probably be included in the ROI work.

dominikl:
Alright, I remove the ROI commit from the PR. Can cherry-pick later again, if needed.

dominikl:
Removed the ROI commit and added another one dealing with importing images (would be useful for the "import image" training example). If there aren't any other objections at the moment, I could add the PR to the standup PR review?

dominikl:
--exclude will conflict with #4423

dominikl:
Added a method to "stream" image to the server (upload plane-by-plane), see 7349b1f  /cc @bramalingam

dominikl:
--exclude I have to exclude this PR for now, don't know how to fix the integration test, works fine locally but not on CI.

dominikl:
Thanks @joshmoore ,you're right, that's a good lead.

dominikl:
Found the problem. I used the image id, not the pixels id! Guess it always worked in my case locally, because running the test against an empty DB, the pixels id == the image id.

joshmoore:
Ah! Good catch. I still think the non-context-based cache could also become a problem, but happy for that to be an RFE.

components/tools/OmeroJava/test/integration/gateway/TransferFacilityTest.java
```diff
@@ -113,4 +169,39 @@ private void initData() throws Exception {
...
+                            it, data);
+                }
+            }
+        }
+
+        transferFacility.closeImage(rootCtx, img.getId());
+    }
+
+    private byte[] createRandomPlaneData(int x, int y) {
+        int size = x * y;
```
mtbc: For this kind of thing http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/math/IntMath.html#checkedMultiply%28int,%20int%29 can be safer. (Java 8 brings something like that into the standard library but that's for another day.)

dominikl: Thanks, yes, good to know. I always forget about the possibility of an int overflow.

components/blitz/src/omero/gateway/facility/BrowseFacility.java
```diff
@@ -240,6 +260,29 @@ public IObject findIObject(SecurityContext ctx, String klassName, long id)
...
+     * @param id
+     *            The object's id.
+     * @return The last version of the object.
+     * @throws DSOutOfServiceException
+     *             If the connection is broken, or not logged in
+     * @throws DSAccessException
+     *             If an error occurred while trying to retrieve data from OMERO
+     *             service.
+     */
+    public DataObject findObject(SecurityContext ctx, String pojoName, long id)
```
joshmoore: Would a generic version of this be easier to use in general, or would that be too constraining based on what the client knows at a given point? e.g. `public <T extends DataObject> T findObject(Class<T> pojoType, ...)` would be an option.

The same holds for `findIObject`.

components/blitz/src/omero/gateway/facility/BrowseFacility.java
```diff
@@ -269,6 +315,50 @@ public IObject findIObject(SecurityContext ctx, String klassName, long id,
...
+     *            Pass <code>true</code> to look for all groups
+     * @return The last version of the object.
+     * @throws DSOutOfServiceException
+     *             If the connection is broken, or not logged in
+     * @throws DSAccessException
+     *             If an error occurred while trying to retrieve data from OMERO
+     *             service.
+     */
+    public DataObject findObject(SecurityContext ctx, String pojoName, long id,
+            boolean allGroups) throws DSOutOfServiceException,
```
joshmoore: All with the growing number of methods, would it be possible to speceify an options object (and/or interface)?

components/blitz/src/omero/gateway/facility/BrowseFacility.java
```diff
@@ -603,12 +723,15 @@ public IObject findIObject(SecurityContext ctx, IObject o)
...
+     * @throws DSOutOfServiceException
+     *             If the connection is broken, or not logged in
+     * @throws DSAccessException
+     *             If an error occurred while trying to retrieve data from OMERO
+     *             service.
      */
     public Collection<DatasetData> getDatasets(SecurityContext ctx,
             long ownerId, Collection<Long> ids) throws DSOutOfServiceException, DSAccessException {
-        if (ids == null)
+        if (ids == null || ids.isEmpty() || ownerId < 0)
```
joshmoore: What does an `ownerId` less than zero mean? Is that explained somewhere?

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
```diff
@@ -497,4 +572,242 @@ public void addImagesToDataset(SecurityContext ctx,
...
+                            rawFileStore.close();
+                        } catch (ServerError e) {
+                        }
+                    }
+                }
+            }
+        };
+
+        if (callback != null) {
+            (new Thread(r)).start();
```
joshmoore: I'd urge against using `new Thread` anywhere that it's not absolutely necessary. If that means we need to look into a `ThreadingFacility` or a `BackgroundService` etc. :+1:

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
```diff
@@ -497,4 +572,242 @@ public void addImagesToDataset(SecurityContext ctx,
...
+                            originalFile);
+
+                    rawFileStore = gateway.getRawFileService(ctx);
+                    rawFileStore.setFileId(originalFile.getId().getValue());
+                    FileInputStream stream = new FileInputStream(file);
+                    long pos = 0;
+                    int rlen;
+                    byte[] buf = new byte[INC];
+                    ByteBuffer bbuf;
+                    while (((rlen = stream.read(buf)) > 0) && !cb.isCancelled()) {
```
joshmoore: Much of this code exists in `omero.client`, though it doesn't include the callback. If the addition of a callback to `omero.client.upload` would make this reusable, that might well be worth it.

components/blitz/src/omero/gateway/facility/Callback.java
```diff
@@ -0,0 +1,87 @@
+package omero.gateway.facility;
+
+/**
+ * A callback handle for asynchronous method calls
```
mtbc: might want the usual copyright boilerplate

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
```diff
@@ -497,4 +572,242 @@ public void addImagesToDataset(SecurityContext ctx,
...
+     *             If the connection is broken, or not logged in
+     * @throws DSAccessException
+     *             If an error occurred while trying to retrieve data from OMERO
+     *             service.
+     */
+    public <T extends AnnotationData> T attachAnnotation(SecurityContext ctx,
+            T annotation, DataObject target) throws DSOutOfServiceException,
+            DSAccessException {
+        if (target != null) {
+            if (target instanceof ProjectData) {
```
joshmoore: As a longer-term RFE, it'd be nice if we could reduce the use of `instanceof` in favor of interface-based methods on the `*Data` objects themselves.

components/blitz/src/omero/gateway/facility/TablesFacility.java
```diff
@@ -0,0 +1,526 @@
...
+            }
+            return new TableData(header, descriptions, types, dataArray);
+        } catch (Exception e) {
+            handleException(this, e, "Could not load table data");
+        }
+        return null;
+    }
+
+    /**
+     * Get all available tables for a the specified object
```
joshmoore: "a the" typo

components/blitz/src/omero/gateway/facility/TablesFacility.java
```diff
@@ -0,0 +1,526 @@
...
+     * @param parent
+     *            The {@link DataObject}
+     * @return See above
+     * @throws DSOutOfServiceException
+     *             If the connection is broken, or not logged in
+     * @throws DSAccessException
+     *             If an error occurred while trying to retrieve data from OMERO
+     *             service.
+     */
+    public Collection<FileData> getAvailableTables(SecurityContext ctx,
```
joshmoore: @manics recently adding something similar to the CLI -- it might be time to move that query to the server.

components/blitz/src/omero/gateway/facility/TransferFacility.java
```diff
@@ -49,6 +61,12 @@
 
     /** Reference to the {@link DataManagerFacility} */
     private DataManagerFacility datamanager;
+    
+    /** Reference to the {@link BrowseFacility} */
+    private BrowseFacility browser;
+
+    /** References to currently open pixel stores; key is the pixels id*/
+    private static Map<Long, RawPixelsStorePrx> pixelStores = new ConcurrentHashMap<Long, RawPixelsStorePrx>();
```
joshmoore: When does this get cleaned?

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
```diff
@@ -497,4 +572,242 @@ public void addImagesToDataset(SecurityContext ctx,
...
+            final File file, String mimetype, final String description,
+            final String namespace, final DataObject target, Callback callback) {
+        final String name = file.getName();
+        String absolutePath = file.getAbsolutePath();
+        final String path = absolutePath.substring(0, absolutePath.length()
+                - name.length());
+
+        final String mime;
+        if (mimetype == null)
+            mime = "application/octet-stream";
```
mtbc: Might be nice to use http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#probeContentType%28java.nio.file.Path%29.

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,175 @@
...
+     * 
+     * @param objects
+     *            The array to generate the object for
+     * @param types
+     *            The types of the elements in the array; it's assumed that
+     *            every element of a sub array (<code>object[i][]</code>) has
+     *            the same type (<code>types[i]</code>).
+     * @return See above
+     */
+    private int objectArrayHashCode(Object[][] objects, Class[] types) {
```
joshmoore: @mtbc : do you know of an existing implementation of this?

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
```diff
@@ -497,4 +572,242 @@ public void addImagesToDataset(SecurityContext ctx,
...
+            public void run() {
+                RawFileStorePrx rawFileStore = null;
+                try {
+                    OriginalFile originalFile = new OriginalFileI();
+                    originalFile.setName(omero.rtypes.rstring(name));
+                    originalFile.setPath(omero.rtypes.rstring(path));
+                    originalFile.setSize(omero.rtypes.rlong(file.length()));
+                    final ChecksumAlgorithm checksumAlgorithm = new ChecksumAlgorithmI();
+                    checksumAlgorithm.setValue(omero.rtypes
+                            .rstring(ChecksumAlgorithmSHA1160.value));
```
mtbc: I don't think anybody will mind this choice but now you have me wondering if http://downloads.openmicroscopy.org/omero/5.2.0/api/slice2html/omero/grid/ManagedRepository.html#suggestChecksumAlgorithm should have been on RFS instead!

components/blitz/src/omero/gateway/model/TableData.java
mtbc: I'm afraid that I'm not aware of anything with the desired equality property.

components/blitz/src/omero/gateway/facility/MetadataFacility.java
```diff
@@ -108,4 +123,119 @@ public ImageAcquisitionData getImageAcquisitionData(SecurityContext ctx,
...
+        return result.get(object);
+    }
+
+    /**
+     * Get the annotations for the given {@link DataObject}s
+     * 
+     * @param ctx
+     *            The {@link SecurityContext}
+     * @param objects
+     *            The {@link DataObject}s to load the annotations for
```
mtbc: Is it required that they are all of the same type?

components/blitz/src/omero/gateway/facility/TablesFacility.java
manics: See `components/tools/OmeroPy/src/omero/plugins/metadata.py` in #4426

components/blitz/src/omero/gateway/facility/TablesFacility.java
```diff
@@ -0,0 +1,526 @@
...
+import omero.model.ImageI;
+import omero.model.OriginalFile;
+import omero.model.OriginalFileI;
+import omero.model.Plate;
+import omero.model.PlateI;
+import omero.model.Roi;
+import omero.model.RoiI;
+import omero.model.WellSample;
+import omero.model.WellSampleI;
+import edu.emory.mathcs.backport.java.util.Arrays;
```
mtbc: :grimacing: Maybe we need a git commit hook!

components/blitz/src/omero/gateway/facility/TablesFacility.java
```diff
@@ -0,0 +1,526 @@
...
+     *             service.
+     */
+    public void addTable(SecurityContext ctx, DataObject target, String name,
+            TableData data) throws DSOutOfServiceException, DSAccessException {
+        try {
+            if (name == null)
+                name = UUID.randomUUID().toString();
+
+            String[] columnNames = data.getColumnNames() != null ? data
+                    .getColumnNames() : new String[0];
```
mtbc: https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/ArrayUtils.html#EMPTY_STRING_ARRAY should save you the new instance allocation. Again, more an FYI than a blocker.

components/blitz/src/omero/gateway/facility/TablesFacility.java
mtbc: "a the" needs editing

components/blitz/src/omero/gateway/facility/TablesFacility.java
```diff
@@ -0,0 +1,526 @@
...
+        }
+
+        if (type.equals(Double[].class)) {
+            double[][] d = new double[data.length][];
+            int l = 0;
+            for (int i = 0; i < data.length; i++) {
+                Double[] src = (Double[]) data[i];
+                double[] dst = new double[src.length];
+                for (int j = 0; j < src.length; j++)
+                    dst[j] = src[j];
```
mtbc: Apache Commons has tons of useful array type stuff like https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/ArrayUtils.html#toPrimitive%28java.lang.Double[]%29

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + Arrays.hashCode(columnNames);
+        result = prime * result + Arrays.hashCode(types);
+        result = prime * result + objectArrayHashCode(data, types);
+        result = prime * result + Arrays.hashCode(descriptions);
+        return result;
```
mtbc: http://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#hash%28java.lang.Object...%29 may help for this kind of thing. You could even use `TableData.class` as the first argument to try to keep the hash particular to instances of this class. Again, FYI, not a blocker.

components/tools/OmeroJava/test/integration/gateway/TransferFacilityTest.java
```diff
@@ -77,6 +82,57 @@ public void testUploadImage() throws DSAccessException,
...
+        DatasetData targetDS = new DatasetData();
+        targetDS.setName("Test Dataset");
+        cb = new ImportCallback();
+        transferFacility.uploadImage(rootCtx, testFile, true, null, targetDS, null, cb);
+        
+        c = 0;
+        while (!cb.isFinished() && c < 10) {
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
```
mtbc: not quite sure why we're swallowing these instead of letting the test throw

components/blitz/src/omero/gateway/facility/TablesFacility.java
dominikl: Ah, I really hate this, Eclipse auto-import always picks this weird class instead of java.util.Arrays, if I don't watch very carefully.

components/blitz/src/omero/gateway/facility/BrowseFacility.java
dominikl: There's a generic version, see line 204 and 229 (if that's what you meant).

components/blitz/src/omero/gateway/facility/BrowseFacility.java
dominikl: You mean for example one method `getImages(SecurityContext ctx, Options options)` instead of the - at the moment - 7 different `getImages()` methods? That would make sense/simplify the API. Should have done that from the start; now it means I'd have to mark most of the methods as deprecated. That'd be a change for the next round anyway, too much for the 5.2.2.

components/blitz/src/omero/gateway/facility/BrowseFacility.java
dominikl: Actually the current implementation doesn't make sense. `<0` as ownerId should mean the ownerId is ignored and the result should be the same like the according method without ownerId paramater is called, in that case `getDatasets(SecurityContext ctx, Collection<Long> ids)`. Returning an empty collection is probably not what one would expect. I'll fix that. Will also remove some code duplication.

components/blitz/src/omero/gateway/facility/TransferFacility.java
dominikl: If the RawPixelsStores are not closed by calling the `closeImage()` method, they will be closed when the Gateway is disconnected, https://github.com/dominikl/openmicroscopy/blob/gateway_followup/components/blitz/src/omero/gateway/Gateway.java#L217 (the Connectors themselves hold references to open services, too).

components/blitz/src/omero/gateway/model/TableData.java
dominikl: Ah that's interesting, thanks. I used and modified the auto-generated one from eclipse. Using `hash(Object... values)` would have been easier.

joshmoore:
NB: @dominikl -- sorry, I missed the `depends-on` in gh-4475 which means that I had to remove the import line from `TransferFacility` in gh-4477. This will therefore soon conflict. Merging in origin/develop once 4477 is merged should solve it.

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
dominikl: I'm not sure how I can access this service. It's a bit strange that you have to provide a list of checksum algorithms for the method, and not just get the suggested checksum algorithm straight away. Might address this in a followup PR, when I found out how to access it.

mtbc: You just tell it what you can support. There's no point it suggesting an algorithm you don't have handy! (The idea is to let changes in the set of supported algorithms not be a breaking change.) The import library solves this by https://github.com/openmicroscopy/openmicroscopy/blob/037f0caf/components/blitz/src/ome/formats/importer/ImportLibrary.java#L153. (Definitely not a blocker for this PR.)

dominikl:
As discussed with @jburel and @sbesson I'm going split this PR into a "Gateway 5.2.2" part and an "experimental" part (which will be excluded for 5.2.2).

dominikl:
Closing in favor for #4483 .
Thanks for reviewing @joshmoore  and @mtbc , I added your suggestions to the new 5.2.2 PR #4483

PR #4348, opened 2015-11-21
Fix duplicate screen target

simleo:
This PR addresses a problem encountered when importing `.screen` files.

 * *Without* this patch, if the user specifies a screen target, plates are linked to that screen *and* to the screen specified by Bio-Formats' `ScreenReader`
 * *With* this patch, the target specified by the user overrides the BF one

**NOTES**

 * It would be nice to simply skip `addReference` in `setScreenPlateRef`, but at that point in the import process `userSpecifiedTarget` has not been set yet (on the server's side)
 * `screenKey` should probably be changed to a list, or removed altogether if scanning the whole `referenceCache` for screen-plate refs is not too intensive
 * BF adds a screen -> plate ref, while `TargetProcessor` [adds a plate -> screen one](https://github.com/openmicroscopy/openmicroscopy/blob/983abca738b10367fe20ba2653ed99d3971e74ed/components/blitz/src/ome/formats/model/TargetProcessor.java#L86). Does this mean one of the two has to be fixed?

### DIFF
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -273,6 +273,8 @@ public class OMEROMetadataStoreClient
      */
     private Map<String, String[]> referenceStringCache;
 
+    private LSID screenKey;
+
     /** Our model processors. Will be called on saveToDB(). */
     private List<ModelProcessor> modelProcessors;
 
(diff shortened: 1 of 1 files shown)


joshmoore:
http://idr-demo.openmicroscopy.org/mito all in!

joshmoore:
Note: before moving to the mainline we'll need to define tests based on our expectations. i.e. will it ever make sense to import one plate into multiple screens?

joshmoore:
--rebased-to #4429

PR #4353, opened 2015-11-25
Fixed integer overflow

dominikl:
Fixes  [Ticket 13109](http://trac.openmicroscopy.org/ome/ticket/13109)

Test: Select a large image ( > 47k x 47k), check that the export options (JPG, PNG, etc.) are all disabled.

Fortunately it could only cause Insight to crash with an OOM exception, no spacecraft disasters...


gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.0-208-5600258-ice35-b167-mac eel user-3

Behaves as expected - could export ID: 21546 (1612x4668) - export options disabled for ID: 21547 (78656x106560).
Good to merge.

PR #4355, opened 2015-11-26
Remove HDF5_DIR overriding in python_deps.sh

sbesson:
This export command currently makes the dependency installation script unusable for PyTables in any other location than the ones under which the HDF libraries are accessible (e.g. /usr/local for Homebrew). Reported by @pwalczysko during testing. 

Without this change, on OSX with Homebrew installed packages, running the following command

```
virtualenv /tmp/venv
source /tmp/ven/bin/activate
bash path/to/python_deps.sh
```

from any other location than `/usr/local` should fail. With this change, the Python packages installation should complete and find the HDF libraries under `/usr/local`.

### DIFF
--- a/docs/install/python_deps.sh
+++ b/docs/install/python_deps.sh
@@ -56,7 +56,6 @@ export CFLAGS="-I/usr/X11/include -I/usr/X11/include/freetype2 -I/usr/X11/includ
 install matplotlib
 
 # PyTables requirements ===========================================
-export HDF5_DIR=`pwd`
 install Cython
 install numexpr
 install tables
--- a/docs/install/python_deps.sh
+++ b/docs/install/python_deps.sh
@@ -58,7 +58,7 @@ install matplotlib
 # PyTables requirements ===========================================
 install Cython
 install numexpr
-install tables
+install tables==2.4
 
 echo "Done."
 


pwalczysko:
Turns out that too recent version of pytables was being downloaded. @sbesson will push a commit where the version of the tables will be fixed, and this shoiuld enable the repetition of the bug and checking of the fix in this PR.

pwalczysko:
I still think I have no chance to repeat the bug. So afaik the situation is as follows:
  - there is a fix for the script in place in this PR, which causes that it will work whether the tables are capped to 2.4 or not (main cause of inability to repeat the bug)
  - the workflow might be: uninstall tables, then try to install it back, then go to venv, then run the script from where ? (if it is run after lets say master branch, then it will not have the last commit either, and it will again succeed, because the version is not capped. If it is run from this PR branch, then it will succeed, because it has the cap AND the fix for it as well)

sbesson:
... of course, this PR contains both the way to reproduce the bug and the fix. Should I split into 2 PRs ?

manics:
Are you capping tables to 2.4 just to reproduce this bug?

pwalczysko:
I am afraid, if I just checkout the branch of this PR, it does not work either. Following the scenario in the comment of this PR I get.

```
Collecting tables==2.4
  Downloading tables-2.4.0.tar.gz (8.9MB)
    100% |████████████████████████████████| 8.9MB 61kB/s 
    Complete output from command python setup.py egg_info:
    .. ERROR:: You need numpy 1.4.1 or greater to run PyTables!
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/j3/9sf00sjs4kng7djn76m0_grh4d75r0/T/pip-build-Xyh1VI/tables
(venv2)[pwalczysko@ls31619 ~/Work/openmicroscopy]$ 
```

pwalczysko:
@manics : afaik NO is the answer, there was another, much more sensible reason for the capping. @sbesson ?

sbesson:
@manics: tables is capped to 2.4 to reproduce this bug and because it seems we need < 3.0 or  less.

manics:
The IDR is currently running `tables==3.1.1`

pwalczysko:
@sbesson : this sugestion https://github.com/openmicroscopy/openmicroscopy/pull/4355#issuecomment-160695951 is reasonable. But also see https://github.com/openmicroscopy/openmicroscopy/pull/4355#issuecomment-160696085 - seems to be much bigger problem.

manics:
Are you capping because there's a bug with `tables >= 3` on OS X? Or is it due to some other Python dependencies on OS X which can't be upgraded?

PR #4356, opened 2015-11-26
Bugfix for ticket 13092

dominikl:
Fixes [Ticket 13092](https://trac.openmicroscopy.org/ome/ticket/13092)

Test: Check that multi-selection (holding shift key) works for wells; check that the bug fixed by #4219 isn't re-introduced again by this PR

/cc @jburel


pwalczysko:
Shift click on wells works. There is no regression of behaviour with respect to https://github.com/openmicroscopy/openmicroscopy/pull/4219.
Ready to merge.

PR #4357, opened 2015-11-26
Webgateway api

will-moore:
I have started to port the webclient /api/ methods over to webgateway so we can discuss various location / naming / api / error handling issues.

There is already a ```webgateway/marshal.py``` module. For now, I've created a new ```webgateway/api_marshal.py``` file to keep the ported code separate from existing marshalling code, but this could just as well go all together in ```marshal.py```

We handle errors in parsing query parameters such that ```webclient/api/containers/?page=one``` returns ```HttpResponseBadRequest("Invalid value 'ome' for parameter 'page'")```

I've preserved the naming of the views methods, e.g. ```api_container_list``` and url ```api/containers/``` both with the api prefix. Maybe we don't need the extra ```api``` prefix, since everything in webgateway should be an api?

In the webclient, ```api_container_list``` takes an "id" query parameter to filter by experimenter. All the other urls use ```id``` to filter by the parent container. E.g. For ```api_images``` id is for dataset ID.
Now I have used ```user``` to filter by experimenter ID for ALL urls.

I notice that there's a lot of stuff in ```webgateway/views.py``` that could be moved out.
This should really just contain the request handling methods, with other helper methods elsewhere.
Also there's stuff here that is required by the webgateway viewer but is maybe not useful as part of the api, such as ROI/Shape thumbnails? This could all be moved to a new app just for the viewer?

It would be good to discuss these issues and develop a consensus, since we want this to be a very stable API. Meanwhile I can continue porting more code using the same pattern as above.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/api_marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/api_marshal.py
@@ -22,11 +22,22 @@
 # import time
 import omero
 
-from omero.rtypes import unwrap   # wrap   # rlong
+from omero.rtypes import unwrap, rlong   # wrap   # rlong
 from django.conf import settings
 # from django.http import Http404
 # from datetime import datetime
 from copy import deepcopy
+
+
+def build_clause(components, name='', join=','):
+    ''' Build a string from a list of components.
+        This is to simplify building where clauses in particular that
+        may optionally have zero, one or more parts
+    '''
+    if not components:
+        return ''
+
+    return ' ' + name + ' ' + (' ' + join + ' ').join(components) + ' '
 
 
 def parse_permissions_css(permissions, ownerid, conn):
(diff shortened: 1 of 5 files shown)


will-moore:
@joshmoore @chris-allan @aleksandra-tarkowska Following discussion the other day, have started to port code to webgateway. Some questions above that we could discuss on this PR, or chat?

aleksandra-tarkowska:
@sbesson @joshmoore have we already diverted develop and 5.2 ?

aleksandra-tarkowska:
@sbesson @joshmoore when we are going to divert develop and 5.2 ?

components/tools/OmeroWeb/omeroweb/webgateway/util.py
```diff
@@ -26,14 +26,20 @@
...
+
+    @param request:     http request
+    @param name:        name of parameter
+    @param default:     value to return if not found in request
+    @return:            Integer or None
+    """
+
+    index = request.GET.get(name, request.POST.get(name, default))
+    if index is not None:
+        index = int(index)
```
aleksandra-tarkowska: Why did you remove ValueError? this will simply throws ValueError if index is non digit. That is a mistake.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -1247,6 +1250,76 @@ def render_col_plot(request, iid, z, t, x, w=1, conn=None, **kwargs):
...
 @login_required()
 @jsonp
+def api_container_list(request, conn=None, **kwargs):
+    # Get parameters
+    try:
+        page = getIntOrDefault(request, 'page', 1)
+        limit = getIntOrDefault(request, 'limit', settings.PAGE)
+        group_id = getIntOrDefault(request, 'group', -1)
+        experimenter_id = getIntOrDefault(request, 'id', -1)
+    except ValueError:
```
aleksandra-tarkowska: How do you know which param is incorrect if you catch them all together without exposing that in error message?

aleksandra-tarkowska:
@will-moore could you add tests?

components/tools/OmeroWeb/omeroweb/webgateway/views.py
will-moore: Yes, I agree. This is just the current code from webclient ported over. I was wondering if there's a nice way to handle all these parameters without a ton of try / except blocks. But perhaps several try / except blocks is the clearest way of handling this.
In any case, I think it is preferable to raise an exception with invalid parameters rather than silently use the default - otherwise bug finding etc becomes much harder

will-moore:
@aleksandra-tarkowska We already have the tests in OmeroWeb/test/integration/test_tree.py which test the marshalling quite well. I have simply changed the import statement in that test to import ```marshal_projects``` from new location in webgateway.
We don't yet have Django web tests for all these api urls. We could add these but there's very little logic in the views.py methods, so these tests would only be needed for testing parameter parsing etc.

aleksandra-tarkowska:
@will-moore sorry missed tests in last file

sbesson:
@aleksandra-tarkowska: answering your question above, give the large amount of work to do notably on the CI front and the upcoming milestones, it is very unlikely the development branches split will happen before 2016.

components/tools/OmeroWeb/omeroweb/webgateway/api_marshal.py
```diff
@@ -0,0 +1,260 @@
...
+    params = omero.sys.ParametersI()
+    service_opts = deepcopy(conn.SERVICE_OPTS)
+
+    # Set the desired group context
+    if group_id is None:
+        group_id = -1
+    service_opts.setOmeroGroup(group_id)
+
+    # Paging
+    if page is not None and page > 0:
```
aleksandra-tarkowska: 'if page' is enough to test greater then 0

components/tools/OmeroWeb/omeroweb/webgateway/api_marshal.py
```diff
@@ -0,0 +1,260 @@
...
```
aleksandra-tarkowska: The same here 'if page'

will-moore:
In discussion with @aleksandra-tarkowska, we considered the option of moving all the iQuery and marshalling code down into the ```BlitzGateway```, since there is nothing web-specific about this and it could be useful to have it available to other Python users.
@chris-allan what do you think?

aleksandra-tarkowska:
Maybe with the new web API we should start python gateway2 that is mostly focused on performance?

chris-allan:
Honestly I'm a bit torn with respect to that, @will-moore.

On the one hand I agree, it's useful to expose to other OMERO Python API users. On the other, I really don't think `BlitzGateway` is a place to hold a large number of task specific queries which are designed to return a subset of data as primitives (ie. not wrapped OMERO model objects).

Things quickly devolve into adding a plethora of options to these task specific queries to handle a variety of different use cases and we're back where we started. This is not just with respect to performance but also readability and maintainability. I am also a firm believer in the use of `IQuery` outside the `BlitzGateway`. I do not think everything has to be "gatewayised".

Regarding the comments from @aleksandra-tarkowska about the use of `if variable` vs. `if variable is None` I fundamentally disagree with changing that style to reflect the former. Using that style implies a deep understanding of **all** potential possible values of `variable` and the intentions of the original code. `if variable is None` is **very** explicit about the boolean condition it is matching. What `if variable` is doing is calling `variable.__nonzero__()` to get a boolean value to compare against and then proceeding accordingly. Which of course is fine if that's what the intention of the original code is.

`if variable` is absolutely **not** sufficient to check for `> 0`:

```
In [5]: a = -1

In [6]: if a:
   ...:     print 'foo'
   ...: 
foo
```

`if variable` also has very defined and potentially unintentional consequences when dealing with empty strings, empty arrays, etc.:

```
In [7]: a = ''

In [8]: if a:
   ...:     print 'foo'
   ...:

In [9]: a = []

In [10]: if a:
   ....:     print 'foo'
   ....:
```

Unless you absolutely know all potential values you are about to apply a boolean condition to and the intentions of the original code be **very** careful about the `if variable` style. It can and will introduce subtle logic errors.

For reference:
 * https://docs.python.org/2/reference/datamodel.html#object.__nonzero__
 * https://www.python.org/dev/peps/pep-0008/#programming-recommendations

will-moore:
Thanks @chris-allan - We can certainly leave the iQuery code in webgateway for now, and move it into BlitzGateway or some similar non-web location later if needed.

will-moore:
Closing this for now, to avoid conflicts with other jsTree work.

PR #4358, opened 2015-11-27
small adjustments to server code/doc

mtbc:
This gets some small server code/doc adjustments into 5.2.1. They're from branches of larger work that will go into later versions, but we may as well avoid making releases without these small improvements in. Everything should remain green.

--no-rebase

### DIFF
--- a/components/blitz/src/omero/cmd/graphs/DuplicateI.java
+++ b/components/blitz/src/omero/cmd/graphs/DuplicateI.java
@@ -23,6 +23,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


jburel:
changes work, 
If a class uses the "old way" (object instead of generic), there will probably be only a warning.
Not sure if we have a good way to represent that case.

mtbc:
As far as I can tell, one can even write a non-generic subclass of `Work` overriding `doWork` and the code still compiles.

mtbc:
Does anything further need doing here?

joshmoore:
If `DummyExecutor` *required* updating to make this compile, then this would count as a breaking change for anyone who had subclassed `Executor`.

mtbc:
Definitely seems best not to touch `Work` so I can get this in before disappearing for the holiday, so have force-pushed that bit away. :smiley:

joshmoore:
https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration/176/ :thumbsup:

PR #4359, opened 2015-11-27
in graph traversal avoid awful removeAll performance

mtbc:
This PR should improve the performance of model graph traversal operations without any change in behavior. @joshmoore managed to run into a performance issue that this change should fix.

--no-rebase

### DIFF
--- a/components/server/src/ome/services/graphs/GraphTraversal.java
+++ b/components/server/src/ome/services/graphs/GraphTraversal.java
@@ -1504,7 +1504,7 @@ public class GraphTraversal {
         /* process the targets forward across links */
         while (!planning.blockedBy.isEmpty()) {
             /* determine which objects can be processed in this step */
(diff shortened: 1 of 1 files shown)


jburel:
Do we have a test to expose that performance issue?

mtbc:
No, I think it was related to IDR data?

joshmoore:
The specific case was:

 * `bin/omero import` a fileset (grouped by a .screen file) which registers as SPW on the client.
 * on the server, no Plate is created, so there's a fileset with thousands of files and images but no top-level container.
 * `bin/omero delete Fileset:$ID` took quite a long time.

Likely, any synthetic fileset with many, many individual files would work.

joshmoore:
--rebased-to #4374

joshmoore:
Based on testing in gh-4375, I'd say this is good to merge.

jburel:
@joshmoore @mtbc: thanks. Merging so it can be included in rc2

PR #4362, opened 2015-12-01
Ticket 12546

dominikl:
This would fix the NPE of ticket 12546, although I don't know the real cause for this (and therefore can't say how to test it). ```wm.getSelectedWell()``` is the only object which theoretically could be ```null``` in the line of code which triggered the NPE, so this PR at least provides a workaround for the issue (i. e. prevents Insight from crashing).

Test: Code review only (as I've not been able to reproduce the NPE)


jburel:
I will also in that case, add a check in ```setThumbnailsFieldsFor (line 1635)```

dominikl:
Yes, good idea, thanks.

PR #4364, opened 2015-12-01
Activities fixes

will-moore:
This fixes a couple of issues noticed as part of 5.2.0 update on nightshade, where ping of 'activities' in web was hanging, but the webclient continued to ping every 5 seconds.
Now, we only re-try to get activities 5 seconds AFTER the previous call returns successfully.
Also add a try/except around the cast of string to proxy object where it was missing for scripts.

To test, check that scripts run OK and results are returned in the activities panel.
Also try an OMERO.figure export.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
@@ -32,7 +32,7 @@
     }
 
     // this is called by the setInterval loop below
-    OME.activitiesUpdate = function(i) {
+    OME.activitiesUpdate = function(recursive) {
         $.get("{% url 'activities' %}", function(data) {
             $('#activities_spinner').hide();
             var inprogress = $("#inprogress", data).text();
@@ -41,13 +41,13 @@
             
             // if we've got no jobs still running, stop checking
             if ((typeof inprogress == 'undefined') || (inprogress.length == 0)) {
-                if (i) clearInterval(i);
+                recursive = false;
                 showJobCount(0);
                 //return;
             }
             inprogress = parseInt( inprogress );
-            if (inprogress==0) {
-                if (i != undefined) clearInterval(i);
+            if (inprogress === 0) {
+                recursive = false;
             }
             
             // display what we recieved, bind events etc
@@ -77,10 +77,14 @@
                 new_results = -1;   // hide notification number
             }
             OME.displayStatus(new_results, inprogress);
+            if (recursive) {
+                setTimeout(function(){
+                    OME.activitiesUpdate(true);
+                }, 5000);
+            }
 
         }).error(function() {
             // this requires jQuery 1.5 or later
-            clearInterval(i);
         });
     }
 
@@ -120,10 +124,7 @@
     }
 
     OME.refreshActivities = function() {
-        var i = setInterval(function (){
-                OME.activitiesUpdate(i);
-        }, 5000);
-        OME.activitiesUpdate(i);
+        OME.activitiesUpdate(true);
     }
     $(document).ready(function() {
         OME.displayStatus(-1);  // reset new_results
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
@@ -85,6 +85,8 @@
 
         }).error(function() {
             // this requires jQuery 1.5 or later
+            $('#activities_spinner').hide();
+            $("#activities_content").html("<table><tbody><tr><td><span>Activity unavailable - timeout</span></td></tr></tbody></table>");
         });
     }
 
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
@@ -86,7 +86,7 @@
         }).error(function() {
             // this requires jQuery 1.5 or later
             $('#activities_spinner').hide();
-            $("#activities_content").html("<table><tbody><tr><td><span>Activity unavailable - timeout</span></td></tr></tbody></table>");
+            $("#activities_content").html("<table id='activitiesTimeoutMsg'><tbody><tr><td><span>Activity unavailable - timeout</span></td></tr></tbody></table>");
         });
     }
 
@@ -152,6 +152,8 @@
         $("#clear_activities").click(function() {
             $.post("{% url 'activities_update' action='clean' %}");
             $("#jobsTable>tbody>tr").filter(":not(.in_progress)").remove();
+            // In case we have shown timeout message...
+            $("#activitiesTimeoutMsg").remove();
         });
 
         // close panels on 'ESCAPE'
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
@@ -33,60 +33,65 @@
 
     // this is called by the setInterval loop below
     OME.activitiesUpdate = function(recursive) {
-        $.get("{% url 'activities' %}", function(data) {
-            $('#activities_spinner').hide();
-            var inprogress = $("#inprogress", data).text();
-            var new_results = $("#new_results", data).text();
-            var failure = $("#failure", data).text();
-            
-            // if we've got no jobs still running, stop checking
-            if ((typeof inprogress == 'undefined') || (inprogress.length == 0)) {
-                recursive = false;
-                showJobCount(0);
-                //return;
-            }
-            inprogress = parseInt( inprogress );
-            if (inprogress === 0) {
-                recursive = false;
-            }
-            
-            // display what we recieved, bind events etc
-            $("#activities_content").empty();
-            $("#activities_content").append( $("#jobsTable", $(data)) );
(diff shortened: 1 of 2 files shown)


aleksandra-tarkowska:
@will-moore thanks for fixing activity. My only comment would be to better handle timeout on activity request. Basically if ``webclient/activities/`` times out just display ``Activity not available``  instead of ``No Activities``. This is because people may not know there is something wrong if they click delete only.

firefoxscreensnapz014

components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -3610,9 +3610,19 @@ def activities(request, conn=None, **kwargs):
...
             if status not in ("failed", "finished"):
                 logger.info("Check callback on script: %s" % cbString)
-                proc = omero.grid.ScriptProcessPrx.checkedCast(
-                    conn.c.ic.stringToProxy(cbString))
-                cb = omero.scripts.ProcessCallbackI(conn.c, proc)
+                try:
+                    proc = omero.grid.ScriptProcessPrx.checkedCast(
+                        conn.c.ic.stringToProxy(cbString))
+                    cb = omero.scripts.ProcessCallbackI(conn.c, proc)
+                except Exception, x:
```
aleksandra-tarkowska: Could you please use the same logic as in delete and wrap the entire code in try/catch block?

```
try:
    proc = omero.grid.ScriptProcessPrx.checkedCast(
    everything...
except Ice.ObjectNotExistException:
except Exception, x:
```

aleksandra-tarkowska: and please make sure ``cb.close()`` is called in final

will-moore:
@aleksandra-tarkowska Looking at the try except logic in activities seems to be horribly complex and unclear what we are trying to achieve. As far as I can tell, with the various nested try finally blocks, it's not even guaranteed that cb.close() will be called for delete or chgrp, depending on where the exception is generated.
I would prefer that each try block contains the minimum amount of code and we try not to nest them.
Also, we do
```
cb = omero.callbacks.CmdCallbackI(conn.c, handle, foreground_poll=True)
```
for delete and e-mail, but not for chgrp or scripts and it would be nice to know if we need this or not?

will-moore:
@joshmoore Can you remind me the difference between:
```
prx = omero.cmd.HandlePrx.checkedCast(conn.c.ic.stringToProxy(cbString))
rsp = prx.getResponse()
```
and
```
prx = omero.cmd.HandlePrx.checkedCast(conn.c.ic.stringToProxy(cbString))
callback = omero.callbacks.CmdCallbackI(conn.c, prx, foreground_poll=True)
rsp = callback.getResponse()
```
If we don't actually want to wait, then can we just go for the first option?

will-moore:
@aleksandra-tarkowska I've refactored the activities() try except blocks to try and clean it up a bit.
However, it turns out that scripts handles things differently, so it doesn't use the same method as the others. I've left this as it is.
It also seems that you can't call cb.close() while a script is running, since this causes an error.
I haven't yet tested that e-mail is still working, but chgrp and delete are OK.

joshmoore:
Kicked travis (HDF5 error)

aleksandra-tarkowska:
@will-moore great, that was really needed. I will check email for you

aleksandra-tarkowska:
@will-moore timout indeed appeared in the activity, that is great, although clean button doesn't work. Would it make sense to fix it as well?

will-moore:
@aleksandra-tarkowska Clear List button should also remove timeout warning now.

jburel:
@will-moore/@aleksandra-tarkowska : did that get re-tested?

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
```diff
@@ -77,10 +77,16 @@
                 new_results = -1;   // hide notification number
             }
             OME.displayStatus(new_results, inprogress);
+            if (recursive) {
+                setTimeout(function(){
+                    OME.activitiesUpdate(true);
+                }, 5000);
+            }
 
         }).error(function() {
```
aleksandra-tarkowska: Here you need to check status of the response otherwise this display timeout even there were 500

aleksandra-tarkowska:
With the new changes unfortunately email notification failed with

```
VariableDoesNotExist: Failed lookup for key [rsp] in u"{'status': 'finished', 'start_time': datetime.datetime(2015, 12, 14, 15, 59, 31, 172261), 'job_type': 'send_email', 'key': '22c8d9cd-d25e-4221-83f2-3ac2eaa61c6e/IHandle430161c2-f536-4984-8764-c39f7a735f71 -t -e 1.0:tcp -h eel.openmicroscopy.org -p 58354', 'error': 0, 'id': 'IHandle430161c2-f536-4984-8764-c39f7a735f71'}"
```

aleksandra-tarkowska:
cleaning up activity doesn't work as expected timed out job remain ``in progress`` and there is no way at this moment of forcing to delete the job. This can be only resolved once we handle jobs via activity directly and check status individually. Probably not important at this moment.

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
aleksandra-tarkowska: $.ajax({
  statusCode: {
    500: function(xhr) {
      ...
    }
  }
});

aleksandra-tarkowska: Also https://api.jquery.com/jquery.get/ The jqXHR.success(), jqXHR.error(), and jqXHR.complete() callback methods introduced in jQuery 1.5 are deprecated as of jQuery 1.8 could you start using new one please?

pwalczysko:
Perfectly happy with what @aleksandra-tarkowska concludes here re: testing this. Sorry, did not have a margin for this today as busy with https://github.com/ome/omero-install/pull/71#issuecomment-164509022. 
Do you think this is ready to merge @aleksandra-tarkowska ?

aleksandra-tarkowska:
I'm afraid there is a bug in the email notification. See the comment above. I also didn't test the failure in the scripts and chgrp.

pwalczysko:
@aleksandra-tarkowska : Great, thanks a lot. I can confirm your observation about email notifications, it fails in UI with "Activities timeout". This is fully working on eel latest => clear regression.

Tested also as you suggested chgrp, the OMERO.figure and scripts (all on Centos7) - these all passed the test.

This means - unless we do not desperately need this PR, maybe exclude from the consideration for 5.2.1 ?

aleksandra-tarkowska:
@pwalczysko did you tested failing chgrp, scripts, etc?

aleksandra-tarkowska:
This PR is needed because activity is flooding with unnecessary requests.

aleksandra-tarkowska:
This PR is needed because activity is flooding with unnecessary requests if anything failed. Although I think it could be fixed in 5.2.2

pwalczysko:
@aleksandra-tarkowska : Ah, sorry, you mean to test **failing** chgrp actions as opposed to passing ones ? Yes, I just had a failed chgrp which ended with a valid message in Activities and an error output as expected 
```
{'error': 'stacktrace: ome.services.graphs.GraphException(message=may not move Image[21939] while WellSample[14048] remains)\n\tat ome.services.graphs.GraphPolicyRule$CleanGraphPolicy.recordChanges(GraphPolicyRule.java:1110)\n\tat ome.services.graphs.GraphPolicyRule$CleanGraphPolicy.reviewWithSingleMatch(GraphPolicyRule.java:951)\n\tat ome.services.graphs.GraphPolicyRule$CleanGraphPolicy.review(GraphPolicyRule.java:891)\n\tat ome.services.graphs.GraphTraversal.reviewObject(GraphTraversal.java:1106)\n\tat 
```
So this looks good for chgrp, both failing and passing.

will-moore:
@pwalczysko I think it's OK to exclude this from 5.2.1.
It is a relatively minor update to prevent repeated requests to activities which we noticed when these requests were completely hanging.
Anyway, I will look at the e-mail error...

pwalczysko:
@will-moore : Would you have a suggestion for a failing script workflow ? The chgrp was easy, but for example Kymograph, although there is no line in the image, is actually reported as Success in Activities.
Will leave it for tomorrow.

jburel:
Marking it on hold so it is clearer.
Thanks

pwalczysko:
Workflow to test here:
   - select a large plate and move it to another group, e.g.
   - https://eel.openmicroscopy.org/merge/webclient/?show=plate-1153
   - check that the Activities show the action as expected

will-moore:
@pwalczysko Just to clarify - bug described in your comment above is from eel today, which doesn't have this PR (since Travis was failing until I closed and re-opened the PR today).
Don't know if that behaviour is caused by other open PRs?

pwalczysko:
@will-moore : Yes, it; is possible that this is caused by some other open PR. 
In any case this PR here 
  - does cause change in the behaviour of this bug in such a manner that it started to be repeatable and the Activities always report  "No Activities"
  - this PR does not fix the behaviour (as comes from the previous bullet point)

In the next step, I will try the workflow on eel latest.

pwalczysko:
Further to the problem https://github.com/openmicroscopy/openmicroscopy/pull/4364#issuecomment-169680384 : 
  - the faulty Activities behaviour is not demonstrating all the time the same, but it does consistently come up with the following setup
  - the plate ID 1153 of user-3 (96 well plate) moving between the RA and RO groups by the owner of the data back and forth
  - being on eel merge with this PR open or with this PR closed (both cases have the bug)
  - on eel latest, the bug does not occur
  - with smaller plate (16 well) the bug cannot be repeated in any setup -> large plate specific

PR #4365, opened 2015-12-01
Remove 'tilt' from stats pie chart

will-moore:
See http://trac.openmicroscopy.org/ome/ticket/13042

As suggested by @mtbc, just removed the 'tilt' attribute from the pie chart plugin.
To test, login as admin and look at Statistics tab for all users, and/or group owner and look at statistics for your owned groups.
In both cases, pie chart should be round, not an ellipse:

screen shot 2015-12-01 at 14 01 00


gusferguson:
@will-moore 

Behaves as expected in Admin Stats and Group Owner Stats all views.
Good to merge.

PR #4366, opened 2015-12-01
Rendering settings warning

dominikl:
Proposal for fixing [Ticket 13104](http://trac.openmicroscopy.org/ome/ticket/13104)

With this PR a warning dialog is only shown when all images of a dataset are affected by a rendering settings change action:
- Save to All
- Paste rendering settings on a dataset
- Reset to owner on dataset level
- Reset to imported on dataset level

But *not* if only one or a selection of images is affected, i. e. deliberately pasting a rendering setting on a selected image (in that case the user is well aware what he's doing - in my opinion). 

This is not in sync with OMERO.web yet. But as I just noticed, the current state is also not in sync with OMERO.web anyway (e. g. web doesn't show a warning on "save to all").

/cc @gusferguson @pwalczysko

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/PojosUtil.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/PojosUtil.java
@@ -1,0 +1,55 @@
+/*
+ *------------------------------------------------------------------------------
+ *  Copyright (C) 2015 University of Dundee. All rights reserved.
+ *
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *  
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ *------------------------------------------------------------------------------
+ */
+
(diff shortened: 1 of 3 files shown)


gusferguson:
@dominikl 

Tested with OMERO.insight-5.2.0-250-398e1b8-ice35-b172-win on eel user-3, user-10

Behaves as expected.
I think this interaction is good and the warnings or lack of them are appropriate.
Good to merge.

dominikl:
Thanks @gusferguson . Forgot to cc @will-moore , the warning dialog in web also needs to be adjusted (at the moment "save to all" doesn't show the warning).

pwalczysko:
I think the positions of the dialogs per se is good, as @gusferguson says. But compare the text of the warning in Web:
screen shot 2015-12-02 at 11 15 57

How about adding the "no undo" sentence as well ?
Agree that in web there should be the warning about ``Save to All`` - often people ask what ``All`` means at the demos.

jburel:
Why do you limit it to a dataset? since the functionality is available for other containers

pwalczysko:
``...will change rendering settings of all the images in the ..../screen...`` -> not precise, replace with ``...will change rendering settings .../plate``

pwalczysko:

components/insight/SRC/org/openmicroscopy/shoola/util/PojosUtil.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Checks if a given class is a container class, i. e. can contain multiple
+     * images
+     * 
+     * @param type
+     *            The class to check
+     * @return See above.
+     */
+    public static boolean isContainerClass(Class<?> type) {
```
joshmoore: Perhaps another RFE: Is it time to start introducing marker classes into the gateway: `obj instanceof ContainerType`?

joshmoore: Note: discussion of such a marker interface shouldn't block this PR.

dominikl:
That definitely would make sense, added to the [Trello card - Gateway improvements](https://trello.com/c/0hvmrXSL/10-gateway-improvements)

PR #4368, opened 2015-12-02
updating django requirements on Windows

aleksandra-tarkowska:
This fix pip requirements file for Windows, as by default there is Python 2.7 and increase Django version to 1.8 LTS.

cc @dominikl


jburel:
is < 1.9 not enough? or we don't install django at all if < 1.8.
Maybe add a comment

aleksandra-tarkowska:
@jburel no it is not enough, we explicitly need 1.8 only

jburel:
@aleksandra-tarkowska: 
```This is a work in progress, filenames may change in the next version```.
 Do you have other name in mind? or we are now happy with it
In that case, we may want to remove that line.

aleksandra-tarkowska:
This is standard comment across all files, @sbesson ?

sbesson:
@aleksandra-tarkowska @jburel: this comment was put into place as part of OMERO 5.2.0 as we did not know whether the `requirements-pyxx-yy.txt` strategy would scale with the variety of deployments. With the large work performed on deployment for OMERO 5.2.1, we might want to either drop these lines and commit to stable naming pattern or keep them flexible if this might be subject to change..

jburel:
We now have the same naming scheme established
We can revisit if required in 5.3 so I will be more of the opinion to remove the "work in progress"

jburel:
Discussed today. Merging this PR so we can move forward with doc
cc @hflynn 
@aleksandra-tarkowska will open a follow-up PR to remove the "work in progress" section

PR #4370, opened 2015-12-02
log parameters and results of graph requests

mtbc:
Adjust ` etc/logback.xml` such that `omero.cmd.graphs` is logged at `DEBUG`, then try some graph requests (delete, move, whatever you like) and see that the request parameters and the response values are logged in `Blitz-0.log`.

--no-rebase

### DIFF
--- a/components/blitz/src/omero/cmd/graphs/Chgrp2I.java
+++ b/components/blitz/src/omero/cmd/graphs/Chgrp2I.java
@@ -66,6 +66,8 @@ import omero.cmd.Response;
  */
 public class Chgrp2I extends Chgrp2 implements IRequest, WrappableRequest<Chgrp2> {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(Chgrp2I.class);
+
     private static final ImmutableMap<String, String> ALL_GROUPS_CONTEXT = ImmutableMap.of(Login.OMERO_GROUP, "-1");
 
     private static final Set<GraphPolicy.Ability> REQUIRED_ABILITIES = ImmutableSet.of(GraphPolicy.Ability.OWN);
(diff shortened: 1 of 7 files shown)


ximenesuk:
This looks good, I can see requests and response logged for simple commands and skiphead commands. (Very useful too!). Good to merge.

sbesson:
Thanks @mtbc, merging this. Follow-up question is whether the `omero.cmd.*` requests/responses should be logged at the `INFO` level like other server methods?

mtbc:
Not sure. There *is* the usual logging from `HandleI`, etc., available already, so you already see a bit of what happened, the steps going by, etc. I suppose the question is, if that's enough. Would want to avoid the full parameter values at INFO as some of them may include tens of thousands of integers.

joshmoore:
Yeah, 10K integers (longs!) sounds bad. What `ServiceHandler` does is to elide anything beyond 3 entries. I've missed the old-service-style logging, so I guess my main question is whether or not we could get `omero.cmd.*` logging to a place where in the general case, we're happy gleaming information from the log *without* needing to restort to `s/DEBUG/INFO/`

mtbc:
It certainly might be worth a card if you have an idea what you'd like to see. Anything that helps us understand what's going on when the server does something undesirable is welcome!

joshmoore:
@mtbc : coming back to this, thoughts on using a method like https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.1/components/server/src/ome/services/util/ServiceHandler.java#L344 for reducing the output

mtbc:
@joshmoore: I definitely want to be able to get to see the full output. But perhaps it could truncate the long lists only if debug level isn't TRACE, would that make sense?

PR #4371, opened 2015-12-03
Correct the stats_info_creation SQL action

jballanc:
This changes the `stats_info_creation` SQL action from using a `select`
subclause to the more appropriate `values (...)`.


joshmoore:
Number of locations making use of this and all jobs today were green. Thanks, @jballanc

PR #4374, opened 2015-12-04
in graph traversal avoid awful removeAll performance (rebased onto metadata)

joshmoore:
This is the same as gh-4359 but rebased onto metadata.

----

This PR should improve the performance of model graph traversal operations without any change in behavior. @joshmoore managed to run into a performance issue that this change should fix.

--no-rebase


joshmoore:
--rebased-from #4359

joshmoore:
Uh, that didn't work!

PR #4379, opened 2015-12-04
Force configuration rewrite in bin/omero admin stop if no configuration exists

sbesson:
If the server was freshly build and downloaded and `bin/omero admin stop` is called, the command will always return "Server not running" independently of whether a server is running or not.

With this change, the configuration files should be regenerated before attempting to stop the server if `--force-rewrite` is passed or there is no configuration file found.

/cc @joshmoore


joshmoore:
Think I've convinced myself that it should be:

```
+        internal_cfg = self._cfglist()[0]
+        if not os.path.exists(internal_cfg):
+            self.ctx.die(45, "Unconfigured! Please run with --force-rewrite is this is what you intended")
```

or similar to handle the case:

 * User downloads 5.x.0 and configures 2 servers, one default and one port `--prefix=1`
 * ...both in production...
 * User downloads 5.x.1 and tries to stop the *second* with `bin/omero admin stop`
 * The *first* is stopped.

sbesson:
The proposal is certainly sensible. Thoughts on other commands that should have a default `die_on_missing` behavior ? `admin status`? `admin waitup` ? `admin waitdown`?

joshmoore:
Perhaps everything that takes `--force-rewrite`. i.e.:

 * new dir; no arg: fails
 * new dir; `force-rewrite`: runs with new config
 * old dir; no arg: runs with old config
 * old dir; `force-rewrite`: runs with new config

sbesson:
Makes sense, except for the case of `admin start` where we don't want a failure. Will work through an exhaustive list of all commands and try to update tomorrow.

joshmoore:
As with other CLI "public API" PRs, these decisions are probably too big for this late in 5.2.1. Marking ON_HOLD. Thanks regardless.

sbesson:
Coming back to this PR, we have really 2 classes of commands here:
- `admin status` is the lowest common denominator to all high-level command
- `admin start`, `admin stop` (and their async variants skipping `waitup`, `waitdown`)

Focusing on the `start/stop` commands, their current behavior is as follows

CLI admin subcommand | config exists | --force-rewrite | behavior
-------------------|--------------|-------------------|------------
start | no | no | checks the server status (always not running), generates the configuration, starts the server
start | no | yes | generates the configuration, checks the server status, starts the server
start  | yes | no | checks the server status, generates the configuration, starts the server 
start  | yes | yes | generates the configuration, checks the server status, starts the server
stop | no | no | checks the server status (always not running)
stop | no | yes |  generates the configuration, checks the server status, stops the server
stop | yes | no | checks the server status, stops the server
stop | yes | yes | generates the configuration, checks the server status, stops the server

My  proposal would be the following to raise ambiguity:
- implement the configuration check proposed by @joshmoore in AdminControl.status and effectively fail the `admin status` subcommand or any subcommand consuming it if the configuration files are not existing
- maintain 82da72e to always regenerate the configuration in `admin start` prior to checking the status

The resulting behavior should be the following

CLI admin subcommand | config exists | --force-rewrite | behavior
-------------------|--------------|-------------------|------------
start | no | no | generates config, checks the server status, starts the server
start | no | yes | generates config, checks the server status, starts the server
start  | yes | no | checks the server status, generates the configuration, starts the server 
start  | yes | yes | generates the configuration, checks the server status, starts the server
stop | no | no | fails
stop | no | yes |  generates the configuration, checks the server status, stops the server
stop | yes | no | checks the server status, stops the server
stop | yes | yes | generates the configuration, checks the server status, stops the server

joshmoore:
Looks quite impressive.

manics:
There's also `omero node` to consider https://github.com/openmicroscopy/openmicroscopy/pull/4422

sbesson:
Pushed. Also added/fixed unit tests to cover the semantics above.

manics:
```
omero@090d305005d7:~/omero-other$ O/bin/omero admin status
Missing internal configuration. Run bin/omero admin rewrite or pass --force-rewrite.
omero@090d305005d7:~/omero-other$ O/bin/omero admin status --force-rewrite
usage: O/bin/omero admin status [-h] [--nodeonly] [node]
O/bin/omero admin status: error: unrecognized arguments: --force-rewrite
```

sbesson:
Good catch. `admin status` does not support `--force-rewrite` and I don't think it is sensible for such a low-level function. I will have a go at rewriting the error message. Thoughts about something like

```
Missing internal configuration. Run bin/omero admin rewrite or pass the --force-rewrite option if you are using the start/stop commands.
```

manics:
That would work. Alternatively either pass a flag to `status()` indicating whether it's a direct or indirect call with different error messages, or move the logic into `_status()` and have separate error messages in the calling methods.

joshmoore:
Re-opening for travis

manics:
Good to merge (unless anyone objects to the last merge commit 19ac9db)

PR #4385, opened 2015-12-07
render.py: new CLI plugin for dealing with rendering settings

joshmoore:

```
```


joshmoore:
This is a work-in-progress but comments on the CLI API are very welcome. cc @ximenesuk @jburel @sbesson @eleanorwilliams @pwalczysko et al.

joshmoore:
Another semi-complete plugin being merged for 0.0.3. @ximenesuk : the methods that are implemented were used for the recent release, but some methods in the docs are not yet implemented. Likely we would just scope back the plugin and handle functions as we need them.

pwalczysko:
Can I redo also non-Plates (i.e. images in Datasets ?) - I am asking because of the recent nightshade black thumbs problem

joshmoore:
@pwalczysko : the plugin currently doesn't support datasets (IDR doesn't have any) but it will be quite simple (5-10 minutes) to add support when this PR is rebased to develop/dev_5_2.

joshmoore:
--rebased-to #4430

PR #4388, opened 2015-12-16
Workaround for Gnome3 bug

dominikl:
This is a workaround for https://bugzilla.gnome.org/show_bug.cgi?id=759492 , respectively  [OME forum post](http://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7966). I wrapped the workaround in a method in UIUtilities so it can be easily removed again when/if this gets fixed in Gnome3.

**Test**:
Start Insight on an OS with Gnome3 desktop (i. e. Debian (Jessie) default desktop). You should see the normal login screen (without the PR you'd just see a grey rectangle). Test the "Show parent project" and "Show location on server" pop up dialogs (buttons see toolbar in right hand side metadata panel). Select an image with a file attachment. Test the "Info..." dialog on the attachment. Again, without the PR these would just be grey rectangles.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/ui/UIUtilities.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/ui/UIUtilities.java
@@ -575,7 +575,7 @@ public class UIUtilities
 	{
 		centerOnScreen(window);
 		window.setVisible(true);
-        applyGnome3Workaround((Frame) window);
+        applyGnome3Workaround(window);
 	}
     
 	/**
@@ -2952,22 +2952,22 @@ public class UIUtilities
      * (making it one pixel bigger); a simple call to repaint() is not
      * sufficient.
      * 
-     * @param w
-     *            The {@link Window}
+     * @param c
+     *            The {@link Component}
      */
-    public static void applyGnome3Workaround(Window w) {
+    public static void applyGnome3Workaround(Component c) {
         if (!GNOME)
             return;
 
         boolean undecorated = false;
-        if (w instanceof Frame)
-            undecorated = ((Frame) w).isUndecorated();
-        if (w instanceof Dialog)
-            undecorated = ((Dialog) w).isUndecorated();
+        if (c instanceof Dialog)
+            undecorated = ((Dialog) c).isUndecorated();
+        if (c instanceof Frame)
+            undecorated = ((Frame) c).isUndecorated();
 
         if (undecorated) {
-            Dimension size = w.getSize();
-            w.setSize(new Dimension(size.width + 1, size.height + 1));
+            Dimension size = c.getSize();
+            c.setSize(new Dimension(size.width + 1, size.height + 1));
         }
     }
 }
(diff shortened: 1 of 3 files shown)


dominikl:
Probably unnecessary, they already fixed it in Gnome, will test again later with an updated Gnome.

dominikl:
According to the user reporting this bug, it will be around for a long time, until the bugfix will be pushed to the different distributions. Therefore reopened this workound PR.

components/insight/SRC/org/openmicroscopy/shoola/util/ui/UIUtilities.java
```diff
@@ -2935,4 +2937,30 @@ public static String replaceWindowsPathSeparator(String path) {
...
+     */
+    public static void applyGnome3Workaround(Window w) {
+        boolean undecorated = false;
+        if (w instanceof Frame)
+            undecorated = ((Frame) w).isUndecorated();
+        if (w instanceof Dialog)
+            undecorated = ((Dialog) w).isUndecorated();
+
+        if (undecorated) {
+            String desktop = System.getenv("XDG_CURRENT_DESKTOP");
```
joshmoore: Would you expect this value to ever be changed at runtime? If not, perhaps:

```
private final static boolean GNOME;
static {
    String desktop = System.getenv("XDG_CURRENT_DESKTOP");
    GNOME = desktop != null && desktop.toLowerCase().contains("gnome");
}
```

followed by:

```
public static void applyGnome3Workaround(Window w) {
    if (!GNOME) return;
```

dominikl:
Thanks @joshmoore . Yes, much better than repeatedly checking it.

joshmoore:
Currently testing with:
```
/tmp/gnome3 $ cat Vagrantfile
Vagrant.configure(2) do |config|
  # Ubuntu 15.04
  config.vm.box = "ubuntu/vivid64"

  config.vm.provider "virtualbox" do |vb|
    # Display the VirtualBox GUI when booting the machine
    vb.gui = true
  end

  # Install xfce and virtualbox additions
  config.vm.provision "shell", inline: "sudo apt-get update"
  config.vm.provision "shell", inline: "sudo apt-get install -y ubuntu-gnome-desktop virtualbox-guest-dkms virtualbox-guest-utils virtualbox-guest-x11"
  config.vm.provision "shell", inline: "sudo service gdm restart"
  # Permit anyone to start the GUI
  config.vm.provision "shell", inline: "sudo sed -i 's/allowed_users=.*$/allowed_users=anybody/' /etc/X11/Xwrapper.config"
end
/tmp/gnome3 $ vagrant up
```

dominikl:
Additional **test**: Perform a delete action (see [Ticket 13127](http://trac.openmicroscopy.org/ome/ticket/13127))

pwalczysko:
Tested the last commit, on Mac only. Works as expected ( i. e. problems along the lines of http://trac.openmicroscopy.org/ome/ticket/13127 are fixed on Mac platform)

dominikl:
@kennethgillen also tested it on ubuntu with gnome3.

pwalczysko:
@dominikl : Then I would think this is sufficient, thank you @kennethgillen

PR #4390, opened 2015-12-17
Elide object ids in standard CLI output

ximenesuk:
This PR adds object id elision to the basic output from CLI delete, chgrp, etc. This has been in the detailed report for some time but had been omitted from the default response. See https://trello.com/c/Y37tzyrK/25-missing-range-collapsing-on-delete

To test import and then move or delete a number of objects and check the response. 
```
$ omero delete Image:1,2,3
...
omero.cmd.Delete2 Image:1-3 ok
```
Try with broken sequences of ids too, the results should not be surprising!

The added test should pass.

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1844,11 +1844,11 @@ class GraphControl(CmdControl):
         objects = []
         for req in doall.requests:
             for type in req.targetObjects.keys():
-                ids = ",".join(map(str, req.targetObjects[type]))
+                ids = self._order_and_range_ids(req.targetObjects[type])
                 if isinstance(req, omero.cmd.SkipHead):
                     type += ("/" + req.startFrom[0])
-                objects.append('%s %s' % (type, ids))
-        return "%s %s... " % (cmd_type, ', '.join(objects))
+                objects.append('%s:%s' % (type, ids))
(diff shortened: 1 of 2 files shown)


ximenesuk:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_delete/TestDelete/testOutputWithElision/

joshmoore:
* `omero.cmd.Delete2 Dataset:186-188` ok :thumbsup:
 * `omero.cmd.Delete2 Dataset:183,184` ok was a slight surprise but :thumbsup:

Now I'm just waiting on someone to request `bin/omero delete Dataset:186-188`. :smile:

OT: 2 minor testing points that we should probably just keep in mind for the moment:
 * There's a *slight* race condition here in that another process could create an image between your existing queue
 * We might eventually want some text-comparison helpers to simplify the maintenance of the output processing.

cc: @sbesson

ximenesuk:
Regarding the race condition, yes I was aware of this and so the test faile early if the image id sequence is not continuous. Not ideal but it means if the test passes it is testing the right thing. At least for this PR I went for a simple test, it could certainly be more robust.

Text-comparison helpers would definitely make sense.

`bin/omero delete Dataset:186-188` is definitely doable, it's just deciding what it means if the permissions over that range vary.

PR #4389, opened 2015-12-17
Jstree move to ome.tree.js

will-moore:
This moves the jsTree code to a separate js file and uses the container.html template for both the main Data page as well as the Tags page.
This means we don't have to maintain a separate container_tags.html template and all the logic for jsTree for these two pages is in a single file.

To test (NB: a lot of this is covered by Robot framework now - once that is passing, we shouldn't need to test this so deeply):
 - Need to test browsing in main Data page and Tags page, comparing data with Insight
 - Test copy, cut, paste, delete on jsTree (via toolbar and right-click menu) on Data page
 - Test filtering by user and 'All Members' in both pages

PR #4392, opened 2015-12-18
Api tagged perms

will-moore:
This uses the ```select map( )``` format of iQuery when loading tagged data, as we already do with other jsTree data, in order to work-around the hibernate projection permissions issues.

This fixes permissions on tagged objects belonging to other users.
Also adds tests to check this.

To review:
 - Check that all tests are passing
 - Look at the queries to check they make sense (that there isn't an obviously better way than the inner ```select``` to ensure that results are distinct)?

### DIFF
--- a/components/tools/OmeroPy/test/integration/helpers.py
+++ b/components/tools/OmeroPy/test/integration/helpers.py
@@ -35,3 +35,26 @@ def createTestImage(session):
                                       "description", dataset=None)
 
     return image.getId().getValue()
+
+
+def createImageWithPixels(client, name, sizes={}):
+        """
+        Create a new image with pixels
+        """
+        pixelsService = client.sf.getPixelsService()
+        queryService = client.sf.getQueryService()
+
+        pixelsType = queryService.findByQuery(
+            "from PixelsType as p where p.value='int8'", None)
+        assert pixelsType is not None
+
+        sizeX = "x" in sizes and sizes["x"] or 1
+        sizeY = "y" in sizes and sizes["y"] or 1
+        sizeZ = "z" in sizes and sizes["z"] or 1
+        sizeT = "t" in sizes and sizes["t"] or 1
+        sizeC = "c" in sizes and sizes["c"] or 1
+        channelList = range(1, sizeC+1)
+        id = pixelsService.createImage(
+            sizeX, sizeY, sizeZ, sizeT, channelList, pixelsType,
+            name, description=None)
+        return id
--- a/components/tools/OmeroPy/test/integration/test_iquery.py
+++ b/components/tools/OmeroPy/test/integration/test_iquery.py
@@ -111,7 +111,7 @@ class TestQuery(lib.ITest):
             rootLinks.append(link)
         rootUpdate.saveAndReturnArray(rootLinks, {'omero.group': str(groupId)})
 
-        q = """select new map(obj.id as id,
+        q = """select distinct new map(obj.id as id,
                obj.name as name,
                obj.details.owner.id as ownerId,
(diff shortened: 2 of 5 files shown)


jburel:
Did you test the performance with the extra select in the map?

will-moore:
OK, I have tagged about 2000 images with a single tag (in group 5), and have added print statements to time the query and iterating through the query.

With this query:
```
select new map(obj.id as id,
               obj.name as name,
               obj.details.owner.id as ownerId,
               obj as image_details_permissions,
               obj.fileset.id as filesetId 
             ,
             pix.sizeX as sizeX,
             pix.sizeY as sizeY,
             pix.sizeZ as sizeZ
             ,
            obj.details.creationEvent.time as date,
            obj.acquisitionDate as acqDate
            )
            from Image obj  left outer join obj.pixels pix
            join obj.annotationLinks alink
            where alink.id = (select max(alink.id)
                from ImageAnnotationLink alink
                where alink.child.id=:tid and alink.parent.id=obj.id)
        
                    order by lower(obj.name), obj.id
```

For a tag with 2000 images this takes approximately ```4 seconds```. 

Interestingly, I just tried this without specifying the group context, E.g. (group is -1 instead of 5) and the exact same query takes about ```1.3 seconds```! Any idea why this might be @joshmoore?

Now, if I remove the second select statement, replace it with
```
where alink.child.id=:tid
```
the time goes down to ```1.2 seconds``` (still with group -1).

If I use group 5 again, without the second select statement, I get about the same ```1.2 seconds```.

So, only a small amount slower (less that 10% increase) with the second select statement.

Strange that use of ```group 5``` has such a big effect on the timing when combined with the extra select statement.
Maybe we could simply ignore the group filter in the tree.py query if we're querying by tag ID?
At least for now I'll remove it from the javascript AJAX query builder.

will-moore:
@joshmoore In general, is it better to keep group context at ```-1``` unless we actually need to filter by group? Currently jsTree is adding group context to every query, even though it doesn't need it.
I think this should be fixed, but not until #4389 is merged, since it would conflict.

joshmoore:
Personally, this sounds like a bug somewhere. I also wouldn't expect for `group:-1` to be faster.

will-moore:
Tried to reproduce the bug in a test, but it mostly passes (```group:-1``` is slower).

will-moore:
So it seems that testing query performance in web today does not show the slow-down 'bug' with ```group: 5``` that I was seeing above.
I'll fix up the test (so it's not so fragile dependent on query time).

will-moore:
I've removed the test, since it's not really valid.

will-moore:
@joshmoore Any suggestions for how to ensure unique query results without the second select statement? Otherwise I guess we have to accept the 10% performance penalty?

joshmoore:
@will-moore : I'm slightly lost. Could you point me to a failing test (i.e. something where non-unique results are returned but expected) that I could attempt to fix?

will-moore:
@joshmoore That commit contains a failing test where non-unique results are returned as expected. A "fix" would just be a nicer (more performant) way to ensure unique results without the need for a second select statement.

joshmoore:
@will-moore: could you try https://github.com/joshmoore/openmicroscopy/commit/d37772ca5acb2f2afc7f28df612a4178c1294746 ?

will-moore:
So, tests should all be passing again. We now use a simper query in tree.py using ```distinct``` instead of a second ```select``` statement.
Check that browsing tags in web shows same data as Insight.

dominikl:
Looks fine. Tests passed. Couldn't spot any difference between Insight and Web browsing through tags.

PR #4395, opened 2016-01-05
adjust log level for probing nested simple properties

mtbc:
RFE documented at https://trello.com/c/XyOslw9x/24-reduce-warn-in-graphpathbean. Check the `Blitz-0.log` after the server starts up for a lack of the mentioned warnings. One needn't do anything with a client.

--no-rebase


ximenesuk:
Log looks correct, good to merge.

PR #4399, opened 2016-01-06
Fixed ticket 13081

dominikl:
Fixed the SelectionDialog, see [Ticket 13081](https://trac.openmicroscopy.org/ome/ticket/13081).
Test: Check that the SelectionDialog works like expected by adding and removing tags for single and multiselection, especially check the workflow mentioned in the ticket.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/util/SelectionWizardUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/util/SelectionWizardUI.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -843,12 +843,10 @@ public class SelectionWizardUI
                 ho = node.getUserObject();
                 if (ho instanceof DataObject) {
                     data = (DataObject) ho;
-                    if (!isImmutable(data)) {
-                        if (data.getId() >= 0 && !isChild(node)) {
-                            availableItems.add(node);
-                        }
-                        toRemove.add(node);
+                    if (data.getId() >= 0 && !isChild(node)) {
+                        availableItems.add(node);
                     }
+                    toRemove.add(node);
                 } else {
                     toRemove.add(node);
                 }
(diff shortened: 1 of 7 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/util/SelectionWizardUI.java
```diff
@@ -843,12 +843,10 @@ private void removeItem()
...
                     data = (DataObject) ho;
-                    if (!isImmutable(data)) {
-                        if (data.getId() >= 0 && !isChild(node)) {
-                            availableItems.add(node);
-                        }
-                        toRemove.add(node);
+                    if (data.getId() >= 0 && !isChild(node)) {
+                        availableItems.add(node);
                     }
+                    toRemove.add(node);
```
mtbc: Why not just bring this out of the `if` and remove the `else` altogether?

dominikl:
Thanks @mtbc , yes, I missed that.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.1-131-144e64a-ice35-b200-mac user-3 read-annotate-1

Issue described in ticket is resolved, but there are some other problems.
- add tag Test 1 to 2 images - works fine
- select those 2 image + 2 more without any tags - add tag - in tag selection window Test 1 appears in left side list - select Test 1 and click right arrow - Test 1 appears in right side list (screenshot 1)
- click Save - Test 1 successfully linked to all 4 images as expected (screenshot 2)
- select 4 images with Test 1 linked + 2 images without any tags linked - add tag - in tag selection window Test 1 and Test 2 appears in left side list - select Test 1 and Test 2 - click right arrow - Test 1 and Test 2 appear in right side list (screenshot 3)
- click Save - Test 2 successfully linked to all 6 images (screenshot 4), but Test 1 is not linked to the 2 images that did not have it linked before - they only have Test 2 linked (screenshot 5)

Holds for own images and other user’s images.

Up to you whether you want to tackle this in this PR or merge one this as a fix for the ticket and open a new PR for the other issue.

ss1

ss2

ss3

ss4

ss5

dominikl:
Thanks for testing all these combinations, that's apparently one I missed to check. I think I'd make sense to fix that within this PR.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.1-138-f2274ec-ice35-b202-mac user-3 read-annotate-1

There seems to be a new problem that is preventing testing of this PR.

First:

Test images - some with tags attached - selected all - tagged and untagged - in right hand pane - clicked on drop-down -Remove tags.

Crashed  with error message in ticket: https://trac.openmicroscopy.org/ome/ticket/13132#ticket

Note - removal of tags with multiselected images only fails with this error when one or more of the images does not have the tag on it.

Second:

Select 2 images - add tag Test 1 - behaves as expected
Select these 2 images + 2 more with no tags - add tag Test 2 - crashes on clicking Save in tag window.

Crashed with error message in ticket: https://trac.openmicroscopy.org/ome/ticket/13133#ticket

Filed second ticket as there were some differences further down in the error messages - is probably the same issue.

dominikl:
That's really horrible to fix... I'll exclude this PR for now from the merge build.

--exclude

dominikl:
Rebased it onto #4405 because of merge conflicts.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.1-155-cc65d52-ice35-b205-mac eel user-3 read-annotate-1

No crashes.

Still issue with adding tag:
- add tag Test 1to 2 images
- select those 2 images + 2 more with no tags
- add tags - select tags Test 1 and Test 2 and add
- tag Test 2 is added to all 4 images - Test 1 is not added to the second 2 images.

dominikl:
I have to give up on this. The problem is, what the tagging dialog does is, it adds the tags to right hand side tags panel. And from this panel it's worked out which tags have to be added or removed. When multiple objects are selected (i. e. tags panel shows all tags of all objects) and you add another tag via tagging dialog, it's just impossible to find out which tags have to added to which objects. 
This is general problem of the annotation/multiselection handling in Insight, not specific for this PR. So I'm not sure how to proceed, close this PR or leave it open for getting merged? It fixes the problem within the tag selection dialog (see ticket), but tagging still doesn't work properly for multiselection.

dominikl:
Closing this PR. Ticket 13081 can't be solved in the terms of the SelectionWizard (i. e. it shouldn't be used for multi-selection in the first place). 
(Note: The really important bug fixes with respect to the metadata panel are also in PR #4405 )

PR #4400, opened 2016-01-06
improve graph operation performance (rebased from develop)

mtbc:
With thanks to @joshmoore, adjust how graph traversal uses Hibernate to determine the exact class of model objects.

--rebased-from #4398 to `metadata` to assist testing

### DIFF
--- a/components/server/src/ome/services/graphs/GraphTraversal.java
+++ b/components/server/src/ome/services/graphs/GraphTraversal.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014-2015 University of Dundee & Open Microscopy Environment.
+ * Copyright (C) 2014-2016 University of Dundee & Open Microscopy Environment.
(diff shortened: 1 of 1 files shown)


mtbc:
Do we expect travis to pass on the `metadata` branch?

jburel:
flake8 issue in the tests, that should be fixed so we have no problem when it is ported to the main line.

joshmoore:
Happy to have someone try, but if it were only flake8, I'm pretty sure I would have gone ahead and done it. If there isn't an immediate fix, it might be simpler to wait on the rebasing which will be ongoing throughout January.

mtbc:
Now #4398 is merged feel free to close or merge this as desired.

joshmoore:
I think I'll merge this to `metadata`, but won't rebase to `metadata52` since it will be based on 5.2.2

PR #4402, opened 2016-01-08
Example for image import

dominikl:
Adds the import example to the Java training examples, goes along with docs PR https://github.com/openmicroscopy/ome-documentation/pull/1377

**Test**: Check that `OMERO-DEV-merge-training` job passes.


jburel:
The second example assumes that there is a dataset.
Will it possible to have the 2 examples to match?
or at least indicate the difference.

dominikl:
Removed the dataset specification, actually just commented it out, because I think it's a useful information anyway. Would have been better to change the Gateway example to import into a particular dataset, but just noticed that the TransferFacility is still lacking an appropriate method for doing this.

dominikl:
I actually wonder why the merge-training job doesn't fail. Run locally, I get a lot of exceptions, like
```
java.lang.Exception: The image must have at least 2 channels.
	at training.CreateImage.CreateNewImage(CreateImage.java:148)
	at training.CreateImage.<init>(CreateImage.java:255)
	at training.Setup.<init>(Setup.java:124)
	at training.Setup.main(Setup.java:143)
```
But for some reason it seems to work on the CI.

jburel:
Now that we have a constructor in ```LoginCredentials``` I am wondering if we should not modify the main method and no longer pass ```null```
This will apply to other examples too.

jburel:
@dominikl: what do you think?

dominikl:
Yes, that'll make the code more compact. I'll change the examples within this PR.

dominikl:
Just to get sure, you mean using the 'command line arguments' constructor ```LoginCredentials(new String[]{"omero.user=root", omero.pass=...", ...})``` throughout the examples, right?

jburel:
@dominikl: yes. sorry I should have been clearer

jburel:
https://ci.openmicroscopy.org/view/OMERO-DEV/job/OMERO-DEV-merge-training/218/ is green

Changes make sense.
doc PR needs to be adjusted.

PR #4405, opened 2016-01-11
Metadata panel bugfixes

dominikl:
The permissions weren't checked properly on the right hand side metadata panel (for Tags, Attachments and Rating), which leads to actions being enabled although the user is not allowed to perform the action (which throws an exception on attempt to execute the action), see [Ticket 13130](http://trac.openmicroscopy.org/ome/ticket/13130).

**Test**: 
- Perform a couple of metadata actions (adding/removing tags, attachments and ratings) within a read-only group on your own data. Make sure you can't perform these actions (add/remove button, rating panel should be disabled) for other users' data in the read-only group.
- Make sure filtering by "added by me", "added by others" works
- Test adding and removing tags with multiselection, while also making sure, same tags aren't shown multiple times.
- Add a key-value pair and directly add a comment as well. Make sure both modifications are stored.
- Make sure Key-Value pairs metadata is not shown for multiselection.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/TagsTaskPaneUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/TagsTaskPaneUI.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2015 University of Dundee. All rights reserved.
+ *  Copyright (C) 2015-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/TagsTaskPaneUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/TagsTaskPaneUI.java
@@ -144,6 +144,9 @@ public class TagsTaskPaneUI extends AnnotationTaskPaneUI {
             l = model.getAllTags();
         
         layoutTags(l);
+        
+        addTagsButton.setEnabled(model.canAddAnnotationLink());
+        removeTagsButton.setEnabled(model.canAddAnnotationLink());
     }
     
     /**
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/TagsTaskPaneUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/TagsTaskPaneUI.java
@@ -96,6 +96,9 @@ public class TagsTaskPaneUI extends AnnotationTaskPaneUI {
         removeAll();
         tagsDocList.clear();
         DocComponent doc;
+        
+        filter(list);
+        
         if (list != null && list.size() > 0) {
             Iterator i = list.iterator();
             DataObject data;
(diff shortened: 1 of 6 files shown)


pwalczysko:
Test from the header of the PR works as expected.
But:
  - login as user-2 to eel merge
  - create some Key Value pairs on your own Project in read-only group, as well as add Tags, comment, file attachment, rating
  - logout and login as user-4 in read-only put user-5 data on display -> you cannot see the Key Value pair of user-5 which you just created, neither the Rating is persuading (no filled in stars visible) -> not expected
  - the ``Show all``, ``Show added by me`` menu -> added by others  does hide Attachments which WERE added by others -> not expected.

Happy if the problem(s) are deemed for fix in a separate PR.

dominikl:
There are still a couple of problems with the metadata panel in general. I'll use this PR to fix them, not only the permissions problem.

components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/TagsTaskPaneUI.java
```diff
@@ -219,91 +280,18 @@ void handleObjectsSelection(Class<?> type, Collection objects, boolean fire)
...
-                if (m.size() < count) {
-                    l.add(annotation);
-                }
-            }
-        }
-        
-        return l;
+        List<AnnotationData> result = new ArrayList<AnnotationData>();
+        for(TagAnnotationData tag : toAdd)
+            result.add(tag);
```
mtbc: just use `new ArrayList<AnnotationData>(toAdd)`?

dominikl: Thanks @mtbc , good point.

pwalczysko:
All seems fine, just cannot repeat the ticket https://trac.openmicroscopy.org/ome/ticket/13135. Will continue tomorrow.

dominikl:
Thanks @pwalczysko . I think for ticket 13135 it's important to tag lots of images at once, so that the tagging action is running for  while. And while it's still ongoing, you have to select the images again (new selection). Then the tag panel is often "flooded" by the same tags.

pwalczysko:
Even after a retry today we were not able with @dominikl to repeat the bug 13135 on a release-version insight.  This seems also fine.
I have found though one other problem concerning multi-tagging of large number of images. Whereas tagging of 300 images with 2 tags (600 tags alltogether) takes 2 sec, the removal of all tags on all 300 images takes minutes.... Is this expected ? 

Possibly the last tagging problem is not something you would want to do in this PR. 
If so, this is ready to merge.

pwalczysko:
After discussion, maybe @ximenesuk would like to have a look at the tagging problems as well (he was able to repeat the original 13135 bug)

jburel:
@dominikl: will you have time to look at the issue before code freeze?

dominikl:
I don't think there's an issue left. @pwalczysko just had trouble to reproduce the original bug with the latest build, but @ximenesuk was able to replicate it. It probably just has to be tested again, with latest build to see the bug and with this PR. I'll list it again on standup.
The issue that the deletion of the tags takes quite a long time, is more a gateway/server related problem, probably better to deal with that in a separate PR.

dominikl:
**Test** for https://trac.openmicroscopy.org/ome/ticket/13135 :
- Try to replicate bug with latest build. Select dataset with several hundred of images. Select them all. Add some tags. While the tagging action is still going on, change the selection. Then again select all images. **Add another tag.** (this seems to be the crucial point) Right hand side metadata panel is now "flooded" by the same tags.
- Test workflow again with merge build.

pwalczysko:
- replicated the bug https://trac.openmicroscopy.org/ome/ticket/13135 on eel latest 
screen shot 2016-02-11 at 11 18 45

- the bug is fixed on the merge build

Ready to merge

jburel:
Thanks all

PR #4407, opened 2016-01-13
travis: Build with Ice from downloads.openmicroscopy.org

rleigh-dundee:
The original PR was closed.  This is the travis-specific part of the PR, so it will build with Ice 3.5 from downloads.openmicroscopy.org.

--------

Testing: check travis is green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ sudo: false
 cache:
   directories:
   - $HOME/.m2
+  - download
 
 matrix:
       fast_finish: true
--- a/components/tools/travis-build
+++ b/components/tools/travis-build
@@ -19,8 +19,28 @@ clean()
     ant clean
 }
 
+get_ice()
+{
+  if [ "$TRAVIS" = "true" ]; then
+    ice=Ice-3.5.1-b1-ubuntu1204-amd64.tar.xz
+    if [ ! -f "download/$ice" ]; then
+      mkdir -p download
+      cd download
+      wget "http://downloads.openmicroscopy.org/ice/experimental/$ice"
+      cd ..
+    fi
+    tar -xvf "download/$ice"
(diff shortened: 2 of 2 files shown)


joshmoore:
Since this file "may be used by hand as well" (which at least I do), can we detect `TRAVIS` in `get_ice` and exit if not?

rleigh-dundee:
@joshmoore Should be taken care of now.  Using TRAVIS=true (from https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables), it's now only triggered if this is set.

components/tools/OmeroPy/scripts
```diff
@@ -1 +1 @@
-Subproject commit 7288f3362beb5a76778e5cd41dc6ef6e0e47e4f7
+Subproject commit 4d96b4f98fcfb88281a9cf0a3c61f9da386d32b8
```
joshmoore: revert?

joshmoore:
> Should be taken care of now. 

ta.

components/tools/OmeroPy/scripts
rleigh-dundee: Done.

rleigh-dundee:
@joshmoore Fixed the rebased submodule SNAFU.  It's now green, and good to go I think.

joshmoore:
Thanks, @rleigh-dundee -- a few points:

 * testing `travis-build py-test` locally: :+1:
 * testing `TRAVIS=true py-test` locally: :+1:
 * testing `travis-build` (no arg) found an unrelated bug -- you ok to cherry-pick https://github.com/joshmoore/openmicroscopy/commit/df8d6028e7737111ce09d86a2e2f55fde733a834 ?

rleigh-dundee:
@joshmoore Picked and pushed your fix.

joshmoore:
Merging on green.

PR #4412, opened 2016-01-15
Polygon python training example

will-moore:
Improve ROIs examples in python with new Polygon example and setting of strokeWidth and stroke + fill colours. See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-January/005804.html

To test, review and run the example.


examples/Training/python/ROIs.py
```diff
@@ -79,21 +96,33 @@
...
-r = updateService.saveAndReturnObject(roi)
-
+createROI(image, [point])
+
+
+def pointsToString(points):
+    """ Returns strange format supported by Insight """
+    points = ["%s,%s" % (p[0], p[1]) for p in points]
+    csv = ", ".join(points)
+    return "points[%s] points1[%s] points2[%s]" % (csv, csv, csv)
```
mtbc: If you need to set it to something other than what's described at http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2015-01/ROI_xsd.html#Polygon_Points then we probably need a card.

joshmoore: @mtbc : I think this is the older format. @jburel / @dominikl : is the support for the newer available / carded?

joshmoore:
Objections to my adding a mask example here?

will-moore:
Please do! (in another PR you mean)?

joshmoore:
Yup, I'll build on this commit but re-use the `ROIs.py` file unless we think there should be a `Masks.py` one.

sbesson:
I would reuse `ROIs.py` minimally to stay consistent with the other training languages (https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.1/examples/Training/java/src/training/ROIs.java, https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.1/examples/Training/matlab/ROIs.m) where `ROIs.ext` is the generic script demonstrating how to create available shapes.

will-moore:
Comment from http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-January/005807.html
"My particular use-case required that the colours set on ROIs include the 
alpha-channel (which I created) so as not to completely hide the segment 
begin the ROI. You might also want to include that in rgbToRGBInt()."

will-moore:
The previous commit added support for Alpha, but if I try to set alpha > 127 (50%) for strokeColor then I get this error:

```
  File "ROIs.py", line 128, in <module>
    createROI(image, [polygon])
  File "ROIs.py", line 56, in createROI
    updateService.saveObject(roi)
  File "/Users/wmoore/Desktop/OMERO/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 4205, in __call__
    return self.handle_exception(e, *args, **kwargs)
  File "/Users/wmoore/Desktop/OMERO/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 4202, in __call__
    return self.f(*args, **kwargs)
  File "/Users/wmoore/Desktop/OMERO/openmicroscopy/dist/lib/python/omero_api_IUpdate_ice.py", line 107, in saveObject
    return _M_omero.api.IUpdate._op_saveObject.invoke(self, ((obj, ), _ctx))
ValueError: invalid value for ::omero::RInt member `_val'
```

Any ideas what I'm doing wrong here?

mtbc:
http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2015-01/ROI_xsd.html#Shape_FillColor and its kin specify a *signed* 32 bit value: if alpha > 127 then do you need to subtract to make the most significant bit act like a sign bit? Just as my example from http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-January/005803.html had #C4C4C4C4 represented by -993737532.

will-moore:
Hmmm - I wonder if anyone knows straight off how to convert from unsigned 32 bit to signed? Don't really want to spend a day trying to research 2's complement (or whatever) and work it out in python. Do we have code to do this in Java anywhere?

mtbc:
If the number is larger than 2^31-1 then subtract 2^32 from it.

will-moore:
Thanks @mtbc. Script is working now with range of alpha values.

screen shot 2016-01-21 at 11 52 22

joshmoore:
Makes sense to me and works for me locally. Merging and I'll 1) remark on the forum post and 2) update by PR to not include your commits.

PR #4413, opened 2016-01-15
React centre panel

will-moore:
Evaluating React.js for updating centre panel

TODO: :construction:

 - [x] focus(): Use scrollIntoViewIfNeeded() instead.
 - [ ] sorting. Need to implement sorting in jsTree so that thumbs stay in same order as tree, since otherwise everything breaks on pagination etc and we might be able to use jsTree selection logic instead of new logic just for thumbnails.
 - [x] Selecting a range of filtered thumbnails. Possible to select a range of thumbnails where some in the range are hidden (filtered). They should not be selected.
 - [x] Drag to select (jQuery plugin) - need to manually update selected thumbnails to jsTree.
 - [x] Select plate - hide top Dataset toolbar (and clear filter).
 - [x] Click handling for Wells
 - [x] Double-click handling for Dataset thumbnails and Wells.
 - [x] Refresh thumbnails (?thumbVersion etc)
 - [x] Pagination
 - [x] Handle /?show=well-123 and /?show=image-456 (image in well)
 - [x] Webpack etc
 - [x] Urls to use proper prefix
 - [x] Robot framework tests!
 - [ ] Tests for new imageCount in views.py
 - [x] Remove old code, checking all functionality is included in this PR: ``` center_plugin.thumbs.js, plate.html, ome.plateview.js, containers_icon.html, container_paging.html```
 - [ ] propTypes
 - [ ] code complexity: More components, smaller functions
 - [ ] shouldComponentUpdate() to improve rendering speed, especially on click/selection change


will-moore:
Close for now, until ready to test etc.

joshmoore:
One question having seen this pass by, @will-moore, would be if it wouldn't be possible to have this as a plugin.

will-moore:
So you could disable it? (like you can with the Acquisition tab and Preview tabs on the right panel)?
Yes, that's certainly possible, but is that question related specifically to this PR, or to the existing centre panel behaviour?

joshmoore:
Well, mostly re: the need to immediately close this. If it were a plugin, then it could just live wherever.

jburel:
and having a plugin for example in web test. It could be easily evaluated using docker
since we now install figure/webtagging/gallery and webtest

will-moore:
I only closed it because it's not ready yet. I just wanted to open a PR to create a PR link I could include in my presentation and I couldn't immediately add a "broken" label to it, so I've closed it while I continue to work on it (or not). The purpose of this PR was primarily the evaluation of React.js, which we'll discuss later. But this is not optional extra behaviour (like webtagging). It is a direct replacement of existing functionality.

joshmoore:
> It is a direct replacement of existing functionality.

Understood. For me, it's mostly about thinking through the concept of pluggability as far as possible. i.e. what would it cost us now or later to be able to make even this component pluggable?

PR #4414, opened 2016-01-15
Don't set Ice.Default.Host by default

manics:
`Ice.Default.Host` is useful when running multiple servers on the same host, but appears to break a multi-host OMERO.grid setup. My guess is that `Ice.Default.Host` is propagated to the slave nodes which then attempt to listen on that invalid IP (feature? bug? your guess).

This replaces `Ice.Default.Host` with a new property `omero.master.host` that allows the master's IP to be set in all templates as described in https://www.openmicroscopy.org/site/support/omero5.2/sysadmins/grid.html#nodes-on-multiple-hosts without actually setting `Ice.Default.Host`.

The single host behaviour should be identical to before, `Ice.Default.Host` is used in all templates and the server should only listen on `Ice.Default.Host`.

See
- https://trello.com/c/3fY2taKP/37-omero-grid-omero-node-bugs-issues
- https://github.com/openmicroscopy/openmicroscopy/pull/4245
- https://github.com/openmicroscopy/openmicroscopy/pull/2705

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1022,7 +1022,8 @@ present, the user will enter a console""")
             '@omero.ports.tcp@': config.get('omero.ports.tcp', '4063'),
             '@omero.ports.registry@': config.get(
                 'omero.ports.registry', '4061'),
-            '@Ice.Default.Host@': config.get('Ice.Default.Host', '127.0.0.1')
+            '@omero.master.host@': config.get('omero.master.host', config.get(
+                'Ice.Default.Host', '127.0.0.1'))
             }
 
         def copy_template(input_file, output_dir):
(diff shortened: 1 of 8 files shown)


joshmoore:
* NB: Just delete `IcePatch`? cc: @sbesson 
 * No property set:
    * `server starts`: :+1:
    * `diagnostics work`: :+1:
    * `web starts`: :+1:
    * `script launch` failed the first time, but after `stop && rm -rf var && start` it worked. Have you seen anything similar?

which all looks quite good. Happy to try to test with another host but don't think I can make it today.

manics:
- `IcePatch`- How about leave for a second PR to keep changes here minimal?
- `script launch`- haven't noticed anything unsual

joshmoore:
Works for me. Passing to @sbesson for last thoughts.

sbesson:
Looks fine to me. Merging.

PR #4415, opened 2016-01-18
Mask python training example

joshmoore:
Carry-on from:

 * gh-4412
 * http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-January/005804.html

Thanks to @emilroz

### DIFF
--- a/examples/Training/python/ROIs.py
+++ b/examples/Training/python/ROIs.py
@@ -101,6 +101,20 @@ line.theT = rint(theT)
 line.textValue = rstring("test-Line")
 createROI(image, [line])
 
+# create an ROI with a single mask
+mask = omero.model.MaskI()
+mask.setTheC(rint(0))
+mask.setTheZ(rint(0))
+mask.setTheT(rint(0))
+mask.setX(rdouble(20))
+mask.setY(rdouble(20))
+mask.setWidth(rdouble(100))
+mask.setHeight(rdouble(100))
+mask.setFillColor(rint(126))
+mask.setTextValue(rstring("test-Mask"))
+mask.setBytes([5]*10000) # WxH
+createROI(image, [mask])
+
 # create an ROI with single point shape
 point = omero.model.PointI()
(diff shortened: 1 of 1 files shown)


will-moore:
@joshmoore https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/214/console failed with
```
File "/opt/hudson/workspace/OMERO-DEV-merge-training/src/examples/Training/python/ROIs.py", line 141, in <module>
    mask_array = mask_array.tobytes()
AttributeError: 'numpy.ndarray' object has no attribute 'tobytes'
```

sbesson:
@will-moore: reading the commit history, I suspect 69c1e50e0afe7ee96de114806a1b8e2c145334f6 will fix this failure in tomorrow's build

will-moore:
This seems to be working fine.
I'd really like to try and display the mask in web (as on https://trello.com/c/H1MePy2A/124-mask-support-in-web) and wondered if @emilroz had the "reverse" code in hand to get from OMERO into a PIL image (mask) ideally. Otherwise I should be able to work something out, but didn't want to duplicate effort etc.

emilroz:
@will-moore we've have never looked into doing it the other way round (it should be rather straight forward), since the rendering engine has the ability to render the masks. In principle you can request an image with a mask on it from the server. The only problem doing it this way is the fact that it supports only the masks that match the image plane size.

will-moore:
@joshmoore I wonder if it's slightly nicer to use something like this instead of random array, since then if you're doing some subsequent work with the mask, you have some idea if it's working/looking right?
```
# Create [0, 1] mask
def f(x, y):
    return (x * y) % 2
mask_array = numpy.fromfunction(f, (mask_w, mask_h))
```

joshmoore:
Rebased and force pushed on top of 4412 and added @will-moore's non-random mask suggestion.

will-moore:
Working fine. Good to merge.

PR #4416, opened 2016-01-18
Also allow mapping of middleName via omero.ldap.user_mapping

PaulVanSchayck:
Being "blessed" with a middle name myself, this is somewhat important to me. This was tested on the active directory installation here. 

I'm unsure about the changes I made to getControls(). I wasn't sure how they are being used, or how to test them.

### DIFF
--- a/components/server/src/ome/security/auth/PersonContextMapper.java
+++ b/components/server/src/ome/security/auth/PersonContextMapper.java
@@ -39,6 +39,7 @@ public class PersonContextMapper extends OmeroModelContextMapper {
         Experimenter person = new Experimenter();
         person.setOmeName(get("omeName", ctx));
         person.setFirstName(get("firstName", ctx));
+        person.setMiddleName(get("middleName", ctx));
         person.setLastName(get("lastName", ctx));
         person.setInstitution(get("institution", ctx));
         person.setEmail(get("email", ctx));
@@ -64,6 +65,7 @@ public class PersonContextMapper extends OmeroModelContextMapper {
 
         final String inst = cfg.getUserAttribute("institution");
         final String email = cfg.getUserAttribute("email");
+        final String middleName = cfg.getUserAttribute("middleName");
         final String[] attrs = new String[]{
             "dn",
             attribute,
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks for this, @PaulVanSchayck. Can you include a change to https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.1/etc/omero.properties#L746 as well?

PaulVanSchayck:
Done, I will then also open a PR against ome-documentation

pwalczysko:
Tested locally using docker for the LDAP server and scc merge develop and locally built server. 
   - checked that I have the changes in ``omero.properties`` as indicated in the second commit https://github.com/M4I-nanoscopy/openmicroscopy/commit/8cd20dca93c613bcb8ba2b394a9109e751603e82, i.e.  ``...institution=department,middleName=middleName``

I added to the omero config manually 
``bin/omero config set omero.ldap.user_mapping "omeName=cn,firstName=givenName,lastName=sn,email=mail,middleName=displayName"``, which, when the ``displayName`` was set in LDAP server (using JXPlorer) resulted in this being taken as ``middleName`` by OMERO and displayed appropriately by Insight. 

Nevertheless, I do not see in JXPlorer any lines like ``department`` or ``middleName``. 
Not sure if this is something very generic for all LDAPs or specific for our docker LDAP server.
Maybe someone who has more experience with LDAP setups can answer this question ?

In summary:
   - commit https://github.com/M4I-nanoscopy/openmicroscopy/commit/ee9feb2eb4cbfa1cb9a748a89dcbe9f805c31e4a seems to be useful and work
   - commit https://github.com/M4I-nanoscopy/openmicroscopy/commit/8cd20dca93c613bcb8ba2b394a9109e751603e82 does nothing for me in my particular setup, but, it did not interfere with me setting up the mapping manually later

PaulVanSchayck:
I believe there is no official consensus on which fields in a LDAP implementation may contain middleName or institution. 

For the ActiveDirectory implementation I was testing the field for middleName was in fact extensionAttribute3.

joshmoore:
`extensionAttribute3` *au weia*... Thanks, @PaulVanSchayck & @pwalczysko. My assumption is that this will be fine and users can actively set another value as they need. The most likely trouble we could get into is needing to *change* the default which would amount to a breaking change. As long as no one foresees that happening, then I'd say let's merge and see who uses it.

PR #4417, opened 2016-01-19
Webgateway url ids

will-moore:
This addresses various errors that arise when non-integer values are used in webgateway urls.
The url regex expressions were previously were not ensuring that numbers were used in the url for IDs etc, hence allowing any values to be used.
E.g. ```/webgateway/render_thumbnail/3933597/Simon-Josh/```
See https://trello.com/c/dPcurAje/41-render-thumbnail-invalid-int
Now, all IDs and sizes in webgateway urls use ```[0-9]``` to match only numerical values.
To test, try to go to that url.
In addition, check that the viewer still works OK, line plot, ROIs etc.

In addition, this fixes https://trello.com/c/H1uBUxL8/39-render-roi-thumbnail-index-out-of-range where you get an error ```IndexError: list index out of range``` at ``` webgateway/render_roi_thumbnail/0 ```. This will now give a simple 404 error (for any ROI ID for which no shapes are found).
To test, try to visit that url.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/urls.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/urls.py
@@ -45,7 +45,7 @@ object type and identifier, optionally traversing object model graph.
 """
 
 render_image = (
-    r'^render_image/(?P<iid>[^/]+)/(?:(?P<z>[^/]+)/)?(?:(?P<t>[^/]+)/)?$',
+    r'^render_image/(?P<iid>[0-9]+)/(?:(?P<z>[0-9]+)/)?(?:(?P<t>[0-9]+)/)?$',
     'webgateway.views.render_image')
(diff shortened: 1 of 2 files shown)


gusferguson:
@will-moore 

Tested using https://cowfish.openmicroscopy.org/merge/webclient/ user-1 public-group

https://cowfish.openmicroscopy.org/merge/webgateway/render_thumbnail/31851/

Text added on to thumbnail URL - in browser gives 404 error message - in web page gives no image icon.
Viewer works as expected - but could not test the line plot and ROIs as the logistics of getting an image with ROIs into cowfish public defeated me.
Otherwise - behaves as expected.

https://cowfish.openmicroscopy.org/merge/webgateway/render_thumbnail/0/
Gives 404 error page.

Good to merge.

PR #4425, opened 2016-01-21
Fix conn.getInterpolateSetting() always return True

will-moore:
This fixes a bug in Blitz Gateway (#4191) where conn.getInterpolateSetting() always returns True.
See https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7991&p=16657

To test
```
$ bin/omero config set omero.client.viewer.interpolate_pixels false
```
Restart omero server and log out of web client.
Then log in and check that viewer has interpolation off by default.

Question: is there any way we could have tests for config service methods within Blitz Gateway? Seems tricky if they require a server restart?

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_config_service.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_config_service.py
@@ -20,14 +20,16 @@ def testInterpolateSetting(gatewaywrapper):
     """
     gatewaywrapper.loginAsAdmin()
     configService = gatewaywrapper.gateway.getConfigService()
-    configService.setConfigValue("omero.client.viewer.interpolate_pixels", "true")
+    configService.setConfigValue(
+        "omero.client.viewer.interpolate_pixels", "true")
 
     gatewaywrapper.loginAsAuthor()
     assert gatewaywrapper.gateway.getInterpolateSetting()
 
     gatewaywrapper.loginAsAdmin()
     configService = gatewaywrapper.gateway.getConfigService()
-    configService.setConfigValue("omero.client.viewer.interpolate_pixels", "false")
+    configService.setConfigValue(
+        "omero.client.viewer.interpolate_pixels", "false")
 
     gatewaywrapper.loginAsAuthor()
     assert not gatewaywrapper.gateway.getInterpolateSetting()
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_config_service.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_config_service.py
@@ -20,17 +20,21 @@ def testInterpolateSetting(gatewaywrapper):
     """
     gatewaywrapper.loginAsAdmin()
     configService = gatewaywrapper.gateway.getConfigService()
-    configService.setConfigValue(
-        "omero.client.viewer.interpolate_pixels", "false")
 
-    gatewaywrapper.loginAsAuthor()
-    assert gatewaywrapper.gateway.getInterpolateSetting()
+    try:
+        configService.setConfigValue(
+            "omero.client.viewer.interpolate_pixels", "false")
 
-    # Set back to 'true' (default value) to clean up
-    gatewaywrapper.loginAsAdmin()
-    configService = gatewaywrapper.gateway.getConfigService()
-    configService.setConfigValue(
-        "omero.client.viewer.interpolate_pixels", "true")
+        gatewaywrapper.loginAsAuthor()
+        assert gatewaywrapper.gateway.getInterpolateSetting()
+
+    finally:
+        # try/finally to make sure that we
+        # set back to 'true' (default value) to clean up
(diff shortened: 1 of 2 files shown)


joshmoore:
> Question: is there any way we could have tests for config service methods within Blitz Gateway? Seems tricky if they require a server restart?

An admin account should be able to set the config value. We'll have to be slightly careful that there aren't race conditions. It you have a small set of properties so that we could define "flavors" of integration servers, we could periodically swap and use [skipIf](https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.1/components/tools/OmeroPy/test/integration/test_mail.py#L86).


Quick aside: the calls to getConfigValue are seeing 30-second hangs on some servers. Considering the number called, this is quite the overhead.

will-moore:
@joshmoore Simple test of conn.getInterpolateSetting() pushed.
This seems to work OK.
Not sure I understand your reference to skipIf?

mtbc:
```
test/integration/gatewaytest/test_config_service.py::testInterpolateSetting PASSED
```

Works just fine in the viewer respecting the setting of `omero.client.viewer.interpolate_pixels`.

CI being green tomorrow morning, seems to me to be good to merge.

joshmoore:
> Not sure I understand your reference to skipIf?

Tests can be written which expect each of a couple of settings. i.e. you could write a `def testInterpolateOn` and a `def testInterpolateOff` but of which start with a call to `skipIf`. If the server has been set with interpolation on, then only `testInterpolateOn` would run for example. The benefit here is that it doesn't require a code change to test the other configuration, just a (possibly temporary) PR against mgmt_tools.

will-moore:
@joshmoore The approach I've used in this test is OK? Setting the config and testing each value (true / false)?

joshmoore:
It will likely work in the short-term, but there is certainly a race-condition here if any other tests (run in parallel) assume either `interpolate` is on or off. Minimally, it might help to reset the value to the original value in a try/finally block.

We might should discuss. cc @sbesson

jburel:
Maybe double-check insight too since the property is also used in insight
i.e. check that both clients are in synch

aleksandra-tarkowska:
Test is failing https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/191/testReport/OmeroPy.test.integration.gatewaytest/test_config_service/testInterpolateSetting/

components/tools/OmeroPy/test/integration/gatewaytest/test_config_service.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    gatewaywrapper.loginAsAdmin()
+    configService = gatewaywrapper.gateway.getConfigService()
+
+    try:
+        configService.setConfigValue(
+            "omero.client.viewer.interpolate_pixels", "false")
+
+        gatewaywrapper.loginAsAuthor()
+        assert gatewaywrapper.gateway.getInterpolateSetting()
```
aleksandra-tarkowska: this code test `True`, you are missing `not`

will-moore:
Thanks @aleksandra-tarkowska - I had it passing locally because my ```target``` BlitzGateway had got out of sync with ```src```. Fixed now.

will-moore:
@jburel I'm not sure if you're asking for anything else in this PR above?

jburel:
@will-moore: it was just a quick check to make sure that both clients have the interpolation on/off depending on the value set in the configuration.

will-moore:
@jburel I tested Insight the other day when I responded to https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7991&p=16657 and it was working.
The fix was tested in web above. So now that tests are passing (https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/194/testReport/OmeroPy.test.integration.gatewaytest/test_config_service/testInterpolateSetting/) I think we are good to merge?

jburel:
@will-moore: thanks 
Merging

components/tools/OmeroPy/test/integration/gatewaytest/test_config_service.py
```diff
@@ -0,0 +1,40 @@
...
+        gatewaywrapper.loginAsAuthor()
+        assert not gatewaywrapper.gateway.getInterpolateSetting()
+
+    finally:
+        # try/finally to make sure that we
+        # set back to 'true' (default value) to clean up
+        gatewaywrapper.loginAsAdmin()
+        configService = gatewaywrapper.gateway.getConfigService()
+        configService.setConfigValue(
+            "omero.client.viewer.interpolate_pixels", "true")
```
aleksandra-tarkowska: you should use ``getConfigService().getConfigDefaults()`` to reset to what is default rather then hard code it. IF default change this test will affect others.

will-moore: See https://github.com/openmicroscopy/openmicroscopy/pull/4461

PR #4159, opened 2015-09-10
Temporarily bump MessageSizeMax to 250mb

joshmoore:
Import of the condensation screens failed with

```
==> Summary
63171 files uploaded, 0 filesets created, 0 images imported, 0 errors in 2:47:28.118
2015-09-10 15:20:46,922 10354261   [1-thread-1] ERROR  me.formats.importer.util.ClientKeepAlive - Exception while executing ping(), logging Connector out:
java.lang.RuntimeException: Ice.CommunicatorDestroyedException
        at ome.formats.OMEROMetadataStoreClient.ping(OMEROMetadataStoreClient.java:763) ~[blitz.jar:na]
```

```
-! 09/10/15 15:20:14.841 OMERO.Glacier2: warning: dispatch exception: OutgoingAsync.cpp:720: Ice::UnknownLocalException:
   unknown local exception:
   Ice::MemoryLimitException
   Ice.MemoryLimitException
       reason = "requested 67108865 bytes, maximum allowed is 67108864 bytes (see Ice.MessageSizeMax)"
```

PR #4428, opened 2016-01-21
bump MessageSizeMax to 250mb (rebased onto develop)

joshmoore:
This is the same as gh-4159 but rebased onto develop.

----

Import of the condensation screens failed with

```
==> Summary
63171 files uploaded, 0 filesets created, 0 images imported, 0 errors in 2:47:28.118
2015-09-10 15:20:46,922 10354261   [1-thread-1] ERROR  me.formats.importer.util.ClientKeepAlive - Exception while executing ping(), logging Connector out:
java.lang.RuntimeException: Ice.CommunicatorDestroyedException
        at ome.formats.OMEROMetadataStoreClient.ping(OMEROMetadataStoreClient.java:763) ~[blitz.jar:na]
```

```
-! 09/10/15 15:20:14.841 OMERO.Glacier2: warning: dispatch exception: OutgoingAsync.cpp:720: Ice::UnknownLocalException:
   unknown local exception:
   Ice::MemoryLimitException
   Ice.MemoryLimitException
       reason = "requested 67108865 bytes, maximum allowed is 67108864 bytes (see Ice.MessageSizeMax)"
```

### DIFF
--- a/etc/templates/grid/templates.xml
+++ b/etc/templates/grid/templates.xml
@@ -30,7 +30,7 @@
     </properties>
 
     <properties id="Basics">
-      <property name="Ice.MessageSizeMax" value="65536"/> <!-- 64 MB -->
+      <property name="Ice.MessageSizeMax" value="250000"/> <!-- 64 MB -->
       <property name="Ice.CacheMessageBuffers" value="0"/>
       <property name="Ice.Override.ConnectTimeout" value="5000"/>
     </properties>
--- a/etc/templates/grid/templates.xml
+++ b/etc/templates/grid/templates.xml
@@ -30,7 +30,7 @@
     </properties>
 
     <properties id="Basics">
-      <property name="Ice.MessageSizeMax" value="250000"/> <!-- 64 MB -->
+      <property name="Ice.MessageSizeMax" value="250000"/> <!-- 250 MB -->
       <property name="Ice.CacheMessageBuffers" value="0"/>
       <property name="Ice.Override.ConnectTimeout" value="5000"/>
     </properties>
--- a/components/blitz/resources/omero/Constants.ice
+++ b/components/blitz/resources/omero/Constants.ice
@@ -43,10 +43,12 @@ module omero {
     const int GLACIER2PORT = 4064;
 
     /**
-     * Default Ice.MessageSizeMax (65536kb). Not strictly necessary, but helps to
+     * Default Ice.MessageSizeMax (250mb). Not strictly necessary, but helps to
      * curb memory issues. Must be set before communicator initialization.
+     *
+     * Temporarily set to 250MB
      **/
-    const int MESSAGESIZEMAX = 65536;
+    const int MESSAGESIZEMAX = 250000;
 
     /**
      * Determines the batch size for sending


joshmoore:
--rebased-from #4159

joshmoore:
I've opened this against the mainline due to recent conversations like https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7980&p=16632#p16632

If we choose to *not* bump, that's fine; I'll maintain it as a patch. As discussed in gh-4159, it would be possible to make the value configurable on server start, but clients would need to be modified to use the larger value. This is likely not worth the cost.

etc/templates/grid/templates.xml
```diff
@@ -30,7 +30,7 @@
     </properties>
 
     <properties id="Basics">
-      <property name="Ice.MessageSizeMax" value="65536"/> <!-- 64 MB -->
+      <property name="Ice.MessageSizeMax" value="250000"/> <!-- 64 MB -->
```
mtbc: update the comment :smiley:

joshmoore: :smile:

mtbc:
Looks great, **good to merge** for me, though might want to wait today to see if other testing on eel throws up any issues or anyone objects to the bump.

jburel:
are the ```condensation screens``` the biggest we have as of today?

joshmoore:
Mitocheck may be larger with the large number of ROIs, but I didn't try importing it until later (after I had already bump MSM). Might be worth trying a plate of each to see how many fails.

joshmoore:
@jburel : next steps here?

manics:
Are there any downsides to increasing it?

joshmoore:
Primary downside is the potentially memory use since we support up to 100 connections (by default) each of which could be streaming 250 MB.

polarise:
Is it possible to configure some form of lazy loading so that only the data needed by the client immediately is fetched? This way we can have the best of all worlds.

joshmoore:
@polarise : not at the moment, no. What method are you currently looking at? If you are accessing pixel data, the suggested method is to load in tiles to not hit the limit. If you are querying metadata, paging is the suggested solution.

polarise:
I'm loading a lot of programmatically generated ROIs. Where would this lie? Tiling or paging? Could you please point me to the docs on how to do this?

joshmoore:
What language are you working in? Is your code available anywhere, even in snippet form?

polarise:
I'm using Python. The MessageSizeMax limit actually affects OMERO.insight client and the webclient. My goal is to generate thousands of ROIs then view them with the available clients.

jburel:
I will discuss testing plan with @pwalczysko and @bramalingam  tomorrow

joshmoore:
At the moment, unfortunately, there isn't support for paging ROIs in the clients. For others benefit, @polarise's exception is:

```
java.lang.Exception: omero.gateway.exception.DSAccessException: Cannot access data. 
Cannot load the ROI for image: 151
	at omero.gateway.facility.Facility.handleException(Facility.java:206)
	at omero.gateway.facility.ROIFacility.loadROIs(ROIFacility.java:239)
	at org.openmicroscopy.shoola.env.data.OMEROGateway.loadROI(OMEROGateway.java:6154)
	at org.openmicroscopy.shoola.env.data.OmeroImageServiceImpl.loadROIFromServer(OmeroImageServiceImpl.java:1578)
...
	at org.openmicroscopy.shoola.util.concur.tasks.AsyncProcessor$Runner.run(AsyncProcessor.java:91)
	at java.lang.Thread.run(Thread.java:745)
Caused by: Ice.UnknownLocalException
    unknown = "Ice::MemoryLimitException
               Ice.MemoryLimitException
                   reason = "requested 67112210 bytes, maximum allowed is 67108864 bytes (see Ice.MessageSizeMax)"
               	at IceInternal.Ex.throwMemoryLimitException(Ex.java:34)
```

which makes me think that you are close to the limit that you need, so bumping MessageSizeMax may work for the moment. How many ROIs are you dealing with here? And will that be roughly the maximimum number in the short-term?

polarise:
The current image I'm working with (the smallest) has 493 ROIs. I have larger images which will end up with thousands of ROIs. I've had to cap them to just under 300, which is below the MessageSizeMax limit.

pwalczysko:
Did locally with yesterdays merge build server ``OMERO.server-5.2.1-529-3988194-ice35-b238``.
Imported a large plate (took 3:35:52.706, this is 3 and a half hours approx).
Size of the plate 173.89 GB. 
All went fine.
```
bin/omero import -- --transfer=ln_s /Users/pwalczysko/Downloads/plate1_1_013/experiment_descriptor.xml 
....
6571
6633
Other imported objects:
Fileset:3102
Plate:551
2016-02-17 23:34:29,566 13079908   [.Client-52] INFO      ome.formats.importer.cli.ErrorHandler - Number of errors: 0

==> Summary
63171 files uploaded, 1 fileset, 1 plate created, 96 images imported, 0 errors in 3:35:52.706
```

jburel:
https://trello.com/c/e0Rw4QiF/136-paging

PR #4429, opened 2016-01-21
Fix duplicate screen target (rebased onto develop)

joshmoore:
This is the same as gh-4348 but rebased onto develop.

----

This PR addresses a problem encountered when importing `.screen` files.

 * *Without* this patch, if the user specifies a screen target, plates are linked to that screen *and* to the screen specified by Bio-Formats' `ScreenReader`
 * *With* this patch, the target specified by the user overrides the BF one

**NOTES**

 * It would be nice to simply skip `addReference` in `setScreenPlateRef`, but at that point in the import process `userSpecifiedTarget` has not been set yet (on the server's side)
 * `screenKey` should probably be changed to a list, or removed altogether if scanning the whole `referenceCache` for screen-plate refs is not too intensive
 * BF adds a screen -> plate ref, while `TargetProcessor` [adds a plate -> screen one](https://github.com/openmicroscopy/openmicroscopy/blob/983abca738b10367fe20ba2653ed99d3971e74ed/components/blitz/src/ome/formats/model/TargetProcessor.java#L86). Does this mean one of the two has to be fixed?

### DIFF
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -263,6 +263,8 @@ public class OMEROMetadataStoreClient
      */
     private Map<String, String[]> referenceStringCache;
 
+    private LSID screenKey;
+
     /** Our model processors. Will be called on saveToDB(). */
     private List<ModelProcessor> modelProcessors;
 
@@ -1370,6 +1372,10 @@ public class OMEROMetadataStoreClient
     @Override
     public void setUserSpecifiedTarget(IObject target)
     {
+        if (target instanceof Screen && null != screenKey) {
+          log.info("deleting screen ref in favor of user-specified one");
+          referenceCache.remove(screenKey);
+        }
         this.userSpecifiedTarget = target;
     }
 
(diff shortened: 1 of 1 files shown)


joshmoore:
--rebased-from #4348

joshmoore:
Note: on develop, it won't be possible to test against `ScreenReader` but both fake files with screens/plates as well as OME-XML files with SPW information could be affected. cc: @ximenesuk

ximenesuk:
@simleo Do I need to use a specific file format for this test?

ximenesuk:
Checking with:
```
omero import -r 101 ~/Work/images/fakes/SPW\&plates\=1\&plateRows\=1\&plateCols\=1\&fields\=1\&plateAcqs\=1.fake
```
Without this PR an extra screen is created, with it that screen isn't created. So in that respect it works. However, I had assumed this problem was with the `FakeReader` and so fixed it in https://github.com/openmicroscopy/bioformats/pull/1984 so I am not sure where this leaves things when this PR meets BF 5.2

joshmoore:
@ximenesuk is away for a week or too. Considering that this is now functional, I'd propose merging and we follow up with testing in OMERO 5.3 taking @ximenesuk's changes in Bio-Formats 5.2. It might "just work", or we may need to revert one or the other.

merge vetoes? @sbesson @melissalinkert

melissalinkert:
Both sets of PRs seem to fix legitimate bugs independent of each other, so no veto here.

PR #4431, opened 2016-01-22
Disable orphaned container

aleksandra-tarkowska:
Allowing to disable orphaned container as tweaked in https://github.com/openmicroscopy/openmicroscopy/pull/4246/

Test defaults:
 - if you tested that PR before unset ``bin/omero config set omero.client.ui.tree.orphans ``otherwise you shouldn't touch it.
 - test own and someone else data as user, group owner and admin. you should see "Orphaned container" everytime 

Test custom name via old properties:
 - set ``bin/omero config set omero.client.ui.tree.orphans.name 'Trash'`` and ``bin/omero config set omero.client.ui.tree.orphans.description "This is trash."'``
 - test if custom name works

Test disabled:
**you can only disable while viewing other users data as regular user. For admin and group owners container will be present**
 - set ``bin/omero config set omero.client.ui.tree.orphans.enabled False``
 - load your own tree, default "Orphaned Images" should be present,
 - chose someone else data, tree should not load orphaned container
 - cut image if you have permissions

**Scenario updated as functionality has changed since https://github.com/openmicroscopy/openmicroscopy/pull/4431#issuecomment-177925015 and https://github.com/openmicroscopy/openmicroscopy/pull/4431#issuecomment-178493209**

### DIFF
--- a/components/common/resources/ome/config.xml
+++ b/components/common/resources/ome/config.xml
@@ -169,6 +169,11 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
+    <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans.enabled">
+       <property name="db" value="false"/>
+       <property name="mutable" value="false"/>
+       <property name="visibility" value="all"/>
+    </bean>
     <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans.name">
        <property name="db" value="false"/>
        <property name="mutable" value="false"/>
--- a/components/common/resources/ome/config.xml
+++ b/components/common/resources/ome/config.xml
@@ -174,6 +174,11 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
+    <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans.enabled">
+       <property name="db" value="false"/>
+       <property name="mutable" value="false"/>
+       <property name="visibility" value="all"/>
+    </bean>
     <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans.name">
        <property name="db" value="false"/>
        <property name="mutable" value="false"/>
--- a/components/common/resources/ome/config.xml
+++ b/components/common/resources/ome/config.xml
@@ -169,7 +169,7 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
-    <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans.enabled">
+    <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans">
        <property name="db" value="false"/>
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -641,7 +641,8 @@ def api_container_list(request, conn=None, **kwargs):
             page=page,
             limit=limit)
         # Get the orphaned images container
-        if (conn.isAdmin() or conn.isLeader() or
+        if (conn.isAdmin() or
+            conn.isLeader(gid=request.session.get('active_group')) or
             experimenter_id == conn.getUserId() or
             request.session['server_settings']
                            ['ui']['orphans']['enabled']):
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -641,7 +641,10 @@ def api_container_list(request, conn=None, **kwargs):
             page=page,
             limit=limit)
         # Get the orphaned images container
-        if request.session['server_settings']['ui']['orphans']['enabled']:
+        if (conn.isAdmin() or conn.isLeader() or
+            experimenter_id == conn.getUserId() or
+            request.session['server_settings']
+                           ['ui']['orphans']['enabled']):
             orphaned = tree.marshal_orphaned(
                 conn=conn,
                 group_id=group_id,
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -641,7 +641,7 @@ def api_container_list(request, conn=None, **kwargs):
             page=page,
             limit=limit)
         # Get the orphaned images container
-        if request.session['server_settings']['ui']['orphans']:
+        if request.session['server_settings']['ui']['orphans']['enabled']:
             orphaned = tree.marshal_orphaned(
                 conn=conn,
                 group_id=group_id,
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -605,43 +605,52 @@ def api_container_list(request, conn=None, **kwargs):
     # parents), screens and plates (without parents). This is fine for
     # the first page, but the second page may not be what is expected.
 
+    r = dict()
     try:
         # Get the projects
-        projects = tree.marshal_projects(conn=conn,
-                                         group_id=group_id,
-                                         experimenter_id=experimenter_id,
-                                         page=page,
-                                         limit=limit)
+        r['projects'] = tree.marshal_projects(
+            conn=conn,
+            group_id=group_id,
+            experimenter_id=experimenter_id,
+            page=page,
+            limit=limit)
 
         # Get the orphaned datasets (without project parents)
-        datasets = tree.marshal_datasets(conn=conn,
-                                         orphaned=True,
-                                         group_id=group_id,
-                                         experimenter_id=experimenter_id,
-                                         page=page,
-                                         limit=limit)
+        r['datasets'] = tree.marshal_datasets(
+            conn=conn,
+            orphaned=True,
+            group_id=group_id,
+            experimenter_id=experimenter_id,
+            page=page,
+            limit=limit)
 
         # Get the screens for the current user
-        screens = tree.marshal_screens(conn=conn,
-                                       group_id=group_id,
-                                       experimenter_id=experimenter_id,
-                                       page=page,
-                                       limit=limit)
+        r['screens'] = tree.marshal_screens(
+            conn=conn,
+            group_id=group_id,
+            experimenter_id=experimenter_id,
+            page=page,
+            limit=limit)
 
         # Get the orphaned plates (without project parents)
-        plates = tree.marshal_plates(conn=conn,
-                                     orphaned=True,
-                                     group_id=group_id,
-                                     experimenter_id=experimenter_id,
-                                     page=page,
-                                     limit=limit)
-
+        r['plates'] = tree.marshal_plates(
+            conn=conn,
+            orphaned=True,
+            group_id=group_id,
+            experimenter_id=experimenter_id,
+            page=page,
+            limit=limit)
         # Get the orphaned images container
-        orphaned = tree.marshal_orphaned(conn=conn,
-                                         group_id=group_id,
-                                         experimenter_id=experimenter_id,
-                                         page=page,
-                                         limit=limit)
+        if request.session['server_settings']['ui']['orphans']:
+            orphaned = tree.marshal_orphaned(
+                conn=conn,
+                group_id=group_id,
+                experimenter_id=experimenter_id,
+                page=page,
+                limit=limit)
+            orphaned['name'] = \
+                request.session['server_settings']['ui']['orphans']['name']
+            r['orphaned'] = orphaned
     except ApiUsageException as e:
         return HttpResponseBadRequest(e.serverStackTrace)
     except ServerError as e:
@@ -649,11 +658,7 @@ def api_container_list(request, conn=None, **kwargs):
     except IceException as e:
         return HttpResponseServerError(e.message)
 
-    return HttpJsonResponse({'projects': projects,
-                             'datasets': datasets,
-                             'screens': screens,
-                             'plates': plates,
-                             'orphaned': orphaned})
+    return HttpJsonResponse(r)
 
 
 @login_required()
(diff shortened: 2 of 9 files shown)


components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -605,43 +605,52 @@ def api_container_list(request, conn=None, **kwargs):
...
+            experimenter_id=experimenter_id,
+            page=page,
+            limit=limit)
         # Get the orphaned images container
-        orphaned = tree.marshal_orphaned(conn=conn,
-                                         group_id=group_id,
-                                         experimenter_id=experimenter_id,
-                                         page=page,
-                                         limit=limit)
+        if request.session['server_settings']['ui']['orphans']:
```
will-moore: Slightly uncomfortable with evaluating an empty dict() as False here.
Since you have 3 settings
```
omero.client.ui.tree.orphans.enabled
omero.client.ui.tree.orphans.name
omero.client.ui.tree.orphans.description
```
they might logically map to a session dict 
```request.session['server_settings']['ui']['orphans']``` with keys of ```enabled```, ```name``` & ```description```.

Then you could (if needed) still get the name & description even if orphans was disabled.

File: components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py [hunk 2]
```diff
@@ -179,12 +179,17 @@ def changeActiveGroup(self, gid):  # TODO: should be moved to ISession
...
+        if toBoolean(self.getConfigService().getConfigValue(
+                "omero.client.ui.tree.orphans.enabled")):
+            orphans['name'] = \
+                self.getConfigService().getConfigValue(
+                    "omero.client.ui.tree.orphans.name") or "Orphaned image"
+            orphans['description'] = \
+                (self.getConfigService().getConfigValue(
+                 "omero.client.ui.tree.orphans.description") or
+                 "This is a virtual container with orphaned images.")
+        return orphans
```
will-moore: When we're accessing config service in Blitz Gateway we now do try/except (see https://github.com/openmicroscopy/openmicroscopy/pull/4191) so it would be nice to do it here too, so we don't risk screwing up the session if this fails.

will-moore:
Functionality is working fine, so just the minor comments above.

aleksandra-tarkowska:
@will-moore thanks for picking it up. try now

will-moore:
Robot test is failing as it's looking for "Orphaned Images" (case sensitive) https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/205/robot/Web%20&%20Web/Web_1/Create%20Scenario/Test%20Container%20Creation%20Enabled/ 

Travis just failed too.

components/tools/OmeroWeb/omeroweb/webclient/views.py
aleksandra-tarkowska: Why do you need to make 2 extra calls to the server if setting is off?

joshmoore:
I wonder if it isn't time to add: `omero.cmd.ui.LoadConfig` which returns all the values (with defaults, etc) and can be run by guest once per worker.

aleksandra-tarkowska:
@joshmoore I think this requires more design discussion. All the parameters are loaded on login (https://github.com/aleksandra-tarkowska/openmicroscopy/blob/disable_orphand/components/tools/OmeroWeb/omeroweb/decorators.py#L250) and they should be really context related rather then global. I'd see them more as Experimenter/ExperimeterGroup annotation. For example admin account should have always access to orphans.

joshmoore:
@aleksandra-tarkowska : agreed, and that makes good sense, but in the shorter-term, we're seeing overhead issues with the large number of calls to `getConfigValue`. Until there's a better/faster session backend (OMERO not Django) that provides these values at less cost, it might still be worth having something to reduce the number of calls to the minimum possible.

aleksandra-tarkowska:
@joshmoore understood, lets get this PR in as it is and change all of them to use ``omero.cmd.ui.LoadConfig`` in a separate PR then. Are you ok with that?

joshmoore:
Yup, exactly. Just mostly didn't want the cost of more calls to make any decisions for this PR, since I think we need to handle that more fully.

will-moore:
Could we have the orphan session dict look like 
```
{'enabled': true,
 'name', 'Orphaned',
 'description', 'Some description'}
```
with the name and description not populated if enabled is false?
That way, we can explicitly state the enabled state, instead of implying that enabled is false by evaluating an empty dictionary.
```
if request.session['server_settings']['ui']['orphans']['enabled']
```
is nicer than
```
if request.session['server_settings']['ui']['orphans']
```

aleksandra-tarkowska:
My last commit did it already. Enable is either true or false. We are trying to limit number of unnecessary calls, as @joshmoore mentioned

aleksandra-tarkowska:
Why do you need {'enabled' False, 'name': None, 'description':None}?

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -179,12 +179,30 @@ def changeActiveGroup(self, gid):  # TODO: should be moved to ISession
...
+        except:
+            orphans['enabled'] = True
+        if orphans['enabled']:
+            try:
+                orphans['name'] = \
+                    (self.getConfigService().getConfigValue(
+                     "omero.client.ui.tree.orphans.name") or
+                     "Orphaned image")
+            except:
+                orphans['name'] = "Orphaned image"
```
will-moore: The default here should be "Orphaned Images"

will-moore:
@aleksandra-tarkowska Sorry for the misunderstanding. Simply ``` {'enabled': False} ``` is fine.

will-moore:
One potential idea to reduce the calls to config service is have the initial call to config service gets an Orphans enabled flag of "true", "false" or "custom". If the value is "true" then we can use hard coded values (as in the try/except blocks) and only if the value is "custom" to we have to look-up name and description from config service. This would mean that for all users who don't customise these values, they only have one call to config service instead of 3.

aleksandra-tarkowska:
I don't feel comfortable with testing string as bool. Boolean has two states. We will be moving that anyway for more practical solutions https://trello.com/c/cJCScHOl.

aleksandra-tarkowska:
@will-moore could you put your ideas in https://trello.com/c/cJCScHOl to discuss them when time comes? This PR covers recent request.

aleksandra-tarkowska:
@joshmoore @gusferguson the request from IDR was to disable orphaned container completely, although I think it should be always visible for admins. Otherwise data become hidden until someone start searching. Any thoughts about that?

gusferguson:
@aleksandra-tarkowska @joshmoore - I presume this will be a positive configuration choice and would only be used in circumstances where Admin completely controls data display - i.e.public repositories.
In those terms I can't see any user issues but if user was able to move data around it would be very confusing. I agree that admin should be able to see at all times.

will-moore:
@aleksandra-tarkowska I don't see that there are any types for config values. They are all strings. Nothing defines ```omero.client.ui.tree.orphans.enabled``` as boolean, although the naming would suggest that the value is "true" or "false".
Maybe we just need to consider different naming, but in any case it seems wasteful for all regular users to have to access 3 config values just to know that the defaults are used in each case, whereas one should be enough to know this.
This is only an issue if the performance of config service is bad, as @joshmoore suggests.

If this is purely for the IDR use-case, you could even restrict the hiding of Orphaned to non-owners of the data. That way, anyone else who is able to move their own data around won't lose it in Orphaned.

aleksandra-tarkowska:
@will-moore thanks probably that is the best option to limit to non-owners only (admin, group owner and owners will access as usual)

gusferguson:
That sounds good.

aleksandra-tarkowska:
@will-moore I see what you mean. In this case the following naming should work:
- *default* `omero.client.ui.tree.orphans.custom false` -> do not call server
- `omero.client.ui.tree.orphans.custom true` -> load from the server, regardless if is the same as default or not.

Does it make sense?

will-moore:
@aleksandra-tarkowska sounds great. For disabling (when ```custom``` is "true"), you could set ```orphans.name``` to an empty string? Or you might want to make it more explicit?

aleksandra-tarkowska:
just updated comment above, empty name = disabled

aleksandra-tarkowska:
@will-moore I have done all you suggested but I strongly disagree with the solution of hard coding that in webclient. The whole point of pushing them to the server is to share them with other clients. You are asking exactly about opposite. In this case I am going to add enable flag to local settings. Once ``omero.cmd.ui.LoadConfig`` is ready we will do it right.

aleksandra-tarkowska:
@will-moore all the discussion I presume putting everything in one property is the best choice. See last commit. I am not sure why this never been pointed in previous PRs like that.

will-moore:
@aleksandra-tarkowska OK, that seems like a nice idea and the code looks good.
So, to disable you could do:
```
$ bin/omero config set omero.client.ui.tree.orphans '{"enabled": false}'
```
But to then enable again, do you do this:
```
$ bin/omero config set omero.client.ui.tree.orphans '{"enabled": true, "name": "${omero.client.ui.tree.orphans.name}", "description": "${omero.client.ui.tree.orphans.description}"}'
```
Seems like it might be nicer to have an additional ```omero.client.ui.tree.orphans.enabled``` parameter, that works in the same way as ```name``` and ```description```. And then we'd have to "hide" the ```omero.client.ui.tree.orphans``` parameter or somehow document that it shouldn't be used directly?

components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -605,43 +605,55 @@ def api_container_list(request, conn=None, **kwargs):
...
+            experimenter_id=experimenter_id,
+            page=page,
+            limit=limit)
         # Get the orphaned images container
-        orphaned = tree.marshal_orphaned(conn=conn,
-                                         group_id=group_id,
-                                         experimenter_id=experimenter_id,
-                                         page=page,
-                                         limit=limit)
+        if (conn.isAdmin() or conn.isLeader() or
```
will-moore: ```conn.isLeader()``` tells you if you are leader of your default group, not necessarily the group you're currently looking at. I think you need to use ```conn.isLeader( gid=request.session.get('active_group') )```

joshmoore: Not that it's something we can change at the moment, but if we make this mistake (and I likely would have too) others will as well.

aleksandra-tarkowska:
@will-moore to disable you can already do just

```
$ bin/omero config set omero.client.ui.tree.orphans '{"enabled": false}'
```

to enable it you just unset ``$ bin/omero config set omero.client.ui.tree.orphans`` and all is back to the values

aleksandra-tarkowska:
@will-moore you right enabled flag will sort it out. Adding now

aleksandra-tarkowska:
@will-moore try now

aleksandra-tarkowska:
If we are happy with wrapping dict I will do the same for ``ui.dropdown`` labels in a separate PR.

joshmoore:
There's been a bunch of backforth (which is great), but to clarify for me: is there anything in this change which deprecates/modifies previous behavior?

aleksandra-tarkowska:
@joshmoore I don;t think so. This is transparent. `.name` and `.description` remain unchanged. Even user interaction with setting config hasn't changed. We just introduced "umbrella property" and `.enabled` flag.

will-moore:
This seems to be working well now.
With ``` $ omero config set omero.client.ui.tree.orphans.enabled false``` this disables orphans for non-owners of the data (even if they are a Leader of another Group), but doesn't disable orphans for Admin or Owner of the current Group.

One thing I noticed, which is more-like something to be aware of rather than a blocker
If you do this:
```
$ omero config set omero.client.ui.tree.orphans.enabled False
```
it will cause the ```json.loads()``` to fail, since ```False``` is not valid (needs to be ```false```).
So you will end up with the defaults.
But I think there's only so much we can do here! 

Good to merge.

aleksandra-tarkowska:
@will-moore let me fix it, just needs ""
```
>>> j = '{"enabled":"False"}';
>>> json.loads(j)
{u'enabled': u'False'}
```

aleksandra-tarkowska:
once your https://github.com/openmicroscopy/openmicroscopy/pull/4425 is merged I will add test for that

components/tools/OmeroWeb/test/integration/test_config.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        if enabled is None:
+            enabled = self.root.sf.getConfigService() \
+                .getConfigDefaults()['omero.client.ui.tree.orphans.enabled']
+        self.root.sf.getConfigService().setConfigValue(
+            "omero.client.ui.tree.orphans.enabled", enabled)
+        conn = omero.gateway.BlitzGateway(client_obj=self.client)
+        orphans = conn.getOrphanedContainerSettings()
+        if enabled is None:
+            assert orphans['enabled'] == toBoolean(enabled)
```
will-moore: Seems you're missing any assertion for all the values that are not None.
``` "foo", "", "False", "false", "True", "true", "" ```

Comment (aleksandra-tarkowska) [inline components/tools/OmeroWeb/test/integration/test_config.py]: sorry I don't know how this line ended up here

File: components/common/resources/ome/config.xml [hunk 6]
```diff
@@ -169,6 +169,16 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
+    <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans">
+       <property name="db" value="false"/>
+       <property name="mutable" value="false"/>
+       <property name="visibility" value="all"/>
+    </bean>
+    <bean class="ome.system.Preference" id="omero.client.ui.tree.orphans.enabled">
```
joshmoore: This this still needed or is this what `...tree.orphans` is now for?

aleksandra-tarkowska: ``omero.client.ui.tree.orphans`` is wrapper to reduce number of calls and shouldn't being touch. People should use only ``.enabled``, ``.name``, and ``.description``. Documentation will be updated accordingly.

jburel:
2 comments:
 * The type returned by the method ```getOrphanedContainerSettings``` cannot be changed during 5.2.x series. This is considered a breaking change. So we will have to mark the existing method, deprecated and add a new method.
 * insight will have to be synchronized so it also uses the ``enabled`` flag.

aleksandra-tarkowska:
@jburel I am voting to push getOrphanedContainerSettings to OmeroPy gateway then and deprecated webclient one

aleksandra-tarkowska:
To simplify, could we merge this PR and I will open new one to sort out drop down menu labels in the same time? All methods from webclietn_gateway.py will be pushed to Python gateway and deprecated internally

PR #4430, opened 2016-01-21
render.py: new CLI plugin for dealing with rendering settings (rebased onto develop)

joshmoore:
This is the same as gh-4385 but rebased onto develop.

----

See the `bin/omero render -h` output and the test for expected behavior. Primary driver for the creation of `render.py` was scripting the rendering settings along with the imports so that everything could be re-done if necessary:

```
PLATE=$(bin/omero import ... | grep Plate:)
bin/omero render copy Image:1 $PLATE
```

 * [ ] https://gist.github.com/emilroz/3858429bccd3012bf478

PR #4432, opened 2016-01-22
Masks in web viewer

will-moore:
This is a first stab at mask support in web viewer, see https://trello.com/c/H1MePy2A/124-mask-support-in-web

It seems to work fine for me, using masks created by @joshmoore's script at https://github.com/openmicroscopy/openmicroscopy/pull/4415 but it would be nice to test against "real" masks.

@emilroz Does that Mask-reading code look right?

screen shot 2016-01-28 at 11 20 24


pwalczysko:
Apparently, we do not have any workflows how to get masks into OMERO.  None of us has ever done this and the insight-IJ plugin does not support masks conversion.

emilroz:
@pwalczysko @will-moore I will write some examples for you that 1) create a mask from a png file and attaches it to the OMERO image and 2) reads a mask from the OMERO.server and saves it to a png file. @pwalczysko is a python script ok for this?

pwalczysko:
@emilroz yes. thank you

joshmoore:
@emilroz : note https://github.com/openmicroscopy/openmicroscopy/pull/4430 if you think there's a case for having CLI tools for converting to/from masks.

@pwalczysko : in general, I think masks will often be coming from third-party tools (CellProfiler) so having it script-based makes sense.

will-moore:
@emilroz if you don't have time to work on examples now, we can just test this PR with the existing script at https://github.com/openmicroscopy/openmicroscopy/pull/4415/files?

emilroz:
@will-moore sorry for delaying, should have something later today.

emilroz:
@will-moore and @pwalczysko sorry it took me while but the script is ready now: https://gist.github.com/emilroz/3858429bccd3012bf478 . I've used Will's code for mask reading and it works fine for me. @pwalczysko you should be able to use the script for creating new masks from image file and reading existing masks to a file:

create mask:
```python mask_utils.py -s some.omero.server.com -u user-name -w pass upload_mask -f test.tif -i 65201 --color red --threshold 20 ```
read mask:
```python mask_utils.py -s some.omero.server.com -u user-name -w password download_mask -id roi_d```

```python mask_utils.py --help``` should give you all the options with explanations. Let me know if you have troubles using it.

@joshmoore will do my best to convert it to a CLI tool if it's useful.

**Comment by pwalczysko** [comment]
@emilroz : Thank you very much. The creation of the mask part of the script works fine. Also addition of the -x and -y options for a shift of the mask works fine.
Nevertheless, the downloaded mask seems quite different to the mask in OMERO.
Compare https://cowfish.openmicroscopy.org/merge/webclient/?show=image-21911 and the mask there iwth the .png which I downloaded from it using the script (below). Or you can compare it with the oritinal image from which the mask was created (the second screenshot - I have just overlaid in ImageJ a Threshold of 100 (the same threshold value I used for the script parameter)



mask_125007_139807
screen shot 2016-02-12 at 15 48 27

**Comment by pwalczysko** [comment]
Nothwithstanding some export issues of the script https://github.com/openmicroscopy/openmicroscopy/pull/4432#issuecomment-183382614, (thank you @emilroz ), all looks good in this PR.

screen shot 2016-02-12 at 16 12 01

We have unrelated, although important problem in Insight though here. When the mask was created using https://gist.github.com/emilroz/3858429bccd3012bf478 @emilroz 's script, then it is displayed okay in Insight (and web) but 
  - Insight does not allow to save any more ROIs on the image after the mask was created
  - it is possible to create some ROIs in Insight, then create the mask using https://gist.github.com/emilroz/3858429bccd3012bf478 and all works fine but
  - no further ROIs after the mask ROI can be saved in Insight (the buttons react, but no action proceeds, after closure of the i mage the number of regions is still the same, although all the dialogs were clicked right previously) @dominikl - should I create a ticket for this ?

**Comment by pwalczysko** [comment]
https://trac.openmicroscopy.org/ome/ticket/13158#ticket created. Ready to merge.

**Comment by jburel** [comment]
@pwalczysko: found this old ticket during trac review
https://trac.openmicroscopy.org/ome/ticket/4336
Suggested file ``squig/flex/003003?``. 
Could you try it?

**Comment by will-moore** [comment]
Unfortunately if there are any Masks on ```test_images_good/flex/003003000.flex``` they are not turned into ROIs in OMERO on import:

screen shot 2016-02-18 at 16 12 37

**Comment by pwalczysko** [comment]
The same result as @will-moore re https://github.com/openmicroscopy/openmicroscopy/pull/4432#issuecomment-185791612 just double checked in Insight. 
screen shot 2016-02-18 at 16 33 39

Is our suspicion right that we actually do not have any reader for the masks in Bioformats ? cc @rleigh-dundee @melissalinkert

**Comment by jburel** [comment]
Thanks for testing, I don't remember seeing any rois with that file.

**Comment by sbesson** [comment]
Looking at the Bio-Formats readers, it looks like MIAS and Trestle have the ability to read masks from the proprietary file format. That being said, reading masks in Bio-Formats may likely fall into one of the current [codegen limitations](https://trello.com/c/UjyhkwhA/8-model-defects-code-generation-limitations-and-mask-bindata) which we should be addressed in Bio-Formats 5.2.0.

**Comment by jburel** [comment]
ok. We can review the support of masks when they are correctly read.
@pwalczysko: insight has limited support for masks. When they were first introduced in the client, we decided not to offer the ability to add other types of rois.

**Comment by jburel** [comment]
@will-moore:
default fill color:
```
/** The default fill color. */
    public final static Color DEFAULT_FILL_COLOUR = new Color(0, 0, 0, 64);
```
That's the value in insight.

jburel:
@chris-allan  @emilroz: any strong feeling about the default fill color? 
Both values have been set arbitrarily

pwalczysko:
There ARE masks on https://eel.openmicroscopy.org/merge/webclient/?show=plate-1210 (the ``mias/frans`` plate). These are not coming from OMERO. They display fine in web.

jburel:
@pwalczysko: thanks

jburel:
it will be nice to have tests for the ```render_*``` methods, tricky
In the new method ```render_shape_mask```, could you add a ``TODO`` instead of "Maybe look at this later?" so we can easily find time so we can spend time investigating.

jburel:
Thanks @will-moore 
Looking at the rendering methods, we could have robot tests similar to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tests/ui/testcases/insight/thumbnails/blank-thumbnails.txt. In the Java case, a library to handle rendered image was required.

This is not for this PR, cc @bramalingam.
I will create a card.

PR #4424, opened 2016-01-20
bring regions back into line with develop

mtbc:
no new features, check that CI passes

PR #4433, opened 2016-01-22
Merge 5.2.1 into regions

sbesson:
This PR is an alternative to https://github.com/openmicroscopy/openmicroscopy/pull/4424 which reintegrates the state of the released mainline back into the `regions` branch. Concretely, the last release tag v5.2.1 is integrated into the branch.

Two additions are made on top of the mainline commits:
- a set of commits to patch the `sql` history on the `regions` branch with the last DB changes brought before the `5.2.0` release
- b442a32 creates a duplicate `version.bioformats` entry under `omero.properties` pointing to `5.2.0-SNAPSHOT` and overriding the mainline Bio-Formats version. This should allow future bumps of the Bio-Formats version for upcoming 5.2.x versions to merge cleanly into this branch in the future.


sbesson:
http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/9/ being green (apart from one transient search failure), I propose to get this merged to unlock the upcoming model work @mtbc. We can discuss our strategy for merging `develop` into `regions` as we make progress in-between releases.

mtbc:
No objections to merging this in but why not #4424 instead? We've already made progress in between releases: why not just merge `develop` in promptly? The benefit is that we then do regions work that builds on top of those subsequently merged PRs instead of doing work on an out-of-date codebase then opening a PR, discovering it conflicts, then having to fix that. What's the downside?

sbesson:
@mtbc: all good points. I am still debating the `develop` vs tag merging strategy mostly because we have more confidence in tags as opposed to the latest development branch. Arguably, part of the SQL workaround in this PR would had been avoided had we had the opportunity to start the `regions` branch on top of `v5.2.0` directly. On the other side, maybe this concern is not that valid once we enter patch releases and I certainly understand your conflict worry.
It might be we need to adopt slightly different strategies between `regions` and `metadata` from which production system are built. /cc @joshmoore

mtbc:
Good to merge, then will open a PR for develop latest.

sbesson:
Discussed with @mtbc. Merging this now and we can open a subsequent PR merging the current state of `develop`.

PR #4434, opened 2016-01-23
Add slack notification to .travis.yml

joshmoore:
This will send notifications to the #travis room.


sbesson:
This PR produces the desired effect. As the notifications go to a dedicated room, I would expect the noise on the team would be minimal. Merging and we can revert later if needed.

PR #4437, opened 2016-01-25
merge develop/latest into regions

mtbc:
Regions CI should remain blue.


mtbc:
http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/10/ is blue.

PR #4438, opened 2016-01-25
Compression and thumbnail scaling

dominikl:
This PR addresses two issues which came up on the [Forum](https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7991).

- Only switch the default to high compression for non-LAN connections (previously the compression for large images ( between 512x512px and 2000x2000px) was automatically set to "high" which led to visible JPEG artifacts). **Test**: Open a large image in full viewer, make sure compression is set to "none" (when your connection is set to be a LAN connection) Note: On this occasion I changed the Insight property "CONNECTION_SPEED" to "IMAGE_QUALITY_LEVEL" because that's what the property is actually about. [Ticket 13148](http://trac.openmicroscopy.org/ome/ticket/13148)

- Preserve the chosen thumbnail size. **Test**: Change the thumbnail size (slider at the bottom left of the central panel). Make sure this setting is preserved, also after a restart of Insight. [Ticket 13149](http://trac.openmicroscopy.org/ome/ticket/13149)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/ThumbnailProvider.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/ThumbnailProvider.java
@@ -43,8 +43,7 @@ import omero.gateway.model.PixelsData;
 /** 
  * The class hosting the thumbnail corresponding to an {@link ImageData}.
  * We first retrieve a thumbnail of dimension {@link #THUMB_MAX_WIDTH}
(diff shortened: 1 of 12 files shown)


gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.1-210-0bc3951-ice35-b212-mac eel user-3 read-only-1

Compression set to none.
Thumbnail size preserved including after restart.
Behaves as expected.
Good to merge.

jburel:
Thanks @dominikl 
Merging

PR #4439, opened 2016-01-25
Node not valid exception

dominikl:
This would "fix" [Ticket 9490](https://trac.openmicroscopy.org/ome/ticket/9490) by simply just not throwing the exception. I couldn't replicate the issue by any of the actions reported in the QAs, but I found a different way which also throws the exception: In the Projects/Datasets tree switch to Group Display (i. e. disable "Display Users"). Expand the groups. Go to search tab. Search for any image. Select an image in the search results view. Open the "Located in" tab on the right hand side metadata panel. Double click on the dataset. This will trigger the exception. With the PR the Project/Datasets tab will be active again, with the dataset selected.
/cc @jburel


gusferguson:
@dominikl 

Tested using OMERO.insight-5.2.1-210-0bc3951-ice35-b212-mac eel user-3 read-only-1

test sequence above behaves as expected and does not throw an exception.
Good to merge.

PR #4440, opened 2016-01-25
Update OMERO build to consume BF5.1.8

bramalingam:
@sbesson @joshmoore @jburel


sbesson:
As discussed with @bramalingam, this should allow to test https://github.com/openmicroscopy/bioformats/pull/2209 via OMERO.

PR #4442, opened 2016-01-25
Delete rois docs

will-moore:
This improves our documentation & examples of deleting ROIs with conn.deleteObjects() and gives an example of using the ```wait=True``` parameter.

See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-January/005822.html

Check that tests are passing for these examples and the code changes & comments make sense.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3882,6 +3882,7 @@ class _BlitzGateway (object):
                                 * 'Well'
                                 * 'Annotation'
                                 * 'OriginalFile'
+                                * 'Roi'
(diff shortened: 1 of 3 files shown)


joshmoore:
One other option here that I've used recently when testing your and my example PRs is:

```
bin/omero delete /Image/Roi:$IMAGEID
```

will-moore:
@joshmoore That's useful for sure, but should probably go under CLI docs rather than here.

ximenesuk:
This looks good to merge unless @joshmoore has a strong reason for a `skiphead` delet being used.

joshmoore:
Yeah, I'll merge since adding another example would require a code change, but I was thinking less of that for the CLI docs, but that such an example would also be useful for the gateway.

PR #4444, opened 2016-01-26
Make lens tool work properly with units

dominikl:
Fixes [QA 16995](https://www.openmicroscopy.org/qa2/qa/feedback/16995/)

Test:
- Use the lens tool with an image which has pixel size information. Toggle between pixels and microns. Make sure the values make sense.
- Use the lens tool with an image without pixel size information. Make sure lens tools also works and doesn't crash. The "Microns" option should be disabled.

screen shot 2016-01-26 at 10 35 44


pwalczysko:

PR #4411, opened 2016-01-14
Improve OMERO build system integration with local Maven build

sbesson:
Previously the local Maven resolver was using the default Ivy pattern (ending with `.xml`) which was making the build incompatible with local artifacts build by Maven i.e. when working on coupled branches. This PR attempts to improve the developer workflow by taking advantage of the resolution of `.pom` files by Ivy.

To test this PR:

1. clean `~/.m2/repository/ome` (or alternatively clean the entire local Maven repository), re-run `./build.py build-dev` and check the build completes and the `~/.m2/repository/ome` is regenerated with the release version of Bio-Formats
2. re-run `./build.py build-dev` and check the build completes without waiting on artifacts retrieval
3. test a local coupled Bio-Formats/OMERO development workflow
  1. from a local Bio-Formats repository, checkout `origin/dev_5_1`, bump the version number to a non-existing published snapshot, e.g. `./tools/bump_maven_version.py 5.9.0-SNAPSHOT` then build the Bio-Formats artifacts using `mvn clean install`
  2.  check the Bio-Formats artifacts have been generated for this version under `~/.m2/repository/ome/` including both the `.jar` and the `.pom` files
  3. from the OMERO repository, modify `versions.bioformats` in `etc/omero.properties` to point at the chosen Bio-Formats version, e.g. 5.9.0-SNAPSHOT
  4. run `./build.py build-dev` and check the OMERO build completes
  5. run `dist/bin/omero import -f test.fake` and check the reported Bio-Formats version

Once signed off, a corresponding documentation PR will be opened to document this workflow.

/cc @mtbc @ximenesuk

PR #4449, opened 2016-01-29
Separate pom and xml filesystem resolvers

sbesson:
This commit follows-up 8aedc97834af6edea75de58747f5c760bce24ac3 and defines two Ivy patterns in the `user-maven` file system resolver:
- the first pattern ending in `.xml` will be used for artifacts downloaded by Ivy from a remote repository (e.g. Maven Central)
- the second pattern ending with `.pom` will be used for components built locally by Maven

Apart from checking the CI builds are green, the same testing procedure as in https://github.com/openmicroscopy/openmicroscopy/pull/4411 could be reapplied here. In a nutshell, clean the Maven repository under `~/.m2`, install the Bio-Formats artifacts locally using `mvn clean install` and then build the server using `./build.py build-dev`. In addition to `build-dev`, one should check the `release-javadoc` target also completes without failure (while it should fail without this PR). This PR will be  tested independently on the `develop` CI devspace where the OMERO build is currently failing.

### DIFF
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -57,8 +57,8 @@
           cache="local" descriptor="required">
         <artifact pattern="${local-maven2-dir}/[orgPath]/[module]/[revision]/[artifact]-[revision].[ext]"/>
         <ivy pattern="${local-maven2-dir}/[orgPath]/[module]/[revision]/[artifact]-[revision].xml"/>
-       <!-- Ivy pattern for artifacts installed locally via Maven -->
-       <ivy pattern="${local-maven2-dir}/[orgPath]/[module]/[revision]/[module]-[revision].pom"/>
+        <!-- Ivy pattern for artifacts installed locally via Maven -->
+        <ivy pattern="${local-maven2-dir}/[orgPath]/[module]/[revision]/[module]-[revision].pom"/>
       </filesystem>
 
       <!-- Remote downloads; cached to '~/.m2/repository' -->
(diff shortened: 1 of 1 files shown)


sbesson:
See http://develop-ci.docker.openmicroscopy.org:8080/job/OMERO-build/6

joshmoore:
If I've reading this right, the only change in the two blocks is another pattern for pom files? Perhaps add just another line to `user-maven`? From http://ant.apache.org/ivy/history/latest-milestone/resolver/filesystem.html, you can have `0..n` of them.

sbesson:
I hadn't thought of using the cardinality but I will give it a try.

sbesson:
Pushed and description updated.

etc/ivysettings.xml
```diff
@@ -56,7 +56,9 @@
           changingPattern=".*SNAPSHOT.*"
           cache="local" descriptor="required">
         <artifact pattern="${local-maven2-dir}/[orgPath]/[module]/[revision]/[artifact]-[revision].[ext]"/>
-        <ivy pattern="${local-maven2-dir}/[orgPath]/[module]/[revision]/[module]-[revision].pom"/>
+        <ivy pattern="${local-maven2-dir}/[orgPath]/[module]/[revision]/[artifact]-[revision].xml"/>
+       <!-- Ivy pattern for artifacts installed locally via Maven -->
```
mtbc: why indent change?

sbesson: laziness? :) Will push another commit

sbesson:
See http://develop-ci.docker.openmicroscopy.org:8080/job/OMERO-build/12/console (ran after clearing local `.m2` on the slave and running BIOFORMATS-maven).

joshmoore:
Changes looks very nice & tidy and the build is green. Merging with the hope that this will simplify end-user development as well.

PR #4446, opened 2016-01-28
Bulk annotations right panel

will-moore:
This is metadata "Bulk Annotations" PR https://github.com/openmicroscopy/openmicroscopy/pull/3875 rebased onto develop (actually it was on develop originally as https://github.com/openmicroscopy/openmicroscopy/pull/3863 but not merged).

Please see #3875 for main testing info.

NB: Also tweaked to add this under "Tables" collapsible tab. To test:
 - When "Tables" tab is not expanded, the data is not loaded.
 - When you expand the "Tables" tab, data will be loaded
 - If the "Tables" tab is expanded, selecting new Well will load new data.
 - "Info" should give tooltip (currently not working - known issue)

screen shot 2016-02-04 at 15 26 32

PR #4455, opened 2016-02-02
Populate metadata example

will-moore:
Now that the webclient will display bulk metadata for Plates #4446 (and Insight has done for some time), this gives users an example of how to use populate_metadata from within a python script to upload Table data to a plate from a csv.

To test:
 - try running the code sample, then check result in webclient (using PR https://github.com/openmicroscopy/openmicroscopy/pull/4446)

### DIFF
--- a/examples/Training/python/ROIs.py
+++ b/examples/Training/python/ROIs.py
@@ -91,8 +91,9 @@ ellipse.theT = rint(theT)
 ellipse.textValue = rstring("test-Ellipse")
 
 # Create an ROI containing 2 shapes on same plane
-# NB: OMERO.insight client doesn't support this
-# The ellipse is removed later (see below)
+# NB: OMERO.insight client doesn't support display
+# of multiple shapes on a single plane.
+# Therefore the ellipse is removed later (see below)
 createROI(image, [rect, ellipse])
 
 # create an ROI with single line shape
@@ -178,7 +179,7 @@ createROI(image, [point])
 
 
 def pointsToString(points):
-    """ Returns strange format supported by Insight """
+    """ Returns legacy format supported by Insight """
     points = ["%s,%s" % (p[0], p[1]) for p in points]
     csv = ", ".join(points)
     return "points[%s] points1[%s] points2[%s]" % (csv, csv, csv)
--- a/examples/Training/python/Tables.py
+++ b/examples/Training/python/Tables.py
@@ -12,10 +12,12 @@ FOR TRAINING PURPOSES ONLY!
 """
 
 import omero
+import os
 import omero.grid
 from omero.gateway import BlitzGateway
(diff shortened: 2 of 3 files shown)


joshmoore:
> Question: Do we want to try and provide more vanilla Python code for doing this (without using populate_metadata.py)? This will be quite a bit more complex.  However, we may still want this for Matlab or Java, (E.g. users asking how to do this from IJ)?

Options I can think of:

 * Turn this into a service (i.e. provide a cmd command)
 * Have a script which allows doing this
 * Allow use of the CLI from the client-side

will-moore:
Test fails https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/227/console
because the plate only has one Well, but test code expects 3 columns: A1, A2, A3.
```
File "/opt/hudson/workspace/OMERO-DEV-merge-training/OMERO.py/lib/python/omero/util/populate_metadata.py", line 423, in post_process
    well = well[well_column.values[i]]
KeyError: -1L
```

will-moore:
@sbesson Where can I fix the DEV-merge-training job to create a Plate with 3 columns instead of one?

@joshmoore Best solution might be to create a "Populate_Metadata.py" script, very similar to
https://github.com/ome/scripts/blob/develop/omero/import_scripts/Populate_ROI.py, which replicates the code in this PR, using a csv file that has been attached to the Plate/Screen. This would even allow regular users to run the script from Insight / Web / CLI.

sbesson:
@will-moore the training data is all generated using the script under https://github.com/sbesson/openmicroscopy/blob/develop/examples/Training/training_setup.sh. You should be adjust the SPW fake filename to match your new requirements /cc @ximenesuk

emilroz:
@will-moore this may be worth looking at: https://github.com/glencoesoftware/omero-user-scripts/blob/master/util_scripts/Populate_Metadata.py
It's not directly from a csv but from a File Annotation that holds a csv.

will-moore:
Thanks @emilroz that's perfect! @joshmoore do we want to add that script into the the main OMERO release, or leave it under glencoe scripts and point users there?

manics:
In case it's relevant there's a whole new `omero metadata` command on the way (e.g. https://github.com/openmicroscopy/openmicroscopy/pull/4037)

joshmoore:
> @joshmoore do we want to add that script into the the main OMERO release, or leave it under glencoe scripts and point users there?

 Better to copy it over, @will-moore.

jburel:
@will-moore: will you take care of the Java/matlab examples too?

simleo:
Looks good. Here is what I get on the right panel for the three wells (I've merged screenshots vertically to a single image):

tabular_data

joshmoore:
:thumbsup:

PR #4038, opened 2015-08-11
Temporary web UI workarounds

manics:
Temporary OMERO.web workarounds so that other metadata PRs can be viewed:
- Display Well annotations as well as Image annotations e.g. Bulk-MapAnnotations from #4037 (normally only Image Annotations are shown when a Well is selected).
- Linkify keys/values in MapAnnotations

--on-hold

PR #4457, opened 2016-02-04
Temporary web UI workarounds (rebased onto metadata52)

joshmoore:
This is the same as gh-4038 but rebased onto metadata52.

----

Temporary OMERO.web workarounds so that other metadata PRs can be viewed:
- Display Well annotations as well as Image annotations e.g. Bulk-MapAnnotations from #4037 (normally only Image Annotations are shown when a Well is selected).
- Linkify keys/values in MapAnnotations

--on-hold


joshmoore:
--rebased-from #4038

will-moore:
Both these features seem to work fine (testing locally).
The Well Annotations aren't removed as expected (E.g. by clicking the X on Tags) since webclient expects them to be linked to the image. But they are displayed just fine.
Linkify works too.

joshmoore:
Ah, thanks, @will-moore. So far most of the IDR hacks are concentrated on public user only, but it's good to realize that deletes in the UI will be confused.

PR #4450, opened 2016-01-29
Insight folders

dominikl:
PR for adding ROI folders support to Insight

**Test:** Check that the measurement tool still works with "plain" ROIs as well as with ROIs in folders.

PR #4458, opened 2016-02-04
Insight folders step2

dominikl:
Step 2 of the Insight integration of ROI folders:
- Fixed minor outstanding issues of #4450 (FolderData null checks and ROIService integration test)
- Fixed ROI propagation for ROIs in multiple folders /cc @pwalczysko 
- Take folder visibility into account to determine visibility of ROIs /cc @pwalczysko 
- Display multi-level hierarchy of ROI folders

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
@@ -523,6 +523,20 @@ public class ROINode
     }
 
     /**
+     * Gathers all sub nodes of this node and adds them to the provided nodes
+     * collection (the node itself will be added to the collection, too)
+     * 
+     * @param nodes
+     *            The collection to put the sub nodes into
+     */
+    public void getAllDecendants(Collection<ROINode> nodes) {
+        nodes.add(this);
+        for (MutableTreeTableNode n : this.getChildList()) {
+            gatherNodes((ROINode) n, nodes);
+        }
+    }
+    
+    /**
      * Gathers all sub nodes of a node and adds them to the provided nodes
      * collection (the node itself will be added to the collection, too)
      * 
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
@@ -25,6 +25,8 @@ package org.openmicroscopy.shoola.agents.measurement.util.roitable;
 
 
 //Java imports
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -32,6 +34,7 @@ import java.util.TreeMap;
 
 //Third-party libraries
 import org.jdesktop.swingx.treetable.MutableTreeTableNode;
+import javax.swing.tree.TreePath;
 
 //Application-internal dependencies
 import org.openmicroscopy.shoola.util.roi.figures.ROIFigure;
@@ -93,6 +96,12 @@ public class ROINode
 		
 	/** The map of the children, FolderData belonging to the ROINode. */
     private HashMap<FolderData, ROINode>           folderMap;
+    
+    /**
+     * Flag to indicate if this node is marked as visible; only used for folder
+     * nodes, which can override shape node visibility
+     */
+    private boolean visible = true;
     
 	/**
 	 * Constructor for parent node. 
@@ -393,11 +402,7 @@ public class ROINode
         } else if (isShapeNode()) {
             return ((ROIShape) getUserObject()).getFigure().isVisible();
         } else if (isFolderNode()) {
-            for (MutableTreeTableNode node : getChildList()) {
-                if (!((ROINode) node).isVisible())
-                    return false;
-            }
-            return true;
+            return this.visible;
         }
         return false;
     }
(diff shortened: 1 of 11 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
```diff
@@ -216,6 +229,19 @@ public ROINode findChild(Coord3D shapeCoord)
...
+    /**
+     * Find the child node representing the given folder
+     * 
+     * @param folder
+     *            see above.
+     * @return see above.
+     */
+    public ROINode findChild(FolderData folder) {
+        if (folderMap.containsKey(folder))
+            return folderMap.get(folder);
```
mtbc: Isn't this line the only of the three that you need?

dominikl: That's true, thanks. Was just copy/pasting the other findChild() method, without really thinking about it.

jburel:
For the methods used to retrieve folder added to the BrowserFacility, we should probably think of adding some parameters like userid even if it is not used now.
I will probably try to match when applicable the signature of the methods used to retrieve Project/Dataset etc

dominikl:
Good point. I'll fix that within this PR. Also noticed I should have some integration tests for that as well.

gusferguson:
@dominikl cc @pwalczysko 

Tested using OMERO.insight-5.3.0-m0-183-ea14612-ice35-b23-mac regions-omero.docker.openmicroscopy.org sample_user 

- Fixed ROI propagation for ROIs in multiple folders /cc @pwalczysko

Behaves as expected.

- Take folder visibility into account to determine visibility of ROIs /cc @pwalczysko

Behaves as expected.

Note the following:
- uncheck Visible for Folder: elliptical [2] - uncheck  Folder: right [2] - ROI ID 375 is not displayed (screenshot 1)
- check visible for  Folder: right [2] - ROI ID 375 becomes visible and Visible is checked for ROIs and  Folder: right [2] but remains unchecked for Folder: elliptical [2] (screenshot 2)

This is actually correct behaviour as there could be other ROIs in that folder that you want to remain visible - it just looks a little odd when there are no other invisible ROIs in folder. So no action needed.

- Display multi-level hierarchy of ROI folders

Two levels of folder hierarchy displays as expected in image ID: 4825

Good to merge from my POV - but probably a good idea for @pwalczysko to check as well.

id- 4825 p-tre_12_r3d_d3d-folders ome tiff

default

pwalczysko:
Note that in the screenshots above (as well as in my own test with Insight) we have ``0`` in place of the ``id`` for a propagated shapes.

pwalczysko:
Collapse of tree when ``Propagate`` is executed: Select a ROI inside a nested folder, click ``Propagate`` and draw the ROI over several z or t planes. Observe that in the moment you let the mouse button go, the hieararchy displayed in the ROI Manager collapses in all levels and folders -> this makes it hard to check the result of your action and do not really see the reason for it.

pwalczysko:
Re: Visibility when Folder checked or unchecked - fully agree with the lowest-level folder visibility test and anyalysis by @gusferguson . What about the higher and highest folder level though ? Atm there is no action at all if you uncheck the visibility of e.g. Folder: by (id 245). All inside this folder will remain visible. I do not mean now that I have a ready solution for this one, but shouldn't we think of this ? (also, can a higher level folder contain ROIs as well as folders ? How do we adjust the behaviour then ?)

pwalczysko:
In summary:
  - one possible bug https://github.com/openmicroscopy/openmicroscopy/pull/4458#issuecomment-181382176
  - one usability issue 
  - one conceptual question https://github.com/openmicroscopy/openmicroscopy/pull/4458#issuecomment-181389850

I would suggest lets investigate the bug, and possibility to fix the usability issue, but also happy for this  to be merged and start another Step for this.

dominikl:
With respect to https://github.com/openmicroscopy/openmicroscopy/pull/4458#issuecomment-181382176 (id == 0): Just checked with a non RoiFolders build of Insight, same there. id == 0 until the ROI changes are saved.

dominikl:
**Test**:
- Last commit, addresses https://github.com/openmicroscopy/openmicroscopy/pull/4458#issuecomment-181384147 ; make sure the table is not collapsed after a ROI is propagated through z/t.
- The two commits before just add more Folder specific methods and tests to the Gateway (to be congruent with methods like getProjects and getDatasets). Test: Code review and make sure Java integration tests pass.

mtbc:
I wonder if it would be good to display something other than "0" to mean "doesn't yet have an ID assigned" (which I suspect is what the `0` signifies).

gusferguson:
@dominikl 

Tested with OMERO.insight-5.3.0-m0-197-d42b269-ice35-b27-win  regions-omero.docker.openmicroscopy.org sample_user

Folder not collapsing after propagation of ROI.
Behaves as expected.
WRT this bug fix - good to merge.
Happy for this PR to be merged and other issues raised by @pwalczysko to be addressed in other PRs if that suits everyone best.
I am not hung up on the ID = 0 before save - don't think it will worry anyone really.

pwalczysko:
Tested on Win7. All as expected. The only open question is the conceptual issue of visibility of high level folders https://github.com/openmicroscopy/openmicroscopy/pull/4458#issuecomment-181389850 - definitely something we need to push. Maybe merge and copy the https://github.com/openmicroscopy/openmicroscopy/pull/4458#issuecomment-181389850 into the next step ? @gusferguson

dominikl:
Would be good if it could be merged. Then I can move on to step3. Can resolve the folder visibility issue in there.

gusferguson:
@dominikl - I would be happy with that - we just need to decide what else would be in step 3. I am just putting final touches to all the right-click and other folder manipulation mockup - will be posted in the issue by EOB today with luck.

sbesson:
The new Java integrations tests and all comments from @jburel on gh-4450 have been addressed here. As per @pwalczysko and @gusferguson recommendations, merging to allow future work to happen on subsequent PRs.

PR #4460, opened 2016-02-04
Make sure pixel dimensions are in microns

chris-allan:
The expectation of all downstream code is that the physical pixel
dimensions are in microns.  This commit restores the behaviour to that
expected prior to the introduction of units to OMERO and the
introduction of blind `getValue()` retrieval.

The downstream code in the OMERO.web that relies on physical dimensions being in microns:

 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/omero_image.js#L346
 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.popup.js#L131

The underlying gateway code in action as an illustration:

```
$ dist/bin/omero shell
Python 2.7.9 (default, Apr  2 2015, 15:33:21) 
Type "copyright", "credits" or "license" for more information.

IPython 2.3.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import omero

In [2]: import omero.clients

In [3]: import omero.gateway

In [4]: pixels = omero.model.PixelsI()

In [5]: pixels.physicalSizeX = omero.model.LengthI(100, omero.model.enums.UnitsLength.MICROMETER)

In [6]: image = omero.model.ImageI()

In [7]: image.addPixels(pixels)

In [8]: wrapper = omero.gateway.ImageWrapper(obj=image)

In [9]: wrapper.getPixelSizeX()
Out[9]: 100

In [10]: pixels.physicalSizeX
Out[10]: 
object #0 (::omero::model::Length)
{
    _value = 100
    _unit = MICROMETER
}

In [13]: wrapper.getPixelSizeX('MICROMETER')
Out[13]: 
object #0 (::omero::model::Length)
{
    _value = 100
    _unit = MICROMETER
}

In [14]: pixels.physicalSizeX = omero.model.LengthI(100, omero.model.enums.UnitsLength.NANOMETER)

In [15]: pixels.physicalSizeX
Out[15]: 
object #0 (::omero::model::Length)
{
    _value = 100
    _unit = NANOMETER
}

In [16]: wrapper.getPixelSizeX()
Out[16]: 100

In [17]: wrapper.getPixelSizeX('MICROMETER')
Out[17]: 
object #0 (::omero::model::Length)
{
    _value = 0.1
    _unit = MICROMETER
}
```

I would extend the marshaling test cases to cover all possible unit types for physical pixel dimensions but this will be incredibly laborious relative to the changes included here. The number of elements to mock up due to interactions with the rendering engine, etc. unfortunately creates a great deal of complexity for potential unit testing of the code in question. Happy to discuss otherwise.

There are no doubt images in our repository that have physical pixel sizes specified in a unit other than microns. These can be used to both examine the bug and this fix.

At a minimum we should assure that images with physical pixel values using the following units work, have no JavaScript errors in the console, that values displayed in the OMERO.web right hand panel correspond (even if the units are not the same) with those in the "Image Information" dialog of the full image viewer and finally that the scale bar is both displaying under the correct conditions and with the correct value:

 * millimeters
 * micrometers
 * nanometers
 * pixels

**Edit:** Added scope for testing from the comments below.

/cc @will-moore, @aleksandra-tarkowska, @knabar, @emilroz

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -163,7 +163,7 @@ def imageMarshal(image, key=None, request=None):
                 size = method('MICROMETER')
                 return size.getValue() if size else None
             except:
-                logger.warn(
+                logger.debug(
                     'Unable to convert physical pixel size to microns',
                     exc_info=True
                 )
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -159,8 +159,15 @@ def imageMarshal(image, key=None, request=None):
 
     try:
         def pixel_size_in_microns(method):
-            size = method('MICROMETER')
-            return size.getValue() if size else None
+            try:
+                size = method('MICROMETER')
+                return size.getValue() if size else None
+            except:
+                logger.warn(
+                    'Unable to convert physical pixel size to microns',
+                    exc_info=True
+                )
+                return None
 
         rv.update({
             'interpolate': interpolate,
--- a/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
@@ -158,6 +158,10 @@ def imageMarshal(image, key=None, request=None):
     interpolate = server_settings.get('interpolate_pixels', True)
 
     try:
+        def pixel_size_in_microns(method):
+            size = method('MICROMETER')
+            return size.getValue() if size else None
+
         rv.update({
             'interpolate': interpolate,
             'size': {'width': image.getSizeX(),
@@ -165,9 +169,9 @@ def imageMarshal(image, key=None, request=None):
                      'z': image.getSizeZ(),
                      't': image.getSizeT(),
                      'c': image.getSizeC()},
-            'pixel_size': {'x': image.getPixelSizeX(),
-                           'y': image.getPixelSizeY(),
-                           'z': image.getPixelSizeZ()},
+            'pixel_size': {'x': pixel_size_in_microns(image.getPixelSizeX),
+                           'y': pixel_size_in_microns(image.getPixelSizeY),
+                           'z': pixel_size_in_microns(image.getPixelSizeZ)},
             })
         if init_zoom is not None:
             rv['init_zoom'] = init_zoom


sbesson:
Thanks @chris-allan. As an edge-case, physical sizes can also be expressed in units which are not convertible into microns. Given the scope of the PR description, I assume we should have the marshaling returning `None` rather than the actual physical size value in this case.

Note that its the current state, the `ImageWrapper.getPixelSize{XYZ}` will throw an exception when handling these non-convertible units /cc @will-moore @aleksandra-tarkowska :

```
>>> pixels = omero.model.PixelsI()
>>> pixels.physicalSizeX = omero.model.LengthI(1.0, omero.model.enums.UnitsLength.PIXEL)
>>> image = omero.model.ImageI()
>>> image.addPixels(pixels)
>>> wrapper = omero.gateway.ImageWrapper(obj=image)
>>> wrapper.getPixelSizeX()
1.0
>>> wrapper.getPixelSizeX('MICROMETER')
  File "<stdin>", line 1, in <module>
  File "/opt/ome/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 6829, in wrapped
    return func(self, *args, **kwargs)
  File "/opt/ome/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 8634, in getPixelSizeX
    self._obj.getPrimaryPixels().getPhysicalSizeX(), units)
  File "/opt/ome/openmicroscopy/dist/lib/python/omero/gateway/__init__.py", line 267, in _unwrapunits
    obj = obj.__class__(obj, unitEnum)
  File "/opt/ome/openmicroscopy/dist/lib/python/omero_model_LengthI.py", line 2111, in __init__
    raise Exception(msg)
Exception: 1.0 PIXEL cannot be converted to MICROMETER
```

As above, we certainly have images in our repository allowing to test this scenario if needed.

chris-allan:
Thanks for the thoroughness, @sbesson.

As an experiment, I've made conditions where physical pixel sizes cannot be converted to microns return `None` which will be serialized to JSON as `null`. This may have adverse affects on the user interface. The scale bar in particular has various template blocks that are activated when `getPixelSizeX()` is `True`, which it will be in these cases because of the fall through to `getValue()`. The JSON data is then `null`, creating a mismatch. Just something we'll have to test.

Template block conditionals are as follows:

 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html#L321
 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html#L947

At a minimum we should assure that images with physical pixel values using the following units work, have no JavaScript errors in the console, that values displayed in the OMERO.web right hand panel correspond (even if the units are not the same) with those in the "Image Information" dialog of the full image viewer and finally that the scale bar is both displaying under the correct conditions and with the correct value:

 * micrometers
 * nanometers
 * pixels

/cc @knabar

chris-allan:
Since OMERO.figure is also likely a heavy user of the image data JSON is there the possibility that scale bars and metadata could have been displayed incorrectly as well, @will-moore?

will-moore:
@chris-allan I have tried to handle Units in OMERO.figure https://github.com/ome/figure/pull/50/files but haven't tested with PIXELS units.

will-moore:
Starting with an image that looks like this:
```
In [9]: image.getPixelSizeX()
Out[9]: 0.050801798701286316
In [10]: image.getPixelSizeY()
Out[10]: 0.050801798701286316
In [11]: image.getPixelSizeZ()
Out[11]: 0.20000000298023224
```
screen shot 2016-02-09 at 13 23 27

screen shot 2016-02-09 at 13 22 21

Set sizes to 9.8 mm, imgData json has
```
pixel_size: {
y: 9800,
x: 9800,
z: 9800
},
```

screen shot 2016-02-09 at 13 34 08
screen shot 2016-02-09 at 13 34 29
screen shot 2016-02-09 at 13 52 17

Now if I set sizes X and Y to 5.5 nm (don't change size Z) I get:
```
pixel_size: {
y: 0.0055,
x: 0.0055,
z: 9800
},
```
screen shot 2016-02-09 at 13 38 02
screen shot 2016-02-09 at 13 38 19
screen shot 2016-02-09 at 13 47 57

And finally set size X and Y to 5.5 Pixels (although I'm not sure what it means for a pixel to measure 5.5 pixels!)...
```
pixel_size: {
y: null,
x: null,
z: 9800
},
```
screen shot 2016-02-09 at 13 43 26
screen shot 2016-02-09 at 13 43 41

Scalebar is not now available in viewer.

Looks like everything is working as expected.

components/tools/OmeroWeb/omeroweb/webgateway/marshal.py
```diff
@@ -158,6 +158,17 @@ def imageMarshal(image, key=None, request=None):
     interpolate = server_settings.get('interpolate_pixels', True)
 
     try:
+        def pixel_size_in_microns(method):
+            try:
+                size = method('MICROMETER')
+                return size.getValue() if size else None
+            except:
+                logger.warn(
```
joshmoore: I'd probably think this is `debug` rather than `warn`, though without the typed exception, it's perhaps dangerous.

sbesson: Also as a potential follow-up of this PR, we could catch such conversion errors directly at the Python gateway level and let `ImageWrapper.getPixelSixe{XYZ}` return a dedicated exception e.g. `UnitConversionException` that could be consumed by consumers like OMERO.web gateway. Thoughts @aleksandra-tarkowska @will-moore ?

chris-allan: I would argue that this is the responsibility of OMERO.py, which of course the gateway would just pass along, @sbesson.

will-moore:
Just looked at OMERO.figure. What I've been doing there is adding units info to the existing json data (which until now just blindly returned size.getValue() with no units). So with sizes set to 9.8mm as above, I see:
```
pixel_size: {
symbolX: "mm",
unitY: "MILLIMETER",
unitX: "MILLIMETER",
y: 9800,
x: 9800,
symbolY: "mm",
z: 9800
},
```
which means that this is displayed incorrectly:

screen shot 2016-02-09 at 16 32 53

So if this PR is merged as is, we will need to release a corresponding fix for OMERO.figure.

Also, there is the question of how to handle the PIXELS units in the viewer above. In order to display the size in Pixels, as we do in the webclient right panel, we need the value (not null) and also need to know the units.
Do we need a breaking change to the imgData json (like OMERO.figure does) for the 5.3.0 release?

chris-allan:
I think at a minimum we would want to make an additive change like you have already made in OMERO.figure's version of the imgData JSON that passes along, at a minimum, the value and unit as encoded, yes, @will-moore.

This is the executive decision we made for `omero-marshal`:

 * https://github.com/openmicroscopy/omero-marshal/blob/master/omero_marshal/encode/__init__.py#L34

will-moore:
Fix for OMERO.figure is in https://github.com/ome/figure/pull/142

jburel:
@will-moore: we should start thinking of a bug-fixing release of figure next month

jburel:
Discussed with @will-moore today, the plan is to have a release of figure in the coming weeks.

PR #4445, opened 2016-01-27
add OMERO graph policies for Folders

mtbc:
Adjusts graph policies to incorporate Folders and adds corresponding integration tests:

* Delete2: http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration.delete/HierarchyDeleteTest/
* Chgrp2: http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration.chgrp/HierarchyMoveAndPermissionsTest/
* Chown2: http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration.chown/PermissionsTest/
* Chmod2: http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration.chmod/PermissionsTest/
* Duplicate: http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration/DuplicationTest/

PR #4462, opened 2016-02-05
bugfixes for OMERO.blitz's graphs operations

mtbc:
Additional integration tests in the regions devspace revealed a couple of bugs in Blitz's graphs code. These affect deep copy (which nobody uses) and moving mixed-ownership graphs (e.g., dataset, image, comment) to other groups (which nobody's complained about). Integration tests should all pass and nothing should regress, though the super-keen may want to try some mixed-ownership moves (especially to private or not-private) to spot regressions of merge with respect to latest. These commits make the additional integration tests in regions also pass and are rebased from #4445.

--no-rebase


joshmoore:
@mtbc: is a test possible that *would* have caught these issues?

mtbc:
Probably: while the tests that caught them are on the regions branch and include use of Folders, I could try to figure out backported tests that catch them without using Folders.

As to why we'd not already noticed: I'm guessing that our permissions testing rarely uses mixed-ownership graphs.

pwalczysko:
Tested a Move to private group case where I moved my dataset containing one image of mine and 21 images of another user. The 21 images of another user stayed in the original (RW) group, in orphaned folder. The dataset and one image of mine moved to private group as expected. Note that the same behaviour was detected in eel latest though.

mtbc:
Anything more I should do here?

pwalczysko:
From the Moves in UI perspective all seems fine.

joshmoore:
Thanks for getting it green, @mtbc. I'd propose merging tomorrow AM. I'll re-list for myself.

PR #4464, opened 2016-02-08
Disable orphaned container

aleksandra-tarkowska:
This PR replace https://github.com/openmicroscopy/openmicroscopy/pull/4431

Allowing to disable orphaned container as tweaked in https://github.com/openmicroscopy/openmicroscopy/pull/4246/

Test defaults:
 - if you tested that PR before unset ``bin/omero config set omero.client.ui.tree.orphans ``otherwise you shouldn't touch it.
 - test own and someone else data as user, group owner and admin. you should see "Orphaned container" everytime 

Test custom name via old properties:
 - set ``bin/omero config set omero.client.ui.tree.orphans.name 'Trash'`` and ``bin/omero config set omero.client.ui.tree.orphans.description "This is trash."'``
 - test if custom name works

Test disabled:
**you can only disable while viewing other users data as regular user. For admin and group owners container will be present**
 - set ``bin/omero config set omero.client.ui.tree.orphans.enabled False``
 - load your own tree, default "Orphaned Images" should be present,
 - chose someone else data, tree should not load orphaned container
 - cut image if you have permissions

cc: @dominikl Insight will need to take these to the account at some point


aleksandra-tarkowska:
@will-moore done, now with try finally and reverting to default

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
```diff
@@ -179,6 +180,12 @@ def changeActiveGroup(self, gid):  # TODO: should be moved to ISession
             return False
 
     def getOrphanedContainerSettings(self):
+        """
+        ** Deprecated ** Use :meth:`BlitzGateway.getOrphanedContainerSettings`.
+        """
+        warnings.warn(
+            "Deprecated. Use BlitzGateway.getOrphanedContainerSettings()",
```
will-moore: Here (and in the line above) you need to use BlitzGateway.getOrphanedImagesSettings()  (renamed)

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1530,6 +1531,29 @@ def getMaxPlaneSize(self):
                 int(c.getConfigValue('omero.pixeldata.max_plane_height')))
         return self._maxPlaneSize
 
+    def getOrphanedImagesSettings(self):
+        default = {
+            "enabled": True,
+            "name": "Orphaned Images",
+            "description": ("This is a virtual container with "
+                            "orphaned images.")
+        }
```
will-moore: I missed this before, but seeing as we're using getConfigDefaults() to avoid hard coding these elsewhere, seems like that makes sense here too, or is that going to have performance implications? Since we're doing this on every login, it might make sense to leave it as is until we know this is going to be quick? cc @joshmoore.

aleksandra-tarkowska: @will-moore test improvement is about to keep state of the default config to prevent unexpected failures. We never used it in gateway as you well know. Without better way of retrieving config from the server (with defaults) I don't see this being improvement. This will be extra call.

aleksandra-tarkowska: If there is such a requirement we can review all functions loading config in gateway and adjust them in another PR

components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
aleksandra-tarkowska: Renamed

will-moore:
This looks fine for me now. Good to merge.

aleksandra-tarkowska:
All tests are passing https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest/test_config_service/

etc/omero.properties
```diff
@@ -819,15 +819,19 @@ omero.client.scripts_to_ignore=/omero/figure_scripts/Movie_Figure.py,\
...
+omero.client.ui.tree.orphans.enabled=true
 # Name of the "Orphaned images" container located in client tree data manager.
-omero.client.ui.tree.orphans.name=Orphaned images
-
+omero.client.ui.tree.orphans.name=Orphaned Images
 # Description of the "Orphaned images" container.
 omero.client.ui.tree.orphans.description=This is a virtual container with orphaned images. These images are not linked anywhere. Just drag them to the selected container.
 
+# helper combining all orphans properties. Do not set manually that property.
+omero.client.ui.tree.orphans={"enabled": "${omero.client.ui.tree.orphans.enabled}", "name": "${omero.client.ui.tree.orphans.name}", "description": "${omero.client.ui.tree.orphans.description}"}
```
sbesson: If I understand correctly reading throught this PR and the previous one, this property represents a configuration node and its principal function is to define the configuration leaves attached to it. Assuming this configuration is never meant to be set, would it be worth blacklisting it from the generated configuration page in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/install/config_parser.py#L167.

aleksandra-tarkowska: @sbesson thanks for looking. We have discussed this morning with @joshmoore that is not necessary to add wrapping property. It should use ``getClientConfigValues``. I am about to open yet another PR. Hopefully this will be the final solution.

PR #4470, opened 2016-02-09
Remove command

jburel:
Use nio package to create hard link and symlink

Test the various in-place import options on several OS.

cc @joshmoore @mtbc @pwalczysko

### DIFF
--- a/components/blitz/src/ome/formats/importer/transfers/AbstractExecFileTransfer.java
+++ b/components/blitz/src/ome/formats/importer/transfers/AbstractExecFileTransfer.java
@@ -35,6 +35,7 @@ import omero.model.OriginalFile;
  * This is only useful where the command "ln -s source target" will work.
  *
  * @since 5.0
+ * @deprecated replaced by AbstractExecFileTransfer2
  */
 public abstract class AbstractExecFileTransfer extends AbstractFileTransfer {
 
--- a/components/blitz/src/ome/formats/importer/transfers/AbstractExecFileTransfer.java
+++ b/components/blitz/src/ome/formats/importer/transfers/AbstractExecFileTransfer.java
@@ -24,9 +24,6 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.StringWriter;
-import java.util.UUID;
-
-import org.apache.commons.io.FileUtils;
 
 import ome.util.checksum.ChecksumProvider;
 import omero.ServerError;
@@ -40,10 +37,6 @@ import omero.model.OriginalFile;
  * @since 5.0
  */
 public abstract class AbstractExecFileTransfer extends AbstractFileTransfer {
-
-    private static final String LINE = "\n---------------------------------------------------\n";
-
-    private static final String SEPARATOR = System.getProperty("line.separator");
 
     /**
      * "Transfer" files by soft-linking them into place. This method is likely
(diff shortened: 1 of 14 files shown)


joshmoore:
The error code handling from `exec` will need to be resurrected as exception handling. Specifically the error messages were requested so that users could figure out what was going on.

Also, since `createProcessBuilder` is protected, this would count as a breaking change since it would no longer do anything for subclasses which implement the `createProcessBuilder` method. Options I can think of:

 * keep `createProcessBuilder` but return an implementation which calls the native Java method
 * leave AbstractExecFileTransfer` as it is, and implement add new classes (`Hardlink2FileTransfer extends AbstractFileTransfer` (that may require moving some of the methods from AEFT up to AFT)

jburel:
@joshmoore: thanks I was not sure about the plan for error handling since the method already returns IOException (not handled)

jburel:
I have introduce new classes. I opted for a new ```AbstractExecFileTransfer2```.
The simplest thing might be to mark some classes deprecated this round and use 5.3 for an update.
1.6 support was dropped a while back and those classes never got cleaned up.

components/blitz/src/ome/formats/importer/transfers/AbstractExecFileTransfer.java
```diff
@@ -74,83 +68,6 @@ public String transfer(TransferState state) throws IOException, ServerError {
     }
 
     /**
-     * Build a path of the form "root.path/root.name/file.path/file.name".
-     *
-     * @param root the root directory
-     * @param ofile a path relative to the root
-     * @return the assembled path with separators suitable for the local filesystem
-     */
-    protected File getLocalLocation(OriginalFile root, OriginalFile ofile) {
```
joshmoore: I'd still avoid the removal of these `protected` methods.

joshmoore: Sorry, see them below.

joshmoore:
Excluding for the moment: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_import/TestImport/testSymlinkImport/ et al.

jburel:
Sorry I did not have a chance to run the tests. I will try to do that this pm

jburel:
Another example of testing headache (plenty with Ice 3.6)
run locally
```test/integration/clitest/test_import.py::TestImport::testSymlinkImport PASSED```

jburel:
@joshmoore @sbesson:
run the tests on breaking with this PR included
no problem see https://ci.openmicroscopy.org/view/OMERO-DEV/job/OMERO-DEV-breaking-integration-python/491/

sbesson:
@jburel: interestingly this PR was not picked in the merge job - see https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-breaking-push/623/.

jburel:
hummm

jburel:
Crap It did not pick when I remove the exclude. 
I am going to run it again

jburel:
Few errors on breaking. I will investigate when I get back since the Symlink test passes locally

jburel:
Tested locally symlink creation with both Java 1.7 and 1.8 no problem noticed

jburel:
The difference between local and server is the java version installed. I will run some tests on docker now that we have the various parameters in place.
Closing this PR for now.

PR #4474, opened 2016-02-11
Load server settings

aleksandra-tarkowska:
This replace #4431 

This PR refactor the way omero properties are loaded into the webclient. It uses ``Iconfig.getClientConfigValues`` to load them all at once

New properties:
- `omero.client.ui.tree.orphans.enabled`

New alias properties, replacing deprecated in 5.3
- `omero.client.ui.menu.dropdown.leaders.label`
- `omero.client.ui.menu.dropdown.colleagues.label`
- `omero.client.ui.menu.dropdown.everyone.label`

-----

**To test:**
- [x] orphaned images:
   - set `omero.client.ui.tree.orphans.enabled false/False/true/True` and check if regular user cannot see Orphaned Images container in other user data tree. Admin and group owner should be able to access OI anytime. A user always should access its own OI
- [x] roi limit:
   - set `omero.client.viewer.roi_limit` and check if more then limit ROIs are not loaded at once
- [x] scripts:
   - set `omero.client.scripts_to_ignore` check if right scripts are hidden from the menu
- [x] download max size:
   - set `omero.client.download_as.max_size` and check if download as jpg/png/tiff is disabled above max pixel count and allowed below.
- [x] interpolate:
   - set `omero.client.viewer.interpolate_pixels false` check if checkbox in viewer is off
- [x] zoom level:
   - set `omero.client.viewer.initial_zoom_level 2` check if big image opens on that zoom level

also:

- [x] dropdown menu:
   - set `omero.client.ui.menu.dropdown.everyone.label` & `omero.client.ui.menu.dropdown.leaders.label` & `omero.client.ui.menu.dropdown.colleagues.label`
   - set `omero.client.ui.menu.dropdown.everyone.enabled false/true` & `omero.client.ui.menu.dropdown.leaders.enabled false/true` & `omero.client.ui.menu.dropdown.colleagues.enabled false/true`
- [x] dropdown menu backwards compatibility:
   - use develop latest or 5.2.1 release server (ask for help if you wish to redeploy eel/latest)
   - set `omero.client.ui.menu.dropdown.everyone` & `omero.client.ui.menu.dropdown.leaders` & `omero.client.ui.menu.dropdown.colleagues`
    - run webclient from this branch and check if that appear correctly on UI

cc: @dominikl Insight will need to take these to the account at some point


jburel:
@aleksandra-tarkowska: the replace properties will break existing code
The strategy used in https://github.com/openmicroscopy/openmicroscopy/pull/3602 should be put in place to offer a smooth transition
You should certainly document it (under maybe contributing) so we apply the same strategy when we apply such changes.

aleksandra-tarkowska:
@jburel could you explain what is exactly breaking? I've fixed properties clean up to make sure correct labels are shown. Other then that all works fine when connecting to 5.2.1 server (seabass and eel-latest).

I am afraid #3602 is not right solution for omero.client* properties as they are server specific.

sbesson:
My understanding of dfee882 is that it addresses cross-version compatibility, i.e. an OMERO.web 5.2.2 should be able to connect to an OMERO.server 5.2.1 and properly interpret the legacy `omero.client` properties. This is definitely a safe approach and it is good to know that OMERO.web has some place for handling these changes incompatibilities (this might become even more important when we decouple it).

A reciprocal question is what should happened to these properties when an OMERO.server is upgraded to 5.2.2. Based on the PR description and the `omero.properties` changes, I understand that some of the `omero.client.*` properties are deprecated and I would also expect them to be auto-upgraded as in #3602. Or are we supporting both `omero.client.ui.menu.dropdown.leaders.label` and `omero.client.ui.menu.dropdown.leaders` properties for the lifetime of OMERO 5.2.x?

Also, is there a possibility that clients have been written using the legacy properties and would be broken by the new property semantics?

jburel:
@aleksandra-tarkowska: as @sbesson described, I was referring to the upgrade paths when pointing to #3602

aleksandra-tarkowska:
@sbesson your assumption is correct. To simplify test I cleaned the code up in https://github.com/openmicroscopy/openmicroscopy/pull/4474/files#diff-7da83cc438f0267a29c9195da7b220f1R278 Let me know if tests answer your questions.

dominikl:
I added a trello card for Insight under the 5.2.3 board: [Trello: Client properties - Insight](https://trello.com/c/wxjdWblh/4-client-properties-insight)

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1530,6 +1530,15 @@ def getMaxPlaneSize(self):
                 int(c.getConfigValue('omero.pixeldata.max_plane_height')))
         return self._maxPlaneSize
 
+    def getOmeroClientSettings(self):
+        """
+        Returns all client properties mathcing omero.client.*
```
will-moore: typo: matching

aleksandra-tarkowska:
@will-moore my appology but I had to revert max_plane_w&h as this will cause issues with backwards compatibility. We are loading them in simpleMarshal anyway, this could be reviewed separately for 5.3.

will-moore:
Testing the dropdown menu config works fine, except for old ``` {% if grp.all %}``` (already fixed above).

Also testing against 5.2.1 (eel latest) worked fine.

Slightly confusing that the settings are called "client settings" when we get them from the server ```conn. getOmeroClientSettings()``` but that the same settings are then called "server settings" in the session ```request.session.get('server_settings')```. I can kinda see the logic behind this (not a blocker), but I wonder if there's a nicer way to name things?

Also, I wonder if ```conn.getOmeroClientSettings()``` really needs the 'Omero'? Maybe ```conn.getClientSettings()``` is slightly more short and sweet?

pwalczysko:
With the new commits and the delayed build on CI, I will wait with testing for Monday.
To test:
The top part of the ``Test`` in the header of this PR (but not the ``also`` part).

aleksandra-tarkowska:
> Slightly confusing that the settings are called "client settings" when we get them from the server conn. getOmeroClientSettings() but that the same settings are then called "server settings" in the session request.session.get('server_settings'). I can kinda see the logic behind this (not a blocker), but I wonder if there's a nicer way to name things?

request.session['server_settings'] has been used for a while and it is not introduced in this PR. They point to settings loaded from the server. I'd prefer not to change it here.

will-moore:
Renaming of ```conn.getClientSettings()``` looks good and I'm fine with sicking with ```request.session['server_settings']```. Thanks.

pwalczysko:
``set omero.client.ui.tree.orphans.enabled FAlsE`` - works as expected, the two normal users cannot see the orphan folders of each other, but they can see their own. Admin and owner of group can see orphan folders of each of them. 

``set omero.client.viewer.initial_zoom_level 2`` - works as expected, the big image opens at the third zoom level (compared with cowfish on the same image)

``set omero.client.viewer.roi_limit 10`` works as expected, tested on big and non-big image, the ROIs number are highlighted, but the Show ROIs is greyed out in case N = 14, and shown in case N = 6

``set omero.client.scripts_to_ignore=/omero/figure_scripts/Split_View_Figure.py,/omero/figure_scripts/Thumbnail_Figure.py`` worked fine as well, the scripts not named in the command are now visible (as compared to cowfish, where the default setup hides some more scripts named in ``omero.properties``)

``omero.client.download_as.max_size=164360`` works as expected, the image of size ``865 x 190`` was exportable as jpeg, tiff, png, the images bigger than that were not (the three export menu options were greyed out)

``set omero.client.viewer.interpolate_pixels false`` works as expected, an image opened in Full Viewer has the checkbox ``Interpolate`` unchecked. Also checked that the checking of the ``Interpolate`` box has indeed the interpolating effect.

All 6 tests from the header of the PR (under ``To test:``, but not including ``also``) passed here.
If @will-moore okays the ``also`` part of the header and the code review, this is good to merge.

aleksandra-tarkowska:
Thanks @pwalczysko and @will-moore for testing

jburel:
Thanks @aleksandra-tarkowska for reworking the PR and all for testing.
Merging

PR #4478, opened 2016-02-15
expand convenience methods for Blitz requests in Java gateway

mtbc:
This PR greatly expands the selection of convenient helper functions for creating various kinds of request for perhaps some adjustment then submitting to Blitz. Assuming CI remains green, the main question is that of how well it trades convenience off against size: Are too many of the variants useless? Are some missing but wanted?

--no-rebase


joshmoore:
Uh, wow!

joshmoore:
(This will need gh-4477 merged in. cc @sbesson)

mtbc:
Easiest way to get an overview is just open in an IDE and look at the outline of available method signatures.

mtbc:
http://pastebin.ca/3374774 is the current generator for the class. It's neither polished nor especially readable as I figured it's not worth working much on until we have more opinion on which methods it should be generating (which this PR is intended to ask) to see what complexity could first be stripped from it or if the whole idea is bad or whatever. The code's probably quite awful from a Pythonic view too but was a good excuse for me to try to learn a little more!

mtbc:
Rebased against `develop`, now Travis passes, huh.

joshmoore:
For those interested, here's a listing for just `chgrp`:

screen shot 2016-02-17 at 12 09 05
screen shot 2016-02-17 at 12 08 55


My thoughts are:
 * in general, making these things simple is a good idea, so :+1:
 * but since no code has been updated to use them, I don't have a feeling for how many of these methods we need :-1:
 * and I do worry somewhat about the long-term support burden of having this many.

Would a builder pattern solve some/all of what you're attempting to support here?

```
new RequestBuilder().chgrp( [ targets-map | instance-list ] ).options(List<ChildOptions).dryRun().group(-1).build()
```

mtbc:
Some existing code uses some that replace the previous and I can certainly think of a few places in integration tests where I'd use some of these extras where I'm not already (i.e. I'm somehow working around their absence). I could add some examples in extra commits if that would help, though that might be easier on the regions branch so I don't end up making conflicting test changes.

The main issue I can think of with support burden is that if we change the arguments for some existing requests (I'm guessing that's pretty rare?) then that might effectively make some existing methods change meaning: for instance, introducing a new String argument just before a List<String> argument might make the previous String-to-List<String> helper change meaning.

A builder pattern would work too though be rather more verbose when several options are being offered; this way is a lot more succinct for the client.

joshmoore:
> A builder pattern would work too though be rather more verbose when several options are being offered; this way is a lot more succinct for the client.

On the flip side, when several arguments are *not* being passed, there would be substantially fewer alternatives to search through.

mtbc:
Once you know roughly which order they're in (it matches the slice docs and is fairly consistent among the different requests) you mostly don't have to search through them, you just guess: indeed, one can get a long way with just "target objects first", "request-specific stuff last".

There are of course opportunities to simply trim down the generation. I opted to omit helpers that simply call methods on what they are passed (for instance, only `long groupId` is offered, nothing's offering to take an `RLong` or an `ExperimenterGroup`) but one could also make the caller, say, wrap in `Collections.singletonList` rather than have a helper argument offer a lone `String`, that'd cut down on the helper method count too. Really I was just trying to think about how to stop code becoming annoyingly long/unreadable.

joshmoore:
Beyond any trimming that may be needed, I don't think I'm in favor of the precedence of having a snippet somewhere which is used to generate a massive class. I see the benefit of having helper methods, but this just doesn't seem like an idiom/pattern that we should be working toward.

https://github.com/mtbc/openmicroscopy/compare/expand-gateway-requests...joshmoore:expand-gateway-requests (or something similar) seems sufficiently terse while being nicely self-documenting.

mtbc:
Builders have the advantage of allowing more alternatives without fear of making the class far more enormous too. Barring contrary views from others writing client-side Java code I could look at offering some builder-based solution instead, especially if I can deprecate the current class' methods (which were certainly tedious to write by hand even for the few that it already has).

joshmoore:
I'm not sure how widely the methods are, but I could imagine both leaving them in place as well as deprecating (either with a builder or in a follow-up release if we realize the builders are working well).

PR #3896, opened 2015-06-23
Hide empty column data

will-moore:
This simply hides tabular data where there is no value to avoid showing empty column names in the right panel or the full viewer.

Before and after:
screen shot 2015-06-23 at 13 52 49

PR #3889, opened 2015-06-19
Metadata/table data url links

will-moore:
This is https://github.com/openmicroscopy/openmicroscopy/pull/3884 rebased onto metadata (on top of https://github.com/openmicroscopy/openmicroscopy/pull/3875).

========================================================================

See: https://trello.com/c/e4Jxa6My/11-adding-links-within-metadata

This PR provides 3 steps for adding urls to Table Data.
All parts of this are speculative and up for discussion.

- Add a 'url' description to the column name in your CSV file. Currently I'm using a double %% as the delimiter, but this could be changed? E.g.
    ``` Gene_Symbol %% url=http://www.ensembl.org/id/%s ```
 This will be parsed by the populate_metadata script, which will attempt to add the data as json in the Column description. E.g.
   Name: ```Gene_Symbol```
   Description: ``` '{"url": "http://www.ensembl.org/id/%s"}' ```

- Webgateway now adds the description to the json Table data, and attempts to parse the description as json. E.g. 
```
data: {
    rows: [
        [
        "ENSG00000117399",
        "CDC20",
        "44",
        ]
    ],
    descriptions: [
        "",
        {
        url: "http://www.ensembl.org/id/%s"
        },
        ""
    ],
    columns: [
        "Entity_Identifier",
        "Gene_Symbol ",
        "Entity_Number",
    ]
}
```

- Finally, if a description url is found, webclient uses it to create a link when displaying the gene name.

screen shot 2015-06-19 at 11 54 33

PR #4482, opened 2016-02-17
Hide empty column data develop

will-moore:
This is https://github.com/openmicroscopy/openmicroscopy/pull/3896 from metadata, which includes https://github.com/openmicroscopy/openmicroscopy/pull/3889
See: https://trello.com/c/e4Jxa6My/11-adding-links-within-metadata

**Discussion**:
This PR includes several IDR customisations that allow the inclusion of links in bulk annotation tables in right panel and in viewer. These could be considered a "first draft" of this workflow and use several new techniques, such as encoding json in a table column's description. We should consider if we want to go down this route for mainline OMERO?
 - Do we support this in Insight?
 - Do we update docs, code samples etc to instruct other users how to use this, or is this just for "internal IDR use" only? (ie, is this an addition to our public API)?
 - Currently this PR, as rebased from 'metadata' has no tests, and is currently on top of #4446 which is not yet merged with develop.
 - Need to decide if this should go into 5.2.2?

==================================================================

**Testing**:
This PR provides 3 steps for adding urls to Table Data.
All parts of this are speculative and up for discussion.

- Add a 'url' description to the column name in your CSV file. Currently I'm using a double %% as the delimiter, but this could be changed? E.g.
    ``` Gene_Symbol %% url=http://www.ensembl.org/id/%s ```
 This will be parsed by the populate_metadata script, which will attempt to add the data as json in the Column description. E.g.
   Name: ```Gene_Symbol```
   Description: ``` '{"url": "http://www.ensembl.org/id/%s"}' ```

- Webgateway now adds the description to the json Table data, and attempts to parse the description as json. E.g. 
```
data: {
    rows: [
        [
        "ENSG00000117399",
        "CDC20",
        "44",
        ]
    ],
    descriptions: [
        "",
        {
        url: "http://www.ensembl.org/id/%s"
        },
        ""
    ],
    columns: [
        "Entity_Identifier",
        "Gene_Symbol ",
        "Entity_Number",
    ]
}
```

- Finally, if a description url is found, webclient uses it to create a link when displaying the gene name.

screen shot 2015-06-19 at 11 54 33

 - From https://github.com/openmicroscopy/openmicroscopy/pull/3896, any empty values in the table should cause the row to be hidden.


manics:
For the IDR I think(?) we've decided on mapping the table rows to formatted map-annotations with links, so effectively the bulk-annotation provides the data and the map-annotation will be the transformed "view".

will-moore:
@manics But is IDR still using this code / functionality? Do we want this in the mainline?

manics:
I like the idea of having json column descriptions, but I'm not sure whether to effectively make the `%%` convention part of the API.

pwalczysko:
Works fine on cowfish. Tested both attaching the csv to a screen as well as to plate. The clickable link is created as expected.

jburel:
Still some points that need to be discussed, points related to "API"
Marking this PR onhold so we can time to assess possible implications. 
The feature when "generic" enough will certainly be useful by others. 
This PR won't be included in 5.2.2

jburel:
https://github.com/openmicroscopy/openmicroscopy/pull/4446 has now been merged. This PR will have to be reviewed post discussion.

will-moore:
Just rebased with origin/develop so we can see that the changes in this PR are quite small.
Time to remove the ON_HOLD label and test etc?

will-moore:
This is conflicting with #4579 - Closing to allow that to progress...

PR #4484, opened 2016-02-18
Fix typo

jburel:
see https://trac.openmicroscopy.org/ome/ticket/13161


pwalczysko:
I think the header links to a wrong ticket. Should be https://trac.openmicroscopy.org/ome/ticket/13088

pwalczysko:
Checked all three messages on cowfish. Works as expected. Ready to merge.

PR #4488, opened 2016-02-19
Set Bio-Formats version to 5.2.0-m1 for upcoming OMERO milestone

sbesson:
This should be in preparation of OMERO 5.3.0-m1

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -963,4 +963,4 @@ versions.bufr=1.1.00
 ###
 
 # Override BIo-Formats version for the model work
-versions.bioformats=5.2.0-SNAPSHOT
+versions.bioformats=5.2.0-m1


etc/omero.properties
```diff
@@ -963,4 +963,4 @@ versions.bufr=1.1.00
 ###
 
 # Override BIo-Formats version for the model work
-versions.bioformats=5.2.0-SNAPSHOT
\ No newline at end of file
+versions.bioformats=5.2.0-m1
```
mtbc: doesn't block this PR but EOL at EOF might be nice in future

mtbc:
CI is all blue.

mtbc:
Yes, build seems to be pulling in `formats-bsd-5.2.0-m1.jar`, etc. successfully. :+1: Good to merge.

PR #4489, opened 2016-02-19
Travis: fold Ice installation step

sbesson:
This should increase the readability of Travis builds since we are installing Ice 3.5.1 manually.

To test this PR, check each step in the Travis log is folded.


joshmoore:
screen shot 2016-02-21 at 20 29 35

LGTM :+1:

PR #4490, opened 2016-02-19
Fix editing of map annotation values containing single quote

will-moore:
Just found this bug (hence the last minute PR).
Fixes editng keys or values in map annotations with single quotes in them.

To test:
 - Edit keys and values in a Map Annotation, including a single quote in the text.
 - When you re-edit the text, the text field should be correctly with the text, without truncating at the quote mark.


mtbc:
Now seems to work excellently. :+1: for using the `attr` approach, probably a good pattern in general.

PR #4492, opened 2016-02-23
refactor the generation of the model object glossary

mtbc:
This PR refactors the graph path bean to offer the information required to construct the model object report. (The bean's API changes are fairly simple and minimal; it isn't worth putting much design work into it until the extended metadata functionality is understood then subsumed.) The model object report code is thus made far simpler without changing the output. Broadly, this PR should bring fewer code lines with no regressions. To generate the report used as the source for http://www.openmicroscopy.org/site/support/omero/developers/Model/EveryObject.html enter something like,
```
java -cp lib/server/\* `bin/omero config get | awk '{print"-D"$1}'` ome.services.graphs.GraphPathReport EveryObject.txt
```
which is how https://github.com/openmicroscopy/ome-documentation/pull/1408 was created.

### DIFF
--- a/components/blitz/src/omero/cmd/graphs/DuplicateI.java
+++ b/components/blitz/src/omero/cmd/graphs/DuplicateI.java
@@ -414,7 +414,7 @@ public class DuplicateI extends Duplicate implements IRequest, WrappableRequest<
                         }
                     }
                     /* process property values that do not relate to edges in the model object graph */
-                    for (final String property : graphPathBean.getSimpleProperties(superclassName)) {
+                    for (final String property : graphPathBean.getSimpleProperties(superclassName, false)) {
                         /* ignore inaccessible properties */
                         if (!graphPathBean.isPropertyAccessible(superclassName, property)) {
                             continue;
--- a/components/server/src/ome/services/graphs/GraphPathBean.java
+++ b/components/server/src/ome/services/graphs/GraphPathBean.java
@@ -566,6 +566,17 @@ public class GraphPathBean extends OnContextRefreshedEventListener {
     }
 
     /**
+     * Get the <q>simple</q> non-nested properties for the given class, not linking to other mapped classes.
+     * @param className the name of a class
+     * @return the <q>simple</q> properties of the given class; never {@code null}
+     * @deprecated instead use {@link #getSimpleProperties(String, boolean)} setting {@code isNested} to {@code false}
+     */
+    @Deprecated
+    public Set<String> getSimpleProperties(String className) {
+        return getSimpleProperties(className, false);
+    }
+
+    /**
      * Get the <q>simple</q> properties for the given class, not linking to other mapped classes.
      * @param className the name of a class
      * @param isNested if nested properties should have all components given separately with dot-notation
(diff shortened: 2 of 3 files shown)


sbesson:
Running the command listed above removes a paragraph in the `EveryObject.txt` file under ome-documentation

```
--- a/omero/developers/Model/EveryObject.txt
+++ b/omero/developers/Model/EveryObject.txt
@@ -4,14 +4,6 @@ Glossary of all OMERO Model Objects
 Overview
 --------
 
-In navigating the model objects used by the :doc:`OMERO API
-<../Modules/Api>` and in :omerocmd:`hql` it is often useful to look up
-the names of the object properties and the types of their values. This
-reference document lists every OMERO model object and their more useful
-properties, with an emphasis on enumerating every direct relationship
-among the objects.
-
-
 Reference
 --------- 
```

Looking at `GraphPathReport.java` this overview paragraph might not have been backported to the rrpot generation class in the past.

mtbc:
Ah yes, I should have mentioned, I always just paste that paragraph in manually from the previous version.

sbesson:
Is there a reason not to add it to `report()` ? Happy to open a mainline PR for that if that makes sense.

mtbc:
I just figured it's harder to review text changes, rewrap paragraphs, etc., when they're buried in code, though I suppose one could put the paragraph into some resource file. Might just want a PR against regions if you think it's a good idea, as OMERO 5.2 won't need it run any more anyway.

components/blitz/src/omero/cmd/graphs/DuplicateI.java
```diff
@@ -414,7 +414,7 @@ public Object apply(Object original) {
                         }
                     }
                     /* process property values that do not relate to edges in the model object graph */
-                    for (final String property : graphPathBean.getSimpleProperties(superclassName)) {
+                    for (final String property : graphPathBean.getSimpleProperties(superclassName, false)) {
```
sbesson: Is it worth defining a `GraphPathBean.getSimpleProperties(String)` signature which would default to `GraphPathBean.getSimpleProperties(String, false)`?

mtbc: I feared that it might be misleading: this current change makes one think about if one wanted nested or not. Depending on what extended metadata consumers need it might be worth splitting out a bunch of these methods by nested-or-not (some now conflate the two); it's mainly just this component-properties issue I have in mind with the "design work" of the PR description. (From memory, I think it mostly crops up for units and details.)

sbesson:
Overall, this works as expected and the migration of the report code to `GraphPathBean` greatly simplifies the report class. Two minor comments:
- the missing overview paragraph could be added as a follow-up to this PR
- one API decision re `GraphPathBean.getSimpleProperties(String, boolean)` vs `GraphPathBean.getDirectSimpleProperties(String)/GraphPathBean.getNestedSimpleProperties(String)`. The second form might have allow us to properly deprecate `GraphPathBean.getSimpleProperties(String)`. /cc @joshmoore

joshmoore:
No strong feelings. Whatever you guys think is the proper way to keep breakage to a minimum, :+1:

PR #4493, opened 2016-02-24
have projection tests take an initial run-through

mtbc:
Attempts to get https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.delete/DeleteProjectedImageTest/ passing more reliably. Also now cleans up the leftover images afterward.


mtbc:
Failed again through lock timeout, so I guess this approach doesn't suffice.

PR #4497, opened 2016-02-26
Insight folders Step 3.0 - Bugfixes - 2

dominikl:
The next bugfix PR, see [Trello - ROI Folders Insight Bugs](https://trello.com/c/XUI38KAp/27-roi-folders-insight-bugs)

Fixed:
- Changes of folder visibility won't affect orphaned ROIs any more.
- Expanded folders will stay expanded after the ROI table is refreshed (e. g. after a "add folders" action); additionally modified folders (e. g. if ROIs are moved to a folder) will be expanded, too (is this helpful, your thoughts @pwalczysko ? ).
- Deleted folders will immediately be removed from the table
- The column width is preserved after table refresh
- Folders created via 'Add ROIs to Folder" dialog won't have "Description" as Folder description any more.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerModel.java
@@ -97,6 +97,7 @@ import omero.gateway.model.GroupData;
 import omero.gateway.model.ImageData;
 import omero.gateway.model.PixelsData;
 import omero.gateway.model.ROIData;
+import omero.gateway.util.Pojos;
 import ome.model.units.BigResult;
 import omero.model.Length;
 import omero.model.LengthI;
@@ -1228,6 +1229,16 @@ class MeasurementViewerModel
      *            The Folders
      */
     void deleteFolders(Collection<FolderData> folders) {
+        // remove from the local data model
+        Collection<Long> ids = Pojos.extractIds(folders);
+        Iterator<FolderData> it = this.folders.iterator();
+        while(it.hasNext()) {
+            FolderData f = it.next();
+            if(ids.contains(f.getId()))
+                it.remove();
+        }
+        
+        // delete on the server
         ExperimenterData exp = (ExperimenterData) MeasurementAgent
                 .getUserDetails();
         currentSaver = new ROIFolderSaver(component, getSecurityContext(),
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerModel.java
@@ -1229,16 +1229,6 @@ class MeasurementViewerModel
      *            The Folders
      */
     void deleteFolders(Collection<FolderData> folders) {
-        // remove from the local data model
-        Collection<Long> ids = Pojos.extractIds(folders);
-        Iterator<FolderData> it = this.folders.iterator();
-        while(it.hasNext()) {
-            FolderData f = it.next();
-            if(ids.contains(f.getId()))
-                it.remove();
-        }
-        
-        // delete on the server
         ExperimenterData exp = (ExperimenterData) MeasurementAgent
                 .getUserDetails();
         currentSaver = new ROIFolderSaver(component, getSecurityContext(),
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerModel.java
@@ -97,7 +97,6 @@ import omero.gateway.model.GroupData;
 import omero.gateway.model.ImageData;
 import omero.gateway.model.PixelsData;
 import omero.gateway.model.ROIData;
-import omero.gateway.util.Pojos;
 import ome.model.units.BigResult;
 import omero.model.Length;
 import omero.model.LengthI;
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
@@ -558,6 +558,8 @@ public class ROINode
                 // therefore take a step of length 2 to get the folder node
                 ROINode folder = (ROINode) path.getPathComponent(path
                         .getPathCount() - 3);
+                if (folder.isRoot())
+                    return shape.isVisible();
                 if (b == null)
                     b = folder.isVisible();
                 else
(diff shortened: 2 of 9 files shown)


pwalczysko:
> Expanded folders will stay expanded after the ROI table is refreshed

@dominikl : Yes, helpful, mainly for the subsequent Drag and Drop behaviour, where this is crucial.

gusferguson:
@dominikl cc @pwalczysko 

Testing using OMERO.insight-5.3.0-m1-36-8785939-ice35-b60-mac regions-omero.docker.openmicroscopy.org DB: user-14 

and OMERO.insight-5.3.0-m1-36-8785939-ice35-b60-win on Windows 8.0

- Changes of folder visibility won't affect orphaned ROIs any more.

Behaves as expected

- Expanded folders will stay expanded after the ROI table is refreshed (e. g. after a "add folders" action); additionally modified folders (e. g. if ROIs are moved to a folder) will be expanded, too (is this helpful, your thoughts @pwalczysko ? ).

Behaves as expected

- The column width is preserved after table refresh

Behaves as expected

- Folders created via 'Add ROIs to Folder" dialog won't have "Description" as Folder description any more.

Behaves as expected

- Deleted folders will immediately be removed from the table

Works for untested folders - any contained ROIs orphaned as expected
For nested folders - start state - Level 1 > Level >2 Level 3 > ROI (propagated) - see screenshot 1
Delete Level 1 Folder - does not appear to delete - random other folder expands - screenshot 2
Delete Level 1 Folder again - or any other activity causing refresh - deletes all of Level 1, Level 2 and Level 3 Folders and orphans ROI - screenshot 3

This is what I expected - I would just expect Level 1 to be deleted and Level 2 remain with Level 3 and ROI remaining nested.

Same for deleting Level 2 - no apparent effect - screenshot 4 - repeated delete or other action such as adding an ROI to another folder - then deletes Level 2 and Level 3 and orphans ROIs - screenshot 4

Deleting Level 3 - same apparent failure - when refresh occurs then Level 3 only deleted and roes orphaned - screenshot 5

Also getting unrelated folders - both empty and containing ROIs - opening with refresh after activity such as adding ROI to different folder e.g. - after adding ROIs to Level 2 and Level 3 - see screenshot 6

Behaviour is the same in Windows except the random folders are not expending when unrelated activity forces a refresh.

id- 4961 62m p-tre_22_r3d_d3d dv 1

id- 4961 62m p-tre_22_r3d_d3d dv 2

id- 4961 62m p-tre_22_r3d_d3d dv 3

id- 4961 62m p-tre_22_r3d_d3d dv 4

id- 4961 62m p-tre_22_r3d_d3d dv 5

id- 4961 62m p-tre_22_r3d_d3d dv 6

pwalczysko:
Tested on Win and Mac. Agree point by point with @gusferguson observations https://github.com/openmicroscopy/openmicroscopy/pull/4497#issuecomment-193714165. 
the only difference found for me is:
   - the Delete action did not work for me neither for nested folders, nor for orphaned folders which contain ROIs (Mac and Windows)
   - coulc not detect any random expansion of folders at refresh (Mac and Windows)

dominikl:
Interesting, I fixed the "delete folder" problem in 9ed96d7e64e but somehow re-introduced the bug with a later commit (and it seems to happen more or less randomly).

dominikl:
**Re-Review**: Test the "delete folder(s)" action again.

components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementViewerModel.java
```diff
@@ -97,6 +97,7 @@
 import omero.gateway.model.ImageData;
 import omero.gateway.model.PixelsData;
 import omero.gateway.model.ROIData;
+import omero.gateway.util.Pojos;
```
mtbc: a strange change

gusferguson:
@dominikl cc @pwalczysko 

Tested using OMERO.insight-5.3.0-m1-41-e798aac-ice35-b63-mac regions-omero.docker.openmicroscopy.org DB: user-14

Delete refresh appears resolved - deletes are now showing as expected.

Still have issue with enclosed folders.
ROI in nested folder Test 1 > Test 2 > Test 3 > ROI 1
Delete folder Test 1 - all folders deleted and ROI 1 orphaned
Delete folder Test 2 - Test 2 and Test 3 folders deleted and ROI 1 orphaned

Other issue noted:
The ROI folder count is now showing the count of ROIs in it or subfolders - screenshot 1
The ROI Folder count should be show the number of subfolders.

id- 4908 20110209_region1_1505_atp zvi

pwalczysko:
Agree with the behaviour described by @gusferguson .
Tested on Mac and Win7 laptops. 

    - when I delete a top folder, all the subfolders get deleted and the ROI gets orphaned as described in https://github.com/openmicroscopy/openmicroscopy/pull/4497#issuecomment-194803637 -> I do not have much problems with that (except that it is slightly inconsistent, treating ROIs and folders differently), but the behaviour of subfolders is one which is expected as compared with the File-tree example -> maybe needs more thoughts for a follow-up PR ? (I can see that this might not be welcomed if the user wants to delete JUST the top folder without having to salvage a complicated hierarchy first)
    - the showing of ROI numbers instead of subfolder numbers is surprising for the user, but we actually need both informations (maybe something for filtering scoping ?)

dominikl:
- Delete folder will delete sub-folders: I thought that's the expected behavior. That's what the current graph implementation does. If that's not intended I have to orphan the subfolders before triggering the delete action. Can do that of course.
- ROI folder count shows number of ROIs: @pwalczysko suggested that; more useful than number of sub folders, because otherwise one has no idea which folders contain ROIs and how many of them. Ideally both counts should be shown in some kind of non-confusing way. Ideas?

pwalczysko:
Would recommend to merge and bring the 2 issues (which are of scoping/conceptual character) https://github.com/openmicroscopy/openmicroscopy/pull/4497#issuecomment-194810730  to scoping issues ? @dominikl @gusferguson

dominikl:
:+1: Yes, would be easier to do that in a separate PR.

gusferguson:
I'm good with whatever works best.

gusferguson:
I don't agree with Petr - I see it as a violation of the accepted paradigm that a count in a folder represents enclosed objects - most usually the next level of the hierarchy down, but definitely not as it is now.

Who is going to open  a new issue for it?

mtbc:
deleting sub-folders -- also can use `excludeType={'Folder'}` on the delete to avoid that.

dominikl:
Thanks @mtbc . I'd do this in follow-up PR (to keep the PRs within a manageable size), if there are no objections? In the follow-up PR I can also add the warning dialog (with option to delete the whole branch, ie "contents"), like stated in https://github.com/openmicroscopy/design/issues/15 .

pwalczysko:
@gusferguson 
> Who is going to open a new issue for it?

See
https://github.com/openmicroscopy/design/issues/25

PR #3966, opened 2015-07-14
improve graphs error detection for dry runs

mtbc:
In https://github.com/openmicroscopy/openmicroscopy/pull/3879#issuecomment-114064985 @will-moore noted that with a graph operation's `dryRun=True` he wasn't seeing an error that he would have if he'd gone ahead with the operation for real. This PR should fix a high fraction of such situations and also reduce the incidence of the confusing "cannot null" exceptions.

PR #4501, opened 2016-02-29
slightly simplify graph policy rules

mtbc:
Enabled by #3966 uses the `OUTSIDE` graph node state less now that error-checking order does not need to be worked around. CI should remain blue and `Chgrp2`, `Delete2` should not show regressions.

### DIFF
--- a/components/blitz/resources/ome/services/graph-rules/blitz-chgrp-rules.xml
+++ b/components/blitz/resources/ome/services/graph-rules/blitz-chgrp-rules.xml
@@ -695,8 +695,8 @@
         <!-- If a well is moved, check if the fields' images can be moved; if so, move the fields. -->
 
         <bean parent="graphPolicyRule" p:matches="WS:WellSample.well = [I], WS.image = I:[E]{i}" p:changes = "I:{r}"/>
-        <bean parent="graphPolicyRule" p:matches="WS:WellSample.well = [I], WS.image = I:[E]{o}" p:changes = "WS:[I]"/>
-        <bean parent="graphPolicyRule" p:matches="WS:WellSample.well = [I], WS.image = I:[I]" p:changes = "WS:[I]"/>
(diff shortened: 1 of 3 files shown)


pwalczysko:
Delete and Chgrp actions in Insight on regions branch work as expected. Job is blue.
(not for this PR @dominikl : My regions which are inside folders do not get moved when I change group of the image - @mtbc is telling me this is due to the policy not being implemented in Insight yet.)

Ready to merge.

mtbc:
`HierarchyMoveAndPermissionsTest.testMoveRois` illustrates the current behavior. Regarding https://trello.com/c/3EQc3guS/14-folder-related-api-methods I could certainly work on https://trello.com/c/5WFgEmPL/581-query-omero-objects-by-graph-spec for one of our milestones.

mtbc:
To clarify: an option of `includeType={'Roi'}` will bring the ROIs but not their folders. A "list model objects containing given object" request could give you all the containing folder IDs recursively in one go if you wanted to include them in the move but of course it gets hairy if as in https://github.com/openmicroscopy/design/issues/2 ROIs from different images may be in the same folder.

joshmoore:
@mtbc is further discussion needed here? Or would it be worth getting the proposed gateway change to see if the behavior is as intended?

mtbc:
No more discussion needed here, no surprises. The integration testing already checks all the combinations (moving image and/or folders, with or without ROIs).

It's a bit like the situation with deleting orphaned tags: the current default makes the other alternatives easily available. I'm currently working on a `FindParents` request to provide all the Folders containing a set of ROIs, whether directly or indirectly, which should help.

mtbc:
> My regions which are inside folders do not get moved when I change group of the image ...

Note that this PR doesn't change any of that behavior anyway, it's simply a tidying up of existing rules.

sbesson:
All understood. Merging based on the state of the tests. We can (re)discuss the gateway graph rules tomorrow.

PR #4503, opened 2016-02-29
Bump Bio-Formats to 5.2.0-SNAPSHOT for the next development cycle

sbesson:
This should allow Bio-Formats/OMERO coupled PRs to be reopened against this line. /cc @mtbc


mtbc:
Good to merge.

PR #4504, opened 2016-02-29
try harder to guess MIME type

mtbc:
Has OMERO's Java code use file content to guess MIME type before falling back to `application/octet-stream`. To test, try attaching files from your local system *using Insight* and check their resulting "mimetype" property. For instance, for an attachment with Annotation ID 123, from the CLI one can do,
```
$ bin/omero hql "SELECT file.mimetype FROM FileAnnotation WHERE id = 123"
```
Generally you should find that sensible things happen for common file extensions like ".pdf". However, where you have files with unknown or no extension, named "myfile" or "foo.bar.wibble" or whatever, with this PR you should now find that they still tend to get a sensible MIME type guessed according to their content: PDF, ASCII text, whatever. Also try a really random file (e.g., `head -c 64 </dev/urandom >/tmp/foo`) to make sure that `application/octet-stream` does remain the fallback if the file content is unrecognizable.

Assists https://trac.openmicroscopy.org/ome/ticket/12041.

--no-rebase

### DIFF
--- a/components/blitz/src/omero/client.java
+++ b/components/blitz/src/omero/client.java
@@ -26,6 +26,7 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -1069,7 +1070,16 @@ public class client {
         }
 
         if (fileObject.getMimetype() == null) {
-            fileObject.setMimetype(rstring("application/octet-stream"));
+            String mimeType = null;
+            try {
+                mimeType = Files.probeContentType(file.toPath());
+            } catch (IOException | SecurityException e) {
+                /* can't guess */
+            }
+            if (mimeType == null) {
+                mimeType = "application/octet-stream";
+            }
+            fileObject.setMimetype(rstring(mimeType));
         }
 
         IUpdatePrx up = sf.getUpdateService();
--- a/components/blitz/src/omero/gateway/model/FileAnnotationData.java
+++ b/components/blitz/src/omero/gateway/model/FileAnnotationData.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2008 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


pwalczysko:
Attached ``gunicorn.err`` to image ID 6711 user-7 eel merge. The annotation ID is 37001.  This is a text file (an error log).

Got an output  from CLI:
```
 # | Col1                     
---+--------------------------
 0 | application/octet-stream 
(1 row)
```

pwalczysko:
Further to the previous comment https://github.com/openmicroscopy/openmicroscopy/pull/4504#issuecomment-190731432: When I attach a text (piece of xml) which is named ``new_xml`` (no extension) I get correctly
```
 # | Col1     
---+----------
 0 | text/xml 
(1 row)


```
But if I rename it adding extension ``.wobble`` I get again
```
# | Col1                     
---+--------------------------
 0 | application/octet-stream 
(1 row)
```

mtbc:
If I reupload that `gunicorn.err` to that same image I get `text/plain`. (Take a look at it now.) I can also get that from a `petr.wobble` file. This probably depends on how good the client-side Java on your local system is: mine is probably better at interacting with the system-provided content guessing than yours. Sorry about that!

No regressions at least? :smiley:

mtbc:
(Interesting it did still pick up the XML though, cool.)

mtbc:
I wonder if Mac users are mostly out of luck: http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7133484

mtbc:
Ha, there may also be Windows issues -- http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8080369 -- it may be some time yet before this PR delivers significant improvement depending on what work is done on the underlying system library! For the meantime we may be in a strange mirror world in which Linux gets the best support.

pwalczysko:
Yes, re https://github.com/openmicroscopy/openmicroscopy/pull/4504#issuecomment-190789718 - when tried the attachment on Win 7 OMERO.insight client, I got again ``octet-stream`` response form the hql command.

manics:
Could you bundle a mime.types file in a jar? https://docs.oracle.com/javase/7/docs/api/javax/activation/MimetypesFileTypeMap.html

pwalczysko:
Also on eel latest when using Mac Insight (from the latest build) I am getting the ``octet-stream`` on the ``new_xml.wobble`` file.

   - No regression here

In summary: Ready to merge after the somment of @manics is answered.

mtbc:
@manics: Unfortunately that wouldn't help with `new_xml.wobble`, etc.; our existing code actually already does reasonably at guessing by extension. The missing part on other systems seems to be Java interacting with libraries that can actually use the file content as well as the name. If we wanted to do well at that everywhere we could use something like Apache Tika but I figured that this way we can at least do better than present on systems where Java does offer content analysis (which might have coverage increasing over time) while not making any codebase any harder to maintain.

joshmoore:
:+1: from me and @manics

PR #4505, opened 2016-02-29
add request builders to Java gateway

mtbc:
Adds request builders to the Java gateway utility class `Requests`. Testing is largely via CI already using these builders in the new implementation (hence plenty of examples) of the now-deprecated previous helper methods for creating graph requests. (Deprecation due to that they are a pain to create and maintain and already falling behind the requests now offered.) Do feel free to try out the builders yourself too and let me know if you want working examples for disk usage and duplicate (helpers newly offered in this PR so without automated indirect testing).

Not changing much of our existing code to use the new builders until `dev_5_2` is branched to avoid conflicts with `regions`.

--no-rebase

### DIFF
--- a/components/blitz/src/omero/gateway/util/Requests.java
+++ b/components/blitz/src/omero/gateway/util/Requests.java
@@ -56,7 +56,9 @@ public class Requests {
      * @param includeType the child type to include in the request's operation
      * @param excludeType the child type to exclude from the request's operation
      * @return the new instance
+     * @deprecated use {@link Requests.ChildOptionBuilder} from {@link #option()}, see this method for an example
      */
+    @Deprecated
     public static ChildOption option(String includeType, String excludeType) {
         ChildOptionBuilder builder = option();
         if (includeType != null) {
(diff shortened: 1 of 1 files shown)


joshmoore:
Few quick comments based on viewing https://patch-diff.githubusercontent.com/raw/openmicroscopy/openmicroscopy/pull/4505.patch since GH doesn't want to show the diff:

 * The examples in the upgraded methods look generally good.
 * I might slightly favor `s/id/ids/` where the method takes a plural, but I don't know the impact of that.
 * Unit tests would be welcome if you think there might be odd edge cases if one doesn't call certain methods, etc.

mtbc:
Huh, my previous response here disappeared into the ether, bah. Basically: I went for `id` to try to reduce the number of different method names and because one *can* pass a singular ID, especially for the varargs variants, and it's shorter. I couldn't think of any odd edge cases that, if time for test-writing were available, it wouldn't be worth adding tests to different classes instead, especially as this will get more tested anyway as I move more existing code to use the builders after the `dev_5_2` branch, but of course in principle if I don't feel like thinking too hard some day then unit tests could certainly be added.

dominikl:
Looks fine to me. When this is merged I can update the `Requests.*` usage in Insight accordingly.

mtbc:
Thank you! Though to avoid conflicts of course do feel free to avoid touching anywhere on Insight on `develop` that you might separately be touching on `regions`.

PR #4508, opened 2016-03-02
Extend obj get command

ximenesuk:
This was a quick fix for the issue raised by @mtbc in https://trello.com/c/sBjOWOWT/588-omero-obj-follow-up  If the requested field has an `id` rather than a `val` return the result in the form `Klass:id`. It now extends the command to handle fields that have units, returning the value and the units. It also adds the more general form to get all reasonable OMERO fields and values for a given object. 

Testing should use fields that have ids. For instance import an image from the command line and then use the returned Pixels id to get linked object ids:
```
$ bin/omero obj get Pixels:101 image
Image:101
$ bin/omero obj get Image:101 fileset
Fileset:51
```

To test a field with units import a fake, e.g.:
```
$ omero import ~/Work/images/fakes/image\&physicalSizeX\=123.fake
Imported pixels:
251
Other imported objects:
Fileset:251
Image:251
$ omero obj get Pixels:251 physicalSizeX
123.0 MICROMETER
```

existing value fields should still work:
```
$ bin/omero obj get Image:101 name
IAGFP-Noc01_R3D.dv
```

Test getting all fields:
```
$ bin/omero obj new Project name=foo description=bar
Project:251
$ bin/omero obj get Project:251
description: bar
id: 251
name: foo
version:

$ omero obj get Pixels:251
dimensionOrder: DimensionOrder:1
id: 251
image: Image:251
methodology:
physicalSizeX: 123.0 MICROMETER
physicalSizeY:
physicalSizeZ:
pixelsType: PixelsType:5
relatedTo:
sha1: ecb315eecb4feab2a177494b656e1653cb6d76f1
significantBits: 8
sizeC: 1
sizeT: 1
sizeX: 512
sizeY: 512
sizeZ: 1
timeIncrement:
version:
waveIncrement:
waveStart:
```

Test list fields, whole and by index value:
```
$ omero obj get ExperimenterGroup:260 config
[(name1,value1),(name2,value2),(name3,value3)]
$ omero obj get ExperimenterGroup:1 config
[]
$ omero obj list-get ExperimenterGroup:260 config 1
(name2,value2)
$ omero obj list-get ExperimenterGroup:260 config -1
(name3,value3)
$ omero obj list-get ExperimenterGroup:260 config 3
Error: field 'config[3]' for ExperimenterGroup:260, list index out of range
$ omero obj list-get ExperimenterGroup:260 name 1
Field 'name' for ExperimenterGroup:260 is not a list
```

Integration tests have been added to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_obj.py the whole of which should still pass.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -358,9 +358,14 @@ class ObjGetTxAction(NonFieldTxAction):
         else:
             try:
                 proxy = current.val
-            except AttributeError, ae:
-                ctx.die(336, "Error: field '%s' for %s:%s : %s" % (
-                    field, self.kls, self.obj.id.val, ae.message))
+            except AttributeError:
+                try:
+                    objId = current.id.val
+                    klass = current.__class__.__name__.rstrip("I")
+                    proxy = klass + ":" + str(objId)
+                except AttributeError, ae:
+                    ctx.die(336, "Error: field '%s' for %s:%s : %s" % (
+                        field, self.kls, self.obj.id.val, ae.message))
 
         self.tx_state.set_value(proxy, dest=self.tx_cmd.dest)
 
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -391,16 +391,15 @@ class ObjGetTxAction(NonFieldTxAction):
                 elif hasattr(current, "_value") and hasattr(current, "_unit"):
                     proxy = str(current._value) + " " + str(current._unit)
                 elif isinstance(current, list):
-                    proxy = "["
-                    for item in current:
-                        if isinstance(item, NamedValue):
-                            proxy += ("(" + str(item.name) + ","
-                                      + str(item.value) + ")")
+                    if len(current) == 0:
+                        proxy = ""
+                    else:
+                        if isinstance(current[0], NamedValue):
+                            proxy = ",".join(
+                                ["(" + str(i.name) + "," + str(i.value) + ")"
+                                    for i in current])
                         else:
-                            proxy += str(item)
-                        proxy += ","
-                    proxy = proxy.rstrip(",")
-                    proxy += "]"
+                            proxy = ",".join([str(i) for i in current])
                 else:
                     raise AttributeError(
                         "Error: field '%s' for %s:%s : no val, id or value" % (
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -357,18 +357,21 @@ class ObjGetTxAction(NonFieldTxAction):
             proxy = ""
         else:
             try:
-                if hasattr(current, "_value") and hasattr(current, "_unit"):
-                    proxy = str(current._value) + " " + str(current._unit)
-                else:
+                if hasattr(current, "val"):
                     proxy = current.val
-            except AttributeError:
-                try:
+                elif hasattr(current, "id"):
                     objId = current.id.val
                     klass = current.__class__.__name__.rstrip("I")
                     proxy = klass + ":" + str(objId)
-                except AttributeError, ae:
-                    ctx.die(336, "Error: field '%s' for %s:%s : %s" % (
-                        field, self.kls, self.obj.id.val, ae.message))
+                elif hasattr(current, "_value") and hasattr(current, "_unit"):
+                    proxy = str(current._value) + " " + str(current._unit)
+                else:
+                    raise AttributeError(
+                        "Error: field '%s' for %s:%s : no val, id or value" % (
+                            field, self.kls, self.obj.id.val))
+            except AttributeError, ae:
+                ctx.die(336, "Error: field '%s' for %s:%s : %s" % (
+                    field, self.kls, self.obj.id.val, ae.message))
 
         self.tx_state.set_value(proxy, dest=self.tx_cmd.dest)
 
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -412,6 +412,42 @@ class ObjGetTxAction(NonFieldTxAction):
         return proxy
 
 
+class ListGetTxAction(NonFieldTxAction):
+
+    def on_go(self, ctx, args):
+
+        if len(self.tx_cmd.arg_list) != 4:
+            ctx.die(335, "usage: list-get OBJ FIELD INDEX")
+
+        field = self.tx_cmd.arg_list[2]
+        try:
+            current = getattr(self.obj, field)
+        except AttributeError:
+            ctx.die(336, "Unknown field '%s' for %s:%s" % (
+                field, self.kls, self.obj.id.val))
+
+        index = int(self.tx_cmd.arg_list[3])
+        if current is None:
+            proxy = ""
+        else:
+            if isinstance(current, list):
+                try:
+                    item = current[index]
+                    if isinstance(item, NamedValue):
+                        proxy = ("(" + str(item.name) + ","
+                                 + str(item.value) + ")")
+                    else:
+                        proxy = str(item)
+                except IndexError, ie:
+                    ctx.die(336, "Error: field '%s[%s]' for %s:%s, %s" % (
+                        field, index, self.kls, self.obj.id.val, ie.message))
+            else:
+                ctx.die(336, "Field '%s' for %s:%s is not a list" % (
+                    field, self.kls, self.obj.id.val))
+
+        self.tx_state.set_value(proxy, dest=self.tx_cmd.dest)
+
+
 class TxState(object):
 
     def __init__(self, ctx):
@@ -488,7 +524,8 @@ Bash examples:
         parser.add_argument(
             "command", nargs="?",
             choices=("new", "update", "null",
-                     "map-get", "map-set", "get"),
+                     "map-get", "map-set",
+                     "get", "list-get"),
             help="operation to be performed")
         parser.add_argument(
             "Class", nargs="?",
@@ -540,6 +577,8 @@ Bash examples:
             return NullTxAction(tx_state, tx_cmd)
         elif tx_cmd.action == "get":
             return ObjGetTxAction(tx_state, tx_cmd)
+        elif tx_cmd.action == "list-get":
+            return ListGetTxAction(tx_state, tx_cmd)
         else:
             raise self.ctx.die(100, "Unknown command: %s" % tx_cmd)
 
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -357,7 +357,10 @@ class ObjGetTxAction(NonFieldTxAction):
             proxy = ""
         else:
             try:
-                proxy = current.val
+                if hasattr(current, "_value") and hasattr(current, "_unit"):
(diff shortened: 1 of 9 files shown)


mtbc:
For `test_obj.py`,
```
========================== 12 passed in 4.15 seconds ===========================
```

mtbc:
Works excellently,
```
$ omero obj get Pixels:2 image
Using session 94cdfdd3-cf92-4406-b464-e576729941da (user-2@localhost:4064). Idle timeout: 10 min. Current group: private-1
Image:2
```

mtbc:
Less thrilling was,
```
$ omero obj get Pixels:2 physicalSizeX
Using session 94cdfdd3-cf92-4406-b464-e576729941da (user-2@localhost:4064). Idle timeout: 10 min. Current group: private-1
REFERENCEFRAME
```
```
# SELECT physicalsizex, physicalsizexunit FROM pixels WHERE id = 2;
 physicalsizex | physicalsizexunit 
---------------+-------------------
         0.253 | µm
```
but that's not this PR's fault so **good to merge**.

manics:
A completely unrelated error:

```
$ omero-DEV-merge obj get nonexistent:82
...
  File "/home/hudson/OMERO-DEV-merge-deploy/lib/python/omero/plugins/obj.py", line 257, in go
    self.obj, self.kls = self.instance(ctx)
  File "/home/hudson/OMERO-DEV-merge-deploy/lib/python/omero/plugins/obj.py", line 169, in instance
    ctx.die(102, "No class named '%s'" % self.class_name())
TypeError: class_name() takes exactly 2 arguments (1 given)
```

ximenesuk:
Thanks both. I'll try to tidy up the error @manics is seeing in this PR.

ximenesuk:
Looking into @mtbc's comment it seems that this Length has an implicit attribute `val` which returns `REFERENCEFRAME` - hence the value output but `obj get`. However, the explicit `_unit` returns, the correct unit, `MICROMETER` with `_value` having the right numeric value. Why should `val` exist? Why should it return a different enum (notably the last one defined for Length.

jburel:
restarted travis

joshmoore:
@ximenesuk, re https://github.com/openmicroscopy/openmicroscopy/pull/4508#issuecomment-191836212 -- this looks like a bug in the code generation:

```
class LengthI(_omero_model.Length, UnitBase):

    try:
        UNIT_VALUES = sorted(UnitsLength._enumerators.values())
    except:
        # TODO: this occurs on Ice 3.4 and can be removed
        # once it has been dropped.
        UNIT_VALUES = [x for x in sorted(UnitsLength._names)]
        UNIT_VALUES = [getattr(UnitsLength, x) for x in UNIT_VALUES]
    CONVERSIONS = dict()
    for val in UNIT_VALUES:            ### <---- This line defines a class variable

```

The `val` item in the static block will needed to be `del val`'d. I can give you a commit for this for this PR if that will help.

ximenesuk:
@joshmoore my code https://github.com/ximenesuk/openmicroscopy/commit/f0b048c963da74726de3c4352bae3ba568da5726 gets around this bug but I'd be happy to add you commit here.

joshmoore:
See: https://github.com/ome/ome-units/pull/5 which generated 1fe76e4bc482c683676d255807cc0d02187b933b which you can cherry-pick

ximenesuk:
@joshmoore thanks!

ximenesuk:
The condition ordering in https://github.com/ximenesuk/openmicroscopy/commit/de20fa1dd0b260e878ff32b52cfd310c910dad40 with effectively tests https://github.com/openmicroscopy/openmicroscopy/commit/1fe76e4bc482c683676d255807cc0d02187b933b and so indirectly tests https://github.com/ome/ome-units/pull/5

joshmoore:
```
/opt/ome0/dist $ touch foo.fake
/opt/ome0/dist $ echo 'physicalSizeX=1 nm' >> foo.fake.ini
/opt/ome0/dist $ PID=$(bin/omero import foo.fake)
/opt/ome0/dist $ bin/omero -q obj get Pixels:$PID | grep SizeX
physicalSizeX: 1.0 NANOMETER
/opt/ome0/dist $ bin/omero -q obj get Pixels:$PID physicalSizeX
```

:+1:

Slight comments re: full object format:

 * should the format match that expected by the `set` command? `image=Image:251`
 * (possibly larger problem) how to differentiate between nulls and empty strings?
 * ...units... The `1.0 NANOMETER` string *can't* be echo'd into foo.fake.ini. Doing so produces no error but is silently ignored. Not sure what this means on what rules we need to put into place regarding what we show users.

ximenesuk:
@joshmoore :

> should the format match that expected by the set command? image=Image:251

commit pushed.

> (possibly larger problem) how to differentiate between nulls and empty strings?

is this necessary for this level of reporting?

> ...units... The 1.0 NANOMETER string can't be echo'd into foo.fake.ini. Doing so produces no error but is silently ignored. Not sure what this means on what rules we need to put into place regarding what we show users.

should the fake reader be able to cope with these unit enum strings? Or should this command try to output a usable format?

I guess for the latter two it depends on how you see the output being used.

joshmoore:
Longer/longest term I see us having a `bin/omero obj edit` where you can edit output similar to `obj get` to modify multiple fields in one go. This is certainly what I was looking to do with the `bin/omero render` plugin for setting viewing options. However, that's definitely beyond this PR, just wanted to point to one possible future use.

Re: the units -- probably a wider discussion with @sbesson, @melissalinkert, et al.

ximenesuk:
The last commits close off the card https://trello.com/c/sBjOWOWT/23-omero-obj-follow-up by handling list fields. In `obj get` the whole list is output, in `obj list-get Object:id field index` the item at `index` is shown. Lists of NamedValue pairs (MapAnnotation and ExperimentGroup config) and lists of strings (Namespace keywords) are handled.

A question from me is the format of the results. For `obj get` the form of the string is that of a Python list of items (tuples for pairs), including `[]` for an empty list. What form would be most useful? I couldn't find any examples of setting lists using the `obj` command that could be reflected in the output.

joshmoore:
* `-h` could benefit from your description on this PR (it still doesn't mention querying objects)
 * list/map usage looks good (see below)
 * regarding the formatting, ideal in order to implement `list-set` would be a form that we could reverse: `NamedValue(name, value)` but that seems overly complicated. `obj new` isn't an option though since `NamedValue`s don't have IDs so can't exist alone. It might be that there is no good answer here (at the moment). I don't find the pythonic representation horrifying, but it might be possible (*cough* @manics *cough*) to come up with something with enough commas, brackets, and quotes that it's unreadable.
 
#### Test run ####

```
[jamoore@ome-c6100-1 dist]$ bin/omero obj map-get ExperimenterGroup:819 config josh
None
[jamoore@ome-c6100-1 dist]$ bin/omero obj map-set ExperimenterGroup:819 config josh test
ExperimenterGroup:819
[jamoore@ome-c6100-1 dist]$ bin/omero obj map-get ExperimenterGroup:819 config josh
test
[jamoore@ome-c6100-1 dist]$ bin/omero obj get ExperimenterGroup:819
config=[(josh,test)]
description=
id=819
ldap=False
name=FD5BDB1EX1C63X4F59XA388X1A81F9F7D1A6
[jamoore@ome-c6100-1 dist]$ bin/omero obj list-get ExperimenterGroup:819 config 0
(josh,test)
```

p.s. thinking about it, is this a valid use case that we should shoot for? (i.e. pipe `get` into `update` and have it "just work"?

```
[jamoore@ome-c6100-1 dist]$ bin/omero -q obj update ExperimenterGroup:819 $(bin/omero -q obj get ExperimenterGroup:819 | xargs | perl -pe 's/test/foo/')
    rv = omero.cli.argv()
  File "/opt/hudson/workspace/OMERO-DEV-merge-deploy/src/dist/lib/python/omero/cli.py", line 1438, in argv
    cli.invoke(args[1:])
  File "/opt/hudson/workspace/OMERO-DEV-merge-deploy/src/dist/lib/python/omero/cli.py", line 952, in invoke
  File "/opt/hudson/workspace/OMERO-DEV-merge-deploy/src/dist/lib/python/omero/cli.py", line 1029, in onecmd
  File "/opt/hudson/workspace/OMERO-DEV-merge-deploy/src/dist/lib/python/omero/cli.py", line 1111, in execute
    args.func(args)
  File "/opt/hudson/workspace/OMERO-DEV-merge-deploy/src/dist/lib/python/omero/plugins/obj.py", line 559, in process
    action.go(self.ctx, args)
  File "/opt/hudson/workspace/OMERO-DEV-merge-deploy/src/dist/lib/python/omero/plugins/obj.py", line 233, in go
    setter(obj)
  File "/opt/hudson/workspace/OMERO-DEV-merge-deploy/src/dist/lib/python/omero/plugins/obj.py", line 67, in __call__
    return getattr(obj, self.setter)(self.value, wrap=True)

```

manics:
Sounds like we need two plugins, a simple one like this for querying and setting single scalar fields, and  one for more complex objects/fields: `omero obj [--yaml | --json] {get,set}`

joshmoore:
That might be a nice way to draw the line on this. `omero-marshal` is likely a good place to look for what representation(s) should be supported.

ximenesuk:
Thanks @joshmoore. I have dropped the `[]` they are not really needed and an empty list now reports the same as an empty string and a null (for good or bad). I think having the NamedValues as tuples is probably safe unless we introduce any other tuple forms. I'll update the help and then hopefully get this PR merged and leave the bigger questions for a card.

joshmoore:
LGTM :thumbsup: (Do you have a follow on card?)

ximenesuk:
@joshmoore I don't have a specific follow-on card yet.

ximenesuk:
See https://trello.com/c/7qWi8uHU/643-further-extensions-to-omero-obj I've added a couple of items feel free to add more! /cc @manics @joshmoore

ximenesuk:
Interestingly we do the same import at the method level elsewhere in this file https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/plugins/obj.py#L286 (I hadn't seen this until grepping the files for imports from `omero.model`)

The file has a whole nest of imports and different levels. I can use the same pattern as above but a big concern is why, if this is the cause, it is not picked up my the CI systems. Why it works with one system and not another. /cc @sbesson @joshmoore 

@aleksandra-tarkowska I'll open a PR for you to test tomorrow using your workflow.

joshmoore:
@ximenesuk : https://github.com/ximenesuk/openmicroscopy/commit/c7456993c11c5b0bcce75f61e639ff82cb23dfc4 has some explanation.

joshmoore:
I still haven't been able to reproduce, so I don't yet know how to pick it up. I'd *think* one way would be to make a failing plugin import a failure. i.e. `export OMERO_STRICT_PLUGINS=1` and then `bin/omero config set` would have failed rather than just print the warning. (Note: this is similar to https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.2/components/tools/OmeroPy/src/omero/plugins/metadata.py#L503)

Regarding the import, https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.2/components/tools/OmeroPy/src/omero/min.py#L22 is probably what is needed. I've been trying to avoid importing `omero.min` (or `omero.all`) at the top-level of a plugin but it may be that it's no longer possible. An alternative is to use the `IceImport` statement as in `plugins/sessions.py`.

ximenesuk:
Harmonising the imports in plugins would be good to do at some stage. At the moment we have a real mishmash of strategies. I'm not 100% sure why this only affects some imports from `omero.model` but then Python imports can be quite complex once you have a long chain of them from different files.

PR #4512, opened 2016-03-04
Travis findbugs

joshmoore:
Enable findbugs in travis java-build. Increase in build time may make this infeasible.
--exclude
See: https://trac.openmicroscopy.org/ome/ticket/2596


sbesson:
Closing as it is not mergeable anymore. Can be reopened.

PR #4515, opened 2016-03-07
use new request builders in DiskUsage and Duplicate tests

mtbc:
Add use of the new builders for `DiskUsage` and `Duplicate` so that those builders actually get some automated testing. On `regions` instead of `develop` because those test classes have already diverged to include folders.

### DIFF
--- a/components/tools/OmeroJava/test/integration/DiskUsageTest.java
+++ b/components/tools/OmeroJava/test/integration/DiskUsageTest.java
@@ -70,10 +70,8 @@ import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
-import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


mtbc:
http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration/DiskUsageTest/ and http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration/DuplicationTest/ are a blue shade of green.

joshmoore:
In general, I find this quite readable. As mentioned elsewhere, if we want to handle `target` and `id` specially, then the new commands could be made a bit shorter, but perhaps it's worth waiting on that. (The same goes for possibly moving the `excludeType` method to the main build as opposed to the option builder.)

Merging as a good test of the changes. Thanks, @mtbc.

mtbc:
Ah, interesting, `excludeType` on the main builder could add to the latest option in the list, I hadn't thought of that.

PR #4516, opened 2016-03-07
rnd float

jburel:
Fix the float point issue.

To test:
 * Log as user-1
 * Open the image in dataset ``17052``
 * Check that the problem described in https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8011&p=16816#p16816 is fixed

When reviewed, I will post a reply to the forum post above.


pwalczysko:
- repeated the problem on eel latest (in web)
   - was not able to repeat the problem on eel merge (web and insight)

Ready to merge.

jburel:
pr4516

PR #4519, opened 2016-03-08
provide new FindParents and FindChildren requests

mtbc:
These are generic model graph traversal requests for finding the parents and children of model objects. Integration tests are provided. They are intended to be useful for existing code such as https://github.com/openmicroscopy/openmicroscopy/commit/dd532abb92ec82803df97983ea999d0759756816 and will be even more useful with arbitrarily nested folders for single-query traversal.

--exclude as seems to be breaking

### DIFF
--- a/components/blitz/resources/ome/services/blitz-graph-rules.xml
+++ b/components/blitz/resources/ome/services/blitz-graph-rules.xml
@@ -76,5 +76,7 @@
     <import resource="graph-rules/blitz-chown-rules.xml"/>
     <import resource="graph-rules/blitz-delete-rules.xml"/>
     <import resource="graph-rules/blitz-duplicate-rules.xml"/>
+    <import resource="graph-rules/blitz-container-rules.xml"/>
+    <import resource="graph-rules/blitz-contained-rules.xml"/>
 
 </beans>
--- a/components/blitz/resources/ome/services/blitz-servantDefinitions.xml
+++ b/components/blitz/resources/ome/services/blitz-servantDefinitions.xml
@@ -368,6 +368,8 @@
           <entry key="omero.cmd.graphs.Chown2I" value-ref="chownTargets"/>
           <entry key="omero.cmd.graphs.Delete2I" value-ref="deleteTargets"/>
           <entry key="omero.cmd.graphs.DuplicateI" value-ref="duplicateTargets"/>
+          <entry key="omero.cmd.graphs.FindParentsI" value-ref="containerTargets"/>
+          <entry key="omero.cmd.graphs.FindChildrenI" value-ref="containedTargets"/>
         </map>
       </constructor-arg>
       <constructor-arg>
@@ -377,6 +379,8 @@
           <entry key="omero.cmd.graphs.Chown2I" value-ref="chownRules"/>
           <entry key="omero.cmd.graphs.Delete2I" value-ref="deleteRules"/>
           <entry key="omero.cmd.graphs.DuplicateI" value-ref="duplicateRules"/>
+          <entry key="omero.cmd.graphs.FindParentsI" value-ref="containerRules"/>
+          <entry key="omero.cmd.graphs.FindChildrenI" value-ref="containedRules"/>
         </map>
       </constructor-arg>
       <constructor-arg>
--- a/components/blitz/resources/ome/services/graph-rules/blitz-contained-rules.xml
+++ b/components/blitz/resources/ome/services/graph-rules/blitz-contained-rules.xml
@@ -1,0 +1,235 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+#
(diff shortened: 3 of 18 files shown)


joshmoore:
Before finalizing the API here, it'd be useful to see how this would be used, for example, in the "delete by query" use case.

Also, I'm slightly missing the stopBefore usage in https://github.com/openmicroscopy/openmicroscopy/pull/4519/files#diff-cedf8ce6e6d93cad295252bc92e98a62R290 Couldn't the ROIs be eliminated in the general case for finding `images -> datasets`?

mtbc:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/FindParentsChildrenTest/ is green.

mtbc:
It may depend on the details of the "delete by query" use case. For instance, using `SkipHead` to delete all a screen's ROIs would target only those that are actually deletable, even if ``dryRun == true`; `FindChildren` would find even those ROIs that you cannot delete.

The `stopBefore` test at https://github.com/openmicroscopy/openmicroscopy/pull/4519/files#diff-cedf8ce6e6d93cad295252bc92e98a62R290 is mostly about ensuring that there isn't a bug in `FindParents` such that if you give it anything for `stopBefore` then it returns no results regardless of the details.

mtbc:
> Couldn't the ROIs be eliminated in the general case for finding `images -> datasets`?

There is a more general point here though: at the moment there *is* too much recursion through the model graph in many cases. (`SkipHead` may be guilty of the same.) For instance, by default in that case it would also check the IDs of the datasets' projects in case those are on the way to further datasets. This is for a few reasons,

1. My schedule for this milestone may be fairly packed and for Folder-Image-ROI queries -- our immediate need -- there are easy workarounds for the folder-related gateway methods to eliminate much useless querying by stopping before annotations and, for children, pixels and ROIs too. Basically, once the code was good enough for now, I figured I should move back onto ROI model changes. (Similarly, there is more refactoring of common graphs code that could be done, but now it's reaching the point of taking more time and thought to refactor less code.)
1. Eliminating the useless querying shouldn't change the requests' results so fixing it is a non-breaking change. (Of course, `stopBefore` has its uses regardless of the fix.)
1. The useless querying oughtn't be too bad: the requests do only load IDs, and then in bulk queries, without initializing whole model objects.
1. It's tricky to figure out how best to eliminate the extra queries in the general case. Complications include:
  1. These are flexible requests: users can look for any classes from any classes.
  1. There's a class hierarchy that users can exploit (e.g., in searching for or stopping before `TextAnnotation` if they so choose).
  1. The graph ascent/descent is actually slightly hairy because of awkwardnesses. For instance, in going down into masks, we then go up from their pixels to their image, and in descending down into acquisition instances we go up a bit too to find instruments that are associated with an image without being directly linked to it (that happens, I don't know why). The inverse is true for ascent.

So my plan for the moment was to create a Trello card for optimizing this, then come back to the issue when I can take a week or so to improve it once the schedule can afford my doing so. I think I see how but there are certainly details to think through.

joshmoore:
This might be easier to talk through, but ultimately this is sounding like we don't have the time needed to do it right. When talking about public API that seems to externalize the burden to the client.

mtbc:
Good `stopBefore`s can be included in the specific gateway methods ("getRoisOfFolders" or whatever).

mtbc:
Could just leave this PR open and work on the optimization after the next batch of ROI work if preferred.

mtbc:
@pwalczysko can't delete between merge/latest today: I think the subclassing of `GraphModify2` must have been a breaking change from Ice's point of view. Excluding this to see if it fixes his problem.

joshmoore:
> Could just leave this PR open and work on the optimization after the next batch of ROI work if preferred.

Guess mostly I'd like to see the API in practice more (I have in my mind via the CLI but also the gateways). I'm not finding it intuitive and so I'm worried others won't either.

mtbc:
If it helps, while I wait for #2160 and @rleigh-dundee's quick whitespace followup I think I've figured out how the optimization could reasonably work so `stopBefore` should usually be omittable altogether.

pwalczysko:
@mtbc : Would you please relist the PR with my name tomorrow to make sure I will retest the particular problem ? Ta

pwalczysko:
@mtbc : Yes, now that this PR is excluded, the deletion works
    - connecting the Insight from DEV-latest build to eel merge server -> as expected
    - connecting the insight 5.2.2 release client to eel merge server -> as expected
    - connecting the insight from DEV-merge build to eel merge server -> as expected

It looks like it really was this PR causing the incompatibility with respect to delete actions in non-merge insight clients connected to merge server.

mtbc:
Thank you @pwalczysko. I'll look at reopening on regions then once I've completed some automated `stopBefore` fill-in and some example gateway methods.

PR #4523, opened 2016-03-10
Display well label on hover over well

will-moore:
This is a partial solution to the problem of knowing which wells are which on a large plate when the row and column labels have scrolled off screen.
See https://trello.com/c/TXLXoyeU/64-critical-loss-of-context-when-looking-for-well
We simply show a well label when you hover over each well:

screen shot 2016-03-10 at 21 47 45

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
@@ -2170,7 +2170,16 @@ div.paging input.button_pagination {
 #spw table td.ui-selecting { border: 1px dashed #555; padding: 1px; background-color:#87ABD2;}
 #spw table td.ui-selected { border: 1px dashed #555; padding: 1px; background-color:#87ABD2;}
 
-
+#spw .well .wellLabel {
+    display: none;
+    position: absolute;
+    background: #eee;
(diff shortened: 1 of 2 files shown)


mtbc:
Code change looks reasonable. Functionally works well in both Iceweasel (i.e. Firefox) and Chromium (i.e. Chrome). Good to merge.

RFE: This PR is already an improvement, but maybe in some new brave new world we can instead have the row and column labels "frozen" and some slight shading for the row/column we're hovering over instead of the edge headings moving out of sight as one pans around the plate.

will-moore:
@mtbc RFE: That's definitely the ideal solution but will be tricky with nested scrolling tables within tables.
A single scrolling table is hard enough http://stackoverflow.com/questions/17067294/html-table-with-100-width-with-vertical-scroll-inside-tbody - will have to think a bit more about header and left column scrolling.

jburel:
@gusferguson: new screenshots for help?

gusferguson:
@jburel - thanks - I will have a look at where it might be relevant.

gusferguson:
@jburel - I don't have any workflows in the current Help that require screenshots to be updated for this, but will include it in the forthcoming "Viewing HCS Data" being planned.

jburel:
@gusferguson thanks. merging

PR #4526, opened 2016-03-14
NPE Importer Dialog

dominikl:
This wee PR should fix [QA 17060](https://www.openmicroscopy.org/qa2/qa/feedback/17060/). I couldn't reproduce the issue, but the only explanation for the NPE is that the "Add to Queue" button must have been hit before the dialog was fully initialized (in particular before the user/group boxes have been populated). With this PR the button is only enabled after full UI initialization.
**Review**: Check if code makes sense; test if you can still add files to the import queue.


jburel:
A similar problem could occur when the user switches group so wrong data might be picked up leading to security violation.

dominikl:
That's a good point, thanks. Will tackle that in this PR, too.

jburel:
Looks good thanks

PR #4527, opened 2016-03-14
merge develop into regions

mtbc:
Contains only commits already merged into develop. http://regions-ci.docker.openmicroscopy.org:8080/ should remain blue.


mtbc:
(The current Travis failure is just the usual `TestTables.testTableSearch`.)

PR #4530, opened 2016-03-15
Lens tool compressed image bugfix

dominikl:
Bugfix for [QA 17054](https://www.openmicroscopy.org/qa2/qa/feedback/17054/)

**Test**: Open an image. Set compression level to high. Open lens tool. Set zoom to > 800%.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/ui/lens/LensModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/ui/lens/LensModel.java
@@ -179,6 +179,13 @@ class LensModel
 					break;
     		}
     	}
+		
+        if (dataBufferType == DataBuffer.TYPE_BYTE
+                && zoomedDataBuffer.getSize() < (w * h * 3))
+            // if type==byte (i.e. compressed images) the buffer size has to be
+            // w * h * 3 at least
+            zoomedDataBuffer = new DataBufferByte((w * h * 3), 1);
+		
 		SampleModel sm = colorModel.createCompatibleSampleModel(w, h);
         return Raster.createWritableRaster(sm, zoomedDataBuffer, null);			
  	}


components/insight/SRC/org/openmicroscopy/shoola/util/ui/lens/LensModel.java
```diff
@@ -303,6 +303,7 @@ BufferedImage getZoomedImage()
 																		null);
 			return scaleBufferedImage(img, zoomFactor, zoomFactor);
 		} catch (Exception e) {
+		    e.printStackTrace();
```
mtbc: intended?

components/insight/SRC/org/openmicroscopy/shoola/util/ui/lens/LensModel.java
```diff
@@ -76,7 +76,7 @@
 	 * The default size of the pre-allocated buffer used to store the 
 	 * zoomed image. 
 	 */
-	final static int		DEFAULT_SIZE = 562500;
+	final static int		DEFAULT_SIZE = 750000;
```
mtbc: any chance of calculating this from the maximum zoom?

components/insight/SRC/org/openmicroscopy/shoola/util/ui/lens/LensModel.java
dominikl: Kind of... The class is not part of an agent package, so I can't get a logger via an agents registry. Is there another method to get access to a 'general' Insight logging instance @jburel ?

components/insight/SRC/org/openmicroscopy/shoola/util/ui/lens/LensModel.java
dominikl: Got this number from the exception, which said the buffer has to be greater than 749999 (when zoom is set to 1000%). Maybe it can be calculated, can do some more investigation.

components/insight/SRC/org/openmicroscopy/shoola/util/ui/lens/LensModel.java
jburel: no,only from registry

pwalczysko:
- repeated the bug on eel latest
   - with the same image (one of the lei-56 series) tried to find the bug on eel merge -> but works as expected

Functionally no problems here. If the conceptual questions above are solved, ready to merge.

dominikl:
Found a way to calculate the needed buffer size. Sorry @pwalczysko , has to be retest again.
I also removed the `printStackTrace()` of the exception again, doesn't make a difference, does not get logged anyway (I was hoping that it somehow will be written into insight.log, but apparently it's not).

pwalczysko:
Tested today on mac. Works as expected. Ready to merge.

PR #4531, opened 2016-03-16
provide new FindParents and FindChildren requests

mtbc:
These are generic model graph traversal requests for finding the parents and children of model objects. Integration tests are provided. They are intended to be useful for existing code such as https://github.com/openmicroscopy/openmicroscopy/commit/dd532abb92ec82803df97983ea999d0759756816 and will be even more useful with arbitrarily nested folders for single-query traversal.

Beyond #4519 this PR adds Folder support and automatic assistance with `stopBefore`.

Integration tests are provided at http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration/FindParentsChildrenTest/ and the curious could turn on debug logging for `omero.cmd.graphs` to see the extra `stopBefore` classes being added.

I am certainly happy to assist with the creation of Java gateway methods.  These may want to use `stopBefore(Mask.class)` so that folder contents queries do not include images or ROIs that are not directly in folders but instead relate to masks of those that are.

--no-rebase as introduction of `GraphQuery` superclass breaks Blitz API

### DIFF
--- a/components/blitz/resources/ome/services/blitz-graph-rules.xml
+++ b/components/blitz/resources/ome/services/blitz-graph-rules.xml
@@ -76,5 +76,7 @@
     <import resource="graph-rules/blitz-chown-rules.xml"/>
     <import resource="graph-rules/blitz-delete-rules.xml"/>
     <import resource="graph-rules/blitz-duplicate-rules.xml"/>
+    <import resource="graph-rules/blitz-container-rules.xml"/>
+    <import resource="graph-rules/blitz-contained-rules.xml"/>
 
 </beans>
--- a/components/blitz/resources/ome/services/blitz-servantDefinitions.xml
+++ b/components/blitz/resources/ome/services/blitz-servantDefinitions.xml
@@ -368,6 +368,8 @@
           <entry key="omero.cmd.graphs.Chown2I" value-ref="chownTargets"/>
           <entry key="omero.cmd.graphs.Delete2I" value-ref="deleteTargets"/>
           <entry key="omero.cmd.graphs.DuplicateI" value-ref="duplicateTargets"/>
+          <entry key="omero.cmd.graphs.FindParentsI" value-ref="containerTargets"/>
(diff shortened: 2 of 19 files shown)


dominikl:
:+1: Looking forward to make use of these methods.

ximenesuk:
I have checked very basic API consumption using https://github.com/ximenesuk/openmicroscopy/commit/583dbdb6258027ec92e1796e3c127979bbe0db60 though this was limited to finding single types of single objects:

```
$ omero obj children Project:1 Annotation
MapAnnotation:1
TagAnnotation:5,6
FileAnnotation:2
CommentAnnotation:4
LongAnnotation:3

$ omero obj children Project:1 Image
Image:1,2,3,4,5,6,7,8,9

$ omero obj parents Dataset:1 Project
Project:1,2
```

It looks fine from this respect.

sbesson:
All looks good. Integration tests are passing and the simple CLI extension from @ximenesuk is instructive. Merging this to unlock the gateway work consuming this request.

One minor comment about the API documentation. The `stopBefore` functionality is described as excluding objects from the search which led me to wonder why it would not simply be called `exclude`.
After looking at the code, we might want to clarify the fact it is also stopping the graph traversal and excluding downstream objects.

joshmoore:
@manics : can you compare what you recently did with `bin/omero children`?

mtbc:
The docs describe it as,

> Classes of model objects to exclude from the recursive search. Search does not include or pass such objects.

so with the "recursive search" and "does not *pass*" I'd already tried to clarify that it is stopping the graph traversal and excluding downstream objects (except for those it can reach by some different path). I am certainly open to better wording!

It's `stopBefore` to further emphasize that it's interrupting a *process* (i.e. affects subgraphs, not just individual instances) and to emphasize the "does not *include*". That is, the traversal stops before nodes of that class. Of course, it could be renamed.

PR #4532, opened 2016-03-16
Re-enable saving of ExternalInfo

joshmoore:
At some point (either during the move to blitz or later
during the hiding of Details implementations), the copying
of ExternalInfo objects from one Details to the next was
lost. Since then, no ExternalInfo objects could have been
linked to any IObject, and it's even unlikely that they
could have been created at all.

### DIFF
--- a/components/model/src/ome/model/internal/Details.java
+++ b/components/model/src/ome/model/internal/Details.java
@@ -151,6 +151,7 @@ public abstract class Details implements Filterable, Serializable {
         setContexts(copy.contexts);
         setContext(copy.getContext());
         possiblySetPermissions(copy);
+        setExternalInfo(copy.getExternalInfo());
         setCreationEvent(copy.getCreationEvent());
         setOwner(copy.getOwner());
         setGroup(copy.getGroup());


joshmoore:
That's certainly one possible outcome of the current investigation. Alternatively, if they are useful, we should at least have tests to catch if we should make this mistake again. :smile:

chris-allan:
With some of the work we are doing having something like `ExternalInfo` is incredibly useful. We are attempting to maintain entity synchronization between multiple disparate systems on a schedule. The frequency of synchronization is 24 hours or less.

Using `ExternalInfo.entityType` and `ExternalInfo.entityId` we can confidently make queries examining whether an entity is already present and coupled with `version` even ascertain when it was updated. Without the presence of such functionality as a "first class citizen", we would be forced to use annotations which provide neither the specificity nor containment desired. I accept that it is somewhat superficial that this rationale is fundamentally based on the fact that `ExternalInfo` cannot be interacted with using our current clients.

As a thought experiment, let's consider synchronizing a "Study" named "Acute Myeloid Leukemia", which has a source identifier of `52` between a digital lab notebook system and OMERO. We'll use an OMERO `Project` to represent the "Study".

Creating the `Project` for the "Study":

```
In [1]: client.enableKeepAlive(60)

In [2]: session = client.getSession()

In [3]: update_service = session.getUpdateService()

In [4]: from omero.rtypes import rlong, rstring

In [5]: p = omero.model.ProjectI()

In [6]: p.name = rstring('Acute Myeloid Leukemia')

In [7]: external_info = omero.model.ExternalInfoI()

In [8]: external_info.entityType = rstring('com.glencoesoftware.sync:study')

In [9]: external_info.entityId = rlong(52L)

In [10]: p.details.externalInfo = external_info

In [11]: p = update_service.saveAndReturnObject(p)

In [12]: p.id.val
Out[12]: 3L

In [13]: p.details.externalInfo.entityType
Out[13]: 
object #0 (::omero::RString)
{
    _val = com.glencoesoftware.sync:study
}

In [14]: p.details.externalInfo.entityId
Out[14]: 
object #0 (::omero::RLong)
{
    _val = 52
}
```

All updates to the "Study" properties can then be performed by referencing the identifier and no string matching need take place. This is incredibly useful in an "upsert" type environment where you are attempting to maintain one and only one corresponding OMERO `Project` instance for the "Study". Querying for the "Study":

```
In [1]: client.enableKeepAlive(60)

In [2]: session = client.getSession()

In [3]: query_service = session.getQueryService()

In [4]: params = omero.sys.ParametersI()

In [5]: params.addId(52L)
Out[5]: 
object #0 (::omero::sys::Parameters)
{
    map = 
    {
        key = id
        value = object #1 (::omero::RLong)
        {
            _val = 52
        }
    }
    theFilter = <nil>
    theOptions = <nil>
}

In [6]: params.addString('entityType', 'com.glencoesoftware.sync:study')
Out[6]: 
object #0 (::omero::sys::Parameters)
{
    map = 
    {
        key = entityType
        value = object #1 (::omero::RString)
        {
            _val = com.glencoesoftware.sync:study
        }
        
        key = id
        value = object #2 (::omero::RLong)
        {
            _val = 52
        }
    }
    theFilter = <nil>
    theOptions = <nil>
}

In [7]: p = query_service.findByQuery('SELECT p from Project AS p 
JOIN FETCH p.details.externalInfo as einfo 
WHERE einfo.entityId = :id and einfo.entityType = :entityType', params, {'omero.group': '-1'})

In [8]: p.id.val
Out[8]: 3L

In [9]: p.details.externalInfo.entityType
Out[9]: 
object #0 (::omero::RString)
{
    _val = com.glencoesoftware.sync:study
}

In [10]: p.details.externalInfo.entityId
Out[10]: 
object #0 (::omero::RLong)
{
    _val = 52
}
```

With annotations, something comparable could of course be achieved using an OMERO `LongAnnotation`. Where `ExternalInfo.entityType == LongAnnotation.Ns` and `ExternalInfo.entityId == LongAnnotation.LongValue`. The query implications of such a construct are quite significant. Especially in situations where the annotation table could contain millions of entries.

Future improvements would likelyinclude ensuring there are indexes on both `ExternalInfo.entityType` and `ExternalInfo.entityId`. Something we have done in a variety of other places.

Food for thought.

/cc @emilroz, @sbesson, @dominikl

mtbc:
I certainly agree that *some* way to reference such external links is valuable, I just wasn't sure that reusing `ExternalInfo` was the desired way given that it was sounding very tied to the failed LSID effort.

chris-allan:
In doing some further testing I realized that the `ExternalInfo` is not being cleaned up when the `Project` is deleted. As the relationship between the two objects is likely intended to be 1:1 I'd expect this is a bug. Test case:

```
In [1]: session = client.getSession()

In [2]: update_service = session.getUpdateService()

In [3]: query_service = session.getQueryService()

In [4]: from omero.rtypes import rlong, rstring

In [5]: p = omero.model.ProjectI()

In [6]: p.name = rstring('Acute Myeloid Leukemia')

In [7]: external_info = omero.model.ExternalInfoI()

In [8]: external_info.entityType = rstring('com.glencoesoftware.sync:study')

In [9]: external_info.entityId = rlong(52L)

In [10]: p.details.externalInfo = external_info

In [11]: p = update_service.saveAndReturnObject(p)

In [12]: p.id.val
Out[12]: 4L

In [13]: p.details.externalInfo.id.val
Out[13]: 3L

In [14]: p.details.externalInfo.entityType.val
Out[14]: 'com.glencoesoftware.sync:study'

In [15]: p.details.externalInfo.entityId.val
Out[15]: 52L

In [16]: update_service.deleteObject(omero.model.ProjectI(4L, False))

In [17]: external_info = query_service.get('ExternalInfo', 3L)

In [18]: external_info.id.val
Out[18]: 3L

In [19]: try:
   ....:     p = query_service.get('Project', 4L)
   ....: except omero.ValidationException:
   ....:     print 'Project does not exist!'
   ....: 
Project does not exist!
```

joshmoore:
```
 entityid    │ bigint                 │ not null
 entitytype  │ character varying(255) │ not null
 lsid        │ character varying(255) │
 uuid        │ character varying(255) │
```

Things I'd like to work through before our buying into `ExternalInfo` (in addition to the indexes & widening of the string values):

 * What is the implication/meaning of an entityid/entitytype that is within the same DB? @chris-allan has pointed out an external usage. Are entity type & id completely a client prerogative? Or will there be any internal usage that we should provide for?
 * Is `lsid` ever intended to be used? Only on import? Should it be removed? Or is it again, just another location that a client *could* use?
 * Similarly for UUID, but perhaps most importantly, is there any relationship between `lsid` and `uuid` (under any circumstances)?

mtbc:
It's a bug: the,
```xml
<bean parent="graphPolicyRule" p:matches="[D].details.externalInfo = EI:[E]" p:changes="EI:[D]/n"/>
```
rule must not be working as I intended.

chris-allan:
It certainly has an LSID column, @mtbc. That's pretty much the extent it or any of our documentation references any sort of LSID effort as far as I'm aware.

chris-allan:
My perspective, @joshmoore, is that the LSID column be dropped, we rename `uuid` to `entityKey` (keeping it at 255 characters), index all three columns and consider all uses to be opaque, client decisions. Furthermore, we explicitly state that `entityKey` is designed to allow for fast lookup of opaque string based identifiers or keys with no intrinsic meaning. If your keys are > 255 characters you can use the hashing algorithm of your choice to give you that shorter string.

joshmoore:
@chris-allan : and drop the NOT NULL on entityId?

chris-allan:
Agreed.

If we wanted to get exotic, we could perhaps keep the `uuid` column, adding the `entityKey`, and implicitly fill it with a server generated `uuid` to give us a globally unique way of referencing every object in every OMERO database.

joshmoore:
> ...implicitly fill it with a server generated uuid to give us a globally unique way of referencing every object in every OMERO database.

This was certainly one of the early ideas, but the cost is an extra row for every object in the DB and an implicitly huge externalinfo table. Currently the "GUID" for every object is `DB uuid + table + PK`. This is certainly tied to the concept of LSID, so we need some detangling one way or the other.

chris-allan:
That "GUID" strategy works for me and certainly avoids the overhead mentioned. Completely safe in my view to drop the `uuid` column then if we're including an `entityKey` instead. I guess from a storage perspective going forward should we just put these fields on the objects themselves rather than add another object every time? Is there any reason to have the separate object?

joshmoore:
*Conceivably* there would be a reason to have 2 objects re-using the same externalinfo (i.e. if re-imported, etc). This of course has **never** happened.

joshmoore:
Recording TODOs from here as well as external conversations on https://trello.com/c/30kEyUh6/46-externalinfo-rationalization

mtbc:
Functional testing without then with this PR shows that it enables the saving of external info via the update service. It also fixes `Delete2`'s deletion of external info.

joshmoore:
@chris-allan : will use of `Delete2` be an acceptable workaround for `IUpdatedeleteObject`'s not cleaning up the `ExternalInfo`?

mtbc:
Note that `IUpdate.deleteObject` is deprecated so if it's actually needed for some reason we should know that soon.

mtbc:
The code change in this PR looks good to me too. I'd say, go ahead and **merge** it, separate from the larger discussion of if we want to change `ExternalInfo`'s properties somehow.

chris-allan:
Happy with all of that and for this to be merged.

jburel:
@mtbc: do you want to move the large discussion to a card/design issue?

mtbc:
Looks to be already captured in https://trello.com/c/30kEyUh6/46-externalinfo-rationalization, thank you @joshmoore.

joshmoore:
--rebased-to #4571

PR #4538, opened 2016-03-24
Gateway loadROIs

dominikl:
This PR adds a more elaborated `loadROIs` method to the Java Gateway, which allows more specific loading of ROIs (Folder ROIs, ROIs without image, orphaned ROIs, etc.). 

**Test**: Code review, check integration test job; there are no implications on Insight yet.

Note: I marked previous methods which return `ROIResult`s as deprecated, as they can be replaced by the new `loadROIs` method. However this method returns `ROIData` objects. I didn't find any specific use of the `ROIResult` class other than being used as wrapper around a collection of `ROIData` objects, so in my opinion just returning the `ROIData` objects directly make things easier.

### DIFF
--- a/components/blitz/src/omero/gateway/facility/ROIFacility.java
+++ b/components/blitz/src/omero/gateway/facility/ROIFacility.java
@@ -32,6 +32,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
+import javax.naming.OperationNotSupportedException;
+
 import org.apache.commons.collections.CollectionUtils;
 
 import omero.ServerError;
@@ -40,6 +42,9 @@ import omero.api.IRoiPrx;
 import omero.api.IUpdatePrx;
 import omero.api.RoiOptions;
 import omero.api.RoiResult;
+import omero.cmd.CmdCallbackI;
+import omero.cmd.FindChildren;
+import omero.cmd.FoundChildren;
 import omero.gateway.Gateway;
 import omero.gateway.SecurityContext;
 import omero.gateway.exception.DSAccessException;
@@ -48,6 +53,7 @@ import omero.gateway.model.ROIResult;
 import omero.gateway.util.ModelMapper;
 import omero.gateway.util.Pojos;
 import omero.gateway.util.PyTablesUtils;
+import omero.gateway.util.Requests;
 import omero.model.FolderRoiLink;
 import omero.model.FolderRoiLinkI;
 import omero.model.IObject;
@@ -94,8 +100,9 @@ public class ROIFacility extends Facility {
     }
 
     /**
-     * Loads the ROI 
-     *
+     * Loads the ROI
+     * 
+     * @deprecated Use {@link #loadROIs(SecurityContext, Collection)} instead
      * @param ctx
      *            The security context.
      * @param roiId
@@ -107,6 +114,7 @@ public class ROIFacility extends Facility {
      *             If an error occurred while trying to retrieve data from OMERO
      *             service.
      */
+    @Deprecated
     public ROIResult loadROI(SecurityContext ctx, long roiId)
             throws DSOutOfServiceException, DSAccessException {
         try {
@@ -158,7 +166,9 @@ public class ROIFacility extends Facility {
     
     /**
      * Loads the ROI related to the specified image.
-     *
+     * 
+     * @deprecated Use {@link #loadROIs(SecurityContext, Long, Collection)}
+     *             instead
      * @param ctx
      *            The security context.
      * @param imageID
@@ -170,6 +180,7 @@ public class ROIFacility extends Facility {
      *             If an error occurred while trying to retrieve data from OMERO
      *             service.
      */
+    @Deprecated
     public List<ROIResult> loadROIs(SecurityContext ctx, long imageID)
             throws DSOutOfServiceException, DSAccessException {
         return loadROIs(ctx, imageID, null, gateway.getLoggedInUser().getId());
@@ -177,33 +188,16 @@ public class ROIFacility extends Facility {
 
     /**
      * Loads the ROI related to the specified image.
-     *
+     * 
+     * @deprecated Use {@link #loadROIs(SecurityContext, Long, Collection)}
+     *             instead
+     * 
      * @param ctx
      *            The security context.
      * @param imageID
      *            The image's ID.
-     * @param measurements The measurements IDs linked to the image if any.
-     * @return See above.
-     * @throws DSOutOfServiceException
-     *             If the connection is broken, or logged in.
-     * @throws DSAccessException
-     *             If an error occurred while trying to retrieve data from OMEDS
-     *             service.
-     */
-    public List<ROIResult> loadROIs(SecurityContext ctx, long imageID,
-            List<Long> measurements) throws DSOutOfServiceException,
-            DSAccessException {
-        return loadROIs(ctx, imageID, measurements, -1);
-    }
-
-    /**
-     * Loads the ROI related to the specified image.
-     *
-     * @param ctx
-     *            The security context.
-     * @param imageID
-     *            The image's ID.
-     * @param measurements The measurements IDs linked to the image if any.
+     * @param measurements
+     *            The measurements IDs linked to the image if any.
      * @param userID
      *            The user's ID.
      * @return See above.
@@ -213,6 +207,7 @@ public class ROIFacility extends Facility {
      *             If an error occurred while trying to retrieve data from OMERO
      *             service.
      */
+    @Deprecated
     public List<ROIResult> loadROIs(SecurityContext ctx, long imageID,
             List<Long> measurements, long userID)
             throws DSOutOfServiceException, DSAccessException {
@@ -520,7 +515,8 @@ public class ROIFacility extends Facility {
      * @param ctx
      *            The security context.
      * @param imageID
-     *            The image's ID.
+     *            The image's ID (can be <code>-1</code> for ROIs not attached
+     *            to an image)
      * @param userID
      *            The user's ID.
      * @param roiList
@@ -539,6 +535,23 @@ public class ROIFacility extends Facility {
         try {
             IUpdatePrx updateService = gateway.getUpdateService(ctx);
             IRoiPrx svc = gateway.getROIService(ctx);
+            
+            Collection<ROIData> updated = new ArrayList<ROIData>();
+
+            if (imageID < 0) {
+                for (ROIData r : roiList) {
+                    if (!r.isClientSide())
+                        throw new OperationNotSupportedException(
+                                "Modification of existing ROIs is not implemented yet.");
+                }
+                
+                for (ROIData r : roiList) {
+                    Roi ri = (Roi) updateService.saveAndReturnObject(r
+                            .asIObject());
+                    updated.add(new ROIData(ri));
+                }
+                return updated;
+            }
             
             RoiOptions options = new RoiOptions();
             if (userID >= 0)
@@ -591,7 +604,6 @@ public class ROIFacility extends Facility {
             ROICoordinate coord;
             int shapeIndex;
 
-            Collection<ROIData> updated = new ArrayList<ROIData>();
             List<Long> deleted = new ArrayList<Long>();
             Image unloaded = new ImageI(imageID, false);
             Roi rr;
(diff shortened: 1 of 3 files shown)


components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,193 @@ else if (shape.isDirty()) {
...
+                    if (roi.copyFolderLinks().isEmpty())
+                        roiIDs.add(roi.getId().getValue());
+                }
+            } else {
+                // get all rois which are part of the given folders
+                Map<String, Collection<Long>> target = new HashMap<String, Collection<Long>>();
+                target.put("Folder", folderIds);
+
+                FindChildren finder = Requests.findChildren().target(target)
+                        .childType("Roi").build();
```
mtbc: Do we want to include ROIs that are on images that are in the folder?

Also if preferred can simply set target with `.target("Folder").id(folderIds)`

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,193 @@ else if (shape.isDirty()) {
...
+                roiIDs = found.children.remove(ome.model.roi.Roi.class
+                        .getName());
+
+                if (imageId != null) {
+                    if (imageId > -1) {
+                        // only take ROIs for the given image into account
+                        target = new HashMap<String, Collection<Long>>();
+                        target.put("Image", Collections.singleton(imageId));
+                        finder = Requests.findChildren().target(target)
+                                .childType("Roi").build();
```
mtbc: Do we want to include ROIs that are on masks of the image?

dominikl:
For now I just used HQL queries for retrieving the ROIs for specific images; will try to also replace these with the `FindChildren` methods introduced by @mtbc .

mtbc:
HQL will probably be more efficient when it's possible to write the query like that.

components/blitz/src/omero/gateway/facility/ROIFacility.java
dominikl: My intention was to get all ROIs of the folders (with and without images) and then filter out with respect to the provided imageId with a second request. Do you suggest that's possible within one request?

components/blitz/src/omero/gateway/facility/ROIFacility.java
dominikl: I forgot about the masks... they are somehow special, and have to be excluded, right? If so, is there a "neat" way how to do that?

components/blitz/src/omero/gateway/facility/ROIFacility.java
mtbc: If you `.stopBefore("Image")` it'll give you only the ROIs directly in the folder, not including the ROIs that are not in the folder directly but are on images in the folder. You can't filter also by imageID in the same request I'm afraid.

components/blitz/src/omero/gateway/facility/ROIFacility.java
mtbc: `.stopBefore("Shape")` should stop it recursing that far down.

components/blitz/src/omero/gateway/facility/ROIFacility.java
dominikl: Ah sorry, I was a bit confused, you were talking about ROIs on Images in Folders (meaning images organized in Folders), in my mind still a Folder == Folder of ROIs. Haven't really considered the case of Folder of Images. Should make the method description more clear, that it only deals with ROI folders.
Thanks for the hint with `stopBefore`, guess that'd also increase the performance.

dominikl:
Implemented @mtbc suggestions. Will leave HQL queries then.

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -933,14 +936,9 @@ else if (shape.isDirty()) {
...
-                Map<String, Collection<Long>> target = new HashMap<String, Collection<Long>>();
-                target.put("Folder", folderIds);
-
-                FindChildren finder = Requests.findChildren().target(target)
-                        .childType("Roi").build();
-
-                Map<String, String> callContext = new HashMap<String, String>();
-                callContext.put("omero.group", "" + ctx.getGroupID());
+                FindChildren finder = Requests.findChildren().target("Folder")
+                        .id(folderIds.toArray(new Long[folderIds.size()]))
```
mtbc: is the `toArray` stuff needed here?

dominikl: You're right; sorry missed the `id(Iterable)` method.

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,190 @@ else if (shape.isDirty()) {
...
+            } else {
+                // get all rois which are part of the given folders
+                FindChildren finder = Requests.findChildren().target("Folder")
+                        .id(folderIds)
+                        .stopBefore("Image").childType("Roi").build();
+
+                CmdCallbackI cb = gateway.submit(ctx, finder);
+                cb.block(10000);
+                FoundChildren found = (FoundChildren) cb.getResponse();
+                roiIDs = found.children.remove(ome.model.roi.Roi.class
```
mtbc: "remove" may be overkill; perhaps "get" suffices?

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,190 @@ else if (shape.isDirty()) {
...
+                                .id(imageId).childType("Roi")
+                                .stopBefore("Shape").build();
+
+                        cb = gateway.submit(ctx, finder);
+                        cb.block(10000);
+                        found = (FoundChildren) cb.getResponse();
+                        List<Long> imageRoiIDs = found.children
+                                .remove(ome.model.roi.Roi.class.getName());
+
+                        Iterator<Long> it = roiIDs.iterator();
```
mtbc: we're sure `roiIDs` isn't null at this point?

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -540,6 +536,23 @@ public ROIResult loadROI(SecurityContext ctx, long roiId)
             IUpdatePrx updateService = gateway.getUpdateService(ctx);
             IRoiPrx svc = gateway.getROIService(ctx);
             
+            Collection<ROIData> updated = new ArrayList<ROIData>();
+
+            if (imageID < 0) {
+                for (ROIData r : roiList) {
+                    if (r.getId() > -1)
+                        throw new OperationNotSupportedException(
```
mtbc: odd choice -- this exception is related to `javax.naming.Context` -- maybe you meant `UnsupportedOperationException`? Also Apache commons has a `NotImplementedException`.

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -94,8 +100,9 @@
     }
 
     /**
-     * Loads the ROI 
-     *
+     * Loads the ROI
+     * 
+     * @deprecated Use {@link #loadROIs(SecurityContext, Collection)} instead
```
mtbc: :+1: for telling them what to use instead!

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,190 @@ else if (shape.isDirty()) {
...
+            if (folderIds == null) {
+                // ignore folders, ROIs can but don't have to be part of folders
+                String query;
+                ParametersI p = new ParametersI();
+                if (imageId == null) {
+                    // ignore imageId, get all rois
+                    query = "select distinct roi from Roi roi";
+                } else if (imageId < 0) {
+                    // get all rois not attached to any image
+                    query = "select distinct roi from Roi roi "
```
mtbc: Is this `distinct` needed?

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,190 @@ else if (shape.isDirty()) {
...
+                String query;
+                ParametersI p = new ParametersI();
+                if (imageId == null) {
+                    // ignore imageId, get all rois
+                    query = "select distinct roi from Roi roi "
+                            + "left outer join fetch roi.folderLinks as links ";
+                } else if (imageId > -1) {
+                    // get rois for the image
+                    query = "select distinct roi from Roi roi "
+                            + "left outer join fetch roi.image as image "
```
mtbc: interesting that you do the join fetch image here but not for the query above -- if you need it just for the where, you can simply write "where roi.image.id = :id"

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,190 @@ else if (shape.isDirty()) {
...
+                        cb = gateway.submit(ctx, finder);
+                        cb.block(10000);
+                        found = (FoundChildren) cb.getResponse();
+                        List<Long> imageRoiIDs = found.children
+                                .remove(ome.model.roi.Roi.class.getName());
+
+                        Iterator<Long> it = roiIDs.iterator();
+                        while (it.hasNext()) {
+                            if (!imageRoiIDs.contains(it.next()))
+                                it.remove();
```
mtbc: just do `roiIds.retainAll(imageRoiIDs)`?

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,190 @@ else if (shape.isDirty()) {
...
+     * @return See above.
+     * @throws DSOutOfServiceException
+     * @throws DSAccessException
+     */
+    public Collection<ROIData> loadROIs(SecurityContext ctx,
+            Collection<Long> ids) throws DSOutOfServiceException,
+            DSAccessException {
+        try {
+            if (ids == null || ids.isEmpty())
+                return Collections.EMPTY_LIST;
```
mtbc: if you're getting a yellow Eclipse warning on this you should find that you can fix it with `Collections.emptyList()`

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -834,71 +846,190 @@ else if (shape.isDirty()) {
...
-            List<IObject> links = qs.findAllByQuery(sb.toString(), param);
+                List<IObject> objs = service.findAllByQuery(query, p);
 
-            Set<Long> roiIds = new HashSet<Long>();
-            for (IObject l : links) {
-                FolderRoiLink link = (FolderRoiLink) l;
-                roiIds.add(link.getChild().getId().getValue());
+                for (IObject obj : objs) {
+                    Roi roi = (Roi) obj;
+                    if (roi.copyFolderLinks().isEmpty())
```
mtbc: to avoid the collection copy, is there no `.sizeOf*` method?

components/tools/OmeroJava/test/integration/gateway/ROIFacilityTest.java
```diff
@@ -77,9 +77,11 @@ protected void setUp() throws Exception {
         initData();
     }
 
+    // The tests should run in the order they have here, hence
```
mtbc: "...they appear here..."

components/tools/OmeroJava/test/integration/gateway/ROIFacilityTest.java
```diff
@@ -89,125 +91,188 @@ public void testSaveROIs() throws DSOutOfServiceException,
...
     public void testGetROIFolders() throws DSOutOfServiceException,
             DSAccessException {
         Collection<FolderData> folders = roifac.getROIFolders(rootCtx,
-                folderImg.getId());
+                img.getId());
         Assert.assertEquals(1, folders.size());
         Assert.assertEquals(folder.getId(), folders.iterator().next().getId());
     }
-
+
```
mtbc: don't need to introduce trailing space

mtbc:
http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration.gateway/ROIFacilityTest/ is reassuring.

dominikl:
Thanks @mtbc for the plenty suggestions, implemented them in the last commit.

*Review*: Last commit only; code + make sure integration tests are green.

components/tools/OmeroJava/test/integration/gateway/ROIFacilityTest.java
```diff
@@ -315,7 +315,7 @@ private ROIData createRectangleROI(int x, int y, int w, int h) {
         roiData.addShapeData(rectangle);
         return roiData;
     }
-
+
```
mtbc: probably don't need this space

dominikl:
This PR might be superfluous, as it's not really clear yet, which kind of methods will be needed for selective loading/browsing through ROIs/Folders. I'm closing it therefore for now; might extract currently useful code to a separate PR.

PR #4543, opened 2016-03-25
Import target fixes for regex templates

ximenesuk:
This fixes the behaviour seen in https://trello.com/c/oBRyeuuI/50-multiple-import-targets. It also fixes a bug seen in passing. Integration tests are added for both fixes.

To test a nested import create a folder structure similar to the one below,
```
~/outer/inner-1
~/outer/inner-2
```
with an image or screen or more in each of the inner folders. Then import using a regex template,
```
omero import ~/outer -T "regex:name:^.*outer/(?<Container1>.*?)" 
```
The result should be that the images/screens are imported into two separate Datasets/Plates, inner-1 and inner-2.

The bug noticed was was that the `-name` discriminator was not being respected for Dataset regex imports . This bug arose because I failed to add an integration test when the initial work was done here! That has been rectified in this PR.

To test for this create a number of Datasets with the same name and and import an image from inside a folder of that name. So if the image is under
```
~/outer/inner
```
then create several Datasets called `inner`.
```
omero import ~/outer -T "regex:-name:^.*outer/(?<Container1>.*?)" 
```
should import into the oldest Dataset `inner`, i.e. the one with the smallest id.
```
omero import ~/outer -T "regex:+name:^.*outer/(?<Container1>.*?)" 
```
should still import into the newest dataset called `inner`.

@joshmoore the problem here was that multiple CommentAnnotations are created, one for each Fileset, but only one if picked up server-side. The code here takes the easier option and picks up the most recent server-side which will correspond to the current Fileset. A neater approach might be to only allow one CommentAnnotation with the correct namespace client-side.

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
@@ -363,12 +363,15 @@ public class ManagedImportRequestI extends ImportRequest implements IRequest {
             return;
         }
 
+        ome.model.annotations.CommentAnnotation lastCA = null;
+        long maxCAId = 0;
         for (Annotation a : settings.userSpecifiedAnnotationList) {
             if (a == null) {
                 continue;
             }
 
             ome.model.annotations.Annotation ann = null;
+
             String ns = null;
             if (a.isLoaded()) {
                 ns = a.getNs() == null ? null : a.getNs().getValue();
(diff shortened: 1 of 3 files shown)


pwalczysko:
The behaviour for the first test is a bit different for Screens than as described in the header of this PR. As there are often nested folders within the imported plates, the whole path gets interpreted as the name of the screen (see below). This is an expected behaviour as discussed with @ximenesuk but maybe it is worth explaining in the doc.

screen shot 2016-03-28 at 16 01 39

pwalczysko:
The second test starting with words
> To test for this create a number of Datasets with.....

passed as expected. 
All good here, the comment about screens https://github.com/openmicroscopy/openmicroscopy/pull/4543#issuecomment-202448281 is just for the doc.

jburel:
@ximenesuk: Will you open a follow-up PR for the doc? cf https://github.com/openmicroscopy/openmicroscopy/pull/4543#issuecomment-202448281

ximenesuk:
@jburel Yes, I'll add an example to the docs in a follow-up.

ximenesuk:
See https://github.com/openmicroscopy/ome-documentation/pull/1436

jburel:
Waiting for @pwalczysko to go over the doc PR before merging, just in case

components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
```diff
@@ -387,30 +390,36 @@ private void detectKnownAnnotations() throws Exception {
...
                 }
-
+            }
+        }
+        if (lastCA != null) {
+            if (settings.userSpecifiedTarget != null) {
+                // TODO: Exception
+                String kls = settings.userSpecifiedTarget.getClass().getSimpleName();
+                long id = settings.userSpecifiedTarget.getId().getValue();
+                log.error("User-specified template target '{}' AND {}:{}",
```
joshmoore: Is this intended to be at "ERROR"?

ximenesuk: You tell me! I just preserved the original line from https://github.com/openmicroscopy/openmicroscopy/blame/develop/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java#L396 which I think is yours.

Would you rather it was set at INFO? I can add a commit here if so.

joshmoore: Touché. Then my apologies. I would think then that this is leave be and there's a follow-up PR or we decide here whether or not it's an error for the user to specify 2 different templates.

mtbc:
Docs PR is now approved and merged: https://github.com/openmicroscopy/ome-documentation/pull/1436#issuecomment-204388058.

joshmoore:
@ximenesuk : anything left here re:

> @joshmoore the problem here was that...

? Separate card? Or just ok for the moment?

ximenesuk:
@joshmoore nothing left from me, that was a question to you about the approach I took - which was pragmatic. As far as I can see my approach works but it may be something we want to look at later so a card as a reminder would be useful.

pwalczysko:
All good here fmpow.

jburel:
card created https://trello.com/c/WevBjua1/661-import-target-follow-up

jburel:
Merging

PR #3264, opened 2014-12-09
Support for collapsing unique light paths (rebased onto develop)

jburel:
This is the same as gh-3261 but rebased onto develop.

----

Logic and unit test cases for collapsing unique light paths. Test cases were also expanded to cover all other existing collapse logic. Some existing, but old and non-functional, test cases have been ported to TestNG and the entire suite of `ome.formats.*` unit tests added to the server unit test suite.

Due to the amount of white space changes that were required, adding `?w=1` to the end of the GitHub "Files Changed" page may be beneficial during review.

Manual testing has been performed:
* Imported one or more Plates with `LightPath` metadata populated (Flex in particular is useful here)
* Confirmed that `LogicalChannel` and `LightPath` reduction has taken place
* Confirmed that the Plates can be deleted
* Confirmed that the Plates can be be moved between groups
* Testing also took place in the presence of a backported version of #3228

`LightPath` and `LogicalChannel` reduction can be confirmed by examining the import log for a sufficiently small number. For example:

```
2014-11-07 11:15:50,320 INFO  [          ome.formats.OMEROMetadataStore] (      main) Unique objective settings: 1
2014-11-07 11:15:50,324 INFO  [          ome.formats.OMEROMetadataStore] (      main) Unique light settings: 1
2014-11-07 11:15:50,324 INFO  [          ome.formats.OMEROMetadataStore] (      main) Unique detector settings: 1
2014-11-07 11:15:50,324 INFO  [          ome.formats.OMEROMetadataStore] (      main) Unique light paths: 2
2014-11-07 11:15:50,324 INFO  [          ome.formats.OMEROMetadataStore] (      main) Unique logical channels: 2
```

Example datasets contained 12 fields, 384 wells and 2 channels resulting in 9216 unique logical channels per Plate. Completely broken on import. With a 2GB heap, takes ~4.5 hours just to get to pixel data processing:

```
...
2014-11-05 06:12:13,751 INFO  [          ome.formats.OMEROMetadataStore] (.Server-10) Unique objective settings: 1
2014-11-05 06:12:13,751 INFO  [          ome.formats.OMEROMetadataStore] (.Server-10) Unique light settings: 2
2014-11-05 06:12:13,751 INFO  [          ome.formats.OMEROMetadataStore] (.Server-10) Unique detector settings: 0
2014-11-05 06:12:13,751 INFO  [          ome.formats.OMEROMetadataStore] (.Server-10) Unique logical channels: 9216
...
```

[...]

PR #4546, opened 2016-03-29
refactor DuplicateI to copy more complex subgraphs

mtbc:
`DuplicateI` failed in duplicating `ome/data_repo/curated/mias/frans/siRNA_PRIM1_03102008/001-365700055641`. This PR adjusts for Hibernate so as to,

* not rely on deferring flushes
* never persist an object whose properties contain transient instances
* never try to reorder an existing collection
* fill in back-references before persisting objects.

Fixes http://trac.openmicroscopy.org/ome/ticket/13171. Will probably conflict with regions so I will do that rebasing myself. (I have kept most graphs changes to regions, but this is a bugfix.)

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/DuplicationTest/ should continue to pass. Also feel free to try submitting `DuplicateI` requests yourself, including for the above plate. Graph structure should be duplicated *except* for anything relating to files on disk, such as thumbnails and pixel data: see https://trello.com/c/8GbXbGqx/6-duplicate-original-files.

### DIFF
--- a/components/blitz/src/omero/cmd/graphs/DuplicateI.java
+++ b/components/blitz/src/omero/cmd/graphs/DuplicateI.java
@@ -273,12 +273,12 @@ public class DuplicateI extends Duplicate implements IRequest, WrappableRequest<
         void execute(Function<Object, IObject> mapping) throws GraphException {
             final Object duplicateValue = GraphUtil.copyComplexValue(mapping, value);
             if (duplicateValue != null) {
-            try {
-                PropertyUtils.setNestedProperty(duplicate, property, duplicateValue);
-            } catch (NestedNullException | ReflectiveOperationException e) {
-                throw new GraphException(
-                        "cannot set property " + property + " on duplicate " + duplicate.getClass().getName());
-            }
+                try {
+                    PropertyUtils.setNestedProperty(duplicate, property, duplicateValue);
+                } catch (NestedNullException | ReflectiveOperationException e) {
+                    throw new GraphException(
+                            "cannot set property " + property + " on duplicate " + duplicate.getClass().getName());
+                }
             }
         }
     }
--- a/components/blitz/src/omero/cmd/graphs/DuplicateI.java
+++ b/components/blitz/src/omero/cmd/graphs/DuplicateI.java
@@ -19,7 +19,6 @@
 
 package omero.cmd.graphs;
 
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -36,7 +35,6 @@ import org.apache.commons.beanutils.NestedNullException;
 import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
-import org.hibernate.FlushMode;
 import org.hibernate.Hibernate;
 import org.hibernate.Session;
 import org.hibernate.proxy.HibernateProxy;
@@ -44,10 +42,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Function;
+import com.google.common.base.Functions;
 import com.google.common.base.Predicate;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
 import com.google.common.collect.SetMultimap;
 
 import ome.model.IObject;
@@ -106,13 +109,15 @@ public class DuplicateI extends Duplicate implements IRequest, WrappableRequest<
     private GraphHelper graphHelper;
     private GraphTraversal graphTraversal;
 
-    private FlushMode flushMode;
     private GraphTraversal.PlanExecutor processor;
 
     private int targetObjectCount = 0;
     private int duplicatedObjectCount = 0;
 
     private final Map<IObject, IObject> originalsToDuplicates = new HashMap<IObject, IObject>();
+    private final Map<Entry<String, Long>, IObject> originalClassIdToDuplicates = new HashMap<Entry<String, Long>, IObject>();
+    private final Multimap<IObject, PropertyUpdate> propertiesToUpdate = ArrayListMultimap.create();
+    private final SetMultimap<IObject, IObject> blockedBy = HashMultimap.create();
 
     /**
      * Given class names provided by the user, find the corresponding set of actual classes.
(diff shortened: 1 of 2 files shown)


mtbc:
throw travis dice

mtbc:
I've managed to figure out an integration test that usually fails without this PR and always passes with. It's built on top of a bunch of regions changes but it'll make it into OMERO 5.3.0. Key insight though: to make duplication fail without this PR, try to duplicate a graph with many multi-channel images whose channels all share one logical channel. The MIAS plate from this PR's description reproducibly suffices for me in this regard: its two-channel fields have a total of 32 channel instances all sharing the same logical channel.

mtbc:
Rebased this PR to #4553 on `regions` and in #4554 added `testDuplicatePlate` to http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/lastCompletedBuild/testReport/integration/DuplicationTest/ which would usually fail without this work.

components/blitz/src/omero/cmd/graphs/DuplicateI.java
```diff
@@ -530,7 +802,7 @@ public void adjustGraphPolicy(Function<GraphPolicy, GraphPolicy> adjuster) {
 
     @Override
     public int getStepProvidingCompleteResponse() {
-        return dryRun ? 0 : 5;
+        return dryRun ? 0 : 6;
```
joshmoore: How does the `6` here match with https://github.com/openmicroscopy/openmicroscopy/pull/4546/files#diff-5e2f58efa615b81632d4ab5b9186dd45R187 ?

mtbc: The significance is that the response is available before all the steps are complete. This isn't actually consumed by anything when we're not doing a dry run but it's conceivable that something may care someday!

ximenesuk:
I've checked this using a prototype command at: https://github.com/ximenesuk/openmicroscopy/commit/7d535006168f0f18ad954cd7070265041d31f1be

Basic duplicating looks fine, single objects, multiple objects and simple graphs (P/D/I). I tried a fake Plate with runs (`PW&plates=2&plateRows=1&plateCols=1&fields=1&plateAcqs=2.fake`)  before this PR:

```
colin$ omero duplicate Plate:1
omero.cmd.Duplicate Plate:1 failed: 'graph-fail'
	message=IllegalArgumentException occurred calling getter of ome.model.core.Pixels.channels
	stacktrace=org.hibernate.PropertyAccessException: IllegalArgumentException occurred calling getter of ome.model.core.Pixels.channels
	at org.hibernate.property.BasicPropertyAccessor$BasicGetter.get(BasicPropertyAccessor.java:198)

...
```

and then with this PR:
```
colin$ omero duplicate Plate:1
omero.cmd.Duplicate Plate:1 failed: 'graph-fail'
	message=not-null property references a null or transient value: ome.model.core.Channel._channelsBackref
	stacktrace=org.hibernate.PropertyValueException: not-null property references a null or transient value: ome.model.core.Channel._channelsBackref
	at org.hibernate.engine.Nullability.checkNullability(Nullability.java:101)
...
```

The dry-run a positive report is returned:
```
ls30778-2:openmicroscopy colin$ omero duplicate Plate:1 --dry-run --report
omero.cmd.Duplicate Plate:1 ok
Steps: 3
Elapsed time: 0.592 secs.
Flags: []
Duplicates
  Arc:1
  Detector:1
  DetectorSettings:1
  Dichroic:1
  Filament:2
  Filter:1,2
  FilterSet:1
  ImagingEnvironment:1,2
  Instrument:1
  Laser:4,5
  LightEmittingDiode:3
  LightPath:1
  LightPathEmissionFilterLink:1
  LightPathExcitationFilterLink:1
  LightSettings:2
  Microscope:1
  Objective:1
  StageLabel:1,2
  TransmittanceRange:1,2
  Channel:51-54
  Image:51,52
  LogicalChannel:51
  Pixels:51,52
  PlaneInfo:201,202
  ChannelBinding:1,2
  QuantumDef:1,2
  RenderingDef:1,2
  Experiment:1
  MicrobeamManipulation:1
  Plate:1
  PlateAcquisition:1,2
  Well:1
  WellSample:1,2
  StatsInfo:51-54
```

It may be that my plugin is not doing everything necessary since it was a quick hack based on the `delete` plugin.

mtbc:
Ah, I probably should have warned that fake plates are weird things anyway, e.g., https://trac.openmicroscopy.org/ome/ticket/13128#comment:5 -- though that issue's worth noting, and I may as well investigate while the work from this PR is fresh in my mind, I'm not inclined to prioritize such fixes without some clearer indication that they correspond to anything real.

mtbc:
(In short: your plugin is probably fine!)

ximenesuk:
@mtbc Thanks! It looks fine with a real Plate.

ximenesuk:
PS @joshmoore @sbesson it would be trivially easy to open the CLI `duplicate` comand as a PR if it would be useful - though it doesn't exploit any of the more advanced aspects of `DuplicateI`. Though would `copy` be a better (ie shorter) name or is that too loaded?

mtbc:
I'm not satisfied that the fake plate isn't revealing a genuine problem, despite its weirdness, so for now hold off on merging while I investigate.

mtbc:
@sbesson: For the fake plate of https://github.com/openmicroscopy/openmicroscopy/pull/4546#issuecomment-205359054 the two plates share one logical channel!

ximenesuk:
@mtbc does this suggest a bug in the fake reader? Have you tried the same in a build using BF 5.2?

mtbc:
@ximenesuk: Yes, it does the same using BF 5.2.

Okay, I'm satisfied: the problem is that, in duplicating one plate, it wants to duplicate the logical channel, so it duplicates the channels of that logical channel, except then it's left with channels without pixels because it's not duplicating the other plate's pixels. So, that seems reasonable. @ximenesuk, did you need to do any further testing, or is this now good to merge?

sbesson:
@mtbc @ximenesuk: still trying to decide whether the issue is at the `FakeReader` level or higher. Couple of thoughts:
- there is no more `LogicalChannel` concept in the model or even Bio-Formats and this is  OMERO-only
- looking at the OME-XML generated by the fake plates, I do not see any sharing of the `Channel` elements
- could we be hitting a consequence of the `LogicalChannel` collapsing for HCS data introduced in https://github.com/openmicroscopy/openmicroscopy/pull/3264 ?

mtbc:
@sbesson: I think you're right. Probably happens only when there are multiple plates in the same fileset. Could it happen for non-HCS MIFs then?

sbesson:
Maybe we should try to reproduce with 2 fake HCS files with one plate only imported separately to be sure

mtbc:
`PW&plates=1&plateRows=1&plateCols=1&fields=1&plateAcqs=2.fake` duplicates fine for me; want to cross-check @ximenesuk?

mtbc:
(I should have been clearer: I import it twice and then duplicate them both in the same request.)

ximenesuk:
@mtbc that workflow works for me. Duplicating a single such plate or several of them in the same command both work okay.

mtbc:
Thank you @ximenesuk.

mtbc:
(I'll check if there's an easy way to improve the error message in this case but that'll be for a separate PR.)

joshmoore:
:thumbsup:

sbesson:
Merging this in order to the conflicts between `develop` and `regions` sorted.

mtbc:
The problem reported in https://github.com/openmicroscopy/openmicroscopy/pull/4546#issuecomment-205359054 should be fixed by #4564.

PR #4548, opened 2016-03-30
Get ROI count

dominikl:
At the moment Insight loads all ROIs and counts them on the client side for showing the number of ROIs in the right hand side meta data panel. This PR adds a method for directly getting the number of ROIs via an HQL query.

**Test**: Check some images with ROIs, make sure the ROI count (right hand side metadata panel) is correct.

### DIFF
--- a/components/blitz/src/omero/gateway/facility/ROIFacility.java
+++ b/components/blitz/src/omero/gateway/facility/ROIFacility.java
@@ -22,18 +22,16 @@ package omero.gateway.facility;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
 import org.apache.commons.collections.CollectionUtils;
 
+import omero.RLong;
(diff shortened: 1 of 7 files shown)


dominikl:
Should I open it against regions branch? That was my initial intention, but then I thought, it might be good to get it into develop, that's why I openend the PR against develop for now.

mtbc:
Just be aware that if these are classes you're also changing on regions then there may well be awkward conflicts to resolve: not a showstopper, but takes extra effort. My policy lately has largely been to do stuff on develop only for bugfixes or for classes I don't expect us to touch on regions before `dev_5_2` is branched but conflicts aren't necessarily terrible, they just take a bit of effort and thought, up to you to judge if it's worth it.

components/blitz/src/omero/gateway/facility/ROIFacility.java
```diff
@@ -83,6 +80,45 @@
...
     }
+    
+    /**
+     * Get the number of ROIs for an image
+     * 
+     * @param ctx
+     *            The {@link SecurityContext}
+     * @param imageId
+     *            The image Id
+     * @return See above
```
mtbc: should probably mention it returns -1 in case of error

dominikl:
You were right, it conflicts with my regions PR. I'll close this PR and open it against regions. Can be back ported to develop later anyway, if needed.

dominikl:
Moved to regions branch, see #4550

PR #4551, opened 2016-04-01
refactor disk usage

mtbc:
Changes the `DiskUsage` request to use the graph machinery instead of its own different implementation of graph traversal. Behavior should be unchanged except for reporting attachments as `FileAnnotation` instead of `Annotation`. Note that this can be accessed from the CLI via the `bin/omero fs usage` subcommand (see the `--report` option).

### DIFF
--- a/components/blitz/resources/omero/cmd/FS.ice
+++ b/components/blitz/resources/omero/cmd/FS.ice
@@ -196,6 +196,7 @@ module omero {
          *   Folder, Screen, Plate, Well, WellSample,
          *   Image, Pixels, Annotation, Job, Fileset, OriginalFile.
          **/
+        ["deprecated:use omero::cmd::DiskUsage2 instead"]
         class DiskUsage extends Request {
             omero::api::StringSet classes;
             omero::api::StringLongListMap objects;
@@ -213,6 +214,7 @@ module omero {
          *   Thumbnail for the image thumbnails
          * The above map values are broken down by owner-group keys.
          **/
+        ["deprecated:use omero::cmd::DiskUsage2Response instead"]
         class DiskUsageResponse extends Response {
             omero::api::LongPairToStringIntMap fileCountByReferer;
             omero::api::LongPairToStringLongMap bytesUsedByReferer;
--- a/components/blitz/resources/omero/cmd/FS.ice
+++ b/components/blitz/resources/omero/cmd/FS.ice
@@ -185,41 +185,6 @@ module omero {
 
         };
 
-        /**
-         * Request to determine the disk usage of the given objects
-         * and their contents. File-system paths used by multiple objects
-         * are de-duplicated in the total count. Specifying a class is
-         * equivalent to specifying all its instances as objects.
-         *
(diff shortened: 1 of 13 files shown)


components/blitz/resources/omero/cmd/FS.ice
```diff
@@ -185,41 +185,6 @@ module omero {
...
-         * and their contents. File-system paths used by multiple objects
-         * are de-duplicated in the total count. Specifying a class is
-         * equivalent to specifying all its instances as objects.
-         *
-         * Permissible classes include:
-         *   ExperimenterGroup, Experimenter, Project, Dataset,
-         *   Folder, Screen, Plate, Well, WellSample,
-         *   Image, Pixels, Annotation, Job, Fileset, OriginalFile.
-         **/
-        class DiskUsage extends Request {
```
joshmoore: I have some small concern over the re-definition of these classes, i.e. giving `DiskUsage` a new parent class. It may be find in a major release, but this may be another case of bumping the class name to `DiskUsage2` since we've done that in this package for other classes.

mtbc: I was thinking that the API has changed very little; indeed, the policy rules should pretty much mirror what the previous did. Should I just rename it or also resurrect the previous as deprecated? (I suppose they could coexist, though of course this diff will then do even worse at showing the changes!)

joshmoore: Less worried about the API behavior than potential weirdnesses in Ice/Python that I'm missing. I'd generally tend toward a traditional `@Deprecate Class1` and add `Class2` strategy.

joshmoore:
A side question: how much overlap is there between the various rules files?

components/blitz/resources/omero/cmd/FS.ice
mtbc: Over a major version I can't see what the issue there would be but I'm happy to adjust anyway.

mtbc:
There is quite a lot of overlap between the rules files. Perhaps an interesting exercise one day might be to try to describe the relationships among them. For instance, chmod to a private group looks remarkably like moving data to a private group. If we add many more graph-based requests than the rule maintenance burden for model or permissions changes may become an issue. In this case, the disk usage rules are very much like the find-contained rules. The two exceptions that jump to my mind are:

* `DiskUsage` includes fileset entries if given just image IDs. `FindChildren` doesn't consider `FilesetEntry` to be a child of `Image`. I preserved the difference so as not to change the behavior of `DiskUsage`.

* `DiskUsage` has rules from `Experimenter` and `ExperimenterGroup` to include that which they own, again preserving previous behavior. `FindContainer` doesn't yet dare to touch the semantic question of how users and groups relate to containment.

mtbc:
throw the travis die

ximenesuk:
The functionality through the CLI as of this morning looks fine. The later commits don't change that. I'll leave @joshmoore  to check the code tomorrow.

joshmoore:
:thumbsup:

PR #4552, opened 2016-04-01
remove obsolete database check

mtbc:
Remove check for `pg_advisory_lock` which has been available since PostgreSQL 8.2.


rleigh-dundee:
Looks fine and good to merge.

Out of interest, were these already removed from the 5.0, 5.1 and 5.2 scripts?  I only see remaining usage in 4.4 and earlier.

joshmoore:
Interesting. It is in 5.2 but not 5.1 or 5.0. Looks like a conflict or similar may have have resurrected removed code in the past, but I don't think that it's 5.3-specific.

PR #4560, opened 2016-04-06
rewrite how to access information on Quartz cron expressions

mtbc:
Unfortunately the version of Quartz we use is so ancient that one must now dig somewhat to find documentation. Companion to https://github.com/openmicroscopy/ome-documentation/pull/1446.


dominikl:
Good to merge :+1:  Hope they keep the link to the last 1.x version on their downloads page.

PR #4561, opened 2016-04-07
Shape renaming: examples review

sbesson:
See https://trello.com/c/LW8l7U8x/20-shape-attribute-renaming-cleanup

This PR reviews the variable/methods used when retrieving Point and Ellipse attributes in the `examples` directory. In particular the output of the two following greps should return nothing:

```
$ git grep -I [sg]et[RrCc][xy] -- examples
$ git grep -I [^a-zA-Z][RrCc][XxYy] -- examples
```


dgault:
Renaming of the examples is in line with the cleanup trello card.
The git greps suggested do not return any instances and additional searches also show no further instances of rx, ry, cx or cy. Looks good to merge.

PR #4568, opened 2016-04-08
Fix potential class cast/NPE for unknown metadata key

dominikl:
Fixes a potential class cast and NPE if metadata contains an unknown entry.

**Test**: Import fake plate `PW&plates=2&plateRows=1&plateCols=1&fields=1&plateAcqs=2.fake`, select a well, view all metadata (i. e. expand all tabs) in right hand side metadata panel.


pwalczysko:
Works as expected. No error hit, neither iwth the plate in the header or with other fake plates tested, which were demonstrating the problem previously. 
Ready to merge.

PR #4567, opened 2016-04-08
Folders step 4

dominikl:
Implementation of https://github.com/openmicroscopy/design/issues/30

**Test**
- Make sure integration tests still pass
- Delete a folder which contains ROIs and sub folders: Make sure ROIs are orphaned and sub folders become top level folders
- Select a couple of ROIs, create a new folder via context menu: Make sure the selected ROIs are added to the newly created folder. Note: The newly created folder will be a top level folder.
- Select a ROI which is part of a Folder. Check Inspector table; the folder(s) should be mentioned there.
- Switch to Results table. Add "Folders"  to the display (via Result Wizard). Check that the folder(s) the ROIs are part of are displayed.

PR #4570, opened 2016-04-11
Ticket 13202

dominikl:
While investigating [Ticket 13202](https://trac.openmicroscopy.org/ome/ticket/13202#ticket) I fixed some problems with the ResultsTable of the MeasurementTool. This should also fix the error the user reported, although I couldn't reproduce it (but my suspicion is, it was some kind of unit conversion problem).

On that occassion I also made sure, that no unit up or down conversion ("round to reasonable unit") happens and that the ResultTable does not contain any unit symbols (numbers only),therefore the unit is shown in the headers of the table only (if it is a real unit and not in pixels). As far as I remember that was also requested in the past. The numbers in the table should all be of the same dimension so that they are easily comparable. This point might need some further discussion ( /cc @jburel @gusferguson ).

**Test**
- Make sure Coord_List(X) and Coord_List(Y) values are populated for Polygons, in the ResultsTable as well as the exported Excel file.
- Toggle between "calibrated" and "pixel value" display, check that the numbers are reasonable.

Note: I opened the PR agains regions, because this will certainly conflict with upcoming region PRs, although it's actually more a bugfix for develop branch.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementResults.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementResults.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.agents.measurement.view.MeasurementResults 
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


pwalczysko:
The change in the values of ``Center(X)`` and ``Center(Y)`` with this PR causes that Excel does not recognize these values as numbers anymore (does arrange them to the left-hand side of the cells, instead of, like in the develop branch insight, to the right). This causes that the numbers are not directly summable or no other direct operations with them are possible in Insight.
Compare in the screenshot below the situation with insight with this PR (above) and without it (below).

screen shot 2016-04-12 at 11 28 04

pwalczysko:
The funcitonality is fine as described in the header of this PR. The parameters are populated in Insight as well as in exported excel sheet, but https://github.com/openmicroscopy/openmicroscopy/pull/4570#issuecomment-208838188 shoiuld be reverted to previous behaviour imho.

dominikl:
I'll close it for now, due to the outstanding issue. Will re-open once #4567 has been merged (otherwise merge conflicts likely). Thanks @pwalczysko  for testing.

PR #4571, opened 2016-04-12
Re-enable saving of ExternalInfo (rebased onto dev_5_1)

joshmoore:
This is the same as gh-4532 but rebased onto dev_5_1.

----

At some point (either during the move to blitz or later
during the hiding of Details implementations), the copying
of ExternalInfo objects from one Details to the next was
lost. Since then, no ExternalInfo objects could have been
linked to any IObject, and it's even unlikely that they
could have been created at all.


joshmoore:
--rebased-from #4532

joshmoore:
cc: @chris-allan

joshmoore:
Following on from today's conversation, closing.

PR #4573, opened 2016-04-13
RoiTable DnD Scrolling

dominikl:
Workaround to enable auto scrolling when dragging items on the ROITable.

**Test**: Select a few ROIs somewhere in the middle of the table, try to drag them to a position which currently lies outside of the visible part of the table by moving the cursor close to the upper or lower edge of the table. Make sure the table automatically scrolls up or down. 

Note: **This might not work on OSX**. Might be fixed in newer JDK version (see http://mail.openjdk.java.net/pipermail/awt-dev/2013-September/005643.html ), but couldn't confirm yet. Suggest to test with Windows.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/ROITable.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/ROITable.java
@@ -26,8 +26,13 @@ package org.openmicroscopy.shoola.agents.measurement.view;
 
 //Java imports
 import java.awt.Component;
+import java.awt.Insets;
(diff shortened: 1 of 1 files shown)


dominikl:
Would be interesting to know if it works on OSX with newer Java version. Mine is `1.8.0_65-b17` and it doesn't. But for some reason I can't update Java at the moment.

dominikl:
Was able to update to latest Java (1.8.0-77), doesn't work. Apparently another JDK bug which never got fixed. Works only on Windows. Doesn't work on Linux either. And it introduces other problems... Will close it for now.

PR #4574, opened 2016-04-13
CLI duplicate command

ximenesuk:
This PR adds a basic duplicate command to the CLI, this should duplicate simple objects and hierarchies. It includes integration tests for simple objects.

## Testing
See:
```
$ omero duplicate --help
...
Examples:

    # Duplicate a dataset
    omero duplicate Dataset:50
    # Do the same reporting all the new duplicate objects
    omero duplicate Dataset:50 --report

    # Do a dry run of a duplicate reporting the outcome
    # if the duplicate had been run
    omero duplicate Dataset:53 --dry-run
    # Do a dry run of a duplicate, reporting all the objects
    # that would have been duplicated
    omero duplicate Dataset:53 --dry-run --report
```
Multiple IDs and objects should also work:
```
$ omero duplicate Dataset:263,264
$ omero duplicate Dataset:263 Dataset:264
```
In addition hierarchies (P/D/I for example) and other complex objects should be duplicated. Create a simple P/D/I structure or import a Screen, then:
```
$ omero duplicate Project:123 --report
$ omero duplicate Screen:321 --report
```
the `--report` option is useful here to check the graph duplication.

To do
-----

Once this PR is merged @pwalczysko plans to add further tests for multiple arguments, simple hierarchies (P/D/I) and complex objects (Screen).

The command itself needs improving to deal with the options in @mtbc's underlying command.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/duplicate.py
+++ b/components/tools/OmeroPy/src/omero/plugins/duplicate.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+   Startup plugin for command-line duplicates
+
+   Copyright 2009-2015 Glencoe Software, Inc. All rights reserved.
+   Use is subject to license terms supplied in LICENSE.txt
+
+"""
+
+import sys
+
+from omero.cli import CLI, GraphControl
+
+HELP = """Duplicate OMERO data.
+
+Duplicate entire graphs of data based on the ID of the top-node.
+
+Examples:
+
+    # Duplicate a dataset
+    omero duplicate Datset:50
+    # Do the same reporting all the new duplicate objects
+    omero duplicate Dataset:50 --report
+
+    # Do a dry run of a duplicate reporting the outcome
+    # if the duplicate had been run
+    omero duplicate Dataset:53 --dry-run
+    # Do a dry run of a duplicate, reporting all the objects
+    # that would have been duplicated
+    omero duplicate Dataset:53 --dry-run --report
+
+"""
+
+
+class DuplicateControl(GraphControl):
+
+    def cmd_type(self):
+        import omero
+        import omero.all
+        return omero.cmd.Duplicate
+
+    def print_detailed_report(self, req, rsp, status):
+        import omero
+        if isinstance(rsp, omero.cmd.DoAllRsp):
+            for response in rsp.responses:
+                if isinstance(response, omero.cmd.DuplicateResponse):
+                    self.print_duplicate_response(response)
+        elif isinstance(rsp, omero.cmd.DuplicateResponse):
+            self.print_duplicate_response(rsp)
+
+    def print_duplicate_response(self, rsp):
+        if rsp.duplicates:
+            self.ctx.out("Duplicates")
+            objIds = self._get_object_ids(rsp.duplicates)
+            for k in objIds:
+                self.ctx.out("  %s:%s" % (k, objIds[k]))
+
+try:
+    register("duplicate", DuplicateControl, HELP)
+except NameError:
+    if __name__ == "__main__":
+        cli = CLI()
+        cli.register("duplicate", DuplicateControl, HELP)
+        cli.invoke(sys.argv[1:])
--- a/components/tools/OmeroPy/src/omero/plugins/duplicate.py
+++ b/components/tools/OmeroPy/src/omero/plugins/duplicate.py
@@ -25,6 +25,7 @@
 """
 
 import sys
+import os
 
 from omero.cli import CLI, GraphControl
 
@@ -77,7 +78,8 @@ class DuplicateControl(GraphControl):
                 self.ctx.out("  %s:%s" % (k, objIds[k]))
 
 try:
-    register("duplicate", DuplicateControl, HELP)
+    if "OMERO_DEV_PLUGINS" in os.environ:
+        register("duplicate", DuplicateControl, HELP)
 except NameError:
     if __name__ == "__main__":
         cli = CLI()
--- a/components/tools/OmeroPy/src/omero/plugins/duplicate.py
+++ b/components/tools/OmeroPy/src/omero/plugins/duplicate.py
@@ -32,10 +32,14 @@ HELP = """Duplicate OMERO data.
 
 Duplicate entire graphs of data based on the ID of the top-node.
 
+Note that no underlying binary data will be duplicated. For example,
+a duplicated Image will refer to the same underlying image files on
+the file system as the original image.
+
 Examples:
 
     # Duplicate a dataset
-    omero duplicate Datset:50
+    omero duplicate Dataset:50
     # Do the same reporting all the new duplicate objects
     omero duplicate Dataset:50 --report
 
--- a/components/tools/OmeroPy/src/omero/plugins/duplicate.py
+++ b/components/tools/OmeroPy/src/omero/plugins/duplicate.py
@@ -33,9 +33,10 @@ HELP = """Duplicate OMERO data.
 
 Duplicate entire graphs of data based on the ID of the top-node.
 
-Note that no underlying binary data will be duplicated. For example,
-a duplicated Image will refer to the same underlying image files on
-the file system as the original image.
+Note that no object that corresponds to a file on disk will be duplicated.
+In some circumstances a duplicate object may reference an original object that
+does have associated binary data but a duplicated image should not be expected
+to include any pixel data.
 
 Examples:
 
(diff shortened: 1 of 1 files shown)


components/tools/OmeroPy/src/omero/plugins/duplicate.py
```diff
@@ -0,0 +1,81 @@
...
+from omero.cli import CLI, GraphControl
+
+HELP = """Duplicate OMERO data.
+
+Duplicate entire graphs of data based on the ID of the top-node.
+
+Examples:
+
+    # Duplicate a dataset
+    omero duplicate Datset:50
```
manics: typo (`Datset`)

ximenesuk: At least that is in my code and not @pwalczysko's! I noiced it when copying it to the PR description but thought I could sneak it through/leave it in case there are other changes.

components/tools/OmeroPy/src/omero/plugins/duplicate.py
```diff
@@ -0,0 +1,81 @@
...
+
+"""
+
+import sys
+
+from omero.cli import CLI, GraphControl
+
+HELP = """Duplicate OMERO data.
+
+Duplicate entire graphs of data based on the ID of the top-node.
```
joshmoore: I'd include a warning about non-duplication of binary data re: images.

ximenesuk: Thanks @joshmoore I was thinking about this yesterday. So a duplicated Image or OriginalFile will point to the same underlying file object?

joshmoore:
Probably the only other question from my side is whether or not to include this in 5.2.3 and/or hide it via `OMERO_DEV_PLUGINS` awaiting https://github.com/openmicroscopy/design/pull/31 discussions.

ximenesuk:
Reopening with @pwalczysko being identified as a contributor.

ximenesuk:
@joshmoore re https://github.com/openmicroscopy/openmicroscopy/pull/4574#issuecomment-209879124 I'd say it works as a stand-alone command even though it may be that another command eventually subsumes it. However, at the moment it is limited by being an all or nothing duplicate. I'd vote for allowing it in to 5.2.3 as a useful utility but, for now, hiding it from non-developers.

components/tools/OmeroPy/src/omero/plugins/duplicate.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def print_duplicate_response(self, rsp):
+        if rsp.duplicates:
+            self.ctx.out("Duplicates")
+            objIds = self._get_object_ids(rsp.duplicates)
+            for k in objIds:
+                self.ctx.out("  %s:%s" % (k, objIds[k]))
+
+try:
+    register("duplicate", DuplicateControl, HELP)
```
joshmoore: This still needs the DEV marker: see https://github.com/openmicroscopy/openmicroscopy/commit/affa1528bda834e06dba0d45f7988a9c8d539896

ximenesuk: I will add today - I hadn't realised you had :+1:ed my comment as I don't seem to get notifications of reactions.

components/tools/OmeroPy/src/omero/plugins/duplicate.py
mtbc: It'll refuse to attempt original files at all. It will not bother with anything that necessarily entails one.

ximenesuk: @mtbc could you please suggest a note that isn't a lie?

mtbc: "Note that no object that corresponds to a file on disk will be duplicated. In some circumstances a duplicate object may reference an original object that does have associated binary data but a duplicated image should not be expected to include any pixel data."

ximenesuk: Thanks!

mtbc:
`omero duplicate --help` doesn't seem to work. Is there a special trick since the `OMERO_DEV_PLUGINS` change?

ximenesuk:
Have you added `OMERO_DEV_PLUGINS` to your env?

mtbc:
No. What value do I put in it? (Is this documented anywhere?)

ximenesuk:
No value, jus `export OMERO_DEV_PLUGINS=` will do.

mtbc:
Goodness, so it does, thank you!

ximenesuk:
I don't think there is any documentation yet other than its use in

https://github.com/openmicroscopy/openmicroscopy/commit/affa1528bda834e06dba0d45f7988a9c8d539896

/cc @joshmoore

manics:
It's for dev/internal use, so we probably don't want to document it.

sbesson:
Tested using the daily DEV merge build. All worked as expected both with and without `--dry-run` mode. Also tested in the permissions context while duplicating other's data. Integration tests are passing. This is ready to merge.

On the `OMERO_DEV_PLUGINS` topic, I would be inclined to list it under [ENV_HELP](https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.2/components/tools/OmeroPy/src/omero/cli.py#L397) with sufficient warnings about its usage  but I can see @manics point about not publicizing it. Looking at the list of CLI plugins, others could be classified under this category as well (`perf`, `testengine`) if not even deprecated.

joshmoore:
> (perf, testengine) if not even deprecated.

This is beyond this PR but as described in https://trello.com/c/tVSjHnee/55-add-warnings-for-windows-end-of-support there are several plugins (perf, testengine, submit) which should be deprecated and removed. In comparison, childre, duplicate, render, are "experimental" and the API will be decided on soon for inclusion. (Another way of doing this would be to have them in a separate repo and versioned low, e.g. 0.1.0, and then install them via pip in a virtualenv.)

manics:
> On the OMERO_DEV_PLUGINS topic, I would be inclined to list it under ENV_HELP with sufficient warnings about its usage but I can see @manics point about not publicizing it.

The alternative is to not include these plugins in the mainline, so I think it's best to treat them as non-existent. Originally they were only in the `metadata` branch, but since they require changes to other files outside of `plugins` this would lead to conflicts as the two branches diverged even more. We could add just a subset of files but this makes things more complicated, and difficult to test.

joshmoore:
> On the OMERO_DEV_PLUGINS topic, I would be inclined to list it under ENV_HELP ...

:+1: Minimally that should help the team remember what the secret invocation was. (Can be external to this PR)

components/tools/OmeroPy/test/integration/clitest/test_duplicate.py
```diff
@@ -0,0 +1,260 @@
...
+        return self.query.findAllByQuery(query, params)
+
+    def get_project(self, did):
+        """Retrieve all the parent projects linked to the dataaset"""
+
+        params = omero.sys.ParametersI()
+        params.addId(did)
+        query = ("select p from Project as p where exists"
+                 "(select l from ProjectDatasetLink as l where "
+                 "l.child.id=:id and l.parent=p.id)")
```
mtbc: this might simply be something like,
`SELECT DISTINCT parent FROM ProjectDatasetLink WHERE child.id = :id`

ximenesuk: Thanks for the suggestion @mtbc, @pwalczysko may have adapted his query from elsewhere in our codebase?

pwalczysko: As @ximenesuk suggested, I indeed did adapt this code from the import test https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_import.py#L189. 
@mtbc : Would it be okay to merge this as it is and return to the cleanup of the queries in the tests later ?

mtbc: Sure, not blocking, just commenting. Be warned that there are some quite exciting examples of HQL around our codebase. :smiley:

jburel:
Reading the comments, especially https://github.com/openmicroscopy/openmicroscopy/pull/4574#issuecomment-212114155
it looks like we need to scope "pip install plugins", we will need the same for web. This could be done under ``Facilities/Institutions management``. Card created https://trello.com/c/KUGBzOdm/22-pip-install-plugins

Better to hold off for 5.2.3

mtbc:
What more needs doing here -- review tests?

joshmoore:
My understanding is that this was just waiting for 5.2.3 to be released.

ximenesuk:
Yes, that was my understanding too. If any further tests were to be needed @pwalczysko would open a PR directly agains develop once this is merged,

pwalczysko:
So should this be relisted ? Would give me a boost if this could go in really :)

joshmoore:
Well in that case .... merging. I will organize a general discussion on the de-DEV-PLUGIN-ification in the summer as we try to get the IDR work back on the mainline. Thanks all!

PR #4575, opened 2016-04-13
Fix saving composite ROIs from ImageJ

stelfrich:
This fixes an issue of composite ROIs not being properly imported from ImageJ to OMERO (somewhat related to #4046).

Should add two rectangular ROIs to OMERO (because composite ROIs are not supported, afaik?) but doesn't add any ROI (tested with `develop`):

1. Open sample image (e.g. Dot blot)
2. Add rectangular ROI
3. Add second rectangular ROI while holding down shift (thus creating a composite ROI)
4. Add to overlay (`Ctrl+B`)
4. Plugins > OMERO > Save Images to OMERO

**Note**: 71c7ca8 contains an optional refactoring.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/ROIReader.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/ROIReader.java
@@ -344,8 +344,15 @@ public class ROIReader {
             } else if (r instanceof ShapeRoi) {
                 Roi[] subRois = ((ShapeRoi) r).getRois();
                 Roi shapeij;
+                
                 for (int j = 0; j < subRois.length; j++) {
-                    shapeij = subRois[j];
+                	roiData = new ROIData();
+                    type = r.getTypeAsString();
+                    if (imageID >= 0) {
+                        roiData.setImage(new ImageI(imageID, false));
+                    }
+                    pojos.add(roiData);
+                	shapeij = subRois[j];
 
                     // Set ImagePlus reference in subROIs for the check in L216 to work
                     ImagePlus imp = r.getImage();
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/ROIReader.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/ROIReader.java
@@ -342,65 +342,9 @@ public class ROIReader {
                     }
                 }
             } else if (r instanceof ShapeRoi) {
-                Roi[] subRois = ((ShapeRoi) r).getRois();
-                Roi shapeij;
+                Roi[] subRois = getProperSubRois((ShapeRoi) r);
                 
-                for (int j = 0; j < subRois.length; j++) {
-                	roiData = new ROIData();
-                    type = r.getTypeAsString();
-                    if (imageID >= 0) {
-                        roiData.setImage(new ImageI(imageID, false));
-                    }
-                    pojos.add(roiData);
-                	shapeij = subRois[j];
-
-                    // Set ImagePlus reference in subROIs for the check in L216 to work
-                    ImagePlus imp = r.getImage();
-                    shapeij.setImage(imp);
-                    // Transfer correct ROI positions (according to IJ) from superROI
-                    int pos = r.getPosition();
-                    int c = r.getCPosition();
-                    int z = r.getZPosition();
-                    int t = r.getTPosition();
-                    if (imp.getNChannels() == 1 && imp.getNSlices() == 1) {
-                        shapeij.setPosition(pos);
-                    } else if (imp.getNChannels() == 1 &&
-                            imp.getNFrames() == 1) {
-                        shapeij.setPosition(pos);
-                    } else if (imp.getNSlices() == 1 &&
-                            imp.getNFrames() == 1) {
-                        shapeij.setPosition(pos);
-                    } else if (imp.isHyperStack()) {
-                        shapeij.setPosition(c, z, t);
-                    }
-                    type = shapeij.getTypeAsString();
-                    if (shapeij instanceof Line) {
-                        shape = convertLine((Line) shapeij);
-                        if (shape != null) {
-                            roiData.addShapeData(shape);
-                        }
-                    } else if (shapeij instanceof OvalRoi) {
-                        shape = convertEllipse((OvalRoi) shapeij);
-                        if (shape != null) {
-                            roiData.addShapeData(shape);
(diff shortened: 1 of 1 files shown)


components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/ROIReader.java
```diff
@@ -342,58 +342,9 @@ private void setImage(Roi[] rois)
...
-                    } else if (imp.getNChannels() == 1 &&
-                            imp.getNFrames() == 1) {
-                        shapeij.setPosition(pos);
-                    } else if (imp.getNSlices() == 1 &&
-                            imp.getNFrames() == 1) {
-                        shapeij.setPosition(pos);
-                    } else if (imp.isHyperStack()) {
-                        shapeij.setPosition(c, z, t);
-                    }
-                    type = shapeij.getTypeAsString();
```
bramalingam: @stelfrich : The ROI type conversion logic seems to be removed. Currently you are just populating the subROI with the position alone. You will have to invoke the conversion logic, to actually create the corresponding omero subROI's. Due to this issue, the testing workflow is failing (silently) without creating any new ROI's on the imported image.

stelfrich: @bramalingam: I couldn't find a difference between the handling of ROIs and their conversion and the handling of subROIs. Hence, I have refactored the code to a recursive call to `read()` with the subROIs as parameters in L347.

My "testing" workflow properly imports (ImageJ) composite ROIs as individual ROIs to OMERO. Am I missing something completely obvious here, @bramalingam?

bramalingam: Ahh apologies, I did not see the read() method properly. And I was assuming that the composite ROI's are similar to Overlays, and thus replicated your testing workflow without doing "Add to ROI-Manager" : thus the lack of ROI's in the resultant image.

Corrected the workflow now and I can see the resultant ROI's. 

But one small issue, there seems to be a ghost (invisible) ROI as well. I drew six roi's (composite) in my workflow but I keep getting a Seventh ghost ROI (which is invisible in the clients and the count seems to be 1 more than the expected number in the Omero side). This happens only for Composite ROI's and only on the omero Side (Please Check ROI Count for the Image in the omero clients), and not for the regular ROI's or Overlay workflows. Do you see this in your testing workflow as well? If its an issue at our end alone, I can dig deeper and troubleshoot the issue in better detail.

Reopening the image on ImageJ gives me just 6 ROI's, so I just wanted to check with you on this regard, before digging deeper into the issue.

stelfrich:
> If its an issue at our end alone, I can dig deeper and troubleshoot the issue in better detail.

I don't get any additional ghost ROIs with my test workflow and my configuration, @bramalingam. That is, an OMERO.insight-ij version cut from this PR together with an OMERO 5.2.1 virtual appliance on my machine. After the import from an ImageJ overlay, I have verified the ROI count to be correct using OMERO.insight 5.2.2.

If I can be of any help resolving that issue, just let me know.

jburel:
@bramalingam: I have understood why you could have a "ghost" roi.
A roi w/o shape was always added.
Note that the change made in the PR implies that for each sub-roi shape an roidata object will be created i.e. we lose the grouping of sub-roi. Previously the sub-roi were converted into shape then added to a single roi data.
@stelfrich correct me if I am wrong, but I understood that the previous behaviour was the correct one

stelfrich:
> @stelfrich correct me if I am wrong, but I understood that the previous behaviour was the correct one

Now that you have typed it out: Yes. While it might be the intended behaviour, it wasn't working for me (that is what got me started on this PR).

I have just tested the workflow mentioned in my initial comment with the latest release of OMERO.insight-ij. This workflow results in only one ROI (correct) with one shape (**should be two shapes**) in OMERO. After all, this PR might just be working around a bug in `ROIData.addShapeData()` (although we might keep the refactoring to remove code duplication).

So I did some more digging (sorry for not doing that earlier):
- `ROIData.addShapeData()` creates a new `List<ShapeData>` for every call if `Shape.getROICoordinate() == null` (see [ROIData#L164](https://github.com/openmicroscopy/openmicroscopy/blob/3d1c97a46f7f42ec3acb9c9f3e7521935382e623/components/blitz/src/omero/gateway/model/ROIData.java#L164))
- This seems to be the default case in my tests because `ShapeData.getT()` returns `-1` due to `shape.getTheT()` returning `null` (see [ShapeData#L342](https://github.com/openmicroscopy/openmicroscopy/blob/3d1c97a46f7f42ec3acb9c9f3e7521935382e623/components/blitz/src/omero/gateway/model/ShapeData.java#L342))
- I vaguely remember some changes that I have proposed about a year ago to the mapping of ImageJ ROI positions and OMERO ROI positions
- This also applies for an adapted workflow using ImageJ's RoiManager

@bramalingam @jburel Can you confirm my findings?

jburel:
A list will be created for a given coordinate only if there is not already a list associated to it.
As you noticed, the roi will have a collection of shapes (coordinates not required) but OMERO.insight does not correctly handle roi with multiple shapes on the same plane or roi with multiple shapes  w/o specified plane (i.e. the shapes should be on all t if t is not set)
That is a limitation of the existing desktop client, this is why you will not be able to view correctly the  shapes.
The changes you implemented (minus the extra empty roi) will act as a workaround but the grouping will be lost. 
This could be potential problem when re-opening in imageJ.

stelfrich:
> This could be a potential problem when re-opening in imageJ since the grouping info will be lost.

That's a good point. I just checked how the current implementation handles that part: It looses the grouping information as well. That is, the individual shapes of a composite ROI are imported to ImageJ as individual ROIs.

> That is a limitation of the existing desktop client, this is why you will not be able to view correctly the shapes.

Okay, that explains a lot.. How do you want to proceed, @jburel? Should I open a new issue with that insight problem? Also, I could investigate the OMERO to ImageJ discrepancies regarding composite ROIs a little more and propose a PR? If so, we should be good to close this PR..

jburel:
The problem in the desktop client will require a major rework of the measurement tool
If you could investigate the OMERO to ImageJ discrepancies regarding composite ROIs a little more, that will be great.
Close this PR and open a new one when ready (the refactoring one will good to have so maybe cherry-pick the commits)

jburel:
Closing this PR, this can re-open when  @stelfrich is done with the investigation

PR #4576, opened 2016-04-13
Shape renaming: insight review

sbesson:
See https://trello.com/c/LW8l7U8x/20-shape-attribute-renaming-cleanup

This PR reviews the variable/methods used when retrieving Point and Ellipse attributes in the Insight component. In particular the following grep should return nothing

```
$ git grep -I [sg]et[RrCc][xy] -- examples
```

and the output of the following one should be reviewed:

```
$ git grep -I [^a-zA-Z][RrCc][XxYy] -- components/insight
```

All remaining usages of the variables above should either used in the context of JHotDraw or SVG  elements or have a different meaning e.g. ratio along x.


dominikl:
Looks alright, also didn't notice any problems in Insight. :+1:

PR #4577, opened 2016-04-14
Workaround for DnD autoscroll

dominikl:
Workaround to enable auto scrolling when dragging items on the ROITable.

Test: Select a few ROIs somewhere in the middle of the table, try to drag them to a position which currently lies outside of the visible part of the table by moving the cursor close to the upper or lower edge of the table. Make sure the table automatically scrolls up or down.


pwalczysko:
The autoscroll definitely is better. There is a "sweet spot" now, which, when the mousepointer is kept over it, causes a smooth autoscroll. I did not notice any difference in this between Mac and Win.
I stil had cases though when 5 ROIs after a very long scroll and apparently succesful drop into a folder did not appear in the new folder. This does not repeat every time though.
I think this PR is an improvement, so I would recommend to merge and investigate the DnD behaviour further as we progress.

PR #4583, opened 2016-04-18
Cut/Copy

jburel:
Change cut/copy for orphaned objects. See https://trello.com/c/Frsb7uGR/49-cut-copy-orphaned-objects
To test:
 * Select an orphaned image/dataset/plane
 * Check that the Copy link is enabled instead of Cut Link

cc @will-moore


dominikl:
Good to merge :+1:

PR #4579, opened 2016-04-15
Right panel annotation tabs

will-moore:
This PR updates the right panel annotations panes so that they don't load data unless expanded.
Also data is loaded via json instead of html.
Batch annotation has not been changed yet.

This will need a lot of testing with all the different annotations and permissions. Not ready for full testing yet. More cleanup work to do, but this is stable enough to open a PR now...

 - [x] Handle all "custom" types of annotation (boolean, xml, long, etc)
 - [x] Remove listAnnotations() from container.py. Leave for now - used by annotate_tags()
 - [x] Tests for tree.py marshal_annotations()
 - [x] Split ome.general_tab.js into separate files for each Annotation Pane.
 - [x] Move js for each annotation Pane out of metadata_general.html into separate classes for each Pane.
 - [x] Filter annotations by user (also in batch_annotation)
 - [x] Date formatting
 - [ ] Long text E.g. ns on tooltips.

To test, compare what's shown in web with Insight - to check correct annotations / permissions etc.
**For all panes**:
 - Check annotations are same as Insight
 - Check tooltips
 - Filter by owner (mine/others)
 - When pane is expanded, it should stay expanded when selecting new P/D/I etc
 - All annotations should take Well-Index into account - E.g. Select Well with index is 2 - add annotations & check that they are shown just on this well & index.
 - Permissions: when can / cannot annotate, check that buttons etc reflect this.
 - Permissions: check you can remove / delete your own / others' annotations as expected.
 - Add and remove annotations - check that display is updated as expected.
 - As discussed with @pwalczysko: test if fast selection changes can load incorrect content into an Annotation pane. E.g. Expand 'tag' pane, select a "slow-to-load" pane (object with lots of tags) followed rapidly by a "fast-to-load" pane. IF the slow content arrives AFTER the fast content, it *might* replace it (incorrectly) and be displayed for the second object.

**Batch annotate**
 - Same panes should stay expanded/collapsed when switching between singe/batch annotate.
 - Tags summarised as before (details in tooltip) - Clicking X removes all links that you can unlink, including those linked by others.

**Specific annotations**
 - Tags:
    - NB: we still show separate tags for each user (as before) instead of summarise them like Insight.
    - Now use a '-' icon for remove instead of 'x' [see ticket](https://trac.openmicroscopy.org/ome/ticket/12867)
 - Key Value Pairs
    - Create & Edit
    - Check display of non "client-editable" map annotations
 - Attachments
    - Check 'scripts' button works (run script with checked files)
 - Rating
[...]

PR #4585, opened 2016-04-18
automatically delete orphaned non-sharable annotations

mtbc:
To assist http://trac.openmicroscopy.org/ome/ticket/2999 in preventing orphaned non-sharable annotations littering databases and in making the graph policy rules more consistent this PR ensures that when a non-sharable annotation is orphaned by being unlinked it is also deleted.

Testing: ensure that http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-test-integration/ is blue. Also, try deleting different things from different clients (e.g., comments, ratings, attachments, images, whatever, from Insight and Web) to watch out for regressions in deletion.

### DIFF
--- a/components/blitz/resources/ome/services/graph-rules/blitz-chgrp-rules.xml
+++ b/components/blitz/resources/ome/services/graph-rules/blitz-chgrp-rules.xml
@@ -268,9 +268,11 @@
         <bean parent="graphPolicyRule" p:matches="!$to_private, L:IAnnotationLink.parent = X:[I], L.child = A:Annotation[D]/d"
                                        p:changes="A:[I]"/>
 
-        <!-- Delete remaining orphaned annotations. -->
+        <!-- Delete orphaned annotations, ignoring permissions for BasicAnnotation and CommentAnnotation. -->
 
         <bean parent="graphPolicyRule" p:matches="A:Annotation[E]{o}/d" p:changes="A:[D]"/>
+        <bean parent="graphPolicyRule" p:matches="A:BasicAnnotation[E]{o}" p:changes="A:[D]/n"/>
+        <bean parent="graphPolicyRule" p:matches="A:CommentAnnotation[E]{o}" p:changes="A:[D]/n"/>
 
         <!-- If an original file is moved then any corresponding file annotation must also be moved. -->
 
--- a/components/blitz/resources/ome/services/graph-rules/blitz-chmod-rules.xml
+++ b/components/blitz/resources/ome/services/graph-rules/blitz-chmod-rules.xml
@@ -98,9 +98,9 @@
 
         <!-- Delete orphaned annotations, ignoring permissions for BasicAnnotation and CommentAnnotation. -->
 
+        <bean parent="graphPolicyRule" p:matches="A:Annotation[E]{o}/d" p:changes="A:[D]"/>
         <bean parent="graphPolicyRule" p:matches="A:BasicAnnotation[E]{o}" p:changes="A:[D]/n"/>
         <bean parent="graphPolicyRule" p:matches="A:CommentAnnotation[E]{o}" p:changes="A:[D]/n"/>
-        <bean parent="graphPolicyRule" p:matches="A:Annotation[E]{o}/d" p:changes="A:[D]"/>
 
         <!-- If an annotation link's parent or child is deleted then delete the link regardless of permissions. -->
 
--- a/components/blitz/resources/ome/services/graph-rules/blitz-delete-rules.xml
+++ b/components/blitz/resources/ome/services/graph-rules/blitz-delete-rules.xml
@@ -191,10 +191,8 @@
 
         <!-- If a basic or comment annotation is unlinked then consider it for deletion regardless of permissions. -->
 
-        <!-- omitted at present because it causes errors in the web client
         <bean parent="graphPolicyRule" p:matches="IAnnotationLink[D].child = A:BasicAnnotation[E]{i}" p:changes="A:{r}"/>
         <bean parent="graphPolicyRule" p:matches="IAnnotationLink[D].child = A:CommentAnnotation[E]{i}" p:changes="A:{r}"/>
-          -->
 
         <!--
              If an annotated object is deleted then consider its basic or comment annotations for deletion regardless of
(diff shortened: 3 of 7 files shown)


mtbc:
DNS issues today. CI is at http://10.2.1.77:8080/ and OMERO at 10.2.1.85 4064.

pwalczysko:
Unfortunately Insight does not connect for me today to regions server. @mtbc is also tellling me the server is not responding. Lets relist for tomorrow.

mtbc:
@manics and @sbesson and @aleksandra-tarkowska have fixed regions devspace. :+1:

sbesson:
All integration tests passed successfully and changes look plausible (most of the diff is actually related to the using of the new builders). I have not tested the Web client extensively so this PR should be relisted for functional testing using both clients. /cc @pwalczysko Is there any code  in Insight that could be simplified by this PR similarly to https://github.com/openmicroscopy/openmicroscopy/pull/4585/commits/eab13c50d1ce122ee0b0a07171047f7a76cc5318  @dominikl?

sbesson:
A more general question looking at the graph rules in more details: what is the rationale for treating comment and basic orphaned annotations (timestamp, numeric...) differently than other orphaned annotations?

mtbc:
Yes, functional testing would be good indeed.

Good question about rationale. To some extent I simply appeal to the ticket, of course, combined with the mentions of non-sharable and sharable annotations around our code; in general there are special rules about different kinds of annotation that to some extent are suggested by patterns in both old documentation/tickets and integration tests. At the one end are things like comments that appear to be considered to be quite ephemeral and at the other are tags which are certainly not; this may be a spectrum of "sharability" or, oppositely, "dispensability", and is not explicit in the OME-XML schema. When I ask about intermediate kinds like MapAnnotation or XmlAnnotation people usually aren't very sure and the graph policy rules are thus commensurately in-between about those.

With very much of the graph policy stuff I've taken my cue from an impression accumulated from snippets from tickets, code, etc. and conversations with @jburel. I've tried to stimulate some clarification -- for example, relevant to your question, with https://trello.com/c/T3UX9YZe/340-annotations-and-permissions I capture the perhaps-more-recent opinion,

> There is a thought that permissions policy on what should be done with annotations when permissions grow tighter should be more about the namespace that they are in than about what kind of annotation they are.

but I get the impression that the original rationale is captured in workflow and design discussions that rather predate me. They might be grounded more in how users were expected to use the graphical clients than in a server-side view.

In short: I'm not sure what the rationale is but I think there is, or at least was, one. Further, this is a specific facet that is fairly typical of many of the other graph policy decisions: the behavior "seems to fit" but if we had more time and energy and stepped back to rethink it all with the current UX team then that might well cause us to change aspects of how things work. In many cases I made precedent-based educated guesses about the edge cases while also trying to use what people were more certain of.

pwalczysko:
sorry, did not manage any testing here today,  https://github.com/openmicroscopy/openmicroscopy/pull/4579 takes lot of resources

pwalczysko:
Still occupied by #4579 . Started testing with insight on regions server, but I do not think I have a 5.3x web server anywhere to play with , @sbesson ?

sbesson:
@pwalczysko: the Web client should be up and running - see standup links.

pwalczysko:
No regressions observed, neither from Insight nor from Web.
Deleted P/D/I (with annotations), ROIs, comments, file attachments, tags, Map Annot (as far as these can be deleted from the UI). All went fine.

sbesson:
Depending on the scoping on production DB annotations distribution, this deletion rule might need to be reviewed/adjusted - see conversation of https://github.com/openmicroscopy/openmicroscopy/pull/4608. How easy would it be to include namespace criteria for instance?

mtbc:
It's certainly possible but not trivial. I have a checklist item for that on https://trello.com/c/zNzeSGlw/429-expand-graph-traversal-rules. So, we then have the overhead of querying every linked annotation's namespace to check if it's a match, and for each circumstance for each of the graph operations in which they delete an annotation (e.g., moving an image to a private group and somebody else made the comment) we need to decide what to do with the annotation, and adjust the policy rules accordingly. (Also is there any other special treatment required for certain annotations, e.g., regarding chgrp and chown?)

(If I adjust graph traversal so that namespace matches are possible at all, maybe somebody else can adjust the rules as required and add integration tests accordingly!)

mtbc:
(Of course, 5.0 didn't do anything special for these either.)

mtbc:
Obvious question is: Will such special annotations ever be affected by this? E.g., are they sometimes actually linked to something and then orphaned, or are they always orphans from the start?

mtbc:
@aleksandra-tarkowska, @manics: this PR enables deletion of orphaned map annotations. Also see https://github.com/openmicroscopy/openmicroscopy/pull/4907.

joshmoore:
Coming back to https://github.com/openmicroscopy/openmicroscopy/pull/4585#issuecomment-214523306 -- I wonder to what extent a _sharability_ `trigger` would be appropriate.

manics:
--rebased-to #5068

manics:
I think this suffers from poor performance when one MapAnnotation is linked to several objects, and just one of those links is deleted: https://github.com/openmicroscopy/openmicroscopy/pull/5068#issuecomment-277303948

mtbc:
It's certainly going to be slower mostly because it now actually looks at those basic/comment annotations and their linkages at all rather than just ignoring them as it did previously. Should still perform far better than similar would have under OMERO 4 though. :smiley:

joshmoore:
Not just slower as I understood it @mtbc: the server OOM'd.

mtbc:
Ah, no hint of that in the linked comment: that *would* be surprising/disappointing given the lengths the graph traversal goes to to avoid actually hydrating Hibernate proxies.

joshmoore:
```
Exception: object #0 (::omero::cmd::ERR)
{
    category = ::omero::cmd::Delete2
    name = graph-fail
    parameters = 
    {
        key = stacktrace
        value = java.lang.OutOfMemoryError: GC overhead limit exceeded
	at java.util.HashMap$KeySet.iterator(HashMap.java:916)
	at java.util.HashSet.iterator(HashSet.java:172)
	at com.google.common.collect.AbstractMapBasedMultimap$Itr.next(AbstractMapBasedMultimap.java:1148)
	at java.util.AbstractCollection.addAll(AbstractCollection.java:343)
	at java.util.HashSet.<init>(HashSet.java:119)
	at ome.services.graphs.GraphTraversal.getLinksToCache(GraphTraversal.java:884)
	at ome.services.graphs.GraphTraversal.cache(GraphTraversal.java:933)
	at ome.services.graphs.GraphTraversal.planOperation(GraphTraversal.java:616)
	at ome.services.graphs.GraphTraversal.planOperation(GraphTraversal.java:513)
	at omero.cmd.graphs.Delete2I.step(Delete2I.java:146)
	at omero.cmd.HandleI.steps(HandleI.java:438)
	at omero.cmd.HandleI$1.doWork(HandleI.java:366)
	at omero.cmd.HandleI$1.doWork(HandleI.java:362)
...     
        key = message
        value = GC overhead limit exceeded
    }
}
```

joshmoore:
Moving conversation to https://trello.com/c/mfW0N1an/272-poor-performance-when-deleting-many-one-annotation-links

PR #4586, opened 2016-04-18
Tag/File annotations

jburel:
Synch UI with web so it is not possible to annotate tag or file.

 * Go to the Tags
 * Select a tag
 * Check that the tag cannot be annotated in the right-hand panel
cc @will-moore


dominikl:
Good to merge :+1:

PR #4587, opened 2016-04-18
cyclic tag sets

jburel:
see https://trac.openmicroscopy.org/ome/ticket/13040
To test:
 * create via cli cyclic tag sets
 * Open insight and go the tag tab

cc @mtbc


mtbc:
Works excellently: tried both cyclic and reflexive tag sets, no problems at all. Good to merge.

PR #4588, opened 2016-04-19
Plate into screen

jburel:
See https://trac.openmicroscopy.org/ome/ticket/12949

To test:
 * Open the importer
 * Select a plate
 * Click the screens tab
 * Create a new screen
 * Add plate to the queue
 * Check that the plate is imported in the screen


mtbc:
Steps 9 and 12 of the ticket description are now fixed. Import otherwise seems to work just fine. Good to merge.

PR #4565, opened 2016-04-07
warn about Windows deprecation

aleksandra-tarkowska:
Following https://github.com/openmicroscopy/ome-documentation/pull/1437 web plugin should also warn

To test:
- `bin\omero web -h`
- `bin\omero web help`

try other commands from help
- `bin\omero web config ...`

they should contain:

```
bin\omero web config nginx
WARNING: Support for Windows will be removed in OMERO 5.3, see http://blog.openmicroscopy.org/tech-issues/future-plans/deployment/2016/03/22/windows-support/
```

cc: @jburel

PR #4589, opened 2016-04-19
Warn of Windows deprecation in fs plugin.

ximenesuk:
As the fs plugin does deal with file level operations, particularly `fs rename` it probably deserves a warning. I've added it at the top level though it may be more warranted at the `rename` level? (Though maybe the plugin is being used remotely?)

The deprecation warning should appear when `omero fs --help` or `omero fs -h` is used on Windows.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -23,6 +23,7 @@
 fs plugin for querying repositories, filesets, and the like.
 """
 
+import platform
 import sys
 
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


aleksandra-tarkowska:
This message should also being displayed for each running command

ximenesuk:
I've added the warning to the only relevant command, the admin-only `rename`.

components/tools/OmeroPy/src/omero/plugins/fs.py
```diff
@@ -435,6 +444,9 @@ def rename(self, args):
 template. By default the original files and import log are also
 moved.
 """
+        if platform.system() == 'Windows':
```
aleksandra-tarkowska: you could use ``self._isWindows()``

ximenesuk: I was using the same condition as on of the other examples I looked at but happy to chnage if there is a preferred style. /cc @jburel

jburel: I will update my PR too

aleksandra-tarkowska: @ximenesuk that is true for the top level check but within a class you can access, see https://github.com/openmicroscopy/openmicroscopy/pull/4565/files#diff-ad256e5e06d156adf521e47a4cdd8c37R68.

My comment was only suggestion, no need to be changed. I am planning to make `_isWindows` public that we will be able to use it in init as well (next PR once all of these warnings are merged).

aleksandra-tarkowska:
as discussed cherry-picked to https://github.com/openmicroscopy/openmicroscopy/pull/4601 can be closed now

PR #4590, opened 2016-04-19
Transform examples

jburel:
Remove transform from example
Code review
tested by training job


dominikl:
Looks alright, training job passes, too. Good to merge :+1: 
Why is the call to `setTransform` actually removed, just because it's not necessary?

jburel:
The examples did not expose very much of the usage. (mainly null) and the no usage in python nor matlab example.

sbesson:
See https://github.com/openmicroscopy/design/issues/33

PR #4592, opened 2016-04-19
warnings db plugin

jburel:
The deprecation warning should appear when --help or  -h is used on Windows.


joshmoore:
I was going to say that `db.py` is platform non-specific since since it only generates SQL, but it does have a dependency on the server jars. Running the unit tests on a future OMERO.cli.zip might be instructive!

aleksandra-tarkowska:
as discussed cherry-picked to #4601 can be closed now

jburel:
closing in favour of #4601

PR #4594, opened 2016-04-20
Session test

jburel:
Add test to check if the ```_del_``` method actually deletes the session
This is not the case with ice version < 3.6
cf. https://trac.openmicroscopy.org/ome/ticket/12921

I have fixed a bug in the disconnect method. We will need to watch if some tests failed (wrong assumption in the tests)

### DIFF
--- a/components/tools/OmeroJava/test/integration/AbstractServerTest.java
+++ b/components/tools/OmeroJava/test/integration/AbstractServerTest.java
@@ -698,8 +698,8 @@ public class AbstractServerTest extends AbstractTest {
      */
     protected omero.client disconnect() throws Exception {
         omero.client oldClient = client;
-        client = null;
         clean();
+        client = null;
         return oldClient;
     }
 
--- a/components/tools/OmeroJava/test/integration/chgrp/HierarchyMoveDatasetTest.java
+++ b/components/tools/OmeroJava/test/integration/chgrp/HierarchyMoveDatasetTest.java
@@ -105,9 +105,6 @@ public class HierarchyMoveDatasetTest extends AbstractServerTest {
         Dataset d = (Dataset) iUpdate.saveAndReturnObject(mmFactory
                 .simpleDatasetData().asIObject());
 
-        // log out
-        //disconnect();
-
         // Create a new group, the user is now a member of the new group.
         ExperimenterGroup g = newGroupAddUser(target, ctx.userId);
 
--- a/components/tools/OmeroJava/test/integration/chgrp/HierarchyMoveDatasetTest.java
+++ b/components/tools/OmeroJava/test/integration/chgrp/HierarchyMoveDatasetTest.java
@@ -106,7 +106,7 @@ public class HierarchyMoveDatasetTest extends AbstractServerTest {
                 .simpleDatasetData().asIObject());
 
         // log out
-        disconnect();
+        //disconnect();
 
         // Create a new group, the user is now a member of the new group.
         ExperimenterGroup g = newGroupAddUser(target, ctx.userId);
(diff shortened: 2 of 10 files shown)


components/tools/OmeroJava/test/integration/AbstractServerTest.java
```diff
@@ -698,8 +698,8 @@ protected void makeGroupOwner() throws Exception {
      */
     protected omero.client disconnect() throws Exception {
         omero.client oldClient = client;
-        client = null;
         clean();
+        client = null;
```
joshmoore: That's a nasty test bug. So is the reason why this wasn't needing in Ice 3.6 that Glacier is doing a better job of cleaning up objects than before?

jburel: The ``disconnect`` method was never invoking the ``del`` method nothing related to ice 3.6

jburel: as expected https://ci.openmicroscopy.org/view/OMERO-DEV/job/OMERO-DEV-merge-integration-java/255/

jburel:
The tests should now turned green.

jburel:
Adding breaking label so the PR can be tested tomorrow with both ice 3.5 and ice 3.6 on breaking. now that https://ci.openmicroscopy.org/view/Breaking/job/OMERO-DEV-breaking-build is green cf. https://ci.openmicroscopy.org/view/Breaking/job/OMERO-DEV-breaking-build/637/

components/tools/OmeroJava/test/integration/chgrp/HierarchyMoveDatasetTest.java
```diff
@@ -106,7 +106,7 @@ private void moveDataDatasetToProject(String source, String target,
                 .simpleDatasetData().asIObject());
 
         // log out
```
mtbc: seems odd to leave the comment behind

mtbc:
Tests all pass on breaking! Can't see from CI that both Ice 3.5 and Ice 3.6 are covered -- do we switch something and then look again tomorrow?

https://github.com/snoopycrimecop/openmicroscopy/compare/develop/breaking/trigger...snoopycrimecop:develop/merge/trigger shows a merge conflict but perhaps I'm looking at the wrong branches.

mtbc:
(Ah, @sbesson tells me the develop one is from the devspace.)

mtbc:
https://github.com/snoopycrimecop/openmicroscopy/compare/develop/breaking/trigger...snoopycrimecop:develop/merge/daily has the same issue.

jburel:
@mtbc:

* integration tests with ice 3.6 http://10.0.51.107:8080/job/OMERO-test-integration/label=testice36/
* integration tests with ice 3.5 http://10.0.51.107:8080/job/OMERO-test-integration/label=testice35/

mtbc:
OMERO-push includes,
```
# PR 4594 jburel 'Session test'
```

Failure of `OmeroPy.test.integration.clitest.test_fs.TestFS.testSetsWithTransfer` involves `Invalid cross-device link` which looks more like a CI setup issue. Seems to me to be good to merge.

jburel:
@mtbc: that's the one I was mentioning this am, not related to ice but more set-up

aleksandra-tarkowska:
@mtbc I have exactly the same issue across all running devspaces, not sure what permissions has to be change and where exactly

mtbc:
Might be interesting to check which directories are being assumed to be on the same partition.

aleksandra-tarkowska:
It turned out that is not a problem of directories or permissions, just lack of right ports build opened. CLI and Insight requires 4063 and 4064. devspace should be green (blue) tomorrow

PR #4595, opened 2016-04-20
Bugfix: Moving multiple folders via context menu didn't work

dominikl:
**Test**: Select multiple folders (also nested ones), move them to another folder (*via context menu*). All of them should be moved (before: only the first one was). For nested folders: Their hierarchy should be preserved. E. g. "B" is a sub folder of "A". Select "A" and "B", move them to folder "C", then "B" should still be a sub folder of "A" (this is just to make sure that I did not reintroduce a bug I fixed earlier).

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/ROITable.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/ROITable.java
@@ -1627,17 +1627,62 @@ public class ROITable
         }
         
         if (SelectionDialog.OBJECT_SELECTION_PROPERTY.equals(name)) {
-            FolderData folder = getSelectedFolders().get(0);
+            // exclude child nodes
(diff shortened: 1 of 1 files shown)


pwalczysko:
All works as expected. Ready to merge.

joshmoore:
Just a general comment: the length of this change is probably approaching something that it would be good to refactor out with a proper javadoc block (and perhaps even a unit test) to keep future maintenance burdens down.

dominikl:
That's true... I squashed far too much stuff into the `ROITable` and it contains lots of copy/pasted code. I'll work on a refactoring/cleaning up PR for the ROI folders issue.

sbesson:
@dominikl: was this refactoring captured anywhere?

dominikl:
Ah sorry, that totally slipped my attention. I can open a refactoring PR after regions is merged into develop again.

PR #4596, opened 2016-04-20
Status field

jburel:
Remove status field from UI See https://trac.openmicroscopy.org/ome/ticket/13066

To Test:
 * Log as user-1 (already imported plane)
 * Browse a plate
 * Check that the "Status" field is no longer displayed.


pwalczysko:
Compared eel latest with eel merge. The field ``Status`` is not displayed on eel merge on comparable plates as opposed to eel latest, where it is displayed.

PR #4600, opened 2016-04-22
Fix potential class cast/NPE for unknown metadata key

dominikl:
Same as #4568 rebased to develop.


mtbc:
The code change looks sensible. I didn't manage to find the `N/A` string in the UI but I couldn't get anything bad to happen either. Good to merge.

PR #4601, opened 2016-04-22
Deprecate windows

aleksandra-tarkowska:
This PR replace https://github.com/openmicroscopy/openmicroscopy/pull/4565, https://github.com/openmicroscopy/openmicroscopy/pull/4589 and https://github.com/openmicroscopy/openmicroscopy/pull/4592

Adds windows decorator to prefs.py to keep message in one location

To test:
 - admin plugin
 - web plugin
 - db plugin
 - fs plugin


components/tools/OmeroPy/src/omero/plugins/prefs.py
```diff
@@ -47,6 +47,11 @@
 
 """
 
+WINDOWS_WARNING = ("WARNING: Support for Windows will be removed in"
```
rleigh-dundee: Maybe qualify with "OMERO.server support" rather than just "Support" since client support will remain.

components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -34,6 +34,7 @@
 
 from omero.plugins.prefs import \
     WriteableConfigControl, with_config, with_rw_config
+from omero.plugins.prefs import windows_warning, WINDOWS_WARNING
```
joshmoore: I'd suggest we not cross-link all the plugins. This can lead to odd loading issues. If need me, let's create a single `OmeroPy/src/omero/install/windows_warning.py` file which can be imported.

components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -1063,6 +1069,7 @@ def copy_template(input_file, output_dir):
 
         return rv
 
+    @windows_warning
```
joshmoore: Conceivably, `with_config` could be updated to include `windows_warning` if that would be cleaner. But this makes sense too.

components/tools/OmeroPy/src/omero/plugins/ldap.py
```diff
@@ -102,6 +109,7 @@ def active(self, args):
         else:
             self.ctx.die(1, "No")
 
+    @windows_warning
```
joshmoore: I'm not completely sure these LDAP ones are necessary. Here `admin_only` differs slightly from the `admin.py` plugin. The former means "member of system group in OMERO" whereas the second means "system administrator with access to the installation directories".

components/tools/OmeroPy/src/omero/plugins/admin.py
aleksandra-tarkowska: `with_config` is used only in admin.py

components/tools/OmeroPy/src/omero/plugins/ldap.py
aleksandra-tarkowska: is ldap a client or a server helper?

joshmoore: I'm not exactly sure of the distinction you mean, but in most cases, ldap.py is a consumer of the API as opposed to a local operation.

joshmoore:
:+1: for reverting ldap.py. And `windows_warning.py` looks nice and clean.

LGTM.

jburel:
Thanks all merging

PR #4603, opened 2016-04-22
CLI command to remove a pyramid file

ximenesuk:
This `fs` subcommand allows a specific pyramid file to be removed given its Image ID, this is regardless of whether it is empty or not. See: https://trello.com/c/TM1Si3UN/38-pyramids This doesn't provide an end to end workflow but it does allow an admin to develop a CLI workflow for dealing with problem pyramids. The command is admin-only currently but it could be relaxed if necessary. There is no test but I will add a basic test.

Testing
-------

On a fresh local system without significant imports one possible test is to import and image and record its ID, then create an artificial pyramid file:

```
touch OMERO/Pixels/X_pyramid
```
where X is the Image ID. Then:

```
$ bin/omero fs fixpyramid Image:X
```
should remove the pyramid file with a confirmation message.

Alternatively import an image that will produce a pyramid file, note its ID, check that the pyramid file has generated and then run the above command. The pyramid file should be deleted. Using a client it should then be possible to force the regeneration of the pyramid file by trying to view the image.

### DIFF
--- a/components/tools/OmeroPy/src/omero/__init__.py
+++ b/components/tools/OmeroPy/src/omero/__init__.py
@@ -6,7 +6,7 @@
    Classes:
       omero.client    -- Main OmeroPy connector object
 
-   Copyright 2007, 2008 Glencoe Software, Inc. All rights reserved.
+   Copyright 2007-2016 Glencoe Software, Inc. All rights reserved.
    Use is subject to license terms supplied in LICENSE.txt
 
 """
@@ -104,6 +104,14 @@ def proxy_to_instance(proxy_string, default=None):
     """
     import omero
     parts = proxy_string.split(":")
+    # If a default is provided and the format is Class:ID
+    # then the default and Class must be the same
+    if len(parts) == 2 and default is not None:
+        if parts[0] != default:
+            raise ClientError(
+                ("Invalid proxy string: %s "
+                 "does not match default %s") % (proxy_string, default))
+
     if len(parts) == 1 and default is not None:
         proxy_string = "%s:%s" % (default, proxy_string)
         parts.insert(0, default)
--- a/components/tools/OmeroPy/src/omero/__init__.py
+++ b/components/tools/OmeroPy/src/omero/__init__.py
@@ -6,7 +6,7 @@
    Classes:
       omero.client    -- Main OmeroPy connector object
 
-   Copyright 2007-2016 Glencoe Software, Inc. All rights reserved.
+   Copyright 2007, 2008 Glencoe Software, Inc. All rights reserved.
    Use is subject to license terms supplied in LICENSE.txt
 
 """
@@ -104,14 +104,6 @@ def proxy_to_instance(proxy_string, default=None):
     """
     import omero
     parts = proxy_string.split(":")
-    # If a default is provided and the format is Class:ID
-    # then the default and Class must be the same
-    if len(parts) == 2 and default is not None:
-        if parts[0] != default:
-            raise ClientError(
-                ("Invalid proxy string: %s "
-                 "does not match default %s") % (proxy_string, default))
-
     if len(parts) == 1 and default is not None:
         proxy_string = "%s:%s" % (default, proxy_string)
         parts.insert(0, default)
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -714,7 +714,7 @@ Examples:
         client = self.ctx.conn(args)
         mib = ManageImageBinaries()
         if not isinstance(args.image, omero.model.ImageI):
-            self.ctx.die(100, "Argument must reference and an Image")
+            self.ctx.die(100, "Argument must reference an Image")
         mib.imageId = args.image.id.val
         mib.deletePyramid = True
         try:
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -271,7 +271,7 @@ class FsControl(CmdControl):
             "image",
             type=ProxyStringType("Image"),
             help=("Image which should have its pyramid removed: "
-                  "ID or Fileset:ID"))
+                  "ID or Image:ID"))
         fixpyramid.add_argument(
             "--wait", type=long,
             help="Number of seconds to wait for the processing to complete "
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -718,7 +718,7 @@ Examples:
             rsp, status, cb = self.response(client, mib, wait=args.wait)
             err = self.get_error(rsp)
             if err:
-                self.ctx.err("Error: " + rsp.parameters['message'])
+                self.ctx.err("Error: " + str(rsp))
             else:
                 if rsp.pyramidPresent:
                     self.ctx.err(
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -266,6 +266,17 @@ class FsControl(CmdControl):
             "fileset",
             type=ProxyStringType("Fileset"))
 
+        fixpyramid = parser.add(sub, self.fixpyramid)
+        fixpyramid.add_argument(
+            "image",
+            type=ProxyStringType("Image"),
+            help=("Image which should have its pyramid removed: "
+                  "ID or Fileset:ID"))
+        fixpyramid.add_argument(
+            "--wait", type=long,
+            help="Number of seconds to wait for the processing to complete "
+            "(Indefinite < 0; No wait=0).", default=-1)
+
         usage = parser.add(sub, self.usage)
         usage.set_args_unsorted()
         usage.add_login_arguments()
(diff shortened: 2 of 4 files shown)


mtbc:
Some observations:

1. The pyramid destroyer is called "fixpyramid"!
1. I get a "removed" message regardless of if there was even a pyramid in the first place.

I am trying to make sense of,
```
Positional Arguments:
  image               Image which should have its pyramid removed: ID or Fileset:ID
```
If I actually put `Fileset:<whatever>` as an argument then it deletes the pyramid of the image with that ID. But if I don't put a prefix then how does it know if I meant image ID or fileset ID?

Having said all that, the pyramids *do* get deleted.

ximenesuk:
Answering @mtbc's observations.

1. The name is intended to match an `admin` subcommand which removes empty pyramids. While it only removes at present it could have an option added to trigger a regeneration, `--regenerate`, say; it might then live up to its name.
1.  The success message has now been suppressed in line with some other commands that succeed silently. This may or may not be reasonable.

The method underlying the CLI argument  `ProxyStringType` was not checking that the class name in `Class:ID` matched the default, if bother were provided. A commit fixes this and a test has been added. However it is **strict**, thus:
```
$omero fs fixpyramid ImageI:1
```
will fail as the default is `Image`. This change affects the `fs` and `metadata` commands. Relevant tests pass but further testing should, at least, check the use of `fs rename Fileset:XX`.

components/tools/OmeroPy/src/omero/__init__.py
```diff
@@ -104,6 +104,14 @@ def proxy_to_instance(proxy_string, default=None):
     """
     import omero
     parts = proxy_string.split(":")
+    # If a default is provided and the format is Class:ID
```
joshmoore: I'm a bit confused where this is coming from. Is something causing incorrect strings to be passed when a default is used? My understanding of the use of the default (e.g. "Fileset") is that:

```
 1         ->  Fileset:1
 Fileset:1 ->  Fileset:1
 Image:1   ->  Image:1
```

ximenesuk: The problem is when the default is 'Image' and the argument is 'Fileset:1' then:
```
Fileset:1 ->  Image:1
```
ie the default is used but it seems to me that this should be an error.

mtbc:
A test for `Fileset:<n>` would be nice too.

ximenesuk:
@mtbc that is effectively tested here: https://github.com/openmicroscopy/openmicroscopy/pull/4603/commits/10d0a3ee6754206def74c8acc7d3a4bf7f55bc92 though the other way around I think.

mtbc:
Ah, clarification: the help is incorrect. :smiley:

mtbc:
Indent issue with `bin/omero fs -h`:
```
  <subcommand>
    images                          List images, filtering for archives, etc.
    rename                          Moves an existing fileset to a new location (admin-only)
    repos                           List all repositories.
    sets                            List filesets by various criteria
    ls                              List all the original files contained in a fileset
    fixpyramid                       Remove the pyramid file associated with an image.
    usage                           Shows the disk usage for various objects.
```

mtbc:
Seems to be working fine for both `fixpyramid` and `rename`. Main thing now is to see that tests pass in CI?

components/tools/OmeroPy/src/omero/__init__.py
joshmoore: Hmmm.... definitely sounds like an error, but I guess I'm not sure who's at fault. I'd think if `Image` is the default and `Fileset` is passed, `Fileset` should be used without error.

mtbc: Safest to throw the error instead of acting on the wrong object? Especially for admin stuff.

joshmoore: Guess the question is *is* it the wrong object? If the user requests "Image" then sure, if the method doesn't work on Image, through. But that may be a cli-level decision and not a proxy_to_string level-decision.

ximenesuk: @joshmoore I can see that being one possible outcome. Given its low coverage and its current use in the CLI it seems better to use this method to check the valid type. Otherwise the CLI command has to first check the whether the type returned by the proxy is valid. Maybe there is an argument for adding an additional `enforce_default` flag to the method?

ximenesuk: I'm happy to revert that commit and deal with catching any error in the `fixpyramid` subcommand. I don't know the original intention of the `proxy_to_instance` method and was only second guessing given its usage.

components/tools/OmeroPy/src/omero/plugins/fs.py
```diff
@@ -712,6 +713,8 @@ def fixpyramid(self, args):
 
         client = self.ctx.conn(args)
         mib = ManageImageBinaries()
+        if not isinstance(args.image, omero.model.ImageI):
+            self.ctx.die(100, "Argument must reference and an Image")
```
mtbc: er, what?

mtbc:
Interesting discovery: if I delete an image's pyramid *and* rename it then I can't regenerate a pyramid for it.

jburel:
To be relisted tomorrow @ximenesuk?

components/tools/OmeroPy/src/omero/__init__.py
joshmoore: > Maybe there is an argument for adding an additional enforce_default flag to the method?

That sounds more non-breaking. Alternative naming suggestion might be to use:

```
 def proxy_to_instance(proxy_string, default=None, enforce=None):
```
And either:
 * only one of default and enforce should be used *or*
 * the enforce check is just applied on top of everything else.

components/tools/OmeroPy/test/integration/clitest/test_fs.py
```diff
@@ -110,3 +113,33 @@ def testRenameAdminOnly(self, capsys):
...
+        pixels_dir = path(cfg.getConfigValue("omero.data.dir")) / "Pixels"
+        iid = self.importSingleImage().id.val
+        subdir = path(long_to_path(iid, root=pixels_dir)).parent
+        subdir.makedirs_p()
+        pyramid_file = subdir / ("%s_pyramid" % iid)
+        pyramid_file.touch()
+        assert pyramid_file.exists()
+        self.args += ["fixpyramid", "Image:%s" % iid]
+        self.cli.invoke(self.args, strict=True)
+        assert not pyramid_file.exists()
```
mtbc: failed today

ximenesuk: Yes! I was looking at it. The final assert shows that the file is still there:

```
>       assert not pyramid_file.exists()
E       assert not True
E        +  where True = <bound method path.exists of path(u'/repositories/OMERO-DEV-merge-integration/Pixels/Dir-002/2423_pyramid')>()
E        +    where <bound method path.exists of path(u'/repositories/OMERO-DEV-merge-integration/Pixels/Dir-002/2423_pyramid')> = path(u'/repositories/OMERO-DEV-merge-integration/Pixels/Dir-002/2423_pyramid').exists
```

as indeed it is:

```
colin$ omero -C fs fixpyramid Image:2423
Server: [eel.openmicroscopy.org:4064]eel.openmicroscopy.org:24064
Username: [root]
Password:
Created session 0e4fc3ba-a3a2-4329-b2ed-848ead0452fe (root@eel.openmicroscopy.org:24064). Idle timeout: 10 min. Current group: system
$
```
doesn't fail, but,:

```
[cblackburn@ome-c6100-1 ~]$ ls -l /repositories/OMERO-DEV-merge-integration/Pixels/Dir-002/
total 12
-rw-rw-r-- 1 hudson hudson    8 Apr 28 07:49 2057
-rw-rw-r-- 1 hudson hudson    8 Apr 28 07:57 2096
-rw-rw-r-- 1 hudson hudson    8 Apr 28 07:58 2101
-rw-rw-r-- 1 hudson hudson    0 Apr 28 05:15 2423_pyramid
```

ximenesuk:
I noticed that this test failed for the last two runs: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest.test_ticket10618/Test10618/test_rw_root_dirTrue_grpFalse_16/

This fails because of the pyramid I generated in my test (or at least because it wasn't deleted) I assume because it is zero size. It doesn't fail on https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/

ximenesuk:
Although the command works locally it seems to fail silently in the CI builds resulting in the failing test. Without further logging in the MIB command I'm struggling to work out what is happening. I'd suggest not including this in 5.2.3 though I'll add some logging to try to figure out what is happening. /cc @jburel

jburel:
If the test failed we won't include it in the release. @will-moore still has an outstanding PR, it will be reviewed on monday so more likely to do the release Tuesday (Assuming we are happy with the other PR)

ximenesuk:
It's still failing and I am unlikely to fix this in the next cycle. I will close the PR for now and retest once 5.2.3 is out of the way.

aleksandra-tarkowska:
Im order to reproduce the failure, would you like to try that in docker? It is failing on devspace as well. This may help with debugging

PR #4606, opened 2016-04-23
cached thumbnail size is 96

aleksandra-tarkowska:
This PR change default thumbnail size loaded to plate grid to 96. This is because cached thumbnails in `omero.data.dir` are 96 not 64. There is no point to render them again

Should we also change default in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/views.py#L308 ? If so in 5.3 if that is ok?

@chris-allan @will-moore do you see any implications?

To test:
 - open plate and inspect thumbnail URL

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -1372,7 +1372,7 @@ def plateGrid_json(request, pid, field=0, conn=None, **kwargs):
     except ValueError:
         field = 0
     prefix = kwargs.get('thumbprefix', 'webgateway.views.render_thumbnail')
-    thumbsize = int(request.GET.get('size', 64))
+    thumbsize = int(request.GET.get('size', 96))
     logger.debug(thumbsize)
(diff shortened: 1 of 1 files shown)


mtbc:
URLs are like `https://cowfish.openmicroscopy.org/merge/webgateway/render_thumbnail/21064/96/` and provide 96×96 images. Good to merge.

joshmoore:
Thanks, @mtbc . Don't know if a discussion took place elsewhere, but it'd be good to hear @chris-allan's and @will-moore's opinions.

will-moore:
This seems fine for me.
Of course, since you can't zoom the thumbnails, you don't benefit from the larger size (more bandwidth for no visual change) but the fact that they are already cached should more than make up for that.
Working fine too. Good to merge from me.

jburel:
This should speed up the retrieval.

chris-allan:
As the 64 pixel longest side thumbnails would be cached after the first retrieval anyway it's certainly an improvement but not an overly significant one. Happy for this to be merged.

PR #4610, opened 2016-04-26
Fix tooltips metadata panel

dominikl:
Fixes some tooltips of the right hand side metadata panel, see [Trello - Insight - Right Hand Pane tooltip changes + one other issue](https://trello.com/c/POjkfKpE/392-insight-right-hand-pane-tooltip-changes-one-other-issue).

**Test**: Check that the tooltips match the proposed ones on the trello card.

Note: When testing the last point (select files for scripts), I noticed that it doesn't work with @emilroz 's "Populate metadata" script which is deployed on eel. Reason for this is, that the scripts define the FileAnnotation parameter as list of Strings, but it should be a list of Longs instead!

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
@@ -199,7 +199,7 @@ public class ImportDialog extends ClosableTabbedPaneComponent
 	private static final String TEXT_TAGS_SELECTION = "Tags Selection";
 
 	/** Text for add tags button */
-	private static final String TOOLTIP_ADD_TAGS = "Add Tags.";
+	private static final String TOOLTIP_ADD_TAGS = "Add tags.";
 
 	/** File naming checkbox text */
 	private static final String TEXT_OVERRIDE_FILE_NAMING =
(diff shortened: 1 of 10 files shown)


jburel:
``Delete rating.`` ``Add tags.`` both have ``.``. The rest does not

jburel:
Also suggested on card ``Attachments - add attachment button - Insight: “Attach a document” - should be “Attach a file"```

jburel:
We have in several classes. Tooltip with a dot at the end. Probably not worth the effort to review all that cc @gusferguson

gusferguson:
@dominikl 

Still some issues with tooltips.

“Filter tags and attachments.” should be “Filter annotations by user"
“Edit the description.” should be “Edit the description”
“Edit channel.” should be “Edit channel names"

dominikl:
Yes , there are hundreds of tooltips with a dot at the end. That'd be quite a task to find and change them all. The last commit addresses the three issues of @gusferguson 's comment. Where shall we draw the line for this PR?

gusferguson:
@dominikl - happy to leave it at that - if the RHP is all consistent now then we can address the rest as and when we are doing changes in those areas.

jburel:
As indicated in a previous comment,  I will not spend time removing a "." in other classes.

PR #4613, opened 2016-04-26
Add support for twitter cards (full image viewer only) (rebased onto metadata52)

manics:
This is the same as gh-4260 but rebased onto metadata52.

----

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -1952,6 +1952,20 @@ def full_viewer(request, iid, conn=None, **kwargs):
         if image is None:
             logger.debug("(a)Image %s not found..." % (str(iid)))
             raise Http404
+        twitter = None
+        twitter_img = None
+
+        if settings.TWITTER_ENABLED and hasattr(settings, 'TWITTER_SITE_USER'):
+            twitter = settings.TWITTER_SITE_USER
+        if twitter:
+            prefix = kwargs.get(
+                'thumbprefix', 'webgateway.views.render_thumbnail')
+
+            def urlprefix(iid):
+                return reverse(prefix, args=(iid,))
+
+            twitter_img = request.build_absolute_uri(urlprefix(iid))
+
         d = {'blitzcon': conn,
              'image': image,
              'opts': rid,
@@ -1962,6 +1976,9 @@ def full_viewer(request, iid, conn=None, **kwargs):
              'viewport_server': kwargs.get(
                  # remove any trailing slash
                  'viewport_server', reverse('webgateway')).rstrip('/'),
+
+             'twitter': twitter,
+             'twitter_img': twitter_img,
 
              'object': 'image:%i' % int(iid)}
 
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -1964,7 +1964,7 @@ def full_viewer(request, iid, conn=None, **kwargs):
 
         if hasattr(settings, 'SHARING_OPENGRAPH'):
             opengraph = settings.SHARING_OPENGRAPH.get(server_name)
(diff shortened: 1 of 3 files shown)


manics:
--rebased-from #4260

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -1941,6 +1942,9 @@ def full_viewer(request, iid, conn=None, **kwargs):
     @return:            html page of image and metadata
     """
 
+    server_id = request.session['connector'].server_id
+    server_name = Server.get(server_id).server
```
manics: Something (I haven't been able to find what) changed in omeroweb which means `Server._registry` is empty, so `Server.get(server_id)` returns None instead of an object.

manics:
Exclude until someone can figure out what's changed between 5.1 and 5.2

components/tools/OmeroWeb/omeroweb/webgateway/views.py
aleksandra-tarkowska:
```
2016-05-12 21:07:28,464 DEBUG [                     omero.gateway.utils] (proc.56823) setOmeroShare():163 Key 'omero.share' not found in <ServiceOptsDict: {'omero.session.uuid': '7066138f-cd75-408f-b913-d62b833fb2ff', 'omero.group': '-1', 'omero.client.uuid': 'ad442d36-7712-46cf-9e23-dcc59062c28d'}>
2016-05-12 21:07:28,465 DEBUG [               omeroweb.webgateway.views] (proc.56823) full_viewer():1946 server_id 1
2016-05-12 21:07:28,465 DEBUG [               omeroweb.webgateway.views] (proc.56823) full_viewer():1948 server_name: omero
```
your PR works for me

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -45,6 +45,7 @@
 from omero.util.decorators import timeit, TimeIt
 from omeroweb.http import HttpJavascriptResponse, HttpJsonResponse, \
     HttpJavascriptResponseServerError
+from connector import Server
```
aleksandra-tarkowska: could you try `from omeroweb.connector import Server` ?

components/tools/OmeroWeb/omeroweb/webgateway/views.py
aleksandra-tarkowska: @manics FastCGI used manage.py to start workers, wsgi uses different application handler, where settings are not imported. if you remove https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/manage.py#L38 it will fail with exactly the same error.

joshmoore: Thanks for the sleuthing, @aleksandra-tarkowska

manics:
Thanks @aleksandra-tarkowska , I've added your fix and it works for me now.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -1952,6 +1956,31 @@ def full_viewer(request, iid, conn=None, **kwargs):
...
             raise Http404
+
+        opengraph = None
+        twitter = None
+        image_preview = None
+        page_url = None
+
+        if hasattr(settings, 'SHARING_OPENGRAPH'):
+            opengraph = settings.SHARING_OPENGRAPH.get(server_name)
+            logger.debug('Open Graph enabled: %s', twitter)
```
joshmoore: `s/twitter/opengraph/`

manics: Fixed

joshmoore:
Tested with Kelli. Merging for IDR-0.0.5

manics:
Note #4792 is only a partial rebase of #4744. If this PR (#4613) us rebased to `develop` the remaining changes from #4744 will be required.

atarkowska:
--rebased-to #5216

PR #4614, opened 2016-04-27
bulk import improvements

joshmoore:
This PR migrates functionality from the https://github.com/IDR/idr-metadata/blob/0.0.4/screen_import.py file as well as the "mini-[idr-setup-2.sh](https://github.com/joshmoore/idr-metadata/blob/f9f768fbbfdc886343f34e545839be512ed356f7/scripts/mini/idr-setup-2.sh)" script into bin/omero import.

New functionality includes:

 * `---bulk` (Python): allow driving import via YAML files
   * `path` and `columns`: allow for importing from tsv, csv, or txt files.
   * `include`: allows for importing a relative path for shared variables
   * `continue`: passes the `-c` flag to continue on errors
   * `dry_run`: print commands rather than execute them
   * other parameters match existing parameters, e.g. `name`, `target`, etc.
  
 * `--output=yaml` (Java) for more modern stdout 

Other changes include:
 * **breaking**: make `-f` fail when no candidates are found
 * `--logprefix` (Python): allows prefixing the values of `---file` and `---err`
 * `--logback` (Python): for passing a logging file
 * begin exposing advanced arguments in Python (e.g. transfer, exclude)
 * deprecate underscore arguments (Java): e.g. `auto_close`
 * deprecate `---` arguments (Python): e.g. `---file`

cc: @ximenesuk @aleksandra-tarkowska 

Note: this requires installing PyYAML which should be discussed project-wide.


joshmoore:
Discussing with @sbesson, the yaml output from `bin/omero import -f --output=yaml` may be a starting point for skipping import-candidates client-side.

joshmoore:
Note updated description with the new argument handling.

joshmoore:
`--debug=DEBUG` handling may need further (manual) testing.

joshmoore:
Merging for IDR-0.0.5

mtbc:
--rebased-to #4731 on `develop`

components/tools/OmeroPy/src/omero/plugins/import.py
```diff
@@ -320,6 +499,140 @@ def importer(self, args):
...
+            while bulkfile:
+                bulkfile = os.path.abspath(bulkfile)
+                parent = os.path.dirname(bulkfile)
+                with open(bulkfile, "r") as f:
+                    data = safe_load(f)
+                    contents.append((bulkfile, parent, data))
+                    bulkfile = data.get("include")
+                    os.chdir(parent)
+                    # TODO: include file are updated based on the including file
+                    # but other file paths aren't!
```
joshmoore: No. All the cleanup for these metadata(52) PRs typically happens when opened against the mainline.

components/tools/OmeroPy/test/unit/clitest/test_import.py
```diff
@@ -235,9 +241,127 @@ def new_client(x):
...
+        assert "fake" in result["group"]
+        assert 1 == len(result["files"])
+        assert "reader" in result
+        assert "spw" in result
+
+    def testBulkNoPaths(self):
+        t = path(__file__) / "bulk_import" / "test_simple"
+        b = t / "bulk.yml"
+        self.add_client_dir()
+        self.args += ["-f", "---bulk=%s" % b, "dne.fake"]
```
ximenesuk: Given the help output and the deprecation of `---` shouldn't this be `--bulk` unless backwards compatibility is specifically being tested?

The same goes for other instances in this file.

joshmoore: Don't think this was specifically for compatibility testing. More likely just missed.

components/tools/OmeroPy/test/unit/clitest/bulk_import/test_cols/bulk.tsv
```diff
@@ -0,0 +1,2 @@
+meta_one	1.fake
+meta_two	2.fake
```
ximenesuk: This `tsv` file references two fake files but only one exists in the directory. This makes no difference to the unit test but is inconsistent as an example.

PR #4617, opened 2016-04-28
Add keepalive for root connection

ximenesuk:
This an attempt to fix the three fails record on the card https://trello.com/c/EavoXnp4/649-importas-and-multigroup-cli-tests

This adds a `keepAlive` for the root connection so that the connection is still available when these three tests need it. This is the most minimal solution and appears to work when testing locally against a remote server - it needs this latency for the tests to take long enough for the connection to be lost, see below. 

Testing
-------

The tests on Jenkins should pass, though given the intermittency the tests will need to be checked for a number of runs.

Local testing is possible though dependent on a slower connection. Set ice.config to point at `eel.openmicroscopy.org:24064` and then:

```
./setup.py test -t test/integration/clitest/test_import.py -m "not broken" -k "not Symlink" -v
```

should pass cleanly.

### DIFF
--- a/components/tests/python/library/__init__.py
+++ b/components/tests/python/library/__init__.py
@@ -116,6 +116,18 @@ class ITest(object):
         cls.root = None
         cls.__clients.__del__()
 
+    def keepRootAlive(self):
+        """
+        Keeps root connection alive.
+        """
+        try:
+            if self.root.sf is None:
+                assert self.root.connect()
+            else:
+                self.root.sf.keepAlive(None)
+        except Exception:
+            raise
+
     @classmethod
     def omeropydir(self):
         count = 10
--- a/components/tests/python/library/__init__.py
+++ b/components/tests/python/library/__init__.py
@@ -122,7 +122,9 @@ class ITest(object):
         """
         try:
             if self.root.sf is None:
-                assert self.root.connect()
+                p = Ice.createProperties(sys.argv)
+                rootpass = p.getProperty("omero.rootpass")
+                self.root.createSession("root", rootpass)
             else:
                 self.root.sf.keepAlive(None)
         except Exception:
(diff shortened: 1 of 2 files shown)


components/tests/python/library/__init__.py
```diff
@@ -116,6 +116,18 @@ def teardown_class(cls):
         cls.root = None
         cls.__clients.__del__()
 
+    def keepRootAlive(self):
+        """
+        Keeps root connection alive.
+        """
+        try:
+            if self.root.sf is None:
+                assert self.root.connect()
```
joshmoore: Is `root` here a `BlitzGateway` or a `BaseClient`? In the latter case, I don't think there's a "connect" method. NB: an alternative option would be [enableKeepAlive](https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.2/components/tools/OmeroPy/src/omero/clients.py#L612) on the `BaseClient` but that will require cleaning them up, otherwise tests may hang.

ximenesuk: Yes, sorry, this is bad pasting from me and because nothing failed not looking further.

ximenesuk: Of course this also means this condition was never true and so is just a simple keepalive (the else clause) valid here of is it better to try to create the session again?

`enableKeepAlive` was mentioned in conversation with @jburel but I shied away from it because some of our tests don't clean up that well. Maybe this is something for a card when we do some test refactoring - as I am sure we will do.

joshmoore: > Maybe this is something for a card when we do some test refactoring - as I am sure we will do.

:+1: -- or if this doesn't work :smile:

jburel:
Good to have it opened so we can hopefully have the tests green

jburel:
Tested from home against eel.
```
test/integration/clitest/test_import.py::TestImport::testImportAsRoot PASSED
test/integration/clitest/test_import.py::TestImport::testImportMultiGroup PASSED
test/integration/clitest/test_import.py::TestImport::testImportAsRootMultiGroup PASSED
```

The same tests always fail w/o the change

jburel:
the "importAs" tests are passing this morning https://ci.openmicroscopy.org/view/OMERO-DEV/job/OMERO-DEV-merge-integration-python/262/

joshmoore:
And the masses rejoice!

jburel:
Last few builds passed with the change.
Merging thanks @ximenesuk

PR #4619, opened 2016-04-28
Load all `omero/plugins` subdirs on the PYTHONPATH

joshmoore:
During plugin loading, loop over the contents of sys.path
(i.e. PYTHONPATH) and for all directories which contain a
subdirectory `omero/plugin` run the loadpath method as if
it had been passed to `bin/omero` with `--path`.

NB: This currently does not handle plugins packaged as eggs.

See https://github.com/ome/omero-cli-duplicate/pull/1 for an example of how this would be used.

Further explanation is now on https://trello.com/c/qMrZfcoU/78-external-cli-plugins

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1348,7 +1348,17 @@ class CLI(cmd.Cmd, Context):
         in the parser
         """
 
-        for plugin_path in self._plugin_paths:
+        paths = set(self._plugin_paths)
+        for x in sys.path:
+            x = path(x)
+            if x.isdir():
+                x = x / "omero" / "plugins"
+                if x.exists():
+                    paths.add(x)
+            else:
+                if self.isdebug:
+                    print "Can't load %s" % x
+        for plugin_path in paths:
             self.loadpath(path(plugin_path))
 
         self.configure_plugins()


sbesson:
Closing for now. Can be reopened when we investigate CLI plugin discovery.

joshmoore:
--rebased-to #5163

joshmoore:
Re-opening in discussion with @jburel so we can consider getting `render.py` in.

jburel:
Finally got a bit of time to go back to it
Tested using https://github.com/ome/omero-cli-duplicate/pull/1

Steps followed for testing;
* start server using this PR
* Run ``omero duplicate -h``. As expected an error is returned
 ``bin/omero: error: argument <subcommand>: invalid choice: 'duplicate'`` 
* Run ``pip install -U omero-cli-duplicate``. Version 0.0.1 of the plugin is installed.
* Run ``bin/omero duplicate -h``. The help is displayed. This means that the plugin has been updated.
* Remove the version installed via pip
* Check the original version

This works as expected.

jburel:
I have also tested option 1 to install ``render.py``
see https://trello.com/c/qMrZfcoU/78-external-cli-plugins

```
bin/omero render -h
usage: bin/omero render [-h] [-C] [-s SERVER] [-p PORT] [-g GROUP] [-u USER]
                        <subcommand> ...

Tools for working with rendering settings
```

jburel:
Point discussed today:
option to have include plugins from "others" folder e.g. ``your_org/plugins``

joshmoore:
Discussed today after @jburel's successful testing. My primary (longer-term) concerns are:

 * flexibility: @chris-allan recently mentioned wanting to load plugins from other subdirectories. However, since `bin/omero` is fundamentally a flat namespace, I'm not sure that we should look in more subdirectories than `omero/plugins`
 * usability: At the moment, a user installing OMERO.server will not have any of these decoupled plugins. Possible deployment options for `suggested`/`recommended` plugins:
   - require users to find them online
   - list plugins on the OMERO docs
   - put the plugins in a requirements file (depends on our virtualenv instructions)
   - somehow bundle the current version of the plugins with OMERO.zip

All of that is open for discussion, but I think this simple initial PR is good for merging.

jburel:
That's definitely a good start. Agree that we should merge and improve later on depending on coming workflow
If no more comments, will merge tomorrow post chat

joshmoore:
:+1: Docs & deployment issues remain: https://trello.com/c/DeEfMN0E/140-doc-cli-plugin-deployment

PR #4624, opened 2016-04-29
Roi table improvements

dominikl:
Some more improvements/bug fixes with respect to ROI Folders handling, see [Trello - ROI Folders Insight Bugs/RFEs](https://trello.com/c/XUI38KAp/27-roi-folders-insight-bugs-rfes).

* Enable Drag/Drop only on the first column of the ROI table
* If a Folder name matches the filter expression, don't filter the sub folders, always show them.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/TableRowTransferHandler.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/TableRowTransferHandler.java
@@ -106,6 +106,8 @@ public class TableRowTransferHandler extends TransferHandler {
      * Check if the current selection is draggable
      */
     private boolean checkDragSource() {
+        if (table.getSelectedColumn() > 0)
+            return false;
         int[] selection = table.getSelectedRows();
         for (int i = 0; i < selection.length; i++) {
             ROINode n = (ROINode) ((ROITable) table).getNodeAtRow(selection[i]);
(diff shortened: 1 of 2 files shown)


gusferguson:
@dominikl 
Tested using regions-omero.docker.openmicroscopy.org DB: user-14:
OMERO.insight-5.3.0-m2-344-3a2679e-ice35-b132-mac
and
OMERO.insight-5.3.0-m2-344-3a2679e-ice35-b132-win

Behaves as expected.
Note - Drag and Drop does not work for moving ROIs/folders inside folders —> orphaned - presume this is intended behaviour.
Filtering works as expected.
Good to merge.

gusferguson:
@dominikl @pwalczysko 

Just noticed that filtering is disabled if Show All ROIs is checked.
This is contrary to the workflow described in Petr's scenario https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/ROIFolders.html Point 5.

I think you should be able to filter when Show All ROIs is checked - i.e. as per scenario.

id- 18 20110209_region1_1505_atp zvi

gusferguson:
@dominikl @pwalczysko 

Sorry - I see that I have misunderstood the scenario. The behaviour is the same as the  scenario.

Probably better to leave as it is and I will see if I can suggest a clarification for the scenario.

dominikl:
DnD moving of ROIs and folders inside folders should work, too. They should only become orphaned when dropped into the empty space of the ROI table. I'll check that again.

gusferguson:
@dominikl - you are correct - it does work - I was not dragging far enough. Sorry!

Good to merge.

PR #4625, opened 2016-04-29
bring BinData changes into OMERO

mtbc:
Stubs new BinData methods so that OMERO still works and metadata mapping to-do's are logged.

See https://github.com/openmicroscopy/bioformats/pull/2319.

### DIFF
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -86,6 +86,7 @@ import ome.util.LSID;
 import ome.xml.meta.MetadataRoot;
 import ome.xml.model.AffineTransform;
 import ome.xml.model.MapPair;
+import ome.xml.model.enums.Compression;
 import ome.xml.model.enums.FillRule;
 import ome.xml.model.enums.FontFamily;
 import ome.xml.model.enums.FontStyle;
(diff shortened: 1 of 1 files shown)


mtbc:
another go with travis

sbesson:
@mtbc: since this PR depends on openmicroscopy/bioformats#2319, I would expect Travis failures until the low-level Bio-Formats PR is merged and the 5.2.0-SNAPSHOT JARs are available via to the OME artifactory.

mtbc:
What's the scope of this? Lots of `ignoreUnneeded` or should I be extending the OMERO model?

joshmoore:
I'd definitely suggest starting off by using one of the method defined at the bottom (https://github.com/mtbc/openmicroscopy/blob/bindata/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java#L8999). In many cases, these are fields which are not intended to be imported -- e.g. we don't *want* to allow someone to create a new Experimenter. Having the correct method, especially `ignoreMissing`, makes later cleanup *much* easier.

rleigh-dundee:
The stubbed out methods added here certainly look fine to me.  What's needed to get travis green here, or can this be tested and merged independently of travis if it can't cope with the bioformats dependency?

mtbc:
The `regions` branch is retiring. @sbesson: Let me know when you want me to rebase this to develop and I'll do so.

mtbc:
Superseded by #4641.

PR #4627, opened 2016-05-03
remove rank and sort by name regardless of node type

aleksandra-tarkowska:
this PR adds ability to sort client tree nodes by type. 

`omero.client.ui.tree.type_order`by default is `tagset,tag,project,dataset,screen,plate,acquisition,image`

to mix types in a tree and sort them by default sorting strategy impleented in client set `omero.client.ui.tree.type_order false`

Note: removing type from the list won't turn off appearance of nodes, it only allows to order them

firefoxscreensnapz026

**Testing: if you update server configuration you must logout from web (even public user) to pick up changes, server config is stored in session and loaded only for the first time when connection is created**

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
@@ -1066,8 +1066,6 @@ $(function() {
             var inst = this;
             var node1 = inst.get_node(nodeId1);
             var node2 = inst.get_node(nodeId2);
-            var name1 = node1.text.toLowerCase();
-            var name2 = node2.text.toLowerCase();
 
             function getRanking(node) {
                 // return rank based on 'omero.client.ui.tree.type_order' list
@@ -1081,8 +1079,9 @@ $(function() {
                 return WEBCLIENT.UI.TREE.type_order.length + 1;
             }
 
-            // if sorting list is turned off mix object and sort by name
-            if (WEBCLIENT.UI.TREE.type_order.indexOf('false') > -1) {
+            function sortingStrategy(node1, node2) {
+                // sorting strategy
+
                 if(node1.type === 'experimenter') {
                     if (node1.data.obj.id === WEBCLIENT.USER.id) {
                         return -1;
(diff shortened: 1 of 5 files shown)


aleksandra-tarkowska:
if this work, do you want me to add `omero.client.ui.tree.split_by_type true/false` to server config?
- false (default) woudl keep ui as it is
- true will allow mixing them.

manics:
How about something more general, along the lines of:

`omero.client.web.ui.tree.type_order:`
- 'mixed'
- `["project", "dataset", "screen", ...]` etc to specify the which top-level groups are shown, and in what order (so if `orphaned` isn't in the list you don't show them).

aleksandra-tarkowska:
ok, can do this, but this wont include orphaned. Orphaned container is a fixed node and it it handled separately via `omero.client.ui.tree.orphans.*`

aleksandra-tarkowska:
ok, can try that. At the moment I am not sure about appearance of orphaned. Orphaned container is a fixed node and it is handled separately via `omero.client.ui.tree.orphans.*`

aleksandra-tarkowska:
@manics try now

aleksandra-tarkowska:
@joshmoore conflict resolved

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
```diff
@@ -1070,27 +1070,31 @@ $(function() {
...
+                    return rank;
                 }
+                // types not specified in 'omero.client.ui.tree.type_order'
+                // are sorted as loaded to jquery based on sql
+                return WEBCLIENT.UI.TREE.type_order.length + 1;
+            }
+
+            // if sorting list is turned off mix object and sort by name
+            if (WEBCLIENT.UI.TREE.type_order.indexOf('false') > -1) {
+                if(node1.type === 'experimenter') {
```
will-moore: This code to 'sort by name' is repeated exactly below.
Would be nicer not to have this duplication.

will-moore:
The code changes looks fine to me (one minor comment) but I've not actually tested locally. @manics you want to re-review (based on your comments above)?

manics:
@will-moore yes, I'll try it out.

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
aleksandra-tarkowska: @will-moore try now

manics:
I'm having trouble with this

```
$ OMERO.py-5.2.3-104-0301d51-ice35-b97/bin/omero config set omero.client.ui.tree.type_order false
$ omero web restart
```

Didn't have any effect on the UI.

will-moore:
@manics Do you have a mixture of Projects, Datasets, Screens etc? Are they sorted by name?

manics:
I've got a mixture, and the sorting seems to be the default instead of by name only.

manics:
I'll re-test this, now that I realise it's a server-side property (I was setting it in a local OMERO.web connecting to a remote OMERO.server)

joshmoore:
Looks good on demo2. Thanks, @aleksandra-tarkowska !

aleksandra-tarkowska:
--rebased-to #4682

PR #4628, opened 2016-05-03
add a description template for pull requests

mtbc:
See https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/ -- perhaps this is a useful start for people to help them remember what to include in their pull request descriptions.

### DIFF
--- a/.github/Pull_Request_Template.md
+++ b/.github/Pull_Request_Template.md
@@ -1,0 +1,21 @@
+# What this PR does
+
+Fill me in.
+
+
+# Testing this PR
+
+1. required setup
+
+2. actions to perform
+
+3. expected observations
+
+
+# Related reading
+
+Link to cards, tickets, other PRs:
+
+1. background for understanding this PR
+
+2. what this PR assists, fixes, or otherwise affects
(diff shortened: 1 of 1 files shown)


ximenesuk:
This will mean no more empty PR description so will that bit of scc need to handle unedited PR descriptions? /cc @sbesson

ximenesuk:
This seems like a good idea. Maybe there are some PRs where addressing every point may not make sense? So how strict would any support policy be or is this simply intended as a reminder to provide as much information as is useful? So would there be some supporting docs if this was merged? (And if so can we add something about descriptive PR names!?)

mtbc:
Intended as a reminder, parts to be deleted as seems reasonable. Good idea about PR names too!

ximenesuk:
I'm happy with the idea and the text as it stands. Maybe it is worth adding a small supporting doc at https://www.openmicroscopy.org/site/support/contributing/using-git.html#pull-requests ? But good to merge from me.

sbesson:
Merging and we can review the template in follow-up PRs.

PR #4630, opened 2016-05-04
Open with

will-moore:
"Open With" plugins allow configuration of additional viewers etc for objects in web.
In this PR, we define the open_with setting and implement this in right-click menu of jsTree.

Docs at https://github.com/openmicroscopy/ome-documentation/pull/1543 are the starting point for testing this PR. More (same) details below if needed.

In the simplest case, viewers are specified as a url that takes no IDs, E.g. ```webtest/myViewer/``` and when an image(s) is opened with the viewer, webclient will open the url as ```webtest/myViewer/?image=1```.
This allows the ```reverse(url_name)``` to be called without any image IDs and selected images etc can be added by JavaScript.

The minimum needed to specify a viewer is "label" and "url_name".
```
$ bin/omero config append omero.web.open_with '["My viewer", "url_name"]'
```
We use ```reverse(url_name)``` to resolve viewer.
The 'Open With...' > viewer will only be enabled for single images by default and will be opened in a new window (same as standard viewer).
To allow multiple images or other objects to be opened in the viewer, the open_with config can include an options dictionary with:
 - ```supported_objects```: list of supported types. E.g. ```["dataset", "images"]``` will enable the viewer when a single Dataset or 1 or more Images is selected.
 - ```target```: set this to ```"_blank"``` to open the viewer in a new tab (instead of new window).

E.g. OMERO.figure - enabled with multiple images selected.
```
$ bin/omero config append omero.web.open_with '["OMERO.figure", "new_figure",  {"supported_objects":["images"],"target":"_blank"}]
```

Further customisation can be provided by including a script with JavaScript functions to define whether the plugin is enabled (for a selected set of jsTree nodes) and how to handle the menu "action".

screen shot 2016-05-05 at 00 22 16

To test 'open with' with script. E.g. with [viewer-ng](https://github.com/waxenegger/viewer-ng/tree/master/plugin/viewer-ng) installed:
 - create a file ``` omeroweb/webclient/static/webclient/viewer-ng-openwith.js``` with this to provide custom Url provider:
```
OME.setOpenWithUrlProvider("viewer-ng", function(selected, url) {
    url += selected[0].id + "/";
    return url;
});
```
 - Config:
```
$ bin/omero config append omero.web.open_with '["viewer-ng", "viewer_ng_index", {"script_url": "webclient/viewer-ng-openwith.js"}]'
```

You can see the config that the webclient consumes by visiting:
``` webgateway/open_with/ ```

### CODE REVIEW

**Comment by gusferguson** [comment]
@will-moore 

Tested with https://cowfish.openmicroscopy.org/merge/webclient/ user-3 read-only-1

Behaves as expected.
Good to merge.

**Comment by joshmoore** [comment]
Before merging, I'd be up for a design discussion on this (sorry if I already missed one) similar to the API one. For example, is there any potential for allowing the registered functions to filter themselves out of a list, for example if a 3D viewer doesn't want to show up for 2D images? Do external URLs work as well or must they be installed as an app in OMERO.web? Will the menu be available for all elements in the UI or just for the images in the left-hand panel? Can we do anything with PDF file annotations or OMERO.tables down the road?

**Comment by sbesson** [comment]
Also before merging anything on `develop`, we should have a conversation on the timeline/plan for splitting `dev_5_2/develop` now 5.2.3 is tagged and we do not have an active 5.2.4 release planned.

**Comment by jburel** [comment]
Actually I wanted to have a call about it this am

**Comment by jburel** [comment]
https://trello.com/c/RLvGaXUK/16-list-of-available-viewer-plugins for further discussion

**Comment by will-moore** [comment]
@joshmoore There's not been a design discussion yet since this was initially considered to be a very simple "list of image viewers". In this PR I've gone a bit beyond that to support all objects that are in the left-hand tree (E.g. "dataset" in the example above).
We could go further to allow adding a JavaScript function that is passed the currently selected tree nodes and returns enabled true/false. This is similar to what we do for the centre/right panel plugins.
However, it will only have the info that is in the jsTree (which currently includes dtype, id, name, sizeX,Y,Z).
I can't imagine how you'd open something in OMERO using an external URL (unless some external site knew how to access your server - E.g. you're using a public repo) but this is way beyond what we're planning here.
We could potentially add the "Open with..." options to anything in the UI including annotations if we want.
Happy to discuss this anytime. I guess @chris-allan might be interested too?

**Comment by pwalczysko** [comment]
The workflow for Figure works for me as well and seems logical. I suppose the Dataset on Figure makes no sense, as it does not open anything, but I understand this is just for now and just for testing.

**Comment by jburel** [comment]
Marking it as onhold so it is not listed. Further discussion to happen

**Comment by will-moore** [comment]
Updated description with how to use JavaScript to handle enabled status and "Open With" action.

**Comment by joshmoore** [comment]
> Open with "Genbank Protein" will only be enabled when a single object is selected and the name of the object is a number. The action will be to open http://www.ncbi.nlm.nih.gov/protein//
> See example in ee70460

A possible next step for the IDR metadata, @will-moore, might be to convert the multiple map annotation rows for a gene (name, URL, identifier, etc) into a single-annotation with a given namespace. For something like this, would using the namespace and/or the contents of the map annotation be possible for triggering the open with?

cc @eleanorwilliams

**Comment by manics** [comment]
You could add a namespace attr to `NamedValue`

**Comment by joshmoore** [comment]
> You could add a namespace attr to NamedValue

That'll need a design issue itself :smile:

**Comment by will-moore** [comment]
@joshmoore You'd need to have the "Open with" menu on the annotation itself, not on the parent in the tree, since the tree has no knowledge of annotations.
Probably the biggest issue is adding a drop-down / context menu to annotations in the right panel. We don't have an easy way to add menus - each needs a bit of JavaScript hand coded etc.

**Comment by will-moore** [comment]
@joshmoore With that last commit, I added experimental support for File Annotations. I've used a simple select chooser (we might want to make a nicer dropdown in due course) that uses the same 'Open With' list as in the jsTree. The json data that is passed to the ```isEnabled()``` and ```handleAction()``` javascript functions is the same format for the jsTree menu and the FileAnnotation menu. E.g. 
```
# Image
{
	type: "image",
	id: 3727,
	date: "2014-08-22T00:30:14Z",
	name: "a-range-start CSFV-GFP09_1_R3D_D3D.dv_proj-1",
	ownerId: 3,
	sizeX: 552,
	sizeY: 552,
	sizeZ: 1
}

# File Annotations
{
	type: "fileannotation",
	id: 17557,
	file: {
		path: "",
		size: 8958,
		id: 22757,
		name: "demo - paris"
	},
	ns: "omero.web.figure.json"
}
```

screen shot 2016-06-17 at 11 47 51

Since there's quite a bit to digest here, I think the easiest would be for me to demo this to those interested and we can discuss where we go from here. cc @aleksandra-tarkowska

**Comment by joshmoore** [comment]
Nice, @will-moore. Looking forward to it.

**Comment by jburel** [comment]
Do you want to demo it during next week web meeting?

**Comment by will-moore** [comment]
@jburel Sure - Tuesday at 10 am. Can you make it @joshmoore?

**Comment by joshmoore** [comment]
We have visitors Tues & Wed. I'll let you know the schedule ASAP.

**Comment by joshmoore** [comment]
Points discussed on zoom with @will-moore:

 * selection of one or more key-value pairs for execution
 * auto-discovery of plugins & javascripts (from third parties)
 * planning for the first breaking change (can we support 2 versions at once?)
 * scalability: what happens if I want to generate a heatmap from several thousand images (Will this require POST'ing json to the plugins?)

**Comment by will-moore** [comment]
Presented this PR today at group meeting: http://will-moore.github.io/presentations/2016/OpenWith-Filtering-June-2016/
Extra functionality suggested:
 - isEnabled() function to get data via AJAX (Douglas)
 - if isEnabled() returns false, include tooltip for reason (Roger)
 
If there is the possibility of future restricting the 'selected' data that gets passed to isEnabled(selected), E.g. images in jsTree may not always have sizeX, sizeY etc. we could defensively document that to future-proof "Open with" plugins, only assume that E.g. id, type & name is available. If other data is missing then return true and it is the responsibility of the viewer/app to check if it can handle data. E.g. OMERO.figure gives warning that it can't handle Big images.

@jburel Can we discuss what needs to be done for this PR. Don't want the PR to get too big.

**Comment by will-moore** [comment]
- Try/catch around isEnabled() & handleAction() to handle errors as well as possible.
 - Move support for FileAnnotations and Full-Viewer button to subsequent PRs.
 - Start conservatively, only passing in id, type & name to js functions.
 - Can pass additional attributes in future PRs, especially when launching from right panel when full image is loaded.

**Comment by jburel** [comment]
@will-moore: Could you add comments not included in this PR to https://trello.com/c/RLvGaXUK/124-list-of-available-viewer-plugins? and if possible add checklists with what will be implemented in the subsequent PRs.

**Comment by will-moore** [comment]
Can someone remove the ON_HOLD label now, as this is ready for testing.

File: components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/openwith.js [hunk 1]
```diff
@@ -0,0 +1,46 @@
+
+    // Example openwith script
+    // TEMP - for evaluation of https://github.com/openmicroscopy/openmicroscopy/pull/4630
+    // TODO: remove this before merging PR!
```
mtbc: Note this. :smiley:

will-moore:
Listed on today's standup. A full testing scenario was requested....

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base_container.html
```diff
@@ -195,6 +195,61 @@
...
         {% endif %}
+
+
+        // ** "Open With" config used by E.g. ome.tree.js **
+        WEBCLIENT.OPEN_WITH = [];
+        {% if ome.open_with %}
+            // This is a json dump, so we can add it directly without escaping
+            WEBCLIENT.OPEN_WITH = {{ ome.open_with|safe }};
+        {% endif %}
+        // Helper can be used by 'open with' plugins to add isEnabled()
```
chris-allan: Could these two functions be added to the JavaScript library rather than being placed here in the template?

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base_container.html
```diff
@@ -195,6 +195,61 @@
...
+                            console.log("Open with " + label + ": " + e);
+                            enabled = false;
+                        }
+                        return enabled;
+                    }
+                }
+            });
+        };
+        // Include any JavaScript snippets provided by 'open with' plugins
+        {% for ow in ome.open_with_scripts %}
```
chris-allan: If we're thinking about plugin based extensions here which are likely installed via `pip install` and then activated via OMERO.web configuration could we dispense with this paradigm? If there are issues with plugin JavaScript it is immensely difficult to debug when using anonymous self-invoking functions. 

Such a plugin would of course have one or more JavaScript files in its static directories like the current paradigm which could be included on the page and be required to register themselves with the "open with" functionality on page load.

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base_container.html
```diff
@@ -195,6 +195,61 @@
...
+            // look for label in OPEN_WITH
+            WEBCLIENT.OPEN_WITH.forEach(function(ow){
+                if (ow.label === label) {
+                    ow.isEnabled = function() {
+                        // wrap fn with try/catch, return true by default
+                        var args = Array.from(arguments);
+                        try {
+                            enabled = fn(args);
+                        } catch (e) {
+                            console.log("Open with " + label + ": " + e);
```
chris-allan: We have a standing rule of not including `console.log()`'s in the final source, yes?

/cc @aleksandra-tarkowska

components/tools/OmeroWeb/omeroweb/webclient/decorators.py
```diff
@@ -191,3 +192,35 @@ def load_settings(self, request, context, conn):
             c_plugins.append({
                 "label": label, "include": include, "plugin_id": plugin_id})
         context['ome']['center_plugins'] = c_plugins
+
+        open_with = settings.OPEN_WITH
```
chris-allan: Like @aleksandra-tarkowska has done with #4706, these settings and this logic need to be tested.

pwalczysko:
>To test:

    Configure OMERO.figure as ....

From the header of this PR, the first test (with OMERO.figure) passes.

pwalczysko:
RFE:
Remove the hardcoded values in 
``bin/omero config get omero.web.open_with
[["Test", "api_paths_to_object"], ["OMERO.figure", "new_figure", {"objects": ["images", "dataset"], "target": "tab"}], [["OMERO.figure", "new_figure", {"objects": ["dataset", "images"], "target": "tab"}]]]
``
and in the example in the header of this PR, adjust the number of square brackets for this line.

pwalczysko:
As for the script-based functionality here, we agreed with @will-moore that this is necessary to be spelled out in a doc-type of text, which @will-moore will open as a doc PR. After this, the test can be designed more precisely.

will-moore:
Closed temporarily to avoid conflicts with Ola's PR.

will-moore:
Since I would like OMERO.figure and other apps to be able to access the 'Open With' config, these should be available via a json API, instead of embedded via Django templates.
I'll look at this approach now...

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base_container.html
will-moore: Certainly for normal use we wouldn't want to `console.log()` but in this case we want to give some feedback for why the plugin is not working. Without a try/catch you'd see any error in the console log anyway, but we need to be more defensive here since an error would otherwise break the whole jsTree menu.

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
```diff
@@ -945,6 +945,38 @@ OME.hideScriptList = function() {
...
+        if (ow.label === label) {
+            ow.isEnabled = function() {
+                // wrap fn with try/catch, since error here will break jsTree menu
+                var args = Array.from(arguments);
+                try {
+                    enabled = fn.apply(this, args);
+                } catch (e) {
+                    // Give user a clue as to what went wrong
+                    console.log("Open with " + label + ": " + e);
+                    enabled = true;
```
waxenegger: I justed wanted to double-check whether the enabled is meant to be set to true in the error case

waxenegger:
I justed wanted to double-check whether the enabled is meant to be set to true in the error case:

https://github.com/will-moore/openmicroscopy/blob/bd98e312bd1669e4cfa3c260f88b4f1017281147/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js#L962

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base_container.html
waxenegger: @will-moore  makes as good point regarding the "communication" of the error.

I think the console message here is the only way really how a user can (when pointed towards the dev tools) share the error, given that it has to be caught and that the browser does not give you a 'backend type of log', the only way one could achieve that would be a logging service with ajax requests sent to it.

I realize, however, that console or any of its methods (despite all major browsers supporting it in their more recent versions - according to MDN) can be potentially undefined but a check for that can address that issue. I would use it very sparingly, this one case seems a potential candidate.

will-moore: console.log is supported in everything apart from IE8: http://caniuse.com/#feat=console-basic so I think we are OK.

will-moore:
@waxenegger Yes, I default to true since I wouldn't want an error in the JavaScript to prevent someone launching the "Open with". The enable/disable functionality is only meant to be a convenience to indicate 'disabled' if we know that plugin won't support it. But in cases where we don't know or there's an error, we allow the "Open with" plugin to be opened, then it can make the final decision on whether it supports the data.

waxenegger:
@will-moore, ok then. 

I tried out all the options according to the docs with OME.setOpenWithActionHandler and OME.setOpenWithEnabledHandler and it worked as described.

I attempted to break things to check whether anything bad would happen to the loading/working of the web client, especially if js is eval(ed) this is always a possibility.  The good news is that you try/catch within the handler and jquery deals with errors from getScript().

If there is something that could perhaps be contemplated it is some reaction to some of the errors but that's more of a philosophical question about error handling and who is to implement the actual open with options:

At the moment any script errors apart from a wrong file path (`/omero config set omero.web.open_with '[["viewer-ng", "viewer_ng_index",{"script_url":"webclient/i_dont_exist.js"}]]'`) which gives a 404 popup are silently swallowed. So in the error case when it prevents the default handler from being overridden than the default handler might not work at all like was the case with the viewer-ng I tried or (maybe worse) not as expected if say the handed in values are compatible enough to make it work but whatever else was supposed to be done in the custom handler was omitted. The open with option would still be enabled in that case as well.

That said the above problem is only ever relevant when the script option is used. Also, the ways to remedy the problem are likely not super elegant. With the jquery version as is all errors are caught in  $(document).ajaxError and one would have to find a way to pin down that the error was due to the script loading from open with but, what's the bigger challenge then is -I'm guessing- to disable the corresponding open with option in a non-messy way. 

Other than that everything's fine and I did not run into any problems/errors.

will-moore:
Thanks @waxenegger  for the review.
To summarise:
 - If the ```script_url``` is invalid (E.g. wrong url or the app is not installed) then we get a 404 and 'Open with' will be enabled for default ```image``` but without custom 'action' behaviour (probably not working).
 - Alternatively we could remove the default ```supported_objects``` and force users to specify the objects they support. Then if you've specified an invalid ```script_url``` (instead of ```supported_objects```) then your plugin would be disabled. Does this make sense?

waxenegger:
I think your suggestion (second bullet point) makes sense but I don't know how overs feel about removing the default. The motivation behind my comment was mainly that I thought it safer to disable an invalid configuration, that's all. That said the negative effects are -as far as I suspect- not an issue for the "main" app but limited to to opened pop-up/window/tab.

will-moore:
@waxenegger I made ```supported_objects``` OR ```script_url``` required and updated docs at https://github.com/openmicroscopy/ome-documentation/pull/1543

waxenegger:
@will-moore: that's working nicely. and if there is a script error in the handed in script the option is disabled which is good as well I'd say.

will-moore:
Thinking more about making this re-usable by other apps, such as OMERO.figure.
I think that instead of an 'action' handler for when the user clicks on the 'Open With' option, plugins could instead implement a 'get_url' function, and then the webclient / figure etc could open that url or provide a link for users to click on. This would mean that the 'Open with' list could be a list of links, instead of dynamically handled actions (as in the jsTree menu).
Question is, do we want to restrict 'Open with' plugins to simply providing a url (instead of doing something fancy in JavaScript)?
I think we can do everything we want to with a url. We could always add back an 'Action' handler later if needed?
cc @waxenegger @joshmoore @aleksandra-tarkowska @jburel @chris-allan.

will-moore:
I changed the `setOpenWithActionHandler()` to instead use `setOpenWithUrlProvider()` and the function you provide here should return a url instead of actually opening a tab/window with that url.
This allows the script to be used to provide links as used in OMERO.figure https://github.com/ome/omero-figure/issues/179.
Need to update docs...

joshmoore:
@will-moore : What impact does that change have on some of the as yet unimplemented ideas expressed earlier like https://github.com/openmicroscopy/openmicroscopy/pull/4630#issuecomment-217093392?

will-moore:
@joshmoore The last change doesn't affect extending the Open With functionality to other types or use in other places. In fact, it will make it more versatile.
Removing the exclude label...

will-moore:
Updated docs at https://github.com/openmicroscopy/ome-documentation/pull/1543 to use `setOpenWithUrlProvider()`. Docs can be used in re-test of this PR along with https://github.com/ome/omero-figure/pull/179

waxenegger:
Open With works perfectly within the webclient, I used both figure and viewer-ng (incl.setOpenWithUrlProvider) to open an image.

I could not manage to set it up for figure as shown in https://github.com/ome/omero-figure/pull/179. The only option available in the info tab was to open it in the webclient.

will-moore:
@waxenegger I tried with your config but didn't see any problems with omero-figure. Maybe we can try again next week?

waxenegger:
no worries, perhaps I was not quite up-to-date with the code, although I did use the latest master.

will-moore:
So, I think we are "Good to merge" on this PR. Let's move discussion of OMERO.figure open_with to https://github.com/ome/omero-figure/pull/179.
@waxenegger Also, if you could comment on the updated docs at https://github.com/openmicroscopy/ome-documentation/pull/1543 with any suggestions etc that would be great.

waxenegger:
Sounds good @will-moore.

I believe the documentation is good, in general. I like the code snippets, they helped me most.

I did forget about syncing after I created a script file in the plugin which is something that I do naturally when I develop. Perhaps it's that this was more of a configuration task almost that I forgot it was necessary. People who might approach it from that direction could forget as well and think all that was needed is call set/append config although one could argue that ideally the files for open with are/should be already part of the plugin. 

Don't know of if a mention of restarting/syncing web is useful or not...up to you.

will-moore:
Thanks @waxenegger I'll move your comment to the docs PR and update the docs.
@jburel It would be great to get this merged now, as it's conflicting with #4708 webgateway_api which is otherwise 'good to merge'. Thanks.

jburel:
Merging: Further improvements can be done in subsequent PRs

PR #4631, opened 2016-05-05
Enable Plate download

dominikl:
Enable Plate download, taking the `omero.policy.binary_access` property into account, [Trello - Allow SPW download in Insight](https://trello.com/c/ydtmsP1Z/75-allow-spw-download-in-insight)

**Test** : Add Plates to the `binary_access` property: `./omero config set omero.policy.binary_access "+read,+write,+image,+plate"` (restart server). Make sure you can download plates on the plate, plate acquisition ("Run") and well level, by the context menu on the treeviewer and central databrowser and download menu in the right hand side metadata panel. Note: This will always download the whole plate even if just specific wells are selected. That should probably raise a warning dialog, as it's not intuitively expected. What are your thoughts about that?

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/DownloadAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/dataBrowser/actions/DownloadAction.java
@@ -24,21 +24,25 @@ import java.awt.event.ActionEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.File;
+import java.util.ArrayList;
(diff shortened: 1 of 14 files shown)


pwalczysko:
Tested using Rundeck docker.
All download options on plates work as expected as indicated in the header of this PR.
Ready to merge.

jburel:
ArchivedLoader: ``@param objects The images to export.`` This should be ``objects`` to export

mtbc:
Last commit looks good to me!

dominikl:
Thanks @mtbc . I consider that as good to merge then, @pwalczysko already did the functional testing.

PR #4623, opened 2016-04-29
Merge develop into regions

dominikl:
This PR merges all the recent develop PRs including Ice 3.6. The last merge commit should fix a trivial conflict related to the rebasing of the `getROICount` PR.

PR #4633, opened 2016-05-05
Fix tests for Ice 3.6

mtbc:
Rebases @jburel's work from #4594 and removes extra `disconnect()`s in tests that either I added recently or that @sbesson accuses. Should improve integration test situation in CI without further regressions.


sbesson:
Thanks for the additional test patching @mtbc. Waiting on https://github.com/openmicroscopy/openmicroscopy/pull/4623 to be validated to merge all these PRs together.

PR #4632, opened 2016-05-05
Merge develop into regions 

dominikl:
This PR merges recent develop PRs ( up to cda8446 (incl), **not** including Ice 3.6).  
The last commit fixes a trivial conflict related to the rebasing of the getROICount PR.

PR #4634, opened 2016-05-05
have AbstractServerTest.resetGroupPerms use Chmod2

mtbc:
Attempts to correct failing tests in `RenderingSettingsServicePermissionsTest`.


mtbc:
I am open to simply removing and inlining the method if that is preferred, it's called only once.

sbesson:
Interestingly with this PR and https://github.com/openmicroscopy/openmicroscopy/pull/4633 and without https://github.com/openmicroscopy/openmicroscopy/pull/4632, all Java integration tests pass. Will relaunch the test suite with the latter PR included to bissect the cause of the RenderingSettingsServicePermissionsTest failures.

sbesson:
If `AbstractServerTest.resetGroupPerms` is effectively used in one class, I vote for removing it and migrating it it to the corresponding tests. This can be done in a follow-up PR. Integrations tests are now happily passing with this included.

PR #4635, opened 2016-05-06
5.2.3: final merge

sbesson:
This PR should be the last PR merging develop into regions before the upcoming reintegration.

The only commit to be reviewed is d0355b7 which includes a trivial conflict fix in `components/tools/OmeroWeb/omeroweb/webclient/controller/container.py`. All integration tests should keep passing.


sbesson:
8134023 pushed to address the flake8 failures due to the `Delete2` import removed in https://github.com/sbesson/openmicroscopy/commit/eab13c50d1ce122ee0b0a07171047f7a76cc5318. The orphaned annotation deletion should now be handled server-side as per #4585. /cc @will-moore

mtbc:
Final commit looks good to me.

sbesson:
@will-moore: is the `remove` method of `container.py` tested via an OMERO.web integration test or should we test the last commit via a manual workflow?

mtbc:
Comments *are* getting deleted from the DB upon deletion in web client so good to merge.

PR #4637, opened 2016-05-09
Metadata column column types and datasets

joshmoore:
This PR carries on from work by @emilroz  (PR 120) which makes it possible to specify the column types for OMERO.tables when using populate_metadata.py to parse the csv files and adds support for datasets and projects as the target object for a metadata table.


joshmoore:
Merging for IDR-0.0.5

atarkowska:
--rebased-to #5218

PR #4640, opened 2016-05-10
inline singly used test method

mtbc:
Inlines a short method that is used only once in integration tests. CI should remain green.

Suggested by https://github.com/openmicroscopy/openmicroscopy/pull/4634#issuecomment-217374147.


sbesson:
The LockTimeout failure from https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-java/269/ is unrelated. Otherwise this change makes sense. Thanks.

PR #4642, opened 2016-05-10
Don't overwrite files with same names

dominikl:
See [Ticket 13223](https://trac.openmicroscopy.org/ome/ticket/13223), respectively point on Trello check list [Trac Tickets to consider](https://trello.com/c/ChsrJ25E/81-trac-tickets-to-consider).
**Test** workflow mentioned in ticket.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AttachmentsTaskPaneUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AttachmentsTaskPaneUI.java
@@ -66,9 +66,6 @@ public class AttachmentsTaskPaneUI extends AnnotationTaskPaneUI {
     /** Flag to indicate if the FileAnnotations should be selectable */
     private boolean selectable;
     
-    /** The collection of annotations to replace. */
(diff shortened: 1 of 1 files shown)


mtbc:
Now if I try to attach another local file of the same name then both end up attached to the image with their differing content preserved.

mtbc:
Attachment tooltip has useful information for figuring which is which.

mtbc:
@kennethgillen: Is the above behavior good with you? If so, then this PR is good to merge.

dominikl:
I think in Web it's the same behaviour, a new local file is just additionally attached not matter if a file with the same name already exists.

sbesson:
@kennethgillen: you okay with merging this?

kennethgillen:
👍  phenomenal folks. This is a big improvement. It'd be worth checking what the 'attached files' list shows now, as I seem to remember there was a listing of attached files available somewhere in one of the UIs. Perhaps multiple repetitions of the same filename? Not sure I remember where in the UIs I could see that, though.

mtbc:
Yes, the name is repeated: click the little widgety things for the attachments in the metadata pane.

PR #4643, opened 2016-05-10
do not have Pixels.sha1 values of "Foo"

mtbc:
Addresses https://trello.com/c/QrfLtInf/68-should-we-really-be-creating-the-value-foo-in-our-user-s-production-databases by *not* leaving "Foo" hashes in the pixels table. Neither old nor new images should have a `Pixels.sha1` of "Foo". For instance,
```
bin/omero hql "SELECT id FROM Pixels WHERE sha1 = 'Foo'"
```
should return no IDs. Supersedes #4608.


sbesson:
Failures in https://ci.openmicroscopy.org/view/Breaking/job/OMERO-DEV-breaking-integration-java/532/ are unrelated to this PR. Also looking at the integration DB

```
[hudson@ome-c6100-3 ~]$ psql -U omero OMERO-DEV-breaking-integration 
psql (9.3.11)
Type "help" for help.

OMERO-DEV-breaking-integration=# select count(*) from pixels where sha1 = 'Foo';
 count 
-------
     0
(1 row)

OMERO-DEV-breaking-integration=# select count(*) from pixels where sha1 = 'Pending...';
 count 
-------
   445
(1 row)
```

Removing `breaking` label for inclusion in merge builds.

sbesson:
```
OMERO-DEV-merge-deploy=# select count(*) from Pixels where Pixels.sha1='Foo';
 count 
-------
     0
(1 row)

OMERO-DEV-merge-deploy=# select count(*) from Pixels where Pixels.sha1='Pending...';
 count 
-------
  2184
(1 row)
```

sbesson:
And importing an image via the command-line using the `--skip all` option also creates an additional Pixels entry with a `Pending...` SHA1. Works as expected. Is there anything else that should be checked here @mtbc ?

mtbc:
No, not that I can think of.

PR #4647, opened 2016-05-11
have tooltip not add units to file size

mtbc:
Attach files to data in Insight. Check that when you hover over the filename in the metadata pane then the tooltip sensibly reports the file size. (It would previously use units like `KBkb`.)


gusferguson:
@mtbc 

Tested with OMERO.insight-5.3.0-m2-428-44a34b4-ice36-b310-mac on eel user-1 private-1

Attached files of xx KB x MB and xx.x MB from local disc.
Tooltips reported appropriate units on all.
Did note a slight discrepancy on one attachment - reported size = 33 MB - compared with size in Finder = 35.5 MB, but this is not significant and I presume due to different ways of calculating.

Good to merge.

main

PR #4648, opened 2016-05-11
index images by ROIs' folders

mtbc:
Testing: Put images' ROIs into folders then see if search in the graphical clients allows you to find those images by the folder names of their ROIs. This should be tested on both (i) a server upgraded from a previous DB and (ii) one whose DB is freshly generated from a `OMERO5.3DEV__5.sql` file. In both cases fiddle with the data however you like to see if the search results (after a brief update delay) keep track with your changes. The courageous may also wish to read through the diff, in which case I am happy to be on hand to explain anything. (The search results may omit images whose ROIs were in folders *before* the upgrade to this DB patch but for users coming from OMERO 5.2 that won't matter.)

This PR does *not* change the actual UI at all. That is for https://trello.com/c/hXo6uNA3/2-scoping-team-ui to consider for subsequent PRs.

Also fixes http://trac.openmicroscopy.org/ome/ticket/13221.


mtbc:
throw travis die again

pwalczysko:
We split testing with @dominikl 
 - PW does the two test in the top paragraph of the description (upgraded and non-upgraded DB search indexing checks)
 - @dominikl does the ticket testing and code review

mtbc:
(Note: could probably use `bin/omero obj` to switch the link for the ticket if that's preferred to `psql`.)

dominikl:
Ticket is fixed, tagged an image, searched for the tag -> image was found; changed the parent of the `ImageAnnotationLink` in the db to an other image, searched again, only the other image was found.

dominikl:
Code looks fine, too (although I'm not very familiar with the (re)indexing process) 👍

pwalczysko:
>... (i) a server upgraded from a previous DB

Tested on trout breaking. Both Insight and Web behave as expected.

pwalczysko:
>  (ii) one whose DB is freshly generated from a OMERO5.3DEV__5.sql file

Tested using Rundeck beluga server omero-markspr.... Both Insight and Web behave as expected.

pwalczysko:
In summary, this seems to be working well. Ready to merge.

sbesson:
Removed `breaking` for inclusion into the daily merge build.

joshmoore:
From my side, a further interesting test would be to apply one of the mega folderification scripts and then check how long it takes the indexer to catch up, which is another way of saying that the number of events produced for re-indexing are within reason.

mtbc:
Relisted accordingly. I'm optimistic: it reuses the same machinery that we added for making the annotation-triggered reindexing more reasonable.

joshmoore:
@mtbc : I don't foresee any particular issues either, but it was the mega addition/deletion of map annotations which pointed to gh-4273

mtbc:
Absolutely: it was your work on this kind of thing that caused me to write "we". :+1:

mtbc:
@sbesson: Can you suggest a good / easy source of good data? I figure if we have a large plate with plenty of ROIs or something, then one can run folderize on that, watch the Indexer log, then delete it and watch the log some more?

sbesson:
Regarding performance testing, along the same lines as https://github.com/openmicroscopy/bioformats/pull/2338#issuecomment-219687452, I wonder if this all does not call us towards a round of testing at scale as we are nearing 5.3.0-m3.
For immediate testing, would fake HCS images with regions be sufficient @mtbc ?

mtbc:
Ah yes, your work on the fake reader, excellent. I'm sure it would be. I *knew* you knew the answer, I just couldn't remember what it was!

pwalczysko:
Imported ``HCSanalysis&plates=1&plateRows=16&plateCols=24&rectangles=100.fake`` to eel merge user-4. Then ran ``folderize-objects.py`` in its original form on the resulting plate. The script took 2 minutes approx. The output of indiexer log see below. Verified the indexing via webclient search as well as via sql. I have the feeling that the indexing was actually faster than reported below (238381 ms). The script finished about 13:37 (to compare with the appearance of the line below, 13:41).

```
2016-05-18 13:41:07,483 INFO  [   ome.services.fulltext.FullTextIndexer] (2-thread-3) INDEXED  385 objects in batch#16467  [ 238381 ms.]  ~100% done (4718263 of 4718263)
2016-05-18 13:41:07,538 ERROR [        ome.services.util.ServiceHandler] (2-thread-3) Method interface ome.services.util.Executor$Work.doWork invocation took 238440
```

pwalczysko:
Note that there is the word ``ERROR`` in the output on the second line https://github.com/openmicroscopy/openmicroscopy/pull/4648#issuecomment-220016516 - something to worry about @joshmoore ?

joshmoore:
4 minutes for the Indexer is not particularly bad, I'd think. We should likely update the default here.

mtbc:
@joshmoore: Anything more you'd want to check, while today's eel has the logs, etc.?

joshmoore:
Sorry I missed this notification. No, in general if the searcher clearly moved on to other work, then :+1:

sbesson:
Having looked at the Indexer WARN/ERROR logic, it looks like the configuration properties defined in https://github.com/openmicroscopy/openmicroscopy/blob/v5.3.0-m2/etc/omero.properties#L647 are never passed to the Indexer `server-template` (they were hardcoded prior to https://github.com/openmicroscopy/openmicroscopy/pull/2846) so at the moment the Indexer service is using the default `omero.throttling.method_time.{error,warn}` values .

sbesson:
Assuming the indexer RFEs gets carded independently, merging this DB change.

pwalczysko:
Filed https://trac.openmicroscopy.org/ome/ticket/13235#ticket

PR #4649, opened 2016-05-11
Adding Mask to OmeroReader

dgault:
Adding storeOmeroMask function to OmeroReader to test Mask BinData integration.


sbesson:
See https://github.com/openmicroscopy/bioformats/pull/2319#issuecomment-219007029 for the export testing. Merging.

PR #4651, opened 2016-05-11
cecog cleanup

jburel:
Remove cecog

Make sure that all ref have been removed.
Not much to test
No mention in the doc/scripts.


sbesson:
Is it preferable to drop immediately or should we go through a 5.3.0 deprecation and remove it in the following breaking version?

jburel:
It was not be used since 4.1 
Some scripts like "Run_cecog.py" no longer exists

Same applies for flim so I don't think it is necessary.

jburel:
The ```cecog``` work was initially introduced when @will-moore was working on em data few years ago

sbesson:
The CLI plugin removal is sensible for all the reasons you mention above. Similar to other deprecation strategies, we could even consider filtering it into a separate unmaintained repository. For the other method, we assume no one would have been relying on the `script_utils.uploadCecogObjectDetails` API. Do we know of anyone who consumed this work ?

jburel:
At the time, only @will-moore used it, maybe ingvard?
but that was a long time ago

will-moore:
As I remember @joshmoore added this mostly as a prototype when we were discussing Cecog integration, but it is not related to EM work. Don't know that it was ever used by anyone outside our group.

PR #4652, opened 2016-05-11
remove long-deprecated code

mtbc:
Removes various code that has been deprecated for some time now, mostly graphs-related. Adjusts tests to match. Make sure that CI is happy and that OMERO seems to continue to work!


mtbc:
throw the travis die

mtbc:
tables tests again

mtbc:
and again

components/blitz/src/omero/cmd/basic/DoAllI.java
```diff
@@ -259,12 +259,7 @@ public void init(Helper helper) {
             Map<String, String> allgroups = new HashMap<String, String>();
             allgroups.put("omero.group", "-1");
             ctx.publishMessage(new ContextMessage.Push(this, allgroups));
-            try {
-                // Process within -1 block.
-                GraphUtil.combineFacadeRequests(this.requests);
-            } finally {
-                ctx.publishMessage(new ContextMessage.Pop(this, allgroups));
-            }
+            ctx.publishMessage(new ContextMessage.Pop(this, allgroups));
```
joshmoore: can the push & pop not be removed if nothing is happening between?

joshmoore:
One minor comment, otherwise :thumbsup:

components/blitz/src/omero/cmd/basic/DoAllI.java
mtbc: I have no idea: I don't know what they do nor who/what's listening for those (ha, that's often an issue with Insight) so I figure they must have been carried across somehow from older code. Happy to try removing them if you like!

joshmoore: Push/Pop were there to all `combineFacadeRequests` do its work. Since it's no longer doing anything, they aren't needed.

PR #4653, opened 2016-05-11
update uses of Requests to avoid deprecated methods

mtbc:
With this PR the only remaining uses of deprecated `Requests` methods are in the integration tests. CI should remain green. This probably mostly affects permissions operations from Insight.

### DIFF
--- a/components/blitz/src/omero/gateway/facility/DataManagerFacility.java
+++ b/components/blitz/src/omero/gateway/facility/DataManagerFacility.java
@@ -43,7 +43,6 @@ import omero.api.IUpdatePrx;
 import omero.cmd.CmdCallbackI;
 import omero.api.RawFileStorePrx;
 import omero.cmd.Delete2;
-import omero.cmd.Request;
 import omero.cmd.Response;
 import omero.cmd.graphs.ChildOption;
 import omero.gateway.Gateway;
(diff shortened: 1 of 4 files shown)


components/blitz/src/omero/gateway/facility/DataManagerFacility.java
```diff
@@ -226,35 +226,11 @@ public CmdCallbackI deleteFolders(SecurityContext ctx,
...
-                final String objectClassName = objectClass.getSimpleName();
-                /* then add the object's ID to the list for that class name */
-                final Long objectId = object.getId().getValue();
-                List<Long> idsThisClass = objectIds.get(objectClassName);
-                if (idsThisClass == null) {
-                    idsThisClass = new ArrayList<Long>();
-                    objectIds.put(objectClassName, idsThisClass);
-                }
-                idsThisClass.add(objectId);
+                request.target(object);
```
dominikl: Could just do `request.target(objects)` instead of for loop, right?

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
```diff
@@ -281,35 +257,11 @@ public CmdCallbackI delete(SecurityContext ctx, List<IObject> objects)
...
```
dominikl: Same here

components/blitz/src/omero/gateway/facility/DataManagerFacility.java
mtbc: Alas no, thanks to Java dreadfully having poor-man's generics: the type erasure doesn't let the builder distinguish all the kinds of collection safely, and `Iterable<IObject>` was one of the alternatives I thus omitted. (Of course there are erasure issues with object arrays too.)

dominikl: Ah yes, of course. Alright then!

dominikl:
Code looks good, integration tests passed, and Insight seems to be happy, too. Good to merge 👍 
Should check at some point if there's more code which could be greatly simplified by using the request builders.

PR #4655, opened 2016-05-12
Prefix linkify_element with OME.

joshmoore:
A rebase issue, this left many parts of the UI unusable
(link button, k/v dropdown, etc) after certain operations.

See: https://trello.com/c/gmCgOvwU/326-metadata52-web-problems

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
@@ -116,6 +116,7 @@ var MapAnnsPane = function MapAnnsPane($element, opts) {
                 $mapAnnContainer.html(html);
 
                 // Finish up...
+                OME.linkify_element($( "table.keyValueTable tbody tr td" ));
                 OME.filterAnnotationsAddedBy();
                 $(".tooltip", $mapAnnContainer).tooltip_init();
             });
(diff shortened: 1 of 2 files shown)


manics:
What are the blockers preventing this from going into mainline?

joshmoore:
Most likely just the agreement that we want to support it long-term. cc @will-moore @gusferguson @jburel @waxenegger @aleksandra-tarkowska etc.

waxenegger:
if it's about the prefix I'm all for it. it's good, I'd go further, necessary to namespace js as well.

unfortunately I was using 'ome', not all capitals but I can change that easily.

once there is a 'little repository for ui components/widgets' it will be even more important that they all live within something like: OME.theirownnamespace.WhateverComponent

joshmoore:
@waxenegger : I'm pretty sure that the move has been generally toward `OME.` so for the moment, that's not too at question. I think @manics is more concerned with evaluating whether or not 

```
                 // Temporary hack to get links to show in MapAnnotations
```

can be removed, i.e. that the link-ification becomes mainline.

will-moore:
This doesn't work in mainline, since we now load the contents of annotation panels on the fly, when they are expanded - so there's nothing there to linkify when this is called.

It needs to go here:
```
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
@@ -116,6 +116,7 @@ var MapAnnsPane = function MapAnnsPane($element, opts) {
                 $mapAnnContainer.html(html);
 
                 // Finish up...
+                OME.linkify_element($( "table.keyValueTable tbody tr td" ));
                 OME.filterAnnotationsAddedBy();
                 $(".tooltip", $mapAnnContainer).tooltip_init();
             });
```

Then it's just a question of how it works when you're able to edit the map annotation. It seems to behave OK just trying it myself now. It depends if you want to linkify as soon as the user edits a key/value or just when the table loads (as above).

joshmoore:
@will-moore : pushed.

joshmoore:
Merging for IDR-0.0.5

PR #4656, opened 2016-05-12
Update guava CacheBuilderSpec link in config

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/343/warnings3Result/ - docs build was unstable this morning because this link is broken. I'm not sure if it's temporary but the `/git/javadoc/com/google/common/cache/CacheBuilderSpec.html` bit is giving a 404 and the `http://docs.guava-libraries.googlecode.com` domain is redirecting to this github repo instead so I've changed the link to point directly at the code.

Will need the autogen docs build to run before this will feed through into the docs page.

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -258,7 +258,7 @@ omero.scripts.cache.cron=0 0 0 * * ?
 # for how long.
 #
 # For more information, see
-# https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/CacheBuilderSpec.java
+# http://google.github.io/guava/releases/17.0/api/docs/com/google/common/cache/CacheBuilderSpec.html
 omero.scripts.cache.spec=maximumSize=1000
 
 #############################################
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -258,7 +258,7 @@ omero.scripts.cache.cron=0 0 0 * * ?
 # for how long.
 #
 # For more information, see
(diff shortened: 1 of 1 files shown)


etc/omero.properties
```diff
@@ -258,7 +258,7 @@ omero.scripts.cache.cron=0 0 0 * * ?
 # for how long.
 #
 # For more information, see
-# http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/cache/CacheBuilderSpec.html
+# https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/CacheBuilderSpec.java
```
sbesson: The main issue here is that the Javadoc link becomes a source code link on the master branch. I would use the occasion to tie the link to the version of guava used in OMERO (currently 17.0 /cc @mtbc). We might also want to use the Javadoc published on GitHub pages i.e. http://google.github.io/guava/releases/17.0/api/docs/ although it is unclear to me on whether this should be the reference Javadoc URL for guava.

hflynn: I can use the api doc link instead if you prefer, doesn't seem to allow linking to the actual class though.

I suppose I could try opening an issue and asking what's going on with their javadoc links...

hflynn: https://github.com/google/guava/issues/2474

hflynn:
Broken link is permanent, project advised that the github hosted API are the definitive reference URL so I've fixed the link to use the release version we are currently using - like for like replacement now.

PR #4659, opened 2016-05-12
fix typo in Insight dialog

mtbc:
This dialog appears for the load ROI button in the measurement tool. See the diff for the change.


pwalczysko:
Tested on latest (typo present) as well as merge (typo fixed). 
This PR is ready to merge.


An aside here: Was actually surprised by this half-forgotten dialog/workflow. As we are now rolling out new features, this might help to offer a UI-y way to import them, see also testing of https://github.com/openmicroscopy/bioformats/pull/2319 and the folders workflows:

   - would I be able to have masks in an xml and upload them like this ?
   - would I be able to have folders in an xml and upload them like this ?

screen shot 2016-05-13 at 10 11 59

@dgault @sbesson @dominikl @jburel

jburel:
@pwalczysko:
The loading of ROI using XML will only handle standard shapes so no folder/masks.
Note that the format of that XML file won't follow the model, this was from the time when rois generated by the measurement tool could not be saved server side. 
This feature should be turned off unless we want people to upload "external ROIs"
but that will require the introduction of "ROIReader"

pwalczysko:
Maybe we should turn it off then ? Obviously it elicites too much hopes ...

jburel:
Better yes

jburel:
@mtbc: do you want to take care of that in this PR?

mtbc:
Would it suffice to just track the button down and do a `setVisible(false)`?

jburel:
yes, since we may want later on to add support for "external rois"

mtbc:
Okay, will do.

mtbc:
Unfortunately I really don't know how to get this past travis. (Overrides in `OMEROMetadataStoreClient`.)

dominikl:
Could build the PR locally without problem, no idea what's going on with Travis. "Load ROIs" button is removed from the toolbar, good to merge 👍

mtbc:
--depends-on #4675

jburel:
remove depends on to check travis again

PR #4662, opened 2016-05-13
Revert "Cleanup local OME Maven cache for every Travis build"

sbesson:
This reverts commit 70ab7a67c2daa72c40c70550e10f89336505bbbf.

With the large part of the breaking Bio-Formats changes having been included, purging the `ome` directory in the Maven cache should not be necessary anymore. With this change Travis should remain green and upcoming builds should not need to fetch from Artifactory.


joshmoore:
Makes sense. Perhaps something that can be included again later based on branch naming?

mtbc:
Good to merge.

PR #4665, opened 2016-05-16
fix method name for Chown2Builder.toGroup

mtbc:
This is a simple code PR that fixes a copy-and-paste slip in method naming. CI should remain green.

(As this affects gateway API it is a 5.2 backport candidate; I've no idea where/how we note those.)

### DIFF
--- a/components/blitz/src/omero/gateway/util/Requests.java
+++ b/components/blitz/src/omero/gateway/util/Requests.java
@@ -2210,8 +2210,18 @@ public class Requests {
          * @param user the user to which to give the target objects, does overwrite previous calls
          * @return this builder, for method chaining
          */
-        public Chown2Builder toGroup(Experimenter user) {
+        public Chown2Builder toUser(Experimenter user) {
             return toUser(user.getId());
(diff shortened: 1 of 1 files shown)


mtbc:
doesn't look to be a real problem, thowing the travis die again ...

mtbc:
and again ...

components/blitz/src/omero/gateway/util/Requests.java
```diff
@@ -2210,9 +2210,19 @@ public Chown2Builder toUser(RLong id) {
...
+        public Chown2Builder toUser(Experimenter user) {
             return toUser(user.getId());
         }
+
+        /**
+         * @param user the user to which to give the target objects, does overwrite previous calls
+         * @return this builder, for method chaining
+         * @deprecated use {@link #toUser(Experimenter)}
+         */
+        @Deprecated
```
joshmoore: Kudos for the deprecation.

joshmoore:
Thanks for battling this one out with travis.

PR #4666, opened 2016-05-16
TablesFacility

dominikl:
Resurrected the `TablesFacility` from  #4347 (which was closed and only a subset merged into 5.2.3).

The TablesFacility would provide an easy way for adding and getting table structured data to/from OMERO, this would be very useful for the R gateway (i.e attaching/loading dataframes to/from OME objects, https://github.com/dominikl/rOMERO/blob/dev_5_3/examples/dataframes.R ).

Probably needs some more discussion, but it would be very handy to open up the tables accessibility, so I could create some more realistic R examples.

**Test**
Basically check if the `TablesFacility.java` 'API' makes sense and the according integration test is sufficient and passes. Optional: Run above mentioned `dataframes.R` example code.

**Status**
In my opinion this could be merged. But there was still some discussion going on about the tables mimetype, without coming to a conclusion yet.

/cc @jburel

### DIFF
--- a/components/blitz/src/omero/gateway/facility/TablesFacility.java
+++ b/components/blitz/src/omero/gateway/facility/TablesFacility.java
@@ -31,7 +31,6 @@ import omero.gateway.exception.DSOutOfServiceException;
 import omero.gateway.model.AnnotationData;
 import omero.gateway.model.DataObject;
 import omero.gateway.model.FileAnnotationData;
-import omero.gateway.model.FileData;
 import omero.gateway.model.ImageData;
 import omero.gateway.model.MaskData;
 import omero.gateway.model.PlateData;
@@ -432,10 +431,10 @@ public class TablesFacility extends Facility {
      *             If an error occurred while trying to retrieve data from OMERO
      *             service.
      */
-    public Collection<FileData> getAvailableTables(SecurityContext ctx,
-            DataObject parent) throws DSOutOfServiceException,
-            DSAccessException {
-        Collection<FileData> result = new ArrayList<FileData>();
+    public Collection<FileAnnotationData> getAvailableTables(
+            SecurityContext ctx, DataObject parent)
+            throws DSOutOfServiceException, DSAccessException {
+        Collection<FileAnnotationData> result = new ArrayList<FileAnnotationData>();
         try {
             MetadataFacility mf = gateway.getFacility(MetadataFacility.class);
 
@@ -448,9 +447,7 @@ public class TablesFacility extends Facility {
             for (AnnotationData anno : annos) {
                 FileAnnotationData fad = (FileAnnotationData) anno;
                 if (fad.getOriginalMimetype().equals(TABLES_MIMETYPE)) {
-                    long fileId = ((FileAnnotationData) anno).getFileID();
-                    OriginalFile file = new OriginalFileI(fileId, false);
-                    result.add(new FileData(file));
+                    result.add(fad);
                 }
             }
 
--- a/components/blitz/src/omero/gateway/facility/TablesFacility.java
+++ b/components/blitz/src/omero/gateway/facility/TablesFacility.java
@@ -1,0 +1,527 @@
+/*
+ * Copyright (C) 2016 University of Dundee & Open Microscopy Environment.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+package omero.gateway.facility;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.UUID;
+
+import omero.gateway.Gateway;
+import omero.gateway.SecurityContext;
+import omero.gateway.exception.DSAccessException;
+import omero.gateway.exception.DSOutOfServiceException;
+import omero.gateway.model.AnnotationData;
+import omero.gateway.model.DataObject;
+import omero.gateway.model.FileAnnotationData;
+import omero.gateway.model.FileData;
+import omero.gateway.model.ImageData;
+import omero.gateway.model.MaskData;
+import omero.gateway.model.PlateData;
+import omero.gateway.model.ROIData;
+import omero.gateway.model.TableData;
+import omero.gateway.model.WellSampleData;
+import omero.grid.BoolColumn;
+import omero.grid.Column;
+import omero.grid.Data;
+import omero.grid.DoubleArrayColumn;
+import omero.grid.DoubleColumn;
+import omero.grid.FileColumn;
+import omero.grid.FloatArrayColumn;
+import omero.grid.ImageColumn;
+import omero.grid.LongArrayColumn;
+import omero.grid.LongColumn;
+import omero.grid.MaskColumn;
+import omero.grid.PlateColumn;
+import omero.grid.RoiColumn;
+import omero.grid.SharedResourcesPrx;
+import omero.grid.StringColumn;
+import omero.grid.TablePrx;
+import omero.grid.WellColumn;
+import omero.model.FileAnnotation;
+import omero.model.FileAnnotationI;
+import omero.model.Image;
+import omero.model.ImageI;
+import omero.model.OriginalFile;
+import omero.model.OriginalFileI;
+import omero.model.Plate;
+import omero.model.PlateI;
+import omero.model.Roi;
+import omero.model.RoiI;
+import omero.model.WellSample;
+import omero.model.WellSampleI;
+
+/**
+ * {@link Facility} to interact with OMERO.tables
+ * 
+ * @author Dominik Lindner &nbsp;&nbsp;&nbsp;&nbsp; <a
+ *         href="mailto:d.lindner@dundee.ac.uk">d.lindner@dundee.ac.uk</a>
+ */
+public class TablesFacility extends Facility {
+
+    /** The mimetype of an omero tables file */
+    public static final String TABLES_MIMETYPE = "OMERO.tables";
+    
+    /**
+     * Creates a new instance
+     * 
+     * @param gateway
+     *            Reference to the {@link Gateway}
+     */
+    TablesFacility(Gateway gateway) {
+        super(gateway);
+    }
+
+    /**
+     * Adds a new table with the provided data
+     * 
+     * @param ctx
+     *            The {@link SecurityContext}
+     * @param target
+     *            The object to attach the table to
+     * @param name
+     *            A name for the table (can be <code>null</code>)
+     * @param data
+     *            The data
+     * @throws DSOutOfServiceException
+     * @throws DSAccessException
+     */
+    public void addTable(SecurityContext ctx, DataObject target, String name,
+            TableData data) throws DSOutOfServiceException, DSAccessException {
+        try {
+            if (name == null)
+                name = UUID.randomUUID().toString();
+
+            String[] columnNames = data.getColumnNames() != null ? data
+                    .getColumnNames() : new String[0];
+
+            String[] description = data.getDescriptions() != null ? data
+                    .getDescriptions() : new String[0];
+
+            Column[] columns = new Column[data.getTypes().length];
+            for (int i = 0; i < data.getTypes().length; i++) {
+                String cname = columnNames.length > i ? columnNames[i] : "";
+                String desc = description.length > i ? description[i] : "";
+                Object[] d = data.getData().length > i ? data.getData()[i]
+                        : new Object[0];
+                columns[i] = createColumn(cname, desc, data.getTypes()[i], d);
+            }
+
+            SharedResourcesPrx sr = gateway.getSharedResources(ctx);
+            if (!sr.areTablesEnabled()) {
+                throw new Exception("Tables feature is not enabled on this server!");
+            }
+            long repId = sr.repositories().descriptions.get(0).getId()
+                    .getValue();
+            TablePrx table = sr.newTable(repId, name);
+            table.initialize(columns);
+            table.addData(columns);
+            table.close();
+
+            DataManagerFacility dm = gateway
+                    .getFacility(DataManagerFacility.class);
+            BrowseFacility browse = gateway.getFacility(BrowseFacility.class);
+
+            OriginalFile file = table.getOriginalFile();
+            file = (OriginalFile) browse.findIObject(ctx, file);
+
+            FileAnnotation anno = new FileAnnotationI();
+            anno.setFile(file);
+            FileAnnotationData annotation = new FileAnnotationData(anno);
+            annotation.setDescription(name);
+
+            annotation = (FileAnnotationData) dm.saveAndReturnObject(ctx,
+                    annotation);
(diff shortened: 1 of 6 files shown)


components/blitz/src/omero/gateway/facility/TablesFacility.java
```diff
@@ -0,0 +1,527 @@
...
+                for (int i = 0; i < columns.length; i++) {
+                    types[i] = Object.class;
+                }
+                return new TableData(header, descriptions, types,
+                        new Object[columns.length][0]);
+            }
+
+            int nRows = (int) table.getNumberOfRows();
+            Object[][] dataArray = new Object[columns.length][nRows];
+            Data data = table.read(colNumbers, 0, table.getNumberOfRows());
```
manics: Is there an expected maximum size for tables you'll be retrieving with this method?

dominikl: As the intention for now is to provide a simple way for adding/retrieving R dataframes the theoretical maximum could be 2^31 columns and 2^31 rows. Certainly doesn't make sense to read/write such a table at once; definitely more elaborated methods needed for big tables later.

joshmoore: @dominikl : but in terms of API design, we should really avoid all definitions of non-paging methods. One alternative would be to just define that a given method will default to only returning N rows. (This is what `bin/omero hql` does) IQuery defaulted to returning everything in the early days and it's not yet been deprecated despite the problems it causes.

jburel: probably time to deprecate the "dangerous" methods

dominikl: Yes, I can see the point. Would be weird to introduce methods now, which are likely to get deprecated later because they cause problems. I'll add pagination to the methods, makes more sense than setting a more or less random limit.

jburel:
It might be good to test it against some IDR data

manics:
What needs to be inside the table for this to be testable? I'm working on some large feature files but they're currently missing IDs inside the tables.

jburel:
``public TableData getTable(SecurityContext ctx, long fileId, long rowFrom, long rowTo, int... columns)``
Add null check for columns.

Also we should probably log the error if we cannot close the table for some reasons.

dominikl:
Added `null` check 19753b9  and logging a5615f1 (rebased to develop, that's why comments and commits are not in sync).

manics:
I've got a 31 GB OMERO.tables file which you're welcome to try

dominikl:
@manics Is it accessible on a server somewhere?

manics:
IDR clone Screen:206, ping @joshmoore for login details

dominikl:
Tested with @manics 31GB table. Only problem was, that the table doesn't show up with `getAvailableTables()`, I guess because its mime type is *not* set to `OMERO.table`, but you can access it directly using the fileID.
If you call `getTable()` without any limitation, the first 1000 rows are fetched, which takes 17 sec. Tried a few more calls explicitly requesting 2000 ( 41sec), 3000 ( 51sec), 10000 (178 sec), and 10000 rows but only 50 columns (59 sec) seems to be pretty linear.

jburel:
maybe add a new method with the mimetype as a parameter.

manics:
I'm not sure adding a mime-type is the right way to go, it might be better to ignore the mime-type and filter by namespace instead since bulk-annotations should have a namespace, and the features table should too (though I forgot to set it).

jburel:
@manics : namespace is probably more suitable, thanks

dominikl:
I don't get it... my intention with `getAvailableTables()` was, to simply list all tables which are attached to an object, not caring about if it's bulk annotations, features or whatever. I expected that all tables have the mime type "OMERO.table", because for example, if I create a table with the Java API, it will automatically have this mime type. Not sure, what to do with a namespace there...

manics:
In that case how about have an option of either the default mimetype or ALL FileAnnotations, and leave the decision on namespaces for a later discussion?

joshmoore:
Any discussion needed here?

dominikl:
Probably needs more discussion. Is there a reason why there are OMERO.tables without the correct mime type, what's the problem with setting it? If I remember correctly, I added a namespace for "user generated" OMERO.tables initially, but the conclusion was, it's not necessary, because there's already a mime type for it, and would just add unnecessary complexity.

manics:
> Is there a reason why there are OMERO.tables without the correct mime type, what's the problem with setting it?

Primarily that no-one knows about it, it's set automatically if you create the table using the OMERO.tables API but not if you create the file through other methods.

dominikl:
Ok. Then for the Java Gateway it would be reasonable to only deal with tables with "OMERO.table" mime type. If you create a table with the Java Gateway, which uses the tables API, it will have the mime type set and you can access it again. If you add tables by other methods, you'd probably also use other methods to deal with it and not the Java Gateway. Does that make sense?

joshmoore:
--rebased-to #4761

jburel:
next steps?
usage in metadata52 will give feedback. (one extra commit 83c7aaf)
something to review after the idr deadline

components/blitz/src/omero/gateway/model/TableDataColumn.java
```diff
@@ -0,0 +1,213 @@
...
+     * Creates a new instance
+     * 
+     * @param name
+     *            The header
+     * @param index
+     *            The index
+     * @param type
+     *            The type of data in this column
+     */
+    public TableDataColumn(String name, int index, Class<?> type) {
```
jburel, line 57: could be replaced by this(name, index, "", class) so there is only one place to assign value

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,294 @@
...
+        this.offset = offset;
+    }
+
+    /**
+     * @return <code>true</code> if the last available row is contained,
+     *         <code>false</code> if there's more data available in the original
+     *         table
+     */
+    public boolean isCompleted() {
+        if (data == null || data.length == 0)
```
jburel: could add a ``isEmpty`` method
i.e.
`` boolean isEmpty() { return (data == null || data.length == 0); }``

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,294 @@
...
+
+        return (offset + data[0].length) == numberOfRows;
+    }
+
+    /**
+     * Manually set completed state (sets the {@link TableData#numberOfRows} to
+     * the last row in the {@link TableData#data} array)
+     */
+    public void setCompleted() {
+        this.numberOfRows = (data == null || data.length == 0) ? 0 : offset
```
jburel: use ``isEmpty`` method

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,294 @@
...
+        return true;
+    }
+
+    /**
+     * Generates a hash code by iterating over all elements and casting them to
+     * their proposed classes
+     * 
+     * @param objects
+     *            The array to generate the object for
+     * @param types
```
jburel: columns is the argument

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,294 @@
...
+
+        if (columns != null) {
+            for (int i = 0; i < columns.length; i++) {
+                sb.append(columns[i]);
+                sb.append('\t');
+            }
+            sb.append('\n');
+        }
+
+        if (data == null || data.length == 0)
```
jburel: use ``isEmpty``

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,294 @@
...
+                    result = prime * result + castedObject.hashCode();
+            }
+        }
+
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
```
jburel, line 285: concern about thread safety?
if yes, ``StringBuffer`` should be used

components/blitz/src/omero/gateway/model/TableDataColumn.java
```diff
@@ -0,0 +1,213 @@
...
+        } else if (!type.getCanonicalName().equals(
+                other.type.getCanonicalName()))
+            return false;
+        return true;
+    }
+
+    @Override
+    public String toString() {
+        String s = name;
+        if (description.trim().length() > 0)
```
jburel: ``null`` check for description required
better to use method from  ``StringUtils``

components/blitz/src/omero/gateway/model/TableData.java
mtbc: I don't think multiple threads can access the same instance.

joshmoore: Agreed. Definitely better to move to `StringBuilder` and not let the instance escape a given method.

dominikl: Now I'm a bit confused, did you mean move to `StringBuffer` ?

joshmoore: Sorry, no. :+1: for `StringBuilder`

dominikl:
There was still the question about the mimetype for OMERO.table files. Within the gateway methods, I assume the mimetype of a table file is set to `OMERO.tables`, because that's the mimetype which is used when a new table is created via SharedResources service: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/ome/services/blitz/impl/SharedResourcesI.java#L328 I use this mimetype to look for tables in the `getAvailableTables()` method. One can still load tables with other/no mimetype, just have to know the file id in that case, see `getTable()` methods.

components/blitz/src/omero/gateway/model/TableData.java
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * Creates a new instance
+     * 
+     * @param columns
+     *            The column definitions
+     * @param data
+     *            The data in form of List of columns
+     */
+    public TableData(List<TableDataColumn> columns, List<List<Object>> data) {
+        this.columns = new TableDataColumn[columns.size()];
```
jburel: maybe some ``null`` check could be added unless you are happy with ``NPE``

dominikl:
Added a test which uses a table with 10 columns and 100000 rows (that'll create a tables file between 3 and 6 GB), can't make it much bigger without running into memory issues.

dominikl:
--depends-on #5139

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -88,12 +96,92 @@ public void testGetSubsetTable() throws Exception {
...
+                        Assert.assertEquals((Long)data2[c][r-(int)td2.getOffset()], (Long)data[index][r]);
+                    } else if (type.equals(Double.class)) {
+                        Assert.assertEquals((Double)data2[c][r-(int)td2.getOffset()], (Double)data[index][r]);
+                    } else if (type.equals(Double[].class)) {
+                        Assert.assertEquals((Double[])data2[c][r-(int)td2.getOffset()], (Double[])data[index][r]);
+                    }
+                }
+            }
+        }
+        System.out.println("Requesting random subsets (average size: "+(nEntries/10)+" entries) took "+(duration/10)+" ms on average");
```
jburel: you could add option like
``@Test(invocationCount = 5, threadPoolSize = 5)``
if you want to run the test few times, I imagine that the test will be over the timeout limit so you will probably have to add ``timeOut=Value``

dominikl:
One more thing which should be clarified. At the moment, the maximum number of rows which are returned is limited to 1000, but **only** if you call `getTable(...)` without specifying `rowFrom` and `rowTo`. If you provide these parameters you can request more than 1000 rows (my comment in the TablesFacilityTest is not correct, I'll update that). Can that be a problem? My initial thought about that was, that if you deliberately request a certain amount of rows, you're probably aware of the performance issues that might cause. The question is, should that be a hard limit instead?

manics:
Sounds like a more general question across the whole API rather than something specific to this PR, e.g. for Hibernate queries which can cause a server OOM. Should there be a limit, and if so do you completely fail or return `limit` results with a status to indicate you should make a follow-up request?

joshmoore:
An absolute limit should probably be in place server-side and the client more has to be ready to handle the appropriate exception (`OverUsageException`)

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -136,75 +84,90 @@ public void testBigTable() throws Exception {
...
+        Object[][] origData = original.getData();
+
+        TableData info = tablesFacility.getTableInfo(rootCtx,
+                original.getOriginalFileId());
+        int rows = (int) info.getNumberOfRows();
+        int cols = info.getColumns().length;
+
+        int rowFrom = rand.nextInt(rows - 1000); // can't request more than 1000
+                                                 // rows at once
+        int rowTo = rowFrom + rand.nextInt(1000);
```
mtbc: it's okay if `rowTo == rowFrom`?

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -136,75 +84,90 @@ public void testBigTable() throws Exception {
...
+
+        for (int x = 0; x < columns.length; x++)
+            columns[x] = rand.nextInt(cols);
+
+        long start = System.currentTimeMillis();
+        TableData td2 = tablesFacility.getTable(rootCtx,
+                original.getOriginalFileId(), rowFrom, rowTo, columns);
+        System.out.println("Requesting random subset of size: "
+                + ((rowTo - rowFrom) * columns.length) + " entries, took "
+                + (System.currentTimeMillis() - start) + " ms");
```
mtbc: maybe could mention more as a "from, to" explicitly giving `rowFrom`, `rowTo`: if the test fails, there might then be enough information on the console to reproduce the problem

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -0,0 +1,187 @@
...
+        int rowTo = rowFrom + rand.nextInt(1000);
+        int[] columns = new int[rand.nextInt(cols)];
+
+        for (int x = 0; x < columns.length; x++)
+            columns[x] = rand.nextInt(cols);
+
+        long start = System.currentTimeMillis();
+        TableData td2 = tablesFacility.getTable(rootCtx,
+                original.getOriginalFileId(), rowFrom, rowTo, columns);
+        System.out.println("Requested random subset row " + rowFrom + " to "
```
jburel: remove ``System.out`` unless you need for clarity

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+        original = new TableData(header, data);
+        long start = System.currentTimeMillis();
+        TableData stored = tablesFacility.addTable(rootCtx, ds, "Big Table",
+                original);
+        long duration = System.currentTimeMillis() - start;
+
+        original.setOriginalFileId(stored.getOriginalFileId());
+
+        System.out.println("Storing table with " + (nCols * nRows)
```
jburel: are you expecting that to be done in a given time interval?
There is a timeout parameter so it will fail if it takes too long

dominikl: Timeout parameter shouldn't be necessary, I'd expect not more than 100-300ms.

jburel:
Tests are now green.
Will merge when today's testing is done

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.gateway/TablesFacilityTest/testAddTable/ failed with timeout error

dominikl:
Increased the timeout to 600sec. But I guess if it takes that long, something else is wrong with the build.

jburel:
Something to keep an eye on. Merging

PR #4667, opened 2016-05-18
Fixed ClassCastException

dominikl:
Tiny bugfix for [QA 17191](https://www.openmicroscopy.org/qa2/qa/feedback/17191/) to prevent potential ClassCastException (`result` object is not guaranteed to be an instance of Set, e.g. see https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/facility/BrowseFacility.java#L1008 ).

**Test**: Code review should be sufficient. Don't know how to easily replicate the bug. Would need an experimenter without user image set, but all experimenters actually have this user icon as default user image.

### DIFF
--- a/components/blitz/src/omero/gateway/facility/BrowseFacility.java
+++ b/components/blitz/src/omero/gateway/facility/BrowseFacility.java
@@ -1162,7 +1162,7 @@ public class BrowseFacility extends Facility {
         } catch (Throwable t) {
             logError(this, "Could not load orphaned images", t);
         }
-        return Collections.emptyList();
+        return Collections.emptySet();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jburel:
The change in that section of the code makes sense but the problem is coming from the gateway method itself.
in ``BrowseFacility#getOrphanedImages``: if the user has images, a set is returned, if no images, an empty list is returned.
but the ``getUserImages`` method always returns a list.

dominikl:
Ah yes, sorry, was refering to the wrong method in my comment. Just changed the method to always return a Set (see last commit). Although List would make more sense. Don't know why `PojoMapper.asDataObjects(...)` was implemented to return a Set in first place, as DataObjects don't overwrite `hashCode()` HashSets of DataObjects won't work as expected anyway.

jburel:
@dominikl: it might be a good opportunity to review the ``PojoMapper`` method: deprecated and introduced a new one

jburel:
@dominik: Do you want to review the ``PojoMapper`` in this PR?
Otherwise I will merge it

dominikl:
Ah yes, was originally planning to do it in this PR, but might be cleaner to do it in a separate PR. Just go ahead and merge it. Thanks.

jburel:
Card created https://trello.com/c/ifii0bT2/116-pojomapper-review

PR #4669, opened 2016-05-18
Chown tests

pwalczysko:
See [chown trello card](https://trello.com/c/RpdAOmKp/37-permissions-system-problems-chown#comment-5734d47e92374a1dec3292d5).
This PR should be a suggestion to cpmplement the tests and changes of existing tests in [chown trello card](https://trello.com/c/RpdAOmKp/37-permissions-system-problems-chown#comment-5734d47e92374a1dec3292d5). It deals with Python tests only. 

The tests added here should deal with the comments [comment1 on trello](https://trello.com/c/RpdAOmKp/37-permissions-system-problems-chown#comment-5734d6a14a903d90c3e3b14a) and [comment2 on trello](https://trello.com/c/RpdAOmKp/37-permissions-system-problems-chown#comment-5734d47e92374a1dec3292d50) as well as other tests which bring the [chown test](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_chown.py) in line with the [delete test](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_delete.py). 

The two parts which are not completely in line between delete and chown tests are "default exclusion" which is missing in chown logic (on CLI) and the Tag/Tagset workflow (which again, is intrinsically different for delete and chown). I have added just two tests representing these two paragraphs in [delete test](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_delete.py), namely https://github.com/openmicroscopy/openmicroscopy/pull/4669/commits/7c59c652cdf502af71b0afbe1983f99de8c64900 and https://github.com/openmicroscopy/openmicroscopy/pull/4669/commits/68385640dfafdaeea6b17fe3f3d79c65c1d9c9f3 which check the default behaviour in the private group of attachments and tags/tagsets. I think the permissions here should be checked via  the java integration test. 

To test:
Check that the integration test [chown test](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_chown.py) is passing and makes sense.

### DIFF
--- a/components/tools/OmeroPy/test/integration/clitest/test_chown.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_chown.py
@@ -268,6 +268,50 @@ class TestChown(CLITest):
             assert obj.id.val == i.id.val
             assert obj.details.owner.id.val == user.id.val
 
+    @pytest.mark.parametrize('number', [1, 2, 3])
+    @pytest.mark.parametrize("ordered", ordered)
+    def testMultipleSkipheadsPlusObjectsSeparated(self, number, ordered):
+        projs = [self.make_project() for i in range(number)]
+        dsets = [self.make_dataset() for i in range(number)]
+        imgs = [self.update.saveAndReturnObject(self.new_image())
+                for i in range(number)]
+
+        for i in range(number):
+            self.link(projs[i], dsets[i])
+            self.link(dsets[i], imgs[i])
+
+        ds = [self.make_dataset() for i in range(number)]
+
+        # Create user and transfer the objects to the user
+        client, user = self.new_client_and_user(group=self.group)
+        self.args += ['%s' % user.id.val]
+        for d in ds:
+            self.args += ['Dataset:%s' % d.id.val]
+        for p in projs:
+            self.args += ['Project/Dataset/Image:%s' % p.id.val]
+        if ordered:
+            self.args += ["--ordered"]
+        self.cli.invoke(self.args, strict=True)
+
+        # Check that only the images and separate datasets
+        # have been transferred
+        for p in projs:
+            obj = self.query.get('Project', p.id.val, all_grps)
+            assert obj.id.val == p.id.val
+            assert obj.details.owner.id.val == self.user.id.val
+        for d in dsets:
+            obj = self.query.get('Dataset', d.id.val, all_grps)
+            assert obj.id.val == d.id.val
+            assert obj.details.owner.id.val == self.user.id.val
+        for i in imgs:
+            obj = self.query.get('Image', i.id.val, all_grps)
+            assert obj.id.val == i.id.val
+            assert obj.details.owner.id.val == user.id.val
+        for d in ds:
+            obj = self.query.get('Dataset', d.id.val, all_grps)
+            assert obj.id.val == d.id.val
+            assert obj.details.owner.id.val == user.id.val
+
 
 class TestChownRoot(RootCLITest):
 
--- a/components/tools/OmeroPy/test/integration/clitest/test_chown.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_chown.py
@@ -499,7 +499,7 @@ class TestChown(CLITest):
 
     def testOutputWithElision(self, capfd):
         DATASETS = 8
-        # Import several images
+        # Create several datasets
         ids = []
         for i in range(DATASETS):
             ids.append(self.make_dataset().id.val)
@@ -508,7 +508,7 @@ class TestChown(CLITest):
         assert ids[-1] - ids[0] + 1 == DATASETS
         ids = [str(id) for id in ids]
 
-        # Create user and transfer some of those images
+        # Create user and transfer some of those datasets
         # to the user, mix up the order
         client, user = self.new_client_and_user(group=self.group)
         self.args += ['%s' % user.id.val]
--- a/components/tools/OmeroPy/test/integration/clitest/test_chown.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_chown.py
@@ -312,6 +312,49 @@ class TestChown(CLITest):
             assert obj.id.val == d.id.val
             assert obj.details.owner.id.val == user.id.val
 
+    @pytest.mark.parametrize('number', [1, 2, 3])
+    @pytest.mark.parametrize("ordered", ordered)
+    def testMultipleSkipheadsPlusObjectsInterlaced(self, number, ordered):
+        projs = [self.make_project() for i in range(number)]
+        dsets = [self.make_dataset() for i in range(number)]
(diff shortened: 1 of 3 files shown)


pwalczysko:
This commit adds some Fileset testing as the deal was ``follow what is in the Delete test```. This is nevertheless probably importing real images, so just heads up.

pwalczysko:
The last commit https://github.com/openmicroscopy/openmicroscopy/pull/4669/commits/7c59c652cdf502af71b0afbe1983f99de8c64900 is the only test which seems to fit for ``chown`` situation as opposed to ``delete`` situation. The other tests in this [group of delete tests](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_delete.py#L348) deal with "Default Exclusion" which is not present for Chown. The only way how to include them in chown tests would be to use Read-annotate group (instead of the private used in all the tests here and in Delete test). @ximenesuk @mtbc - is it something worthwile ?

pwalczysko:
This test https://github.com/openmicroscopy/openmicroscopy/pull/4669/commits/c8bb3d7421918c188e83dbcbb7ff31091712f976 seem to take quite a while - it imports 8 images ? I overtook it from the https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_delete.py#L460 but maybe it strains resources to also include it to chown ?

mtbc:
It's always fine to use whatever kind of group makes the most sense.

Chown does have a sense of what is here called "default exclusion": for instance, in a private group I think that if I give someone my image then it might give them my comments on it too but not my attachments. However, you *will* find that the behavior is suboptimal in some edge cases, perhaps because that it was too tricky to get it right, too tricky to even realize could happen, not clear enough that anybody would care, or not even clear what "right" is. Chown is lightly tested even after I added https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroJava/test/integration/chown/PermissionsTest.java so all manner of issues could have fallen through the cracks.

mtbc:
`importSingleImage` is quite probably overkill in a number of cases: the Java integration testing to which I just linked takes a much faster approach to getting images on to the server than importing fakes:
```java
final Image image = (Image) iUpdate.saveAndReturnObject(mmFactory.createImage()).proxy();
```
This is definitely a general issue to keep watch for: When you are testing rich acquisition data or import logs or whatever then, sure, a full import may well be warranted but, as you have spotted, it *does* come at a price.

ximenesuk:
re: https://github.com/openmicroscopy/openmicroscopy/pull/4669#issuecomment-225090241

I think we have to be careful here not to conflate what I meant be default exclusion and what is done by default according to the graph rules. I am explicitly refering to this method of `delete`:

```
    def default_exclude(self):
        """
        Don't delete these three types of Annotation by default
        """
        return ["TagAnnotation", "TermAnnotation", "FileAnnotation"]
```

This is to mirror the behaviour of Insight which asks before deleting annotations. These classes are added to the exclude classes in the command object. For `chown` and `chgrp` this is an emplty list so there is no default exclusion as understood by the CLI. I.e. the default graph behaviour is not overridden.

mtbc:
Thank you @ximenesuk for the clarification.

pwalczysko:
The commit https://github.com/openmicroscopy/openmicroscopy/pull/4669/commits/9550ccd036aacba7095c2a26e21d77fe3ac77122 is a test suggestion for the ``slash, capital I, nothing or both`` - 4 options of the input of objects to the CLI which we discussed @ximenesuk

components/tools/OmeroPy/test/integration/clitest/test_chown.py
```diff
@@ -57,17 +58,18 @@ def testChownBasicUsageWithId(self, object_type, user_prefix):
         assert obj.id.val == oid
         assert obj.details.owner.id.val == user.id.val
 
-    def testChownBasicUsageWithName(self):
-        oid = self.create_object("Image")
+    @pytest.mark.parametrize("object_form", object_forms)
+    def testChownBasicUsageWithName(self, object_form):
+        oid = self.create_object("Dataset")
```
ximenesuk: As you are using a Dataset I would suggest different parametrerization here:
```
@pytest.mark.parametrize("object_prefix", ["", "/"])
@pytest.mark.parametrize("object_suffix", ["", "I"])
def testChownBasicUsageWithName(self, object_prefix, object_suffix):
...
```
then create that args using the prefic, object name and suffix

pwalczysko:
> in a private group I think that if I give someone my image then it might give them my comments on it too but not my attachments. 
@mtbc : This is not the case. I have checked attachments are also "given" in this sense. I think the behaviour is logical (only when the attachment is used somewhere else as well, it is not "given").

pwalczysko:
I think the PR is ready for review.

ximenesuk:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_chown/

components/tools/OmeroPy/test/integration/clitest/test_chown.py
```diff
@@ -78,7 +84,552 @@ def testChownDifferentGroup(self):
...
+        self.cli.invoke(self.args, strict=True)
+
+        # Check the image has not been transferred
+        assert self.query.find('Image', img.id.val)
+        obj = self.query.get('Image', img.id.val, all_grps)
+        assert obj.id.val == img.id.val
+        assert obj.details.owner.id.val == self.user.id.val
+
+    # Test combinations of include and exclude other than annotations
+    @pytest.fixture()
```
ximenesuk: This fixture is now in both delete and chown tests. It could be sensible to refactor it out into the superclass `CLITest`. Then it is in just one place and easier to maintain.

ximenesuk:
The tests look good with good coverage of the CLI usage of `chown`. Just the one comment on refactoring common code up one class.

components/tools/OmeroPy/test/integration/clitest/test_chown.py
```diff
@@ -78,7 +84,552 @@ def testChownDifferentGroup(self):
...
+    def simpleHierarchy(self):
+        proj = self.make_project()
+        dset = self.make_dataset()
+        img = self.update.saveAndReturnObject(self.new_image())
+        self.link(proj, dset)
+        self.link(dset, img)
+        return proj, dset, img
```
ximenesuk: From here up to the parent class

pwalczysko:
This should fix Colin's comment.

ximenesuk:
Tests all pass, I'm happy with the changes. Good to merge.

PR #4670, opened 2016-05-19
Bugfix: Point figure offset when close to x and or y axis

dominikl:
Tiny bugfix for [Ticket 13232](http://trac.openmicroscopy.org/ome/ticket/13232) / [Trello - Insight drawing bug](https://trello.com/c/YFAxjYtL/60-insight-drawing-bug). 

**Test**: Draw a point ROI close (< 10px) to the upper and/or left edge of the image, check the x,y coordinates via Results tab and make sure the point is displayed in the place where it's supposed to be.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/roi/figures/MeasurePointFigure.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/roi/figures/MeasurePointFigure.java
@@ -411,8 +411,8 @@ public class MeasurePointFigure
     {
         if (shape == null) return;
         UnitPoint c = getMeasurementCentre();
-        AnnotationKeys.CENTREX.set(shape, transformX(c.x.getValue()));
-        AnnotationKeys.CENTREY.set(shape, transformY(c.y.getValue()));
+        AnnotationKeys.CENTREX.set(shape, c.x);
+        AnnotationKeys.CENTREY.set(shape, c.y);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mtbc:
This fixes the issue I was seeing at least! :smiley:

gusferguson:
@dominikl 

Tested using OMERO.insight-5.3.0-m2-480-dce14cf-ice35-b322-mac user-1 read-only-1

There doesn't seem to be accurate mapping of the point ROI coordinates shown on the image and the coordinates shown in the Results tab - even when not near edge - see screenshots.

e.g. ROI ID 2 at the top of the image shows 18.92µm and 0.65µm on image and 3.87µm and blank in Results tab. Looking at the scale bar on the image - the on-image measurement appears accurate.

It also looks as if it is allowing ROIs to be moved off the actual image - left hand side - reports Y coordinate and blank for X.

main

main

dominikl:
Interesting, I always tested it with an image with no pixelssize set, that's why I missed that problem. Can replicate it with a calibrated image; there seems to be something wrong with the point conversion from pixel to um.

dominikl:
**Test 2** (for last commit): Draw a point ROI on a calibrated image. Go to Result tab, check Center(x) and Center(y) values, switch between units display in Pixels and micrometer, make sure the values are converted correctly.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.3.0-m2-480-2eebd69-ice36-b323-mac eel user-3 read-only-1

Behaves as expected for both calibrated - µm - and uncalibrated images - pixels.

Good to merge.

PR #4672, opened 2016-05-19
convert SVG string-based transforms to objects with properties

mtbc:
Introduces a new OMERO model object `AffineTransform` that `Shape.transform` now takes as a value instead of an SVG-style string as previously. For background reading see https://github.com/openmicroscopy/design/issues/33.

Testing notes:

The upgrade script from `OMERO5.3DEV__5` includes SQL unit tests. Feel free to locally add any more you like after the `INSERT INTO transform_tests`. As it is, when one runs the script it should report the seventy passes:
```
 passes | count 
--------+-------
 t      |    70
(1 row)
```

https://trello.com/c/DAoNKA5E/63-synthetic-shape-transform-examples describes how I created an example set of images. You need to do this for a *previous* patch version then upgrade. I can't do it all for you because of https://trello.com/c/InWdSfGf/62-ome-xml-transform-in-bf. However, in the comment below I've made some files available so you can start at the card's step 4. Of course, you can instead test using whatever other data you like.

If you want to check those test files (or your own) then use Insight. Web will have to wait until https://trello.com/c/mlO22Gaw/64-adjust-roi-drawing-in-omero-web is addressed.

If you want to work with your own transforms then with this PR `bin/omero obj` works with them just fine, e.g.,
```
bin/omero obj new AffineTransform a00=0 a10=1 a01=1 a11=0 a02=0 a12=0
bin/omero obj update Point:12345 transform=AffineTransform:123
```

In a later PR I'll address the related pre-upgrade checking of https://trello.com/c/nLwzmAPc/38-preparatory-sql-scripts.


mtbc:
These are some of the files generated by step 3 of https://trello.com/c/DAoNKA5E/63-synthetic-shape-transform-examples: [transform-tests.zip](https://github.com/openmicroscopy/openmicroscopy/files/272707/transform-tests.zip).

mtbc:
again...

ximenesuk:
@pwalczysko, @mtbc had to commit a small change to the CLI here: https://github.com/openmicroscopy/openmicroscopy/pull/4672/commits/d88adc4db0b7f8affe5c9ff797dca25e93d386c0 which allows for object field names to contain, but not start with, numerals. This might be something you could investigate writing a test for. I don't know how many objects have such field names, but there is an example in the description.

dominikl:
#4670 Just fixes a general issue with the Point ROI in Insight, it doesn't fix the tranformations problem. Apparently transformations just don't work in Insight, looks like a JHotDraw issue. If an ROI has an AffineTransform, it is still drawn like it wouldn't have one, but the selection rectangle is transformed (although not correctly either). I don't think you can use Insight to verify that the transformation is applied correctly.

mtbc:
Just actually select the shapes in turn to make sure that the selection marking appears where it should. You need to select anyway because you can't otherwise tell the points apart for sure.

dominikl:
Tested the following workflow:
Started local merge build server, imported image, added some ROIs, added some transforms (translate, rotate, scale) by manually editing `transform` field in the db; also added some ellipses and rotated them (this adds a transform to the shape in the db). Then I upgraded the server to the breaking build. Checked the db, all `transform` strings were converted to the `AffineTransform` relations. All ROIs still worked in Insight like before the update. Good to merge from my point of view 👍

joshmoore:
> Good to merge from my point of view

Let me amend that to "good to come out of breaking" but please see my upcoming email.

pwalczysko:
Went through all the 29 transform ome-tiffs prepared by Mark, and they work as expected.

mtbc:
Thank you! I confess I hadn't had the patience to.

(For those noting that my https://trello.com/c/DAoNKA5E/63-synthetic-shape-transform-examples procedure generates 10 more OME-TIFFs than that, I omitted the ones that are effectively the identity transform, not moving the point at all, and not creating an `AffineTransform` on upgrade: working through all 29 more interesting cases is already rather tedious! The others are still covered in the DB unit tests.)

sbesson:
Any reason not to include the `transform_tests` into the production `5.2__0 -> `5.3__DEV6` upgrade script?

mtbc:
They clutter/complicate it and we already know from the other that they'll pass. They're really just there for the PR review and to make it easier for reviewers to check anything extra.

pwalczysko:
@mtbc : To be able to write a test such as suggested by @ximenesuk in https://github.com/openmicroscopy/openmicroscopy/pull/4672#issuecomment-220339279, it seems to me that the only reasonalbe iteration (except the 3 cases we spoke about, i.e. ``letters only are okay``, ``letters followed by numbers are okay`` and ``leading numbers are not okay``) is to iterate over the  objects which have such field names. It seems the https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_obj.py is not going to help me here, as it deals only with objects which are familiar to me (Project, Dataset...) and which are not interesting in this respect as they do not accept such field naming patterns as described here for ``AffineTransform``.
 - in case this is a reasonable plan, how would I get a list of such ``AffineTransrorm-like`` objects please ?
 - in case this is too overambitious, is that enough to simply test only the three cases discussed on ``AffineTransform`` only ?

mtbc:
It's enough to simply test only the three cases but there are a few other such properties (like sha1); for some diversity, maybe look also at Line which has x1, y1 and x2, y2 and with AffineTransform too that'll be plenty.

mtbc:
Added a couple of integration tests: https://ci.openmicroscopy.org/view/Breaking/job/OMERO-DEV-breaking-integration-java/testngreports/

pwalczysko:
@ximenesuk : see https://github.com/mtbc/openmicroscopy/pull/13 for the tests discussed https://github.com/openmicroscopy/openmicroscopy/pull/4672#issuecomment-220339279

mtbc:
See @pwalczysko's new tests at https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-integration-python/552/testReport/OmeroPy.test.integration.clitest.test_obj/TestObj/.

sbesson:
@chris-allan: does the proposal above make sense to you in the context of your suggestions in https://github.com/openmicroscopy/design/issues/33?

sbesson:
Sorry for the delay, removing the `breaking` label for inclusion in the merge build.

pwalczysko:
Not sure why does it say here on the top "You have been unsubscribed from this thread." ?

mtbc:
No idea but you should have a Subscribe button in the right margin!

pwalczysko:
@mtbc : yes, it says ``Unsubscribe`` so this looks fine. Wonder what that message was.

mtbc:
Non-breaking CI is green, anything further needed here?

chris-allan:
I'm pretty happy with this as is provided that we are collectively happy with the schema style format and I have no reason to suggest that we aren't.

Good to merge from my perspective.

jburel:
Thanks.

PR #4674, opened 2016-05-20
Select roifolders on display

dominikl:
This PR tackles the first two checkpoints on [Trello - ROI Folders Insight follow-up](https://trello.com/c/CONobAmi/57-roi-folders-insight-follow-up):

- Replaces the "Display Folders" dropdown menu with a dialog, where you can choose which folders should be put on display
- In the "Add to Folders" dialog you can now switch if you want to choose from all folders or just from the folders which are on display.

Note: The dialog still use Lists, no TreeTable, so the folders are shown in flat hierarchy with their paths as name, e. g. " Folder A > Sub Folder B > Sub Folder C". I'll investigate for 5.3.1, if/how I could replace the List with a TreeTable, so that you can browse the hierarchy also in the dialog.


mtbc:
```
components/insight/SRC/org/openmicroscopy/shoola/agents/util/SelectionWizard.java:217: error: local variable box is accessed from within inner class; needs to be declared final
                     SelectionWizardDataSource ds = (SelectionWizardDataSource) box.getSelectedItem();
```

components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/ObjectManager.java
```diff
@@ -300,331 +279,106 @@ public void actionPerformed(ActionEvent e) {
...
+                    Map m = (Map) evt.getNewValue();
+                    if (m == null || m.size() != 1)
+                        return;
+                    Set set = m.entrySet();
+                    Entry entry;
+                    Iterator i = set.iterator();
+                    Class type;
+                    while (i.hasNext()) {
+                        entry = (Entry) i.next();
+                        type = (Class) entry.getKey();
```
mtbc: wouldn't need to cast if used generic in declaring `m`

components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/ObjectManager.java
```diff
@@ -300,331 +279,106 @@ public void actionPerformed(ActionEvent e) {
...
+                        if (FolderData.class.getName().equals(type.getName())) {
```
mtbc: should just look at `m.get(FolderData.class)`?

will-moore:
Is there a reason that you can't create a new Folder in the "Displayed Folders" dialog? (you can create a new Folder when you right-click on ROI and "Add to Folder").

The "Display ROI Folders" button should probably not look so much like a drop-down menu now. Just a regular button. Would "Add ROI Folders" but more concise/clearer (just a suggestion)?

I think the TreeTable would be slightly nicer than the list in future, but with the filtering of the list the functionality is not too bad.

pwalczysko:
Bug: The filtering dropdown menu in the ``Add to Folder`` dialog which filters according to ownership of the folders is completely not working. It just shows random bunch of folders, especially selecting ``mine`` and ``others`` after I have switched from ``Displayed folders`` to ``All folders`` in the other dropdown menu.
screen shot 2016-06-09 at 10 05 57

pwalczysko:
Other than the bug https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-224840479 and the comments by Will this PR improves the layout and workflows and gets rid of the "long menu small list bug" as well. ToDos:
   - fix the bug filtering by owner https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-224840479
   - exchange the button for ``Display ROI Folders``, get rid of the arrow, maybe replace with PLUS and go for ``Add...`` as suggested by Will in second point in https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-224840190
   - investigate possibility of having the creation of new folder option available in the ``Display ROI Folders`` dialog, point 1 in https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-224840190
   - as for point 3 of https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-224840190 it seems after the discussion with @will-moore and @dominikl that this is optional and hard to do at the same time, so probably leave for now ?

dominikl:
Yes, it would make sense to enable folder creation in the `Display Folders` dialog; I'll add it to this PR.

pwalczysko:
Small Bug: Filtering term does not persist in one single case of menu swap.
  - get some Folders on display
  - leave some Folders out of display
  - draw a ROI, select it, righticlick, open the ``Add to Folders`` dialog, in the filtering line put in ``h``
  - now go to the dropdown menus (3 of them under the box)
  - Switch between the ``All``, ``Show my``, ``Show others``, all works and the filtering of ``h`` persists
  - Switch between the ``anywhere in name``, ``beginning of name`` works and the filtering persists
  - Switch between ``All folders``, ``Displayed Folders`` -> after you switched to ``All folders``, the filtering DOES NOT persist, all folders are displayed, whether they have ``h`` in their name or not

pwalczysko:
Other than https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-226464679 there are no problems and all works as expected, happy with the new button too. 
The question is whether the new demands coming out of todays UX meeting are going to be implemented in this PR or not.

will-moore:
@pwalczysko I don't think anything from today needs to be added to this PR.
Seems the checked/visibility status of ROIs/Shapes is a bigger and less important change.

pwalczysko:
>  don't think anything from today needs to be added to this PR.

Fine by me. 
Then all that is left here is [the bug](https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-226464679} - unless @dominikl wants to fix the bug in a different PR.

dominikl:
I'll fix the bug and the minor issues raised today (remove the "Folder: " label and the "Show all" checkbox). The visible/enabled/displayed issue seems to be a bit more complex, i. e. separate PR when we decide on what/how to do.

jburel:
I will rename ``visible`` by  ``show``
The visible flag leads to a "view" discussion i.e. we keep the state displayed and that go well beyond the roi and MT. This will have to be investigated. The flag in the model discussed today will not solve the larger issue.
Note that in the current release version, as far as I can tell, web ignores the visible flag i.e. even if it sets to false, shapes will still be displayed.

dominikl:
I renamed the "Visible" column to "Show" and implemented the logic @will-moore suggested this morning, i.e. the checkbox is now only a flag to indicate that an ROI (or ROIs within folders/subfolders) shall be displayed; an ROI itself is only shown if all it's parents are marked as "Show".

dominikl:
I just realised with the last change, the behaviour for ROIs in multiple folders changed a bit. If you now set all parents of such an ROI on "Show" it will be drawn, even it is also in another folder branch which in not marked as "Show". I think that's ok, because if you deliberately want to display the ROIs of a folder, you would assume that you see all the ROIs of this folder, otherwise some ROIs (in an non-"Show" folder) might not be visible and you have no indication that this is the case! Your thoughts?

pwalczysko:
We just agreed with @dominikl that re https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-226709545 this is not expected, and a regression to a behaviour which everybody was happy with (i.e. have a ROI which is in two folders, make one folder hidden, the other folder shown, the ROI is still shown). 
Would suggest not to do the change of Show (Visible) behaviour in this PR, as this is obviously causing regressions.

pwalczysko:
The intended ROI Show behaviour (see https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-226719383) would need more checks in the UI code. The danger of too many checks of what is visible, what should be visible, what is the visibility status of the parent etc. is that we might loose performance as @dominikl indicated.

pwalczysko:
Except the change in Show behaviour which I think we should not accept https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-226719383, the new changes here are as expected.
  - renaming of Visible to Show
  - filtering behaviour fixed
  - "Folder" prefix gone as agreed

dominikl:
Reverted the last commit (which changed the "set visible" behaviour), and only renamed the "Visible" column to "Show".

dominikl:
**Test** : Just check that the display ROI behaviour is like it was before.

components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/IconManager.java
```diff
@@ -195,11 +195,13 @@
     /** The <code>Filter Menu</code> icon. */
     public static final int           FILTER_MENU = 48;
     
+    public static final int           ADD_16 = 49;
```
sbesson: Should this be `48`?

jburel: ??
a new entry is added. Everything seems correct

sbesson: Yes... it's late - don't mind me

dominikl: Well, but I forgot the comment... will add it now.

pwalczysko:
Apparently this is not in the merge build today. Postponing testing.

pwalczysko:
All problems highlighted in https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-226719383 and all intended work here is as expected.
But

1. BUG: Some folders on eel merge (login as user-4) are not displayed when a ROI > Add to folder workflow is executed. I know that these folders exist because they appear in the "+ Add ROI Folder`` dialog. Putting the folders missing from the ``Add to folder`` on display does not help. 

On the screenshots  below note that the fodler ``c`` is missing in the dialog menu.
screen shot 2016-06-21 at 11 02 00

screen shot 2016-06-21 at 11 01 38

pwalczysko:
2. BUG: If I create a ROI, right-click, go ``Add to folder``, select an existing folder and put it into the right-hand side of the ``Add to folder`` dialog. Then create a new folder in the same dialog, put it onto the right-hand side. Expected behaviour is that the existing and the newly created folder are on display and both contain the ROI. But observation is that the newly created folder is not on display. This bug is not intermittent, but does not happen in all situations for all folders.

pwalczysko:
Summary: All good, but two new bugs https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-227396082 and https://github.com/openmicroscopy/openmicroscopy/pull/4674#issuecomment-227396789 detected.

dominikl:
I just realised, bug 1 is not a bug, it's a feature ;-) . In the "Add ROI to Folder" dialog, only Folders are shown, which the user has the permission to `link` to (ie. create a ROI-Folder link). I see now, that this can be very confusing... Is there a way, to better indicate this behaviour somehow?

dominikl:
Note about the last commit: There was an ambiguity with the enabled/disabled state of the "Create Folder" context menu (was always enabled) and the "Create new Folder" panel in the "Add ROI to Folder" and "Display ROI Folders" dialog (only present, if the current active image was annotatable).
Now both are in sync, and only enabled/present if the user is allowed to annotate the current image (although the ROI folders are actually not linked to image).

Following a discussion with @pwalczysko , it turned out that this is not a simple issue and should be discussed more thoroughly. **When should the user actually be allowed to create new folders?** Always, ie also in read-only environments? Only if the active image is annotatable, which is kind of the working context for ROI folders in the measurement tool? Etc. 

That's goes beyond the scope of this PR. Should I open a design issue for it, trello card, trac ticket, ...? /cc @gusferguson

jburel:
I will open a trello card.

gusferguson:
@dominikl @pwalczysko - suggest raise it at the client meeting tomorrow and decide what to do.

pwalczysko:
Understood about the Bug 1. Bug 2. is fixed. Ready to merge.

PR #4675, opened 2016-05-23
override

jburel:
Remove override annotation to stop build failure noticed while looking at https://github.com/openmicroscopy/openmicroscopy/pull/4659
We can decide to remove the methods if unused.

cc @mtbc


mtbc:
#4659 has now passed travis. Thank you!

PR #4676, opened 2016-05-24
7th polygon problem

dominikl:
Fix for [Ticket 12883](http://trac.openmicroscopy.org/ome/ticket/12883)

I noticed that the `MeasureBezierFigure.calculateMeasurements()` is called a couple of times on each point for a polygon you draw. There's a weird add-up effect the more polygons you draw, which ends up calling the method hundreds of thousand times when you've drawn >= 7 polygons, which makes Insight practically unresponsive. I haven't found out the reason for this yet, however preventing to set the `SHOWMEASUREMENT` attribute (which triggers the call), when it's not necessary, stops this odd behavior.

**Test**: Draw a couple of polygons (at least more than 7), check that the measurement tool doesn't slow down with the number of polygons you add.


pwalczysko:
Drawed 50 polygons, also checking the saving is okay. All went fine. Great improvement. Ready to merge.

PR #4677, opened 2016-05-24
Removed (not working) tooltips from GraphPane to save memory

dominikl:
While investigating [QA 17196](https://www.openmicroscopy.org/qa2/qa/feedback/17196) I noticed that  a map of tooltips is generated for the measurement "Graph Pane", which actually doesn't work anyway (there are no tooltips displayed on the graph), but can block for a large ROI for example up to 1 GB of memory. This PR removes that.

**Test**:  Draw/Select an ROI, check that the Graph Pane still works (for large ROIs (eg rectangle nearly completely covering a 4kx4k image) you should also notice a slight performance increase). Also check with latest build (not including this PR) that there are really no tooltips shown on the Graph (maybe I just missed them).

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/GraphPane.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/GraphPane.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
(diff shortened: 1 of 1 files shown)


gusferguson:
@dominikl 

Tested with OMERO.insight-5.3.0-m2-498-645c183-ice36-b325-mac eel user 3 read-only-1 and with version without this PR.

Behaves as expected. 
No tooltips in original graph pane, and no change in this build.
Good to merge.

PR #4682, opened 2016-05-25
remove rank and sort by name regardless of node type (rebased onto develop)

aleksandra-tarkowska:
This is the same as gh-4627 but rebased onto develop.

----

this PR adds ability to sort client tree nodes by type. 

`omero.client.ui.tree.type_order`by default is `tagset,tag,project,dataset,screen,plate,acquisition,image`

to mix types in a tree and sort them by default sorting strategy impleented in client set `omero.client.ui.tree.type_order false`

Note: removing type from the list won't turn off appearance of nodes, it only allows to order them

firefoxscreensnapz026

**Testing: if you update server configuration you must logout from web (even public user) to pick up changes, server config is stored in session and loaded only for the first time when connection is created**

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
@@ -1082,6 +1082,7 @@ $(function() {
             function sortingStrategy(node1, node2) {
                 // sorting strategy
 
+                // If the nodes are experimenters and one of them is the current user.
                 if(node1.type === 'experimenter') {
                     if (node1.data.obj.id === WEBCLIENT.USER.id) {
                         return -1;
@@ -1105,7 +1106,6 @@ $(function() {
             }
             // If the nodes are the same type then just compare lexicographically
             if (node1.type === node2.type && node1.text && node2.text) {
-                // Unless they are experimenters and one of them is the current user.
                 return sortingStrategy(node1, node2);
             // Otherwise explicitly order the type that might be siblings
             } else {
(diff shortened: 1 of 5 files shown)


aleksandra-tarkowska:
--rebased-from #4627

mtbc:
Not sure how `acquisition` fits in here: aren't those always displayed within their plate, with nothing else at their level?

mtbc:
(Also not sure about `image`: a loose image will be inside "Orphaned Images" at the bottom anyway, so does that ordering of that have any effect?)

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
```diff
@@ -1066,47 +1066,47 @@ $(function() {
...
                 return name1 <= name2 ? -1 : 1;
+            }
+
+            // if sorting list is turned off mix object and sort by name
+            if (WEBCLIENT.UI.TREE.type_order.indexOf('false') > -1) {
+                return sortingStrategy(node1, node2);
+            }
+            // If the nodes are the same type then just compare lexicographically
+            if (node1.type === node2.type && node1.text && node2.text) {
+                // Unless they are experimenters and one of them is the current user.
```
mtbc: not sure how the comment fits this code but with luck it all works okay :smiley:

mtbc:
Some minor questions but functionally this works just fine and is a neat feature: good to merge as far as I'm concerned!

aleksandra-tarkowska:
> Not sure how acquisition fits in here: aren't those always displayed within their plate, with nothing else at their level?

this is true if you look at the default tree, but there is nothing stopping people from customizing their tree. The idea was to put all type of node in case. I hope that make sense

mtbc:
That indeed makes sense. :+1:

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
aleksandra-tarkowska: comment moved

will-moore:
Last commit looks right. Good to merge.

jburel:
https://github.com/openmicroscopy/openmicroscopy/pull/4682/commits/4170112c624fa59619b3d1c02c3d4cfb9fd8f33f probably property to handle in the insight too. cc @dominikl 
Commit added to https://trello.com/c/SUO2JeJx/97-client-properties-insight

PR #4683, opened 2016-05-25
make the CLI cleanse operation admin-only

mtbc:
This PR limits cleansing to admins. Test that it otherwise works exactly as before: so, test with both admin and non-admin users. Also test both `bin/omero admin cleanse` and the direct `cleanse.py`.

--rebased-to #4691

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/cleanse.py
+++ b/components/tools/OmeroPy/src/omero/util/cleanse.py
@@ -189,12 +189,9 @@ class Cleanser(object):
             (len(self.cleansed), self.bytes_cleansed)
 
 
-def initial_check(admin_service, config_service):
+def initial_check(config_service, admin_service):
     if admin_service is None:
-        print("No admin service provided! "
-              "Waiting 10 seconds to allow cancellation")
-        from threading import Event
-        Event().wait(10)
+        raise Exception("No admin service provided!")
 
     ctx = admin_service.getEventContext()
     if not ctx.isAdmin:
@@ -224,7 +221,7 @@ def cleanse(data_dir, client, dry_run=False):
     config_service = client.sf.getConfigService()
     shared_resources = client.sf.sharedResources()
 
-    initial_check(admin_service, config_service)
+    initial_check(config_service, admin_service)
 
     try:
         cleanser = ""
@@ -306,7 +303,7 @@ def is_empty_dir(repo, directory, may_delete_dir, to_delete):
 
 def fixpyramids(data_dir, admin_service, query_service,
                 dry_run=False, config_service=None):
-    initial_check(admin_service, config_service)
+    initial_check(config_service, admin_service)
 
     # look for any pyramid files with length 0
     # if there is no matching .*.tmp or .*.pyr_lock file, then
--- a/components/tools/OmeroPy/src/omero/util/cleanse.py
+++ b/components/tools/OmeroPy/src/omero/util/cleanse.py
@@ -189,7 +189,17 @@ class Cleanser(object):
             (len(self.cleansed), self.bytes_cleansed)
 
 
-def initial_check(config_service):
+def initial_check(admin_service, config_service):
+    if admin_service is None:
+        print("No admin service provided! "
+              "Waiting 10 seconds to allow cancellation")
+        from threading import Event
+        Event().wait(10)
+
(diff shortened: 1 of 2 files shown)


sbesson:
Similar to https://github.com/openmicroscopy/openmicroscopy/pull/3158, we should be able to test the `admin_only` decorator logic for the `admin cleanse` subcommand in the Python integration tests.

mtbc:
@pwalczysko: Want a look at doing that?

pwalczysko:
okay, can do.

components/tools/OmeroPy/src/omero/util/cleanse.py
```diff
@@ -189,7 +189,17 @@ def __str__(self):
             (len(self.cleansed), self.bytes_cleansed)
 
 
-def initial_check(config_service):
+def initial_check(admin_service, config_service):
```
joshmoore: I don't think this code is used elsewhere, but the API-safer pythonic way of doing this would be:

```
def initial_check(config_service, admin_service=None):
```

though I'd err on the side of:

```
def initial_check(config_service, admin_service):
    if admin_service is None: raise...
```

pwalczysko:
With this PR, when I try to cleanse as non-admin
```
[pwalczysko@ls31619 ~/Downloads/OMERO.server-5.3.0-m2-521-6639652-ice35-b326]$ bin/omero admin cleanse --dry-run ~/var/testing53
Previously logged in to localhost:4064 as root
Server: [localhost:4064]
Username: [root]testu
Password:
Created session e...2 (testu@localhost:4064). Idle timeout: 10 min. Current group: test
SecurityViolation: Admins only!
[pwalczysko@ls31619 ~/Downloads/OMERO.server-5.3.0-m2-521-6639652-ice35-b326]$ 
```

pwalczysko:
With this PR in, I am getting analogical output as in https://github.com/openmicroscopy/openmicroscopy/pull/4683#issuecomment-221870457 also for a group owner, which is good.

pwalczysko:
With this PR in, created a new admin, and I have the same output as root's for the cleanse command, which is also as expected.

pwalczysko:
With this PR in, when admin really executes the command (i.e. NOT a --dry-run), the output is not informative (not a fault of this PR, but something to look into certainly). Note that 2 files were actually deleted, there is nothing about this in the output (see below);

```
[pwalczysko@ls31619 ~/Downloads/OMERO.server-5.3.0-m2-521-6639652-ice35-b326]$ bin/omero admin cleanse  ~/var/testing53
Using session 5a5bc4c3-3011-425b-8aeb-69d084763e04 (admin2@localhost:4064). Idle timeout: 10 min. Current group: system
/Users/pwalczysko/var/testing53/Pixels does not exist. Skipping...
[pwalczysko@ls31619 ~/Downloads/OMERO.server-5.3.0-m2-521-6639652-ice35-b326]$ 
```

pwalczysko:
All as expected here. I would suggest to keep https://github.com/openmicroscopy/openmicroscopy/pull/4683#issuecomment-221872307 for later improvement and merge if other reviewers are happy. Do you want me to create a card or similar for https://github.com/openmicroscopy/openmicroscopy/pull/4683#issuecomment-221872307 @mtbc ?

mtbc:
@pwalczysko: Yes, card would be great, thank you.

pwalczysko:
@mtbc https://trello.com/c/Hp4KjSRW/683-improve-cleanse-output# created.

ximenesuk:
This looks fine, still works as expected via `omero admin cleanse` and directly as a Python script. Good to merge from me.

sbesson:
Thanks @ximenesuk, any last addition we can think of @joshmoore @jburel ?

joshmoore:
No objections on the data loss front. I did miss a similar API-breakage in `fixpyramids`, i.e. the new argument should have gone to the end of the list. I really doubt anyone has extended the script though, so leave the judgement call there to someone else.

jburel:
better to be non-breaking since we need to rebase it.

sbesson:
If we want to prevent breaking the API, I think we need to use the 2 following:

```
def initial_check(config_service, admin_service=None):
def fixpyramids(data_dir, query_service, dry_run=False, config_service=None, admin_service=None):
```

PR #4684, opened 2016-05-25
PixelsService: use Memoizer.failIfMissing

joshmoore:
To prevent DoS'ing a server, this flag can be set which
requires that all files have a valid cached Memoizer. A
next step will be to allow selectively permitting access
based on some flag.

See https://github.com/openmicroscopy/bioformats/pull/2435

NB: The boolean added to PixelsService should likely be refactored
    into a more general "options" class to reduce future method
    clutter.

### DIFF
--- a/components/romio/src/ome/io/nio/PixelsService.java
+++ b/components/romio/src/ome/io/nio/PixelsService.java
@@ -764,6 +764,7 @@ public class PixelsService extends AbstractFileSystemService
      * @param filePath Non-null.
      * @param series series to use
      * @param store Min/max store to use with the min/max calculator.
+     * @param raise an exception if no cache file is available
      */
     protected BfPixelBuffer createMinMaxBfPixelBuffer(final String filePath,
                                                       final int series,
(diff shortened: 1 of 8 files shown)


components/romio/src/ome/io/nio/PixelsService.java
```diff
@@ -740,11 +767,12 @@ protected void handleMissingPyramid(Pixels pixels,
      */
```
mtbc: add to javadoc (this comment does not block the PR)

components/server/src/ome/security/SecuritySystem.java
```diff
@@ -12,6 +12,8 @@
 // Third-party libraries
 
 // Application-internal dependencies
+import java.util.Map;
```
mtbc: is not application-internal but again, non-blocking to this PR

joshmoore:
Merging for IDR-0.0.5. As mentioned in the description, this may require further API work on the mainline.

jburel:
--rebased-to #5106

PR #4236, opened 2015-10-01
Django18 upgrade

aleksandra-tarkowska:
This PR helps choosing right Django version depends on Python.
 - Python 2.6 Django 1.6
 - Python 2.7 Django 1.8

The aim of this PR is:
 - to deploy Omero.web using both python versions,
 - clarify use of ``request.GET`` and ``request.POST``
 - fix deprecated warnings.

To test:
- check if https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-daily/
  - OMERO-DEV-merge-integration-Python27 -> https://cowfish/integration
  - OMERO-DEV-merge-integration - > https://trout/integration
  are green
- log in to both Omero.web

To test web plugin:
- uninstall django
- bin/omero web start should tells ``Django is not installed!``.

-----------

**Known issues:**
 - request.REQUEST still in webgateway_cache. That will be fixed in another PR if ok

---------------

**To test ui:**

**testing servers: Django 1.6 on eel/merge and Django 1.8 on cowfish/integration
when reporting bugs please indicate which server was used and exact timestamp when error occurred**

- feedback: send comment and error
 - go to https://eel and cowfish/merge or integration/feedback/comment/ and send comment.
 - go to https://eel and cowfish/merge or integration/weberror/error_500/ and submit error.
- login:
 - go to https://eel and cowfish/merge or integration/webclient/login/?url=%2Fmerge%2Fwebclient%2Fuserdata%2F%3Fexperimenter%3D-1, log in, you should be redirected to https://eel and cowfish/merge or integration/webclient/userdata/?experimenter=-1
- webadmin:
 - Reset Password https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/Resetpwd.html
 - add new user / group https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/AdminGroupUserCreate.html (handled by robot test)
 - add new user / group https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/AdminGroupUserEdit.html (handled by robot test)
 - edit my account (handled by robot test)
 - crop avatar
- webclient:
 - check direct link takes to the image webclient/userdata/?show=image-123
 - search using global searching as well as advance searching
 - annotation (test with multiple selected images) https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/Annotate.html
 - create Project, Dataset, Screen
 - create and edit Share https://www.openmicroscopy.org/private/ome-internal/testing_scenarios/Share.html
 - edit channel name
[...]

PR #4687, opened 2016-05-26
thumbnail_default_size_96

will-moore:
See https://trello.com/c/abPmgATV/90-thumbnail-96-by-default

This changes the default size for webgateway thumbnails from 64 to 96, which
can then use the cached thumbnails from OMERO.

To test:
 - go to: ``` /webgateway/render_thumbnail/<imageId>/ ```
 - Should get a thumbnail that is 96 long on longest side.
 - ``` webgateway/render_birds_eye_view/<imageId>/ ``` should continue to give size 96 thumbnail as before (no change).
 - Dataset thumbnails no-longer have size specified (now use default size). Right-click on thumbnail and "Open image in new tab". Url should be ```webclient/render_thumbnail/<imageId>/``` with no '96' size.
 - Do the same for SPW thumbnails.
 - Check test is green at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/350/testReport/OmeroWeb.test.integration.test_plategrid/TestPlateGrid/test_get_plate_grid_metadata/


gusferguson:
@will-moore 

Tested with https://cowfish.openmicroscopy.org/webmerge/webclient/ user-3 read-only-1

Behaves as expected.
Both types of image returned 96 px wide or high for the largest dimension.
Good to merge.

jburel:
Could you add some tests, it will be nice to see that we have the requested size.

chris-allan:
This is actually not a bug in the `views.py` but actually a bug in the way that `ome.plateview.js` uses `gs_json()` to retrieve the plate grid. It does so with a `data` parameter which when passed to `gs_json()` and by extension `jQuery.ajax()` is ignored due to `gs_json` setting the request method to `POST`.

As `ome.plateview.js` is **already** attempting to request thumbnails with a longest size of 96, there are two ways that I can see fixing this without making potentially dangerous hacks to `views.py` that may have unintended consequences:

 * Switch all `gs_json()` calls to `GET`
 * Modify the `ome.plateview.js`, `WeblitzPlateview.load()` function to directly encode the `data` parameters in the URL

Example diff of the latter option:

```
...
-    gs_json(opts.baseurl+'/plate/'+pid+'/'+field+'/', {size:opts.size}, _reset);
+    gs_json(opts.baseurl+'/plate/'+pid+'/'+field+'/?size='+opts.size,
+            null, _reset);
...
```

I expect this has been a bug since very early 5.1.

References:

 * http://api.jquery.com/jquery.ajax/

/cc @emilroz, @stick

chris-allan:
Based on discussions with @aleksandra-tarkowska, #4606 likely also made some changes to this so we should be very careful with this PR, its implications and what we are trying to achieve.

aleksandra-tarkowska:
The source of the bug mentioned above is https://github.com/openmicroscopy/openmicroscopy/pull/4236/files#diff-fc3a666f42c5099bec2593d6f4b4cf2dR1308

will-moore:
@chris-allan All that I'm trying to achieve with this PR is to make ``` /webgateway/render_thumbnail/<imageId>/ ``` use the size 96 thumbnail by default (when no size is specified).
Not sure what it has to do with the ```GET``` vv ```POST``` bug?

chris-allan:
Since this is most obvious for Plate thumbnails I had wrongly assumed that you were fixing the long standing 5.x bug I mentioned coupled with @aleksandra-tarkowska's, #4606. This was made worse by the fact we deployed a series of 5.2.x upgrades last week that ended up generating 1000's of Plate thumbnails because we had a fix for the `gs_json()` on our 5.1.x branches that hadn't been back ported. Thumbnails for Plates and the bird's eye view have long (since 4.4.x) been supposed to load at a longest side of 96.

I'm fairly certain that part of the reason for the forcing of `direct` was the multi-user case and a wide variety of permissions conditions where rendering settings have not been generated. It's been a long time since we messed with this last and I recall some rather ugly fixes that we had to put in place.

We've been through all of this a number of times in somewhat recent history:
 * #1246
 * #1127
 * #746

In keeping with that we should be really careful about this, understand exactly what we want to achieve, and ensure we have thorough test cases for it so that we're not breaking this all again. Especially after @aleksandra-tarkowska has fixed the `gs_json()` issue in #4718.

will-moore:
Seems that CLI import (including plates) generates thumbnails straight off, so they are immediately available to other users in read-only group before owner has viewed them. Works fine with and without specifying size. E.g. ```http://localhost:4080/webgateway/render_thumbnail/56851/``` (defaults to 96) and ```http://localhost:4080/webgateway/render_thumbnail/56851/200/```

chris-allan:
Yep, that's a standard feature of the importer as a whole. What happens if you turn off generating thumbnails?

will-moore:
When imported with ```--skip thumbnails``` then viewed by another user in read-only group, we get ```Failed to render thumbnail``` with or without specifying size.
After the owner goes to ```/webgateway/render_thumbnail/:imageId/``` then the other user can view thumb at same url just fine.
Using a rectangular image: strangely, if the owner *first* goes to ```/webgateway/render_thumbnail/:imageId/200``` to generate a thumbnail (199 x 75) then the other user goes to ```/webgateway/render_thumbnail/:imageId/``` they get a *square* thumbnail (96 x 96) instead of rectangular! After the owner has gone to ```/webgateway/render_thumbnail/:imageId/``` to create a 96 size thumbnail, then the other user sees a regular 96 x 35 thumbnail.

jburel:
Excluding this PR for tomorrow's build. Discussed with @will-moore

will-moore:
@chris-allan Maybe we can discuss this PR in tomorrow's web meeting and either close or define what needs doing?

chris-allan:
So that we don't screw this up again and have the desired default size hardcoded in half a dozen places I'd suggest at a minimum we want to remove the hardcoded setting in `plateGrid_json()` as well as `ome.plateview.js` and rely on the view method to dictate this default. Then we need the test cases to enforce that.

/cc @aleksandra-tarkowska

jburel:
maybe something to consider is to move the default size to the clients properties
something like `omero.client.browser.thumbnail_default_size` so that can be easily changed and consumed by both clients 

currently insight has `org.openmicroscopy.shoola.util.image.geom.Factory.THUMB_DEFAULT_WIDTH =96` and web will have something similar

chris-allan:
Worth some consideration but probably not in scope for this PR, @jburel.

jburel:
I was not thinking for this PR but as a potential followup
I will add a note the properties card as "candidate"

chris-allan:
Do we have any current integration tests that cover thumbnails, @will-moore? It would be great to extend them to check for these default sizes and if not it's probably time to add at least one.

The fact that we were passing `w=80` by default to `render_thumbnail` seems really odd. Any idea where that stems from?

Understood, @jburel. Add a card?

atarkowska:
looks like was changed in https://github.com/will-moore/openmicroscopy/commit/c6d3739a7089219f8567040f235cd1666d19ae5f or even https://github.com/will-moore/openmicroscopy/commit/3427069ad445ee41f6ef7e926049773c643fdca7#diff-e226bce909b5b71ca51f5a26c7db1823R2482, but I am not sure why

jburel:
Added a new checklist to https://trello.com/c/Ib4Rk0bw/25-cleanup-omero-properties

will-moore:
Don't know why size 80 was chosen in https://github.com/will-moore/openmicroscopy/commit/3427069ad445ee41f6ef7e926049773c643fdca7#diff-e226bce909b5b71ca51f5a26c7db1823R2482. I'll look in to adding some web integration thumbnail tests.

gusferguson:
@will-moore 

Tested using http://web-dev-merge.openmicroscopy.org/webclient/ user 3

Thumbnail in centre pane appears as expected.
When thumbnail viewed in new tab URL is as expected:
http://web-dev-merge.openmicroscopy.org/webclient/render_thumbnail/9461/?version=1
Confirmed downloaded thumbnail is 96 px on longest dimension for ordinary, big images and SPW.

Test passed:
OmeroWeb.test.integration.test_plategrid.TestPlateGrid.test_get_plate_grid_metadata (from pytest).

jburel:
Thanks all

PR #4688, opened 2016-05-26
allow customizing lefthand tabs 

aleksandra-tarkowska:
This PR allows customizing names and appearance of left hand tabs.

Tab can be:
- removed
  
  ```
  bin/omero config remove omero.web.ui.left_plugins '{"label": "Shares", "menu": "public"}'
  ```
- added
  
  ```
  bin/omero config append omero.web.ui.left_plugins '{"label": "Shares", "menu": "public"}'
  ```
- updated
  
  ```
  bin/omero config set omero.web.ui.left_plugins '[{"label": "IDR", "menu": "userdata"}, {"label":"Tags", "menu": "usertags"}]'
  ```

Note: removing tab, doesn't invalidate any urls. It is just purely ui manipulation. it is a early step to allow custom apps like https://trello.com/c/s0HD7Vi6/173-tagexport-py

--exclude

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -646,6 +646,20 @@ CUSTOM_SETTINGS_MAPPINGS = {
          ("Add plugins to the right-hand panel. "
           "Plugins are ``['Label', 'include.js', 'div_id']``. "
           "The javascript loads data into ``$('#div_id')``.")],
+    "omero.web.ui.left_plugins":
+        ["LEFT_PLUGINS",
+         ('[["Explore",'
+          ' "userdata",'
+          ' {"viewname": "load_template", "args":["userdata"]}],'
+          ' ["Tags",'
+          ' "usertags",'
+          ' {"viewname": "load_template", "args":["usertags"]}],'
+          ' ["Shares",'
+          ' "public",'
+          ' {"viewname": "load_template", "args":["public"]}]]'),
+         json.loads,
+         ("Add plugins to the left-hand panel. "
+          "Plugins are ``['label', 'id', 'url']``. ")],
     "omero.web.ui.center_plugins":
         ["CENTER_PLUGINS",
          ('['
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -646,20 +646,6 @@ CUSTOM_SETTINGS_MAPPINGS = {
          ("Add plugins to the right-hand panel. "
           "Plugins are ``['Label', 'include.js', 'div_id']``. "
           "The javascript loads data into ``$('#div_id')``.")],
-    "omero.web.ui.left_plugins":
-        ["LEFT_PLUGINS",
-         ('[["Explore",'
-          ' "userdata",'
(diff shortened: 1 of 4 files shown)


will-moore:
One issue that is not a bug with this PR, but a general problem with configs like this:
Tried the "removed" command above:
```
$ omero config remove omero.web.ui.left_plugins '["shares", "public", {"viewname": "load_template", "args":["public"]}]'
["shares", "public", {"viewname": "load_template", "args":["public"]}] is not defined in omero.web.ui.left_plugins
```
It's not an exact match for what's in the config.
And there's not an easy way to see what's in the config, since this gives you nothing.
```
$ omero config get omero.web.ui.left_plugins
$
```
You can 'trick' it to listing what's there...
```
$ omero config append omero.web.ui.left_plugins '"test"'
$ omero config get omero.web.ui.left_plugins
[["Explore", "userdata", {"args": ["userdata"], "viewname": "load_template"}], ["Tags", "usertags", {"args": ["usertags"], "viewname": "load_template"}], ["Shares", "public", {"args": ["public"], "viewname": "load_template"}], "test"]
```
which shows that "Shares" is upper case. Using "Shares" works:
```
$ omero config remove omero.web.ui.left_plugins '["Shares", "public", {"viewname": "load_template", "args":["public"]}]'
$ omero config get omero.web.ui.left_plugins
[["Explore", "userdata", {"args": ["userdata"], "viewname": "load_template"}], ["Tags", "usertags", {"args": ["usertags"], "viewname": "load_template"}], "test"]
```
Now to clean up!
```
$ omero config remove omero.web.ui.left_plugins '"test"'
```
This successfully removes the Shares tab.
Adding tabs, setting tabs and reverting to default all work fine too.

will-moore:
I'm not sure that you really need the second parameter in each plugin: E.g. "userdata" that is converted to html: ``` id="userdata_tab" ```. I don't think this ID is used anywhere else or gives us any functionality?

You may want to catch more different exceptions around
```
url = parse_url(lt[2])
```
E.g. I made a config error ```{"args": [], "viewname": "figure"}``` (instead of ```{"args": [], "viewname": "figure_index"}```) and got:
```
Exception Type: NoReverseMatch at /webclient/userdata/
Exception Value: Reverse for 'figure' with arguments '()' and keyword arguments '{}' not found. 0 pattern(s) tried: []
```
Alternatively you could argue that it's best to have this immediate warning instead of a broken url, in which case you could just remove the ``` try / except Http404 ``` there too?

aleksandra-tarkowska:
@will-moore thank for pointing CLI issue, did you carded that?

Regarding your attempt to add custom view, I think I allowed to much to configure. At the moment left hand panel doesn't benefit from JS tab selector like is nicely done on right hand side. Selected tab based on `menu` that is rendered in template. I move that settings to development and limit to minimum.

aleksandra-tarkowska:
> Alternatively you could argue that it's best to have this immediate warning instead of a broken url, in which case you could just remove the try / except Http404 there too?

we need proper view parser in settings, but not at the moment

joshmoore:
Would anything here be of use with mapr?

aleksandra-tarkowska:
This PR is for turning off left hand panel tabs. For example IDR doesn't need tags and shares

joshmoore:
So you'll re-open when/if you want this? Or should we list this somewhere else?

PR #4690, opened 2016-05-26
Arrow shapes

will-moore:
This adds support for ```markerStart``` and ```markerEnd``` of ```Arrow``` to the display of Lines in web image viewer.

To test, you should draw a bunch of lines, of various widths in Insight (and a few other shapes too), then Save and check that they look correct in web image viewer. 

Then use this script to add  ```markerStart``` and/or ```markerEnd``` to the lines: https://gist.github.com/will-moore/f22695dbc8d2f8d0752cb6fa44431815

Then refresh in web:

screen shot 2016-05-26 at 23 36 35


mtbc:
This certainly doesn't seem to be acting as I'd expect. For instance,
```
$ omero hql "SELECT id, x1, y1, x2, y2, markerStart, markerEnd FROM Line WHERE roi.image.id = 24601"
Using session a1a747ef-783e-4f88-b40c-e896220b968d (user-2@eel.openmicroscopy.org:4064). Idle timeout: 10 min. Current group: private-1
 # | Col1 | Col2 | Col3 | Col4 | Col5 | Col6  | Col7  
---+------+------+------+------+------+-------+-------
 0 | 5808 | 20.0 | 20.0 | 80.0 | 20.0 | None  | None  
 1 | 5802 | 20.0 | 60.0 | 80.0 | 60.0 | None  | Arrow 
 2 | 5804 | 20.0 | 80.0 | 80.0 | 80.0 | Arrow | Arrow 
 3 | 5806 | 20.0 | 40.0 | 80.0 | 40.0 | Arrow | None  
(4 rows)
```
```
$ omero hql "SELECT id, points, markerStart, markerEnd FROM Polyline WHERE roi.image.id = 24601"
Using session a1a747ef-783e-4f88-b40c-e896220b968d (user-2@eel.openmicroscopy.org:4064). Idle timeout: 10 min. Current group: private-1
 # | Col1 | Col2                 | Col3  | Col4  
---+------+----------------------+-------+-------
 0 | 5807 | 20,100 50,120 80,100 | None  | None  
 1 | 5805 | 20,120 50,140 80,120 | Arrow | None  
 2 | 5803 | 20,140 50,160 80,140 | None  | Arrow 
 3 | 5801 | 20,160 50,180 80,160 | Arrow | Arrow 
(4 rows)
```
in Web looks like,
lines-web
but in Insight,
lines-insight
so Web seems to miss the markers but also some of the lines!

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.roidisplay.js
```diff
@@ -118,7 +208,20 @@ $.fn.roi_display = function(options) {
...
+                var arrowPath = getArrowPath(shape);
+
+                newShape = paper.path(arrowPath);
+                newShape.attr({'stroke-width': 0});
+                // We don't want to apply strokeWidth later
+                shape['strokeWidth'] = 0;
+                shape['fillColor'] = shape['strokeColor'];
+                shape['fillAlpha'] = shape['strokeAlpha'];
+              } else {
+                newShape = paper.path("M"+ shape['x1'] +" "+ shape['y1'] +"L"+ shape['x2'] +" "+ shape['y2'] );
```
mtbc, line 224: it's okay not to have a space before the L?

mtbc:
(Note: understood we don't yet expect arrow heads on Polylines.)

will-moore:
@mtbc I think the issue here is that the lines have no colour, so they're black!
You can still select the lines in the table, and you'll see the 'selected' outline of arrows:

screen shot 2016-05-27 at 11 03 48

Not sure how you ended up with no colour, since Insight gives shapes a grey be default.

mtbc:
Thank you, will further experiment accordingly!

will-moore:
Support for arrow heads on PolyLines will need investigation of a different strategy. For single lines, I calculate the outline of the arrow as a single path for Raphael.js. For a PolyLine this would become a nightmare! The solution would be to combine a PolyLine with separate Arrow head shapes, but the current code (E.g. indicating selected object) expects a single Raphael shape.

mtbc:
Made them Insight-gray:
lines-web
and everything else I've tried has worked. Good to merge.

will-moore:
@mtbc - I thought that ```markerStart``` and ```markerEnd``` were enumerations, but it seems they're just strings. I've used "Arrow" to indicate arrow head, but this isn't defined anywhere else?

mtbc:
@will-moore: Only in the XSD as far as I recall.

joshmoore:
> it seems they're just strings. I've used "Arrow" to indicate arrow head, but this isn't defined anywhere else?

Is this a discrepancy we missed?

mtbc:
@joshmoore: As well as restricted in XSD, is also commented in OMERO mapping file. Could even go so far as making it an enum in OMERO if we think it's worth the extra table join.

will-moore:
I looked a bit more into supporting Polylines with arrow heads:
Instead of creating one shape (by calculating the outline of an Arrow as I do now) I would need to 
 - Use separate shapes to represent a PolyLine and Arrow head(s) - add click handlers etc to both.
 - Change the shape selection code to handle multiple shapes selected at once.
 - Parse the points string "M 294 280 L 305 341 L 383 319 L 409 380" to calculate the x & coordinates of arrow heads at each end.
 - Recalculate the points string so that the PolyLine doesn't overlap with the arrow heads (make the line end at the base of the Arrow head)

Given that I've never seen PolyLines with arrow heads (lower priority than for single lines) and that this work will take some time, I will leave this for the time being.

jburel:
I will assume that the arrow will be displayed on the last line composing the polyline
not on each line

will-moore:
I suggest that we don't try to support arrow heads on PolyLine until we are sure that the need is  sufficient to justify the extra work.

jburel:
@will-moore: as suggested in previous comments, we should add arrow on the start/end lines of the polylines to finish this work

will-moore:
Is there a card for supporting Arrows in Insight? cc @dominikl @jburel

mtbc:
https://trello.com/c/sh86TLSl/51-add-markerstart-markerend-for-java-clients looks to be covering it.

jburel:
no activity on this PR for a while
last todo is: `add arrow on the start/end lines of the polylines` cf. https://github.com/openmicroscopy/openmicroscopy/pull/4690#issuecomment-239795001

To be client in synch https://trello.com/c/sh86TLSl/51-add-markerstart-markerend-for-java-clients will have to be moved to 5.3.0 cc @dominikl

will-moore:
@jburel. Given that support of Arrows on Polylines is a large change (will require refactoring a lot of code that is not part of this PR) and the use-case for it is very edge case (I've never seen this needed) I suggest that we don't do it for 5.3.0. If it's not even sure that Arrows support will be in Insight for 5.3.0, this is certainly lower priority for me than other features such as web API and SPW grid view etc.

jburel:
@will-moore: agree that it is low priority considering the rest of the tasks to complete
I will add a note to the card. 
@mtbc are you happy with the current support (excluding no arrow head on polylines)?

mtbc:
Yes, I am fine with this PR being merged and the remaining arrowhead support being pushed to 5.3.x.

jburel:
Note added to https://trello.com/c/sh86TLSl/51-add-markerstart-markerend-for-java-clients

jburel:
Could you add tests to `unit/test_marshal.py`?
Testing the fully `shapeMarshal` will be good.

will-moore:
Added tests

sbesson:
@will-moore: thanks for the additional unit tests, this made me realize we never added support for the new `MarkerEnd/MarkerStart` shape attributes in `omero_marshal`. I'll try to open a PR tomorrow unless you already have staging work for this.

will-moore:
Tests passing: see https://s3.amazonaws.com/archive.travis-ci.org/jobs/168249157/log.txt

```
     [exec] test/unit/test_marshal.py::TestShapeMarshal::test_line_marshal PASSED
     [exec] test/unit/test_marshal.py::TestShapeMarshal::test_arrow_marshal PASSED
```

waxenegger:
Tests are still passing after changes: https://s3.amazonaws.com/archive.travis-ci.org/jobs/168562890/log.txt

will-moore:
As suggested on the Insight PR: https://github.com/openmicroscopy/openmicroscopy/pull/4887#issuecomment-254461470 it is nicer if web and Insight have same sized arrow heads. Last commit should make arrow heads in web the same size as Insight.
To test, create arrows in Insight, save and compare in Web.

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.roidisplay.js
```diff
@@ -139,10 +139,10 @@ $.fn.roi_display = function(options) {
             var arrowPoint1x, arrowPoint1y, arrowPoint2x, arrowPoint2y;
             // if line starts with arrow...
             if (arrowStart) {
-                arrowPoint1x = x1 - (f * Math.sin(lineAngle - 0.4) * headSize);
-                arrowPoint1y = y1 - (f * Math.cos(lineAngle - 0.4) * headSize);
-                arrowPoint2x = x1 - (f * Math.sin(lineAngle + 0.4) * headSize);
-                arrowPoint2y = y1 - (f * Math.cos(lineAngle + 0.4) * headSize);
+                arrowPoint1x = x1 - (f * Math.sin(lineAngle - 0.35) * headSize);
```
jburel: couldn't we have a value n = 0.35 and maybe 2 variables v1 = f \* Math.sin(lineAngle - n) \* headSize and v2 = f \* Math.cos(lineAngle - n) \* headSize

so if we decide to change we do not have to update 8 lines since one line could be easily missed and introduced a bug

jburel:
in order to be synch the same value should be applied to the `Figure_To_Pdf` script
i.e. `arrowPoint1x = x2 + (f * sin(lineAngle - 0.4) * headSize)`

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.roidisplay.js
will-moore: Refactored the 0.35 out of every line

jburel:
Thanks @will-moore 
I have created a card for figure see https://trello.com/c/50DYmJnY/18-arrow-shapes

PR #4692, opened 2016-05-27
Add 5.2.4 history

hflynn:
Will need the autogen docs running before it can be staged at https://www.openmicroscopy.org/site/support/omero5.2-staging/users/history.html

### DIFF
--- a/history.txt
+++ b/history.txt
@@ -1,6 +1,19 @@
 OMERO version history
 =====================
 
+5.2.4 (May 2016)
+----------------
+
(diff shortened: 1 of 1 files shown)


sbesson:
Works for me. I don't know if it's worth adding a comment about the restriction of `admin fixpyramids` to admin-only at the same time.

sbesson:
As we do not have the autogen docs in-place, can this change be opened directly against ome-documentation ?

hflynn:
Ah I forgot about that! Yeah, should be fine, I'll open another PR now

PR #4693, opened 2016-05-27
sanitize user and group name for import paths

mtbc:
Fixes the problem described at https://trello.com/c/Plm8Mrfu/12-usernames-can-contain-filepath-separators:

> Users such as `foo/bar` and `../ManagedRepository` can be created legitimately but they are not able to import any data!

Also adds a new integration test https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ImporterTest/testImportSimpleImageOddlyNamedUser/.

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
@@ -664,7 +664,7 @@ public class ManagedRepositoryI extends PublicRepositoryI
          */
         @SuppressWarnings("unused")  /* used by create() via Method.invoke */
         public String expandUser(String prefix, String suffix) {
-            return prefix + ctx.userName + suffix;
+            return prefix + serverPaths.getPathSanitizer().apply(ctx.userName) + suffix;
         }
 
         /**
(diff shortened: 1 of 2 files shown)


ximenesuk:
Creating a could of crazy users and groups with an import template that uses the group name seems to work okay:

```
$ ls ~/var/omero/ManagedRepository/
a!;b!;c_53	a!b!c_52	user-0_2
$ ls ~/var/omero/ManagedRepository/a\!\;b\!\;c_53/
odd ! group
$ ls ~/var/omero/ManagedRepository/a\!
a!;b!;c_53/ a!b!c_52/
$ ls ~/var/omero/ManagedRepository/a\!\;b\!\;c_53/odd\ \!\ group/
2016-06-08
```

Would this have any adverse effect of a system that, for whatever reason, already had existing imports in crazily named groups?

mtbc:
The only "adverse" effect I can think of is that in the managed repository there might be two directories made for the same user or group, one with old imports in, one with newer, though this would be pretty transparent for normal end users.

mtbc:
Any further review / discussion required?

ximenesuk:
This is fine to merge for me.

PR #4697, opened 2016-06-06
Fixed potential NPE

dominikl:
Just added a `null` check, fix for [QA17213](https://www.openmicroscopy.org/qa2/qa/feedback/17213/).
Test: Just code review should be sufficient (couldn't replicate QA issue).


jburel:
Change makes sense, one of this weird bug with MT

dominikl:
Yes, ideally it shouldn't be `null` in first place, but couldn't find out in which case this could possible happen.

PR #4605, opened 2016-04-23
Copy ome.api.* documentation to omero.api.* (4)

sbesson:
See https://trello.com/c/KLCaZX4X/19-copy-ome-api-documentation-to-omero-api

Follow-up of #4547, #4569 and #4578, this PR copies more of the `ome.api.*` Javadoc documentation into the slice classes.

To review this PR, either build the slice documentation via release-slice2html or use the published CI documentation or OMERO.docs* artifacts. Check the links are all working and the method parameters match.

PR #4578, opened 2016-04-14
Copy ome.api.* documentation to omero.api.* (3)

sbesson:
See https://trello.com/c/KLCaZX4X/19-copy-ome-api-documentation-to-omero-api

Follow-up of #4547 and https://github.com/openmicroscopy/openmicroscopy/pull/4569, this PR copies more of the ome.api.* Javadoc documentation into the slice classes.

To review this PR, either build the slice documentation via release-slice2html or use the published CI documentation or OMERO.docs* artifacts. Check the links are all working and the method parameters match.

PR #4569, opened 2016-04-11
Copy ome.api.* documentation to omero.api.* (2)

sbesson:
See https://trello.com/c/KLCaZX4X/19-copy-ome-api-documentation-to-omero-api

Follow-up of https://github.com/openmicroscopy/openmicroscopy/pull/4547, this PR copies more of the ome.api.*  Javadoc documentation into the slice classes.

To review this PR, either build the slice documentation via `release-slice2html` or use the published CI documentation or `OMERO.docs*` artifacts. Check the links are all working and the method parameters match.

PR #4547, opened 2016-03-29
Copy ome.api.* documentation to omero.api.* (1)

sbesson:
See https://trello.com/c/KLCaZX4X/19-copy-ome-api-documentation-to-omero-api

This PR initializes the copy of the Javadoc documentation from the first set of classes of `ome.api.*` classes into the slice classes.

To review this PR, either build the slice documentation via `slice2html` or use the published CI documentation. Check the links are all working and there is no obvious error.

PR #4702, opened 2016-06-08
 Copy ome.api.* documentation to omero.api.* (5)

sbesson:
See https://trello.com/c/JQ9AcNa2/77-copy-ome-api-documentation-to-omero-api

Follow-up of #4547, #4569, #4578 and #4605, this PR copies the remaining `ome.api.*` Javadoc documentation into the slice `omero.api.*` classes. It also includes a minor fix reported in https://github.com/openmicroscopy/openmicroscopy/pull/4605#discussion_r66062485. This should be the final set of migration PRs associated with this documentation epic. Thanks to @ximenesuk for his infinite patience.

To review this PR, either build the slice documentation via release-slice2html or use the published CI documentation or OMERO.docs* artifacts. Check the links are all working and the method parameters match.

### DIFF
--- a/components/blitz/resources/omero/api/RenderingEngine.ice
+++ b/components/blitz/resources/omero/api/RenderingEngine.ice
@@ -205,7 +205,7 @@ module omero {
                  */
                 ["deprecated: use omero::romio::PlaneDefWithMasks instead"] idempotent void setOverlays(omero::RLong tablesId, omero::RLong imageId, LongIntMap rowColorMap) throws ServerError;
 
-                /** Creates a instance of the rendering engine. */
+                /** Creates an instance of the rendering engine. */
                 idempotent void load() throws ServerError;
 
                 /**
--- a/components/blitz/resources/omero/api/RenderingEngine.ice
+++ b/components/blitz/resources/omero/api/RenderingEngine.ice
@@ -20,57 +20,488 @@ module omero {
     module api {
 
         /**
-         * See <a href="http://downloads.openmicroscopy.org/latest/omero5.2/api/omeis/providers/re/RenderingEngine.html">RenderingEngine.html</a>
+         * Defines a service to render a given pixels set.
+         *
+         * A pixels set is a <i>5D</i> array that stores the pixels data of an
+         * image, that is the pixels intensity values. Every instance of this
+         * service is paired up to a pixels set. Use this service to transform
+         * planes within the pixels set onto an <i>RGB</i> image.
+         *
+         * The RenderingEngine allows to fine-tune the settings that
+         * define the transformation context &#151; that is, a specification
+         * of how raw pixels data is to be transformed into an image that can
+         * be displayed on screen. Those settings are referred to as rendering
+         * settings or display options. After tuning those settings it is
+         * possible to save them to the metadata repository so that they can
+         * be used the next time the pixels set is accessed for rendering; for
+         * example by another RenderingEngine instance. Note that the display
+         * options are specific to the given pixels set and are experimenter
+         * scoped &#151; that is, two different users can specify different
+         * display options for the <i>same</i> pixels set. (A RenderingEngine
+         * instance takes this into account automatically as it is always
+         * bound to a given experimenter.)
+         *
+         * This service is <b>thread-safe</b>.
          **/
         ["ami", "amd"] interface RenderingEngine extends PyramidService
             {
+                /**
+                 * Renders the data selected by <code>def</code> according to
+                 * the current rendering settings.
+                 * The passed argument selects a plane orthogonal to one
+                 * of the <i>X</i>, <i>Y</i>, or <i>Z</i> axes. How many
+                 * wavelengths are rendered and what color model is used
+                 * depends on the current rendering settings.
+                 *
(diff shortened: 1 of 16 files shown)


components/blitz/resources/omero/api/RenderingEngine.ice
```diff
@@ -20,57 +20,488 @@ module omero {
...
+                /**
+                 * Sets the quantization map, one per channel.
+                 *
+                 * @param w  The channel index.
+                 * @param family The mapping family.
+                 * @param coefficient The coefficient identifying a curve in
+                 *                    the family.
+                 * @param noiseReduction Pass <code>true</code> to turn the
+                 *                       noise reduction algorithm on,
+                 *                       <code>false</code> otherwise.
```
ximenesuk: I don't think these need fixing now - as it was in the original docs - but this use of 'otherwise' for a parameter is not quite right. It's fine for a return value which is dependent on something but here it is simply 'The value passed to the noise reduction algorithm' or 'The boolean value...'  This occurs a good number of times in this file where booleans are passed in.

ximenesuk:
This all looks fine, rendering and links look fine. The one comment is more for the future if you want to get this merged now.

jburel:
``Creates a instance of the rendering engine`` => ``an``

jburel:
for ``getColorModel`` and ``setColorModel`` I will add ``@see #getAvailableColorModels``

jburel:
No doc for ``setChannelLookupTable`` and ``getChannelLookupTable``
Plan to be added in another PR?

jburel:
Search.ice: missing double-quote e.g.`` "by*"``. This will create issue with ice 3.6 due to a parsing error.

sbesson:
@jburel:  I do not have any plan to open a PR adding extra content yet. Instead, the assumption was that the documentation will be reviewed in both places as part of future feature work.
Good catch about the double-quote but should our own Ice 3.6 builds have caught this in the first place? I will push commit for the minor fixes.

jburel:
If I remember correctly, the problem showed up when I started the server and the service was used.

jburel:
I should have mentioned: 
``the argument is "cell", one return value might have as its
    * textValue: "cellular" while another has "cellularize".``

jburel:
Looking at the unimplemented LUT methods, now I am wondering if it will not be better to pass an ID instead of a string for the LUT.
this is outside the scope of the PR so the doc can be left blank for now
we can add doc when we start working on LUT

sbesson:
Briefly discussed with @jburel and @ximenesuk. I will push a final commit here that reviews all worrying usages of double-quote in the Slice documentation (with regard to Ice 3.6.2) and replaces them using other types of markup. One follow-up PR here might be to codify all the stylistic decisions made as part of this set of PRs into a guideline document.

sbesson:
Pushed. Ideally the output of 

```
$ git grep '\"' -- components/blitz/resources/omero/api/ | grep -v "\[\"" | grep -v href
```

should be reviewed to make sure we have not left any unsafe double quote.

ximenesuk:
With this PR the output looks fine.
```
colin$ git grep '\"' -- components/blitz/resources/omero/api/ | grep -v "\[\"" | grep -v href
components/blitz/resources/omero/api/IContainer.ice:                 *            Timestamp.valueOf("YYYY-MM-DD hh:mm:ss.ms");
components/blitz/resources/omero/api/IContainer.ice:                 *              Timestamp.valueOf("YYYY-MM-DD hh:mm:ss.ms").
components/blitz/resources/omero/api/IScript.ice:                 *     path = script.path.val\[1:\] # First symbol is a "/"
components/blitz/resources/omero/api/IScript.ice:                 *     path = path.replace("/",".")
components/blitz/resources/omero/api/IScript.ice:                 *     print "Possible import: %s" % path
components/blitz/resources/omero/api/MetadataStore.ice:        const string METADATASTORE = "omero.api.MetadataStore";
```
I assume it builds okay?

Probably not in the scope of this PR but:

https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-build/ICE=3.5,jdk=8_LATEST,label=octopus/javadoc/slice2html/omero/api/ITimeline.html#ITimeline

is a bit of a mess. This wasn't one of the files dealt with in this series of PRs.

sbesson:
I can try and tidy up `ITimeline` in a follow-up PR. From an previous discussion with @jburel, the line that should be the most worrisome is the one with a trailing `"` symbol  i.e.

```
components/blitz/resources/omero/api/IScript.ice:                 *     path = script.path.val\[1:\] # First symbol is a "/"
```

If we want to be on the  safe side, I could drop the double quote (or add a trailing dot).

jburel:
we can clean ITimeline in another PR, since this PR will probably now conflict with the work I am doing on LUT support (https://github.com/jburel/openmicroscopy/tree/lut-repo)
I will remove the ``"``.

sbesson:
@jburel:  see b7f9f49

jburel:
@sbesson: thanks, Looks good. @ximenesuk happy with the status of this PR. 
If no objection, I will merge it

ximenesuk:
If `slice` is happy I'm happy!

PR #4703, opened 2016-06-09
histograms

will-moore:
This adds histograms to Preview panel of web (will also add to image viewer - not ready yet).
Based on designs at https://github.com/openmicroscopy/design/issues/37
Also https://trello.com/c/6gKn8FT2/40-histogram

To test (NB: start with refresh of browser to avoid cached js files)
- Open image in Preview Tab
- Click "Show Histogram"
- Need full test of editing rendering settings (few code changes in this)
- Histogram should show channel being edited (slider/colour-picker/toggle)
- Lines on histogram showing start/end of channel sliders should be shown on histogram and updated when adjusting sliders.
- Also test histogram updates to 'current channel' and start/end lines in sync with slider values during Save, Save To All, Undo / Redo, Copy / Paste, 'Imported' settings, greyscale etc.
- Test similar in full image viewer.

screen shot 2016-06-09 at 23 22 52

NB: I have removed the 'rangewidget' which gives channel start/end inputs their up/down buttons. I think this wasn't useful functionality and looks better with it gone. See before (above) and after (below)

screen shot 2016-06-12 at 23 59 21


joshmoore:
Think a file is missing:

```
https://cowfish.openmicroscopy.org/webmerge/static/webgateway/js/ome.histogram.js Failed to load resource: the server responded with a status of 404 (Not Found)
```

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base_container.html
```diff
@@ -113,6 +115,7 @@
     <script type="text/javascript" src="{% static "3rdparty/aop-1.3.js" %}"></script>
     <script type="text/javascript" src="{% static "webgateway/js/ome.postit.js"|add:url_suffix %}"></script>
     <script type="text/javascript" src="{% static "3rdparty/farbtastic-1.2/farbtastic.js" %}"></script>
+    <script type="text/javascript" src="{% static "webgateway/js/ome.histogram.js" %}"></script>
```
atarkowska: you are missing url_suffix

components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
```diff
@@ -80,6 +80,10 @@
     <script type="text/javascript" src="{% static "3rdparty/raphael-2.1.0/raphael-min.js" %}"></script>
     <script type="text/javascript" src="{% static "3rdparty/raphael-2.1.0/scale.raphael.js" %}"></script>
 
+    <script type="text/javascript" src="{% static '3rdparty/underscore/underscore-1.8.3-min.js' %}"></script>
+    <script type="text/javascript" src="{% static "3rdparty/d3-3.5.17/d3.min.js" %}"></script>
+    <script type="text/javascript" src="{% static "webgateway/js/ome.histogram.js" %}"></script>
```
atarkowska: missing url_suffix

pwalczysko:
Behaviour:
   - turn one channel off
   - then observe that the turned off channel is still displayed on histogram
   - there is a discrepancy now between the preview image and the histogram shown
   - suggest: do get in sync when channel are being turned off with the Preview, if a channel is turned off, show the next channel which is ON in the histogram
   - when the user starts adjusting the bars of a channel which is OFF, put the histogram of that channel on display (this is a discrepancy with the preview image, but directly user-caused one)

pwalczysko:
RFE:
   - the bar in the histogram on the right-hand side tends to completely disappear as I drag the right slider to the right, this happens quite readily even when I am at Min/Max settings 
   - I understand that the Histogram is meant to show just the "relevant" part of the range, but what is it actually showing then when not Min and Max ?

pwalczysko:
RFE: 
   - be on second channel (second channel is selected in Histogram)
   - click "Imported"
   - observe that the "Imported" Rnd settings have been indeed applied, but
   - there is a third channel now displayed in Histogram (the last one on the image)
   - I think the channel selection should not change if I apply "Imported" (or anything else indeed), as there is a clear "deal" that the Histogram channel changes only when I drag the bars or switch the channel on

pwalczysko:
Bug: Unsolicited switching of channels in histogram after first open.
   - In Preview panel, navigate to image https://cowfish.openmicroscopy.org/webmerge/webclient/?show=image-21920
   - this image has two channels, green and yellow
   - start dragging the green right bar to the left, let it go
   - observe that the Histogram became yellow -> not expected, the yelow channel was not touched nor was it switched on or off (both channels are still on as they were when you opened the image)

pwalczysko:
Couple of RFEs and one bug. Stopping testing for today.

gusferguson:
Issues with default channel shown by histogram:
- Opens - shows top channel by default - leave unchanged or change any channel - click Imported - histogram changes to bottom channel. Ditto for the Min/Max and Full Range buttons.

- As user-3 went to image belonging to user-5 which had been changed by owner - adjusted top channel using slider - on release of slider histogram jumped to bottom channel.

- Saved settings as user-3 - clicked on user-5 settings - histogram showed bottom channel by default. Ditto when clicking back onto user-3 settings.

Selecting slider on another channel or clicking into text box does not change histogram - have to change value to trigger - no way to just look at the histogram of a channel as it is without having to change something and then change it back.

If I decrease the minimum value in a channel or increase the maximum value then there is no change to range displayed in the histogram

The font size of the values displayed on the histogram needs to be reduced - too intrusive.

Find it very confusing only having the max value text on the histogram only appearing as the line moves to the left. Only having the min value displayed by default just triggers “what is that number?”. Should have the max value to the left of the max line, not to the right. I think it may also be better to have the text colour the same as the channel and line as this will make context more obvious.

If I increase the max value beyond the default using the text box, then use the slider to reduce it and - without releasing slider - go back to max - the value decreases. Releasing the slider and then repeating this - the max value continues to decrease progressively until the default maximum is reached and then it seems to stablise on that. I think this may have something to do with the slider still taking in values after the cursor on the slider has gone beyond the max point of the slider. In Insight the moment you click on the max slider button the max value returns to the default value.  Min slider button/values behaves the same as max in both. This does occur in 5.2.3 but is intermittent.

gusferguson:
@will-moore @pwalczysko 
Another thought - I would prefer the Show histogram and Greyscale checkboxes to be swapped - makes Histogram more visible - will be used more than the Greyscale checkbox - and puts Greyscale over colour pickers - which is a bit better logic/association.

will-moore:
Should be mostly fixed now.
 - Histogram channel changes when you Toggle a channel (if toggle off, histogram switches to different channel)
 - Also changes channel when you click on the text start/stop inputs or drag slider.
 - Doesn't change channel when you Paste/Reset/Undo/MinMax etc.
 - I removed the numbers on histogram markers as they were redundant and confusing.
 - Fixed the disappearance of histogram markers going off the right of the chart.

Other things I forgot to mention before:
 - Test that histogram updates when you change Z/T sliders. 
 - Also check that histogram can't be turned on for Big images.

gusferguson:
Better without the numbers in the histogram.

Toggling channel works as expected.
Not defaulting to bottom channel when resetting to imported/min/max etc. - fixed.
Greyscale selection works as expected.
Changes with Z and T change works as expected.
Disabled with big images works as expected.

Histogram not changing colour when selecting a slider on another channel - only when moving - is this because selecting a slider without moving it does nor trigger an event?
Does change when clicking into range text box.

Like @petr - finding the behaviour between full range/min/max confusing. Histogram only shows between min and max and lines only appear on histogram when min/max is reached.
It is logical but the behaviour is not transparent.
Not sure if possible/worth trying to make it easier to grasp.

Would still like histogram and greyscale boxes swapped sides.

pwalczysko:
@will-moore : Detected problems with Projection in Full Viewer (the histogram does not change after you projected the image)

pwalczysko:
Bug: The Big image still "Shows" Histogram when Full Viewer is opened.
screen shot 2016-06-17 at 11 42 45

pwalczysko:
> Like @pwalczysko - finding the behaviour between full range/min/max confusing. Histogram only shows between min and max and lines only appear on histogram when min/max is reached.
It is logical but the behaviour is not transparent.
Not sure if possible/worth trying to make it easier to grasp.

Maybe adding a x-axis scale under the Histogram ? Now you have no comparison where actually the "hill" of the histogram is if the bars are missing from the graph.

pwalczysko:
Would it be possible to keep the histogram open when changing images ? (same logic as expanded/collapsed panel in General)

pwalczysko:
2 bugs and 2 RFEs. Much better behaviour, the channel switching is really great now.

will-moore:
@pwalczysko Fixed the max intensity projection update of histogram.
@gusferguson I tried switching the Histogram and Greyscale checkboxes (see below) but I didn't like it. When you use the greyscale button, you are then going to want to toggle the channels (also on the left) not use colour-pickers (since they won't have any effect).

screen shot 2016-06-17 at 15 57 34

screen shot 2016-06-17 at 15 56 32

will-moore:
@pwalczysko Histograms should now stay open when moving between images (but not open for Big images).

gusferguson:
@will-moore 

Tested with https://cowfish.openmicroscopy.org/webmerge/webclient/ user-3

Histograms displayed by default in Preview pane if new images selected when previous one displayed histogram.
Histogram display state persists with opening and closing Edit settings in full-viewer (same image).
display histogram selection greyed out in full viewer with big images.

pwalczysko:
The two bugs https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-226738163 and https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-226738583 are fixed. The RFE with the Histogram being kept open is implemented. 
Actually I quite like the functionality now. Good to merge by me.

jburel:
UI interaction now completed.

Discussed with @chris-allan today, it is preferable to have a method to the rendering engine to return the histogram per channel. So that other viewer can take advantages of the functionalities.
This will avoid having the logic (rendered greyscale image) implemented in both clients. Since this could lead to have different histogram if the logic is not followed
The UI workflow will have to be implemented in insight now. i.e. insight prototype https://github.com/jburel/openmicroscopy/tree/histogram 

@will-moore: will you add the histogram to the full viewer?

I have also concern with projection (the same i have for thumbnail generation) 
This could be problematic for large plane and a good number of z-sections.

jburel:
@will-moore ignored my comment about histogram in full viewer, already there

will-moore:
@jburel I guess having the rendering engine create a histogram would remove the need for ```d3.js``` in the web. I was thinking that it's nice to have the data in hand so we can "interact" with it - E.g. draw the start/end rendering positions on the data, but I guess we could still do that with an image if we know the range and image width etc.
However, the d3 histogram looks pretty nice, so I'd hope we could get one looking just as nice from the rendering engine too? I guess we'd need to specify: ```renderHistogram(theZ, theC, theT, rangeStart, rangeEnd, width, height)``` with the pixel counts (y-axis) auto-scaled to full height and intensity range (x-axis) scaled to full width.
If we are going to support projections, we will need to allow Z AND T projections and specify start/end for each. In web now, we only support Max Intensity of the full range, but Insight supports a chosen range and Mean intensity, so I guess we need all the options that the rendering engine supports.

mtbc:
I wonder where else we can use D3. Could be nice for interactive, responsive, pretty things, especially if we can thereby move (distribute) load from server to clients.

jburel:
Before merging that PR, some rnd changes will be required. Excluding this PR for now so it does not block other PR.

chris-allan:
The discussion on having the rendering engine provide the histogram was, at least as far as I understood it, not to have it draw the histogram but to provide it as an array so that the client could draw it as it saw fit. This at a minimum reduces the amount of data transferred to the client, results in a single implementation of the histogram calculation, and allows for caching of the result.

mtbc:
Ah, providing the array and still using D3 or whatever to draw it sounds far more agreeable, ta.

jburel:
Moving to breaking so it is available when @dominikl starts working on histogram

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#397](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/397/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/397/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/base/base_container.html

will-moore:
Last commit fixes updating of histogram color when using color-picker as noticed in https://github.com/openmicroscopy/openmicroscopy/pull/4767#issuecomment-238906258. Also check behaviour when picking LUTs etc. (histogram should go grey).

pwalczysko:
this works as described https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-239379774. 
RFE: Changing the color on a channel does not switch to a histogram for this particular channel.
  - have 2 channel image
  - channel 1 histogram is on display
  - pick another color for channel 2 -> observe that the color is accepted successfully but
  - the channel 1 histogram is still on display (instead of switching to channel 2 histogram as would be natural after a change has been performed there, compare also the behaviour when I do any other change on channel 2 - the histogram would switch to channel 2 then)

jburel:
while preparing a document describing the synchronization between the components, I noticed  a bug:
* log in as user-3 read-annotate, view data of user-6, select ``5.0-data/leica-original``
* go to preview and select the settings of user-3 (the blue channel is off). Click show histogram.
The histogram of blue channel is displayed, it should be the histogram for the first active channel to be in synch with the rest.

jburel:
Ignoring the way the data is loaded, see https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-240401569

will-moore:
Fixed https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-240401569. To test, turn off first channel before clicking 'show histogram'. First active channel will be shown in histogram.

will-moore:
Last commit should fix the changing of channel when a new colour is picked, as reported in https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-239421766

will-moore:
Will remove the exclude label to allow testing of today's fixes...

pwalczysko:
I still have some sub-variant of the bug with non-active channel histogram https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-240401569. 
Login as user-3 and in RA group go to user-6's data. At first, all is good with the histogram displayed only of first switched-on channel. Both when you are selecting user-3's and user-6's Rnd settings. 

Bug: 
- have the blue channel (first channel) off for both user-3 and user-6's settings, 
- being user-3, select the user-6's settings and switch the blue channel on
- display histogram (it is blue as expected)
- select the thumb of user-3's settings and observe that although the blue channel is off there, the histogram is still blue
- even if you now alternate between user-3's and user-6's settings, (one of which is with blue channel on, the other with blue channel off), you will still have the blue histogram in each case, until
- you switch the blue channel on and off (in this manner you can get rid of the bug)

Fair to say this is a bit edge case.

pwalczysko:
The https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-239421766 is fixed. 

In summary, just one edge case bug, https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-254448654, will leave it up to @will-moore and @jburel whether this is easy enough to fix.

will-moore:
@pwalczysko I found and fixed another bug: when switching between users' different settings the histogram would revert to using the last channel.
But, this is getting tricky to determine when to switch channels, since the events we get from the viewer don't easily distinguish from E.g. change color of single channel (channel-picker) and change color of all channels (change of user settings). So if it's only an edge case I think we should leave it to avoid adding worse bugs.

pwalczysko:
The bug intended to be fixed by the last commit (switching to last channel when switching between different users settings) is fixed. 
I agree, we would probably want to leave https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-254448654 as it is, because it is not very disturbing and very edge case.
Ready to merge fmpov.

will-moore:
Excluding this to allow codomain web PR to be merged....

will-moore:
Codomain PR merged now. Removed 'exclude' flag.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#490](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/490/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/490/consoleText) for more details.
Possible conflicts:
- Upstream changes
  - components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/omero_image.js

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#491](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/491/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/491/consoleText) for more details.
Possible conflicts:
- Upstream changes
  - components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/omero_image.js

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#492](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/492/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/492/consoleText) for more details.
Possible conflicts:
- Upstream changes
  - components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/omero_image.js

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#493](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/493/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/493/consoleText) for more details.
Possible conflicts:
- Upstream changes
  - components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/omero_image.js

will-moore:
Excluding again to avoid merge conflicts. see https://trello.com/c/6gKn8FT2/98-histogram

-- exclude

will-moore:
We now use the Blitz Gateway image.getHistogram(). To test, check that histogram still works OK. Would be nice to compare to histogram from ImageJ or Icy, to validate (as above).

will-moore:
@jburel can you now remove the 'ON_HOLD' label? Thanks.

pwalczysko:
Comparison with Fiji: 
screen shot 2016-11-25 at 10 49 51
screen shot 2016-11-25 at 10 51 15
screen shot 2016-11-25 at 10 57 06
The shapes of histograms are overall the same. The Fiji histograms might look a bit better in first case above, as the low-counts in higher intensity pixels are more pronounced - is it y-axis scaling difference ?

pwalczysko:
A problem:
screen shot 2016-11-25 at 11 03 22
In the situation as in screenshot above (Greyscale checked on a multi-channel image). 
Lets manipulate the bars on blue channel (which is not on) and observe:
  - during manipulation the histogram is blue
  - after I release the mouse button, the histogram reverts to red (red channel is on, blue channel is not on)
  - this behaviour is slightly confusing, although I do not have a better suggestion atm
  - note that when Greyscale is not checked, the behaviour is normal, in the sense that the manipulated channel will remain to be shown as histogram even when this channel is off

pwalczysko:
One minor question on [histogram comparison with Fiji](https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-262932965) and one [minor problem](https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-262933819). 

@will-moore seems to be keen to fix the [minor problem](https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-262933819) - fmpov this is not a blocker.
The [histogram comparison with Fiji](https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-262932965) question is hopefully easily explained by y-axis scaling.

will-moore:
TODO: remove rendering-engine code for histograms of projected images.
Either disable in the UI when projection enabled or temporarily use non-projected histogram.

will-moore:
- Removed rendering-engine based code for histograms of Z-projection images.
 - Fixed 'problem' with greyscale flipping of histogram channel reported above https://github.com/openmicroscopy/openmicroscopy/pull/4703#issuecomment-262933819
 - Added ?bins=100 parameter to the json histogram. Test with E.g. ```webgateway/histogram_json/IMAGEID/channel/0/?bins=10``` or remove bins parameter to get 256 bins.
 - Added histogram test to test this. Should be green.

gusferguson:
@will-moore 

Tested using http://web-dev-merge.openmicroscopy.org/webclient/  user-3

Behaves as expected.
Works with Z-projected images.
Greyscale issue appears to be fixed.
Did not repeat the comparison with Fiji - @pwalczysko to do that if required.

Good to merge from my POV.

components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -1570,7 +1570,11 @@ def load_metadata_preview(request, c_type, c_id, conn=None, share_id=None,
             'c': ",".join(chs),
             'm': r['model'] == 'greyscale' and 'g' or 'c'
             })
+    maxW, maxH = conn.getMaxPlaneSize()
```
jburel: could you for the new variables use ``max_w`` etc.?

components/tools/OmeroWeb/test/integration/test_histogram.py
```diff
@@ -0,0 +1,58 @@
...
+from django.core.urlresolvers import reverse
+
+
+class TestHistogram(IWebTest):
+    """Tests loading of histogram json data."""
+
+    @pytest.mark.parametrize("bins", [None, 10])
+    def test_histogram_bin_count(self, bins):
+        """
+        Test that the we get histogram json of the expected size.
```
jburel: typo

components/tools/OmeroWeb/test/integration/test_histogram.py
```diff
@@ -0,0 +1,58 @@
...
+
+        Default size is 256 bins.
+        """
+        size_x = 125
+        size_y = 125
+        img_id = self.create_test_image(size_x=size_x, size_y=size_y,
+                                        session=self.sf).id.val
```
jburel, line 42: getId().getValue() is preferrable

components/tools/OmeroWeb/omeroweb/webgateway/urls.py
```diff
@@ -315,6 +315,16 @@
...
 """
 
+histogram_json = url(
+    r'^histogram_json/(?P<iid>[0-9]+)/channel/(?P<theC>[0-9]+)/',
+    'webgateway.views.histogram_json',
+    name="histogram_json")
+"""
+Gets a histogram of 256 columns (grey levels) for the chosen
+channel of an image. A single plane is specified by ?theT=1&theZ=2
+or use ?p=intmax for a projection.
```
jburel: this is no longer the case 
no pintmax

components/tools/OmeroWeb/test/integration/test_histogram.py
will-moore: Hmmm - why is ```.getValue()``` preferable?
We use ```.val``` a lot everywhere. E.g.
```
OmeroWeb wmoore$ grep ".val" -r test/integration/ | wc
     727    3229   64263
```

chris-allan: In the Python stack we almost universally prefer the `id.val` style rather than using the accessor methods.

jburel: understood so will need to go over the inconsistency I have noticed in various locations
The style probably evolves through time

waxenegger:
Looks all good, tested on eel as well as locally.

PR #4705, opened 2016-06-11
add list of ProjectionDef to RenderingDef

mtbc:
# What this PR does

Creates a new `ProjectionDef` list for `RenderingDef.projections`.

# Testing this PR

Note that this PR introduces a new DB patch 7.

Ensure that there are no regressions related to rendering settings.

# Related reading

https://trello.com/c/JkRcxqVU/96-z-projection-thumbnails


jburel:
It might be also a good time to remove the codomain context since we have not been using it since we move the code from insight to the server.
No code uses it.
inverse intensity will be handled by the LUT
cc @joshmoore

mtbc:
That's not in the OME-XML schema at all, right, so unaffected by imminent model freeze? Then maybe better to do in a subsequent DB patch because we should first ask mailing lists about uses, as we did for ROI keywords / namespaces?

jburel:
not in the ome schema. 
This can obviously be done in a follow-up PR

jburel:
I am not sure of the ``StartPlane, EndPlane`` terms, probably ``Start/End`` will be enough

mtbc:
"start" and "end" turn out to be a rather confusing pain because they are reserved words in ISO SQL 1999 and elsewhere so, while still possible, they're tricky property values.

jburel:
fair enough, let's add plane in that case. No point to make things more difficult than needed.

jburel:
currently there is no way to create any of the projectionDef object using any of the methods

mtbc:
One can fetch a RenderingDef via iQuery and add ProjectionDefs to it via iUpdate. (The rendering engine API I have no idea about, I'm afraid!)

jburel:
Sorry I was unclear, this was more if you were planning to modify methods to create such object.

mtbc:
Ah, I wasn't, but certainly worth breaking out on the card checklist. There is definitely server-side work to be done about rendering engine / settings stuff to support these projection definitions better, though I'm not quite sure where or what. (The HQL tweak in PixelsImpl seemed a pretty safe guess though!)

jburel:
I have added a test to the RenderingEngineTest class:
you can run that individual test
``./build.py -f components/tools/OmeroJava/build.xml test -DMETHODS=integration.RenderingEngineTest.testRenderingEngineProjectionDef``
branch
https://github.com/jburel/openmicroscopy/tree/ProjectionDef

The test is currently failing

jburel:
I had a look at the DB and no ``projectionDef`` saved.

mtbc:
Excellent, thank you for adding the test. https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-integration-java/lastCompletedBuildtestngreports/integration/RenderingEngineTest/testRenderingEngineProjectionDef/

jburel:
@mtbc: ignore my comment about codomain map. No point to go over the effort of removing it since it will be removed when we rework the rendering engine

jburel:
Ah, I missed the ``add`` method ;-)

mtbc:
Okay, noted. :smiley:

dominikl:
Looks alright to me 👍

dominikl:
I'm just wondering what this `renderingdef_index` field in the database is, can't figure out where it comes from/what it is for.

mtbc:
It arises from the `ordered="true"` in the mapping file: a RenderingDef's ProjectionDef list is kept in order because it may matter in which order multiple projections are performed.

@sbesson, should I remove the `--breaking`?

sbesson:
Sorry coming back late to this. Assuming everyone is okay with the semantics of the new `ProjectionDef`, I would say let's get this into the merge build.

mtbc:
See https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/RenderingEngineTest/testRenderingEngineProjectionDef/.

jburel:
Looks good. Thanks @mtbc

mtbc:
Anything further to be done?

jburel:
Not that I can think of, I will have to see when I start making the change to the Rendering engine if things need adjustment
I am currently looking at LUT in RE

jburel:
if no objections, probably time to have it in cc @sbesson

sbesson:
:+1:

PR #4707, opened 2016-06-14
Use hyphens in id lists

ximenesuk:
# What this PR does

This PR allows the use of elided ID sequences in the arguments to CLI graph commands. As this can be dangerous the default action for such usage is a dry-run which needs to be overridden with a `--force` flag.

# Testing this PR

As this needs access to the CLI a local server build is useful. It's also useful to be able to create objects with sequential IDs so a local running server makes that more manageable.

Create some objects and then try using various graph commands. E.g.
```
$ bin/omero obj new Dataset name=1
Dataset:251
$ bin/omero obj new Dataset name=2
Dataset:252
$ bin/omero obj new Dataset name=3
Dataset:253
$ bin/omero delete Dataset:251-253
$ bin/omero obj get Dataset:251 name # check one of those hasn't been deleted
1
$ bin/omero delete Dataset:251-253 --force
omero.cmd.Delete2 Dataset:251 ok
$ bin/omero obj get Dataset:251 name # check one of those has been deleted
No object found: Dataset:251
```
This should work with sequences of length 1 upwards, `251-251`, `251-252`, `251-351`, etc.

It won't work with negative sequences, `253-251` - should it, this would be trivial?

The output is the same whether `--force` is used or not - this is because of the use of `--dry-run` under the hood. Should the output be more informational (maybe just when `--report` is used)? 

If one of the IDs refers to an object outside of the permissions for the user and command then the whole command should fail. 

# Related reading

See: https://trello.com/c/81tvEXR8/545-cli-hyphen-syntax-for-object-ids

There are no integration tests as yet for this, once some initial comments have been received it might be that @pwalczysko wants to look at adding some basic tests to the various graph commands.

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1517,7 +1517,17 @@ class GraphArg(object):
             assert '+' not in parts[0]
             parts[0] = parts[0].lstrip("/")
             graph = parts[0].split("/")
-            ids = [long(id) for id in parts[1].split(",")]
+            ids = []
+            needsForce = False
+            for id in parts[1].split(","):
+                if "-" in id:
+                    needsForce = True
+                    low, high = map(long, id.split("-"))
+                    if high < low:
+                        raise ValueError("Bad range: %s", arg)
+                    ids.extend(range(low, high+1))
+                else:
+                    ids.append(long(id))
             targetObjects[graph[0]] = ids
             cmd.targetObjects = targetObjects
             if len(graph) > 1:
@@ -1526,7 +1536,7 @@ class GraphArg(object):
                 skiphead.targetObjects = targetObjects
                 skiphead.startFrom = [graph[-1]]
                 cmd = skiphead
-            return cmd
+            return cmd, needsForce
         except:
             raise ValueError("Bad object: %s", arg)
 
@@ -1687,6 +1697,9 @@ class GraphControl(CmdControl):
             "--dry-run", action="store_true",
             help=("Do a dry run of the command, providing a "
                   "report of what would have been done"))
+        parser.add_argument(
+            "--force", action="store_true",
+            help=("Force an action that otherwise defaults to a dry run"))
         self._pre_objects(parser)
         parser.add_argument(
             "obj", nargs="*", type=GraphArg(self.cmd_type()),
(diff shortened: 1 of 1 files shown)


pwalczysko:
```
[pwalczysko@ls31619 ~/Work/openmicroscopy/dist]$ bin/omero delete Dataset:551-553
Using session 5fdb9a07-6c66-4a86-b20a-f9b0a128677b (root@localhost:4064). Idle timeout: 10 min. Current group: system
omero.cmd.Delete2 Dataset:551-553 ok
```
This is as expected. But very confusing, I think that the answer to the question whether more info in this case is needed is a clear yes for me. If I now do not read the docs I will be firmly persuaded that the Datasets were deleted from this output.

joshmoore:
cF. `git clean` requires *either* `-f` *or* `-n` explicitly.

ximenesuk:
@joshmoore re https://github.com/openmicroscopy/openmicroscopy/pull/4707#issuecomment-226202735 is this the time to introduce the same idea here for `GraphControl` commands? Without `--force` or `--dry-run` the command will do nothing and return an informational message?

@pwalczysko supports this idea for commands that might cause deletion, which most of these commands could. This would be another breaking change to the CLI.

pwalczysko:
RFE: There is no mention of this functionality in ``bin/omero delete -h`` output. I understand that this is to be done after the precise behaviour is trashed out but definitely one not to forget.

pwalczysko:
The manual test passes. The dry-run as well as the ``--force`` cases work. Tested also the mixed case, where there was a mixture of ``canDelete`` objects and ``cannotDelete`` objects. The delete did not proceed at all, as expected.

pwalczysko:
> that @pwalczysko wants to look at adding some basic tests to the various graph commands.

Will start the investigation.

ximenesuk:
Re https://github.com/openmicroscopy/openmicroscopy/pull/4707#issuecomment-226206846 I will add basic help examples in a separate PR if this is okay to be merged, that way I can resolve conflicts elsewhere. I will also update any relevant doc pages.

pwalczysko:
Fine by me to merge.

mtbc:
> commands that might cause deletion

For what it's worth, regarding `Graphs.ice`, the commands that inherit from `GraphQuery` instead of `GraphModify2` shouldn't be risky.

ximenesuk:
Is this okay to me merged?

jburel:
@mtbc any objections?

mtbc:
No objections.

PR #4710, opened 2016-06-14
Fix cleanse for null repositories

ximenesuk:
# What this PR does

This attempts to fix the problem in ticket 13251 where an erroneous null repository would break cleanse. This PR uses a modified client call to get a non-null managed repository.

# Testing this PR

I'm not sure how easy it is to test the fail case here since I'm not sure how to create the set-up of a null repository. Any advice @joshmoore ? For now cleanse should work - with slightly improved output for any empty directories. And the integration tests should pass.

Following Josh's comment below here is a way to reproduce the underlying problem. This needs to be locally unless you really want to annoy everyone on the merge build server!! Without this PR:

 - start a  clean local server
 - import a simple image (ADDED in response to @dominikl as the output is different with no data imported)
 - try `omero admin cleanse /OMERO --dry-run` it should not produce an error and the output should end like this:
```
...
Cleansing context: 0 files (0 bytes)
```

 - stop the server
 - use `chmod 000 /OMERO/ManagedRepository/.omero` (this has created an inaccessible ManagedRepository)
 - start the server
 - try `omero admin cleanse /OMERO --dry-run` again, this should cause an exception:
```
...
Cleansing context: 0 files (0 bytes)
...
File ".../dist/lib/python/omero/util/cleanse.py", line 285, in is_empty_dir
    for entry in repo.listFiles(directory):
AttributeError: 'NoneType' object has no attribute 'listFiles'
```
 - stop the server

Now, with this PR included (rebuild `./build.py build-py` should be enough)

 - start the server (the ManagedRepository is stil linaccessible)
 - try `omero admin cleanse /OMERO --dry-run` it should not produce an error but should end cleanly like this:
```
...
Cleansing context: 0 files (0 bytes)
```
 - stop the server
 - use `chmod 755 /OMERO/ManagedRepository/.omero` (recovering the ManagedRepository)
 - start the server
 - try `omero admin cleanse /OMERO --dry-run` again, this should still be okay though the output should now end differently:
```
...
Cleansing context: 0 files (0 bytes)
Removing empty directories from...
 /OMERO/ManagedRepository
```

# Related reading

http://trac.openmicroscopy.org/ome/ticket/13251

### DIFF
--- a/components/tools/OmeroPy/src/omero/clients.py
+++ b/components/tools/OmeroPy/src/omero/clients.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 """
 
-   Copyright 2009 - 2014 Glencoe Software, Inc. All rights reserved.
+   Copyright 2009 - 2016 Glencoe Software, Inc. All rights reserved.
    Use is subject to license terms supplied in LICENSE.txt
 
 """
--- a/components/tools/OmeroPy/src/omero/clients.py
+++ b/components/tools/OmeroPy/src/omero/clients.py
@@ -716,16 +716,20 @@ class BaseClient(object):
         finally:
             self.__lock.release()
 
-    def getManagedRepository(self):
-        repoMap = self.getSession().sharedResources().repositories()
+    def getManagedRepository(self, description=False):
+        repos = self.getSession().sharedResources().repositories()
+        repoMap = zip(repos.proxies, repos.descriptions)
         prx = None
-        for prx in repoMap.proxies:
+        for (prx, desc) in repoMap:
             if not prx:
                 continue
             prx = omero.grid.ManagedRepositoryPrx.checkedCast(prx)
             if prx:
                 break
-        return prx
+        if description:
+            return(prx, desc)
+        else:
+            return prx
 
     def getRouter(self, comm):
         """
--- a/components/tools/OmeroPy/src/omero/util/cleanse.py
+++ b/components/tools/OmeroPy/src/omero/util/cleanse.py
@@ -5,7 +5,7 @@ Reconcile and cleanse where necessary an OMERO data directory of orphaned data.
 """
 
 #
-#  Copyright (c) 2009-2015 University of Dundee. All rights reserved.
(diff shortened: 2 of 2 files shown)


joshmoore:
Placing the ManagedRepository on an NFS share (or somewhere where a lock is not possibly) should do it. Perhaps `chmod 000 /OMERO/ManagedRepository/.omero`?

ximenesuk:
Thanks @joshmoore, the "Perhaps" works a treat. I would be nice to also have a valid managed repository as well as a null one, but the `chmod` workflow reproduces the eror.

dominikl:
Tested the workflow. Without the PR I get the error messages if `.omero` is not accessible, with the PR I don't get the error messages. But I didn't see
```
Cleansing context: 0 files (0 bytes)
Removing empty directories from...
 /OMERO/ManagedRepository
```

The output with the PR (as well as without PR but accessible `.omero`) was:
```
./omero admin cleanse /Users/dlindner/Testing/dev_data --dry-run
Previously logged in to localhost:4064 as root
Server: [localhost:4064]   
Username: [root]
Password:
Created session c15d03d7-1d0a-4bfc-a1bf-6bd8ba3996d9 (root@localhost:4064). Idle timeout: 10 min. Current group: system
/Users/dlindner/Testing/dev_data/Files does not exist. Skipping...
/Users/dlindner/Testing/dev_data/Pixels does not exist. Skipping...
/Users/dlindner/Testing/dev_data/Thumbnails does not exist. Skipping...
```
(I didn't use` /OMERO/ManagedRepository`but `/Users/dlindner/Testing/dev_data` instead)

dominikl:
Tested the workflow, works as described 👍

ximenesuk:
Does this need any further checking before merging?

jburel:
@joshmoore: any more comments before you break off

joshmoore:
Nothing else from me.

ximenesuk:
@jburel could this be merged? I've added a card, https://trello.com/c/ercxHW6Q/698-cli-bin-omero-fs-ls#, to remind me to check the issue seen in my comment above. I'll check while looking at other fs work.

PR #4712, opened 2016-06-15
Fix window position of login screen of Insight on dual screen Linux.

PaulVanSchayck:
# What this PR does

The login screen of Insight and Importer are for our dual screen setups (Ubuntu 14.04) positioned in the middle of the two screens. Meaning that the login-window spans over two screens. This is quite annoying, especially because the window is not movable. 

This PR fixes that by setting setResizable to true. There seem to my (limited) testing no negative side effects from this. Looking back at the history of this code, there seem to be no particular reason why setResizable was set to false. 

# Testing this PR

## required setup

* Linux (Ubuntu 14.04 in our case)
* OMERO.insight

## actions to perform

* Start OMERO.insight
* Watch position on screen of splash window and login window

## expected observations

* With this PR the login window should be positioned in the middle of the primary screen
* Without this PR the login window is positioned in the middle of both screen (spanning both screens)

I'm worried that this change will effect other OS'es. So needs some testing in those.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/ui/login/ScreenLogin.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/ui/login/ScreenLogin.java
@@ -1013,7 +1013,10 @@ public class ScreenLogin
 	{
 		setIconImage(frameIcon);
 		setDefaultCloseOperation(EXIT_ON_CLOSE);
-		setResizable(false);
+		// Setting resizable to false causes the login screen to be positioned in
+		// the center of a dual screen setup, instead of the center of the primary
+		// screen.
+		setResizable(true);
 		setUndecorated(true);
 		toFront();
 	}


jburel:
Another option is to add a flag to the configuration file
see https://github.com/M4I-nanoscopy/openmicroscopy/pull/1

PaulVanSchayck:
Hi Jean-Marie,

Thanks for the suggestion. But do you think it is necessary to make this a configurable option? For me, the login screen does not look different or acts different. I'm still unable to resize it, just as before. The only thing that is fixed is the positioning across two screens.

Does it look or act different for you? Which OS?

Cheers,

Paul

jburel:
I have not noticed any problem on mac.
I need to do some testing on windows.

dominikl:
Tested with Windows 8, Gnome and XFCE, couldn't see a difference (although not tested with dual screen setup!).

pwalczysko:
Tested with Win7 laptop switched to dual screen. No difference in behaviour with respect to latest build. Interestingly, the Windows warning pops up in the screen which the insight is being started from, but the insight login screen is appearing in the main screen. The behaviour is identical with the latest, this means no regressions caused by this PR on windows.
I wonder, shouldn't we have a look also on Linux ? @dominikl ? (see header of this PR)

pwalczysko:
Discussed with @dominikl . As Windows and Mac were tested in the proper real-laptop and screen setup and Linux was tested using VM by @dominikl  and no differences are expected between the OS-es here, this is considered to be tested functionally.

So from functional perspective, this PR is all good.

PaulVanSchayck:
Thanks for all the testing. Just to be clear, no one else has (so far) tested the dual screen setup in Linux? It may be good to do so, to double that this line indeed was the problem.

pwalczysko:
@PaulVanSchayck @mtbc Just kindly performed a realLinux dual screen check (Debian) with and without your PR.
The result: No regression, the behaviour improved with your PR in such a way that the login screen is now displayed centrally in the screen you started insight from. (before it was displayed "off -to-the-left" which was actually quite close to the border between the two screens. 
In summary, this looks like it is doing what is supposed to on Linux. Thanks a lot @PaulVanSchayck and @mtbc

jburel:
Thanks for all. merging

PR #4718, opened 2016-06-17
modify the WeblitzPlateview.load() to tell gs_json to use GET

aleksandra-tarkowska:
# What this PR does

As pointed in https://github.com/openmicroscopy/openmicroscopy/pull/4687#issuecomment-226798195 plate should load data using GET rather then POST 

# Testing this PR

load a plate and check if query string contains size like ``http://.../webgateway/plate/PLATE_ID/INDEX/?size=96&callback=...``

cc @chris-allan @will-moore

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.plateview.js
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.plateview.js
@@ -164,7 +164,7 @@ jQuery._WeblitzPlateview = function (container, options) {
     if (field === undefined) {
       field = 0;
     }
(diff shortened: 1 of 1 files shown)


will-moore:
The testing does exactly what is described.
But do we ever use this?
Seems the default size in ```WeblitzPlateview``` is 96 and we don't override this with any size options in plate.html
```
wpv = $.WeblitzPlateview($('#spw'), {baseurl: '{{ baseurl }}', useParentPrefix: false});
```
It it ever changed from 96?
Regardless, the bug is fixed so this is good to merge.

aleksandra-tarkowska:
@chris-allan you are ok with that?

chris-allan:
We certainly change it around from time to time, @will-moore. Yes, I'm fine with the change,  @aleksandra-tarkowska.

PR #4719, opened 2016-06-18
Lut repo

jburel:
# What this PR does

This PR allows to store, manage Lookuptable. Follow the same approach as the one used for scripts.
I have added methods to retrieve and upload official LUT to the ``IScript``
This can be modified.
The mimetype will probably have to be changed.
More LUT will be added later on.

# Testing this PR
Integrations have been added
# Related reading
https://trello.com/c/9iAe3Sg9/95-lut-support


jburel:
I agree that it won't scale if we start adding more files type. Since I have not yet potential method like "delete"
Without full "scoping" of what is currently required, the larger work will have to be done in step and only focus on the "extension" part.
Need to discuss that.

jburel:
Closing in favour of #4723

PR #4720, opened 2016-06-20
Bugfix: Point ROI drawn offset

dominikl:
# What this PR does

Previously if an ROI was smaller than a minimum size, its bounds were set to this minimum size. In case of a Point ROI this led to an offset by usually 2px into x and y direction when created by a single click. This PR removes this minimum size constraint to fix this problem, as I can't see a particular reason for this constraint.

# Testing this PR

Draw a Point ROI by a single click (no dragging) and make sure it appears where you expect it.

# Related reading

[Trello - BUG: Point ROI drawn offset](https://trello.com/c/rrYdO13a/91-bug-point-roi-drawn-offset)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/ui/drawingtools/creationtools/DrawingObjectCreationTool.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/ui/drawingtools/creationtools/DrawingObjectCreationTool.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.util.ui.drawingtools.creationtools.DrawingObjectCreationTool 
  *
  *------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gusferguson:
@dominikl 

Tested with OMERO.insight-5.3.0-m2-712-d347833-ice35-b346-mac eel user-3

Still seem to be getting a 2 px displacement to right and down, so top right corner of 4 x 4 adjustment box for point ROI is where the cursor was clicked - see screenshot which shows cursor and position of drawn ROI.
On image and results table positions for ROI are in agreement.

main

gusferguson:
PR not in build - waiting for rebuild - will retest.

gusferguson:
@dominikl 

Tested with OMERO.insight-5.3.0-m2-738-9a3a362-ice35-b348-mac eel user-3

Behaves as expected.
Good to merge.

PR #4722, opened 2016-06-20
Ticket 13202

dominikl:
While investigating [Ticket 13202](https://trac.openmicroscopy.org/ome/ticket/13202) I fixed some problems with the ResultsTable of the MeasurementTool. This should also fix the error the user reported, although I couldn't reproduce it (but my suspicion is, it was some kind of unit conversion problem).

On that occassion I also made sure, that no unit up or down conversion ("round to reasonable unit") happens and that the ResultTable does not contain any unit symbols (numbers only),therefore the unit is shown in the headers of the table only (if it is a real unit and not in pixels). That means the numbers in the table should all be of the same dimension so that they are easily comparable and regonized by Excel as numbers when the results are exported as excel file.

**Test**
-  Make sure Coord_List(X) and Coord_List(Y) values are populated for Polygons, in the ResultsTable as well as the exported Excel file.
- Toggle between "calibrated" and "pixel value" display, check that the numbers are reasonable.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
@@ -148,6 +148,11 @@ public class MeasurementTableModel extends AbstractTableModel
                     buffer.append(UIUtilities.formatToDecimal(d));
                     buffer.append(" ");
                 }
+                if (v instanceof Length) {
+                    Length n = (Length) v;
+                    buffer.append(UIUtilities.twoDecimalPlaces(n.getValue()));
+                    buffer.append(" ");
+                }
             }
             if (total > 0) {
                 buffer.append("= "+UIUtilities.formatToDecimal(total));
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
@@ -164,16 +164,15 @@ public class MeasurementTableModel extends AbstractTableModel
                     }
                 }
                 return buffer.toString();
-            } else {
-                if (value instanceof Number) {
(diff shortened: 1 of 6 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/MeasurementTableModel.java
```diff
@@ -188,12 +164,54 @@ public Object getValueAt(int row, int col)
                     }
                 }
                 return buffer.toString();
-            } else if (value instanceof Number) {
-                double d = ((Number) value).doubleValue();
-                return UIUtilities.twoDecimalPlacesAsNumber(d);
+            } else {
+                if (value instanceof Number) {
```
mtbc: shouldn't this if and the next just continue the enclosing "else if"...? (similar below)

gusferguson:
@dominikl 

Tested with OMERO.insight-5.3.0-m2-743-04d2e4e-ice36-b349-win eel user 3 Win 8.1

Behaves as expected.
Results table and Excel export agree with both calibrated and pixel selected.
Good to merge.

microsoft excel

dominikl:
Thanks @gusferguson . I'll just correct the odd `else` `if` grouping remarked by @mtbc . I don't think a retesting is necessary therefore.

mtbc:
Certainly looks better, thank you.

dominikl:
Now also removed the unnecessary 'else' again.

jburel:
@dominikl: you probably force pushed, @mtbc @gusferguson  happy with the status of this PR?

mtbc:
Sufficiently happy at least. :smiley: :+1:

dominikl:
Yes sorry, I think I force pushed to re-trigger travis build. Only the last commit was added after @gusferguson review of the PR, but it shouldn't affect functionality, so no re-review necessary in my opinion.

jburel:
Thanks merging in that case.

gusferguson:
Happy.

PR #4724, opened 2016-06-21
remove old jstree lib

aleksandra-tarkowska:
# What this PR does

This PR clean up 3rdparty common libraries. 

# Testing this PR

We are using jstree 3.0.8 and there is no reason to keep old one

cc @will-moore


will-moore:
Looks good. Can't see that we use any of these since jsTree 3.
Good to merge.
Thanks!

PR #4725, opened 2016-06-22
ROI/Folders visibility

dominikl:
# What this PR does
Proposal follow up to #4674 .
This PR changes the logic of the ROI visibility. Without this PR the *Show* state is propagated down in the folder hierarchy when clicking on the *Show* checkbox, ie. all subsequent folders and ROIs will be set to the same state. With this PR that's not the case any more, only the *Show* state of the folder (or ROI) which is clicked on, will be changed. The conditions for an ROI to be displayed or not will stay the same: It will be displayed if any branch the ROI is part of is *shown* (a branch is *shown* when all its nodes are *shown*). This should match the behaviour @will-moore was suggesting in the recent ROI/Folders meeting.

# Testing this PR
Create some nested folders and ROIs (also ROIs which are part of multiple folders) and play around with the *Show* checkbox. Check if the behaviour matches the expected behaviour.


components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/roitable/ROINode.java
```diff
@@ -517,12 +521,69 @@ public void setValueAt(Object value, int column)
...
+     * @param show
+     *            The show flag
+     */
+    void setShow(boolean show) {
+        this.show = show;
+        if (isShapeNode()) {
+            ROINode roinode = (ROINode) getPath().getParentPath()
+                    .getLastPathComponent();
+            roinode.show = show;
+        }
```
mtbc: these aren't elses?

dominikl:
Implemented the changes which were discussed in todays client meeting:
- If you change the *Show* status of a ROI it's propagated down to it's Shapes (and vice versa)
- Added new folder icon; provided by @gusferguson 
- Added folder icon for folders which are owned by another user; provided by @gusferguson 
- Setting the *Show* status of a folder will en-/disable the *Show* checkbox of all  it's children

Also noticed/fixed another bug:
- Now it's not possible any more to select not visible ROIs

will-moore:
Show looks good now. Working as I'd expect.
Non-related minor issue I noticed while testing: I can't right-click on a Folder and "Add to Folder"? Instead I had to "Create a Folder" but that added the new Folder as a child of the selected folder instead of parent. Then I had to drag it out (orphaned) before I could make the selected folder into a child of the new folder as I'd wanted. Is there a reason you can't "Add to Folder" for a folder, as you an for an ROI?

Regardless, the Show functionality is working great and good to merge.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.3.0-m2-756-e8fec79-ice36-b351-mac eel user-3

Show stars works as expected.
New icons look OK to me.
Cannot select invisible ROI or shape.

WRT on "other user" icon: - should these not be shown in the Display/Add ROI Folder dialog boxes as well - so you can see if you are selecting a Folder owned by another user?

An issue that may not be related to this PR:
- add 1 existing and one new folder to Management display
- select another ROI - click Add ROI Folders
- previously added Folders appear in right-hand pane.
This implies to me that the ROI I have selected will be added to those Folders as well as any others I now add to the RH panel. This is not what happens.
I suspect this is a hang-over from when this dialog was being used only to add ROI Folders to the display rather than add ROI Folders to the display and any selected ROIs to a those new Folders. Perhaps borne out by the header of the window "Displayed Folders".
I think these already added folders should just not be available in the left panel and not visible in the right. If they absolutely have to remain in the right panel then they should be greyed out as they are not available for any action on them - i.e. can't add selected ROI to them in this dialog.

Happy for this to be deferred to another PR so this one can be closed and merged when the target issues are addressed.

main

dominikl:
Thanks @will-moore . Yes, instead of the "Add to Folder" for folder, there's the "Move folder" option, using the "Add to Folder" dialog would be difficult and I guess a bit confusing, because a folder can always only be part of one other folder. If you create a folder when another folder is selected the new folder will be added as child of the selected one, that was intentional. Do you think it's too confusing/unexpected?
@gusferguson I'll try to replicate and fix the bug. Yes, definitely agree that the "owned by other user" icons should be used in the dialogs as well. Will fix that.

dominikl:
@gusferguson The bug you mentioned, is that just a confusion about the name "Add ROI Folders"? Because "Add ROI Folders" only adds or removes folders to the table, it has nothing to do with what is selected there, it doesn't modify any folders/rois just displays (or doesnt' display) them in the table.

gusferguson:
@dominikl 

I can't replicate the selected ROI going into a new folder via that dialogue now - maybe I was just mistaken.
I think the current use of "Display" in the dialog is a bit confusing.
I suggest the title of the window should be: "Add ROI Folders to Display"
And text in the top bar: "Add ROI Folders to the Display. Select existing folders in the left pane, or create new folders to add to the display. Remove folders from the right hand pane to remove them from the display."
Then the Right hand pane header should be "Display:" rather than "Selected:"

cc: @pwalczysko

gusferguson:
@dominikl 

Sorry to nitpick, but a couple of issues come up with my draft Help screenshot making:
- please can you to change the Filter by drop-down menu options to "Anywhere in name" and "Start of name" to keep the capitalisation consistent with the drop-box below.
- when the dialog opens the text label for the new folder boxes at the bottom reads: "Add a new folder and select it immediately:" However when I filter on the left pane, the "folder" changes to "tag" - see screenshots below. The refresh must cause a reversion to an old label.

I thought I was going mad and kept on redoing screenshots until I worked out what was happening!

displayed folders

displayed folders

gusferguson:
@dominikl 

- New icons for folder (without stack) and owned by other are showing in the dialogue box - look good
- Tag/Folder label issue fixed
- capitalisation of drop-down menus now consistent.

Good to merge.

jburel:
we will need a corresponding web PR to replace ``Visibility`` by ``Show``

will-moore:
Web renaming of ```Visibility``` to ```Show``` is at https://github.com/openmicroscopy/openmicroscopy/pull/4745

PR #4726, opened 2016-06-23
give LoginCredentials a Session ID constructor

mtbc:
# What this PR does

It is not obvious from our code how to use `LoginCredentials` to connect to the server using a session ID instead of a username and password. This PR adds a constructor to fit the session ID pattern and an integration test that uses it.

# Testing this PR

Check that the code looks clear and that https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.gateway/GatewayUsageTest/ passes.

# Related reading

https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8071

### DIFF
--- a/components/blitz/src/omero/gateway/LoginCredentials.java
+++ b/components/blitz/src/omero/gateway/LoginCredentials.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2015 University of Dundee. All rights reserved.
+ *  Copyright (C) 2015-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -18,6 +18,7 @@
  *
  *------------------------------------------------------------------------------
  */
+
 package omero.gateway;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


components/blitz/src/omero/gateway/LoginCredentials.java
```diff
@@ -118,6 +119,36 @@ public LoginCredentials(String username, String password, String host,
...
+     * Note that the given session ID is passed via the user name property
+     * and is subsequently available from {@link UserCredentials#getUsername()}.
+     *
+     * @param sessionId
+     *            The session UUID
+     * @param host
+     *            The server hostname
+     */
+    public LoginCredentials(String sessionId, String host) {
+        this(sessionId, null, host, omero.constants.GLACIER2PORT.value);
```
aleksandra-tarkowska: Does it mean I cannot connect to custom port?

mtbc: Use the other constructor below.

dominikl:
That's a good idea to add an explicit constructor for this. I just thought about adding a code example to the developer's doc ( https://trello.com/c/DW5EwzQq/14-documentation ), but having a constructor is certainly better.

components/tools/OmeroJava/test/integration/gateway/GatewayUsageTest.java
```diff
@@ -56,6 +60,24 @@ public void testLoginWithCredentials()
...
     }
 
+    /**
+     * Check logging in using a session ID.
+     * @throws DSOutOfServiceException unexpected
+     */
+    @Test
+    public void testLoginWithCredentialsSessionId()
+            throws DSOutOfServiceException {
+        final omero.client client =  new omero.client();
```
sbesson: Looking at this test, this raises a general design question to me. From  the point of view of a consumer of this libary, is there any benefit into migrating this logic to the gateway itself. For instance,  define a  simpler

```
 public LoginCredentials(String sessionId)
```

which would built an `omero.client()` using the ice.config set by envvar and pass the requires properties to  the constructor? Or should this logic be migrated away from `LoginCredentials` by design?

joshmoore: 2 related cents are that `Ctor(stringA, stringB, ...)` is easy to confuse.

mtbc: I was just aiming for something straightforward and simple but maybe it's not sufficiently futureproof. Probably better then if I close this PR and create some larger scoping card and have @dominikl document the issue in the meantime?

jburel:
@mtbc: probably better to create an issue on design or a card for scoping and close the PR in the meantime

mtbc:
Closing in favor of https://trello.com/c/JF3VjeWk/694-refactor-logincredentials.

PR #4727, opened 2016-06-23
Deprecate OmeroPopo module

sbesson:
See https://trello.com/c/0dkqUU9f/51-deprecate-omeropopo-py

The `omero.util.OmeroPopo` module is no longer actively used in our repositories. As suggested in the card, this PR deprecates the various classes as well as the two isolated methods `omero.util.script_utils.registerNamespace()` and `omero.util.script_utils.findROIByImage()` which were consuming it.

To test this PR, check all tests remain green and that there is no other place in this code or elsewhere (scripts) where these methods are used.


jburel:
If I remember correctly the idea behind the popo is what omero-marshall is.
So we may look at using omero-marshall instead of popo, This is outside the scope of this PR

PR #4709, opened 2016-06-14
Initial implementation of "group shares"

joshmoore:
Rather than placing multiple elements from a single-group in
a share, this allows adding an entire group to a share creating
a proper read-only environment. Primary concern here is under
what conditions should this be **disallowed** since, for example,
it allows illegally opening up a private group.

PR #4730, opened 2016-06-25
Initial implementation of "group shares" (rebased onto develop)

aleksandra-tarkowska:
This is the same as gh-4709 but rebased onto develop.

----

Rather than placing multiple elements from a single-group in
a share, this allows adding an entire group to a share creating
a proper read-only environment. Primary concern here is under
what conditions should this be **disallowed** since, for example,
it allows illegally opening up a private group.


aleksandra-tarkowska:
--rebased-from #4709

PR #4734, opened 2016-06-29
Refactor PojoMapper

dominikl:
# What this PR does

Refactoring the PojoMapper class: Replaces various similar methods with one method `convertToDataObjects` (previous methods are marked as deprecated). Makes more use of generics. Use general `Collection` interface instead of returning `Set` and `List`. Marks unused methods as deprecated.

# Testing this PR

Nothing particular to test. PojoMapper class is used throughout Insight. Just leave the PR in the build for a while and see if something breaks? I know that's generally not a good strategy for testing... Your thoughts @jburel , @pwalczysko ?

There was one piece of code (the Importer component) where the program logic depended on wether the Collection was a Set or List, which was quite a bad design (fixed that, see last commit bd966ac ). I'm a bit worried that there might be other similar cases. They will compile fine, but either crash Insight or change the behaviour.

### DIFF
--- a/components/blitz/src/omero/gateway/util/PojoMapper.java
+++ b/components/blitz/src/omero/gateway/util/PojoMapper.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
@@ -25,6 +25,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -229,6 +230,8 @@ public class PojoMapper
      * @return A set of {@link DataObject}s.
      * @throws IllegalArgumentException If the set is <code>null</code>, doesn't
      * contain {@link IObject} or if the type {@link IObject} is unknown.
+     * 
+     * @deprecated Use {@link #convertToDataObjects(Collection)} instead
      */
     public static Set asDataObjects(Collection objects)
     {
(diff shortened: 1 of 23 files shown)


components/blitz/src/omero/gateway/util/PojoMapper.java
```diff
@@ -252,6 +255,30 @@ public static Set asDataObjects(Collection objects)
      * @throws IllegalArgumentException If the set is <code>null</code>, doesn't
      * contain {@link IObject} or if the type {@link IObject} is unknown.
      */
+    public static <T extends DataObject> Collection<T> convertToDataObjects(Collection objects)
+    {
+        if (objects == null) return Collections.EMPTY_LIST;
```
mtbc: no Eclipse warning here?

components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/ExperimenterDataLoader.java
```diff
@@ -240,7 +240,7 @@ public void handleResult(Object result)
     		Class klass = parent.getUserObject().getClass();
     		long id = parent.getUserObjectId();
     		if (TagAnnotationData.class.equals(klass)) {
-    			viewer.setLeaves((Set) result, parent, expNode);
+    			viewer.setLeaves((Collection) result, parent, expNode);
```
mtbc: There's a few bare collections around. I assume that there's nothing that makes sense in the `<>`, even `?`? (I can imagine issues might just propagate too far for now.)

jburel: @mtbc: that will propagate...

mtbc: No problem, fine to limit creep!

jburel:
I have not noticed any obvious problem.
We probably need to way a bit longer

jburel:
@pwalczysko: did you notice any issues?

pwalczysko:
No issues noticed during July.

jburel:
This PR has been opened for over a month now. no problems noticed. Merging

PR #4737, opened 2016-06-30
Tree type order

dominikl:
Implements #4682 for Insight.

**Test**: Also see above mentioned PR.
- Start Insight, check the tree in left hand side panel. Make sure, Projects are listed before Datasets, Screens are listed before Plates, TagSets are listed before Tags.
- On the server set the `omero.client.ui.tree.type_order` property *to something else* than `tagset,tag,project,dataset,screen,plate,acquisition,image` (that's the default).
- Restart the server
- Now check again and make sure the different data types are listed in the order you specified

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/browser/BrowserUI.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/browser/BrowserUI.java
@@ -1396,9 +1396,15 @@ class BrowserUI
 			}
 			else if (uo instanceof TagAnnotationData) {
 				if (TagAnnotationData.INSIGHT_TAGSET_NS.equals(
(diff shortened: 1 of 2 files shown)


pwalczysko:
Bug: Tag and Tagsets ordering does not work in Insight in the non-default case.
In case
``omero.client.ui.tree.type_order=tag,tagset,dataset,project,image,screen,acquisition,plate``
The ordering of Tagsets and Tags should be as it is in Web (see lower screenshot).
Instead, the Tagsets and Tags seem to be ordered alphabetically in Insight.
screen shot 2016-07-01 at 12 29 56
screen shot 2016-07-01 at 12 30 24

pwalczysko:
The bug https://github.com/openmicroscopy/openmicroscopy/pull/4737#issuecomment-229924802 is fixed.
Ready to merge.

PR #4740, opened 2016-07-01
Change Pascal case to PASCAL

rleigh-dundee:
# What this PR does

- Search and replace of `Pascal` to `PASCAL`
- Add PASCAL workaround for windows by adding a private header sourced by the generated Ice sources during compilation.
- May need additional workarounds at a higher level where the units headers get included, as done for the other C++ components.  But there are no such users at this time.

Due to the nature of the Ice code generation, it's not possible to put the workaround in the public header itself as for the other C++ code; we don't have control over this.  But `slice2cpp` does make provision for a single header to be added to the generated source, which we use here.

# Testing this PR

Check builds remain green.  The OMERO C++ jobs should be unchanged if this is all working.

# Related reading

https://trello.com/c/4NOpYKza/39-header-problems-on-windows
https://github.com/openmicroscopy/bioformats/pull/2474/files

Windows defines `PASCAL` in the system headers, which interferes with `PASCAL` as a unit enum.  This PR works around that.  Previously we renamed it to `Pascal` as a workaround, but if this fix works then the nasty hack is no longer necessary.

### DIFF
--- a/components/blitz/resources/omero/model/Units.ice
+++ b/components/blitz/resources/omero/model/Units.ice
@@ -148,7 +148,7 @@ module omero {
                 KILOPASCAL,
                 HECTOPASCAL,
                 DECAPASCAL,
-                Pascal,
+                PASCAL,
                 DECIPASCAL,
                 CENTIPASCAL,
                 MILLIPASCAL,
--- a/components/blitz/resources/templates/combined.vm
+++ b/components/blitz/resources/templates/combined.vm
@@ -152,6 +152,7 @@ DECLARATION BLOCK:
 [$ice]
 #end
 
+[$cpp] \#include <omero/internal/fixes.h>
 #foreach( $property in $type.propertyClosure )
 #if($property.type.startsWith("ome"))
 #if($property.isLink)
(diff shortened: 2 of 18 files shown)


rleigh-dundee:
This is mostly done.  It needs a final tweak to do this for the implementations in addition to the interfaces.

sbesson:
Code changes here read well, unify the unit semantics and are in agreement with the OME Files C++ proposal in https://github.com/openmicroscopy/bioformats/pull/2474. With the latest commits both unit tests and the [cpp build](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-cpp-win/) are now passing.

The only remaining question is whether we are all fine with the breaking change (which should be included in the 5.3.0 documentation /cc @hflynn). To some extent, the only place in the model where we use pressure units is the `ImagingEnvironment.AirPressure` (only 1 row with a non-null value in the nighthsade DB) so the impact here should be pretty minimal from my point of view.

mtbc:
Ah, does this require a fragment of DML too to switch any values in that column?

rleigh-dundee:
@mtbc If it's storing the enum name then yes, but if it's storing the symbolic name then no.  I'm not familiar with the OMERO units, so I'm not certain what needs doing here.

mtbc:
The OMERO DB usually has physical pixel sizes in `MICROMETER` and the time increments in `SECOND` if that's a useful clue?

joshmoore:
Pretty sure it's stored as `Pa`.

sbesson:
```
omero=> select * from ImagingEnvironment i where airpressure is not null;
  id   | owner_id | group_id | creation_id | update_id | permissions | external_
id | version | temperature | airpressure | humidity | co2percent | airpressureun
it | temperatureunit 
-------+----------+----------+-------------+-----------+-------------+----------
---+---------+-------------+-------------+----------+------------+--------------
---+-----------------
 20315 |     3302 |        2 |    38605870 |  38605870 |        -120 |          
   |         |           0 |           0 |      0.5 |        0.5 | mbar         
   | °C
(1 row)
````

So I think the symbolic unit is used which should mean no DB change required.

mtbc:
Gah, sorry, was defeated by whatever magic is done to HQL queries.

PR #4743, opened 2016-07-04
populate_metadata many rows

will-moore:
# What this PR does

Fix bug with populate_metadata.py reported at http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2016-July/003701.html where the script fails if the plate has more than 26 rows.

# Testing this PR

 - Use the Dataset_To_Plate.py script to create a Plate with more than 26 rows.
 - Create a csv file that names Wells using A->Z then AA->AZ etc. 
E.g.
```
Well,	Well Type,	TestValue
A01,	Treatment,	1
A02,	Control,	2
AA01,	Treatment,	3
AB02,	Treatment,	4
```
 - Attach CSV to Plate and run Populate_Metadata as described http://help.openmicroscopy.org/scripts.html#metadata
 - The Script should complete without error. Selecting Wells specified in the csv file should show the corresponding values in the Tables section, with the Well Name of that well.

screen shot 2016-07-04 at 13 00 27

cc @chris-allan

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/populate_metadata.py
+++ b/components/tools/OmeroPy/src/omero/util/populate_metadata.py
@@ -235,6 +235,9 @@ class ValueResolver(object):
     """
 
     AS_ALPHA = [chr(v) for v in range(97, 122 + 1)]  # a-z
+    # Support more than 26 rows
+    for v in range(97, 122 + 1):
+        AS_ALPHA.append('a' + chr(v))
     WELL_REGEX = re.compile(r'^([a-zA-Z]+)(\d+)$')
 
     def __init__(self, client, target_object):


manics:
Could you generalise this further by making it into a function?

```
ALPHA = [chr(v) for v in range(ord('a'), ord('z') + 1)]
def as_alpha(n):
    return (ALPHA[n / 26] if n >= 26 else '') + ALPHA[n % 26]
```
```
In [11]: as_alpha(0)
Out[11]: 'a'

In [12]: as_alpha(675)
Out[12]: 'zz'
```

chris-allan:
You certainly could, @manics but we do use this array to translate in both directions so you would have to implement two functions. Given its limited use I can't imagine the performance impact to be that significant function calls vs. array indexing. We have an implementation of this in the forward direction in `omero.gateway._letterGridLabel()`. Maybe time to unify all of this and at the same time make sure that the "Plate Grid" infrastructure works properly with 1536 well Plates? I don't recall it not working when we've encountered them but that doesn't mean that users haven't just not complained.

/cc @emilroz

manics:
@chris-allan 53ddee2cc69e934cc6737b309be534aec8fd1e72 handles `a`-`az` (52 rows) which is fine for the current problem, but doesn't seem particularly future-proof to me.

FYI there's a reverse function in https://github.com/openmicroscopy/openmicroscopy/blob/v5.3.0-m2/components/tools/OmeroPy/test/integration/metadata/test_populate.py#L56

chris-allan:
Indeed it does not, @manics. Switching to a functional and unified way of handling this in Python is exactly what I was suggesting. Consolidating all of this seems like the logical course of action whether that's in the forward or reverse direction.

Even addresses your pesky `TODO`. :)

joshmoore:
--rebased-to #4755

joshmoore:
Merging based on @eleanorwilliams' testing of gh-4755

PR #4744, opened 2016-07-04
Configuration of web link-to URL protocol/host

manics:
# What this PR does

The web metadata panel `Link to this <object>` contains a server side URL created from the HTTP `Host` header. This causes problems with server-side caching.

- Wherever possible client-side URLs should be constructed in javascript using the `location` object, so that the `/path` will be cached but the `protocol://host` will be handled by the client browser
- Where this isn't possible (e.g. embedded non-javascript metadata such as URLs parsed by social media) this can be overridden using `omero.web.ui.omero.web.ui.external_link_baseurl` (this must include the protocol).

# Testing this PR

1. Setup OMERO.web, optionally with a public user (makes testing easier)
2. Check `Link to this <object>` in the web RH metadata panel. The URL should use whatever the browser has in the location bar, and in additional examination of the returned html such as `http://host/webclient/metadata_details/image/ID/` should show `var lnk = location.protocol + "//" + location.host + "/webclient/";` instead of a URL.
3. Set `omero.web.ui.omero.web.ui.external_link_baseurl`: If twitter or opengraph is enabled the html for the full image viewer should contain URLs with this value instead of the Host header.

# Related reading

- See https://trello.com/c/KETiUCeR/373-metadata-cache-refresh-incorrect-link-to for a particular use-case

# Note

The only other place I'm aware of that let's you copy a link is in the full image viewer where it is already generated by client-side javascript: https://github.com/manics/openmicroscopy/blob/metadata52-linkto/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html#L665

This is unchanged, but maybe it should be unified?

Note: Description updated 2016-07-07

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -103,12 +103,7 @@
                 // show a link to the current object
                 $("#show_link_btn").click(function(){
                     $("#link_info_popup").show();
-                {% if webclient_path %}
                     var lnk = location.protocol + "//" + location.host + "{{ webclient_path }}";
-                {% else %}
-                    var lnk = location.href.substring(0,location.href.length - location.search.length);
-                    if (lnk.charAt( lnk.length-1 ) == "#") { lnk = lnk.substring(0, lnk.length-1)}
-                {% endif %}
                     var obj_type = "{{manager.obj_type}}";
                     if (obj_type === "acquisition") {
                         obj_type = "run";
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -105,6 +105,8 @@
                     $("#link_info_popup").show();
                 {% if webclient_path %}
                     var lnk = "{{ webclient_path }}";
+                {% elif webclient_pathname %}
+                    var lnk = location.protocol + "//" + location.host + "{{ webclient_pathname }}";
                 {% else %}
                     var lnk = location.href.substring(0,location.href.length - location.search.length);
                     if (lnk.charAt( lnk.length-1 ) == "#") { lnk = lnk.substring(0, lnk.length-1)}
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -104,9 +104,7 @@
                 $("#show_link_btn").click(function(){
                     $("#link_info_popup").show();
                 {% if webclient_path %}
-                    var lnk = "{{ webclient_path }}";
-                {% elif webclient_pathname %}
-                    var lnk = location.protocol + "//" + location.host + "{{ webclient_pathname }}";
+                    var lnk = location.protocol + "//" + location.host + "{{ webclient_path }}";
                 {% else %}
                     var lnk = location.href.substring(0,location.href.length - location.search.length);
                     if (lnk.charAt( lnk.length-1 ) == "#") { lnk = lnk.substring(0, lnk.length-1)}
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -680,15 +680,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
          None,
          identity,
          ("Use this host (including protocol) for absolute urls displayed in "
-          "the client. If omero.web.ui.use_client_host=True this will only be "
-          "used for absolute URLs which cannot be created with client-side "
-          "javascript.")],
-    "omero.web.ui.use_client_host":
-        ["WEB_URL_USE_CLIENT_HOST",
-         False,
-         bool,
-         ("If True build link-to URLs using client-side javascript, otherwise"
-          " use the server side HOST header")],
+          "the client. This does not include client-side generated URLs.")],
 }
 
 DEPRECATED_SETTINGS_MAPPINGS = {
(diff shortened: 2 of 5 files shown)


components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
```diff
@@ -105,6 +105,8 @@
                     $("#link_info_popup").show();
                 {% if webclient_path %}
```
aleksandra-tarkowska: is it not enough to just delete that if?

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
```diff
@@ -105,6 +105,8 @@
...
                     var lnk = "{{ webclient_path }}";
+                {% elif webclient_pathname %}
+                    var lnk = location.protocol + "//" + location.host + "{{ webclient_pathname }}";
                 {% else %}
                     var lnk = location.href.substring(0,location.href.length - location.search.length);
```
aleksandra-tarkowska: if this JS is invalid then maybe better is to fix it rather then introduce another condition.

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
aleksandra-tarkowska: I've tested that briefly in https://github.com/aleksandra-tarkowska/mapannotations/commit/7d911dff2f3fdfd069a591c448af20b63571dd17
and it work http://10.0.51.133/web/mapannotations/gene/?show=map.value-tea1

manics: I kept it at first in case anyone wanted the old behaviour. If you're happy (as far as I know the Javascript below should work) I'll remove it.

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
manics: I've no idea under what condition this would be reached, prior to this PR `webclient_path` should always be defined.

aleksandra-tarkowska: jsut delete `context['webclient_path']` from the `load_metadata_details` it is not used for anything else

manics:
@aleksandra-tarkowska Updated

aleksandra-tarkowska:
thx, but I think you also need to fix batch annotations. Why we need new property in a viewer?

manics:
The last commit changes batch-annotate to use client-side JS.

The additional property `omero.web.ui.default_client_baseurl` is for absolute URLs that need to be generated as HTML instead of JS. The only uses I'm aware of are the URLs for Twitter and Opengraph integration #4613, currently only on `metadata52`, so I put those changes in a separate commit f8d827266c947a9db6b8f92907f157f4510bae06

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
aleksandra-tarkowska: if you want to maintain the old logic then you need to keep `request.build_absolute_uri(reverse('webindex'))` rather then JS created link twice. The code below has no sense to me.

aleksandra-tarkowska: From the previous discussion my impression was that we want to remove usage of `build_absolute_uri`

aleksandra-tarkowska: @will-moore what was the logic behind https://github.com/manics/openmicroscopy/commit/a9d87d9d8bf1a3ad04e89a7137a55823f9726ce5#diff-5ecd969a9790526e02d15fcb673d6ca5R119 vs https://github.com/manics/openmicroscopy/commit/4eebe363cd36bdc0e358162c12dd057cdfbb7f19#diff-5ecd969a9790526e02d15fcb673d6ca5R116 At the moment we have both in place and we are wondering when `else` is used?

aleksandra-tarkowska: despite the fact that doing https://github.com/manics/openmicroscopy/commit/a9d87d9d8bf1a3ad04e89a7137a55823f9726ce5#diff-e226bce909b5b71ca51f5a26c7db1823R748 is dangerous!

manics: > From the previous discussion my impression was that we want to remove usage of build_absolute_uri

I agree, I was just a bit hesitant to do the full cleanup because I don't understand why the `else` is there. I'll remove it if no-one can objects.

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
will-moore: I think the final ```else``` code was simply what I had originally, and maybe I was simply leaving it there in case the ```webclient_path``` didn't work in some situations. But no evidence that it was needed at all.

manics:
I've removed the `else`. I've also renamed the new property to `omero.web.ui.external_link_baseurl`.

aleksandra-tarkowska:
looks good to me

joshmoore:
Merging for metadata52. I assume before hitting the mainline this will need a discussion re: the unification that's mentioned in the description as well as documentation on what *all* values may need to be set and kept in sync. (If this is the one end-all-be-all property, then likely a shorter name would be appropriate)

manics:
--rebased-to #4792

manics:
--rebased-to https://github.com/openmicroscopy/openmicroscopy/pull/5244
--rebased-to https://github.com/openmicroscopy/openmicroscopy/pull/5245

PR #4723, opened 2016-06-20
Lut repo 2

jburel:
# What this PR does
This PR replaces https://github.com/openmicroscopy/openmicroscopy/pull/4719
 * LUT folder is now under /scripts (i.e. we can see it as script for channels display0
 * ``getScripts`` method will not return the LUT. This implies that the clients displaying the available scripts do not need to be modified.
 *  New method ``getScriptsByMimetype``. This is the method to use to load the LUT (or any script e.g. ``matlab`` scripts). To load the luts, ``scriptingService.getScriptsByMimetype("text/x-lut")``
 * Also possible to upload official luts.

# Testing this PR
Integration tests added/adjusted.

# Related reading

Link to cards, tickets, other PRs:

PR #4746, opened 2016-07-07
Lut selection insight

dominikl:
# What this PR does

Adds Lookup Tables support to Insight.

# Testing this PR

Select an Image. Go to right hand side Preview panel. Make sure you can select the lookup tables in the Channel color picker. This won't have an effect yet, because it's not implemented server side.

# Still ToDo

~~For now I just used the same icon for the 'lookup table' entry as the 'color list' entry in the tool bar. We'd need a new icon for that~~ /cc @gusferguson  -- **Done**
screen shot 2016-07-07 at 11 30 06

# Related reading

This PR is based on #4723


gusferguson:
@dominikl 

This is made using 4 of the imageJ standard LUTs. 32 x 32 px.

What do you think?

icons6

dominikl:
Looks good! Can you make a 24px one? I think the toolbar icons are 24x24.

gusferguson:
24 x 24 px

lut-icon-24

dominikl:
👍

PR #4751, opened 2016-07-15
TF8 import

dominikl:
# What this PR does

Treat `*.tf8` images like `ome.tiff` and import them as image (not as plate).

I'm not sure if this is good solution, as I still have lots of question marks about this process, e. g. why do these files have a plate definition in the metadata in first place, when they are not plates; and why do they get imported as images and not as plates, although they have plate definitions.

# Testing this PR

Import a `*.tf8` image into a (Project/)Dataset. Without the PR the image would have been recognised as plate (but not imported as such, for some reason...), and ended in `Orphaned Images` folder.
Test image: 206157 (on demo server)

# Related reading

[Ticket 13243](http://trac.openmicroscopy.org/ome/ticket/13243)

http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-July/006095.html

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/ImportableObject.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/ImportableObject.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2010 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -122,8 +122,10 @@ public class ImportableObject
 		
 
 		IFormatReader reader = new OMEXMLReader();
-		OME_SUFFIXES = (List<String>) Arrays.asList(reader.getSuffixes());
-
+		OME_SUFFIXES = new ArrayList<String>();
+		OME_SUFFIXES.addAll((List<String>) Arrays.asList(reader.getSuffixes()));
+		OME_SUFFIXES.add("tf8");
+		
 		ARBITRARY_FILES_EXTENSION = new ArrayList<String>();
 		ARBITRARY_FILES_EXTENSION.add("text");
 		ARBITRARY_FILES_EXTENSION.add("txt");
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroImageServiceImpl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroImageServiceImpl.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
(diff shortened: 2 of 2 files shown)


rleigh-codelibre:
Note that `.tf8` can be a plain TIFF as well as being a `.ome.tf8` OME-TIFF.  The former will never be a plate, but the latter could be either.  Not sure if that has any bearing on this PR though because I'm not sure why OMERO cares about the file extension--this is fully described by the image metadata, so I'm not sure why the file extension would play any part in this!

dominikl:
I was thinking that, too. The metadata in the file must be wrong, because it has a plate definition, but not supposed to be a plate. And for some reason, it's also not imported as plate. But because it has a plate definition, Insight Importer treats the file as plate. So how could the Insight Importer otherwise determine that this is a file with plate definitions but it's actually not a plate?

components/insight/SRC/org/openmicroscopy/shoola/env/data/model/ImportableObject.java
```diff
@@ -122,8 +122,10 @@
 		
 
 		IFormatReader reader = new OMEXMLReader();
-		OME_SUFFIXES = (List<String>) Arrays.asList(reader.getSuffixes());
-
+		OME_SUFFIXES = new ArrayList<String>();
+		OME_SUFFIXES.addAll((List<String>) Arrays.asList(reader.getSuffixes()));
+		OME_SUFFIXES.add("tf8");
```
joshmoore: I understand OME_SUFFIXES to have only the XML endings, no the TIFF endings.

joshmoore:
Re: the `tf8` ending -- it was mostly a hunch on my part since there is other code in insight to specifically check for `.ome.tif` and `.ome.tiff`. I don't know for a fact that this is the correct solution.

dominikl:
Apparently the server somehow knows that the file is not a plate, because it's not imported as such. But I couldn't find out yet, how that works.

jburel:
@dominikl: Check the output of ``isHCS`` in ``OmeroImageServiceImpl`` this might be indicating that might not be returning the expected value

joshmoore:
> Apparently the server somehow knows that the file is not a plate, because it's not imported as such. But I couldn't find out yet, how that works.

Initially, `ImportCandidates` just checks for whether or not `HCS_DOMAIN` is present:

```
                boolean isSPW = Arrays.asList(domains).contains(FormatTools.HCS_DOMAIN);
```

Later, I think there would need to be a `Plate` properly linked in order to import as HCS.

dominikl:
Insight determines `isHCS` via ImportCandidates `isHCS`, which is set if the Reader reports `HCS_DOMAIN`, which in case of the OMETiffReader is true if the file has plate definitions https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/OMETiffReader.java#L275 , https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/OMETiffReader.java#L523 
Actually everything is correct. I guess to reply to the mailing lists, the answer would be the metadata of the file if wrong (if it's not supposed to be a plate). But still I wonder why the file doesn't get imported as plate, although the metadata says so.

dominikl:
Ok, I close that again then...

PR #4752, opened 2016-07-19
LUT selection Insight

dominikl:
# What this PR does

Adds Lookup Tables support to Insight.

# Testing this PR

Select an Image. Go to right hand side Preview panel. Make sure you can select the lookup tables in the Channel color picker. ~~This won't have an effect yet, because it's not implemented server side.~~ See @jburel 's  https://github.com/openmicroscopy/openmicroscopy/pull/4752#issuecomment-233752569 , test against breaking.


jburel:
I have opened https://github.com/openmicroscopy/openmicroscopy/pull/4753 so you can test against breaking.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#388](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/388/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/388/consoleText) for more details.
Possible conflicts:
  - PR #4734 dominikl 'Refactor PojoMapper'
    - components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroImageServiceImpl.java
  - PR #4751 dominikl 'TF8 import'
    - components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroImageServiceImpl.java

components/insight/SRC/org/openmicroscopy/shoola/util/ui/colourpicker/RGBModel.java
```diff
@@ -484,6 +501,56 @@ boolean isOriginalColor(Color color)
...
+     * @param The
+     *            lookup table
+     * @return <code>true</code> if it is, <code>false</code> if it is not
+     */
+    boolean isOriginalLut(String lut) {
+        if (originalLut == null ^ lut == null)
+            return false;
+        if (originalLut == null && lut == null)
+            return true;
+        return originalLut.equals(lut);
```
mtbc: looks like you want `ObjectUtils.equals` from Apache Commons

pwalczysko:
screen shot 2016-07-21 at 10 46 58
I genuinely did not know what to pick out of the four icons. 
The first icon is maybe telling me something, the 2, 3, and 4th icon are just a colorful splodges for me, do not see anything to identify them as LUTs or otherwise

pwalczysko:
I applied a "magenta" LUT on my blue channel. The obvious problem is that now I want to adjust the slider, but the slider is still blue....
screen shot 2016-07-21 at 10 50 58
I completely lost correspondence between the slider and the image in this way.

pwalczysko:
RFE: Alphabetical order needed here:
screen shot 2016-07-21 at 10 59 28

pwalczysko:
Discussed with @rleigh-dundee . The UI needs some thinking. Comparing with the Zeiss ZEN software - the colors are basically lookup tables, which can be switched in between. This means that the lookup tables live under one "icon", when this icon is clicked and a LUT was previously selected, the LUT is automatically applied, and also the LUT selection menu applies (in order to be able to change the selection of the LUT). Either you are on a single-color table (Red, Green etc.) or on a LUT. Insight have now implemented a strange hybrid, where the color persists on the slider and the channel button, even if the image has a LUT applied.
Also, the LUTs have mini-icons on each menu item in ZEN, which basically gives you an idea how the LUT looks like (atm in Insight you will never get any hint about how the LUT looks like, even after you applied it to the image).

zen-lut

pwalczysko:
BUG: The HiLo LUT (asked for by users already) is not showing blue on the low-end but white instead in insight. Compare the very same image imported to Fiji and OMERO.insight - in Fiji is the HiLo blue as it should be, in insight it is white.

screen shot 2016-07-21 at 12 04 23
screen shot 2016-07-21 at 12 03 38

jburel:
@pwalczysko: 
 * The backend is still work in progress, this is not related to this PR necessarily
 * luts are files on disk similar to scripts i.e. no icon associated to them. Unless we generate an icon every time on the fly we will have to use a generic icon. We will have to check the performance
This is also related to the update of the button/slider point.

Also:
 * ``.lut`` extension should be removed and maybe some "name cleaning" as we do for script e.g. remove ``_``

dominikl:
With the last couple of PRs I fixed these things:
- Name cleaning: A proper name for the lookup table is now shown instead of the plain filename
- Removed the different color pickers, only the color wheel and lookup table pickers will be displayed; if a lut is specified the dialog will show the lut picker by default (if not the color picker)
- If a color is picked the lookup table will be unset (otherwise picking a color won't have an effect because lut takes precedence).
- If a lut is set, the color of the channelbutton and slider is removed (background is set to default gray panel background). This could be improved to use the real lut colors in future.

jburel:
when this work is complete, we should take a screenshot that we can use for a tweet cc @hflynn

will-moore:
I think we need to keep the ability to pick simple colours, E.g. red, green, blue, yellow, white - maybe magenta and others. See Gus' mock-up at https://github.com/openmicroscopy/design/issues/36#issuecomment-218185643. Here (as in other tools discussed) the simple colours are in the same list as the lookup tables.

I think the default view of the ```Color Picker``` should be simple colours (first time I open it), with LUTs further down the same list OR a separate tab, and the full colour picker in another tab too. (2 or 3 tabs).
Usually you'd only want the full color-picker is rare cases, and the icon for it is easier to understand for users wanting to find the color-picker.

If I have an image with "grayscale" enabled and I pick a color in the color-picker then the "grayscale" checkbox is unchecked so that I can see the color I picked. However this doesn't happen when I pick a LUT (image stays gray, so I don't see the LUT colors rendered).

pwalczysko:
Save to All does not work as expected. 
Select in first channel of the 56-lei MIF a Magenta color. In the second channel, select HiLo LUT.
On this very image, all is as expected. Hit Save to All. The first channel (color) is propagated as expected. The second channel (LUT) is not. See screenshots.

The originally changed  image
screen shot 2016-07-28 at 11 43 23

One of the images in the dataset after Save to All was applied
screen shot 2016-07-28 at 11 43 58

jburel:
@pwalczysko: i probably need now review all the save methods in backend. I have not done that yet

jburel:
https://github.com/openmicroscopy/openmicroscopy/pull/4752#issuecomment-235861943 fixed by
https://github.com/openmicroscopy/openmicroscopy/pull/4753/commits/433cf2ed5504e713431ddc6bda33fb4f50524958

dominikl:
Closing in favor of #4764 .
The Lookup tables are now shown in the Color selection list. Quite a different approach, therefore it's easier to open a new PR instead of trying to merge that into this PR.

PR #4426, opened 2016-01-21
Extenstions to populate metadata 52

joshmoore:
A rebase to 5.2 of @emilroz' port of all `metadata` PRs involving `populate_metadata`.

----

This includes:

 * [New CLI plugin: `metadata`](https://github.com/openmicroscopy/openmicroscopy/pull/4426/files#diff-14ca86402c0d762d3b1cdaa4c9890fdfR30) and
 * [New CLI command: `fs ls`](https://github.com/openmicroscopy/openmicroscopy/pull/4426/files#diff-8b2f1200f3424459187cc6f91a80fbedR686) which need design thought
 * [A workaround for binary download](https://github.com/openmicroscopy/openmicroscopy/pull/4426/files#diff-68d677662210452db071b951f04e9a4fR105) which likely needs discussion
 * As well as:
   * Numerous fixes to `populate_metadata.py` incl. a yml config format (**Requires PyYAML**)
   * Fixes to `populate_roi.py`
   * A number of new tests
   * Closes https://trac.openmicroscopy.org/ome/ticket/12235

----

Description of cherry-picking:

> Extensions to the populate metadata functionality.
> 
> Cherry-picked from:
> * https://github.com/openmicroscopy/openmicroscopy/pull/3876,
> * https://github.com/openmicroscopy/openmicroscopy/pull/3889, 
> * https://github.com/openmicroscopy/openmicroscopy/pull/4037, 
> * https://github.com/openmicroscopy/openmicroscopy/pull/4240,
> * https://github.com/openmicroscopy/openmicroscopy/pull/4251
>
> To ensure no merge conflicts also picked a commit https://github.com/openmicroscopy/openmicroscopy/commit/6ed797d365186737d9a6b53a5c653c32792974e5 from https://github.com/openmicroscopy/openmicroscopy/pull/3875
>
> https://github.com/openmicroscopy/openmicroscopy/pull/4251 and  https://github.com/openmicroscopy/openmicroscopy/pull/4240 are conflicting so added extra commit from @joshmoore at the end here.
>
> Because of some weirdness trying to git merge Simon's branch for https://github.com/openmicroscopy/openmicroscopy/pull/4240 was also causing some merge conflicts but they looked absurd so I cherry-picked the branch.

PR #4765, opened 2016-07-31
Revert "Disable metadata.py with OMERO_DEV_PLUGINS"

manics:
# What this PR does

Reverts the requirement to set `OMERO_DEV_PLUGINS` when using the `omero metadata` plugin. This env-var makes sense on the mainline, but since this plugin is in integral part of the work in the `metadata52` development branch it should be enabled by default.

# Testing this PR

1. Check `OMERO_DEV_PLUGINS` isn't set: `$ set | grep OMERO_DEV_PLUGINS`
2. Run `omero metadata ...`, the plugin should be enabled

# Related reading

- Original `develop` PR: https://github.com/openmicroscopy/openmicroscopy/pull/4426


joshmoore:
Additionally, for OMERO 5.3 we likely want to solidify the API and rebase this to the mainline as well.

cc: @ximenesuk @sbesson @jburel

atarkowska:
--rebased-to #5221

PR #4766, opened 2016-08-01
Rgb to hex

will-moore:
# What this PR does

This separates some changes that were in the histogram work: https://github.com/openmicroscopy/openmicroscopy/pull/4703 into a separate branch to avoid conflicts between #4703 and LUT work.
Here, we always use hex notation for colors in the color-picker and color-picker buttons instead of switching between hex and rgb notation.
Previously, we'd use hex notation for channel colors when the viewer / Preview panel first loads, then when a color is picked we'd use rgb notation.

# Testing this PR
This PR should not show any functional changes to the UI. 
Check that color-picker still works OK and that updated robot tests are passing (Both of these are already tested as part of #4703)

### DIFF
--- a/components/tests/ui/testcases/web/rdef_test.txt
+++ b/components/tests/ui/testcases/web/rdef_test.txt
@@ -10,7 +10,7 @@ Suite Teardown      Close all browsers
 
 *** Variables ***
 
-${importedChColor}                  rgb(128, 128, 128)
+${importedChColor}                  808080
 ${importedMax}                      255
 ${sizeZ}                            3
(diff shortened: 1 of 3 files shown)


bramalingam:
Changes to the Color Picker work as expected, but there seems to be some issue with eel and cowfish servers (None of the save,delete operations are going through from the web-client). And thus will wait till the issue is resolved to redo the test.

jburel:
@bramalingam: did you redo the tests?

will-moore:
Closing since this PR is included in #4767 and is tested there.

PR #4768, opened 2016-08-01
 Ticket 13269

dominikl:
# What this PR does

Fixes [Ticket 13269](https://trac.openmicroscopy.org/ome/ticket/13269) (NPE when a Well without PlateAcquisition is selected).

Instead of checking the PlateAcquisition now the Well of the WellsampleData is checked if it is restricted with respect to downloading. But I don't really understand the implications here. Can a WellsampleData have a PlateAcquisition and a Well or both, and what to check in that case?

# Testing this PR

Import the file mentioned in the ticket. Click on the Well.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/PojosUtil.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/PojosUtil.java
@@ -100,7 +100,7 @@ public class PojosUtil {
             WellSampleData w = (WellSampleData) obj;
             return w.getImage().isArchived()
                     && !w.asWellSample()
-                            .getPlateAcquisition()
+                            .getWell()
                             .getPlate()
                             .getDetails()
                             .getPermissions()


mtbc:
http://www.openmicroscopy.org/site/support/omero5.2/developers/Model/EveryObject.html#wellsample would lead one to expect always a well and maybe also a plate acquisition.

mtbc:
(In testing this kind of thing watch out for http://trac.openmicroscopy.org/ome/ticket/11486.)

dominikl:
Thanks @mtbc . Then it should be safe to only check the Well for download restrictions.

mtbc:
The code change makes sense and this PR fixes the bug I reported and other plates seem to still work fine. Good to merge once a daily round of CI is passed, thank you!

mtbc:
Good to merge. :+1:

PR #4772, opened 2016-08-02
Improve CLI upload plugin

ximenesuk:
# What this PR does

This PR standardises the `upload` output to be of the form `Object:id` (including multiple ids and elision if necessary). The integration tests have been modified. Finally, it removes the unused `--pytable` option.

# Testing this PR

Use a locally installed server (or installation of OMERO.py) execute against a local or remote server some `upload` commands, single and multiple files, and check the output:
```
$ omero upload new.tmp new.xml
OriginalFile:93,94
$ omero obj get OriginalFile:93 name
new.tmp
```
Note that large files may take a while to upload so testing is best done with minimal files.

Note that the old-style output was:
```
$ omero upload analysis.csv
Uploaded analysis.csv as 275
```

The integration tests should pass.

# Additional discussion

Historically, certain file extensions are excluded according to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/util/originalfileutils.py this is intended prevent image files being uploaded. The extension list is probably hopelessly out of date. Also, are there occasions when an image file may want to be uploaded? I have left this as is but it may warrant further discussion.

The removed `--pytable` option did nothing at all. There may have been an intention to have the option do something specific.


manics:
It's probably worth mentioning at this point that @dominikl assumes all tables have the mimetype `OMERO.table` in https://github.com/openmicroscopy/openmicroscopy/pull/4666#issuecomment-228736261

I was under the impression this is what the `--pytables` flag did, but apparently it's a no-op!

manics:
--rebased-to #4773

ximenesuk:
Re https://github.com/openmicroscopy/openmicroscopy/pull/4772#issuecomment-236932070 I'm happy to add it back if there is a good reason, and suggested code! It has been there a long time doing nothing at all!

manics:
Given that it's never done anything how about leave it out, and in future consider a proper way of defining mimetypes instead of something specific to tables?

components/tools/OmeroPy/src/omero/plugins/upload.py
```diff
@@ -64,9 +70,27 @@ def upload(self, args):
...
+        if args.annotate:
+            c = self.ctx.conn(args)
+            update = c.sf.getUpdateService()
+            query = c.sf.getQueryService()
+
+            for fId in fileIds:
+                # obj new FileAnnotation file=OriginalFile:275
+                ann = FileAnnotationI()
+                # f = file get by id
+                ann.file = query.find("OriginalFile", fId)
```
manics: Do you need to query the file? Does `ann.file = omero.model.OriginalFileI(fId, False)` work?

ximenesuk: Yes, that form works fine, I'd just reused a pattern from elsewhere without looking any further. I've added a commit to use this neater pattern. And removed a couple of my temporary note comments.

manics:
Oh hang on, now that I've read this properly I've realised I was envisaging actually annotating an object (e.g. `--annotate Image:1234`). Is that too far out of scope?

manics:
If you want I'm happy to take over this PR

ximenesuk:
I have no problem working on this PR but I do think we need some scope as to what is needed here. The most common use of an uploaded file is as a FileAnnotation and so the current enhancement seems reasonable. Given that the command as it stands returns the FileAnnotation in a format that can be parsed and fed to the `obj` command which is designed for creating links and further annotating could easily be scripted. There is a danger of packing too many bells and whistles into CLI commands that should be deferred to command line scripting.

If the `upload` command was to also create the links would it do it with all the provided files? Would multiple target objects be required, `--annotate Images:1,2,3`? In the case of links being made what would be the expected output, the FileAnnotations?

/cc @joshmoore @sbesson

manics:
> Given that the command as it stands returns the FileAnnotation in a format that can be parsed and fed to the obj command which is designed for creating links and further annotating could easily be scripted. There is a danger of packing too many bells and whistles into CLI commands that should be deferred to command line scripting.

That's a fair point. Based on that I don't think just creating the `FileAnnotation` offers much- I'd go for either encouraging scripting for everything, so reverting to the old behaviour of uploading only (no `--annotate`) but returning the improved string for parsing, or go for a full upload and annotate workflow.

Here's my current workflow, which is probably similar to @eleanorwilliams:

```
omero upload XXX.h5
omero obj new FileAnnotation file=OriginalFile:NN
omero obj new ScreenAnnotationLink parent=Screen:NN child=FileAnnotation:NN
omero obj update FileAnnotation:NN ns=openmicroscopy.org/omero/features/0.2
```

manics:
> If the upload command was to also create the links would it do it with all the provided files?

Until I looked at the PR I wasn't aware you could upload multiple files in one go! So I think everything should be linked to the target.

ximenesuk:
Looking at your standard workflow, and as two complete asides:

Can you not set the namespace when you create the FileAnnotation? Or can it only be done once the link is made?

I've often wondered why a
```
omero obj link Screen:NN FileAnnotation:NN
```
wasn't available for convenience? (But that's another story)

Back to this PR, the options are, in increasing complexity!

1. - revert a few commits and simply provide upload?
2. - link all uploaded files to a single target?
3. - allow multiple targets of one object type? `Image:1,2,3`
4. - allow multiple targets of multiple object types - now the arg parsing gets complicated!

Do you have a use case where a FileAnnotation needs linking to multiple objects?

I would probably go for 1 (KISS)  as even if 2 was available you would still need to set the namespace in an additional step.

manics:
> Can you not set the namespace when you create the FileAnnotation? Or can it only be done once the link is made?

Good point. Haven't tried it, mostly because the file was uploaded before we'd agreed on a namespace. I've only ever linked the FileAnnotation to one object.

Out of the four choices I vote for either (2) with the FineAnnotation returned so you can set the namespace, or (1) and maybe add some functionality to the `metadata` plugin.

will-moore:
This is breaking the Robot Framework setup script, (see https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/355/console)
since the script tries to get fileId from upload: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tests/ui/robot_setup.sh#L122

ximenesuk:
Thanks @will-moore, I'll fix the robot test in this PR.

ximenesuk:
@manics after considering this I have gone for 1 as the simplest solution as everything else can now be easily scripted.

Providing an Object:ID as an annotation target would involve YAKIP (Yet Another Klass Id Parser) beyond simply splitting the input. We have parsers under `GraphControl` and `TxAction` (and possibly others). I didn't want to create another for such a limited CLI command.

ximenesuk:
See https://ci.openmicroscopy.org/job/OMERO-DEV-merge-robotframework/356/

pwalczysko:
Re:
>Historically, certain file extensions are excluded according to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/util/originalfileutils.py this is intended prevent image files being uploaded. The extension list is probably hopelessly out of date. Also, are there occasions when an image file may want to be uploaded? I have left this as is but it may warrant further discussion.

See my test below (note that the ``dv`` extension is in the list. This strongly suggests that the code actually does not anything (i.e. does not work).
I would thus suggest to remove the code restraining an upload on any file extensions, as this is 
A) not what we want and tell users 
B) it does not work anyway

```
[pwalczysko@ls31619 ~/Work/openmicroscopy/dist]$ bin/omero upload ~/Downloads/dv/SMN10ul03_R3D_D3D.dv ~/Downloads/dv/IAGFP-Noc01_R3D.dv
Using session 01aecddc-3d54-408d-8a81-a3131f559692 (user-11@eel.openmicroscopy.org:4064). Idle timeout: 10 min. Current group: read-annotate-1
OriginalFile:28934,28935
```

pwalczysko:
Other than https://github.com/openmicroscopy/openmicroscopy/pull/4772#issuecomment-238203376, the funtional tests fully passed.

ximenesuk:
@pwalczysko it looks like no files are excluded from being imported because the utility library very badly broken. On the upside it is only used by this command and its effect may be a positive one. See: https://trello.com/c/tlPrvcvi/699-cli-originalfileutils-py-badly-broken

ximenesuk:
This final commit removes the faulty file entirely and refactors the current behaviour back to the plugin,  which is to use the `mimetypes` package. The behaviour should stay the same, no files should be excluded from upload. 

If some form of `omero mimetypes` based on the file format is required than that should be added back in a later PR once the requirement has been decided.

pwalczysko:
As expected, the last commit did not change the behaviour (checked the commands in the header of the PR and the command in https://github.com/openmicroscopy/openmicroscopy/pull/4772#issuecomment-238203376).

Some questions about the 5 failed robot tests https://ci.openmicroscopy.org/job/OMERO-DEV-merge-robotframework/358/robot/ @will-moore ?

One failed integration test as well https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/339/testReport/. 

Otherwise all fine.

ximenesuk:
The failed integration test https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/339/testReport/.  is unrelated and due to gh-4776

I don't think the robot tests are related but I'll let @will-moore confirm after tomorrow's builds.

ximenesuk:
Just for confirmation: https://ci.openmicroscopy.org/job/OMERO-DEV-merge-robotframework/360/

With robot and integration tests passing and @pwalczysko's okay I think this is now good to merge.

ximenesuk:
@sbesson @jburel would it be possible to get this merged?

jburel:
@ximenesuk: I was waiting for you to be back to double-check

PR #4774, opened 2016-08-04
Search result tables 13255

will-moore:
# What this PR does

Fixes https://trac.openmicroscopy.org/ome/ticket/13255. When an Image from SPW is found in a search and loaded in the right panel, we don't know that this is in a Well.
Therefore, the *Tables* tab was not shown so you couldn't query for tables data (needs Well Id).
Now, we always show the Tables tab, and if a user expands it then we do an AJAX call to check whether the image is in a Well (uses the paths_to_object query). If we get a Well Id then we do a second query to load the Tables data as normal.

# Testing this PR

See ticket above. 
 - Need to have tables data in SPW as described http://help.openmicroscopy.org/scripts.html#metadata
 - Add table data for a bunch of wells in plate (doesn't have to be all)
 - Table data should be displayed as normal when the Well is loaded in right panel.
 - Add unique tag name to Wells (images) to aid searching for them.
 - Search for tag name - click on Images in search results
 - In right panel expand Tables tab - should load Tables data as before.
 - Clicking on another SPW image in search results should auto-load Tables data if tab is left open.
 - For non-SPW images (either in search results or main Data tab) clicking Tables tab will open tab and query for wellId, but well Id won't be found so the Tables tab will not be opened automatically for subsequent images.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/show.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/show.py
@@ -685,7 +685,8 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
             select coalesce(sowner.id, plowner.id, aowner.id, wsowner.id),
                    slink.parent.id,
                    plate.id,
-                   acquisition.id
+                   acquisition.id,
+                   well.id
             from WellSample wellsample
             left outer join wellsample.details.owner wsowner
             left outer join wellsample.plateAcquisition acquisition
@@ -746,6 +747,12 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
                 path.append({
                     'type': 'acquisition',
                     'id': e[3].val
+                })
+
+            if e[4] is not None:
+                path.append({
+                    'type': 'well',
+                    'id': e[4].val
                 })
 
             paths.append(path)
--- a/components/tools/OmeroWeb/omeroweb/webclient/show.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/show.py
@@ -749,7 +749,8 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
                     'id': e[3].val
                 })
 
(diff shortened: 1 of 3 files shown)


components/tools/OmeroWeb/omeroweb/webclient/show.py
```diff
@@ -685,7 +685,8 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
             select coalesce(sowner.id, plowner.id, aowner.id, wsowner.id),
                    slink.parent.id,
                    plate.id,
-                   acquisition.id
+                   acquisition.id,
+                   well.id
             from WellSample wellsample
             left outer join wellsample.details.owner wsowner
```
manics: Unrelated to this PR, but why are `wsowner` and `aowner` the same: `wellsample.details.owner`?

manics:
Otherwise good to merge.

components/tools/OmeroWeb/omeroweb/webclient/show.py
will-moore: We need the owner of the top element in the path. E.g. Screen OR Plate. Probably never need Acquisition or WellSample. Even if lower elements in the path are owned by a different owner, that doesn't matter. We just need to know the context for browsing the jsTree, starting at Experimenter -> Screen or Experimenter -> Plate.

jburel:
Maybe add some robot tests cc @bramalingam

aleksandra-tarkowska:
could robot rest be added in a separate PR, would be great if that is rebased onto metadata branch and we could test along side mapr

will-moore:
--rebased-to #4788

joshmoore:
Filed card to come back to this before 5.3.0: https://trello.com/c/QlAb4anu/158-search-result-tables-robot-tests

Merging along with #4788, any refactoring or bug fixing should happen in paired PRs (develop/metadata52)

PR #4776, opened 2016-08-09
Remove pyramid

ximenesuk:
This PR supersedes gh-4603. See that PR for earlier discussion.

This `fs` subcommand allows a specific pyramid file to be removed given its Image ID, this is regardless of whether it is empty or not. See: https://trello.com/c/TM1Si3UN/38-pyramids This doesn't provide an end to end workflow but it does allow an admin to develop a CLI workflow for dealing with problem pyramids. The command is admin-only currently but it could be relaxed if necessary. There is no test but I will add a basic test.

Testing
-------

On a fresh local system without significant imports one possible test is to import and image and record its ID, then create an artificial pyramid file:

```
touch OMERO/Pixels/X_pyramid
```
where X is the Image ID. Then:

```
$ bin/omero fs fixpyramid Image:X
```
should remove the pyramid file with a confirmation message.

Alternatively import an image that will produce a pyramid file, note its ID, check that the pyramid file has generated and then run the above command. The pyramid file should be deleted. Using a client it should then be possible to force the regeneration of the pyramid file by trying to view the image.

--breaking


ximenesuk:
@joshmoore As I'm away for a few days, should I close this temporarily or is it fine to be open on breaking for a few more days - particularly is anyone wants to try to work out what is happening!

joshmoore:
I'd think it's ok to leave it open. @mtbc and I will keep brainstorming.

jburel:
@ximenesuk could you fix the conflict? unless a solution is found this week, somebody will have to take over
Details have already been adde to the trello card listed above

ximenesuk:
Conflicts fixed. I doubt a solution will be found this week as @mtbc has no bandwidth, I guess @joshmoore is busy and I have little idea what is happening here.

jburel:
Closing this PR now that @ximenesuk is gone.
The work will have to be re-activated

mtbc:
It _may_ be somehow related to some old library version; might be worth trying reopening after we drop CentOS 6.

PR #4775, opened 2016-08-04
Mapr annotations

manics:
# What this PR does

Implements the grouping of MapAnnotations in separate namespaces required for https://trello.com/c/VVUqCeZC/467-populate-map-annotations

The processing of existing bulkmap configuration files is unchanged.

New support is added for a `group` column type in the configuration file, which must contain `groupname` indicating the namespace for the MapAnnotation and `columns` indicating the columns to be converted into the MapAnnotation. See

- `test/integration/metadata/bulk_to_map_annotation_context_ns.csv`
- `test/integration/metadata/bulk_to_map_annotation_context_ns2.yml`

For examples.

Currently this implements deduplication of MapAnnotations within a single invocation only. This means if existing MapAnnotations are already present `omero metadata populate` may create a duplicate.

# Testing this PR

Import a plate:
- `bioformats/tools/mkfake -rows 2 -columns 3 test-plate.fake`
- `omero import test-plate.fake`

Add a bulk annotation to the plate:
- `omero metadata populate --file openmicroscopy/components/tools/OmeroPy/test/integration/metadata/bulk_to_map_annotation_context_ns.csv Plate:ID`

Copy the bulk-map-annotation config file `components/tools/OmeroPy/test/integration/metadata/bulk_to_map_annotation_context_ns2.yml` and add the following at the end:
```
advanced:
    well_to_images: yes
```
This is a hack to get annotations on wells to display in the webclient.

Convert to MapAnnotations:
- `omero metadata populate --context bulkmap --cfg bulk_to_map_annotation_context_ns2.yml Plate:ID`

The table data should be formatted as described by `bulk_to_map_annotation_context_ns.yml`

i.e. some key-value pairs should go into a MapAnnotation with a custom namespace, the remainder should go into the bulk-annotations namespace

screen shot 2016-08-08 at 13 48 51

# Related reading

- https://trello.com/c/VVUqCeZC/467-populate-map-annotations
- https://github.com/eleanorwilliams/idr-metadata/pull/17
- https://github.com/eleanorwilliams/idr-metadata/pull/18

## Edits

- **Only** map annotations with a primary key defined are de-duplicated

PR #4786, opened 2016-08-16
Mapr annotations: query database for existing map-annotations

manics:
# What this PR does

This is a follow-up to https://github.com/openmicroscopy/openmicroscopy/pull/4775 to use/update existing map-annotations in the database instead of creating duplicates.

There are several limitations which is why I've kept this as a separate PR:
- De-duplication only occurs between different screens/projects. This mean you can't re-run `metadata populate` on the same screen with the same `bulkmap-config.yml` file unless you first delete the map-annotations on the target.
- The database query loads all map-annotations in the given namespace, this is potentially resource intensive on both server and client
- Haven't figured out proper deletion semantics yet
  - Can you delete only annotation links, unless it's the last one in which case also delete the annotation?

# Testing this PR

1. Setup is similar to https://github.com/openmicroscopy/openmicroscopy/pull/4775, but this time create two screens with at least one overlapping primary key
2. `populate metadata` both screens.
3. Only one map-annotation for each primary key should exist, regardless of how many screens it is found in

# For discussion: upsert

There's two use cases here:

1. CreateIfNotExist: Create XxxAnnotationLink(parent,chlid) only if that link doesn't already exist. This would help fix the current lack of implementation for modifying annotations on existing screens. Example:
  - A screen contains `Images 1-10`. All of the images are linked to `MapAnnotation A`.
  - `Images 11-100` are subsequently added to the screen, and also need to be linked to `MapAnnotation A`.
  - The easiest way is to create 100 ImageAnnotationLinks with the same child, and use `UpdateServer.saveAndReturnArray`. This will fail because some of the `ImageAnnotationLink` already exist
  - In practice the Images to be annotated are filtered according to some other criteria, so the existence of the link would have to be checked for each image
2. Creating/merging mMapAnnotations. In #4775 I added support for deduplicating MapAnnotations before they were saved. This PR now takes existing MapAnnotations into account. Most of the merging logic is handled by `components/tools/OmeroPy/src/omero/util/metadata_mapannotations.py`

# TODO:
- [x] Rename `groupname` to `namespace`
- [ ] Maybe delete?

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/metadata_mapannotations.py
+++ b/components/tools/OmeroPy/src/omero/util/metadata_mapannotations.py
@@ -49,11 +49,9 @@ class CanonicalMapAnnotation(object):
     ma: The omero.model.MapAnnotation object
     primary_keys: Keys from key-value pairs that will be used to form the
         primary key.
-    unique_keys: If False duplicate keys (with different values) aren't
-        allowed, default True (allowed)
     """
 
-    def __init__(self, ma, primary_keys=None, unique_keys=True):
+    def __init__(self, ma, primary_keys=None):
         # TODO: should we consider data and description
         self.ma = ma
         ns = unwrap(ma.getNs())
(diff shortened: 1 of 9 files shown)


manics:
I've changed `groupname` to `namespace`

components/tools/OmeroPy/src/omero/util/metadata_mapannotations.py
```diff
@@ -58,18 +61,13 @@ def __init__(self, ma, primary_keys=None, unique_keys=True):
...
             mapvalue = []
         self.kvpairs, self.primary = self.process_keypairs(
-            mapvalue, primary_keys, unique_keys)
+            mapvalue, primary_keys)
         self.parents = set()
 
-    def process_keypairs(self, kvpairs, primary_keys, unique_keys):
+    def process_keypairs(self, kvpairs, primary_keys):
         if len(set(kvpairs)) != len(kvpairs):
             raise ValueError('Duplicate key-value pairs found: %s' % kvpairs)
```
manics: @eleanorwilliams ran into a problem with idr0018 where there are duplicate name-values within the same MapAnnotation. Should this be allowed?

manics:
@joshmoore @aleksandra-tarkowska Given the deadline for the mapr work we should figure out what else needs to go into this PR.

joshmoore:
@aleksandra-tarkowska says this is working as expected. Merging for a final testing in demo3. Before backporting to the mainline, this will need to be checked for breaking behavior in the non-IDR case. Also, de-duplicated delete still needs implementing.

atarkowska:
--rebased-to #5223

PR #4787, opened 2016-08-17
properly count warnings and errors in OMERO server logs

mtbc:
# What this PR does

This PR attempts to increase accuracy in counting errors and warnings in OMERO.server logs.

# Testing this PR

Arrange some warnings and/or errors among the regular logs (like `Blitz-0.log`) and also `master.err` and suchlike. Run `bin/omero admin diagnostics` and check that they are correctly counted.

If it's too tricky to stimulate such log entries locally, just copy logs from some other system or append them to the end of yours: [from CI](https://www.openmicroscopy.org/private/ome-internal/instructions/server-logs.html), some bug report in a ticket, etc.

# Related reading

https://trello.com/c/ecsPd9WE/87-diag-false-positives-on-error

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1140,7 +1140,7 @@ OMERO Diagnostics %s
                 if not p.exists():
                     self.ctx.out("n/a")
                 else:
-                    warn_regex = '(-! )?[\d\-/]+\s+[\d:,.]+\s+warn(ing:)?\s'
+                    warn_regex = '(-! )?[\d\-/]+\s+[\d:,.]+\s+warn(i(ng:)?)?\s'
                     err_regex = '(-! )?[\d\-/]+\s+[\d:,.]+\s+error:?\s'
                     warn = 0
                     err = 0
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1140,18 +1140,17 @@ OMERO Diagnostics %s
                 if not p.exists():
                     self.ctx.out("n/a")
                 else:
+                    warn_regex = '(-! )?[\d\-/]+\s+[\d:,.]+\s+warn(ing:)?\s'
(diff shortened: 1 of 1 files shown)


manics:
Maybe we should just incorporate slack into omero :) https://github.com/manics/omero-fenton/blob/master/taillog.py#L17

mtbc:
:smiley: Of course note that some like `master.err` can exhibit a different format. :+1: for naming the groups though! (I was scared that the comma was locale-specific.)

manics:
> I was scared that the comma was locale-specific.

Maybe it is, I've never tested omero-fenton in a non-English locale!

joshmoore:
In Python, the warning may include an "I":

```
Processor-0.log:2016-08-02 17:20:47,966 WARNI [                             omero.utils] (MainThread) Reason: exception ::Ice::NoEndpointException
```

since the logging format specifies 5 characters:

```
LOGFORMAT = "%(asctime)s %(levelname)-5.5s [%(name)40s] " \
            "(%(threadName)-10s) %(message)s"
```

(https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.4/components/tools/OmeroPy/src/omero/util/__init__.py#L30) which matches the Java format (https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.4/etc/logback.xml#L31).

mtbc:
Good catch, thank you.

mtbc:
In trying to find a way to get errors into `master.err` for @pwalczysko's review I come across https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7660 which to me suggests that I need to loosen the regex further for lines starting `!!` @joshmoore?

mtbc:
@joshmoore: Not sure how to authentically test that last commit.

pwalczysko:
with the last commit, the workflow works also on the ``master.err`` with an extra word. The master.err was artificially creted using bits from https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7660. 

The other logs (Pixel, Indexer and Blitz) work also (with WARN as well as WARNI) .
Probably good to merge if no ideas about how to create a real master error.

joshmoore:
Far fewer false positives is probably better then the possibility of a few false negatives. Thanks all.

PR #4788, opened 2016-08-18
Search result tables 13255 (rebased onto metadata52)

will-moore:
This is the same as gh-4774 but rebased onto metadata52.

----

# What this PR does

Fixes https://trac.openmicroscopy.org/ome/ticket/13255. When an Image from SPW is found in a search and loaded in the right panel, we don't know that this is in a Well.
Therefore, the *Tables* tab was not shown so you couldn't query for tables data (needs Well Id).
Now, we always show the Tables tab, and if a user expands it then we do an AJAX call to check whether the image is in a Well (uses the paths_to_object query). If we get a Well Id then we do a second query to load the Tables data as normal.

# Testing this PR

See ticket above. 
 - Need to have tables data in SPW as described http://help.openmicroscopy.org/scripts.html#metadata
 - Add table data for a bunch of wells in plate (doesn't have to be all)
 - Table data should be displayed as normal when the Well is loaded in right panel.
 - Add unique tag name to Wells (images) to aid searching for them.
 - Search for tag name - click on Images in search results
 - In right panel expand Tables tab - should load Tables data as before.
 - Clicking on another SPW image in search results should auto-load Tables data if tab is left open.
 - For non-SPW images (either in search results or main Data tab) clicking Tables tab will open tab and query for wellId, but well Id won't be found so the Tables tab will not be opened automatically for subsequent images.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/show.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/show.py
@@ -685,7 +685,8 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
             select coalesce(sowner.id, plowner.id, aowner.id, wsowner.id),
                    slink.parent.id,
                    plate.id,
-                   acquisition.id
+                   acquisition.id,
+                   well.id
             from WellSample wellsample
             left outer join wellsample.details.owner wsowner
             left outer join wellsample.plateAcquisition acquisition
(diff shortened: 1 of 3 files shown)


will-moore:
--rebased-from #4774

joshmoore:
See https://github.com/openmicroscopy/openmicroscopy/pull/4774#issuecomment-240776552

Issue #4597, opened 2016-04-21
Idempotent configuration commands


dpwrussell:
During our recent ansible configuration of OMERO I had to do some less than desirable things to make it work because some of the configuration can not be done idempotently.

Examples:

When configuring `omero.web.ui.top_links` it is not possible to use `append` because this will result in (after multiple runs) repeated entries, so instead it is necessary to use `set` which necessitates specifying all of the defaults (which could easily change with a new OMERO version).
```yaml
omero.web.ui.top_links: '[["Data", "webindex", {"title": "Browse Data via Projects, Tags etc"}],["History", "history", {"title": "History"}],["Help", "http://help.openmicroscopy.org/", {"title": "Open OMERO user guide in a new tab", "target": "new"}],["Figure", "figure_index"],["Tag Search", "tagsearch"]]'
```

I think the best solution to this could be a new command `exists` for things which are lists.

For the `omero.web.ui.top_links` example:
```bash
omero config exists omero.web.ui.top_links '["Figure", "figure_index"]'
omero config exists omero.web.ui.top_links '["Tag Search", "tagsearch"]'
```

This one could probably be configured by manipulating XML instead, but as that is not the recommended way to do these changes then I think this makes more sense. Here is another example which could not be done that way though.

Plugin scripts are installed with `upload` and then replaced with `replace`. This is fine when doing things manually, but when using config management, it has no way of knowing what to do. If you run `upload` every time this results in the original script not being replaced. On the flip side, even if the script is already installed, it can't be replaced without some nasty bash foo to get the ID.

``` yaml

- name: check if figure script has been installed (database)
  shell: "{{ omero_base }}/OMERO.server/current/bin/omero -u root -w {{ omero_root_password }} -s localhost script list | grep 'Figure_To_Pdf' | gawk -F '|' '{ print $1 }'"
  sudo: yes
  sudo_user: "{{ omero_user }}"
  register: figure_script_id

# Install the figure script (only the first time)
- name: load plugin scripts (first time)
  shell: "{{ omero_base }}/OMERO.server/current/bin/omero -u root -w {{ omero_root_password }} -s localhost script upload omero/figure_scripts/Figure_To_Pdf.py --official"
  sudo: yes
  sudo_user: "{{ omero_user }}"
  args:
    chdir: "{{ omero_base }}/OMERO.web/current/lib/python/omeroweb/figure/scripts/"
  when: figure_script_id.stdout == ""

[...]

PR #4791, opened 2016-08-19
Add --set to config append, --report to config {set,append,remove}

manics:
# What this PR does

Adds `--set` option to `omero config append`, to indicate a value should only be added if it's not already in the list, taking into account the defaults, see https://github.com/openmicroscopy/openmicroscopy/issues/4597

Adds `--report` option to `omero config {append,set,remove}`. This will always print out `Changed: ...` if a change was made, and nothing if no change was made.

# Testing this PR

1. Check `omero config append/set/remove` behaves as before.

2. Try something like

        omero config append --set omero.web.ui.top_links '["History", "history", {"title": "History"}]'

   this should have no effect (E.g. `omero config get`, or `omero web restart` and check the top-links) because this value is part of the standard defaults.

3. Try it without `--set`, you should see the old behaviour i.e. the duplicate value is appended

        omero config append omero.web.ui.top_links '["History", "history", {"title": "History"}]'

4. Add a different value, e.g.

        omero config append --set omero.web.ui.top_links '["BBC", "http://www.bbc.co.uk/"]'

5. Try `omero config {append,set,remove} --report`. If a change was made you should see a message such as `Changed: Appended omero.web.ui.top_links:["BBC", "http://www.bbc.co.uk/"]`

The idea behind this is to assist with Ansible deployments where you want to programmatically know whether or not a change was made.

# Related reading

- https://github.com/openmicroscopy/openmicroscopy/issues/4597
- Motivation: Deploy [mapr](https://github.com/aleksandra-tarkowska/mapannotations) on the IDR with Ansible
  - https://github.com/openmicroscopy/infrastructure/pull/104

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -165,6 +165,9 @@ class PrefsControl(WriteableConfigControl):
         for x in [set, append, remove]:
             x.add_argument(
                 "KEY", help="Name of the key in the current profile")
+            # report is intended for use with cfg mgmt tools
+            x.add_argument("--report", action="store_true",
+                           help="Report if changes are made")
         set.add_argument(
             "-f", "--file", type=ExistingFile('r'),
             help="Load value from file")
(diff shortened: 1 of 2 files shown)


joshmoore:
Generally looks very useful. Primary thoughts are:

 * `append --set` isn't very intuitive (for me). `append --once` ?
 * `remove` could use a reciprocal flag. `--all` ?

cc: @ximenesuk -- with @manics away, do we want to copy this to the mainline for further investigation?

joshmoore:
Merging for the 0.0.7 tag. If we decide to rename the properties, then this will need a follow-up to metadata52 (though on the mainline we can do it straight-away)

joshmoore:
Note: new mapr PR which makes use of this feature: https://github.com/openmicroscopy/infrastructure/pull/108

aleksandra-tarkowska:
--rebased-to #4828

PR #4792, opened 2016-08-19
Partial rebase of 4744 Use client-side protocol/host in URL

manics:
Partial rebase of #4744 

# What this PR does

The web metadata panel `Link to this <object>` contains a server side URL created from the HTTP `Host` header. This causes problems with server-side caching.

- Wherever possible client-side URLs should be constructed in javascript using the `location` object, so that the `/path` will be cached but the `protocol://host` will be handled by the client browser

# Testing this PR

1. Setup OMERO.web, optionally with a public user (makes testing easier)
2. Check `Link to this <object>` in the web RH metadata panel. The URL should use whatever the browser has in the location bar, and in additional examination of the returned html such as `http://host/webclient/metadata_details/image/ID/` should show `var lnk = location.protocol + "//" + location.host + "/webclient/";` instead of a URL.

The following additions from #4744 related to twitter/opengraph support (#4613) have not been rebased:
- URLs in embedded non-javascript metadata such as URLs parsed by social media can be overridden using `omero.web.ui.omero.web.ui.external_link_baseurl` (this must include the protocol).

# Related reading

- See https://trello.com/c/KETiUCeR/373-metadata-cache-refresh-incorrect-link-to for a particular use-case

# Note

The only other place I'm aware of that let's you copy a link is in the full image viewer where it is already generated by client-side javascript: https://github.com/manics/openmicroscopy/blob/metadata52-linkto/components/tools/OmeroWeb/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html#L665

This is unchanged, but maybe it should be unified?

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -663,6 +663,12 @@ CUSTOM_SETTINGS_MAPPINGS = {
           " 'webtest/webclient_plugins/center_plugin.overlay.js.html',"
           " 'channel_overlay_panel']``. "
           "The javascript loads data into ``$('#div_id')``.")],
+    "omero.web.ui.use_client_host":
+        ["WEB_URL_USE_CLIENT_HOST",
+         False,
+         bool,
+         ("If True build link-to URLs using client-side javascript, otherwise"
+          " use the server side HOST header")],
 }
 
 DEPRECATED_SETTINGS_MAPPINGS = {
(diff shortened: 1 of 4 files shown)


manics:
Partially
--rebased-from #4744

will-moore:
Sorry - missed this yesterday. Looking now...

will-moore:
Not sure that I understand the "problems with server-side caching" but this certainly seems to be working fine.
On cowfish I see
```
var lnk = location.protocol + "//" + location.host + "/webmerge/webclient/";
```
and all the link functionality works fine as described.

Good to merge.

joshmoore:
> Not sure that I understand the "problems with server-side caching" but this certainly seems to be working fine.

@will-moore, the problem we were seeing is that the URL that someone used to access a resource (http://idr-demo.openmicroscopy.org/webclient/?show=plate-2551 vs http://idr-demo2.openmicroscopy.org/webclient/?show=plate-2551, aliases for each other) was being stored in the metadata cache for the right-hand panel.

PR #4654, opened 2016-05-11
Load annotations on demand

dominikl:
Up to now whenever an object is selected all annotations for this object are loaded. As we now have the different taskpanes on the right hand side metadata panel, which are closed by default, this is not necessary any more. With this PR only the specific annotations are loaded on demand, i.e when a taskpane is expanded. While the annotations are loaded from the server, the expanded taskpane displays a simple "Loading..." label.

I started this more as a kind of proof of concept, but it turned into a quite extensive PR. Which means there's a risk that it'll break something and needs a lot of testing (basically everything annotation related). Is it worth it? Taking into account, that this PR won't add a noticeable performance increase in most cases, only for data objects with a very large amount of annotations attached to them.

/cc @jburel @pwalczysko 

--exclude not intended for 5.3.0-m3

PR #4794, opened 2016-08-22
have metadata browser show annotation counts

mtbc:
# What this PR does

In Insight the right-hand metadata browser panel should now show annotation counts even before accordion tabs are expanded.

# Testing this PR

Try adding and removing annotations and see that the annotation counts generally stay correct, especially after selecting a different entity then moving back. Counts under multiselect should also be reasonable. Will generally correspond to *total* that is visible *across* different filtering options in right pane (e.g., show mine or others').

(Sync isn't always immediately perfect but trying to keep implementation sane.)

# Related reading

https://trello.com/c/bc6ZtV0x/144-rfe-indication-of-number-of-annotations-on-an-image-in-brackets-5-on-accordion-tabs-in-right-hand-pane

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AnnotationTaskPane.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/editor/AnnotationTaskPane.java
@@ -1,7 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2015 University of Dundee. All rights reserved.
- *
+ *  Copyright (C) 2015-2016 University of Dundee. All rights reserved.
  *
  *  This program is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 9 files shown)


jburel:
``
int getUnfilteredAnnotationCount() {
         if (model.isMultiSelection()) {
            return model.getAllAttachments().size();
         } else {
             return model.getAttachments().size();
         }
     }
``
The ``else`` statement is not needed

mtbc:
I figure the `else` makes it easiest to see the logical structure that the alternatives truly are two comparable branches; indeed, at the same indent they're also easier to compare by eye. Should I remove them?

joshmoore:
My code tends to have the `else` statements as well. Probably worth reviewing https://github.com/openmicroscopy/openmicroscopy/tree/v5.2.4/docs/styles and making sure we're all on the same page.

dominikl:
In this context, might be worth resurrecting #4654 (although @pwalczysko won't be happy testing it).

mtbc:
Ah, that's interesting. Might want to at least eagerly load the fact that there are any annotations.

dominikl:

jburel:
Certainly time to review the template: https://github.com/openmicroscopy/openmicroscopy/pull/4794#issuecomment-241652325
since I tend not to do that :-)

jburel:
@dominikl: will you resurrect your PR?

PR #4793, opened 2016-08-22
Codomain per channel db

jburel:
# What this PR does

Allow to save the codomain map context per channel instead of per rendering

# Testing this PR
Check that the scripts are correct.
Nobody has ever used the codomain
If somebody did, the rendering engine cannot start due to a ``ClassCastException``
This has been fixed in another PR 

# Related reading

see https://trello.com/c/9iAe3Sg9/95-lut-support

cc @mtbc

PR #4796, opened 2016-08-23
have metadata service collate delete requests into one

mtbc:
# What this PR does

This PR adjusts how Insight often submits delete requests, collating deletes for many objects into a single request.

# Testing this PR

Tag a couple of hundred images and in Insight multiselect them and click in the right pane to remove tags from all the images. With this PR the unlinking should occur quite quickly. (Unlinking is deletion of a link.)

This PR's changes are more far-reaching: they affect Insight's deletion more widely. Try other bulk deletes, especially if possible mixing objects that can be deleted with ones that can't (e.g., include in multiselect some else's images or tags) to ensure there are no regressions.

# Related reading

http://trac.openmicroscopy.org/ome/ticket/13139

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroMetadataServiceImpl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroMetadataServiceImpl.java
@@ -25,6 +25,7 @@ import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


pwalczysko:
I just did not see any situation where insight would allow you to delete at all if you multiselect things that you can delete as well as things you cannot delete. In such cases (quite rightly I would think) NO deletion from the UI is possible.

Thus, no regressions spotted by now.
Will continue testing once https://github.com/openmicroscopy/openmicroscopy/pull/4793/commits/1572baf547517733dc73c6191e1d98c18ffab165 fix is in

pwalczysko:
The speedup thanks to this PR is as follows:
  - tag 270 images with a tag on eel merge (this PR included)
  - remove the tag from these 270 images in one go
  - takes some 2-3 seconds
  - repeat the same on eel latest (without this PR)
  -  takes around 30 seconds

dominikl:
👍  Code looks good.

pwalczysko:
Checked the deletion in bulk of things which you can delete together with things you cannot delete. 
  - ROIs
  - Tags
  - Attachments
Checked one permissions situation only, the admin annotating your object in read-only group
All works as expected, only the objects which you have permissions for are deleted, the others are not.

mtbc:
Thank you for hunting these down and testing them.

pwalczysko:
Ready to merge FMPOV.

joshmoore:
Strategy makes sense to me.

PR #4797, opened 2016-08-23
Ignore stroke width < 1

dominikl:
# What this PR does

Makes sure that the line width used for drawing ROIs is at least 1 px.

# Testing this PR

Draw an ROI; set the line width in the DB to a value < 1; make sure ROI tool doesn't crash.

# Related reading
[[ome-users] OMERO Client crashes when lauching ROI tool](http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-August/006117.html)
The user imported ROIs with stroke width < 1, which made Insight ROI tool practically unusable.
I'm not sure if it's worth fixing in Insight, as it's a problem with the ROIs, but on the other hand, using a 1px default is better than crashing Insight in these cases.


gusferguson:
@dominikl 

Tested using OMERO.insight-5.3.0-m2-1162-597cc31-ice36-b406-mac user-3 read-only-1

Tested using supplied image with ROI line width <1.0 px - no crashing.
Changed line width to 0.5 px - showed as 1.0 px in UI - no crash.

Good to merge.

PR #4798, opened 2016-08-23
in Insight keep the Intensity Results View table sorted

mtbc:
# What this PR does

Insight's "intensity results view" table is essentially unsorted. This PR sorts it sensibly regardless of the order in which ROI results were added to the table.

(Excel export is unaffected: one may easily sort in Excel however one wishes.)

# Testing this PR

Ensure that the intensity results view table continues to work properly and always maintains a sensible row sort order.

# Related reading

http://trac.openmicroscopy.org/ome/ticket/9992

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityResultsView.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/view/IntensityResultsView.java
@@ -45,9 +45,12 @@ import javax.swing.JFileChooser;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JTable;
+import javax.swing.RowSorter;
+import javax.swing.SortOrder;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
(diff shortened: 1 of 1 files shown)


dominikl:
Looks good. Just one point: The default row sorter for the other columns seems to use string/alphabetical sorting (ie for the 'max' column, see screenshot). I wonder if that could be fixed by setting the column class for all columns?

screen shot 2016-08-24 at 11 26 58

mtbc:
It probably could, yes; I hadn't even noticed they were manually sortable! But then we probably want to use the table cell renderers for numbers too and get more right-justification, etc., so maybe worth a bit of UX discussion and a new PR. (For instance, *if* right-justifying numbers, what about channel names?)

dominikl:
I was a bit surprised, too, that adding row sorting for some columns automatically made all rows sortable. Well, seems to be quite an effort, not sure if it's worth it. Is the "Intensity Results View" feature actually used a lot by the users?

mtbc:
Maybe I should make them not sortable then for now.

dominikl:
No, I think it's much better than the random order now and not even having the possibility to chance it. From my point of view, good to merge.

PR #4799, opened 2016-08-23
File filter from B-F

jburel:
# What this PR does

Add try/catch block to handle error when loading supported files

# Testing this PR

Open the importer, 
Make sure the application does not need to crash if the list of
supported file formats cannot be retrieved.
This can be triggered by not having a "readers.txt" file, the file is read by B-F

The "readers.txt" file is in the``formats-api.jar``. This is part of the jars shipped with insight
Open the ``formats-api.jar``, and remove the file.

# Related reading

See https://www.openmicroscopy.org/qa2/qa/feedback/17307/

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroImageServiceImpl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/OmeroImageServiceImpl.java
@@ -1355,17 +1355,24 @@ class OmeroImageServiceImpl
 	 */
 	public FileFilter[] getSupportedFileFormats()
 	{
-		if (filters != null) return filters;
-		//improve that code.
-		ImageReader reader = new ImageReader();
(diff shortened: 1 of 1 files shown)


dominikl:
Tested with merge build, only removed the `readers.txt` from the `formats-api.jar` file. Works like expected, Insight/Importer doesn't crash, only indicates that the format is not recognized, when trying to import a file (`loci.formats.UnknownFormatException: Unknown file format` in the log).
👍

dominikl:
I just noticed, when the import fails, you have to "force quit" the importer when you want to quit Insight. Not related to this PR, will open ticket for it.

jburel:
@sbesson @joshmoore: could one you merge this PR?

joshmoore:
Sorry, missed this in standup. Merging.

jburel:
Thanks

PR #4802, opened 2016-08-23
Tagging bugfixes

dominikl:
# What this PR does

If you create a new Tag or TagSet while browsing another user's Tags, the newly created Tag/TagSet might not be shown, because it gets listed within the logged in user's branch (which might not be expanded), see example screenshot (logged in as user-6, browsing user-3 tags). The create new Tag/TagSet menus should therefore be disabled in that case.

screen shot 2016-08-23 at 15 18 06

(Also fixes two `ClassCastExceptions` when creating Tags/TagSets which probably have been introduced by  #4734)

# Testing this PR

Make sure that you *cannot* create a new Tag/TagSet while browsing through another user's Tags (another user's tag, tagset, object is selected), check the right click context menu and toolbar menu. 
Make sure you *can* create a new Tag/TagSet while you're browsing your own tags. 

# Related reading

https://trello.com/c/WN0W4D3l/69-minor-tag-bugs-insight

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/actions/CreateTopContainerAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/actions/CreateTopContainerAction.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2014 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


gusferguson:
@dominikl 

Tested using OMERO.insight-5.3.0-m2-1162-597cc31-ice36-b406-mac user-3 user-6 read-annotate-1 and read-only-1

Behaves as expected.
Good to merge.

jburel:
Good catch @dominikl for ClassCastException
@pwalczysko: maybe a round of testing of the loading of data/annotations is required to see if we have few more of them

dominikl:
I'm just trying to get  #4654 'Load annotations on demand' working again. Maybe we could wait with a general annotation test round until that PR is also in the build, that would avoid testing the same scenarios again.

PR #4803, opened 2016-08-23
change terminator method name to "close"

mtbc:
# What this PR does

This PR deprecates the offensively named `seppuku` method in the Python gateway replacing it with `close` and not providing its `softclose` option.

# Testing this PR

Integration testing suffices, see https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest.test_connection/TestConnectionMethods/.

# Related reading

https://trello.com/c/bS7vTB10/114-rfe-replace-conn-seppuku-with-less-offensive-term-in-code-base

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1748,7 +1748,7 @@ class _BlitzGateway (object):
     def seppuku(self, softclose=False):  # pragma: no cover
         """
         Terminates connection with killSession(). If softclose is False, the
-        session is really terminate disregarding its connection refcount.
+        session is really terminated disregarding its connection refcount.
 
         :param softclose:   Boolean
 
@@ -1762,7 +1762,7 @@ class _BlitzGateway (object):
     def terminateAllClients(self, softclose=False):  # pragma: no cover
         """
         Terminates connection with killSession(). If softclose is False, the
-        session is really terminate disregarding its connection refcount.
+        session is really terminated disregarding its connection refcount.
 
         :param softclose:   Boolean
         """
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1757,15 +1757,6 @@ class _BlitzGateway (object):
         """
         warnings.warn("Deprecated. Use terminateAllClients",
                       DeprecationWarning)
-        self.terminateAllClients(softclose)
-
-    def terminateAllClients(self, softclose=False):  # pragma: no cover
-        """
-        Terminates connection with killSession(). If softclose is False, the
-        session is really terminated disregarding its connection refcount.
-
-        :param softclose:   Boolean
-        """
         self._connected = False
         oldC = self.c
         if oldC is not None:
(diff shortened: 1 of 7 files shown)


components/tools/OmeroPy/test/integration/gatewaytest/test_connection.py
```diff
@@ -77,6 +78,42 @@ def g2_getTestImage():
...
+        g2.terminateAllClients(softclose=True)
+        pytest.raises(Ice.ConnectionLostException, g2_getTestImage)
+        assert gatewaywrapper.getTestImage() is not None
+        g2 = gatewaywrapper.gateway.clone()
+        assert g2.connect(gatewaywrapper.gateway._sessionUuid)
+        assert gatewaywrapper.getTestImage() is not None
+        assert g2_getTestImage() is not None
+        g2.terminateAllClients(softclose=False)
+        pytest.raises(Ice.ConnectionLostException, g2_getTestImage)
+        pytest.raises(Ice.ObjectNotExistException, gatewaywrapper.getTestImage)
```
will-moore: Not sure I understand why this is an ObjectNotExistException but everywhere else we get an Ice. ConnectionLostException (including for the same call above)?

mtbc: I have no idea, I just copied the other test. :smiley:

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1751,6 +1751,20 @@ def seppuku(self, softclose=False):  # pragma: no cover
...
+        Our apologies for any offense caused by this previous method name.
+        """
+        warnings.warn("Deprecated. Use terminateAllClients",
+                      DeprecationWarning)
+        self.terminateAllClients(softclose)
+
+    def terminateAllClients(self, softclose=False):  # pragma: no cover
+        """
+        Terminates connection with killSession(). If softclose is False, the
+        session is really terminate disregarding its connection refcount.
```
will-moore: termintate -> terminated
Actually, this whole doc string could do with clarifying as I've never really understood what's going on here.
Even now, looking at the code, the logic is not really clear to me.
```terminateAllClients()``` seems a little strong. What about ```terminateClients()``` since if softClose is True then you don't terminate all?

mtbc: @joshmoore, thoughts? (I can certainly rename, but as for a clearer docstring...)

joshmoore: I don't understand softClose either. Why not just use closeSession in that case?! Ignoring that for a second, the functionality is: "lower reference count for each client joined to this session". This should really just be done server side as `omero.cmd.TerminateRegardlessOfSessionCount` (or some better name)

components/tools/OmeroPy/test/integration/gatewaytest/test_connection.py
joshmoore: @will-moore : I would assume it's some error handling in the wrappers.

components/tools/OmeroPy/src/omero/gateway/__init__.py
mtbc: Hmm, looks like @will-moore added the option and @cneves its implementation.

joshmoore:
I apologize, I've confused the issue. I inverted the meaning of `softClose` in previous discussions. Apparently "soft close" meant "decrement the reference count"

mtbc:
So this PR as-is describes things correctly?

joshmoore:
No, currently `terminateAllClients` only reduces the reference count by one.

mtbc:
Even though `terminateAllClients` copies the `softclose == False` path? Will chat tomorrow to make sure I understand! :smiley:

will-moore:
Looks good. Getting rid of softClose makes sense and renaming is nicer. Thanks.
Good to merge from me.

jburel:
Sorry this PR should have been merged before to avoid conflict cc @chris-allan

PR #4814, opened 2016-08-30
have OmeroReader populate channel colors

mtbc:
# What this PR does

The OMERO reader is a Bio-Formats reader interface to images on the OMERO server. This PR has it read the channel colors from the server and write them to the underlying metadata store so as to be available for reading.

# Testing this PR

I would expect that looking at the code would probably do, I don't think we have any code that uses it and it's a simple change.

I use it in https://github.com/mtbc/omero-downloader and I can show you how to use that and note here the round-trip workflow that this PR improves but you'd have to be extra keen to want to try. Or, just set a channel color for some OMERO image and try to read it from an OmeroReader instance. I could have tried to figure out how to add a corresponding integration test but we don't seem to have any others for it.

### DIFF
--- a/components/blitz/src/loci/ome/io/OmeroReader.java
+++ b/components/blitz/src/loci/ome/io/OmeroReader.java
@@ -55,6 +55,7 @@ import omero.RString;
 import omero.RTime;
 import omero.ServerError;
 import omero.api.IAdminPrx;
+import omero.api.IPixelsPrx;
 import omero.api.RawPixelsStorePrx;
 import omero.api.RoiOptions;
 import omero.api.RoiResult;
@@ -320,7 +321,8 @@ public class OmeroReader extends FormatReader {
 
             long pixelsId = img.getPixels(0).getId().getValue();
 
-            pix = serviceFactory.getPixelsService().retrievePixDescription(pixelsId);
+            final IPixelsPrx iPixels = serviceFactory.getPixelsService();
+            pix = iPixels.retrievePixDescription(pixelsId);
             store.setPixelsId(pixelsId, false);
 
             final int sizeX = pix.getSizeX().getValue();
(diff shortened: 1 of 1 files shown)


jburel:
The ``looktuptable`` string is also present i.e. ``channel.getLookupTable()`` at that level but cannot be set in the store. Something to consider for next model change. cc @sbesson

jburel:
Note that the color that is set in this PR is the color set at import time not the one that the user might have used when viewing the image in the OMERO clients.

mtbc:
This PR will probably be good for users who were hoping that it would have set either the import-time color or the rendering-settings color. :smiley:

jburel:
but they can different that's the problem. we will need an option

mtbc:
Should it use `IPixels.retrieveRndSettings`? Do those always have the same number of wave renderings as there are channels?

I'm wondering if this is a more general issue then. For instance, in the client, if I export OME-TIFF I get the import colors, if I export PNG I get the rendering settings colors; it hadn't occurred to me that users may form a strong a priori expectation!

jburel:
the number of ``channelbinding`` equals the number of channel.
Certainly a general issue to discuss, currently we do not export the rendering settings since it is not part of the ome model 
but most of the time the "import" color will not be set
possible logic: if import color, we use it otherwise use rendering color

mtbc:
Added a simple switch which at least gets the rendering setting color retrieval into the code, in the longer term we could consider an order-of-preference enum list or something.

mtbc:
That would be good. Probably outwith this PR: Don't deprecate until we can say where it's going? :smiley:

jburel:
@mtbc: agree, I also need to check its usage internally

jburel:
card: https://trello.com/c/b8jrapBD/174-omero-reader-out-of-blitz-jar
I am not also not convinced that in its actual form we will keep that file as it is.
While looking at this PR to generate ome-tiff, I had a look at the Exporter 
Since we are seriously considering deprecating that file, the suggested changes are probably not required.
The point was discussed after this PR was opened.
I will vote to close this PR for now

mtbc:
Improvements are not appropriate for OMERO reader at this time.

mtbc:
cf. https://github.com/mtbc/omero-downloader/commit/83aec0dda758c6e05ac36f0a20a9662ecdb2ad5b

PR #4817, opened 2016-09-02
adjust how Java clients have Blitz close sessions

mtbc:
This PR has the client's `closeSession()` also destroy the session via the router as already occurs in Python. It is probably one to keep in a bit to see if any problems are noticed. See https://trello.com/c/NWT3qUma/172-client-del-session-still-alive for more specific discussion.


jburel:
Clarified one point with @joshmoore yesterday.
Test is green.
@chris-allan: anything more noticed during the various upgrades/testing of CP?

chris-allan:
Have made my own tests in addition to those here. Looks good to me.

jburel:
Thanks. Merging.

PR #4818, opened 2016-09-02
Close RawPixelsStore when done

emilroz:
# What this PR does

Close RawPixelsStore before getPixelRange method returns otherwise the service stays open.

# Testing this PR
Connect to the same session uuid using BlitzGateway and omero.client. Check session.activeService() before and after calling getPixelRange() and make sure the RawPixelsStore is not on the list.

# Related reading

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7501,8 +7501,11 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
     def requiresPixelsPyramid(self):
         pixels_id = self._obj.getPrimaryPixels().getId().val
         rp = self._conn.createRawPixelsStore()
-        rp.setPixelsId(pixels_id, True, self._conn.SERVICE_OPTS)
-        return rp.requiresPixelsPyramid()
+        try:
+            rp.setPixelsId(pixels_id, True, self._conn.SERVICE_OPTS)
+            return rp.requiresPixelsPyramid()
+        finally:
+            rp.close()
 
     @assert_pixels
     def getPrimaryPixels(self):
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7487,13 +7487,15 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
 
         pixels_id = self._obj.getPrimaryPixels().getId().val
         rp = self._conn.createRawPixelsStore()
-        rp.setPixelsId(pixels_id, True, self._conn.SERVICE_OPTS)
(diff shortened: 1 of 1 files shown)


joshmoore:
Good find. (Only additional point would be to feel free to always introduce a finally for a close)

Looking fruther, I don't know how often this is used, but it seems awfully expensive remote call considering the [implementation](https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.4/components/romio/src/ome/io/bioformats/BfPixelsWrapper.java) is:

```
    public int getByteWidth() {
        return FormatTools.getBytesPerPixel(reader.getPixelType());
    }
```

or

```
  public static int getBytesPerPixel(int pixelType) {
    switch (pixelType) {
      case INT8:
      case UINT8:
      case BIT:
        return 1;
      case INT16:
      case UINT16:
        return 2;
      case INT32:
      case UINT32:
      case FLOAT:
        return 4;
      case DOUBLE:
        return 8;
    }
```

emilroz:
@joshmoore will change the code to try blocks.

Other methods which forget about closing the RawPixelsStore:
* https://github.com/emilroz/openmicroscopy/blob/5fbf3eba715f7a434aa0a7ad5cfdefd086e9cc21/components/tools/OmeroPy/src/omero/gateway/__init__.py#L7499
* https://github.com/emilroz/openmicroscopy/blob/5fbf3eba715f7a434aa0a7ad5cfdefd086e9cc21/components/tools/OmeroPy/src/omero/gateway/__init__.py#L7754

Shall I fix them in this PR?

joshmoore:
From my point-of-view, definitely. Happy to discuss doing further refactoring (decorator? don't use RPS at all?) if anyone is interested.

emilroz:
It would make sense for the getPixelRange not to use the PixelStore at all.

jburel:
cf.https://github.com/openmicroscopy/openmicroscopy/pull/4818#issuecomment-244462722
@joshmoore: we moved to that code in rendering engine to avoid mismatch between the 2.
(That's where @emilroz got it from)
We can certainly revert that.

will-moore:
Looking at ```getPixelLine()``` I see it calls ```getChannels()``` (twice!) which will init the RenderingEngine and this is not closed.
But this is probably true of many other calls that use ```getChannels()```.
I guess we treat Rendering Engine slightly differently since it's cached as image._re, which the ```RawPixelsStore``` is not, so we need to be more careful about closing ```_re```.
Anyway, closing of the ```RawPixelsStore``` in this PR looks good.

jburel:
The "exporter" should probably be closed in a ``try/finally`` cf. ``fileread`` and ``fileread_gen``

will-moore:
Card for improving docs https://trello.com/c/3vGyIffT/192-blitz-gateway-memory-leak-advice

jburel:
thanks @will-moore

snoopycrimecop:
--rebased-to #4987

PR #4821, opened 2016-09-06
Java integration test fixes

dominikl:
# What this PR does

- Slight improvement of the `testPerformanceAttachFile` integration test. The test now uploads all 25 attachments synchronously and then compares the average to the asynchronous upload of the same files. Before just a single attachment upload was performed and compared to the asynchronous upload. 

**Update**: That still didn't fix the flakiness of the test running on the CI. Therefore I disabled the job (marked it as 'broken').

- Added a delay before disconnect and attempt to re-join the session (as suggested by @mtbc ). This is expected to fail, otherwise it would mean the session has not been closed probably. Without the delay it was sometimes possible to re-join the session (ie. integration job was failing, flaky). 

# Testing this PR

Check that the java merge integration job passes.

### DIFF
--- a/components/tools/OmeroJava/test/integration/ClientUsageTest.java
+++ b/components/tools/OmeroJava/test/integration/ClientUsageTest.java
@@ -165,6 +165,13 @@ public class ClientUsageTest extends AbstractServerTest {
         String session = ec.sessionUuid;
         //delete the active client
         disconnect();
+        
+        try {
+            // wait a bit before trying to join the session
+            Thread.sleep(2000);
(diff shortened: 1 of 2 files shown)


mtbc:
Good to merge, thank you @dominikl.

jburel:
@mtbc: could you cherry-pick adb89f4? 
It will be good to have the same on dev_5_2 cf.
https://github.com/openmicroscopy/openmicroscopy/pull/4834

mtbc:
Sure.

PR #4826, opened 2016-09-08
Cap pytest to below 3.0

ximenesuk:
# What this PR does

This PR caps the  `pytest` version to below `3.0.0` due to the bug at: https://github.com/pytest-dev/pytest/issues/1905 

# Testing this PR

The test previously failing at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-breaking-integration-Python27/26/testReport/OmeroPy.test.integration.clitest/test_download/TestDownload/ should pass.

When the issue above if resolved the commit in this PR should be reverted or otherwise amended.

See: https://trello.com/c/FjmLqCum/709-uncap-pytest

--rebased-to #4848

### DIFF
--- a/components/tools/OmeroFS/setup.py
+++ b/components/tools/OmeroFS/setup.py
@@ -33,4 +33,4 @@ setup(name="OmeroFS",
       package_dir={"": "target"},
       packages=[''],
       cmdclass={'test': PyTest},
-      tests_require=['pytest<3.0.0'])
+      tests_require=['pytest != 3.0.0, != 3.0.1, != 3.0.2'])
--- a/components/tools/OmeroFS/setup.py
+++ b/components/tools/OmeroFS/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
-   Copyright 2008 Glencoe Software, Inc. All rights reserved.
+   Copyright 2008-2016 Glencoe Software, Inc. All rights reserved.
    Use is subject to license terms supplied in LICENSE.txt
 
 """
--- a/components/tools/OmeroFS/setup.py
+++ b/components/tools/OmeroFS/setup.py
@@ -33,4 +33,4 @@ setup(name="OmeroFS",
       package_dir={"": "target"},
       packages=[''],
       cmdclass={'test': PyTest},
-      tests_require=['pytest'])
+      tests_require=['pytest<3.0.0'])
--- a/components/tools/OmeroPy/setup.py
+++ b/components/tools/OmeroPy/setup.py
@@ -67,4 +67,4 @@ setup(
         'omero.gateway': ['pilfonts/*'],
         'omero.gateway.scripts': ['imgs/*']},
     cmdclass={'test': PyTest},
-    tests_require=['pytest<3.0.0'])
+    tests_require=['pytest != 3.0.0, != 3.0.1, != 3.0.2'])
--- a/components/tools/OmeroPy/setup.py
+++ b/components/tools/OmeroPy/setup.py
@@ -67,4 +67,4 @@ setup(
         'omero.gateway': ['pilfonts/*'],
         'omero.gateway.scripts': ['imgs/*']},
     cmdclass={'test': PyTest},
-    tests_require=['pytest'])
+    tests_require=['pytest<3.0.0'])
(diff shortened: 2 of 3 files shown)


mtbc:
Maybe also need a card to help us remember to undo this once 3.0.3 is out.

sbesson:
Agreed. There might be a race condition on project roadmaps and we should definitely watch the `pytest` releases. I would assume this PR could stay on hold for a few weeks and consider to be merged when we reach 5.3.0-m5 if nothing has been released upstream then. /cc @jburel

ximenesuk:
If the tests pass tomorrow then I assume the breaking tag will need to be removed to protect other builds?

ximenesuk:
See: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-breaking-integration-Python27/27/ Relevant tests now passing.

simleo:
To avoid coming back to this in the future we need to change the clause so that only bugged versions are excluded. [The PR that fixes the bug](https://github.com/pytest-dev/pytest/pull/1908) has already been merged and is scheduled for inclusion in the 3.0.3 release. I've checked 3.0.{0,1,2} and they're all bugged, while 2.9.2 is OK. So if we change the clause to:

```
tests_require=['pytest != 3.0.0, != 3.0.1, != 3.0.2']
```

we should be OK now and in the (near) future. AFAIK, [PEP440](https://www.python.org/dev/peps/pep-0440) does not allow to specify the clause in a more compact way.

ximenesuk:
Thanks @simleo, I'll defer to @sbesson whether this PR needs a longer term solution or if it is just a temporary PR until `pytest` is fixed. I may not be here when it is!

sbesson:
Unless there is some work in `breaking` which requires Python tests, I would propose to include it in the merge build to ensure Python tests are run consistently. I hope this is a temporary solution but I would assume we do not have to decide on the inclusion until the next development milestone (5.3.0-m5) /cc @jburel . @simleo: agreed, the only compacter way I can think of would be `pytest >= 3.0.3` once the latter is release (not necessary but sufficient condition).

jburel:
Let's move it out of breaking. It does not hurt to keep this PR opened until the new version of ``pytest`` is release

jburel:
Removed ``--breaking``

simleo:
@sbesson the problem with `pytest >= 3.0.3` is that it would also unnecessarily block `2.*`, forcing an upgrade to the latest version in the short term.

mtbc:
Could we have another version of this on breaking so that build goes green again?

ximenesuk:
@sbesson Is there a way for snoopy to merge it into both breaking and develop?

sbesson:
@ximenesuk @mtbc: It's all part of the semantics of the `breaking` queue. Right now, the push job is using `merge develop --no-ask --reset  -D none -I breaking -E exclude --push develop/breaking/trigger --update-gitmodules` i.e. include no PR except for the ones labelled or commented as breaking. It is possible to change this contract to `merge develop --no-ask --reset  -I breaking --push develop/breaking/trigger --update-gitmodules` which would include all PRs AND the breaking ones. The only downside is that there might be conflicts to resolve. No strong opinion either way. 

The alternative would be to implement @simleo's https://github.com/openmicroscopy/openmicroscopy/pull/4826#issuecomment-245867107 and merge this PR. Leaving the OMERO 5.3 team to decide on the most convenient way

ximenesuk:
Let's leave the merge system as it is! I'll modify the PR inline with @simleo's suggestion and if the tests are okay on Monday the PR can be merged.

ximenesuk:
Commit added to exclude broken versions of `pytest` rather than cap it. If the tests pass then this should be merged to allow breaking to also go green.

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/370/
is green
Merging this PR so breaking become green.
We can always revisit later on.

aleksandra-tarkowska:
--rebased-to #4858

atarkowska:
I am afraid they haven;t fixed that in 3.0.3 as mentioned in https://github.com/openmicroscopy/openmicroscopy/pull/4826#issuecomment-245867107. 

```
05:24:34 ============================= test session starts ==============================
05:24:34 platform linux2 -- Python 2.7.5, pytest-3.0.3, py-1.4.31, pluggy-0.4.0 -- /opt/hudson/workspace/OMERO-
```

I will open a new PR with

```
tests_require = ['pytest<3']
```

PR #4822, opened 2016-09-06
fix SQL syntax errors from database upgrade scripts

mtbc:
# What this PR does

@chris-allan reports syntax errors in our PL/pgSQL code. This PR fixes SQL upgrade scripts since OMERO 4.4 and ensures that upgrade to 5.3 fixes the issues for those who already upgraded.

# Testing this PR

Attempt whichever upgrade paths you like, anything from 4.4 onward to fresh databases. Ensure that the upgrade scripts introduce to the database no functions with the overquoted `%%` and, if you try upgrades from old OMERO *without* this PR, that any such overquoting is fixed by upgrading to this latest database patch 11.

# Related reading

#4201
https://github.com/openmicroscopy/openmicroscopy/pull/4793#discussion_r77590764
https://trello.com/c/2tDzi0xP/114-detect-buggy-sql-scripts
--rebased-to #4836 only https://github.com/openmicroscopy/openmicroscopy/commit/53885287e68258ab3e7d48d578aec60ca76c2e75

PR #4201, opened 2015-09-23
fix typo in 5.0 to 5.1 SQL upgrade character encoding error

mtbc:
Accidentally used .vm file quoting syntax. Now with a wrong DB character encoding, upgrade from 5.0 should give an appropriate error message.

--no-rebase unless 5.1.5

PR #4827, opened 2016-09-08
add patch script for format syntax errors in PL/pgSQL functions

mtbc:
# What this PR does

This PR provides a patch script for correcting PL/pgSQL format string syntax errors reported by @chris-allan.

# Testing this PR

Test that the script fixes for OMERO 5.2 users the mistake also addressed by https://github.com/openmicroscopy/openmicroscopy/commit/53885287e68258ab3e7d48d578aec60ca76c2e75.

# Related reading

#4201 - the problem report
#4822 - the fix for the existing scripts and for OMERO 5.3
https://github.com/openmicroscopy/ome-documentation/pull/1520 - describes this new script
https://trello.com/c/2tDzi0xP/114-detect-buggy-sql-scripts -- related to testing
https://trello.com/c/lF0aMP5I/7-db-upgrade-script-errors -- tracking fix for 5.2

### DIFF
--- a/sql/psql/OMERO5.2__0/OMERO5.2-5.1-format-syntax-patch.sql
+++ b/sql/psql/OMERO5.2__0/OMERO5.2-5.1-format-syntax-patch.sql
@@ -1,0 +1,127 @@
+-- Copyright (C) 2012-4 Glencoe Software, Inc. All rights reserved.
+-- Use is subject to license terms supplied in LICENSE.txt
+--
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation; either version 2 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


mtbc:
Ideas for testing include:

* Make sure that the patched functions become the same as one gets from a fresh database installation.
* Make sure that import still works.

joshmoore:
Strategy with docs & community announcement make sense. I was hoping to find time to script an upgrade path in docker testing across each major version of postgresql but haven't yet.

mtbc:
An easy way to test is simply to upgrade from OMERO 4.4 to 5.1 or 5.2 then have a look at the definition of the affected functions `filesetentry_fileset_index_move` and `filesetjoblink_parent_index_move` before and after running this script rather like @will-moore did in reviewing https://github.com/openmicroscopy/openmicroscopy/pull/4822. You could also dump the whole schema and check for any other `%%` in case I missed them.

mtbc:
The script should refuse to run if the database was not upgraded from 4.4 or if the current database is not 5.1 or 5.2.

ximenesuk:
Everything looks fine after upgrade via gh-4836. `pg_dump -s` doesn't show any other `%%` in the output. Good to merge from me.

rleigh-codelibre:
It looks OK to me from a read through, and the jobs appear to be passing.  Probably fine to merge.

PR #4829, opened 2016-09-08
prevent clients from saving deleted objects

mtbc:
# What this PR does

This causes the update service to throw a validation exception if a client attempts to save an object with an ID that does not exist in the database.

# Testing this PR

See https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.test_iupdate/TestIUpdate/

# Related reading

https://trello.com/c/qWNt9vLN/178-save-deleted-object

### DIFF
--- a/components/server/src/ome/logic/UpdateImpl.java
+++ b/components/server/src/ome/logic/UpdateImpl.java
@@ -1,7 +1,5 @@
 /*
- *   $Ids$
- *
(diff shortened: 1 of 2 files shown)


ximenesuk:
This looks fine and is tested both by the new test and the previously failing one in gh-4830. Good to merge.

PR #4830, opened 2016-09-09
Fix gateway test

ximenesuk:
# What this PR does

This PR fixes a gateway test that failed due to the changes in gh-4829.

It looks like the test had a benign error in it where the created OriginalFile object had its ID set to 0. The changes in the above PR uncovered this. It might be that Carlos intended something else here but with this change the test still checks the same functionality.

# Testing this PR

The integration test should pass.


mtbc:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/363/testReport/OmeroPy.test.integration.gatewaytest.test_get_objects/TestFindObject/ is happy and code fix makes sense. Good to merge.

PR #4831, opened 2016-09-09
OMERO_Matlab:Fix the casting from *unit8 to int8

bramalingam:
# What this PR does

ref:
https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8117

# Testing this PR

1. Test the Matlab toolbox with the code snippet given in the forum post.

2. Check if the test-code runs without any errors.

### DIFF
--- a/components/tools/OmeroM/src/annotations/getOriginalFileContent.m
+++ b/components/tools/OmeroM/src/annotations/getOriginalFileContent.m
@@ -42,7 +42,7 @@ store.setFileId(originalFile.getId().getValue(), context);
 % Read data and cast into int8
 fid = fopen(path, 'w');
 byteArr  = store.read(0,originalFile.getSize().getValue());
-fwrite(fid,byteArr,'*uint8');
+fwrite(fid,byteArr,'int8');
 fclose(fid);
 
 % Close the file store


sbesson:
I wonder if it would make sense here to take advantage of the minimal example provided by the user in the forum post and extend the `WriteData` training example to upload a mask created with `imread`, retrieve it and compare both input and output are equal?

bramalingam:
Yes that makes sense, will add that minimal test to the writeData training example.

jburel:
This is a candidate PR for 5.2.6.  see https://trello.com/c/oWtA5Pod/9-omero-matlab-regression#

imunro:
PR Looks Good to me. FYI I've been doing some testing locally and it looks as if any string starting with '*' is invalid and therefore Matlab (Silently) uses the default which is uint8. Writing test data to a file it seems as if 0-7F hex  are written fine but anything outside this range is written , to the file, as 00.  The results will be unpredictable depending on the original data type.

jburel:
@bramalingam could you rebase this PR to dev_5_2?

bramalingam:
--rebased-to #4849

jburel:
@imunro's comment: see card https://trello.com/c/Wq0xVNHc/191-matlab-invalid-string

PR #4832, opened 2016-09-11
Fix rendering engine leak on missing or malformed request arguments (dev_5_2)

chris-allan:
When certain request arguments are missing or malformed the
`render_image_region` view allows an exception to bubble up the stack
where it is handled by Django.  As the `BlitzGateway` relies on a
`__del__` destructor to clean up the rendering engine stateful service
this can lead to situations where the destructor is not called.
The rendering engine is then leaked for the duration of the session's
lifetime unless it is explicitly cleaned up by re-instantiating a
proxy to the service and the `close` method is called.

The semantics under which a destructor is called when an exception is
thrown are quite complicated and get into Python interpreter internals.
In short, objects with a `__del__` method are not subject to garbage
collection, they are collected using reference counting.  Raising an
exception may artificially keep objects alive due to the stack traceback
containing references to local variables which were in scope at the
point when the exception was thrown.  This of course increases the
reference count.  The traceback is usually globally accessible until
the next exception is thrown.

It is __essential__ that we not allow view methods which use the
rendering engine in __any way__ to throw exceptions which bubble up
the stack to Django.  We __will__ end up leaking rendering engine
stateful services otherwise.

This PR fixes a series of such conditions in the
`render_image_region` view.  An HTTP status code of 400 has been chosen.
Test cases and an OMERO.web integration test library extensions have also
been added to validate this.

References:
 * http://eli.thegreenplace.net/2009/06/12/safely-using-destructors-in-python/
 * https://mail.python.org/pipermail/python-list/2004-June/240474.html
 * http://python-3-patterns-idioms-test.readthedocs.io/en/latest/InitializationAndCleanup.html
 * https://emptysqua.re/blog/pythons-thread-locals-are-weird/

/cc @knabar @emilroz

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -18,7 +18,8 @@ import json
 import omero
 import omero.clients
 
-from django.http import HttpResponse, HttpResponseServerError
+from django.http import HttpResponse, HttpResponseBadRequest, \
+    HttpResponseServerError
 from django.http import HttpResponseRedirect, HttpResponseNotAllowed, Http404
 from django.template import loader as template_loader
 from django.views.decorators.http import require_POST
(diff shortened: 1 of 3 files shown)


will-moore:
Tests are all passing at https://10.0.51.137:8443/job/OMERO-test-integration/2/testReport/OmeroWeb.test.integration.test_rendering/TestRenderImageRegion/

I don't think there's too much to add to discussion at https://github.com/openmicroscopy/openmicroscopy/issues/4833.

Some documentation on the test methods would be great.

jburel:
Thanks for the documentation.

PR #4708, opened 2016-06-14
Webgateway api2

will-moore:
This is a subset of webgateway api methods explored in https://github.com/openmicroscopy/openmicroscopy/pull/4572.

User guide for the API which covers all urls added in this PR can be found at https://github.com/will-moore/design/blob/webgateway_json_api/webgateway_json_api.md

I have also added an example test script at
https://github.com/will-moore/openmicroscopy/blob/webgateway_api2/examples/Training/python/Json_Api/Login.py
which can be used for testing this PR.

Browse API at http://10.0.51.145/web/api/

This PR uses `omero-marshal` which is included in the build.
We also add /version/ into the url to allow versioning, E.g.

```
/api/v1/m/projects/
```

The supported versions are specified in settings `API_VERSIONS = ['0.1']` 
Details of our versioning strategy are to be decided.
This PR splits the querying and marshalling into api_query.py and api_marshal.py respectively.

New tests are under OmeroWeb/test/integration/test_api_*
I have ported some integration tests from test_tree.py, and turned them into Django session-based tests to test the actual json api, not just the underlying querying and marshalling. These tests also use `omero-marshal`.

TODO / discuss
- [x] OmeroWeb Tests
- [x] Filter by owner & group
- [x] Single project  `/webgateway/api/v1/m/projects/:id/`
- [x] Test page, limit, normalize, childCount & callback query params in test_api_projects.py
- [x] Test Project creation, update and delete in test_api_projects.py
- [x] Test 404 in test_api_projects.py
- [x] Complete test_api_login.py with success and error handling as in [gist](https://gist.github.com/will-moore/3677206681b1ea4bd966bf838f7ec63d)
- [x] IF we're dropping Django 1.6, can use 1.7 Http imports in omeroweb/http See https://github.com/openmicroscopy/openmicroscopy/pull/4708/commits/440ce90c9bb40efbd36333700872e5f4b05ca092
- [x] Make sure GET, PUT & POST return correct status codes See https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5

Future TODOs:
- Investigate CORS instead of jsonp().
- Register `OMERO.webapi` user agent with QA
- Split 'api' out of webgateway into separate app
- Handle 'normalize' by unloading groups & experimenters?
- Test cases for `_internal` and `_raw` flags in `@jsonp` decorator
- Handle inactive user login. Currently users don't get logged in and can't tell why. 
[...]

PR #4833, opened 2016-09-11
Fix rendering engine leak on missing or malformed request arguments (develop)

chris-allan:
Version of #4832, atop of develop.

When certain request arguments are missing or malformed the
`render_image_region` view allows an exception to bubble up the stack
where it is handled by Django.  As the `BlitzGateway` relies on a
`__del__` destructor to clean up the rendering engine stateful service
this can lead to situations where the destructor is not called.
The rendering engine is then leaked for the duration of the session's
lifetime unless it is explicitly cleaned up by re-instantiating a
proxy to the service and the `close` method is called.

The semantics under which a destructor is called when an exception is
thrown are quite complicated and get into Python interpreter internals.
In short, objects with a `__del__` method are not subject to garbage
collection, they are collected using reference counting.  Raising an
exception may artificially keep objects alive due to the stack traceback
containing references to local variables which were in scope at the
point when the exception was thrown.  This of course increases the
reference count.  The traceback is usually globally accessible until
the next exception is thrown.

It is __essential__ that we not allow view methods which use the
rendering engine in __any way__ to throw exceptions which bubble up
the stack to Django.  We __will__ end up leaking rendering engine
stateful services otherwise.

This PR fixes a series of such conditions in the
`render_image_region` view.  An HTTP status code of 400 has been chosen.
Test cases and an OMERO.web integration test library extensions have
also been added to validate this.

References:
 * http://eli.thegreenplace.net/2009/06/12/safely-using-destructors-in-python/
 * https://mail.python.org/pipermail/python-list/2004-June/240474.html
 * http://python-3-patterns-idioms-test.readthedocs.io/en/latest/InitializationAndCleanup.html
 * https://emptysqua.re/blog/pythons-thread-locals-are-weird/

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -18,7 +18,8 @@ import json
 import omero
 import omero.clients
 
-from django.http import HttpResponse, HttpResponseServerError
+from django.http import HttpResponse, HttpResponseBadRequest, \
+    HttpResponseServerError
 from django.http import HttpResponseRedirect, HttpResponseNotAllowed, Http404
 from django.template import loader as template_loader
 from django.views.decorators.http import require_POST
@@ -829,9 +830,10 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
             x = int(zxyt[1])*w
             y = int(zxyt[2])*h
         except:
-            logger.debug("render_image_region: tile=%s" % tile)
-            logger.debug(traceback.format_exc())
-
+            logger.debug(
+                "render_image_region: tile=%s" % tile, exc_info=True
+            )
+            return HttpResponseBadRequest('malformed tile argument')
     elif region:
         try:
             xywh = region.split(",")
@@ -841,8 +843,12 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
             w = int(xywh[2])
             h = int(xywh[3])
         except:
-            logger.debug("render_image_region: region=%s" % region)
-            logger.debug(traceback.format_exc())
+            logger.debug(
+                "render_image_region: region=%s" % region, exc_info=True
+            )
+            return HttpResponseBadRequest('malformed region argument')
+    else:
+        return HttpResponseBadRequest('tile or region argument required')
 
     # region details in request are used as key for caching.
     jpeg_data = webgateway_cache.getImage(request, server_id, img, z, t)
--- a/components/tools/OmeroWeb/test/integration/test_rendering.py
+++ b/components/tools/OmeroWeb/test/integration/test_rendering.py
@@ -157,10 +157,21 @@ class TestRenderImageRegion(IWebTest):
     """
 
     def assert_no_leaked_rendering_engines(self):
+        """
+        Assert no rendering engine stateful services are left open for the
+        current session.
+        """
         for v in self.client.getSession().activeServices():
             assert 'RenderingEngine' not in v, 'Leaked rendering engine!'
 
     def test_render_image_region_incomplete_request(self):
+        """
+        Either `tile` or `region` is a required request argument to
+        `render_image_region()`.  If `c` is also passed, the rendering
+        engine will also be initialised.  This test ensure that the correct
+        HTTP status code is used and that consequently, any and all
+        rendering engines that were created servicing the request are closed.
+        """
         image_id = self.createTestImage(sizeC=1, session=self.sf).id.val
 
         request_url = reverse(
@@ -177,6 +188,13 @@ class TestRenderImageRegion(IWebTest):
             self.assert_no_leaked_rendering_engines()
 
     def test_render_image_region_malformed_tile_argument(self):
+        """
+        Either `tile` or `region` is a required request argument to
+        `render_image_region()`.  This test ensure that if a malformed `tile`
+        is requested the correct HTTP status code is used and that
+        consequently, any and all rendering engines that were created
+        servicing the request are closed.
+        """
         image_id = self.createTestImage(sizeC=1, session=self.sf).id.val
 
         request_url = reverse(
@@ -193,6 +211,13 @@ class TestRenderImageRegion(IWebTest):
             self.assert_no_leaked_rendering_engines()
 
     def test_render_image_region_malformed_region_argument(self):
+        """
+        Either `tile` or `region` is a required request argument to
+        `render_image_region()`.  This test ensure that if a malformed
+        `region` is requested the correct HTTP status code is used and that
+        consequently, any and all rendering engines that were created
+        servicing the request are closed.
+        """
         image_id = self.createTestImage(sizeC=1, session=self.sf).id.val
 
         request_url = reverse(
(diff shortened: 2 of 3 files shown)


will-moore:
Looks good.
Shame it's not easier to make sure we don't miss these cases elsewhere.
E.g. it seems possible that ```imageMarshal()``` initialises rendering engine and could ```raise``` an Exception that bubbles up to Django.

Would it be possible to have our Django error handling to ensure that BlitzGateway is cleaned up?
What about other stateful services? Presumably it's not just Rendering Engine that could cause memory leaks like this?

chris-allan:
Agreed, it's a shame it's not easier. I don't have any bright ideas to make it easier either, other than refactoring rendering engine use completely. I can't see a way to have Django error handling cover this.

Yes, `imageMarshal()` initialises a rendering engine and yes it could `raise` an exception that bubbles up to Django causing a leak. I think it would have to be a broken image or issues with the server to cause that though.

This issue definitely extends to the other stateful services at a high level. Their service wrappers don't use destructors however. @emilroz found a couple occurrences with the raw pixel store in his `BlitzGateway` usage. They are fixed in #4818.

ximenesuk:
See: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/

will-moore:
@chris-allan Could we not have ```@login_required``` wrap the ```retval = f(request, *args, **kwargs)``` in a ```try/finally``` to always do the conn closing that follows, instead of bubbling up to Django without cleaning up?

chris-allan:
I'm not sure exactly what you're suggesting we put in such a `try/finally`.

At a minimum you would need to have global access to any and all `ImageWrapper` objects that were created in during the execution of the view method. Closing the session, which we always to anyway, is insufficient as these sessions are joined using `detachOnDestroy()`. Until the reference count on the session is zero, the leaked rendering engine stateful services stick around.

will-moore:
@chris-allan This is what I was suggesting https://github.com/will-moore/openmicroscopy/commit/7d77264c9406fc0383e06d2129ac60cfb54479b4

I commented out your ```return``` statements in ```render_image_region()``` so that it throws an exception again, but this is now handled by the ```try/finally``` I added in ```@login_required``` and the services are closed.
The exception then goes on to bubble up to Django (so I needed to add the ```except: pass``` to your tests), but when the tests ```assert_no_leaked_rendering_engines()``` they still find that they have been closed, so the tests still pass.
Does this make sense, or am I missing something here?

chris-allan:
Wow, that is some old code there. Goes back to pre-4.4.0 era stuff that we probably never quite cleaned up properly or just forgot about. We even [still supported Python 2.4](https://www.openmicroscopy.org/community/viewtopic.php?f=11&t=6073) back then. Now that you've brought up what the decorator is doing in some ways I'm even more unsure of the overarching semantics now. In reading more deeply we most definitely have cyclical references between the gateway and the proxy wrappers. This makes the destructor semantics even more complicated and means we probably have Python memory leaks.

While what you're proposing would in principle work for this case, I'm not sure of all the semantics and potential implications. It probably also needs another `try/except` around each `close()` at a minimum. I'd be quite reticent to do this in the 5.2.6 PR (#4832) without exploring the details more thoroughly and adding a **ton** of tests. However, we could certainly try something like will-moore/openmicroscopy@7d77264c9406fc0383e06d2129ac60cfb54479b4 here for 5.3.x and at the same time try to do away with the taint feature and try to remove the destructor, and look at the cyclical references.

The fundamental problem is in the `BlitzGateway` itself. We should probably fix the issue at its core rather than continuing to paper over it in OMERO.web. There are lots of ways to do this far more safely and elegantly now that we're unhampered by Python < 2.7. If we don't, we run the risk of leaking stateful services in OMERO.scripts, people's own standalone scripts which use `BlitzGateway`, etc.

/cc @aleksandra-tarkowska, @joshmoore

will-moore:
I certainly agree that the implications of my suggested change could be quite large. However, we are already attempting to do that conn cleanup on every request anyway.
Also agree that problem is really in Blitz Gateway and it would be nice to fix it there, but tricky to see how to do that without a massive change since we tend to keep stateful services open (image._re in particular) in order not to have to re-open them on multiple calls. E.g. ```webgateway.render_row_plot()``` calls ```image.getChannels()``` 4 times, so if ```image._re``` wasn't cached this would be loaded 4 times (or require a complete refactor). Looked into that as part of https://github.com/openmicroscopy/openmicroscopy/pull/4818.

chris-allan:
You're right, we are trying to clean them up on every request. We're also not trying to close them if there's an exception. I'm not comfortable speculating as to whether that is deliberate or not. I just can't remember. Lacking a definitive statement or test cases enforcing the semantics , which as far as I can tell are nonexistant, and regardless of whether it's a perceived bugfix I'm not comfortable changing such semantics in a point release. We are not the sole consumers of `@login_required()`.

Making this more safe and elegant will require design and refactoring. The current "magic" API is just not fit for purpose. If the fact that we as the core developers screw this stuff up is not evidence of that I don't know what is.

will-moore:
Docs look good, thanks. Good to merge.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#451](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/451/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/451/consoleText) for more details.
Possible conflicts:
  - PR #4687 will-moore 'thumbnail_default_size_96'
    - components/tools/OmeroWeb/omeroweb/webgateway/views.py
  - PR #4708 will-moore 'Webgateway api2'
    - components/tools/OmeroWeb/omeroweb/webgateway/views.py
  - PR #4803 mtbc 'change terminator method name to "close"'
    - components/tools/OmeroWeb/omeroweb/webgateway/views.py

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#453](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/453/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/453/consoleText) for more details.
Possible conflicts:
  - PR #4687 will-moore 'thumbnail_default_size_96'
    - components/tools/OmeroWeb/omeroweb/webgateway/views.py
  - PR #4708 will-moore 'Webgateway api2'
    - components/tools/OmeroWeb/omeroweb/webgateway/views.py
  - PR #4803 mtbc 'change terminator method name to "close"'
    - components/tools/OmeroWeb/omeroweb/webgateway/views.py

jburel:
https://github.com/openmicroscopy/openmicroscopy/pull/4803 has been merged

will-moore:
Although these tests were run locally by @chris-allan and me, it seems they have not been run by ci, since the last run of `OMERO-DEV-merge-integration-web` was on 31st August:  https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-web/351/testReport/OmeroWeb.test.integration.test_rendering/

joshmoore:
See https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_rendering/

PR #4841, opened 2016-09-14
getProjectedThumbnail bugfix

will-moore:
# What this PR does

Fixes error in logs on getProjectedThumbnail.
See https://trello.com/c/gl6sfCdI/146-getprojectedthumbnail-bug

# Testing this PR

1. Need to check web logs:  ```$ tail -f dist/var/log/OMEROweb.log```
2. Pick a Z-stack image, open in full web image viewer, turn Projection ON.
3. Save rendering settings (may need to tweak sliders to enable Save button)
4. Logs should not show any error on Save (see trello card exception).

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7381,6 +7381,8 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
             else:
                 w = w * size[0] / h
                 h = size[0]
+        elif len(size) == 2:
+            w, h = size
         img = img.resize((w, h), Image.NEAREST)
         rv = StringIO()
         img.save(rv, 'jpeg', quality=70)
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
@@ -59,6 +59,12 @@ class TestImage (object):
         thumb.verify()  # Raises if invalid
         assert thumb.format == 'JPEG'
         assert thumb.size == (64, 64)
+        # Projection
+        self.image.setProjection('intmax')
+        pThumb = self.image.getThumbnail()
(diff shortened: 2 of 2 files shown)


jburel:
could you add a test to reflect the change?

will-moore:
The bug had no effect on user experience (only seen once in nightshade logs) and was a trivial fix. However, writing a test to parse OMERO.web logs is certainly not trivial. Probably not a good use of time. We've already tested that the error is no longer thrown. So the test would only be for regression testing that we don't accidentally introduce a bug that logs an Error. If we consider that such tests are useful then we'd want a strategy that would apply log checking coverage over everything, not just this one tiny fix.

jburel:
We are not talking about the user experience but at a change in an underline method used to generate thumbnail.
For sure, monitoring the log file is certainly not the only way to check if an error occurred.
I would image that a test with a ``try/except`` will work in ``gatewaytest/test_image.py``.

chris-allan:
I don't think you need to parse the logs. You're fixing a legitimate bug where if `len(size) == 1` an exception is thrown due to `w` and `h` not being defined. Set up a fixture that exposes the bug and write an integration test for it.

jburel:
this could be also a good candidate to a rebase to dev_5_2 since
the thumbnail store will not be closed if the error occurs

will-moore:
Added simple test of ```_getProjectedThumbnail()``` which fails without this PR fix.

will-moore:
Travis failed with
```
[nomemoryjava] /home/travis/build/openmicroscopy/openmicroscopy/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java:160: error: cannot access IObject
[nomemoryjava] import omero.model.IObject;
```
flaky? trying again...

will-moore:
--rebased-to #4860

will-moore:
Test is passing at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/335/testReport/OmeroPy.test.integration.gatewaytest.test_image/TestImage/

PR #4844, opened 2016-09-15
Include hyphen syntax in graph command help

ximenesuk:
# What this PR does

This PR adds examples to the help for `delete`, `chgrp` and `chown` that use the hyphen syntax for specifying id ranges. See gh-4707. This form of input is only currently available for the `GraphControl` commands, hence this PR's limited scope.

# Testing this PR

Check that the command line help for the above commands makes sense.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/delete.py
+++ b/components/tools/OmeroPy/src/omero/plugins/delete.py
@@ -30,6 +30,8 @@ Examples:
 
     # Delete three images and two datasets including their contents
     omero delete  omero delete Image:101,102,103 Dataset:201,202
+    # Delete five images and four datasets including their contents
+    omero delete  omero delete Image:106-110 Dataset:203-205,207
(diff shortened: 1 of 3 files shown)


components/tools/OmeroPy/src/omero/plugins/delete.py
```diff
@@ -30,6 +30,8 @@
 
     # Delete three images and two datasets including their contents
     omero delete  omero delete Image:101,102,103 Dataset:201,202
+    # Delete five images and four datasets including their contents
+    omero delete  omero delete Image:106-110 Dataset:203-205,207
```
mtbc: why does it say "omero delete" twice?

ximenesuk: Why indeed! That typo has been there a long time now. I'll fix.

mtbc:
:+1: Good to merge.

PR #4857, opened 2016-09-21
leak FileSystem Service

jburel:
# What this PR does

Close formatter to avoid leak
Notice the issue when investigating the thumbnail problem reported on https://github.com/openmicroscopy/openmicroscopy/pull/4823
Code review should be enough

### DIFF
--- a/components/romio/src/ome/io/nio/AbstractFileSystemService.java
+++ b/components/romio/src/ome/io/nio/AbstractFileSystemService.java
@@ -115,10 +115,11 @@ public class AbstractFileSystemService {
             remaining /= 1000;
 
             if (remaining > 0) {
-                Formatter formatter = new Formatter();
-                dirno = remaining % 1000;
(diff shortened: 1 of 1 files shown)


mtbc:
Note that these days one may also use,
```java
try (final Formatter formatter = new Formatter()) { ...
```
but this is **good to merge** as-is. :+1:

joshmoore:
...which of course never made it into our templates.

jburel:
@mtbc: do you want to start a list of elements we want to add to our templates on a doc card?
We have comment in passing on PR but we all tend to forget it when the time comes to edit code

mtbc:
Created https://trello.com/c/qBBCrOs6/512-code-template-improvements.

jburel:
@mtbc Thanks for the card, I have also changed the code to match the new template guideline :-)

mtbc:
Code is looking good, thank you for the various improvements.

PR #4862, opened 2016-09-23
PASTING state exception

dominikl:
# What this PR does

Fixes (removes) an exception sometimes being thrown on "paste and save" action.

# Testing this PR

Open at least two images (with 'compatible' renderings settings, i.e. same number of channels) in full image viewer. In Insight main window: Copy the rendering settings of one image. Select multiple images (including the images opened in the full viewer). "Paste and Save". 

# Related reading

https://trac.openmicroscopy.org/ome/ticket/12050

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerComponent.java
@@ -2849,9 +2849,8 @@ class ImViewerComponent
 	 */
 	public void setSettingsToPaste(RndProxyDef rndProxyDef)
 	{
-		if (model.getState() != PASTING)
-			throw new IllegalArgumentException("This method should be " +
-					"invoked in the PASTING state.");
+		if (!model.checkRendSettingsId(rndProxyDef)) 
+		    return;
 		try {
 			model.resetMappingSettings(rndProxyDef);
 			view.resetDefaults();
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerComponent.java
@@ -2849,15 +2849,12 @@ class ImViewerComponent
 	 */
 	public void setSettingsToPaste(RndProxyDef rndProxyDef)
 	{
-		if (!model.checkRendSettingsId(rndProxyDef)) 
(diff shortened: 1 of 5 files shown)


mtbc:
Successfully reproduced the error without this PR.

mtbc:
With this PR, in playing around with opening images and copying rendering settings among them I keep running into dialogs: 
dialog

Does that make sense?

dominikl:
That's a bit worrying... I try to reproduce that.

dominikl:
When playing around with copy/paste and 'save to all' and sometimes get 
```
Failed to join the session Glacier2.PermissionDeniedException
    reason = "Internal error. Please contact your administrator:
               Wrapped Exception: (org.springframework.ldap.CommunicationException):
              minke.openmicroscopy.org:10389; nested exception is javax.naming.CommunicationException: minke.openmicroscopy.org:10389 [Root exception is java.net.ConnectException: Connection refused]"
```
That's really weird, considering getting this error in the Insight logs (not Blitz, etc) connected to eel!

dominikl:
Reverted most of the first commit again. Now basically I just don't throw the exception. Seems to work reasonably well. Still occasionally get the above mentioned error log about some failed LDAP communication with some server I'm not even connected to, but I don't think that has anything to do with this PR.

waxenegger:
I tested this on eel and I did not run into any exceptions any more when copy & pasting. 

I, too, encountered some connection issues as @dominikl described when testing on eel but then again I tested remotely using vpn. Just to make sure that these connection issues have nothing to do with the PR I also tried the changes locally on my machine and everything behaved as intended, and without connection problems this time.

dominikl:
@pwalczysko probably found a bug: Modify rendering settings of one image (in full viewer) copy, paste to dataset. Settings are pasted to all images. Then "Set imported" rendering settings on dataset. Do all actions while full viewer is open. Rendering settings get reset for all images *except* the one open in full viewer.

dominikl:
Can't replicate the issue with today's build any more.

dominikl:
Tested this again. There's still some weirdness occassionally, when you modify an image's rendering settings in full viewer and preview panel at the same time, like thumbnail going black or not getting updated. But that's not an issue of this PR, it only prevents the exception being thrown, and therefore prevents crashing Insight.

Maybe one more review, then could be merged in my opinion.

jburel:
The black thumbnail is related to Optimist lock exception, related to the loading of the LUT. Already reported on https://trello.com/c/9iAe3Sg9/95-lut-support

PR #4866, opened 2016-09-26
Importer ConcurrentModificationExceptions

dominikl:
# What this PR does

Fixes some rare, randomly occurring `ConcurrentModificationException`s in the Importer component.

# Testing this PR

Make sure the Importer component still works. I couldn't find a way to trigger the initial bug, it's a random timing issue, so there's not really a specific test to perform to make sure the bug is fixed.

# Related reading

https://trac.openmicroscopy.org/ome/ticket/13266

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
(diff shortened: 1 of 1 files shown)


jburel:
Did you notice some performance issue when working with a large number of items?

jburel:
Importer works. 
one thing I noticed see screenshot, in one case the thumbnail was not updated (folder)
folder

files
.
I don't think it is related to this PR.

dominikl:
Thanks @jburel . I've not tested it with hundreds of files. I'll do so, to check if there are performance issues.

dominikl:
Imported a folder with 1800 images, noticed no problems. Well, the UI struggles a bit at the end, but same without the PR.

jburel:
The UI struggles mainly due to the number of components to paint
Thanks for the fix

PR #4869, opened 2016-09-28
dataset_thumb_drag_select_fix

will-moore:
# What this PR does
Fixes the drag-select of Dataset thumbnails, allowing you to start the drag from anywhere in the centre panel (don't need to start dragging on a thumbnail).
cc @pwalczysko 

# Testing this PR

1. Drag select Dataset thumbnails with OR without starting on a thumbnail.


waxenegger:
Can confirm that one can drag select anywhere now with height not being zero any more.

PR #4871, opened 2016-09-29
Fixed ClassCastException in ChannelMetadataLoader

dominikl:
# What this PR does

Tiny bug fix (ClassCastException). Noticed that bug when testing something else. Can be triggered by selecting image 42401 (user-3 on eel), which triggers the metadata loading.

# Testing this PR

Select above mentioned image (without PR that already crashed Insight), browse the metadata.


dominikl:
Hm, can't see the bug with latest build, so might be introduced with on my other PRs, closing again. Sorry.

PR #4876, opened 2016-10-03
more customizations for demo 3

aleksandra-tarkowska:
# What this PR does

This PR:
- remove ajax cache to allow caching all urls, especially API
- allow manage and reorder of metadata pane accordion. Any suggestions about var naming, dict

# Testing this PR

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -653,6 +653,19 @@ CUSTOM_SETTINGS_MAPPINGS = {
           " ] or [\"Repository\", {\"viewname\": \"webindex\", "
           "\"query_string\": {\"experimenter\": -1}}, "
           "{\"title\": \"Repo\"}]'``")],
+    "omero.web.ui.metadata_pane":
+        ["METADATA_PANE",
+         ('['
+          '{"name": "tag", "index": "0"},'
+          '{"name": "map", "index": "1"},'
+          '{"name": "table", "index": "2"},'
+          '{"name": "file", "index": "3"},'
+          '{"name": "comment", "index": "4"},'
+          '{"name": "rating", "index": "5"},'
+          '{"name": "other", "index": "6"}'
+          ']'),
+         json.loads,
+         ("Manage Metadata pane accordion.")],
     "omero.web.ui.right_plugins":
         ["RIGHT_PLUGINS",
          ('[["Acquisition",'
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -656,13 +656,13 @@ CUSTOM_SETTINGS_MAPPINGS = {
     "omero.web.ui.metadata_pane":
         ["METADATA_PANE",
          ('['
-          '{"name": "tag", "label": "Tags", "index": "0"},'
-          '{"name": "map", "label": "Key-Value Pairs", "index": "1"},'
-          '{"name": "table", "label": "Tables", "index": "2"},'
-          '{"name": "file", "label": "Attachements", "index": "3"},'
-          '{"name": "comment", "label": "Comments", "index": "4"},'
-          '{"name": "rating", "label": "Ratings", "index": "5"},'
-          '{"name": "other", "label": "Others", "index": "6"}'
+          '{"name": "tag", "label": "Tags", "index": 0},'
+          '{"name": "map", "label": "Key-Value Pairs", "index": 1},'
+          '{"name": "table", "label": "Tables", "index": 2},'
+          '{"name": "file", "label": "Attachements", "index": 3},'
+          '{"name": "comment", "label": "Comments", "index": 4},'
+          '{"name": "rating", "label": "Ratings", "index": 5},'
+          '{"name": "other", "label": "Others", "index": 6}'
           ']'),
          json.loads,
          ("Manage Metadata pane accordion.")],
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -659,7 +659,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
           '{"name": "tag", "label": "Tags", "index": 0},'
           '{"name": "map", "label": "Key-Value Pairs", "index": 1},'
           '{"name": "table", "label": "Tables", "index": 2},'
-          '{"name": "file", "label": "Attachements", "index": 3},'
+          '{"name": "file", "label": "Attachments", "index": 3},'
           '{"name": "comment", "label": "Comments", "index": 4},'
           '{"name": "rating", "label": "Ratings", "index": 5},'
           '{"name": "other", "label": "Others", "index": 6}'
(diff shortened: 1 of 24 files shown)


components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -653,6 +653,19 @@ def leave_none_unset_int(s):
           " ] or [\"Repository\", {\"viewname\": \"webindex\", "
           "\"query_string\": {\"experimenter\": -1}}, "
           "{\"title\": \"Repo\"}]'``")],
+    "omero.web.ui.metadata_pane":
+        ["METADATA_PANE",
```
atarkowska, line 657: this allows fully customize accordion unloading labels

atarkowska: @will-moore @manics @pwalczysko @joshmoore could you have a look if you are ok with that?

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -653,6 +653,19 @@ def leave_none_unset_int(s):
...
+          '{"name": "tag", "label": "Tags", "index": "0"},'
+          '{"name": "map", "label": "Key-Value Pairs", "index": "1"},'
+          '{"name": "table", "label": "Tables", "index": "2"},'
+          '{"name": "file", "label": "Attachements", "index": "3"},'
+          '{"name": "comment", "label": "Comments", "index": "4"},'
+          '{"name": "rating", "label": "Ratings", "index": "5"},'
+          '{"name": "other", "label": "Others", "index": "6"}'
+          ']'),
+         json.loads,
+         ("Manage Metadata pane accrodion.")],
```
will-moore: typo 'accordion'

components/tools/OmeroWeb/omeroweb/settings.py
will-moore: This seems like a nice idea. Is this planned to come into develop for 5.3.0?

atarkowska: yes if we are happy with that config, this PR can be rebased against main line

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
```diff
@@ -719,149 +718,20 @@ <h1 class="can-collapse defclose" data-name="details">
         
     <!-- ANNOTATIONS -->
 
+        <!-- templates are loaded based on config -->
+        {% for t in ome.metadata_pane|dictsort:"index" %}
```
will-moore, line 722: Why do you need an "index" on each item in the list for sorting? The list is already in the correct order, isn't it?

atarkowska: this is true if you set, but for append it is easier to have that extra var, e.g. lets say you want to move `'{"name": "table", "label": "Tables", "index": 2},'` for the very end. You only have to:

```
bin/omero config remove omero.web.ui.metadata_pane '{"name": "table", "label": "Tables", "index": 2},'
bin/omero config append omero.web.ui.metadata_pane '{"name": "table", "label": "Tables", "index": 7},'
```

we do not have `bin/omero append index`

atarkowska: I was also thinking if index shouldn't be 10, 20, 30,... that will help to shovel them in between easier

will-moore: Right, if you don't have indexes them spaced out, 10, 20, 30, etc, then you can't insert one.
I guess it's just a decision as to what's the nicest default indexes to use. E.g can you insert `"1.1"` into `["1", "2"]`?

atarkowska: sorry they should be numbers not strings :) then user can add 1, 1.1, 2

components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -372,9 +372,8 @@ def logout(request, conn=None, **kwargs):
 
 
 ###########################################################################
-@login_required()
-@render_response()
-def load_template(request, menu, conn=None, url=None, **kwargs):
+def _load_template(request, menu, conn=None, url=None, **kwargs):
```
atarkowska, line 375: @will-moore I made a clean up in load_template and now allow to override it. Could you please have a look if that won't break anything?

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -653,6 +653,19 @@ def leave_none_unset_int(s):
           " ] or [\"Repository\", {\"viewname\": \"webindex\", "
           "\"query_string\": {\"experimenter\": -1}}, "
           "{\"title\": \"Repo\"}]'``")],
+    "omero.web.ui.metadata_pane":
```
will-moore, line 656: This should probably be a plural since it's a list. E.g. `metadata_panes or metadata_panels`?

will-moore:
Seems to be working fine (testing locally).
You might want to start with 1 as the first index instead of 0, so you're not forced to use negative numbers to add at the first slot (although this works OK):

```
$ omero config append omero.web.ui.metadata_pane '{"name": "comment", "label": "Things people said", "index": -0.9}'
```

screen shot 2016-10-07 at 14 00 29

joshmoore:
Do we no longer need https://github.com/openmicroscopy/openmicroscopy/pull/3896/files#diff-32b25f7075b9602021940173873c7224R119 then?

atarkowska:
I wasn't sure how to backport that as it conflicts. I only handled webclient part.

joshmoore:
Ok. Added a note to https://trello.com/c/1h5YDICh/205-populate-metadata

components/tools/OmeroWeb/omeroweb/webclient/tree.py
```diff
@@ -1787,7 +1787,7 @@ def marshal_annotations(conn, project_ids=None, dataset_ids=None,
             join fetch ch.details.creationEvent
             join fetch ch.details.owner
             left outer join fetch ch.file as file
-            where %s
+            where %s order by ch.ns
```
atarkowska, line 1790: order of maps is hardcoded by namespace, I found that the most suitable and guarantee the same order across all the studies. Not sure the best solution

joshmoore:
Merging this for the IDR-0.3.0-rc1 tag. Testing took place on the demo3 servers. Thanks all!

atarkowska:
@jburel @will-moore  in this PR I added `omero.web.ui.metadata_pane` which probably is not the best choice for rebase PR to go to the main line. Does it make sense to you if in a main line we'll create `omero.client.ui.metadata.accordion.labels` or similar ? That would need to be applied in insight as well?

will-moore:
The `omero.web.ui.metadata_pane` config is nice in that it allows you to remove panes you don't want. However, it's probably not very easy to add new panes from external apps here.
So, I'm not sure we could document it as a way to add content - Just to rename or remove panes.
Don't know how easy it is to implement the same in Insight? If it's doable, then we could consider supporting that in both clients. If not, then we could just go for the 'labels' only option.

joshmoore:
@aleksandra-tarkowska : do you have a follow-up card for these items?

atarkowska:
it is not really a follow up, but the place to report to https://trello.com/c/Ds2XzT9J/8-demo3-mapr-cache-web-api. For a bigger discussion or RFE put new card in the same column

PR #4887, opened 2016-10-11
Arrow shapes (Insight)

dominikl:
# What this PR does

Corresponding Insight PR to #4690 .
Adds the possibility to specify Arrows as start and/or end of a Line ROI.
# Testing this PR

Draw some lines, add Arrows as MarkerStart/MarkerEnd, compare to the corresponding Web PR.
# Related reading

screen shot 2016-10-12 at 10 25 06

### DIFF
--- a/components/blitz/src/omero/gateway/model/ShapeSettingsData.java
+++ b/components/blitz/src/omero/gateway/model/ShapeSettingsData.java
@@ -30,6 +30,7 @@ import omero.RString;
 import omero.rtypes;
 import omero.model.Length;
 import omero.model.LengthI;
+import omero.model.Line;
 import omero.model.Shape;
 import omero.model.enums.UnitsLength;
 
@@ -455,8 +456,13 @@ public class ShapeSettingsData
      *
      * @return See above.
      */
-    public String getMarkerStart()
-    {
+    public String getMarkerStart() {
+        Shape shape = (Shape) asIObject();
+        if (shape instanceof Line) {
+            Line l = (Line) shape;
+            return l.getMarkerStart() != null ? l.getMarkerStart().getValue()
+                    : "";
+        }
         return "";
     }
 
@@ -467,27 +473,41 @@ public class ShapeSettingsData
      */
     public String getMarkerEnd()
     {
+        Shape shape = (Shape) asIObject();
+        if (shape instanceof Line) {
+            Line l = (Line) shape;
+            return l.getMarkerEnd() != null ? l.getMarkerEnd().getValue()
+                    : "";
+        }
         return "";
     }
 
     /**
-     * Returns the marker start.
+     * Sets the marker start.
      *
      * @param start The value to set.
      */
-    public String setMarkerStart(String start)
+    public void setMarkerStart(String start)
     {
-        return "";
+        Shape shape = (Shape) asIObject();
+        if (shape instanceof Line) {
+            Line l = (Line) shape;
+            l.setMarkerStart(rtypes.rstring(start));
+        }
     }
 
     /**
-     * Returns the marker end.
+     * Sets the marker end.
      *
      * @param end The value to set.
      */
-    public String setMarkerEnd(String end)
+    public void setMarkerEnd(String end)
     {
-        return "";
+        Shape shape = (Shape) asIObject();
+        if (shape instanceof Line) {
+            Line l = (Line) shape;
+            l.setMarkerEnd(rtypes.rstring(end));
+        }
     }
 
     /**
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/FigureTableModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/FigureTableModel.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2007 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -28,9 +28,9 @@ import java.util.List;
 import javax.swing.table.AbstractTableModel;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.jhotdraw.draw.ArrowTip;
 import org.jhotdraw.draw.AttributeKey;
 import org.openmicroscopy.shoola.agents.measurement.MeasurementAgent;
-import org.openmicroscopy.shoola.agents.util.EditorUtil;
 import org.openmicroscopy.shoola.env.data.util.StructuredDataResults;
 import org.openmicroscopy.shoola.util.roi.figures.ROIFigure;
 import org.openmicroscopy.shoola.util.roi.model.annotation.AnnotationKey;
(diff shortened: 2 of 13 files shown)


components/blitz/src/omero/gateway/model/ShapeSettingsData.java
```diff
@@ -467,27 +473,41 @@ public String getMarkerStart()
...
-     * Returns the marker start.
+     * Sets the marker start.
      *
      * @param start The value to set.
      */
-    public String setMarkerStart(String start)
+    public void setMarkerStart(String start)
     {
-        return "";
+        Shape shape = (Shape) asIObject();
```
jburel, line 492: I will add a `null` check for `start`

components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/InputServerStrategy.java
```diff
@@ -526,6 +528,17 @@ private void addShapeSettings(ROIFigure figure, ShapeSettingsData data)
 		FONT_ITALIC.set(figure, data.isFontItalic());
 		FONT_BOLD.set(figure, data.isFontBold());
 		TEXT_COLOR.set(figure, data.getStroke());
+		
+        if (CommonsLangUtils.isNotBlank(data.getMarkerStart())) {
+            if (data.getMarkerStart().equalsIgnoreCase("Arrow"))
```
jburel: `Arrow` should be declared as a static field, in case it changes and it is used in 4 different locations

dominikl:
Is the functionality sufficient for now, or do we need an explicit "arrow" drawing tool (like line, point, etc) in the toolbar?

will-moore:
Looks good. There's a bit of difference in terms of Arrow head size between web and Insight. Don't know which we prefer (or how hard it is to change in Insight)? Pretty easy to change in web if we want to align them?

screen shot 2016-10-18 at 10 42 01

I think an "Arrow" tool might be nice, just to make it a bit easier and to help users discover this feature.
I guess it depends on how hard that would be. Looks like we have enough space in the toolbar for an Arrow icon (ImageJ puts it under a right-click on the line icon).
I think you should still be able to create an Arrow from a line (and vice versa) just as you can now (rather than have a fixed Arrow shape).

dominikl:
Changing the arrow size would be pretty easy in Insight, too. Should that be hard coded or configurable? And yes, I'll try to add an "arrow" drawing tool to the toolbar.

dominikl:
Just tried to change the size of the arrows in Insight, then they slightly change their shape, too. So might be easier to adjust that in OMERO.web.

jburel (changes requested):

components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/FigureTableModel.java
```diff
@@ -164,7 +165,18 @@ else if (AnnotationKeys.FOLDERS.equals(key)) {
 					                value += "; ";
 					        }
 					    }
-					}
+                    } else if (MeasurementAttributes.START_DECORATION
+                            .equals(key)) {
+                        if (value != null) {
```
mtbc: don't need this null check, also below

components/insight/SRC/org/openmicroscopy/shoola/agents/measurement/util/model/FigureTableModel.java
```diff
@@ -219,7 +231,7 @@ public int getColumnCount()
 	public int getRowCount()
 	{
 		return keys.size();
-	}
+	}
```
mtbc: don't need this space :smiley:

components/blitz/src/omero/gateway/model/ShapeSettingsData.java
dominikl: I don't think that's necessary, `rtypes.rstring(start)` will return empty string in case of `start == null`.

jburel: I had memory of an exception be thrown, that's why.

components/insight/SRC/org/openmicroscopy/shoola/util/roi/figures/MeasureArrowFigure.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * org.openmicroscopy.shoola.util.roi.figures.MeasureLineFigure
```
jburel: we have opted to remove that line i.e. line 2 (+this is not correct name)

components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/InputServerStrategy.java
dominikl: Created an Enum for that instead ( ecbf122 ), makes it easier to add more line decorations if needed.

will-moore:
The "Arrow" toolbar button works as expected.
I updated arrow head size in webclient to match Insight.
I noticed a couple of minor issues in Insight. NB: we don't support Arrow heads on PloyLines in web.
- If I create a PolyLine I can't add Arrow heads to it. Was that intended (to match web)?
- However, I can take a Line with Arrows and "break" it into a PolyLine. But when I save this the Arrow heads are lost:

screen shot 2016-10-19 at 11 49 07

screen shot 2016-10-19 at 11 49 18

Another minor issue (unrelated to this PR): If I hit the Delete button to delete a shape (only single shape in ROI) the ROI and Shape get removed from the Manager table. But when I hit Save, they show up again (not deleted). However, if I use right-click -> Delete then they are deleted correctly.

dominikl:
The polyline thing in Insight doesn't work properly anyway. If you save a line ROI and then "break" it and try to save it, you'll get an `org.hibernate.exception.ConstraintViolationException` `constraint [shape_roi_roi_index_key]` exception. Shall we just disable the possibility to "break" line ROIs into polylines? /cc @jburel

will-moore:
I think we should add a PolyLine button to the toolbar and disable the breaking of lines into PolyLines. 
Most users probably think that you can't draw PolyLines in Insight - it's not very intuitive to draw a line and then break it. Also makes it hard to trace an edge with a PolyLine.
We probably want to keep the ability to break PolyLines and Polygons into smaller segments.

jburel:
I agree with @will-moore: remove the breaking of lines into polylines (very few people know you can do that!)

components/blitz/src/omero/gateway/model/ShapeSettingsData.java
jburel: you are correct it will return and empty string

dominikl:
To test last commits : Double click somewhere on a line. The line simply should get selected (like single click), nothing else (before: the line was split into two connected lines (Polyline) at the point of the double click).

components/insight/SRC/org/openmicroscopy/shoola/util/roi/io/InputServerStrategy.java
jburel: great

jburel (approved):

components/insight/SRC/org/openmicroscopy/shoola/util/roi/figures/MeasureLineFigure.java
```diff
@@ -828,7 +831,8 @@ public void setStartPoint(Point2D.Double p)
 	 */
 	public int splitSegment(Point2D.Double split) 
 	{
-		if (isReadOnly() || !interactable) return -1;
+		if (isReadOnly() || !interactable || !allowSplitSegment)
```
jburel: maybe having a private method `canBeModified` or something like that will be more suitable
in case we add another constaint

will-moore:
The line can't be broken now, which is great. But we really need a new PolyLine button on the toolbar so that users can still create PolyLines that way.
I guess that could be done in a separate PR if it's a fair bit of work - but then strictly speaking it should probably be in the same PR as the disabling of line breaking.

dominikl:
I suppose adding a specific polyline drawing tool will be quite some work. Shall I enable the "line split" functionality again, and create a separate PR for the polyline issue, so that this PR can merged?

jburel:
Preferable to be done in another PR. @dominikl could you create a card?

dominikl:
I'll enable the split functionality for Line ROIs again. Created a [trello card](https://trello.com/c/l3IiAu9u/136-roi-tool-bugs-improvements) for the outstanding ROI tool issues.

jburel:
The "breaking line" is back, this will be removed when/if further work is done on MT

PR #4889, opened 2016-10-12
cpp: cmake: Update FindIce for 3.6.3

rleigh-codelibre:
# What this PR does

Add CMake support for Ice 3.6.3 now it's released.  Sync with latest FindIce.cmake from upstream, with 3.6.3 support added (I'll push this upstream later today).
# Testing this PR

Check builds remain green.  If you want to test against Ice 3.6.3 itself, you can do so; it will continue to support older Ice versions as well.

It's needed for building on Windows, but Windows with Ice 3.6 is broken due to the symbol count limit so this is not useful to test at present.
# Related reading

[Ice 3.6.3 release](https://forums.zeroc.com/discussion/46475/ice-ice-e-and-ice-touch-3-6-3-released)
[Upstream change](https://gitlab.kitware.com/rleigh/cmake/tree/ice-3.6.3)

### DIFF
--- a/components/tools/OmeroCpp/cmake/FindIce.cmake
+++ b/components/tools/OmeroCpp/cmake/FindIce.cmake
@@ -1,3 +1,6 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+
 #.rst:
 # FindIce
 # -------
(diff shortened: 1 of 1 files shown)


mtbc:
Good to merge.

rleigh-codelibre:
@jburel Could we merge this please?  There are no blockers, it doesn't depend upon or require the other PR.

PR #4820, opened 2016-09-05
add enumeration values for acquisition mode

mtbc:
# What this PR does

To the OMERO model object mapping adds acquisition modes,
* BrightField
* SweptFieldConfocal
* SPIM

# Testing this PR

The BIOFORMATS-DEV-merge-full-repository job's logs record various "Unknown AcquisitionMode value" warnings. Pick one of the corresponding images and import it into OMERO. Without these changes (including the entailed Bio-Formats changes) the Blitz log will warn that the enumeration does not exist and will set the image's acquisition mode to "Other". With this PR, having run the `enums_update.sql` script or used a build with #4837 included, the correct acquisition mode should be noted among the image's metadata for subsequent imports. (With these changes in OMERO but without running that database script import is as if OMERO hadn't been changed at all.)

# Related reading

https://github.com/openmicroscopy/bioformats/pull/2553
https://trello.com/c/0zNp8HvU/167-enumerations-addition

PR #4891, opened 2016-10-14
Enum test

jburel:
# What this PR does

Add test to check that the enumerations defined in the various xml files under https://github.com/openmicroscopy/openmicroscopy/tree/develop/components/model/resources/mappings
are correctly inserted during the installation process

The test checks that the values read from the files match the ones in the DB.
It uses method from `ITypes`
# Testing this PR

check https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ConfiguratuionServiceTest/testSupportedEnumerations/
# Related reading

Related to https://github.com/openmicroscopy/openmicroscopy/pull/4820

cc @mtbc

### DIFF
--- a/components/tools/OmeroJava/test/integration/ConfigurationServiceTest.java
+++ b/components/tools/OmeroJava/test/integration/ConfigurationServiceTest.java
@@ -14,10 +14,38 @@ import java.util.Set;
 
 import loci.formats.IFormatReader;
 import loci.formats.ImageReader;
+import omero.RString;
 import omero.api.IConfigPrx;
 import omero.api.ITypesPrx;
+import omero.model.AcquisitionMode;
+import omero.model.ArcType;
+import omero.model.Binning;
+import omero.model.ChecksumAlgorithm;
+import omero.model.ContrastMethod;
+import omero.model.Correction;
+import omero.model.DetectorType;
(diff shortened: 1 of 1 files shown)


components/tools/OmeroJava/test/integration/ConfigurationServiceTest.java
```diff
@@ -203,4 +234,191 @@ public void testSupportedFormats() throws Exception {
...
+
+    /**
+     * Returns the enumeration value.
+     *
+     * @param object The enumeration object.
+     * @return See above.
+     */
+    private String getEnumValue(IObject object)
+    {
+        RString value = null;
```
mtbc: For convenience another option might be some ugly trick with `PropertyUtils.getNestedProperty(object, "value.value")`.

jburel: good point, I extended some existing code from insight.
Let me try that

jburel: better that way so we do not have to modify the test when a new enum is added

mtbc:
Good to merge.

PR #4892, opened 2016-10-14
Update bioformats version to 5.2.2

rleigh-codelibre:
# What this PR does

Updates the version of the bioformats component to 5.2.2.  It was using an old 5.2.0-SNAPSHOT version, which seems odd given that there have been three stable releases since then.
# Testing this PR

Check builds are green
I'm not sure what specifically needs checking here; if there were any fundamental API changes then the build would break.  If there are more subtle behavioural changes, that might need testing but I'm not aware of anything specific

--breaking

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -904,7 +904,7 @@ versions.asm=1.5.3
 versions.axis=1.4
 versions.backport=3.1
 versions.batik=1.8pre-jdk6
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
Oops, this was based on a old copy of openmicroscopy I thought was current.  Closing.

PR #4893, opened 2016-10-15
Use Ice 3.6.3 jars from maven repository

rleigh-codelibre:
# What this PR does

Build with the Ice 3.6.3 jars.

Continue to use omero-shares 3.6.2 until it's regenerated and
uploaded.

Continue to use ice-db since it's not provided by the upstream
maven repository.  It's probably not suitable for us to use on
all platforms either, since it's a JNI wrapper and should be
built against the platform Berkely DB.
# Additional changes

Also needs omero-shares updating for 3.6.3 in OME artifactory.  This PR updates make-data.py to generate using the jars from artifactory.
# Testing this PR

Should continue to build with Ice 3.5.1 and Ice 3.6.2.  Will now also build with Ice 3.6.3.  I would suggest testing with all three.

With the move to maven from upstream, it should now be possible to build from arbitrary ice versions so long as they are in artifactory.  The approach we're currently taking by hardcoding the Ice version means there can be a mismatch between the Ice jars and the slice2xxx tool.  If we select the jar version based upon the slice2xxx version, then they are guaranteed to match.

The Ice DB jar is a special case.  They don't provide it in maven and prefer you to build from source on each platform.  This suggests that the approach we're taking here is wrong and maybe always has been.  ice-db may be broken on all platforms except the one it's build for, or at the least non-functional on platforms without the native DB library for the JNI wrapper to use.
# Related reading

1. https://trello.com/c/Mf5R0YeD/207-ice-3-6-3-broke-omero-build
2. https://forums.zeroc.com/discussion/46478/ice-3-6-3-regression-slice2java-generates-invalid-code#latest


aleksandra-tarkowska:
thanks, can confirm job passed https://10.0.51.106:8443/job/OMERO-build/2/consoleFull

atarkowska:
thanks, can confirm jobs pass now on:
- newly deployed devspace with Ice 3.6.3 https://10.0.51.106:8443/job/OMERO-build/2/consoleFull and integration https://10.0.51.106:8443/job/OMERO-test-integration/1/console (where original problem occurred)
- cowfish with Ice 3.5.1 https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-integration-Python27/63/consoleFull
- trout with Ice 3.4.2 https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-integration/854/consoleFull

rleigh-codelibre:
@aleksandra-tarkowska Thanks!  I'll remove the breaking tag since it looks safe to go in the main build.

joshmoore:
Does this continue to build on 3.6.2? I don't think a patch version specificity is ever something we've had before.

rleigh-codelibre:
@joshmoore The breaking and merge builds are all passing, so I believe so.  They are using Ice 3.6.2.

Regarding the specificity, we've always hardcoded the point release jar in etc/omero.properties and provided the jars in lib/repository (now artifactory/maven).  This has always had the possibility of mismatching with the point release of ice installed on the system, and given the diversity of the platforms and ice versions we use, it's almost certain we've been running mismatched builds in the past both on the CI infrastructure and on the team's systems.  But on this occasion there's been an overt incompatibility.

As a followup to this PR, we can relax the point release hardcoding so that we pull the exact version of the Ice jars to match the installed ice.  That would greatly improve the robustness and flexibility of the build, and eliminate this type of failure in the future.  Now we're pulling directly from ZeroC's maven repo, we can pull any arbitrary version with some minor tweaks to the build properties.

atarkowska:
> Does this continue to build on 3.6.2?

@rleigh-codelibre, merge and breaking integration jobs are run on octopus, cowfish and trout, you would need to check what is installed on octopus, alternatively see https://10.0.51.117:8443/job/Trigger/86/console 3.6.2-1.el7

components/server/src/ome/services/sharing/make_data.py
```diff
@@ -92,9 +92,14 @@ def compile(dir=dat):
             stderr=subprocess.PIPE,
             shell=True)
         version = proc.communicate()[1].strip()
-        pat = "%s/ice*%s.jar" % (rep, version)
-        cp = ":".join(glob.glob(pat))
-        javac_cmd = "javac -target 1.7 -cp "
+        # Current Ice version
+        pat1 = "~/.m2/repository/com/zeroc/*/%s/*.jar" % (version)
```
mtbc, line 96: The repository location is locally configurable. Consider something like `mvn help:evaluate -Dexpression=settings.localRepository`?

rleigh-codelibre: We're using ivy rather than maven here, and it's hardcoded in `etc/ivysettings.xml`.  This script is only ever run by hand when ice is updated, so if a nonstandard location is used, it might be easier to tweak the script when running it.  Ideally this script wouldn't even exist!

mtbc: Gah. Thank you for the context.

jburel:
support for 3.6.3 will have to be back-ported to 5.2  but we still must the jar from artifactory i.e. do not include commit with repo change i.e. `zeroc` to `com.zeroc`
So we will need to push the 3.6.3 jars
The usage of com.zeroc implied that the glacier classes were not longer included in `omero_client.jar`.
We cannot make that change in a point release (testing burden, etc.)

jburel:
Discussed today with @sbesson and @rleigh-codelibre 
3.5.1: One sha1 diff in freeze see https://trello.com/c/Mf5R0YeD/207-ice-3-6-3-broke-omero-build (change in build number and manifest). This is probably due to a security rebuild (w/o version changes) done by zeroc at the time and the we did not pick up
Note that the "new version of freeze 3.5.1" is tested daily on develop

This PR can probably be back-ported as it is

jburel:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/432/  
and 
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-matlab/349/ 
are both back to green

jburel:
@rleigh-codelibre: we can now move the jars on artifactory
and remove `simple.repository=ome.experimental`

jburel:
jars have now been moved

rleigh-codelibre:
@jburel I've pushed the change to use the normal repo

jburel:
@rleigh-codelibre: thanks. Waiting for tomorrow's build, and if everything is green, we should be good.

sbesson:
Just realized with @mtbc of  another effect of this PR for downstream consumers of the OMERO artifacts. Since the Ice 3.5 artifacts are no longer defined in our Artifactory and only the Ice 3.6+ artifacts are present on Maven Central, any developer using Ice 3.5 will need to manually the Zeroc Nexus repository in the `repositories` section of their `pom.xml` to resolve the artifacts.

This new requirement should minimally be reflected in the `minimal-omero-client` but also worth adding to the developer documentation.

mtbc:
Probably also worth resurrecting `OMERO-DEV-merge-maven-client`.

jburel:
This will have to be included in 5.2.7 cc @hflynn

hflynn:
Can someone open a PR to fix https://www.openmicroscopy.org/site/support/omero5.2/developers/installation.html and/or https://www.openmicroscopy.org/site/support/omero5.2/developers/build-system.html as appropriate please?

jburel:
@sbesson, @mtbc: Ice 3.5 jars are in our Artifactory
only 3.6.3 are not in our Artifactory
http://artifacts.openmicroscopy.org/artifactory/webapp/browserepo.html?7

Maybe a quick chat tomorrow

sbesson:
@jburel: this PR also modifies the `groupId` of the Ice artifacts. The Ice 3.5 JARs served by our artifactory are under the `zeroc` groupId and will not resolve as per these changes.

jburel:
I was confused by your comment.

jburel:
Discussed this am with @sbesson 
We are not back-porting as it is this PR to 5.2 since this will be a breaking change
ice 3.6.3 jars have been upload to http://artifacts.openmicroscopy.org/
So no documentation changes reguired

PR #4895, opened 2016-10-18
Script dialog bugfix

dominikl:
# What this PR does

Tiny bug fix, which makes sure the "Run script" button is disabled when not all required fields are set.
# Testing this PR

1) Upload a script with a non-optional parameter in a specific grouping , e. g. use example script https://gist.github.com/dominikl/e64b6ef335288177175fc979a988b713 
2) Open the scripting dialog for this script in Insight.
3) Make sure the "Run script" button is disabled when the required field is empty.
# Related reading

[Ticket 13305](https://trac.openmicroscopy.org/ome/ticket/13305)


pwalczysko:
Works fine with the test.py script in the header of this PR. 
Tried also some of our existing scripts, both execution (Batch image export) and deleting the compulsory field content (Channel Offsets). Both cases worked as expected too.

Ready to merge

PR #4896, opened 2016-10-18
Use jinja2 when populating metadata maps

manics:
# What this PR does

The IDR bulk-annotation -> map-annotations process involves building URLs. Previously these were simple identifiers, but now they may include phrases including characters which need to be escaped when used in a URL. This PR now uses the Jinja templating module if it's installed, so standard Jinja2 filters such as `{{ value | urlencode }}` can be used.
# Testing this PR

Current use case: Mapr @aleksandra-tarkowska which creates mapr URLs with spaces.
# Potential issues

This adds a new optional dependency, `jinja2`, to OMERO.py. Should it be made mandatory along with yaml in order to keep things clearer?

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/metadata_utils.py
+++ b/components/tools/OmeroPy/src/omero/util/metadata_utils.py
@@ -30,6 +30,13 @@ from collections import deque
 from omero.constants import namespaces
 import re
 
+# TODO: Make jinja2 mandatory?
+try:
+    import jinja2
+    JINJA2_ENABLED = True
+except ImportError:
+    JINJA2_ENABLED = False
+
 
 # Namespace for Bulk-Annotations configuration files
 NSBULKANNOTATIONSCONFIG = namespaces.NSBULKANNOTATIONS + "/config"
@@ -357,6 +364,10 @@ class KeyValueListTransformer(object):
         """
 
         def valuesub(v, cv):
+            if JINJA2_ENABLED:
+                t = jinja2.Template(cv)
+                return t.render(value=v)
+            # Fallback to previous implementation
             return re.sub("\{\{\s*value\s*\}\}", v, cv)
 
         key = cfg["name"]
--- a/components/tools/OmeroPy/src/omero/util/metadata_utils.py
+++ b/components/tools/OmeroPy/src/omero/util/metadata_utils.py
@@ -33,9 +33,9 @@ import re
 # TODO: Make jinja2 mandatory?
 try:
     import jinja2
-    JINJA2_ENABLED = True
-except ImportError:
-    JINJA2_ENABLED = False
+    JINJA2_MISSING = None
+except ImportError as j2exc:
+    JINJA2_MISSING = j2exc
 
 
 # Namespace for Bulk-Annotations configuration files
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/src/omero/util/metadata_utils.py
```diff
@@ -30,6 +30,13 @@
 from omero.constants import namespaces
 import re
 
+# TODO: Make jinja2 mandatory?
```
atarkowska, line 33: minimum version jinja2.7 http://jinja.pocoo.org/docs/dev/templates/#urlencode

atarkowska: might be good idea to add that as mandatory, otherwise you will ends up with:

```
Key: some URL
Value: https://someurl?key={{ value|urlencode }}
```

atarkowska:
tested via http://10.0.51.150/web/mapr/compound/NIFLUMIC%20ACID/

Thank you for a prompt help, looks good, works as expected

manics:
@aleksandra-tarkowska How's this?

atarkowska:
by running:

```
bash-4.2$ bin/omero metadata populate --context bulkmap --cfg /tmp/idr-metadata/idr0016-wawer-bioactivecompoundprofiling/screenA/idr0016-screenA-bulkmap-config.yml Screen:1251
Using session c8faf254-c2a0-43d5-9549-142f1874f137 (demo@localhost:4064). Idle timeout: 10 min. Current group: demo
...
    raise JINJA2_MISSING
ImportError: No module named jinja2
```

aleksandra-tarkowska:
although urlencode doesnt work anymore

```
idr-20160803=# select * from annotation_mapvalue where value like 'NIFLUMIC%';
 annotation_id |     name      |     value     | index 
---------------+---------------+---------------+-------
       2828971 | Compound Name | NIFLUMIC ACID |     0
(1 row)
```

atarkowska:
```
idr-20160803=# select * from annotation_mapvalue where value like '%NIFLUMIC%';
 annotation_id |       name        |                            value                             | index 
---------------+-------------------+--------------------------------------------------------------+-------
       2828971 | Compound Name     | NIFLUMIC ACID                                                |     0
       2828971 | Compound Name URL | https://www.ncbi.nlm.nih.gov/pccompound?term=NIFLUMIC%20ACID |     1
(2 rows)
```

looks good, good to merge

joshmoore:
The general idea of using a proper templating framework for this makes a lot of sense. I am worried about the TODOs, though. Though it's somewhat contradictory, my gut feeling is that these _shouldn't_ be mandatary at runtime while we _should_ require them for tests to pass. More generally, though, this is perhaps a test of the general requirements/virtualenv improvements: if we are sufficiently confident about our deployment strategy, then we can feel more comfortable about introducing Python libraries (whether or not they are `optional_`)

**tl;dr**: for metadata52, this makes good sense. At the latest when its proposed for mainline though, it will need cleaning up, documentation, etc. @manics : do you want to do anything else here before I merge?

manics:
@joshmoore Nothing to add here.

It would be good to get a final policy on python/web requirements though, e.g. for yaml, jinja2, and anything else, and perhaps a standard way of checking for them in code if they're optional?

atarkowska:
that really sounds to me like `pip install omero-metadata`

manics:
Yes, if we resurrect @joshmoore's https://github.com/openmicroscopy/openmicroscopy/pull/4619 PR then all requirements could be handled at the plugin level.

joshmoore:
Merging along with https://github.com/IDR/idr-metadata/pull/169 for 0.3.0-rc1. We can come back to decoupling `populate_metadata` later.

atarkowska:
--rebased-to #5226

PR #4899, opened 2016-10-18
Thumbnails and plate fixes

atarkowska:
This PR is the same as https://github.com/openmicroscopy/openmicroscopy/pull/4687 and https://github.com/openmicroscopy/openmicroscopy/pull/4718

---
# What this PR does

This changes the default size for webgateway thumbnails from 64 to 96, which
can then use the cached thumbnails from OMERO.
# Testing this PR
- go to: `/webgateway/render_thumbnail/<imageId>/`
- Should get a thumbnail that is 96 long on longest side.
- `webgateway/render_birds_eye_view/<imageId>/` should continue to give size 96 thumbnail as before (no change).
- Dataset thumbnails no-longer have size specified (now use default size). Right-click on thumbnail and "Open image in new tab". Url should be webclient/render_thumbnail/<imageId>/ with no '96' size.
- Do the same for SPW thumbnails.
- Check test is green at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/350/testReport/OmeroWeb.test.integration.test_plategrid/TestPlateGrid/test_get_plate_grid_metadata/
# Related reading

See https://trello.com/c/abPmgATV/90-thumbnail-96-by-default

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/data/icon_thumbnails_underscore.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/data/icon_thumbnails_underscore.html
@@ -27,7 +27,7 @@
                 <!-- we wrap img with <a> so you can right-click -> open link in new tab -->
                 <a href="<%= webindex %><% if(img.shareId) {print( img.shareId + '/')} %>img_detail/<%= img.id %>/">
                     <img alt="image"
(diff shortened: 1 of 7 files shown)


will-moore:
Testing at http://10.0.51.150/web/...
/webgateway/render_thumbnail/<imageId>/ gives size 96 thumb by default.
Plate images have correct url.

NB: This will change the url that thumbnails are loaded for Plates and Datasets. E.g. plates goes from webgateway/render_thumbnail/1230744/96/ -> webgateway/render_thumbnail/1230744/
and similar for Datasets (with /webclient/render_thumbnail/) so might need to regenerate caches for those urls in production.

Don't know if there are any other IDR-specific issues with this PR (or if there's a merge build with tests etc)? but otherwise looks good to merge.

PR #4905, opened 2016-10-25
reduce number of clauses for wildcard search results

mtbc:
# What this PR does

This PR is a stopgap for search tests to make the wildcarding less general.
# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.test_search/TestSearch/ should pass.

### DIFF
--- a/components/tools/OmeroPy/test/integration/test_search.py
+++ b/components/tools/OmeroPy/test/integration/test_search.py
@@ -136,8 +136,8 @@ class TestSearch(lib.ITest):
         assert 5 == len(res)
 
     def _3164_search(self, searcher, runs=10, pause=1):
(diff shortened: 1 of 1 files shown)


will-moore:
Tests are certainly passing now.
But I wonder what caused them to start failing (and only on python 26)?
Good to merge.

joshmoore:
Perhaps the DB not longer being purged? Or more jobs running against the same server?

will-moore:
But it appears to me that the searches are within the groups that are created in the test, so it shouldn't matter how much other data is in the DB?

joshmoore:
Don't think any group filtering is currently taking place at the Lucene level.

PR #4907, opened 2016-10-26
test behavior of deleting orphaned map annotations

mtbc:
# What this PR does

Adds integration testing to verify that in deletion of objects, associated map annotations are deleted if thus orphaned.
# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.delete/HierarchyDeleteTest/ should pass.

### DIFF
--- a/components/tools/OmeroJava/test/integration/delete/HierarchyDeleteTest.java
+++ b/components/tools/OmeroJava/test/integration/delete/HierarchyDeleteTest.java
@@ -39,12 +39,16 @@ import omero.model.Experimenter;
 import omero.model.ExperimenterGroup;
 import omero.model.ExperimenterGroupI;
 import omero.model.Folder;
+import omero.model.FolderAnnotationLink;
+import omero.model.FolderAnnotationLinkI;
 import omero.model.IObject;
 import omero.model.Image;
 import omero.model.ImageAnnotationLink;
 import omero.model.ImageAnnotationLinkI;
 import omero.model.ImageI;
 import omero.model.Instrument;
+import omero.model.MapAnnotation;
+import omero.model.MapAnnotationI;
 import omero.model.Pixels;
 import omero.model.Plate;
 import omero.model.PlateAcquisition;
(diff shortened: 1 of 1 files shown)


pwalczysko:
The two concrete tests added here https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/400/testngreports/integration.delete/HierarchyDeleteTest/testDeleteOrphanedMapAnnotationViaFolders and the one right below it are passing. So is the whole build.
Read the tests and understood.
Ready to merge fmpov.

joshmoore:
I think this leaves me raising the same question @sbesson did elsewhere about whether or not map annotations should in fact be non-shareable. cc: @eleanorwilliams @manics @aleksandra-tarkowska regarding mapr usage.

mtbc:
The deletion machinery can also cope with comments that annotate multiple images given that everything in our model allows them: it does the same kind of orphaning check, though with the difference that it is more eager to delete others' orphaned comments even outside read-write groups. Even if the graphical clients were to encourage treatment of map annotations as non-sharable then if the server doesn't enforce that then it has to cope somehow with their being shared. (Its willingness to let others' map annotations be orphaned in read-annotate groups is separate from this PR's tests.)

mtbc:
(Currently the graph rules treat map annotations similarly to XML annotations.)

joshmoore:
> Even if the graphical clients encourage treatment of map annotations as non-sharable then if the server doesn't enforce that then it has to cope somehow with their being shared.

And I think I'm proposing that they be orphaned and not deleted.

> (Its willingness to let others' map annotations be orphaned in read-annotate groups is separate from this PR's tests.)

How do you mean?

mtbc:
> And I think I'm proposing that they be orphaned and not deleted.

Ah, okay; I had read "non-sharable" as "more dispensable". Can we merge this as is so it at least helps document current behavior and open a card (whether 5.3.0 or IDR) for changing behavior? Probably worth a wider discussion of other annotations like XML ones. Also has larger implications for, e.g., what to leave behind of others' annotations if moving data to another group or downgrading a group to private.

> How do you mean?

This PR doesn't test anything about _others'_ map annotations.

mtbc:
(Note that in the meantime clients can have the map annotation orphans never deleted by setting `excludeType=['MapAnnotation']`; this can also select by namespace.)

mtbc:
Created https://trello.com/c/Gnoi9mTM/141-never-delete-orphaned-map-annotations so that this PR may now be merged.

jburel:
Merging now that the discussion has been captured.

PR #4908, opened 2016-10-27
allow customizing django apps using AppConfig (rebased onto metadata52)

atarkowska:
This is the same as gh-4843 but rebased onto metadata52.

---
# What this PR does

allows to customize app urls regardless of original web app name is and allow loading apps that has no urls.py

if the app urls need to be different then package name, custom app would have to provide `label` value using AppConfig, see https://github.com/openmicroscopy/weberror/pull/9 as an example

```
from django.apps import AppConfig
class MyWebAppConfig(AppConfig):
    name = "module_name"
    label = "custom_label"  # this is used by urls
```
# Testing this PR

To test this PR is required to have a basic understanding of web deployment including nginx and Django.

**testing matrix: Python2.7 Django 1.8.(latest)**
- test custom label
  
  install omeroweb http://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/unix/install-web/install-web-trial.html or use https://github.com/ome/omeroweb-install Do not use development server!
  
  ```
  # App with AppConfig
  pip install "git+git://github.com/aleksandra-tarkowska/omero-mapr.git@maps"
  
  bin/omero config append --set omero.web.apps '"omero_mapr"'
  ```
  
  test urls:
  - http://host/mapr/
# Related reading

Django doc https://docs.djangoproject.com/en/1.8/ref/applications/#application-configuration

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/urls.py
+++ b/components/tools/OmeroWeb/omeroweb/urls.py
@@ -80,8 +80,13 @@ for app in settings.ADDITIONAL_APPS:
         urlmodule = 'omeroweb.%s.urls' % app
     else:
         urlmodule = '%s.urls' % app
-    regex = '^(?i)%s/' % label
-    urlpatterns += patterns('', (regex, include(urlmodule)),)
+    try:
+        __import__(urlmodule)
+    except ImportError:
+        pass
+    else:
+        regex = '^(?i)%s/' % label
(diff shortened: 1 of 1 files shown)


atarkowska:
--rebased-from #4843

will-moore:
This is already merged into develop and is working well there.
Also seems nothing broken at http://10.0.51.193/ where this is deployed.
So, good to merge for me.

PR #4900, opened 2016-10-20
Customize payload on jstree init

atarkowska:
# What this PR does

- allow passing custom payload on jstree init. @will-moore suggested that trick
- move search to GET
- always select root if nothing selected
- and few other minor fixes

PR #4909, opened 2016-10-27
Customize payload on jstree init (rebased onto develop)

atarkowska:
This is the same as gh-4900 but rebased onto develop.

---

# What this PR does

allow passing custom payload on jstree init. @will-moore suggested that trick

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
@@ -390,8 +390,11 @@ $(function() {
             'data' : function(node, callback, payload) {
                 // Get the data for this query
                 if (payload === undefined) {
-                    payload = {};
+                    payload = this.element.data('payload');
+                    // clear data
+                    $.removeData(this.element[0], "payload");
                 }
+                payload = payload || {}
                 // We always use the parent id to fitler. E.g. experimenter id, project id etc.
                 // Exception to this for orphans as in the case of api_images, id is a dataset
                 if (node.hasOwnProperty('data') && node.type != 'orphaned') {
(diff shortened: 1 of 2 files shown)


atarkowska:
--rebased-from #4900

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
```diff
@@ -390,8 +390,11 @@ $(function() {
             'data' : function(node, callback, payload) {
                 // Get the data for this query
                 if (payload === undefined) {
-                    payload = {};
+                    payload = this.element.data('payload');
```
will-moore, line 393: I think a bit more info could be handy to remind us (me) or the use-case here, E.g:
// Check for existing 'payload' data, used to initialise the jsTree

will-moore:
What's the requirement for the `get_bool_or_default` fix?
This makes the queries a bit more relaxed. E.g. ?orphaned=t would previously be evaluated as False but is now True. Same for 'ON', 'T', 'Y' etc.
I guess this is unlikely to lead to any bugs, but I'm not sure it's expected behaviour.

atarkowska:
becuse there is a bug

```
  File "/home/omero/OMERO.py/lib/python/omeroweb/webclient/views.py", line 148, in get_bool_or_default
    if val_raw.lower() == 'true' or int(val_raw) == 1:
ValueError: invalid literal for int() with base 10: 'false'
```

atarkowska:
`toBoolean` is a method in `omero.gateway`. I don't see anything wrong with unifying web and python api. `on` is a valid result from checkbox input field.

will-moore:
To retain the same behaviour and fix that bug in `get_bool_or_default()` something like this should work:

```
    val = request.GET.get(name, default)
    return str(val).lower() in ['true', '1']
```

atarkowska:
`toBoolean` doesn't change any behaviour. It only expands for more options.  Hard-coding them has no sense to me really.

atarkowska:
closing in fav of https://github.com/openmicroscopy/openmicroscopy/pull/4923

PR #4910, opened 2016-10-28
tag removal test wait

will-moore:
# What this PR does

This reverts async removal of tags (deletion of tag links).
Now we wait for tags to be removed before updating tags in web UI.

Also updates tests accordingly.

To test:
 - Select a large number E.g > 100 objects P/D/I.
 - Add tags, then remove them via Tag dialog
 - Check that when tag is removed, it disappears from lists of tags in right panel
 - Also check that removing via (-) on Tag itself still works as expected.
 - Check that tests are passing.

### DIFF
--- a/components/tools/OmeroWeb/test/integration/test_tags.py
+++ b/components/tools/OmeroWeb/test/integration/test_tags.py
@@ -167,7 +167,7 @@ class TestTags(IWebTest):
         _post_response(self.django_client, request_url, data)
         _csrf_post_response(self.django_client, request_url, data)
         # Check that tag is removed - short delay to allow async delete
-        sleep(0.1)
+        sleep(1)
         request_url = reverse("api_annotations")
         data = {'image': img.id.val, 'type': 'tag'}
         data = _get_response_json(self.django_client, request_url, data)
--- a/components/tools/OmeroWeb/test/integration/test_tags.py
+++ b/components/tools/OmeroWeb/test/integration/test_tags.py
@@ -166,8 +166,7 @@ class TestTags(IWebTest):
         }
         _post_response(self.django_client, request_url, data)
         _csrf_post_response(self.django_client, request_url, data)
-        # Check that tag is removed - short delay to allow async delete
-        sleep(1)
+        # Check that tag is removed
         request_url = reverse("api_annotations")
         data = {'image': img.id.val, 'type': 'tag'}
         data = _get_response_json(self.django_client, request_url, data)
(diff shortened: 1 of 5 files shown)


jburel:
restarted travis

components/tools/OmeroWeb/test/integration/test_tags.py
```diff
@@ -167,7 +167,7 @@ def test_add_edit_and_remove_tag(self):
         _post_response(self.django_client, request_url, data)
         _csrf_post_response(self.django_client, request_url, data)
         # Check that tag is removed - short delay to allow async delete
-        sleep(0.1)
+        sleep(1)
```
joshmoore: Slight heads up that we can only handle so many calls to `sleep(second or more)`. Eventually, this would be better handled with a `@retry` (of some form) on the assertion itself.

will-moore: Probably 0.5 seconds would be sufficient.
Elsewhere I do retry, E.g. https://github.com/will-moore/openmicroscopy/blob/2e750cb9a64646e09d7b734059156a6bdd83e9a3/components/tools/OmeroWeb/test/integration/test_tags.py#L236 but still with a sleep. If we retry, how long should we sleep between retries? (and how many retries)?

will-moore:
Incidentally, I just noticed a web UI bug that is due to the async tag removal. After removing tags with the tag dialog, we refresh the tags in right panel, but this can happen during tag removal, showing incomplete tag removal.
Will need to look at this in more detail - either make the call synchronous and show a spinner while waiting for it to return & refresh OR calculate how to update right tags immediately (needs a lot more logic in JavaScript).

components/tools/OmeroWeb/test/integration/test_tags.py
joshmoore: `0.1` x 10 or even 20 seems fine. Blazej had setup a backoff property which can be used to scale up the number of repetitions if that seems worth it.

atarkowska:
Is tag deletion handled via activity? Maybe just check what is in a queue, and don't show them until status complete

will-moore:
Last commit not directly related to this PR, but should fix failing robot test: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/443/robot/Web%20&%20Web/Web_1/Search/Test%20Key%20Word%20Search/

@aleksandra-tarkowska Tag deletion is added to Activities (same as P/D/I deletion) but _not_ tag removal (since this should generally be much quicker). 
I think the better option is to show spinner during synchronous call and update tags when done.
Maybe a ?wait=true by default and ?wait=false on the request to allow both synchronous and async calls.

pwalczysko:
I can see a serious regression in RHP not allowing to remove the tags using RHP at all in case multiselection of objects is being done in central and LHPanes. @will-moore is telling me the regression was introduced by https://github.com/openmicroscopy/openmicroscopy/pull/4942 (?). 

I would wait till this is clarified, as this workflow should surely work.

pwalczysko:
@aleksandra-tarkowska apparently, both https://github.com/openmicroscopy/openmicroscopy/pull/4923 and https://github.com/openmicroscopy/openmicroscopy/pull/4942 are introducing the bug mentioned in the above comment, @will-moore suggests this is because of https://github.com/openmicroscopy/openmicroscopy/pull/4942/files#diff-c07e169e315bebeca32f30117cad4e6cR1793 motive (false for batch annotate) - analogically for the other PR.

pwalczysko:
The [regression](https://github.com/openmicroscopy/openmicroscopy/pull/4910#issuecomment-261908493) is fixed, thank you @aleksandra-tarkowska .

pwalczysko:
Tests are still passing https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_tags/TestTags/.

pwalczysko:
All the other checks in the UI went as expected. Tried 200 images as well as multi-P/Ds and wells.
Ready to merge.

PR #4911, opened 2016-10-28
Java test chown all

pwalczysko:
# What this PR does

See https://trello.com/c/eaTcjBfW/173-cli-chown-hand-over-all-user-s-data. 
This PR (@mtbc's commits) introduces new chown option targetUsers(), which allows the chowner to chown everything any targetUser owns to another user in one step (permissions permitting of course).

General behaviour of chown.targetUsers()
In case the chowner is a group owner / non admin, then the command will not fail even if the targetUser has some data in another group (of which the chowner is not an Owner of). Instead, all the data which the chowner is permitted to transfer will be transferred, all the data which the chowner is not permitted to transfer will stay in the ownership of the targetUser after the command is executed.

Passing more than one targetUser
Further, the chown.targetUsers() method can take more than one parameter (= target User) so that the chown.targetUsers() might act on two or more user at the same time.
In case the chowner does not have permissions to transfer data of any one of the users in the argument of the chown.targetUsers() then the command fails as a whole, and no data are transferred.

Unique constraint violation
During writing of the integration tests for the above workflows, the violation of unique constraint was hit. See https://trello.com/c/jNUwWsnl/720-accommodate-unique-constraints#comment-5811e0bd6a483d0f614509f7. The case is that if Alice tags Bobs image with exaclty the same tag as Bob did tag it before, then the chown of all Alices data to Bob will fail because Bob would end up with two identical links on his image. For managing the fallout of such case (which is not really a desired user workflow, as an identical entity has been linked twice to the same object by two different users), @mtbc has improved the callback error and in case Group Owner is performing the action, the error has explicite pointer to the problematic link. This might be later consumed by clients. Further plan (not in this PR) is also to offer a script severing such superfluos links.

Tests:
All commits of mine are integration tests (java) which construct and test the abovementioned cases.

Give this PR a descriptive title then expand on that here.
# Testing this PR

Make sure that integration tests are passing. Read the tests and the other code for understandability and sensibility.

@jburel @joshmoore


components/blitz/src/omero/cmd/graphs/Chown2I.java
```diff
@@ -157,31 +188,123 @@ public void init(Helper helper) {
...
+        }
+        for (final Class<? extends IObject> queryClass : classesToQuery) {
+            final String queryClassName = queryClass.getName();
+            final String idProperty = graphPathBean.getIdentifierProperty(queryClassName);
+            String hql = "SELECT " + idProperty + " FROM " + queryClassName + " WHERE details.owner.id IN (:owners)";
+            if (acceptableGroupsFrom != null) {
+                hql += " AND details.group.id IN (:groups)";
+            }
+            List<Long> idList = targetObjects.get(queryClassName);
+            if (idList == null) {
```
jburel, line 268: can't multimap be used for targetObjects?

components/blitz/src/omero/gateway/util/Requests.java
```diff
@@ -2223,6 +2246,22 @@ public Chown2Builder toUser(Experimenter user) {
...
+         */
+        public Chown2Builder targetUsers(Long... ids) {
+            return targetUsers(Arrays.asList(ids));
+        }
+
+        /**
+         * @param user the user whose data is to be targeted for this operation, does not overwrite previous calls
+         * @return this builder, for method chaining
+         */
+        public Chown2Builder targetUsers(Experimenter user) {
```
jburel, line 2262: can user be null? if yes, ok to return a NPE?

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,497 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+        /* set up the users and group for this test case 
+         * Note that two pairs of importers (1 and 2) are
+         * necessary, because also the multi-argument for targetUser
+         * (i.e. two users' data will be chowned in one command)
+         * will be tested.*/
+        final EventContext importerTargetUser1, otherImporter1, importerTargetUser2, otherImporter2;
+        final EventContext chowner, recipient;
+        final ExperimenterGroup dataGroup1;
+
+        importerTargetUser1 = newUserAndGroup("rwra--", false);
```
jburel: tests in other permissions context e.g. `rwrw`

components/blitz/src/omero/cmd/graphs/Chown2I.java
mtbc: No, it's an Ice-generated field from the slice files.

components/blitz/src/omero/gateway/util/Requests.java
mtbc: I'd say so, the Javadoc is quite clear that we expect a user! We can add a bunch of explicit not-null notes but would prefer to do so in a separate PR because there are plenty of other such methods.

jburel: do people read the javadoc ;-)
Happy if the checks are added in another PR

components/blitz/src/omero/cmd/graphs/Chown2I.java
jburel: thanks I was not sure.

pwalczysko:
This should address @jburel comment.

pwalczysko:
It would be of help to have this merged, as the new-role context needs the feature from here to get the two properties (targetUser and new-role) tested in combination. 
I have once more revised the feature test and it works fine (in all group types).
Will relist for tomorrow.

jburel:
Trying to go over the various test coverage. Do you have a "graphical" representation somewhere?
I could not find anything. I will try to produce something

jburel:
Some trailing whitespace introduced in the new tests in ``PermissionsTest``

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,571 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+        final List<ImageAnnotationLink> linksOwnToOthersAnnOwnImage2 = new ArrayList<ImageAnnotationLink>();
+        final List<ImageAnnotationLink> linksOwnToOthersAnnOthersImage2 = new ArrayList<ImageAnnotationLink>();
+        final List<ImageAnnotationLink> linksOthersToOwnAnnOwnImage2 = new ArrayList<ImageAnnotationLink>();
+        final List<ImageAnnotationLink> linksOthersToOwnAnnOthersImage2 = new ArrayList<ImageAnnotationLink>();
+
+        /* First pair of importers imports their respective images, note
+         * that importerTargetUser1 is member of 2 groups (dataGroup1 and otherGroup),
+         * so care has to be taken to import an image for this user in 
+         * each of his groups.*/
+        init(importerTargetUser1);
```
jburel: why ``init`` then ``login``? 
I think it is the same user

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,571 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+        final List<TagAnnotation> tagsetOwn = createTagsets(1);
+        init(otherImporter1);
+        final List<TagAnnotation> tagsetOthers = createTagsets(1);
+        init(importerTargetUser1);
+        final List<TagAnnotation> tagsOwn = createTags(2);
+        init(otherImporter1);
+        final List<TagAnnotation> tagsOthers = createTags(1);
+
+        /* group the tagsets to one pot in order to be able to reuse the 
+         * definteLinikingTags method
```
jburel: typo

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
pwalczysko: Yes, it is the same user. This might be too clanky as you say. I tried to make sure in the first init to get the right user, in the second loginUser to get the right group (the group contex must be spelled precisely too). Probably I should have done it in some one-step method ?

pwalczysko:
Further to our conversation about non-unique links after chown error - the line in the test explaining this is https://github.com/openmicroscopy/openmicroscopy/pull/4911/files#diff-53c3c6ca199387e381cd004fe8986924R986 

Edit: the trick there (and that is why this workflow is exposed as not ideal) is that Chown with targetUser can take more than one parameter (i.e. more than one user), which is tested in the test mentioned and linked above. In such case, the two different links from two different users on the same image linking the same annotation would cause an error after chowning data of both of these users to one recipient.

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
jburel: Discussed with @pwalczysko: "tagset" section might be removed. This was in the initial test which became quite complex

pwalczysko:
@jburel : The last four commits should fix all the comments posted and discussed in zoom call. Hope this will add to the simplification effort.

pwalczysko:
Started to work on my own graphic explanation in team/pwalczysko/testChownAllBelongingToUser/...ppt

pwalczysko:
The graphical Power Point explanation for this test is on ``team/pwalczysko/testChownAllBelongingToUser-explanation/testDescription.pptx``

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,508 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
     }
 
     /**
+     * Test a specific case of using {@link Chown2} with targetUsers option, transferring everything
+     * belonging to a targetUser (importerTargetUser) to a recipient completely,
+     * where targetUser's shared annotations, images and tag sets are in a read-annotate group,
```
jburel: minor: tag sets have been removed from the tests

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,508 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+
+        final long dataGroupId1 = importerTargetUser1.groupId;
+        dataGroup1 = new ExperimenterGroupI(dataGroupId1, false);
+        otherImporter1 = newUserInGroup(dataGroup1, false);
+        
+        if (areDataOwnersInOneGroup) {
+            importerTargetUser2 = newUserInGroup(dataGroup1, false);
+            otherImporter2 = newUserInGroup(dataGroup1, false);
+        }else {
+            importerTargetUser2 = newUserAndGroup("rwra--", false);
```
jburel, line 519: we could parametrize the permissions of the other group too. Not in this PR

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,508 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+         * mission might differ from the first pair of users according to
+         * the areDataOwnersInOneGroup variable */
+
+        final boolean users2CanAnnotateOthers = (importerTargetUser2.groupPermissions.toString() == "rwra--"
+                || importerTargetUser2.groupPermissions.toString() == "rwrw--");
+
+        recipient = newUserInGroup(isRecipientInGroup ? dataGroup1 : otherGroup, false);
+
+        /* Add importerTargetUser1 also to "otherGroup" in order to be able
+         * to test the case where a user has data in two different groups */
```
jburel: ``when`` instead of ``where``

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,508 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+        final List<IObject> linksOthersToOthersAnnOtherImage2 = new ArrayList<IObject>();
+        final List<ImageAnnotationLink> linksOwnToOthersAnnOwnImage2 = new ArrayList<ImageAnnotationLink>();
+        final List<ImageAnnotationLink> linksOwnToOthersAnnOthersImage2 = new ArrayList<ImageAnnotationLink>();
+        final List<ImageAnnotationLink> linksOthersToOwnAnnOwnImage2 = new ArrayList<ImageAnnotationLink>();
+        final List<ImageAnnotationLink> linksOthersToOwnAnnOthersImage2 = new ArrayList<ImageAnnotationLink>();
+
+        /* First pair of importers imports their respective images, note
+         * that importerTargetUser1 is member of 2 groups (dataGroup1 and otherGroup),
+         * so care has to be taken to import an image for this user in
+         * each of his groups.*/
```
jburel: should be ``his/her``

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,508 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+                    final ImageAnnotationLink linkOtherImage = (ImageAnnotationLink) annotateImage(otherImage1, (Annotation) annotation);
+                    linksOwnToOthersAnnOthersImage1.add((ImageAnnotationLink) linkOtherImage.proxy());
+                }
+            }
+        }
+
+        loginUser(importerTargetUser2);
+        if (users2CanAnnotateOthers) {
+            for (final IObject annotation : annotationsOthersForTripleLinking2) {
+                if (!(annotation instanceof Roi) & !(annotation instanceof Thumbnail) & !(annotation instanceof RectangleI)) {
```
jburel: should be ``&&`` or maybe cleaner a ``!(a instance of ROI || ...)``

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,508 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+         * which belongs to this other user (otherImporter).
+         * as well as to the first image (image) which belongs to the importerTargetUser
+         * Note that ROI and thumbnail cannot be multiply linked, and thus will
+         * remain just singly linked objects in the annotationsOwnForTripleLinking batch of annotations
+         * Again, do the same sequence of linking for both sets of annotations/images, 1 and 2.*/
+        loginUser(otherImporter1);
+        /* check group permissions */
+        if (users1CanAnnotateOthers) {
+            for (final IObject annotation : annotationsOwnForTripleLinking1) {
+                if (!(annotation instanceof Roi) & !(annotation instanceof Thumbnail) & !(annotation instanceof RectangleI)) {
```
jburel: same

components/tools/OmeroJava/test/integration/chown/PermissionsTest.java
```diff
@@ -467,6 +471,508 @@ public void testChownAnnotationPrivate(boolean isDataOwner, boolean isAdmin, boo
...
+                    linksOthersToOwnAnnOthersImage1.add((ImageAnnotationLink) linkOtherImage.proxy());
+                    final ImageAnnotationLink linkOwnImage = (ImageAnnotationLink) annotateImage(image1, (Annotation) annotation);
+                    linksOthersToOwnAnnOwnImage1.add((ImageAnnotationLink) linkOwnImage.proxy());
+                }
+            }
+        }
+        loginUser(otherImporter2);
+        if (users2CanAnnotateOthers) {
+            for (final IObject annotation : annotationsOwnForTripleLinking2) {
+                if (!(annotation instanceof Roi) & !(annotation instanceof Thumbnail) & !(annotation instanceof RectangleI)) {
```
jburel: same

pwalczysko:
@jburel : this should address your comments.

jburel:
Thanks @pwalczysko 
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/463/ is green

Will merge post tagging of 5.3.0-m7

PR #4917, opened 2016-11-01
have disk usage plugin cope with multiple IDs given for same class

mtbc:
# What this PR does

@joshmoore [found that](https://openmicroscopy.slack.com/archives/general/p1478000835010648) `bin/omero fs usage` ignores all but the last of arguments of the form Class:X Class:Y Class:Z so that for identical classes "Class" only the Z is targeted.

# Testing this PR

Check that the `usage` feature of the `fs` CLI plugin now gives sensible results when the same class is mentioned multiple times on the command line.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -840,10 +840,7 @@ Examples:
                 else:
                     ids = parts[1].split(",")
                     ids = map(long, ids)
-                    if klass in objects:
-                        objects[klass].extend(ids)
-                    else:
-                        objects[klass] = ids
+                    objects.setdefault(klass, []).extend(ids)
             except:
                 raise ValueError("Bad object: ", o)
 
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -840,7 +840,10 @@ Examples:
                 else:
                     ids = parts[1].split(",")
                     ids = map(long, ids)
-                    objects[klass] = ids
+                    if klass in objects:
+                        objects[klass].extend(ids)
+                    else:
+                        objects[klass] = ids
             except:
                 raise ValueError("Bad object: ", o)
 
--- a/components/tools/OmeroPy/src/omero/plugins/fs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/fs.py
@@ -838,8 +838,7 @@ Examples:
                 if '*' in parts[1]:
                     classes.add(klass)
                 else:
-                    ids = parts[1].split(",")
-                    ids = map(long, ids)
+                    ids = [long(id) for id in parts[1].split(",")]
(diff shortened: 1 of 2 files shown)


mtbc:
@pwalczysko: This bug may indicate a gap in `integration/clitest/`.

components/tools/OmeroPy/src/omero/plugins/fs.py
```diff
@@ -840,7 +840,10 @@ def _usage_obj(self, obj):
                 else:
                     ids = parts[1].split(",")
                     ids = map(long, ids)
-                    objects[klass] = ids
+                    if klass in objects:
+                        objects[klass].extend(ids)
+                    else:
+                        objects[klass] = ids
             except:
```
simleo, line 843: The above four lines can be shortened to `objects.setdefault(klass, []).extend(ids)`

simleo (changes requested):
Suggestion for shortening a common Python idiom

mtbc:
For testing there may be useful memories recorded in #3507.

pwalczysko:
Manual test passes. The order of the Classes is not important, repetition in sense `Image:1 Image:2...` is taken into account correctly, so is interspersing like `Project:1 Dataset:20 Project:2`.

pwalczysko:
I will need a bit of time for the integration test, unless this is an urgent priority.

mtbc:
No, things like OMERO 5.3.0-m5 and idr-demo3 are more urgent; if you like I could add a testing card  for this to the QA board. For now your manual testing suffices, thank you.

pwalczysko:
@mtbc : Yes, please, add a card (and me to it). Thank you

mtbc:
@simleo: any suggestions for the nested "if" at https://github.com/mtbc/openmicroscopy/blob/94d75e74/components/tools/OmeroPy/src/omero/cli.py#L1862 ?

simleo:
@mtbc I'd replace lines 1857 to 1871 with:

``` python
    shmap = {}
    for req in skipheads:
        combined = shmap.setdefault(req.startFrom[0], req)
        if combined is req:
            continue
        type, ids = req.targetObjects.items()[0]
        combined.targetObjects.setdefault(type, []).extend(ids)
    rv.extend(shmap.itervalues())
```

I've only tested that with mock objects though. By the way, why is this just extracting a single arbitrary type -> ids mapping from `targetObjects`? Shouldn't it do something like `for type, ids in req.targetObjects.iteritems() [...]`? Same goes for `StartFrom`, why are we ignoring elements beyond the first?

mtbc:
Thank you. Both good questions. I haven't been able to figure out what this code is doing or why I'm afraid!

mtbc:
I'm wussing out of fiddling with that bit, it can be for later work.

joshmoore:
LGTM

mtbc:
Accidentally closed the PR and now I get to wrestle with Travis again.

joshmoore:
Quickly merging lest anything happen again.

PR #4918, opened 2016-11-01
Deprecate enum

jburel:
Noticed when writing following the formats enum Db patch,

# What this PR does
Mark 2 methods in IPixels as deprecated
Remove usage of the methods in insight

# Testing this PR
Check that insight can start.
Enum are loaded at start up.

cc @mtbc


dominikl:
No problems noticed with Insight. Good to merge from my point of view 👍

PR #4914, opened 2016-10-31
Pillow and numpy issue

jburel:
Check that the integration tests are green

one commit will be required to include the training examples in the daily build
This will done when gh-4921 is merged to avoid conflict

PR #4921, opened 2016-11-02
Training

jburel:
# What this PR does

Reorganise the examples so they can be included in the daily build
Fix unit

Other examples have been reviewed as part of https://github.com/openmicroscopy/openmicroscopy/pull/4914


examples/Training/python/Advanced/Create_Image_advanced.py
```diff
@@ -24,8 +28,11 @@
 
 # Configuration
```
will-moore: This is no-longer `#Configuration` here but something like
" Find another image by getting first image in Dataset"

examples/Training/python/Advanced/Create_Image_advanced.py
```diff
@@ -89,17 +96,11 @@ def planeGen():
...
-if physicalSizeY is not None:
-    newPix.setPhysicalSizeY(rdouble(physicalSizeY))
+newPix.setPhysicalSizeY(pixels.getPhysicalSizeY())
 
-physicalSizeZ = pixels.getPhysicalSizeZ()
-if physicalSizeZ is not None:
-    newPix.setPhysicalSizeZ(rdouble(physicalSizeZ))
+newPix.setPhysicalSizeZ(pixels.getPhysicalSizeZ())
 
 conn.getUpdateService().saveObject(newPix)
```
will-moore: Can probably remove empty lines above, making this all one block, up to `newPix = ...`

examples/Training/python/Advanced/Create_Image_advanced.py
```diff
@@ -7,14 +7,18 @@
 # Use is subject to license terms supplied in LICENSE.txt
 #
 
+import sys
+import os
+sys.path.append(os.path.join('..', 'python'))
```
sbesson: Two thoughts to prevent the Python path workaround assuming it is in place for importing `Parse_OMERO_Properties`:
- either use relative imports e.g. `from ..Parse_OMERO_Properties import...`
- or maybe migrate all the `Advanced` examples together with the rest of the other examples

will-moore:
The `Task_Scripts/Write_Data_3.py` and `Task_Scripts/Write_Data_4.py` are meant to be sample OMERO.scripts that can be used as starting points for writing your own scripts. But now these are not OMERO.scripts any more. Either they need to be left as they are or moved somewhere else if we don't want un-testable code there. Best solution would be to actually upload them via CLI and run them as part of the test, but that would be a fair bit harder?

jburel:
We already have some "samples" scripts under "example/ScriptingService" (that are not tested!)
Let's discuss where we want them.
The main problem here is that they are not run or tested on a regular basis and some have not been updated since the units work so wrong examples!
Either we test them or we delete them not much choice in my opinion

examples/Training/python/Advanced/Create_Image_advanced.py
jburel: `from ..Parse_OMERO_Properties import...` does not work with the actual set-up

jburel:
currently the "training" user (who "runs" the examples) is not an admin so we can't upload as official then run the script to test its contents
Better to have a general testing infrastructure for https://github.com/openmicroscopy/omero-example-scripts

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/446/ is green again

examples/Training/python/Advanced/Create_Image_advanced.py
```diff
@@ -11,30 +11,38 @@
...
 
 # Create a connection
 # =================================================================
 conn = BlitzGateway(USERNAME, PASSWORD, host=HOST, port=PORT)
 conn.connect()
 
-
-# Configuration
+print HOST
+print PORT
```
will-moore: Is there a reason to have these prints, or are they just left-over from debugging? Bit cleaner without them.

will-moore:
Output from https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/446/ looks good.
NB: I noticed a bunch of Exceptions from the Java examples, but they have been there for a while.

One more thought: The great thing about having the Training examples running is that the same code is used on the docs pages https://www.openmicroscopy.org/site/support/omero5.2/developers/Python.html etc. 
The `Training/markup.py` script should do this, but it's getting fragile (output from script is deviating more and more from what's on the docs pages.
Would be nice to try and port the changes here to the Python docs page where appropriate.

examples/Training/python/Advanced/Create_Image_advanced.py
jburel: I was debugging.
I will remove them

jburel:
Wrt to doc
Certainly to look at but I won't do it in this PR
I will open a card

jburel:
@will-moore: are you okay with the last commit f6427ca?
Travis is now green

so I can integrate and activate the remaining examples as part of https://github.com/openmicroscopy/openmicroscopy/pull/4914

will-moore:
Yes, all good, thanks

jburel:
Merging so I can enable the other examples

PR #4928, opened 2016-11-07
Clitest

jburel:
# What this PR does

Remove usage of ``tinytest.dv`` in CLI test
Noticed while investigating a test failure on CI

# Testing this PR

Make sure the tests are green

### DIFF
--- a/components/tools/OmeroPy/test/integration/clitest/test_download.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_download.py
@@ -28,6 +28,7 @@ from omero.cli import NonZeroReturnCode
 from test.integration.clitest.cli import CLITest
 from omero.rtypes import rstring
 from omero.model import NamedValue as NV
+from omero.util.temp_files import create_path
 
 
 class TestDownload(CLITest):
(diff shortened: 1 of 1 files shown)


joshmoore:
A pattern used elsewhere is: 

```
fake = create_path(name, "&%s.fake" % append)
```

so that the temp files are cleaned up on exit.

See: https://github.com/jburel/openmicroscopy/blob/5bda8ef85c886d0fcbde1f9bf42c14d9cc02a9b6/components/tests/python/library/__init__.py#L351

will-moore:
Failing test https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/373/testReport/OmeroWeb.test.integration.test_marshal/TestImgDetail/test_image_detail/ tests for a lot of details of tiny.dv.

Elsewhere we I don't think it matters much what image we use. In several places we may be able to use `createTestImage()` instead of importing.
E.g. `test_scripts.py testSpeedOfThumbnailFigure(self)`, where we don't need to have an original file.

jburel:
@joshmoore: Thanks for the example I missed it
@will-moore: in that case, instead of using tiny.dv we may want to look at generating an ome-xml file with metadata as we do in the import test in Java.
That will require some further changes so probably in another PR

jburel:
To clarify I roll back the change made in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tests/python/library/__init__.py
This PR only removes the usage of `tiny.dv` in the `download` test and uses the method suggested by @joshmoore

will-moore:
OK, test is passing again. 
But confusing to understand the history of test failing and passing again when we don't have the corresponding commit history for the PR. 
I would always prefer to keep the commit history. Otherwise we might as well squash every PR into a single commit?
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/375/testReport/OmeroWeb.test.integration.test_marshal/TestImgDetail/test_image_detail/

jburel:
You are right this is a single commit PR and it should have been from the start. I did not realise that a test in web was relying on tiny.dv
As discussed before your comment, I did not want to keep 2 extra commits (one to add/one to remove) in the code history on such a small PR i.e. 3 commits vs 1.

Happy to discuss further but not on the PR.
Merging since the modified tests are passing

PR #4930, opened 2016-11-08
Test api project dataset update

will-moore:
# Tests saving of Project with linked Dataset.

Following discussion at https://github.com/openmicroscopy/omero-marshal/pull/26#pullrequestreview-6765516
These tests highlight the discussed bug: If we load a Project without child Datasets, then update
the Project and save, all Datasets will get unlinked.
To fix this, we need to call ```p.unloadDatasetLinks()```.

Similar tests and logic added for updating of Datasets and Screens.
This now covers all objects that we allow to Save via the json api.

cc @chris-allan @emilroz @joshmoore


joshmoore:
Mostly an informational comment here since I missed the original discussion: with Ice, there's no difference between `None` and `[]` so there is an explicit boolean to say that a particular field is unloaded:

```
      def unloadDatasetLinks(self, current = None):
          self._datasetLinksLoaded = False
          self._datasetLinksSeq = None;
```

I don't know if it's possible, but if there's a way to represent the difference between "empty set" and "missing set" then it would be possible to track the callers' intent.

will-moore:
It may be that we decide that the web api never breaks links between objects when simply updating those objects. In webclient currently we use `/webclient/api/links/` to DELETE or create links between some objects.

will-moore:
Adding an equivalent test for Project Annotation links (instead of Project-Dataset links). This seems to pass without needing to `unloadAnnotationLinks()` since they are already unloaded, unlike Project-Dataset links.

will-moore:
@chris-allan Thoughts on this PR and how to address this issue for web api?
I think unloading the Datasets for a Project when saving is safest, to avoid breaking links.
Then we can provide other ways of creating/breaking links OR a flag to differentiate loaded from unloaded links in future PRs?

chris-allan:
I haven't really had time to devote to a holistic approach to dealing with unloaded OMERO model objects and am unlikely to for at least another week to ten days. Certainly unloading a collection is a solution to this particular problem, is safe and avoids breaking links. However, unloaded objects are also a problem with "member" objects such as enumerations. The Image-Pixels-Channel hierarchy is littered with examples of this. So, we're going to have to come up with an implementation wide solution to the round-trip problem.

will-moore:
It looks like we're not going to be able to support round-tripping for all the objects that we can read, at least for 5.3.0. I think this is OK as the majority of use-cases initially will be read-only.
Marking some objects (E.g. Images) as "Can't Save" will at least allow us to add support for reading them without exposing round-tripping bugs.

will-moore:
As discussed in meeting http://will-moore.github.io/presentations/2017/Web-Api-Update-Jan-2017/, omero-marshal will need to be smarter to handle decoding of unloaded data so that info is not lost on save.
So the 'fix' from this PR will not be needed but the tests will. This PR can be re-opened when needed.

components/tools/OmeroWeb/test/integration/test_api_containers.py
```diff
@@ -410,6 +410,29 @@ def test_screens(self, user1, user_screens):
...
+                             kwargs={'api_version': version,
+                                     'object_id': screen.id.val})
+        save_url = reverse('api_save', kwargs={'api_version': version})
+        # Get Screen, update and save back
+        rsp = _get_response_json(django_client, screen_url, {})
+        screen_json = rsp['data']
+        screen_json['Name'] = 'renamed Screen'
+        _csrf_put_json(django_client, save_url, screen_json)
+
+        # Check screen has been updated and still has child Datasets
```
jburel: typo ``child Plates``

components/tools/OmeroWeb/test/integration/test_api_projects.py
```diff
@@ -28,7 +28,7 @@
 from django.test import Client
 import pytest
 from omero.gateway import BlitzGateway
-from omero.model import ProjectI, DatasetI
+from omero.model import ProjectI, DatasetI, TagAnnotationI
```
jburel: alphabetical order

components/tools/OmeroWeb/test/integration/test_api_projects.py
```diff
@@ -600,6 +600,87 @@ def test_project_update(self, user1):
...
+        project_json['Name'] = 'renamed Project'
+        _csrf_put_json(django_client, save_url, project_json)
+
+        # Check Project has been updated and still has child Datasets
+        proj = conn.getObject('Project', project.id.val)
+        assert proj.getName() == 'renamed Project'
+        assert len(list(proj.listChildren())) == dataset_count
+
+        # Get Dataset, update and save back
+        dataset_url = reverse('api_dataset',
```
jburel: I split this test into 2 
Projects/Datasets
Datasets/Images

components/tools/OmeroWeb/test/integration/test_api_projects.py
```diff
@@ -600,6 +600,87 @@ def test_project_update(self, user1):
...
+        dataset_json = rsp['data']
+        dataset_json['Name'] = 'renamed Dataset'
+        _csrf_put_json(django_client, save_url, dataset_json)
+
+        # Check Dataset has been updated and still has child Images
+        dataset = conn.getObject('Dataset', dataset.id.val)
+        assert dataset.getName() == 'renamed Dataset'
+        assert len(list(dataset.listChildren())) == image_count
+
+    def test_project_tags_update(self, user1):
```
jburel: This should be tested for other types too

will-moore: I assumed that Datasets and Screens behave the same way as Projects when being linked to Tags etc. But if we need to test them all, I can parametrize that test.

jburel: Better to parametrize, we had surprises before

components/tools/OmeroWeb/test/integration/test_api_projects.py
will-moore: I'll use parametrize to split this into separate tests.

components/tools/OmeroWeb/omeroweb/api/views.py
```diff
@@ -700,6 +700,15 @@ def _save_object(self, request, conn, object_json, **kwargs):
         # Therefore we ignore any details for now:
         obj.unloadDetails()
 
+        # Unlink children for Projects, Datasets and Screens to avoid
```
jburel: Probably need a TODO so we can easily find it
and we need a card to review the graph round tripping since we will face it next  round with image and roi saving for example

will-moore:
Card created at https://trello.com/c/DrXcE9cH/29-unloaded-objects-round-tripping

PR #4931, opened 2016-11-09
Histogram (Server)

dominikl:
# What this PR does

Adds a server-side method for generating histogram data.

# Testing this PR

Check the integration test is correct and passes.

# Related reading

[Trello - Histogram](https://trello.com/c/6gKn8FT2/98-histogram)

# ToDo

- `BfPixelsStoreI` implements the interface, too, but I have no idea what this class is or how it should be implemented.
- ~~What to do if a given `PlaneDef` is not `XY`?~~
- ~~Does not support projections, yet.~~ Next round.

/cc @jburel @will-moore

### DIFF
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -331,9 +331,10 @@ module omero {
                  * @param channels the channels to generate the histogram data for
                  * @param binSize the number of the histogram bins (optional, default: 256)
                  * @param plane the plane (optional, default: whole region of first z/t plane)
+                 * @param globalRange use the global minimum/maximum to determine the histogram range, otherwise use plane minimum/maximum value
                  * @return See above.
                  **/
-                idempotent IntegerIntegerArrayMap getHistogram(IntegerArray channels, int binSize, omero::romio::PlaneDef plane) throws ServerError;
+                idempotent IntegerIntegerArrayMap getHistogram(IntegerArray channels, int binSize, bool globalRange, omero::romio::PlaneDef plane) throws ServerError;
                 
                 /**
                  * Returns the byte width for the pixel store.
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -329,7 +329,7 @@ module omero {
 				/**
                  * Retrieves the histogram data for the specified plane and channels
                  * @param channels the channels to generate the histogram data for
-                 * @param binSize the size of the histogram bins
+                 * @param binSize the number of the histogram bins (optional, default: 256)
                  * @param plane the plane
                  * @return See above.
                  **/
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -11,8 +11,8 @@
 
 #include <omero/ModelF.ice>
 #include <omero/Collections.ice>
-#include <omero/api/PyramidService.ice>
 #include <omero/ROMIO.ice>
+#include <omero/api/PyramidService.ice>
 
 module omero {
 
@@ -327,15 +327,13 @@ module omero {
                 idempotent void setTimepoint(Ice::ByteSeq buf, int t) throws ServerError;
 
 				/**
-                 * Retrieves the histogram data for specified plane and channels
+                 * Retrieves the histogram data for the specified plane and channels
                  * @param channels the channels to generate the histogram data for
                  * @param binSize the size of the histogram bins
-                 * @param renderingDef the rendering definition
-                 * @param projectionDef the projection definition (optional
-                 * @param planeDef the plane definition
+                 * @param plane the plane
                  * @return See above.
                  **/
-                idempotent LongIntegerArrayMap getHistogram(IntegerList channels, int binSize, omero::model::RenderingDef renderingDef, omero::model::ProjectionDef projectionDef, omero::romio::PlaneDef planeDef) throws ServerError;
+                idempotent IntegerIntegerArrayMap getHistogram(IntegerArray channels, int binSize, omero::romio::PlaneDef plane) throws ServerError;
                 
                 /**
                  * Returns the byte width for the pixel store.
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -326,7 +326,7 @@ module omero {
                  **/
                 idempotent void setTimepoint(Ice::ByteSeq buf, int t) throws ServerError;
 
-				/**
+                /**
                  * Retrieves the histogram data for the specified plane and channels
                  * @param channels the channels to generate the histogram data for
                  * @param binSize the number of the histogram bins (optional, default: 256)
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -12,6 +12,7 @@
 #include <omero/ModelF.ice>
 #include <omero/Collections.ice>
 #include <omero/api/PyramidService.ice>
+#include <omero/ROMIO.ice>
 
 module omero {
 
@@ -325,6 +326,17 @@ module omero {
                  **/
                 idempotent void setTimepoint(Ice::ByteSeq buf, int t) throws ServerError;
 
+				/**
+                 * Retrieves the histogram data for specified plane and channels
+                 * @param channels the channels to generate the histogram data for
+                 * @param binSize the size of the histogram bins
+                 * @param renderingDef the rendering definition
+                 * @param projectionDef the projection definition (optional
+                 * @param planeDef the plane definition
+                 * @return See above.
+                 **/
+                idempotent LongIntegerArrayMap getHistogram(IntegerList channels, int binSize, omero::model::RenderingDef renderingDef, omero::model::ProjectionDef projectionDef, omero::romio::PlaneDef planeDef) throws ServerError;
+                
                 /**
                  * Returns the byte width for the pixel store.
                  * @return See above.
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -327,7 +327,7 @@ module omero {
                 idempotent void setTimepoint(Ice::ByteSeq buf, int t) throws ServerError;
 
                 /**
-                 * Retrieves the histogram data for the specified plane and channels
+                 * Retrieves the histogram data for the specified plane and channels. This method can currently only handle non-pyramid images.
                  * @param channels the channels to generate the histogram data for
                  * @param binCount the number of the histogram bins (optional, default: 256)
                  * @param plane the plane (optional, default: whole region of first z/t plane)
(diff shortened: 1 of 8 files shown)


jburel:
@dominikl: I won't worry for now about the `slice` option in `PlaneDef`, currently the rendering engine only support `XY`

components/server/src/ome/services/RawPixelsBean.java
```diff
@@ -665,6 +667,51 @@ public synchronized void setTimepoint(byte[] arg0, int arg1) {
             handleException(e);
         }
     }
+    
+    @RolesAllowed("user")
+    public synchronized Map<Integer, int[]> getHistogram(int[] channels,
+            int binSize, PlaneDef plane) {
+        errorIfNotLoaded();
```
jburel, line 679: check if the binSize is valid

dominikl:
Started to do the client work and noticed a problem there. When looking at some dv images, I noticed that these usually are of type UINT16, which means the value range is 0 to  65355, but most of the pixels seem to be in the range 0-255. When I now request the histogram data with `binSize=256`, basically all pixels end up in the first bin `data[0]`.  Does that mean I shouldn't use `PixelData.getPixelValue()` (see https://github.com/dominikl/openmicroscopy/blob/0a07f72015a4ca5bf93139c23631a7fa662dcfc5/components/server/src/ome/services/RawPixelsBean.java#L710 ) for determining in which bin a pixel goes? But what else? (/cc @jburel )

jburel:
If the values are between 0-255 this probably means that the values have already been mapped.
I will have a closer look at the code

Need to use similar code that the one used in the RawFacility that we then used in insight to give the pixels intensity in MT

will-moore:
But you don't need to use the full pixel range (0 - 65355) for the histogram. Just use the range of darkest to brightest pixel for that channel, which will be more like 50 - 2000 or something more reasonable. 
That's what I was doing for web histograms: https://github.com/openmicroscopy/openmicroscopy/pull/4703/files#diff-fc3a666f42c5099bec2593d6f4b4cf2dR2373 

```
channel.getStatsInfo().getGlobalMin() / Max
```

dominikl:
I don't have the `ChannelData` object in the method, so I just iterate over the pixels to find the min and max pixel value, in order to determine the value range. Is that equivalent to `channel.getStatsInfo().getGlobalMin() / Max` ?

jburel:
the globalmin/max is the min/max across all planes
the histogram is per plane

will-moore:
It must be more expensive to get min/max by iterating all pixels in the plane than by looking the value up from the DB.
Also, it's nicer to use the global min/max because the histogram is plotted above the slider that has global min/max range with markers on the plot to indicate where the slider handle is, so this wouldn't work as well if the ranges of histogram and slider were different.
Also, when you move through Z/T, it's probably nicer to have the histogram plotted to the same range with all planes, so that the plots are comparable.

jburel:
you could use `retrievePixDescription` from `IPixels` to load the stats info
you have a ref in the class `private transient IPixels metadataService;`

will-moore:
What might be nicest is to allow you to specify the histogram range min/max for each channel in the API call.
If the most interesting part of the histogram is a very small portion of the global pixel intensity range, you  might just want to get that data (although we're not supporting that in the webclient UI).

jburel:
we could introduce a boolean flag: use min/max of the plane or global min/max
min/max plane will means that we will have to loop twice but that will be a choice by user

components/blitz/resources/omero/api/RawPixelsStore.ice
```diff
@@ -325,6 +326,16 @@ module omero {
                  **/
                 idempotent void setTimepoint(Ice::ByteSeq buf, int t) throws ServerError;
 
+				/**
```
mtbc: weird indent, at least for new blocks try to get tabs/spaces sorted in IDE

components/server/src/ome/services/RawPixelsBean.java
```diff
@@ -665,10 +671,91 @@ public synchronized void setTimepoint(byte[] arg0, int arg1) {
...
+            min = channel.getStatsInfo().getGlobalMin();
+            max = channel.getStatsInfo().getGlobalMax();
+            // if max == 1.0 the global min/max probably has not been
+            // calculated; fall back to plane min/max
+            if (max > 1.0)
+                return new double[] { min, max };
+        }
+
+        min = Double.MAX_VALUE;
+        max = 0;
```
mtbc: is it possible that the max is negative?

jburel: it could

dominikl: What does that mean in "image terms", is the minimum value then the one closest to zero or most negative one (and maximum vice versa)? I.e. do I have to use the absolute values for the histogram?

jburel: the default min/max as a fallback should be set according to the pixels type
see StatsFactory in rendering code as an example

dominikl: When I try to use the StatsFactory...

```
StatsFactory sf = new StatsFactory();
double[] minmax = sf.initPixelsRange(metadataService
                        .retrievePixDescription(id));
```

... this gives me again `0` as min and `65535` as max value for an UINT16 image (that's with a test image, where the the pixel values have been set manually, not a "real" imported image).

jburel: sorry I might be missing something, probably need a quick chat

dominikl:
Would the last commit address the negative pixel value issue? If not, yes, a brief chat might be required.

will-moore:
Seems to be working well for me.
Opened a PR to add `getHistogram()` to the Blitz image wrapper. #4955

will-moore:
Minor suggestion: Do you think `binSize` should really be called `binCount`? More bins means that the size of each bin is smaller.

dominikl:
That's true, I'll rename the parameter!

will-moore:
If I try this with a Big image, E.g 70k \* 70k I get

```
  File "/Users/wmoore/Desktop/OMERO/openmicroscopy/dist/lib/python/omero_api_RawPixelsStore_ice.py", line 1641, in getHistogram
    return _M_omero.api.RawPixelsStore._op_getHistogram.invoke(self, ((channels, binSize, globalRange, plane), _ctx))
{
    serverStackTrace = ome.conditions.ApiUsageException: Converting Long 5639164907 to Integer is an overflow.
    at ome.io.nio.RomioPixelBuffer.safeLongToInteger(RomioPixelBuffer.java:123)
    at ome.io.bioformats.BfPixelBuffer.getPlane(BfPixelBuffer.java:178)
    at ome.services.RawPixelsBean.getHistogram(RawPixelsBean.java:705)
```

dominikl:
What shall we do with big images? Disable histograms for these images? I guess otherwise I'd have to do some tiling?

jburel:
If the image is big, we will need to use the region of planeDef
Basically the client should in that case request the region corresponding to the viewport
if the region is not set for big images, I will not try to build the histogram

will-moore:
Currently in webclient I disable histogram for Big images. I think that mostly big images (in pathology) are 8-bit and so histograms are probably not required much. Suggest that we disable histograms for big images now, and consider adding it later. The region of a big image in viewport will often be the entire image, so getting histogram for that is not really an option.

dominikl:
With the last commit, an `ApiUsageException` will be thrown, if the image is a big image. Is that ok, or is there a better way to indicate the problem?

components/server/src/ome/services/RawPixelsBean.java
```diff
@@ -678,6 +678,10 @@ public synchronized void setTimepoint(byte[] arg0, int arg1) {
             int binCount, boolean globalRange, PlaneDef plane) {
         errorIfNotLoaded();
 
+        if (requiresPixelsPyramid())
+            throw new ApiUsageException(
+                    "This method can not handle tiled images yet.\n");
```
mtbc: the `\n` is odd

dominikl: That's true... must have slipped in with copy/paste.

pwalczysko:
@dominikl : Inspired by your last comment, had a look at UI behaviour of Big images when Histogram is selected. Although the Preview pane is disabled for Big images in Insight, in Full Viewer you can expand the rendering pane and click the "Histogram" box which results in empty graph being shown.
Not the best behaviour.

screen shot 2016-11-22 at 10 29 12

dominikl:
Thanks. Yes, forgot to handle it on the client side. I'll disable the histogram button for big images in #4946 .

pwalczysko:
The test https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/419/testngreports/integration/RawPixelsStoreTest/ passes. 

Other than the [UI issue](https://github.com/openmicroscopy/openmicroscopy/pull/4931#issuecomment-262204624) all is fine here. Note that the Insight log is showing the API usage exception as expected after I have ticked the "Histogram" box on a big image.

pwalczysko:
Comment added to https://github.com/openmicroscopy/openmicroscopy/pull/4946. Ready to merge.

will-moore:
I was just adding a test to https://github.com/openmicroscopy/openmicroscopy/pull/4703 and noticed a couple of strange issues.

For a test image created in python test with a grey gradient, I'd expect all the histogram bins to be similar numbers, but I see that alternative bins are full. 
E.g. with 
```
omero import "histogram&sizeX=125&sizeY=125.fake"
```
at http://web-dev-merge.openmicroscopy.org/webclient/?show=image-46401 (user-1)
using 256 bins I see 
```
"data": [615, 0, 115, 0, 115, 0, 115,
```
json data at http://web-dev-merge.openmicroscopy.org/webgateway/histogram_json/46401/channel/0/

Also, since that image is in a private group, if I try to get the histogram as 'root' E.g.
Go to http://web-dev-merge.openmicroscopy.org/webgateway/histogram_json/46401/channel/0/ as root: 

I get:
```
InternalException: exception ::omero::InternalException
{
    serverStackTrace = ome.conditions.InternalException:  Wrapped Exception: (java.lang.NullPointerException):
null
	at ome.services.RawPixelsBean.getHistogram(RawPixelsBean.java:708)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
```

jburel:
@will-moore, @dominikl: the PR has been merged so need to add more tests in another PR

jburel:
added entry to https://trello.com/c/6gKn8FT2/98-histogram with link to the comment

dominikl:
The higher number for bin[0] looks right, because the image contains a black square; but yes, the alternating zero values look a bit worrying. I'll check that again (and add an integration test using a uint8 image (maybe it's related to that)).

dominikl:
@will-moore I think the data looks right, like I mentioned previously the image contains a block of black pixels, hence the high pixel count for `bin[0]`. I looked at the pixel values for these `.fake` images, they range from `0` to `100`. That's why there are so many bins with value `0` in between (when using a binCount of 256).

I'm not sure about the `NullPointerException`. How can I replicate this?

will-moore:
@dominikl If the image is not owned by "root" and is in a private group and the "root" user tries to get histogram data, you should get the exception above.

dominikl:
The only thing which can be `null` in that case is ` ome.api.IPixels.retrievePixDescription(long arg0)` returns `null`. No idea why, or how to fix that. Could only catch it and return no histogram data for these cases, because without the `Pixels` object I can't determine the min/max of the histogram.

joshmoore:
https://github.com/openmicroscopy/openmicroscopy/blob/v5.2.5/components/server/src/ome/logic/PixelsImpl.java#L68

If it's known to exist, the only reason that this method should return a `null` is if the security context is such that the pixels can't be loaded for the current user. `omero.group=$GROUP_ID` or `omero.group=-1` may fix the issue.

PR #4933, opened 2016-11-10
Paths to object well image fix

will-moore:
# What this PR does

This fixes a bug found during IDR testing today and reproduced on develop:
https://www.openmicroscopy.org/qa2/qa/feedback/17421/
Introduced in https://github.com/openmicroscopy/openmicroscopy/pull/4774

To reproduce / test:
 - Need to have more than 200 (1 page) of orphaned images (so that orphaned images are paginated)
 - Search for an Image that is in a Well
 - In search results, select image and expand the Tables panel to the right (NB: don't need to have any Tables data).
 - This makes a call to ```webclient/api/paths_to_object/?image=ID``` which previously failed if the image is in a Well (not orphaned) AND there are more than 1 page of orphaned images.
 - With this fix, should get no error when Tables panel is expanded.
 - Also, test that going to ```webclient/api/paths_to_object/?image=ID``` for the Image in Well returns json that includes the WellID

Check that orphaned pagination still works
 - Browse to orphaned images and pick an image that is on page 2 (or more).
 - Copy the link to it, paste into browser and check you are returned to the correct page of orphaned images and image is selected.
 - Also can check that ```webclient/api/paths_to_object/?image=ID``` for this image has correct pagination details returned.

The bug was because we try to find the index of orphaned image in list of ALL orphaned images (to know which page it is on) *BEFORE* checking whether the image is in a Well.
This was also very slow on IDR because it has > 10,000 orphaned images.

Now, we only try to load ALL orphaned images to find page AFTER we've checked to see if image is in a Well.

cc @joshmoore @aleksandra-tarkowska @manics

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/show.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/show.py
@@ -508,7 +508,6 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
 
     # Hierarchies for this object
     paths = []
-    orphanedImage = False
 
     # It is probably possible to write a more generic query instead
     # of special casing each type, but it will be less readable and
@@ -586,31 +585,45 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
                 path.append(ds)
 
             # If it is orphaned->image
+            paths_to_img = []
             if e[2] is None:
-                orphanedImage = True
-                orph = {
-                    'type': 'orphaned',
-                    'id': e[0].val
-                }
-                iids = get_image_ids(conn, groupId=e[5].val, ownerId=e[0].val)
-                if len(iids) > page_size:
-                    try:
-                        index = iids.index(imageId)
-                        page = (index / page_size) + 1  # 1-based index
-                        orph['childCount'] = len(iids)
-                        orph['childIndex'] = index
-                        orph['childPage'] = page
-                    except ValueError:
-                        # If image is in Well, it won't be in orphaned list
-                        pass
-                path.append(orph)
+                # Check if image is in Well
+                paths_to_img = paths_to_well_image(
+                        conn, params,
+                        well_id=well_id, image_id=image_id,
+                        acquisition_id=acquisition_id,
+                        plate_id=plate_id,
+                        screen_id=screen_id,
+                        experimenter_id=experimenter_id,
+                        orphanedImage=True)
+                if len(paths_to_img) == 0:
+                    orph = {
+                        'type': 'orphaned',
+                        'id': e[0].val
+                    }
+                    iids = get_image_ids(conn, groupId=e[5].val,
+                                         ownerId=e[0].val)
+                    if len(iids) > page_size:
+                        try:
+                            index = iids.index(imageId)
+                            page = (index / page_size) + 1  # 1-based index
+                            orph['childCount'] = len(iids)
+                            orph['childIndex'] = index
+                            orph['childPage'] = page
+                        except ValueError:
+                            # If image is in Well, it won't be in orphaned list
+                            pass
+                    path.append(orph)
 
-            # Image always present
-            path.append({
-                'type': 'image',
-                'id': imageId
-            })
-            paths.append(path)
+            if len(paths_to_img) > 0:
+                paths = paths_to_img
+            else:
+                # Image always present
+                path.append({
+                    'type': 'image',
+                    'id': imageId
+                })
+                paths.append(path)
 
     elif lowest_type == 'dataset':
         q = '''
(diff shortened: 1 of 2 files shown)


atarkowska:
--rebased-to #4934

pwalczysko:
All checks have passed.
The json output in the image-in-the-well case shows that the image is in the well
screen shot 2016-11-11 at 10 30 55

The json output in the orphaned-image-on-second-page shows that the image is on second page
screen shot 2016-11-11 at 10 35 48

pwalczysko:
Ready to merge fmpov.

jburel:
could you please add tests to `test_show.py`?

atarkowska:
@will-moore as #4934 was merged could you rebase last commit please

mtbc:
Is the test outcome reported in CI anywhere?

will-moore:
Test is passing at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/386/testReport/OmeroWeb.test.integration.test_show/TestShow/test_well_image/

mtbc:
Aha, thank you! I was confused through not seeing it at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python/lastCompletedBuild/testReport/.

components/tools/OmeroWeb/omeroweb/webclient/show.py
```diff
@@ -896,6 +846,87 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
...
+    q = '''
+        select coalesce(sowner.id, plowner.id, aowner.id, wsowner.id),
+               slink.parent.id,
+               plate.id,
+               acquisition.id,
+               well.id
+        from WellSample wellsample
+        left outer join wellsample.details.owner wsowner
+        left outer join wellsample.plateAcquisition acquisition
+        left outer join wellsample.details.owner aowner
```
chris-allan, line 865: The same object `wellsample.details.owner` is being applied to two separate aliases (`wsowner` and `aowner`) here. As `details` are always present and the base of your `JOIN`s is `wellsample` you also don't need `left outer` here.

Maybe add a failing test case for this condition and then fix the bug? Or is this just completely irrelevant and can be removed from the query?

components/tools/OmeroWeb/omeroweb/webclient/show.py
```diff
@@ -896,6 +846,87 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
...
+        join wellsample.well well
+        left outer join well.plate plate
```
chris-allan, line 867: `plate` is a required element of `well` as is `details` as above so do not need to be `left outer` joined.

chris-allan (changes requested):
Based on past experience, I'm fairly concerned here about the number of objects that are being joined in the WellSample-Well-Plate-Screen hierachy being queried. This query has the potential of taking many seconds to return and should be evaluated, potentially in isolation, on a Plate and/or Screen of realistic size.

/cc @joshmoore, @aleksandra-tarkowska, @jburel

atarkowska:
@chris-allan this PR is fixing the issue where unnecessarily image in the well were tested as an orphaned image. @will-moore  changed the order of the logic identifying images in a well. HQL contently needs improvement as you pointed out, but lets not block this PR as there are no changes to HQL. We will review that in a following up PR to keep it inline with metadata52 branch. Would that be ok with you?

chris-allan:
Completely understood, @aleksandra-tarkowska. Your call on the scope, I don't have enough visibility there.

atarkowska:
follow up in a card https://trello.com/c/VEXIxsT1/237-review-paths-to-objects-hql-queries

will-moore:
I'm happy to fix queries in this PR, then when this is merged we can rebase to IDR all the commits missing from #4934 (this already includes one commit: https://github.com/openmicroscopy/openmicroscopy/pull/4933/commits/103e87fd0221998e72bcde254acebe1b2c04d6f9)

Alternatively this could be merged as is and queries improved in subsequent PR?

chris-allan:
Either option is fine from my perspective.

will-moore:
@chris-allan A sample query from IDR http://idr-demo.openmicroscopy.org/webclient/api/paths_to_object/?image=696619 takes about 150 ms (even with a cache-busting ```?_=random```)

chris-allan:
That's pretty reasonable then @will-moore.

will-moore:
Let's fix queries in another PR, since we now have a trello card for that. Smaller PRs are generally best!

jburel:
@chris-allan: any more comments? since the "red" change requested is still on

chris-allan (approved):

PR #4934, opened 2016-11-10
Paths to object well image fix (rebased onto metadata52)

atarkowska:
This is the same as gh-4933 but rebased onto metadata52.

----

# What this PR does

This fixes a bug found during IDR testing today and reproduced on develop:
https://www.openmicroscopy.org/qa2/qa/feedback/17421/
Introduced in https://github.com/openmicroscopy/openmicroscopy/pull/4774

To reproduce / test:
 - Need to have more than 200 (1 page) of orphaned images (so that orphaned images are paginated)
 - Search for an Image that is in a Well
 - In search results, select image and expand the Tables panel to the right (NB: don't need to have any Tables data).
 - This makes a call to ```webclient/paths_to_object/?image=ID``` which previously failed if the image is in a Well (not orphaned) AND there are more than 1 page of orphaned images.
 - With this fix, should get no error when Tables panel is expanded.
 - Also, test that going to ```webclient/paths_to_object/?image=ID``` for the Image in Well returns json that includes the WellID

Check that orphaned pagination still works
 - Browse to orphaned images and pick an image that is on page 2 (or more).
 - Copy the link to it, paste into browser and check you are returned to the correct page of orphaned images and image is selected.
 - Also can check that ```webclient/paths_to_object/?image=ID``` for this image has correct pagination details returned.

The bug was because we try to find the index of orphaned image in list of ALL orphaned images (to know which page it is on) *BEFORE* checking whether the image is in a Well.
This was also very slow on IDR because it has > 10,000 orphaned images.

Now, we only try to load ALL orphaned images to find page AFTER we've checked to see if image is in a Well.

cc @joshmoore @aleksandra-tarkowska @manics

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/show.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/show.py
@@ -508,7 +508,6 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
 
     # Hierarchies for this object
     paths = []
-    orphanedImage = False
 
     # It is probably possible to write a more generic query instead
     # of special casing each type, but it will be less readable and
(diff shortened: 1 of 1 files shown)


atarkowska:
--rebased-from #4933

joshmoore:
Merging based on https://github.com/openmicroscopy/openmicroscopy/pull/4933#issuecomment-259930059

PR #4936, opened 2016-11-11
Test for server errors

mtbc:
# What this PR does

This PR adds a broken test for where I would expect the Blitz API to give me a `ServerError`. The intent is simply to note how to reproduce the problem that it doesn't: nobody has the time to investigate and fix right now.

# Testing this PR

Check that`RawFileStoreTest.testBadFileId` is listed as failing at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-broken/testngreports/ and that you would expect some kind of `ServerError` for passing a non-existent file ID to `RawFileStorePrx.setFileId`.

### DIFF
--- a/components/tools/OmeroJava/test/integration/RawFileStoreTest.java
+++ b/components/tools/OmeroJava/test/integration/RawFileStoreTest.java
@@ -1,7 +1,5 @@
 /*
- * $Id$
- *
- *   Copyright 2006-2010 University of Dundee. All rights reserved.
+ *   Copyright 2006-2016 University of Dundee. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  */
 package integration;
@@ -9,6 +7,7 @@ package integration;
 import java.util.Iterator;
 import java.util.List;
 
+import omero.ServerError;
 import omero.api.IScriptPrx;
 import omero.api.RawFileStorePrx;
 import omero.model.OriginalFile;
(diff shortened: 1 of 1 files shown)


mtbc:
When `expectedExceptions` is used then TestNG fails the test if no exception is thrown.

mtbc:
>  Expected exception omero.ServerError but got Ice.UnknownException

dominikl:
Looks good. But yes, the `Ice.UnknownException` is strange. Is there a PR planned to fix the server method? And what's the reason for running the test in `groups = "broken"`?

mtbc:
It's in broken because it fails (don't want to fail CI). I think there should be a PR someday to fix the server method, which can also remove this test from `broken`, but we have so much other higher-priority backlog that I figure that for now we can settle for having a test that at least records the problem and maybe we'll have more time someday when we go back and review the broken tests and remember the issue: this PR is basically my device for moving on having noticed the issue at all.

dominikl:
Thanks @mtbc . Good to merge from my point of view.

jburel:
@mtbc: do you have a card recording the problem?
I vaguely remember a "generic"card  "review broken tests" but I cannot find it.

mtbc:
@jburel: I don't: I am afraid of Trello turning into Trac, especially given its rubbish search. I'd agree that there _should_ be a generic "review broken tests" card somewhere though, happy to add one.

jburel:
@mtbc: that's my fear too, let's keep looking before creating a new one ;-)

mtbc:
Perhaps https://trello.com/c/dQfzcVXC/14-broken-tests-status can be repurposed.

jburel:
certainly. I did not look at that board!!

PR #4923, opened 2016-11-03
customize accordion (rebased onto develop)

atarkowska:
# What this PR does

rebase https://github.com/openmicroscopy/openmicroscopy/pull/4876 and https://github.com/openmicroscopy/openmicroscopy/pull/4900 

# Testing this PR

1. test accordion when single object is selected
2. test accordion when multiple objects are selected
3. test if long values in a maps are shown in multiple lines
4. customize accordion labels (test locally)

PR #4942, opened 2016-11-15
prevent from editing map annotations in batch_annotation pane

atarkowska:
# What this PR does

partial rebase from https://github.com/openmicroscopy/openmicroscopy/pull/4923
only commits related to metadata panel

# Testing this PR

1. Login to web
2. Select multiple images
3. check if editing mode is not available on key-value pairs. Maps should be present


pwalczysko:
This PR might have introduced a regression in Right-hand-pane when multiselection of objects is performed, see https://github.com/openmicroscopy/openmicroscopy/pull/4910#issuecomment-261908493

atarkowska:
@joshmoore, @eleanorwilliams heads up, @will-moore added quickly a way to display which map belongs to which object (when multiple selected). Do we want to style it, move to tooltip?

firefoxscreensnapz019

eleanorwilliams:
Not sure we would want to show this to IDR users, but would be useful for us to see.  In the case of mapr, would it display multiple sources if the de-duplicated annotation is linked to many images or would it show the first one the annotation was linked to?

atarkowska:
@eleanorwilliams try http://10.0.51.150/web/webclient/?show=screen-3|screen-102, Certainly I will update mapr to do similar :)

atarkowska:
@jburel this PR has a wrong label, in case it matters

joshmoore:
Fixed the label; it doesn't have any impact in terms of builds but to prevent confusion.

atarkowska:
tested in https://github.com/openmicroscopy/openmicroscopy/pull/4923#issuecomment-263884540 (this is a partial rebase PR to keep two branches in line)

joshmoore:
Looks good. Thanks, @aleksandra-tarkowska. This will be in the upcoming 0.3.1 release.

PR #4943, opened 2016-11-15
add test for robustness of script service listing scripts

mtbc:
# What this PR does

This PR adds a broken test for where I would expect the script service to be able to list scripts for me. The intent is simply to note how to reproduce the problem that it doesn't: nobody has the time to investigate and fix right now.

# Testing this PR

Check that `ScriptServiceTest.testGetScriptsFiltersUnreadable` is listed as failing at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-broken/testngreports/ and that you would expect `getScriptsByMimetype` to work just fine.

### DIFF
--- a/components/tools/OmeroJava/test/integration/ScriptServiceTest.java
+++ b/components/tools/OmeroJava/test/integration/ScriptServiceTest.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2010 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


mtbc:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-broken/414/testngreports/integration/ScriptServiceTest/testGetScriptsFiltersUnreadable/ shows expected (but unfortunate) error because `getScriptsByMimetype` threw `SecurityViolation`.

dominikl:
Looks alright, good to merge 👍 

If `getScriptsByMimetype` fails after another script is uploaded, doesn't that also mean that retrieving the lookup tables will fail, if someone uploads a new script to the server?

mtbc:
Thank you. If the LUT code is similarly relying on `ScriptRepoHelper.loadAll` then it may indeed face the same problem.

(Things probably work fine if `uploadScript` is used as everyone intended, it's the test's other route that seems to cause the problem.)

mtbc:
(just pushed the wrong thing again, ignore the noise)

mtbc:
It does look like mornings are the time to try.

dominikl:
:-) Now I quickly have to try to get my PRs green...

PR #4944, opened 2016-11-16
JNA compliancy

dominikl:
# What this PR does

In newer versions of JNA the method `getFieldOrder()` for `Structure` objects is abstract and has to be implemented by subclasses (see https://github.com/java-native-access/jna/blob/master/src/com/sun/jna/Structure.java#L891 ). 

# Testing this PR

I noticed this error when testing the Insight-ImageJ plugin with latest Fiji. So that might be a good test. Get latest Fiji, add ImageJ plugin, connect to an OMERO server. Without the PR you'd get an `AbstractMethodError`. Also test plugin with ImageJ, just to be sure, it doesn't break anything there.

# Related reading

https://github.com/java-native-access/jna/blob/master/www/GettingStarted.md

Thanks @waxenegger for pointing me in the JNA direction!

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/appdata/LANGANDCODEPAGE.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/model/appdata/LANGANDCODEPAGE.java
@@ -1,7 +1,7 @@
 /*
  *
  *------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


pwalczysko:
Works fine in Fiji and ImageJ. Was able to reproduce the original problem (on Windows).
Ready to merge.

PR #4946, opened 2016-11-16
Histogram insight

dominikl:
# What this PR does

Adds the histogram display to Insight's preview panel. 
Based on #4931 .

# Testing this PR

Check if the histogram works like expected.

### DIFF
--- a/components/tools/OmeroJava/test/integration/gateway/RawDataFacilityTest.java
+++ b/components/tools/OmeroJava/test/integration/gateway/RawDataFacilityTest.java
@@ -98,7 +98,8 @@ public class RawDataFacilityTest extends GatewayTest {
 
         int[] exp = new int[256];
         for (byte b : rawData) {
-            exp[Byte.toUnsignedInt(b)]++;
+            int bin = ((int) b) & 0xFF;
+            exp[bin]++;
         }
 
         Map<Integer, int[]> histo = rawdataFacility.getHistogram(rootCtx,
--- a/components/tools/OmeroJava/test/integration/gateway/RawDataFacilityTest.java
+++ b/components/tools/OmeroJava/test/integration/gateway/RawDataFacilityTest.java
@@ -22,6 +22,7 @@ package integration.gateway;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Random;
 import java.util.UUID;
 
@@ -90,6 +91,26 @@ public class RawDataFacilityTest extends GatewayTest {
         Assert.assertEquals(planeData, rawDataPart);
     }
     
+    @Test
+    public void testGetHistogram() throws DataSourceException,
+            DSOutOfServiceException, DSAccessException {
+        ImageData img = browseFacility.getImage(rootCtx, imgId);
+
+        int[] exp = new int[256];
+        for (byte b : rawData) {
+            exp[Byte.toUnsignedInt(b)]++;
+        }
+
+        Map<Integer, int[]> histo = rawdataFacility.getHistogram(rootCtx,
+                img.getDefaultPixels(), new int[] { 0 }, 0, 0);
+        int[] data = histo.entrySet().iterator().next().getValue();
+        Assert.assertEquals(data.length, 256);
+
+        for (int i = 0; i < 256; i++) {
+            Assert.assertEquals(data[i], exp[i]);
+        }
+    }
+    
     private void initData() throws Exception {
         ProjectData p = createProject(rootCtx);
         DatasetData d = createDataset(rootCtx, p);
@@ -98,12 +119,13 @@ public class RawDataFacilityTest extends GatewayTest {
         IPixelsPrx svc = gw.getPixelsService(rootCtx);
         List<IObject> types = svc
                 .getAllEnumerations(PixelsType.class.getName());
+        PixelsType type = (PixelsType) types.get(2); // unit8
         List<Integer> channels = new ArrayList<Integer>();
         for (int i = 0; i < 3; i++) {
             channels.add(i);
         }
         imgId = svc.createImage(100, 100, 1, 1, channels,
-                (PixelsType) types.get(1), name, "").getValue();
+                type, name, "").getValue();
 
         List<Long> ids = new ArrayList<Long>(1);
         ids.add(imgId);
@@ -122,7 +144,10 @@ public class RawDataFacilityTest extends GatewayTest {
         store.setPixelsId(img.getDefaultPixels().getId(), false);
         Random rand = new Random();
         rawData = new byte[100 * 100];
-        rand.nextBytes(rawData);
+        for (int i = 0; i < rawData.length; i++) {
+            int r = rand.nextInt(256);
+            rawData[i] = (byte) r;
+        }
         store.setPlane(rawData, 0, 0, 0);
         gw.closeService(rootCtx, store);
     }
--- a/components/blitz/resources/omero/Collections.ice
+++ b/components/blitz/resources/omero/Collections.ice
@@ -139,6 +139,7 @@ module omero {
         dictionary<long,   ByteArray>                  LongByteArrayMap;
         dictionary<long,   omero::model::Pixels>       LongPixelsMap;
         dictionary<int,    string>                     IntStringMap;
+        dictionary<int,    IntegerArray>               LongIntegerArrayMap;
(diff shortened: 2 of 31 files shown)


components/tools/OmeroJava/test/integration/gateway/RawDataFacilityTest.java
```diff
@@ -90,6 +91,26 @@ public void testGetTile() throws DataSourceException, DSOutOfServiceException, D
...
     }
     
+    @Test
+    public void testGetHistogram() throws DataSourceException,
+            DSOutOfServiceException, DSAccessException {
+        ImageData img = browseFacility.getImage(rootCtx, imgId);
+
+        int[] exp = new int[256];
+        for (byte b : rawData) {
+            exp[Byte.toUnsignedInt(b)]++;
```
sbesson: This caused https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration/487/ to fail. Is it Java8 specific?

sbesson (changes requested):

components/tools/OmeroJava/test/integration/gateway/RawDataFacilityTest.java
jburel: it is Java 1.8 specific
no such method in Java 1.7 see https://docs.oracle.com/javase/7/docs/api/java/lang/Byte.html
and  for Java 1.8 https://docs.oracle.com/javase/8/docs/api/java/lang/Byte.html

gusferguson:
@dominikl 

Tested using OMERO.insight-5.3.0-m5-212-d2b95e4-ice36-b493-mac and OMERO.insight-5.3.0-m5-212-d2b95e4-ice36-b493-win eel user-3

Displays on check of checkbox
Channel displayed changes to that channel when sliders is moved
Drop-down selection of channel to display works as expected
Lines move when sliders moved and released when Live Update not selected - move dynamically - with considerable lag) when Live Update is selected in Preview pane. No live update in Full Viewer.
Cannot drag lines on histogram (don’t know if this is expected)
Apart from issue with Live Update - Preview pane and Full Viewer appear to behave the same

Issues:
- moving slider does not change line on histogram dynamically when Live Update is not selected and there is no Live Update in Full Viewer - only updates when released - this means you cannot move the slider until it is where you want it in on the histogram, but have to use trial and error approach
- histogram display does not change when displayed channel is toggled off or new channel toggled on using buttons - it would be expected that the display would change to the next on channel when a channel is toggled off, or the newly activated channel if one is toggled on
- colour of histogram is red no matter what colour is allocated to the channel. If colour of channel cannot be used (problematic with LUT) then I think it would be better to use a neutral grey colour for the histogram as current approach is confusing.
- drag on histogram enlarges section 
  - accidentally did it trying to move line and it left me with a completely blank histogram - screenshot 1 - very confusing
  - this only works in top down direction - confusing
  - no way of getting out of enlargement without moving a slider or changing channel
  - I presume this is intended to allow a magnified view of a section of the histogram to enable more accurate placing of lines, but it is ineffective as even if the lines are displayed in the enlarged section, you cannot adjust them dynamically when Live Update enabled as when the slider is moved, the enlargement is lost
  - I think the enlargement should be dropped unless dynamic adjustment of line placement is possible in it.
- the selected/mouse-over colour of the drop-down channel selection menu is invisible - should either have label background change to darker on selection/mouseover, or make selection/mouse-over text darker grey - particularly bad on Windows - screenshots 2 and 3

Incidental:
Don’t know if it is a combination of eel being slow and my working from home, but the rendering in the Full Viewer is **painfully** slow and in the Preview pane not much better.

I think we need to get @pwalczysko to review this PR as well.

user-3 user-3 connected to eel openmicroscopy org

parallels desktop

main

dominikl:
- Shall we use grey in general for the histogram color, otherwise it's not only the problem with LUTs, but also very light colors have to taken into account, e. g. a channel could be white, in which case the histogram bars would be invisible.
- I wasn't even aware that the chart is zoomable. For the sake of simplicity I'd just disable this feature (if possible)? Other thoughts?
- That's true, there's no immediate feedback when the slider is moved. I'll try to implement that. In the way that only the markers in the histogram are adjusted "live", not via "live update" where the image is rendered on every move.
- Also going to fix the channel dropdown, it's really hardly readable at the moment.

dominikl:
**Review** : Make sure issues noticed by @gusferguson  https://github.com/openmicroscopy/openmicroscopy/pull/4946#issuecomment-261500325 are fixed.

gusferguson:
@dominikl 

Tested using OMERO.insight-5.3.0-m5-235-6e38d40-ice36-b495-win and OMERO.insight-5.3.0-m5-235-6e38d40-ice36-b495-mac using eel and user-3

Grey colour for bars of histogram much better.
Zoom on drag in histogram no longer occurring.
Live movement of lines with movement of sliders working as expected.
Background of channel selection drop-down now fixed.

Issue noticed - the labels in the channel dropdown in the Windows version are being truncated at 5 characters and ... inserted. This means just the 5 digit channel label 457.0 is being truncated to 45... which is a bit odd.  See screenshot 1 below.
This is not occurring in the Mac version where the control is expanding to accommodate the name - see screenshot 2 below.
I think the issue in Windows needs to be sorted.

parallels desktop


main

dominikl:
Thanks @gusferguson . And looking at the screenshots, I should fix the strange grey backgrounds some elements have (e.g. the Greyscale and Histogram checkbox), Windows seems to handle the opacity different from OSX.

gusferguson:
@dominikl - fixing the grey backgrounds would be a bonus - they have always been there but were not so apparent in older versions - screenshot below from September.

parallels desktop

pwalczysko:
Another ToDo here is disabling the histogram checkbox for big images in full viewer see https://github.com/openmicroscopy/openmicroscopy/pull/4931#issuecomment-262204624

gusferguson:
@dominikl 

Tested with OMERO.insight-5.3.0-m5-238-8940b5d-ice36-b497-win and OMERO.insight-5.3.0-m5-238-8940b5d-ice36-b497-mac - eel - user 3

Backgrounds of Live update and Greyscale checkboxes now appear normal in Windows.
Truncation of long channel names better in Windows, but still not getting the expansion to fill available space that occurs with the Mac. Can live with this as it is now if solution is not straight forward.

Issue:
Windows: in the Preview pane when the Histogram checkbox is selected the area below the image with the Z/T numbers and theLive update checkbox disappears, along with half of the button toolbar.
Mac: same happening but just affecting the border above an below the button toolbar - Z/T nos. and Live update checkbox are still visible.
Looking back at previous screenshots, this has always been present but was not noticed before.
Doesn’t affect full viewer as layout is different.

Screenshots show with Histogram not selected and then with Histogram selected:

**Windows:**

parallels desktop

parallels desktop

**Mac:**

user-3 user-3 connected to eel openmicroscopy org

user-3 user-3 connected to eel openmicroscopy org

gusferguson:
Histogram disabled when viewing big images - just seems to be confused at times whether to enable the Preview pane or not - sometimes does, sometimes doesn't.

pwalczysko:
Confirming histogram on big images disabled on Mac as expected.

dominikl:
If it's ok, I'd just leave it as it is for now. Very hard to fix the layout issues of the right hand side panel, just too many variable components wrapped inside eachother.

dominikl:
I re-list it for tomorrows standup for a (hopefully) final review. Like mentioned before, I won't tackle the right hand panel layout issues within this PR.

jburel:
@dominikl: I agree, don't tackle it in this PR

gusferguson:
@dominikl 

Tested using  OMERO.insight-5.3.0-m6-263-feeee39-ice36-b503-mac and win eel user-3

Mac - still issues with the toolbar immediately below the preview pane image, but no worse than before and it sorts itself out if the width of the right hand pane is adjusted wider.

Windows - issue with the toolbar immediately below the preview pane image actually improved a bit - also normalises with pane width increase. 

All other functionality as expected.

Happy to settle with this and go ahead and merge.

Just for the record:

Had this weirdness happen when I closed the full viewer - not saving changes I had made. Could not replicate and did not appear when different image selected. 
Suggest no action needed at this stage, but if it starts being seen a lot may need looking in to.

parallels desktop

jburel:
https://github.com/openmicroscopy/openmicroscopy/pull/4931 has now been merged

jburel:
Requested change no longer exists. Merging

PR #4948, opened 2016-11-17
RGBA shape colors

dominikl:
# What this PR does

Makes sure that stroke and fill colors for Shapes are always RGBA integers.

# Testing this PR

Check that the measurement tool still works with respect to setting stroke and fill colors.

# Related reading

[Trello - Color in shape settings](https://trello.com/c/sGlYbbvB/213-color-in-shape-settings)

/cc @jburel  - Does that already fix the issue for Insight or are there other things which have to be adjusted?

### DIFF
--- a/components/blitz/src/omero/gateway/model/ShapeSettingsData.java
+++ b/components/blitz/src/omero/gateway/model/ShapeSettingsData.java
@@ -81,6 +81,37 @@ public class ShapeSettingsData
     public final static String FONT_REGULAR = "Normal";
 
     /**
+     * Returns the Color's RGBA value as integer
+     * 
+     * @param c
+     *            The {@link Color}
+     * @return See above.
+     */
+    public static int getRGBA(Color c) {
+        int rgba = 0;
+        rgba |= (c.getRed() & 255) << 24;
+        rgba |= (c.getGreen() & 255) << 16;
+        rgba |= (c.getBlue() & 255) << 8;
+        rgba |= (c.getAlpha() & 255);
+        return rgba;
+    }
+
+    /**
+     * Creates a {@link Color} object from an RGBA integer value
+     * 
+     * @param rgba
+     *            The RGBA value
+     * @return See above.
+     */
+    public static Color getColor(int rgba) {
+        int r = (rgba >> 24) & 0xFF;
+        int g = (rgba >> 16) & 0xFF;
+        int b = (rgba >> 8) & 0xFF;
+        int a = rgba & 0xFF;
+        return new Color(r, g, b, a);
+    }
+    
+    /**
      * Returns the font style is supported.
      *
      * @param value The value to format.
@@ -147,8 +178,7 @@ public class ShapeSettingsData
         Shape shape = (Shape) asIObject();
         RInt value = shape.getFillColor();
         if (value == null) return DEFAULT_FILL_COLOUR;
-        Color c = new Color(value.getValue(), true);
-        if (c.getAlpha() == 0) return new Color(value.getValue(), false);
+        Color c = getColor(value.getValue()); 
         return c;
     }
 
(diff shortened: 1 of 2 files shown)


jburel:
Need to review what web is doing
i.e. reading

jburel:
we will need tests added to RenderingEngine tests for example to make sure the color is set correctly.

dominikl:
Added integration test.
Just checked Web, looks it still interprets the color as ARGB, is that possible @will-moore ?

screen shot 2016-11-22 at 10 11 36

pwalczysko:
Stroke color and fill color work as expected, except that they do not update on the shape immediately unlike e.g. width and actually all other parameters, which update instantly). In order to see the change in the color you have to press "Save".

Would be nice to have the behaviour the same for all the features in the Inspector tab (instant change when changed). Not a blocker though.

dominikl:
Could repeat it, the crucial point is, to de-select the Roi (by clicking on some empty space on the image) while the ROI inspector still shows the properties of the previously selected ROI. But as it's not related to this PR specifically, I'd tackle that issue in a separate PR.

jburel:
@will-moore: https://github.com/openmicroscopy/openmicroscopy/pull/4948#issuecomment-262201059

will-moore:
Yes, we expect ARGB: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/marshal.py#L349

```
def rgb_int2css(rgbint):
    """
    converts a bin int number into css colour and alpha fraction.
    E.g. -1006567680 to '#00ff00', 0.5
    """
    alpha = rgbint // 256 // 256 // 256 % 256
    alpha = float(alpha) / 256
    r, g, b = (rgbint // 256 // 256 % 256, rgbint // 256 % 256, rgbint % 256)
    return "#%02x%02x%02x" % (r, g, b), alpha
```

Does the model specify RGBA integer? So we've been doing it wrong the whole time?
Or has the model changed?
If anyone's been storing colours as ARGB before now, they might get a nasty surprise when we change to using RGBA?

mtbc:
Yes, wrong the whole time: see the properties of Shape on http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2015-01/ROI_xsd.html#Shape

I figured a color-flipping DB upgrade at http://trac.openmicroscopy.org/ome/ticket/13146#comment:9 for those who want to apply it with a "WHERE ..." appropriate for their broken data.

mtbc:
Note also that one would expect alpha=0 to indicate transparency; don't know what web does for that.

joshmoore:
> appropriate for their broken data.

What's the heuristic for knowing whether it is broken or not?

mtbc:
I guess one has to work from the direction of understanding the provenance of the data -- from where came what and how -- or by manually looking through it for patterns of which looks wrong and which doesn't and what rule might fit that. An nearly inverse PL/pgSQL function is also supplied and either can be applied anytime.

I suppose one could be aware of especially popular colors and use them as a big hint either way or look for which way around brings the most contrast but I'd fear that trying to devise and apply heuristics would be a fair bit of work for maybe very little gain, even some loss. Perhaps the contents of the alpha byte might be a large hint. It's an avenue we could explore given enough spare time.

jburel:
Discussed with @chris-allan, for the shapes generated with Path Viewer an annotation is created so no need to apply the patch in that case

@will-moore: we have been doing it wrong since the introduction of the Measurement tool. The model has not changed!

jburel:
Will the web part be handled in another PR?
@dominikl @will-moore

will-moore:
Yes, card created at https://trello.com/c/YLP2lElr/248-shape-colour-read-as-rgba

jburel:
Corresponding web PR https://github.com/openmicroscopy/openmicroscopy/pull/4995

jburel:
Merging the clients PR
Last remaining part will be the upgrade script
see https://trello.com/c/sGlYbbvB/213-color-in-shape-settings

PR #4950, opened 2016-11-17
Blitz buildQuery params

will-moore:
# What this PR does

This is a subset of https://github.com/openmicroscopy/openmicroscopy/pull/4945 containing just the BlitzGateway and test changes.

This moves more logic for building queries to the BlitzGateway, to reduce duplication within api_query.py.

This means that the _getQueryString() for each BlitzGateway Wrapper object now returns a base query string, params and clauses. These are then augmented in conn.buildQuery(), which is used both by conn.getObjects() (and by the web /api_query.py - see #4945).

The params argument in
```conn.getObjects("Project", params)```
can now be an ```omero.sys.Parameters``` object (as before) OR can be a dictionary of parameters to support additional filtering.
E.g. ```{'page': 1, 'limit': 100, 'order_by': 'name', 'owner': 1, 'project': 5}```

cc @chris-allan

### CODE REVIEW

**Comment by will-moore** [comment]
@chris-allan If you could have a look at this when you get time, that would be great, thanks. And #4930 too (lower priority). Cheers.

File: components/tools/OmeroPy/src/omero/gateway/__init__.py [hunk 1]
```diff
@@ -268,14 +269,26 @@ def _unwrapunits(self, obj, units=None):
             return obj
         return obj.getValue()
 
-    def _getQueryString(self):
+    def _getQueryString(self, opts=None):
```
chris-allan, line 273: As this is going to be used more extensively now and its return value has changed to a tuple the function needs pretty thorough documentation and we should definitely mark it for detailed review in the changelog. We have many examples where this method is overridden and/or its return value is used.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3076,28 +3093,54 @@ def buildQuery(self, obj_type, ids=None, params=None, attributes=None):
...
-        if params is None:
-            params = omero.sys.Parameters()
-        if params.map is None:
-            params.map = {}
+        owner = None
+        order_by = None
+        opts = None
+        inputParams = None
+
+        if isinstance(params, dict):
```
chris-allan: I'm really not sure about this. At a minimum it needs documentation and to detailed review in the changelog as above. While it certainly keeps the prototype is it not feasible to add `opts` as an additional keyword argument?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4507,19 +4554,19 @@ def __eq__(self, a):
...
         """
         Used for building queries in generic methods such as
         getObjects("Annotation")
         """
-        return ("select obj from Annotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from Annotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 4594: Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4671,15 +4718,15 @@ def __init__(self, *args, **kwargs):
...
         """
         Used for building queries in generic methods such as
         getObjects("FileAnnotation")
         """
-        return ("select obj from FileAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent join fetch obj.file")
+        query = ("select obj from FileAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 4759: As above. Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4813,15 +4860,15 @@ class TimestampAnnotationWrapper (AnnotationWrapper):
...
         """
         Used for building queries in generic methods such as
         getObjects("TimestampAnnotation")
         """
-        return ("select obj from TimestampAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from TimestampAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 4906: As above. Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4862,15 +4909,15 @@ class BooleanAnnotationWrapper (AnnotationWrapper):
...
         """
         Used for building queries in generic methods such as
         getObjects("BooleanAnnotation")
         """
-        return ("select obj from BooleanAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from BooleanAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 4960: As above. Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4948,15 +4995,15 @@ def listParents(self, withlinks=True):
...
         """
         Used for building queries in generic methods such as
         getObjects("TagAnnotation")
         """
-        return ("select obj from TagAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from TagAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 5051: As above. Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4990,15 +5037,15 @@ class CommentAnnotationWrapper (AnnotationWrapper):
...
         """
         Used for building queries in generic methods such as
         getObjects("CommentAnnotation")
         """
-        return ("select obj from CommentAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from CommentAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 5098: As above. Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5030,15 +5077,15 @@ class LongAnnotationWrapper (AnnotationWrapper):
...
         """
         Used for building queries in generic methods such as
         getObjects("LongAnnotation")
         """
-        return ("select obj from LongAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from LongAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 5143: As above. Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5071,15 +5118,15 @@ class DoubleAnnotationWrapper (AnnotationWrapper):
...
         """
         Used for building queries in generic methods such as
         getObjects("DoubleAnnotation")
         """
-        return ("select obj from DoubleAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from DoubleAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 5189: As above. Is loosing the `details.group` fetch here not a regression?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5113,15 +5160,15 @@ class TermAnnotationWrapper (AnnotationWrapper):
...
         """
         Used for building queries in generic methods such as
         getObjects("TermAnnotation")
         """
-        return ("select obj from TermAnnotation obj "
-                "join fetch obj.details.owner as owner "
-                "join fetch obj.details.group "
-                "join fetch obj.details.creationEvent")
+        query = ("select obj from TermAnnotation obj "
+                 "join fetch obj.details.owner as owner "
```
chris-allan, line 5236: As above. Is loosing the `details.group` fetch here not a regression?

chris-allan (changes requested):

chris-allan:
Just as a bit of general feedback, having a large number of flake8 fixes embedded in the commits with code changes makes review incredibly time consuming. Every changed line has to be essentially checked on a character by character basis.

Parenthetically, and I'm all for keeping a consistent and readable style, if flake8 is throwing up an error and telling us to replace this:

```
        return (type(a) == type(self)
                and self._obj.id == a._obj.id
                and self.getName() == a.getName())
```

with this:

```
        return (type(a) == type(self) and
                self._obj.id == a._obj.id and
                self.getName() == a.getName())
```

... that's taking the pedantic style application way too far in my opinion and is going to create a lot of needless and conflict inducing churn in our repositories.

/cc @sbesson, @joshmoore

will-moore:
@chris-allan Yes, I had to add a lot of flake8 fixes because of travis failures in #4945, see https://travis-ci.org/openmicroscopy/openmicroscopy/jobs/176394210.
Without them the PR wouldn't even be mergeable, although I guess I could port them all to another PR, merge that, then merge develop back into this PR?!

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: Understood. I wasn't sure about this either. I was just trying to avoid too many arguments, thinking that use of params would mostly be covered by opts instead (they cover mostly the same functionality). But it can get confusing in the implementation.
I'm happy to add opts as additional argument. Then there's the question of whether params or opts takes precedence (E.g. if you specify different pagination or owner in each)?

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: See https://github.com/openmicroscopy/openmicroscopy/pull/4945/commits/ce55154736502cda50fadec2c1aa0d0c48948bae in original PR #4945 for when and why ```details.group``` was removed from these queries. cc @joshmoore

will-moore:
Error found by @aleksandra-tarkowska at https://github.com/ome/omero-gallery/pull/11
```
File "/home/hudson/virtualenv/lib/python2.7/site-packages/omero_gallery/views.py", line 125, in show_group
   for d in conn.listOrphans("Dataset", eid=user_id):
 File "/opt/hudson/workspace/WEB-DEV-merge-deploy/OMERO.py/lib/python/omero/gateway/__init__.py", line 2599, in listOrphans
   query = wrapper()._getQueryString()
TypeError: _getQueryString() takes exactly 2 arguments (1 given)
```

components/tools/OmeroPy/src/omero/gateway/__init__.py
chris-allan: Adding `opts` as a keyword argument is certainly preferable from my perspective. @aleksandra-tarkowska would probably like the opportunity to comment.

I think you can choose one and document it as taking precedence. Whichever one is the subset of the other is probably the a good candidate but I can see arguments both ways. That is, if you have the possibility to configure 50 things with `opts` and have `params` only able to configure 10 the settings from `params` should probably override those set in `params` if a call were made where both were present.

components/tools/OmeroPy/src/omero/gateway/__init__.py
chris-allan: Cool, no problem. Just something we should document and probably schedule time to fix at a fundamental level. During work with `omero-es` I haven't noticed issues like this and I'm **definitely** loading multiple groups that are not the default group. So I guess it's a combination of factors?

/cc @aleksandra-tarkowska

chris-allan:
Regarding the flake8 fixes even just having those fixes themselves in a separate commit would have made it a lot easier. The reviewer can then work on a commit by commit basis after those changes have been applied and it's then much easier to see what has been changed.

Was this code completely devoid of flake8 checking previously and thus we have all these changes to do or is it simply that one or more flake8 version increments have made certain stylistic or structural constructs more rigid?

Edit: Yep, regarding the differences in `_getQueryString()` method prototype we're "breaking" a largely internal API but one we use pretty liberally at least in our own software. We're going to have to document this quite thoroughly as mentioned above.

jburel:
@will-moore: wrt flake8 better to have a separate PR that we can merge quickly 
and then move forward with this PR

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: @chris-allan See https://trac.openmicroscopy.org/ome/ticket/11362#comment:9 for how to reproduce.

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: where is the unit test?

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: could you add unit test as well

will-moore:
@chris-allan:
- flake8 fixes handled in #4982 
- Added ```opts``` parameter to ```buildQuery()```
- Added docs to ```buildQuery()```, ```getObjects()``` and ```_getQueryString()```

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: @aleksandra-tarkowska Unit tests for every ```_getQueryString()``` would be a lot to add to this PR, but maybe just for the ones that handle ```opts``` parameter.

aleksandra-tarkowska: why? one parametrized test should be more then enough

atarkowska: why? we need one at least for parent. You could parametrize the test. In my opinion well written test is much better then the doc

chris-allan:
Thanks, @will-moore. I think the only thing I'd like to see added to the documentation are to explicitly define the supported set of `opts` keys for each method. Those need to be declared somewhere. You could alternatively switch to an object to define those which would make it somewhat more explicit and self documenting but that's really up to you. I'm fine with either technical approach.

will-moore:
@aleksandra-tarkowska I thought about that, but I don't think I really need a series of parameters for ```opts``` and ```params```.

components/tools/OmeroPy/test/unit/gatewaytest/test_build_query.py
```diff
@@ -0,0 +1,91 @@
...
+import pytest
+from omero.rtypes import wrap
+
+
+class TestBuildQuery(object):
+    """Test the conn.buildQuery() method for all Object Wrappers."""
+
+    @classmethod
+    @pytest.fixture(autouse=True)
+    def setup_class(cls, tmpdir, monkeypatch):
```
atarkowska: didn't you mean setup_method? setup_class and autouse doesn't sound right to me

components/tools/OmeroPy/test/unit/gatewaytest/test_build_query.py
```diff
@@ -0,0 +1,91 @@
...
+
+"""Gateway tests - buildQuery() as used by conn.getObjects()."""
+
+from omero.gateway import _BlitzGateway, BlitzObjectWrapper, KNOWN_WRAPPERS
+from omero.sys import Parameters, ParametersI, Filter
+import pytest
+from omero.rtypes import wrap
+
+
+class TestBuildQuery(object):
```
aleksandra-tarkowska: missing teardown

components/tools/OmeroPy/test/unit/gatewaytest/test_build_query.py
```diff
@@ -0,0 +1,91 @@
...
+    """Test the conn.buildQuery() method for all Object Wrappers."""
+
+    @classmethod
+    @pytest.fixture(autouse=True)
+    def setup_class(cls, tmpdir, monkeypatch):
+        """Prepare BlitzGateway with ICE_CONFIG."""
+        ice_config = tmpdir / "ice.config"
+        ice_config.write("omero.host=localhost\nomero.port=4064")
+        monkeypatch.setenv("ICE_CONFIG", ice_config)
+        cls.g = _BlitzGateway()
```
atarkowska: shouldn't you teardown that object

atarkowska (commented):
Why actually `buildQuery` is not a static method?

will-moore:
I've looked at turning ```def _getQueryString()``` into a ```classmethod``` but this seems to need quite a lot of changes.
Since the base ```BlitzObjectWrapper._getQueryString()``` uses the ```OMERO_CLASS``` from each subclass, this fails because the ```OMERO_CLASS``` is currently only set in the ```def __bstrap__(self):``` method of each subclass, which is called when the class is instantiated.
So, while I agree that ```buildQuery``` and ```_getQueryString``` should be static / class methods, I think that this cleanup should be done in a separate PR, since it's not directly related to the change in functionality of this PR (and will confuse the changes in this PR with unrelated fixes).

I'll place my current work on this into a separate branch to create a new PR later.

chris-allan (approved):

will-moore:
If everyone is happy with this, it would be nice to get it merged so I can cleanup the other PRs that depend on it: #4945 and #4993.
Thanks.
cc @jburel

will-moore:
#4993 is ready to review, but I really need to get this PR merged first to clean up the changes.
 Thanks.

jburel:
To check tomorrow if https://github.com/ome/omero-gallery/pull/11 is working before merging

jburel:
"gallery" is back up on cowfish
https://cowfish.openmicroscopy.org/webtrial/gallery/show_group/0/
Merging this PR

PR #4955, opened 2016-11-18
Histogram blitz gateway

will-moore:
# What this PR does

Adds ```getHistogram()``` to the Blitz imageWrapper.
Also adds test to gatewaytest/test_pixels.py

# Testing this PR

Check test is green.
I will next use this in the web PR #4703 

cc @chris-allan

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7823,6 +7823,37 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         (1, False, True): 'b',  # signed char
         }
 
+    @assert_pixels
+    def getHistogram(self, channels, binSize, globalRange=True,
+                     theZ=0, theT=0):
+        """
+        Get pixel intensity histogram of a single plane for specified channels.
+
+        Returns a map of channelIndex: integer list.
+        If globalRange is True, use the min/max for that channel over ALL
+        planes.
+        If False, use the pixel intensity range for the specified plane.
+
+        :param channels:        List of channel integers we want
+        :param binSize:         Number of bins in the histogram
+        :param globalRange:     If false, use min/max intensity for this plane
+        :param theZ:            Z index of plane
+        :param theT:            T index of plane
+        :return:                Dict of channelIndex: integer list
+        """
+
+        pixels_id = self.getPixelsId()
+        rp = self._conn.createRawPixelsStore()
+        try:
+            rp.setPixelsId(pixels_id, True, self._conn.SERVICE_OPTS)
+            plane = omero.romio.PlaneDef(self.PLANEDEF)
+            plane.z = long(theZ)
+            plane.t = long(theT)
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7823,6 +7823,37 @@ def setInvertedAxis(self, inverted):
         (1, False, True): 'b',  # signed char
         }
 
+    @assert_pixels
+    def getHistogram(self, channels, binCount, globalRange=True,
```
jburel: rename the method get_histogram and use _ instead of upper case

jburel:
I will also add test for big images

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel, line 7827: maybe time to unify method name with other part of the python code i.e not using upper case in the method name and variables, using "_"
Changing the existing ones will be a major effort
but maybe the new ones. 
I realise that it will not "match" the current pattern of the file but we need to start at some point

will-moore: I think we should either change all or none. Having two different styles in the same file will not be nice. I guess IF we're going to do it then 5.3.0 is as good a time as any. cc @chris-allan

jburel: Changing in 5.3 is a major job and with the rest of the tasks to be done I don't think this is realistic

chris-allan: Keeping this as `getHistogram()` seems fine to me and is in line with the existing style. While this differs from much of the other Python code, the gateway primarily deals with OMERO objects which do not follow the underscore style anyway.

will-moore:
Test is passing: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/384/testReport/OmeroPy.test.integration.gatewaytest.test_pixels/TestPixels/testGetHistogram/

dominikl:
Looks alright to me 👍 With the last commit on #4931 the server will throw an `APIUsageException` when `getHistogram` is called for a big image. Not sure if you want to deal with that in the Gateway.

will-moore:
I'm not sure there's much I can do to improve that. I'll simply let the APIUsageException get raised.

chris-allan:
That seems like a fine strategy to me. We'll be handling the reaction to an `APIUsageException` in a generic manner anyway. Might be worth documenting I guess but other than that I think this is good to merge assuming that it reflects the API you're expecting @jburel and @dominikl.

jburel:
I know "regions" support is not yet supported in the underline method i.e. histogram for a portion of the image (maybe that will not be useful!)
but you will need another method to support it and another method to support projection
cf. last week discussion.

jburel:
Discussed with @will-moore today, using "omero object" has not been the strategy in the gateway
Happy with the current implementation.
Merging

PR #4961, opened 2016-11-24
Limit search results to the first 500

joshmoore:
In order to protected the server from unnecessary
overhead, limit the number of search results to 500.
This can be improved by adding a per-call setting
and/or an overriding server-wide configuration.

cc: @aleksandra-tarkowska

### DIFF
--- a/components/server/src/ome/services/search/FullText.java
+++ b/components/server/src/ome/services/search/FullText.java
@@ -301,6 +301,7 @@ public class FullText extends SearchAction {
         ftQuery
         .setProjection(ProjectionConstants.SCORE,
                 ProjectionConstants.ID);
+        ftQuery.setMaxResults(500);
     }
 
     @Transactional(readOnly = true)
--- a/components/tools/OmeroPy/test/integration/test_search.py
+++ b/components/tools/OmeroPy/test/integration/test_search.py
@@ -50,6 +50,7 @@ class TestSearch(lib.ITest):
         r = q.findAllByFullText("TagAnnotation", uuid, None)
         assert 0 == len(r)
 
+    @pytest.mark.skipif(True, reason="potentially fails with hard-limit")
     def test3164Private(self):
         group = self.new_group(perms="rw----")
         owner = self.new_client(group)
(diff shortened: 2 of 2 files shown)


atarkowska:
is that failing tests https://10.0.51.106:8443/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.test_search/TestSearch/ ?

joshmoore:
I don't think so. I think 113 is still unhappy from previous DB changes. Probably a good time to get us moved to a new devspace.

atarkowska:
Search testes were ok yesterday :) hmm I didn't have problem with DB changes, need help?

joshmoore:
@aleksandra-tarkowska : I take it back; I think I know what's happening. The server must have a larger number of matching annotations and the limit is getting the older ones which aren't visible to the current user. This isn't something that will impact IDR, so I will likely need to disable these permission-related search tests.

atarkowska:
could you fix `./components/tools/OmeroPy/test/integration/test_search.py:36:19: E221 multiple spaces before operator` as well please

joshmoore:
Will do.

atarkowska:
thx

components/server/src/ome/services/search/FullText.java
```diff
@@ -301,6 +301,7 @@ protected void initializeQuery(FullTextQuery ftQuery) {
         ftQuery
         .setProjection(ProjectionConstants.SCORE,
                 ProjectionConstants.ID);
+        ftQuery.setMaxResults(500);
```
atarkowska, line 304: Do you think this could become param, that will correlate with batchSize https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/gateway/__init__.py#L4132 and do proper pagination?

joshmoore:
Pushed.

```
================================= 23 passed, 5 skipped in 265.44 seconds ==================================
```

aleksandra-tarkowska:
--ola150

aleksandra-tarkowska:
--ola150

components/server/src/ome/services/search/FullText.java
joshmoore: Somehow I missed this comment. We can certainly try for 0.3.2.

joshmoore:
Merging for IDR 0.3.1

atarkowska:
--rebased-to #5228

PR #4971, opened 2016-11-29
Show path

jburel:
# What this PR does

Fix class cast exception
The application will crash w/o the fix

# Testing this PR

 * Select an image
 * Go to the right-hand panel
 * Click on the "show file paths on server" button (3rd one starting from the left)
 * The application should not crash

cc @dominikl


dominikl:
Thanks @jburel . Noticed that exception too, but then it slipped my mind again. Works like expected, good to merge 👍  (Tested it locally, due to Travis failure)

dominikl:
Tested with today's merge build, works like expected 👍

jburel:
Thanks @dominikl

PR #4974, opened 2016-11-30
make movie

jburel:
# What this PR does

Mark make_movie as "broken"

We have not found yet a reliable repo to install via ansible mencoder on centos7
https://github.com/openmicroscopy/infrastructure/pull/174

Discussed today during the sysadmin meeting, for now we disable the script

# Testing this PR
Make sure https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/ is green (cowfish)

### DIFF
--- a/components/tools/OmeroPy/test/integration/scriptstest/test_export_scripts.py
+++ b/components/tools/OmeroPy/test/integration/scriptstest/test_export_scripts.py
@@ -23,6 +23,7 @@
    Integration test for export scripts.
 """
 
+import pytest
 import omero
 import omero.scripts
 from test.integration.scriptstest.script import ScriptTest
@@ -53,6 +54,7 @@ class TestExportScripts(ScriptTest):
         c = self.new_client(user=user)
         check_file_annotation(c, ann)
 
+    @pytest.mark.broken(reason="needs mencoder installed")
     def test_make_movie(self):
         script_id = super(TestExportScripts, self).get_script(make_movie)
         assert script_id > 0


joshmoore:
Another option if you'd prefer (from 2c75a6e2):

```
+    def skipIfNot(self):
+        cfg = self.root.sf.getConfigService()
+        cfg = cfg.getConfigValue("omero.mail.fake")
+        if "true" != str(cfg).lower():
+            pytest.skip("omero.mail.fake not configured")
```

jburel:
excluding this PR for tomorrow's build so that https://github.com/ome/scripts/pull/128 can be tested

@mtbc: python27 job will have a "test-movie failure" since mencoder is not installed. This is expected

jburel:
--exclude

jburel:
removing the exclude.

mtbc:
This PR seems to be working well ?

jburel:
Ideally I would like the test to be run but that requires installation of mencoder
We can keep it open for now

jburel:
The test might be re-activated when carp is up

jburel:
Merging this PR since we also have an annoying failure in devspace.
Hopefully we can find a reliable site to install it from or use another encoder.

components/tools/OmeroPy/test/integration/scriptstest/test_export_scripts.py
```diff
@@ -53,6 +54,7 @@ def test_batch_image_export(self):
         c = self.new_client(user=user)
         check_file_annotation(c, ann)
 
+    @pytest.mark.broken(reason="needs mencoder installed")
```
joshmoore, line 57: NB: `xfail` might have been preferable. On the scale of false positives:
```
broken > skipIf > xfail
```

PR #4976, opened 2016-11-30
history 5.2.7

jburel:
# What this PR does

Add 5.2.7 entry

### DIFF
--- a/history.txt
+++ b/history.txt
@@ -4,7 +4,7 @@ OMERO version history
 5.2.6 (December 2016)
 ---------------------
 
-This is a release aimed at system administrators or developers willing to
+This is a release aimed at system administrators or developers wanting to
 build OMERO with Ice 3.6.3.
 
 5.2.6 (October 2016)
--- a/history.txt
+++ b/history.txt
@@ -7,6 +7,9 @@ OMERO version history
 This is a release aimed at system administrators or developers wanting to
 build OMERO with Ice 3.6.3.
 
+This release also upgrades the version of Bio-Formats which OMERO uses to
+5.1.10. `For more information <https://www.openmicroscopy.org/site/support/bio-formats5.1/about/whats-new.html>`_.
+
 5.2.6 (October 2016)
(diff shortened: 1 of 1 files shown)


history.txt
```diff
@@ -1,8 +1,14 @@
 OMERO version history
 =====================
 
+5.2.6 (December 2016)
+---------------------
+
+This is a release aimed at system administrators or developers willing to
+build OMERO with Ice 3.6.3
```
mtbc: end with period

history.txt
```diff
@@ -1,8 +1,14 @@
 OMERO version history
 =====================
 
+5.2.6 (December 2016)
+---------------------
+
+This is a release aimed at system administrators or developers willing to
```
hflynn: 'wanting' instead of 'willing'? Or are we recommending people use it?

jburel: 3.6 is not the recommended version on 5.2.x
Will replace by "wanting"

hflynn: 👍

hflynn:
Perhaps we should clarify that no other fixes are included and specifically that BF support remains at version 5.1.10(?) since various people have been asking for us to support BF 5.2

jburel:
@hflynn: that's certainly something we need to mention.
Thanks for the reminder

jburel:
The B-F will need to be bumped to 5.1.10 it is currently at 5.1.9
I will add a note about B-F version in history

hflynn:
👍

jburel:
see https://github.com/openmicroscopy/openmicroscopy/pull/4978

hflynn:
If I was being really nitpicky, I'd say it would be better to have `5.1.10` as the link text for the hyperlink to the BF version history and that we don't usually write OME Model with a hyphen but neither of these are blockers to getting this and the accompanying docs PR merged as staging looks fine as is.

jburel:
done

hflynn:
Good to merge

PR #4984, opened 2016-12-05
getHistogram integration test (unit8)

dominikl:
# What this PR does

- Adds another integration test for histogram generation using an `unit8` image.
- Fixes an offset in the histogram bins when `binCount` is bigger than the min/max value range. (see https://github.com/openmicroscopy/openmicroscopy/pull/4984/files#diff-91a64e96538d38370f77e8dc4535df46R731 , if someone has a better idea to address that, please let me know)
- Adds some more `null` checks (~~to prevent NPE like~~ https://github.com/openmicroscopy/openmicroscopy/pull/4931#issuecomment-264214132 - this has been been fixed by last commit)

# Testing this PR

Check that the [test](https://github.com/openmicroscopy/openmicroscopy/blob/e990861486bfc0ff629ef3b4720e52d8a40c678b/components/tools/OmeroJava/test/integration/RawPixelsStoreTest.java#L398) makes sense, and [integration test job](https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-java/) passes.

# Related reading

- [Trello - Histogram](https://trello.com/c/6gKn8FT2/98-histogram)
- https://github.com/openmicroscopy/openmicroscopy/pull/4931#issuecomment-264214132

### DIFF
--- a/components/tools/OmeroJava/test/integration/RawPixelsStoreTest.java
+++ b/components/tools/OmeroJava/test/integration/RawPixelsStoreTest.java
@@ -30,10 +30,8 @@ import ome.api.RawPixelsStore;
 import ome.io.nio.RomioPixelBuffer;
 import omero.ApiUsageException;
 import omero.api.RawPixelsStorePrx;
-import omero.model.IObject;
 import omero.model.Image;
 import omero.model.Pixels;
-import omero.model.PixelsType;
 import omero.romio.PlaneDef;
 import omero.romio.RegionDef;
 
@@ -78,7 +76,7 @@ public class RawPixelsStoreTest extends AbstractServerTest {
 
     @BeforeMethod
     public void localSetUp() throws Exception {
-        localSetUp(1, ModelMockFactory.SIZE_X, ModelMockFactory.SIZE_Y);
+        localSetUp(1, ModelMockFactory.SIZE_X, ModelMockFactory.SIZE_Y, ModelMockFactory.UINT16);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


jburel:
but this is strange that the other method does not allow us to read, an admin should be allowed to do so
could you create a card for that, we will need to look at it

components/tools/OmeroJava/test/integration/RawPixelsStoreTest.java
```diff
@@ -364,55 +367,148 @@ public void testSetRegionEveryPlane() throws Exception {
...
+    public void testGetHistogramFail() throws Exception {
         // Test fail, if called on big images
-        localSetUp(1, 10000, 10000);
+        localSetUp(1, 10000, 10000, ModelMockFactory.UINT16);
         try {
             svc.getHistogram(new int[] { 0 }, -1, true, new PlaneDef(
                     omeis.providers.re.data.PlaneDef.XY, 0, 0, 0, 0, null, -1));
-            Assert.fail("The method getHistogram() can not handle big images and should have thrown an ApiUsageException.");
+            Assert.fail("The method getHistogram() can not handle big images and should "
+                    + "have thrown an ApiUsageException.");
```
mtbc, line 385: :+1: for explanation

dominikl:
The last commit should fix the `null` issue properly. There's no need to load the pixels object again (which can result in being `null`), as it is already loaded on `setPixelsId()`. Just modified the method to also fetch the channel info; I don't think this should cause a decrease in performance, does it?

**Test:** See https://github.com/openmicroscopy/openmicroscopy/pull/4931#issuecomment-264214132 ; As root, try to display the histogram for an image of another user in a private group (which root is not a member of!). Has to be done in OMERO.web, as in Insight you can only do that, when root is a member of this private group, too (which does not trigger the issue).

jburel:
The overall performance "should be" improved since the call to ``retrievePixDescription`` for each channel has been removed.
``retrievePixDescription`` was also loading a larger part of the graph
``
iQuery.findByQuery("select p from Pixels as p "
				+ "left outer join fetch p.pixelsType as pt "
				+ "left outer join fetch p.channels as c "
				+ "left outer join fetch c.logicalChannel as lc "
				+ "left outer join fetch c.statsInfo "
				+ "left outer join fetch lc.photometricInterpretation "
				+ "left outer join fetch lc.illumination "
				+ "left outer join fetch lc.mode "
				+ "left outer join fetch lc.contrastMethod "
				+ "where p.id = :id",
``

jburel:
Have you tried with float images?

dominikl:
Good point. No I didn't test float images. Do we have some example images? Or is there a way to create `.fake` float images?

jburel:
@will-moore created a while ago  a zip of float images
it is on squig: team/will/float-data.zip

I don't think you can create float ``.fake`` images

dominikl:
Thanks. Yes, works fine with float images, too. Also negative float values, e.g. `emd_1001.map`. Visually compared histogram to Icy, looks the same.

jburel:
thanks

PR #4985, opened 2016-12-07
reverting: prevent from caching ajax requests

atarkowska:
# What this PR does

this PR reverts ajax cache to prevent browser from caching ajax response, that will be handled by nginx


joshmoore:
Cache configuration seems to be working in production on demo31

atarkowska:
thx @manics for your kind help

PR #4986, opened 2016-12-08
Ice363 dev 5 2

jburel:
This should be the non breaking change for 5.2.x
The ice dependencies have been renamed ("ice-") and pushed to artifacts.openmicroscopy.org

cc @sbesson @joshmoore 

To test:

 * Download the client from devspace
 * Connect to the server build using ice 3.6.3

ice jars on devspace are of the "form" ``ice-`` e.g. ice-db.jar


dominikl:
Looks good, devspace client working like expected 👍

PR #4987, opened 2016-12-08
Close RawPixelsStore when done (rebased onto metadata52)

snoopycrimecop:
This is the same as gh-4818 but rebased onto metadata52.

----

# What this PR does

Close RawPixelsStore before getPixelRange method returns otherwise the service stays open.
# Testing this PR

Connect to the same session uuid using BlitzGateway and omero.client. Check session.activeService() before and after calling getPixelRange() and make sure the RawPixelsStore is not on the list.
# Related reading


snoopycrimecop:
--rebased-from #4818

joshmoore:
Oops. Opened as snoopy, but I guess no harm in that.

joshmoore:
Looks good. Merging for IDR 0.3.1.

PR #4754, opened 2016-07-21
populate_metadata.py: add batches to write_to_omero

joshmoore:
# What this PR does

The new `QueryContext` implementations in `populate_metadata.py` were not batching in their `write_to_omero` methods. For very large screens (e.g. idr0016):

 * `BulkToMapAnnotationContext` tends to hit MESSAGESIZEMAX limits and
 * `DeleteMapAnnotationContext` hits `Caused by: java.io.IOException: Tried to send an out-of-range integer as a 2-byte value: 109734`

Now both are done in batches of 1000.

*Note:* projections are also causing issues so batching is being added to invocations of `projection` as well.

# Testing this PR

1. A bulk annotation YML configuration is required. This is minimal possible via `test/integration/metadata/test_populate.py`.
2. There should be no change in functionality and no overt slowdown in creation or deletion.

# Related reading

 * https://trello.com/c/urQCwi0v/470-error-creating-map-annotations-in-demo2-for-idr0016
 * https://trello.com/c/88jfqEem/469-error-deleting-map-annotations-for-idr0001-screena

cc: @eleanorwilliams

PR #4988, opened 2016-12-08
Plugin fixes

joshmoore:
# What this PR does

Performance improvements for the new CLI plugins. Goal would be to have all methods be able to function on sizable datasets without timeouts, OOMs, or similar. Primarily, `render.py` should close all stateful services and `metadata.py` should not timeout when deleting, etc. large batches.

# Testing this PR

1. render.py (cc: @emilroz)
   - Primarily, all plugin behavior should remain identical
   - Large re-renderings should no longer bring the server down
1. metadata.py (cc: @aleksandra-tarkowska)
   - allow waiting indefinitely

# Related reading

 * https://trello.com/c/3vGyIffT/192-blitz-gateway-memory-leak-advice
 * https://github.com/openmicroscopy/openmicroscopy/pull/4542 - original render.py PR
 * https://github.com/openmicroscopy/openmicroscopy/pull/4818 - render.py close RPSes
 * https://github.com/openmicroscopy/openmicroscopy/pull/4754 - batch metadata.py calls

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -802,6 +802,12 @@ class BaseControl(object):
             break
         return root_pass
 
+    def _add_wait(self, parser, default=-1):
+        parser.add_argument(
+            "--wait", type=long,
+            help="Number of seconds to wait for the processing to complete "
+            "(Indefinite < 0; No wait=0).", default=default)
+
     def get_subcommands(self):
         """Return a list of subcommands"""
         parser = Parser()
@@ -1560,10 +1566,7 @@ class CmdControl(BaseControl):
 
     def _configure(self, parser):
         parser.set_defaults(func=self.main_method)
-        parser.add_argument(
-            "--wait", type=long,
-            help="Number of seconds to wait for the processing to complete "
-            "(Indefinite < 0; No wait=0).", default=-1)
+        self._add_wait(parser, default=-1)
 
     def main_method(self, args):
         client = self.ctx.conn(args)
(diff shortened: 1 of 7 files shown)


joshmoore:
Pushing further commits to deal with `getChannels` in the gateway. Primary discovery:

```
            img = gw.getObject('Image', iid)
            # Note: calling _prepareRE below does NOT suffice!
            img._prepareRenderingEngine()  # Call *before* getChannels
            # Passing noRE to getChannels below also prevents leaking
            # the RenderingEngine but then Nones are returned later.
            channels = img.getChannels()
```

The only way to prevent leaked REs (that I could discover) when calling `getChannels()` is to manually call `_prepareRenderingEngine` *beforehand*.

joshmoore:
close/open cycle since the last commit doesn't look to be picked up.

components/tools/OmeroPy/src/omero/cli.py
```diff
@@ -802,6 +802,12 @@ def _ask_for_password(self, reason="", root_pass=None, strict=True):
             break
         return root_pass
 
+    def _add_wait(self, parser, default=-1):
```
atarkowska, line 805: I think default could still be 60 sec. Otherwise that is failing right away

joshmoore: Then that's likely something in the logic somewhere (i.e. a bug) since `-1` should mean "wait forever"

joshmoore:
This likely has one small issue when using the default `--wait` of `-1`. Merging for IDR-0.3.1 since `bin/omero render edit` is dangerous without it, but this will need to be fixed for 0.3.2.

PR #4993, opened 2016-12-09
Blitz static get query string

will-moore:
# What this PR does

See discussion at https://github.com/openmicroscopy/openmicroscopy/pull/4950#issuecomment-266041950
Since BlitzGateway ```wrapper_getQueryString()``` doesn't need to be an instance method, this is now a ```classmethod```.

# Testing this PR

 Check that tests are green etc.

cc @aleksandra-tarkowska


will-moore:
Probably this PR is causing this error:
```
 File "/opt/hudson/workspace/WEB-DEV-integration-deploy/OMERO.py/lib/python/omero/gateway/__init__.py", line 774, in listParents
   for x in t])
TypeError: <lambda>() takes exactly 1 argument (2 given)
```

chris-allan:
While I appreciate the work being done here what exactly are we intending to achieve structurally by changing this to a class method? Are we really so sure that we're never going to need anything from the instance to generate the query?

__Edit:__ Moving things out of the bootstrap method also has some pretty serious downstream implications.

will-moore:
I think that ```_getQueryString()``` *has* to be callable without an instance of the object, since we need the query in order to retrieve any objects from the server. If you needed an instance to get the query, then we'd have a chicken and egg problem.
I think the main benefit from these changes is that we wouldn't have to instantiate a dummy object just to get the query. Currently we're doing:
```
wrapper()._getQueryString(opts)
```

This PR would allow us to get the query without creating a wrapper instance (in most cases) see https://github.com/openmicroscopy/openmicroscopy/pull/4993/commits/ab9eba9b97986f65a972699567ee7cc7dd4f075a

chris-allan:
A ha, makes sense. Also then makes complete sense to refactor away the bootstrap method. Do we have any cases where we are using it at all after the changes you've made in 1ae9fff? We also need to know a priori what the wrapper class type mapping is anyway correct? That is, `wrapper()` is providing us with no additional convenience or abstraction?

will-moore:
The error above in https://github.com/openmicroscopy/openmicroscopy/pull/4993#issuecomment-266382166 is caused by removal of ```__bstrap__()``` in ```WellSampleWrapper``` and can be fixed by adding back the ```__bstrap__``` like this:
```
-    LINK_PARENT = lambda x: x
     LINK_CHILD = 'image'
 
+    def __bstrap__(self):
+        self.LINK_PARENT = lambda x: x
```

Or by providing E.g. a static method:
```
-    LINK_PARENT = lambda x: x
     LINK_CHILD = 'image'
+    
+    @staticmethod
+    def LINK_PARENT(parent):
+        return parent
```
I don't know if there's any reason why ```self.LINK_PARENT = lambda x: x``` should be better than this alternative?

The only other remaining use of ```__bstrap__()``` is in ```LaserWrapper```
```
    def __bstrap__(self):
        super(_LaserWrapper, self).__bstrap__()
        self._attrs += (
            '#laserMedium',
            'frequencyMultiplication',
            'tuneable'....
```

chris-allan:
There's already a `LINK_PARENT` static method on `BlitzObjectWrapper`. It's not 100% clear to me exactly what the intention was overriding it this manner in the `WellSampleWrapper`. Certainly overriding the static method with another static method would work. We should probably put test cases in, if we don't have them already, to cover this functionality before messing around with it however; `listParents()` is _super_ old code.

will-moore:
Yes, I'm already writing more tests for ```listParents()```, since we only test it for PDI and it should work for a lot of others.

will-moore:
@chris-allan The ```BlitzObjectWrapper.LINK_PARENT(link)``` returns ```link.parent``` since that is how most objects are linked but there's no link between ```WellSample``` and ```Well```, so ```WellSample.listParents()``` returns ```Wells``` rather than links.

will-moore:
Robot tests that were failing yesterday
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/468/robot/
due to
https://github.com/openmicroscopy/openmicroscopy/pull/4993#issuecomment-266382166 
are now passing
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/469/robot/

will-moore:
testListParents is passing at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-breaking-integration-Python27/127/testReport/OmeroPy.test.integration.gatewaytest.test_get_objects/TestListParents/

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -270,7 +270,8 @@ def _unwrapunits(self, obj, units=None):
             return obj
         return obj.getValue()
 
-    def _getQueryString(self, opts=None):
+    @classmethod
+    def _getQueryString(klass, opts=None):
```
atarkowska: I don't understand arg `klass`, http://legacy.python.org/dev/peps/pep-0008/#function-and-method-arguments

joshmoore: This is probably a historical (i.e. pre-PEP8) Josh/Carlos-style thing which should be dropped in favor of `cls`.

chris-allan: The first argument to a class method is a reference to the class. As `class` is a "reserved" word in Python so we have to use something. `klass` is a classical style that we inherited from Java. `cls` is also fine and in line with the PEP8 suggestions as @joshmoore has already mentioned.

chris-allan (approved):

jburel:
Reviewed with @chris-allan, merging.

PR #5000, opened 2016-12-14
Zoom scale bar

dominikl:
# What this PR does

Bugfix: For big images the zoom factor for drawing the scale bar was not set correctly.

# Testing this PR

Open a big (tiled) image. Display the scale bar. Zoom in/out. Make sure the scale bar is zoomed correctly, too.

# Related reading

[Trello - Insight Scalebar-big image issue](https://trello.com/c/OwIMt9M2/241-insight-scalebar-big-image-issue)

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/actions/UnitBarSizeAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/actions/UnitBarSizeAction.java
@@ -152,6 +152,32 @@ public class UnitBarSizeAction
     	return values[index];
     }
     
+    /**
+     * Returns the index corresponding to the passed value or
+     * <code>CUSTOMIZED</code>.
+     * 
+     * @param value
+     *            The value to handle.
+     * @return See above.
+     */
+    public static int getIndex(double value) {
+        if (value < values[TWO])
+            return ONE;
+        if (value < values[FIVE])
+            return TWO;
+        if (value < values[TEN])
+            return FIVE;
+        if (value < values[TWENTY])
+            return TEN;
+        if (value < values[FIFTY])
+            return TWENTY;
+        if (value < values[HUNDRED])
+            return FIFTY;
+        if (value < 1000)
+            return HUNDRED;
+        return CUSTOMIZED;
+    }
+    
     /** One of the constant defined by this class. */
     private int     index;
     
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/Browser.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/Browser.java
@@ -30,6 +30,7 @@ import java.awt.image.BufferedImage;
 import javax.swing.Icon;
 import javax.swing.JComponent;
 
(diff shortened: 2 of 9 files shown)


will-moore:
Tried with an SVS file and I don't see the scalebar atall with latest build. b512. On previous build b510 I see it displayed for the same image but it doesn't zoom (as known).

dominikl:
Is it possible that it's just too small? I think the default is 5um, that might be too small to display it.

will-moore:
Ah, yes that's it.
When set to 100um then I see the scale bar and it is re-sizing on zoom. Seems to change correctly according to zoom levels.
If I Hide and then Show the scale-bar it reverts to 5um, even though I still have it set to 100um

screen shot 2016-12-16 at 00 28 28

It would be nice if the auto chosen size wasn't so small it's invisible, but I guess if you're familiar with your own data then you'd pick an appropriate scalebar size (and this may only be an issue on Big images).

dominikl:
Yes, I'll do that. Shouldn't be too much work, will check how this is handled in Web.

dominikl:
**Test** last commit:
Open some images with different sizes and perform the "Show scale bar" action on different zoom levels. Check that the default scale bar length is reasonable and the correct scale bar length drop down menu is ticked ( for scale bars >= 1000 the "Custom" menu should be ticked).

If you change the scale bar length, the setting is not preserved, ie. if you untick "Show scale bar" and tick it again, a new default length will be set according to the current zoom level. But this can be seen as feature, too, not necessarily a bug :-)

jburel:
Tested with Today's build
The unit in the menu does not match the unit display as a scale bar.
see screenshot (user-1)
scalebar

dominikl:
Is there a way to reproduce that? I think I tested it with the same image, can't get the discrepancy between menu and display. But noticed another bug: When the image viewer window is bigger than the image, the scalebar is drawn outside of the image and hidden beneath the background.

jburel:
I just opened the image on eel using user-1

dominikl:
Hm, when I open the image (id 1474 isn't it?) and tick the "Show scale bar" menu, I get a scale bar of 1000000 um and the "Custom" scale bar length menu ticked. I just have to zoom in a bit to see the scale bar, because it's drawn offside the image.

jburel:
it is correctly ticked, that's not the problem
but the units in the scale bar is in ``mm`` which is not the unit used when you display

dominikl:
Ah ok, thanks!

jburel:
I still have the same mistmatch between display and unit in menu
I also have the following error.
pr5000

jburel:
if it is too much to fix, let's not spend too long on it

dominikl:
Last commit, a final attempt the get the scale bar behaving properly...

waxenegger:
Could not find anything suspicious. Checked the 2 images used above. Both had scalebars that showed the correct length, checking against measured lines, and they had the color and length I set, as well as the matching unit from the toolbox.

scalebar1

scalebar2

jburel:
A minor thing noticed (dimension next to z in status bar) could match the scale bar but since insight is in maintenance mode, not worth spending time on it
The major issue is fixed. Merging

PR #5002, opened 2016-12-15
Disable copy rnd on multiselection

dominikl:
# What this PR does

Disables the "copy rendering settings" context menu, when there is more than one image selected. *Not sure if that is the right approach to do these things...*

# Testing this PR

Select multiple images. Make sure the "Copy Renderings Settings" option in the context menu is disabled. Select one image. Make sure the option is enabled again. Test with selection in tree as well as center panel.

# Related reading

[Trello - Web bugs](https://trello.com/c/F4c6yUZz/183-web-bugs)

/cc @will-moore


waxenegger:
Works as intended. Copy is possible on single selection but not for multiple.

image

PR #5007, opened 2017-01-04
display PhysicalPixelSize value only if set

atarkowska:
# What this PR does

Fixes invalid value displayed when PhysicalPixelSize is not set
https://trello.com/c/vQ6uJKoJ/24-pixel-sizes-in-idr0009-are-all-0

# Testing this PR

1. check images with or without Pixel sizes

PR #5008, opened 2017-01-04
display PhysicalPixelSize value only if set (rebased onto develop)

atarkowska:
This is the same as gh-5007 but rebased onto develop.

----

# What this PR does

Fixes invalid value displayed when PhysicalPixelSize is not set
https://trello.com/c/vQ6uJKoJ/24-pixel-sizes-in-idr0009-are-all-0

# Testing this PR

1. check images with or without Pixel sizes

cc @will-moore

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/core_metadata.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/core_metadata.html
@@ -46,8 +46,19 @@
         <td id='pixels_size'>
             <div class='tooltip'>
                 {% if vX or vY or vZ %}
-                    {{ vX|floatformat:2|default:'-' }} {% if uX != uY or uX != uZ %}({{ uX }}){% endif %}
-                    x {{ vY|floatformat:2|default:'-' }} {% if uX != uY or uX != uZ %}({{ uY }}){% endif %}
+                    {% with vXd=vX|floatformat:2|default:'-' %}
+                        {{ vXd }}
+                        {% if uX != uY or uX != uZ %}
+                            {% if vXd != '-' %} ({{ uX }}) {% endif %}
+                        {% endif %}
+                    {% endwith %}
+                    x 
+                    {% with vYd=vY|floatformat:2|default:'-' %}
+                        {{ vYd }}
+                        {% if uX != uY or uX != uZ %}
+                            {% if vYd != '-' %} ({{ uY }}) {% endif %}
+                        {% endif %}
+                    {% endwith %}
                     {% if vZ %}
                         x {{ vZ|floatformat:2|default:'-' }} {% if uX != uY or uX != uZ %}({{ uZ }}){% endif %}
                     {% endif %}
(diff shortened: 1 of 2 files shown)


atarkowska:
--rebased-from #5007

atarkowska:
just discussed UI with @will-moore, `not available` will go to tooltip and we will display `-` instead. if statement is also missing Z

will-moore:
Travis failure, but otherwise good to merge - see https://github.com/openmicroscopy/openmicroscopy/pull/5007#issuecomment-270640224

jburel:
looking at https://github.com/openmicroscopy/openmicroscopy/pull/5007#issuecomment-270640224
why do you display a unit if the value is not available?
for me, ``-x-x value(unit)`` will make more sense

jburel:
I also find confusing to have Pixels Size (XYZ) and only have 2 values if Z is not set
I will expect ``value x value x -``
sizedisplay

atarkowska:
@jburel unit is provided always, that is how it was designed https://github.com/openmicroscopy/openmicroscopy/pull/5008/files#diff-7bcff1e41a9e76739c322db1ca40c091R2443. I haven't change the UI. I don't know the story behind the logic here and why that was implemented that way. If you feel existing way of displaying values should be improved that would have to be discussed, start design issue?

jburel:
It is not always displayed next to XYZ if this what you mean. The change to ``-`` with unit next to it when the value is not valable should be only ``-`` with no unit. This should be tackled as part of the fix
I don't think there is a need to create a design issue

atarkowska:
I think they appear when units are different for each dimension. My impression is that there were specific decisions made to achieve that layout. Could you come back to the original design and provide detailed spec what exactly you would like to be changed? This is bug fixing PR.

jburel:
I will try to go over old decisions but I don't think we had any detailed spec at the time.
Displaying the units next to the value when we have different values is absolutely fine.
The point I am making is that we do not have any value but we have a unit. That does not really make sense. 
Previously we had 0.0 (which was wrong and fix here) but at least it was a value e.g. 0.0 nm and that was ok. Your fix changes the display so ``- nm`` does not seem correct

jburel:
PR with suggested changes https://github.com/aleksandra-tarkowska/openmicroscopy/pull/19

jburel:
@aleksandra-tarkowska: ok to remove the exclude label?

will-moore:
@jburel. It may be more technically correct to have `-` for pixel size Z if not set, but it doesn't look so nice alongside the sizeX x sizeY above. 

screen shot 2017-01-11 at 10 40 07

With non-standard units, when values are missing units are not now displayed:
screen shot 2017-01-11 at 10 41 53

But in fixing this edge case I think the more common use case above is less nice.
However, if it's agreed that this is how it should be then "good to merge" from me.

jburel:
if we do not want to have the ``-`` at the end, we should in that case only have ``Pixels Size (XY):`` 
otherwise I think it is confusing.

atarkowska:
thx all, rebased now!

PR #5010, opened 2017-01-04
Bugfix: Save rendering settings

dominikl:
# What this PR does

Fixes two bugs with respect to saving rendering settings:
- The up/down buttons mentioned on the Trello card [web rendering details up/down buttons don't enable the save icon](https://trello.com/c/DFN0mRSf/135-web-rendering-details-up-down-buttons-don-t-enable-the-save-icon) don't exist any more, but there was still a similar issue: Changing the value in the text field and then moving the cursor somewhere else (without hitting the enter key), didn't update the preview image, respectively the save button. With this PR the preview image and save button also gets updated in these cases (ie. on focus loss of the text field).
- If an inactive channel's start and/or end value was changed, this change didn't get saved, see [Trello -  Critical: channel color change is not saved for inactive channels](https://trello.com/c/p5ujlK67/559-critical-channel-color-change-is-not-saved-for-inactive-channels)

# Testing this PR

- Change a channel's start and/or end value via the text field. Make sure the save button is enabled when the cursor is removed from the text field.
- Change an **inactive** channel's start and/or end value. Save the rendering settings. Make sure the settings really have been saved.
- Check that robot test 'Test Rdef Save' passes

# Related reading

- [Trello - web rendering details up/down buttons don't enable the save icon](https://trello.com/c/DFN0mRSf/135-web-rendering-details-up-down-buttons-don-t-enable-the-save-icon) 
- [Trello -  Critical: channel color change is not saved for inactive channels](https://trello.com/c/p5ujlK67/559-critical-channel-color-change-is-not-saved-for-inactive-channels)

/cc @will-moore

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -774,8 +774,16 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
         return
     if 'c' in r:
         logger.debug("c="+r['c'])
-        channels, windows, colors, reverses = _split_channel_info(r['c'])
-        if not img.setActiveChannels(channels, windows, colors, reverses):
+        activechannels, windows, colors, reverses = _split_channel_info(r['c'])
+        allchannels = []
+        for ch in activechannels:
+            allchannels.append(abs(int(ch)))
+        # First save properties of all channels
+        if not img.setActiveChannels(allchannels, windows, colors, reverses):
+            logger.debug(
+                "Something bad happened while setting the active channels...")
+        # Save the active/inactive state of the channels
+        if not img.setActiveChannels(activechannels):
             logger.debug(
                 "Something bad happened while setting the active channels...")
     if r.get('m', None) == 'g':
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -779,11 +779,11 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
         for ch in activechannels:
             allchannels.append(abs(int(ch)))
         # First save properties of all channels
-        if not img.setActiveChannels(allchannels, windows, colors, reverses):
+        if saveDefs and not img.setActiveChannels(allchannels, windows, colors, reverses):
             logger.debug(
                 "Something bad happened while setting the active channels...")
         # Save the active/inactive state of the channels
-        if not img.setActiveChannels(activechannels):
+        if not img.setActiveChannels(activechannels, windows, colors, reverses):
             logger.debug(
                 "Something bad happened while setting the active channels...")
     if r.get('m', None) == 'g':
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -779,11 +779,13 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
         for ch in activechannels:
             allchannels.append(abs(int(ch)))
         # First save properties of all channels
-        if saveDefs and not img.setActiveChannels(allchannels, windows, colors, reverses):
+        if saveDefs and not img.setActiveChannels(allchannels, windows,
+                                                  colors, reverses):
             logger.debug(
                 "Something bad happened while setting the active channels...")
         # Save the active/inactive state of the channels
-        if not img.setActiveChannels(activechannels, windows, colors, reverses):
+        if not img.setActiveChannels(activechannels, windows, colors,
+                                     reverses):
             logger.debug(
                 "Something bad happened while setting the active channels...")
     if r.get('m', None) == 'g':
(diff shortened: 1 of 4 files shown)


jburel:
Is the PR ready for review @dominikl?

dominikl:
Yes, can be reviewed. Apparently forgot to re-list it in standup...

will-moore:
I think we don't want to iterate through all the ```setActiveChannels()``` twice for every call of ```_get_prepared_image()``` since we call that a lot and we're mostly just interested in rendering active channels.

A question: Should we change the behaviour of ```conn.setActiveChannels()``` so that color, LUTs, channelWindow, codomain (reverse Intensity) are updated for inactive channels?
cc @chris-allan @aleksandra-tarkowska 

This would be a simple fix, but might slow performance. e.g. when rendering a single channel from a *many* channel image.
If not, then simply moving the ```img.setActiveChannels(allchannels, ...)``` down under the ```if saveDefs:``` would be an improvement, so we only call it when we're saving.

jburel:
The strategy will not work for example for some life time images

will-moore:
The focus-out events update image and Save as expected.
One potential change we could make here is to use chaining to avoid calling the jQuery selector for each event we want to bind. E.g.
```
            $('#wblitz-ch'+i+'-cw-start')
                .val(channels[i].window.start)
                .unbind('change').bind('change', start_cb(i))
                .keyup(keyup_cb())
                .focusout(focusout_cb());
```
Although the performance gains are negligible, this is slightly clearer when reading it that all the events etc are on the same element and if we change the selector it's only 1 place to update.

will-moore:
@dominikl I guess you don't want to move your ```img.setActiveChannels(allchannels, ...)``` down because it will be called in the wrong order, so just and an extra ```if saveDefs:``` in place above it.

dominikl:
I think `img.setActiveChannels(allchannels, ...)` is in the right place, has to be called before `img.setActiveChannels(activechannels)`, because after `img.setActiveChannels(allchannels, ...)` all channels will be enabled. But yes, I probably only have to call that if `saveDefs` is true.

dominikl:
Deleted my previous comment, because I just misread your comment above @will-moore . But still, can't add `if saveDefs`, because then preview doesn't get updated when the slider or value is changed.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -774,8 +774,16 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
...
     if 'c' in r:
         logger.debug("c="+r['c'])
-        channels, windows, colors, reverses = _split_channel_info(r['c'])
-        if not img.setActiveChannels(channels, windows, colors, reverses):
+        activechannels, windows, colors, reverses = _split_channel_info(r['c'])
+        allchannels = []
+        for ch in activechannels:
+            allchannels.append(abs(int(ch)))
+        # First save properties of all channels
+        if not img.setActiveChannels(allchannels, windows, colors, reverses):
```
will-moore: I think you want to only do this if we're saving. so
```
if saveDefs and not img.setActiveChannels(allchannels, windows, colors, reverses):
    ....
```

will-moore:
For tests, you're going to need to use Robot framework tests for the JavaScript change and could also include the save inactive channels test here (or use a web integration test).
I think you could add these to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tests/ui/testcases/web/rdef_test.txt
Again, let me know if you need any pointers here.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
dominikl: I added it with the last commit. But what happens now, is that the preview doesn't get updated any more. I don't think that's the expected behaviour.

dominikl:
Ok, the robot test just failed on https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/ ... was pretty sure, it worked locally. Will try again.

will-moore:
New Robot test ```Test Rdef Save``` is passing in both browsers: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/508/robot/
Good to merge.

will-moore:
Oops - Sorry, I missed the fact that you are still calling ```setActiveChannels()``` twice, even when not saving.
To address your comment above about Preview viewer not updating, you need to change the other  call too (see 2 lines changed below):
```
         # First save properties of all channels
-        if not img.setActiveChannels(allchannels, windows, colors, reverses):
+        if saveDefs and not img.setActiveChannels(allchannels, windows, colors, reverses):
             logger.debug(
                 "Something bad happened while setting the active channels...")
         # Save the active/inactive state of the channels
-        if not img.setActiveChannels(activechannels):
+        if not img.setActiveChannels(activechannels, windows, colors, reverses):
```

dominikl:
Ah of course, that works, thx @will-moore .

dominikl:
Just swapped the order of the tests, now they're passing. Re-review necessary?

will-moore:
Looks good. Thanks.
Good to merge.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -774,8 +774,18 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
         return
     if 'c' in r:
         logger.debug("c="+r['c'])
-        channels, windows, colors, reverses = _split_channel_info(r['c'])
-        if not img.setActiveChannels(channels, windows, colors, reverses):
+        activechannels, windows, colors, reverses = _split_channel_info(r['c'])
+        allchannels = []
```
jburel, line 778: @will-moore This code will have to be reviewed in https://github.com/openmicroscopy/openmicroscopy/pull/5082
Merging so you can move forward

PR #5012, opened 2017-01-05
Bump Bio-Formats version to 5.3.2-SNAPSHOT

sbesson:
Prior to the release of Bio-Formats 5.3.2, this PR bumps the version of Bio-formats to consume merge SNAPSHOT. This should allow among others https://github.com/openmicroscopy/bioformats/pull/2714 to be exposed and consumed in the OMERO build and integration tests.


sbesson:
Closing as it this bump impacted the changes in https://github.com/openmicroscopy/openmicroscopy/pull/5016 (`Error fetching URL: http://downloads.openmicroscopy.org/bio-formats/5.3.2-SNAPSHOT/api/`).
This might also mean we should use a different approach when trying to propagate a SNAPSHOT version of Bio-Formats into the OMERO for testing. One solution might be to spin up a dedicated ephemerous devspace as we have successfully done for model changes /cc @pwalczysko @jburel @joshmoore

PR #5017, opened 2017-01-06
Api list datasets screens

will-moore:
# What this PR does

Adds support for listing Datasets, Screens & Plates.
Also adds support for showing (and updating) single Plates.

This uses subclasses of a new ```ObjectsView``` class to reduce code duplication for listing Projects, Datasets, Screens & Plates.
Processing of the ```request``` and any other ```kwargs``` into an ```opts``` dict is handled by the base class and can be extended by subclasses, E.g. to add filtering of Datasets by Project.

# Testing this PR

- Check tests are green.
- Functionality below is largely covered by tests, so just pick a few different options and check that they correspond to what you see in webclient/Insight.
- Browse to /api/ and then click on the 'base_url' and then the 'datasets_url', 'screens_url' and 'plates_url'.
- These should list datasets, screens or plates.
- Filter datasets by project ```/api/v0.1/m/datasets/?project=:project_id```
    - Also try ```/api/v0.1/m/projects/:project_id/datasets/```
- Filter plates by screen ```/api/v0.1/m/plates/?screen=:screen_id```
    - Also try ```/api/v0.1/m/screens/:screen_id/plates/```
- Filter by datasets AND plates to only show orphaned ```?orphaned=true```
- Also test paginate via ```?page=2&limit=5```.
- Can also filter by ```?owner=:expId``` and ```?group=:gid```
- Test display of single Plate via E.g. ```/api/v0.1/m/plates/:plate_id/```
- Create, Update and Delete of single Plate covered by ```test_container_crud```

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -291,7 +291,7 @@ class BlitzObjectWrapper (object):
         """
         extra_select = ""
         child_count = False
-        if opts is not None and 'child_count' in opts:
+        if opts is not None and opts.get('child_count'):
             child_count = opts['child_count']
         if child_count and cls.LINK_CLASS is not None:
             extra_select = """, (select count(id) from %s chl
@@ -5701,7 +5701,7 @@ class _DatasetWrapper (BlitzObjectWrapper):
             query += ' join obj.projectLinks plink'
             clauses.append('plink.parent.id = :pid')
             params.add('pid', rlong(opts['project']))
-        if opts is not None and 'orphaned' in opts and opts['orphaned']:
+        if opts is not None and opts.get('orphaned'):
             clauses.append(
                 """
                 not exists (
@@ -5968,7 +5968,7 @@ class _PlateWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
 
         Also handles filtering of Plates by Screens.
         Returns a tuple of (query, clauses, params).
-        Supported opts: 'plate': <screen_id> to filter by Screen
+        Supported opts: 'screen': <screen_id> to filter by Screen
                         'orphaned': <bool>. Filter by 'not in Screen'
 
         :param opts:        Dictionary of optional parameters.
@@ -5986,7 +5986,7 @@ class _PlateWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         if opts is not None and 'screen' in opts:
             clauses.append('spl.parent.id = :sid')
             params.add('sid', rlong(opts['screen']))
-        if opts is not None and 'orphaned' in opts and opts['orphaned']:
+        if opts is not None and opts.get('orphaned'):
             clauses.append(
                 """
                 not exists (
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5984,8 +5984,7 @@ class _PlateWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         clauses = []
         params = omero.sys.ParametersI()
         if opts is not None and 'screen' in opts:
-            query += ' join obj.screenLinks slink'
-            clauses.append('slink.parent.id = :sid')
+            clauses.append('spl.parent.id = :sid')
             params.add('sid', rlong(opts['screen']))
         if opts is not None and 'orphaned' in opts and opts['orphaned']:
             clauses.append(
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5684,6 +5684,7 @@ class _DatasetWrapper (BlitzObjectWrapper):
         Extend base query to handle filtering of Datasets by Projects.
         Returns a tuple of (query, clauses, params).
         Supported opts: 'project': <project_id> to filter by Project
+                        'orphaned': <bool>. Filter by 'not in Project'
 
         :param opts:        Dictionary of optional parameters.
         :return:            Tuple of string, list, ParametersI
@@ -5694,6 +5695,15 @@ class _DatasetWrapper (BlitzObjectWrapper):
             query += ' join obj.projectLinks plink'
             clauses.append('plink.parent.id = :pid')
             params.add('pid', rlong(opts['project']))
+        if opts is not None and 'orphaned' in opts and opts['orphaned']:
+            clauses.append(
+                """
+                not exists (
+                    select pdlink from ProjectDatasetLink as pdlink
+                    where pdlink.child = obj.id
+                )
+                """
+            )
         return (query, clauses, params)
 
     def __loadedHotSwap__(self):
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5960,12 +5960,14 @@ class _PlateWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
 
     def _getQueryString(self, opts=None):
         """
-        Returns a query string for constructing custom queries,
-        loading the screen for each plate.
+        Custom query to load Screen with Plate.
+
+        Also handles filtering of Plates by Screens.
         Returns a tuple of (query, clauses, params).
+        Supported opts: 'plate': <screen_id> to filter by Screen
+                        'orphaned': <bool>. Filter by 'not in Screen'
 
         :param opts:        Dictionary of optional parameters.
-                            NB: No options supported for this class.
         :return:            Tuple of string, list, ParametersI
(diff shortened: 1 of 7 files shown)


chris-allan:
Are the various test cases that you have outlined in the description covered by integration tests?

will-moore:
@chris-allan Mostly yes.
- In ```test_api_login.py``` we test for presence of the various urls, but don't actually follow/check them
- Listing of Screens is covered by ```test_screens()```.
- Listing of Datasets and filtering by Project is covered by ```test_datasets_plates()```.
- Listing of Plates and filtering by Screen is also covered by ```test_datasets_plates()```.
- These tests also include pagination testing.
- Create, Update, Read and Delete of a single Project, Dataset, Screen & Plate is covered by ```test_container_crud.py```.
- Both GET of individual objects via ```ObjectView``` and listing of objects via ```ObjectsView``` are covered by lots of tests on Project in ```test_api_projects.py```. These are not all duplicated for all the other objects but I think this is OK since all the subclasses of these view classes are are very similar. E.g. ```normalize``` of marshalling and filter by group/owner is identical in all objects.
 - I could add tests for ```childCount``` which are more Object-dependent.

What's missing is gateway tests for the changes in Blitz Gateway (although this functionality is covered by the tests above). I would have added these tests but the existing gateway test infrastructure doesn't include creation of Screens & Plates as it does for Projects & Datasets, so this would require a bit more work (in another PR).

will-moore:
@chris-allan Added test for childCount of images in Datasets.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5954,20 +5964,38 @@ def exportOmeTiff(self):
     @classmethod
     def _getQueryString(cls, opts=None):
         """
-        Returns a query string for constructing custom queries,
-        loading the screen for each plate.
+        Custom query to load Screen with Plate.
+
+        Also handles filtering of Plates by Screens.
         Returns a tuple of (query, clauses, params).
+        Supported opts: 'plate': <screen_id> to filter by Screen
```
jburel: shouldn't this be 'screen'?

chris-allan:
Great, thanks. That's very helpful.

So you would say that during review we're not relying on the reviewers to test each one of the conditions outlined by hand? In other words: you're actually confident in the PRs fitness for purpose based on the unit and integration tests alone, yes? By extension, we should then not have to depend on someone painstakingly testing all these conditions by hand in order to satisfactorily merge this PR.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5700,6 +5701,15 @@ def _getQueryString(cls, opts=None):
             query += ' join obj.projectLinks plink'
             clauses.append('plink.parent.id = :pid')
             params.add('pid', rlong(opts['project']))
+        if opts is not None and 'orphaned' in opts and opts['orphaned']:
```
chris-allan: Just as a matter of style, you could use the construct `opt.get('orphaned')` here.

will-moore:
@chris-allan That's correct. I would just check one or two queries with one or two parameters just to see for yourself that it's working (maybe compare with what you see in webclient).

components/tools/OmeroWeb/omeroweb/api/api_query.py
```diff
@@ -19,74 +19,52 @@
...
-            %s
-            order by lower(project.name), project.id
-            """ % (withChildCount, where_clause)
+    qs = conn.getQueryService()
 
-    projects = []
+    objects = []
     extras = {}
-    if childCount:
+    if opts is not None and 'child_count' in opts and opts['child_count']:
```
chris-allan: As above.

chris-allan:
Awesome, sounds good. I'll leave another member of the team to do that sanity check.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5954,20 +5964,38 @@ def exportOmeTiff(self):
...
                  "left outer join fetch spl.parent sc")
-        return query, [], omero.sys.ParametersI()
+        # NB: we don't use base _getQueryString.
+        # since child_count wouldn't be supported anyway (no LINK_CLASS)
+        clauses = []
+        params = omero.sys.ParametersI()
+        if opts is not None and 'screen' in opts:
+            clauses.append('spl.parent.id = :sid')
+            params.add('sid', rlong(opts['screen']))
+        if opts is not None and 'orphaned' in opts and opts['orphaned']:
```
jburel: same possible style change as suggested by @chris-allan

components/tools/OmeroWeb/omeroweb/api/api_query.py
```diff
@@ -19,74 +19,52 @@
...
 
     @param conn:        BlitzGateway
-    @param childCount:  If true, also load Dataset counts as omero:childCount
-    @param group:       Filter by group Id
-    @param owner:       Filter by owner Id
-    @param page:        Pagination page. Default is 1
-    @param limit:       Page size
-    @param normalize:   If true, marshal groups and experimenters separately
+    @param object_type: Type to query. E.g. Project
+    @param opts:        Options dict for conn.buildQuery()
```
jburel, line 42: why do you remove the param ``group`` and ``normalize`` from the description?

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel: Same style suggestion could apply here

components/tools/OmeroWeb/omeroweb/api/views.py
```diff
@@ -159,37 +162,97 @@ class ScreenView(ObjectView):
...
+                }
+        return opts
+
+    def get(self, request, conn=None, **kwargs):
+        """GET a list of Projects, filtering by various request parameters."""
+        opts = self.get_opts(request, **kwargs)
+        group = getIntOrDefault(request, 'group', -1)
+        normalize = request.GET.get('normalize', False) == 'true'
+        # Get the data
+        marshalled = query_objects(conn, self.OMERO_TYPE,
```
jburel: maybe return the result of ``query_objects`` to avoid assignment

components/tools/OmeroWeb/test/integration/test_api_containers.py
```diff
@@ -38,12 +60,15 @@ def marshal_objects(objects):
...
-                   group='-1'):
+                   group='-1', extra=None):
     """
     Load objects from OMERO, via conn.getObjects().
 
     marshal with omero_marshal and compare with json_objects.
     omero_ids_objects can be IDs or list of omero.model objects.
+
+    @param: extra       List of dicts containing expected extra json data
+                        E.g. {'omero:childCount': 1}
```
jburel: typo ``e.g``

jburel:
We may want an ``api_screen_plates`` to match ``api_project_datasets``

components/tools/OmeroWeb/omeroweb/api/api_query.py
will-moore: Refactor error - I'll add them back.

will-moore:
@jburel Yes, the ```api_project_datasets``` allows you to use ```/projects/:pid/datasets/``` instead of ```/datasets/?project=:pid``` which is a bit more natural if you're coming from the parent at ```/projects/:pid```.
However, I mostly added it in this PR as a proof that I could use the same ```DatasetsView``` class for both, handling the filtering in ```get_opts```.
But since this is working fine and it's a useful feature, I'll add ```api_screen_plates``` too.

chris-allan (approved):
Good to merge once we have the go ahead from an in browser sanity check.

components/tools/OmeroWeb/omeroweb/api/urls.py
```diff
@@ -128,6 +128,15 @@
 GET all plates, using omero-marshal to generate json
 """
 
+api_screen_plates = url(
+    r'^v(?P<api_version>%s)/m/screens/'
+    '(?P<screen_id>[0-9]+)/plates/$' % versions,
+    views.PlatesView.as_view(),
+    name='api_screen_plates')
+"""
+GET Datasets in Project, using omero-marshal to generate json
```
jburel: Doc not correct: Should be ``Plates in Screen``

jburel:
Could you add test for ``projects`` similar to ``test_screens``?
Also I cannot see any tests for the retrieval of datasets within a project (same for screen/plate) unless I am missing something

will-moore:
Listing of projects is already covered in ```test_api_projects.py``` from previous PRs.
Testing of the ```/projects/:id/datasets/``` url is not in yet.
I was going to add more tests for that in a follow-up PR that also adds that url to the parent project json.

I could do it in this PR, but would prefer to get this merged since it's blocking #5025.

jburel:
If you add the tests in the follow-up that will be great. 
Done some testing in the browser and did not notice any problems
Merging so we can move with the next PR.

PR #5019, opened 2017-01-09
add missing version to javascript links

atarkowska:
that never been backported to m52
--no-rebase


will-moore:
This now matches the state of develop. Good to merge.

PR #5020, opened 2017-01-09
remove cmdtest Python integration tests

mtbc:
There appears to be nothing of real value in `components/tools/OmeroPy/test/integration/cmdtest/`. This PR thus deletes it all.


will-moore:
Agreed - doesn't appear to be anything useful in these tests.
Good to merge.

PR #5021, opened 2017-01-09
trim non-existent Java packages from logback.xml

mtbc:
`logback.xml` configures the logging level for various Java classes. This PR removes the `ome.services.chgrp` and `ome.services.chmod` packages from its configuration because there are no longer any classes in those packages.

### DIFF
--- a/etc/logback.xml
+++ b/etc/logback.xml
@@ -89,8 +89,6 @@
   <logger name="omero.cmd" level="INFO"/>
   <logger name="omero.cmd.graphs" level="INFO"/>
   <logger name="ome.services.graphs" level="INFO"/>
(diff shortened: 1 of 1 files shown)


dominikl:
👍  Makes sense, these packages don't exist any more. Good to merge from my point of view.

joshmoore:
Are there any new packages that would cover these use cases? For debugging purpose (back in the day) it was definitely useful to tell people to flip these switches. But if there's no clear replacement with Graphs2, then GTM.

mtbc:
Good question. None come to mind. Might be interesting to think about if specific classes, not just packages, could be useful to add someday. Though, we dream of the day when that gets to the top of our to-do list!

PR #5023, opened 2017-01-11
Annotation count web

dominikl:
# What this PR does

Show the number of annotations available in the right hand side panel.

screen shot 2017-01-11 at 11 03 18

# Testing this PR

Annotate different objects (Images, Datasets, etc.). Select these objects (also some non-annotated objects) and make sure the annotation count shown is correct. Test with single as well as multi (batch) selection.

# Related reading

[Trello - RFE: Indication of number of annotations on an image in brackets (5) on accordion tabs in right hand pane.](https://trello.com/c/bc6ZtV0x/144-rfe-indication-of-number-of-annotations-on-an-image-in-brackets-5-on-accordion-tabs-in-right-hand-pane)

/cc @will-moore

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3425,7 +3425,7 @@ class _BlitzGateway (object):
             return counts
 
         ctx = self.SERVICE_OPTS.copy()
-        ctx.setOmeroGroup(None)
+        ctx.setOmeroGroup(-1)
 
         params = omero.sys.ParametersI()
         params.addIds(obj_ids)
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3425,7 +3425,7 @@ class _BlitzGateway (object):
             return counts
 
         ctx = self.SERVICE_OPTS.copy()
-        ctx.setOmeroGroup(self.group)
+        ctx.setOmeroGroup(None)
 
         params = omero.sys.ParametersI()
         params.addIds(obj_ids)
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3454,18 +3454,18 @@ class _BlitzGateway (object):
         for r in queryResult:
             ur = unwrap(r)
             if ur[3] == '/basic/num/long/':
-                counts['LongAnnotation'] += int(ur[0])
-                counts['OtherAnnotation'] += int(ur[1])
+                counts['LongAnnotation'] += ur[0]
+                counts['OtherAnnotation'] += ur[1]
             elif ur[3] == '/basic/text/comment/':
-                counts['CommentAnnotation'] += int(ur[2])
+                counts['CommentAnnotation'] += ur[2]
             elif ur[3] == '/basic/text/tag/':
-                counts['TagAnnotation'] += int(ur[2])
+                counts['TagAnnotation'] += ur[2]
             elif ur[3] == '/type/OriginalFile/':
-                counts['FileAnnotation'] += int(ur[2])
+                counts['FileAnnotation'] += ur[2]
             elif ur[3] == '/map/':
-                counts['MapAnnotation'] += int(ur[2])
+                counts['MapAnnotation'] += ur[2]
             else:
-                counts['OtherAnnotation'] += int(ur[2])
+                counts['OtherAnnotation'] += ur[2]
 
         return counts
 
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1014,6 +1014,9 @@ class BlitzObjectWrapper (object):
         Get the annotion counts for the current object
         """
 
+        if self.OMERO_CLASS is None:
+            return countAnnotations()
+
         ctx = self._conn.SERVICE_OPTS.copy()
         ctx.setOmeroGroup(self.details.group.id.val)
 
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -141,71 +141,6 @@ def fileread_gen(fin, fsize, bufsize):
     fin.close()
 
 
-def countAnnotations(queryService, ctx, obj_type, obj_ids=[]):
-    """
-    Count the annotions linked to the given object
-
-    :param queryService:  Reference to the query service
-    :param ctx:           The security context
-    :param obj_type:      The type of the object the annotations are linked to
-    :param obj_ids:       List of object ids
-    :return:              Dictionary of annotation counts per annotation type
-    """
-
-    counts = {
-        "TagAnnotation": 0,
-        "FileAnnotation": 0,
-        "CommentAnnotation": 0,
-        "LongAnnotation": 0,
-        "MapAnnotation": 0,
-        "OtherAnnotation": 0}
-
-    if obj_type is None or not obj_ids:
-        return counts
-
-    params = omero.sys.ParametersI()
-    params.addIds(obj_ids)
-    params.add('ratingns',
-               rstring(omero.constants.metadata.NSINSIGHTRATING))
-
-    q = """
-        select sum( case when an.ns = :ratingns
-                    and an.class = LongAnnotation
-                    then 1 else 0 end),
-            sum( case when (an.ns is null or an.ns != :ratingns)
-                           and an.class = LongAnnotation
-                           then 1 else 0 end),
-           sum( case when an.class != LongAnnotation
-                    then 1 else 0 end ), an.class
-           from Annotation an where an.id in
-                (select distinct(ann.id) from %sAnnotationLink ial
-                    join ial.child as ann
-                    join ial.parent as i
-            where i.id in (:ids))
-                group by an.class
-        """ % obj_type
-
-    queryResult = queryService.projection(q, params, ctx)
-
-    for r in queryResult:
-        ur = unwrap(r)
-        if ur[3] == '/basic/num/long/':
-            counts['LongAnnotation'] += int(ur[0])
-            counts['OtherAnnotation'] += int(ur[1])
-        elif ur[3] == '/basic/text/comment/':
-            counts['CommentAnnotation'] += int(ur[2])
-        elif ur[3] == '/basic/text/tag/':
-            counts['TagAnnotation'] += int(ur[2])
-        elif ur[3] == '/type/OriginalFile/':
-            counts['FileAnnotation'] += int(ur[2])
-        elif ur[3] == '/map/':
-            counts['MapAnnotation'] += int(ur[2])
-        else:
-            counts['OtherAnnotation'] += int(ur[2])
-
-    return counts
-
-
 def getAnnotationLinkTableName(objecttype):
     """
     Get the name of the *AnnotationLink table
@@ -3440,7 +3375,7 @@ class _BlitzGateway (object):
                                and an.class = LongAnnotation
                                then 1 else 0 end),
                sum( case when an.class != LongAnnotation
-                        then 1 else 0 end ), an.class
+                        then 1 else 0 end ), type(an.class)
                from Annotation an where an.id in
                     (select distinct(ann.id) from %sAnnotationLink ial
                         join ial.child as ann
@@ -3453,16 +3388,16 @@ class _BlitzGateway (object):
 
         for r in queryResult:
             ur = unwrap(r)
-            if ur[3] == '/basic/num/long/':
+            if ur[3] == 'ome.model.annotations.LongAnnotation':
                 counts['LongAnnotation'] += ur[0]
                 counts['OtherAnnotation'] += ur[1]
-            elif ur[3] == '/basic/text/comment/':
+            elif ur[3] == 'ome.model.annotations.CommentAnnotation':
                 counts['CommentAnnotation'] += ur[2]
-            elif ur[3] == '/basic/text/tag/':
+            elif ur[3] == 'ome.model.annotations.TagAnnotation':
                 counts['TagAnnotation'] += ur[2]
-            elif ur[3] == '/type/OriginalFile/':
+            elif ur[3] == 'ome.model.annotations.FileAnnotation':
                 counts['FileAnnotation'] += ur[2]
-            elif ur[3] == '/map/':
+            elif ur[3] == 'ome.model.annotations.MapAnnotation':
                 counts['MapAnnotation'] += ur[2]
             else:
                 counts['OtherAnnotation'] += ur[2]
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1021,7 +1021,7 @@ class BlitzObjectWrapper (object):
 
     def getAnnotationCount(self):
         """
-        Loads the annotion counts for the current object
+        Get the annotion counts for the current object
         """
 
         params = omero.sys.ParametersI()
@@ -3478,7 +3478,7 @@ class _BlitzGateway (object):
 
     def countAnnotations(self, objects):
         """
-        Loads the annotion counts for the given objects
+        Get the annotion counts for the given objects
         """
 
         obj_type = None
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1046,11 +1046,7 @@ class BlitzObjectWrapper (object):
         Get the annotion counts for the current object
         """
 
-        ctx = self._conn.SERVICE_OPTS.copy()
-        ctx.setOmeroGroup(self.details.group.id.val)
-
-        return countAnnotations(self._conn.getQueryService(), ctx,
-                                self.OMERO_CLASS, [self._oid])
+        return self._conn.countAnnotations(self.OMERO_CLASS, [self.getId()])
 
     def listAnnotations(self, ns=None):
         """
@@ -3408,6 +3404,71 @@ class _BlitzGateway (object):
         for r in result:
             yield AnnotationLinkWrapper(self, r)
 
+    def countAnnotations(self, obj_type, obj_ids=[]):
+        """
+        Count the annotions linked to the given objects
+
+        :param obj_type:   The type of the object the annotations are linked to
+        :param obj_ids:    List of object ids
+        :return:           Dictionary of annotation counts per annotation type
+        """
+
+        counts = {
+            "TagAnnotation": 0,
+            "FileAnnotation": 0,
+            "CommentAnnotation": 0,
+            "LongAnnotation": 0,
+            "MapAnnotation": 0,
+            "OtherAnnotation": 0}
+
+        if obj_type is None or not obj_ids:
+            return counts
+
+        ctx = self.SERVICE_OPTS.copy()
+        ctx.setOmeroGroup(self.group)
+
+        params = omero.sys.ParametersI()
+        params.addIds(obj_ids)
+        params.add('ratingns',
+                   rstring(omero.constants.metadata.NSINSIGHTRATING))
+
+        q = """
+            select sum( case when an.ns = :ratingns
+                        and an.class = LongAnnotation
+                        then 1 else 0 end),
+                sum( case when (an.ns is null or an.ns != :ratingns)
+                               and an.class = LongAnnotation
+                               then 1 else 0 end),
+               sum( case when an.class != LongAnnotation
+                        then 1 else 0 end ), an.class
+               from Annotation an where an.id in
+                    (select distinct(ann.id) from %sAnnotationLink ial
+                        join ial.child as ann
+                        join ial.parent as i
+                where i.id in (:ids))
+                    group by an.class
+            """ % obj_type
+
+        queryResult = self.getQueryService().projection(q, params, ctx)
+
+        for r in queryResult:
+            ur = unwrap(r)
+            if ur[3] == '/basic/num/long/':
+                counts['LongAnnotation'] += int(ur[0])
+                counts['OtherAnnotation'] += int(ur[1])
+            elif ur[3] == '/basic/text/comment/':
+                counts['CommentAnnotation'] += int(ur[2])
+            elif ur[3] == '/basic/text/tag/':
+                counts['TagAnnotation'] += int(ur[2])
+            elif ur[3] == '/type/OriginalFile/':
+                counts['FileAnnotation'] += int(ur[2])
+            elif ur[3] == '/map/':
+                counts['MapAnnotation'] += int(ur[2])
+            else:
+                counts['OtherAnnotation'] += int(ur[2])
+
+        return counts
+
     def listOrphanedAnnotations(self, parent_type, parent_ids, eid=None,
                                 ns=None, anntype=None, addedByMe=True):
         """
@@ -3504,15 +3565,12 @@ class _BlitzGateway (object):
                 obj_ids.append(o.id)
 
         if obj_type is None:
-            return countAnnotations()
+            return self.countAnnotations()
 
         obj_type = obj_type.title().replace("Plateacquisition",
                                             "PlateAcquisition")
 
-        ctx = self.SERVICE_OPTS.copy()
-        ctx.setOmeroGroup(self.group)
-
-        return countAnnotations(self.getQueryService(), ctx, obj_type, obj_ids)
+        return self.countAnnotations(obj_type, obj_ids)
 
     def createImageFromNumpySeq(self, zctPlanes, imageName, sizeZ=1, sizeC=1,
                                 sizeT=1, description=None, dataset=None,
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -141,7 +141,7 @@ def fileread_gen(fin, fsize, bufsize):
     fin.close()
 
 
-def countAnnotations(annotationlinks=None):
+def countAnnotations(annotationlinks=[]):
     """
     Count the different (unique) annotions from the
     provided annotation links
@@ -154,9 +154,6 @@ def countAnnotations(annotationlinks=None):
         "LongAnnotation": 0,
         "MapAnnotation": 0,
         "OtherAnnotation": 0}
-
-    if annotationlinks is None:
-        return counts
 
     atypes = {
         omero.model.TagAnnotationI: "TagAnnotation",
@@ -1033,9 +1030,8 @@ class BlitzObjectWrapper (object):
 
         q = """
             select al from %sAnnotationLink al
-            left outer join al.parent as pa
             left outer join fetch al.child as an
-            where pa.id = :id
+            where al.parent.id = :id
             """ % self.OMERO_CLASS
 
         if ns:
@@ -3484,21 +3480,20 @@ class _BlitzGateway (object):
         for e in q.findAllByQuery(sql, p, self.SERVICE_OPTS):
             yield AnnotationWrapper._wrap(self, e)
 
-    def getAnnotationCounts(self, objDict=None):
+    def getAnnotationCounts(self, objDict={}):
         """
         Get the annotion counts for the given objects
         """
         obj_type = None
         obj_ids = []
-        if objDict is not None:
-            for key in objDict:
-                for o in objDict[key]:
-                    if obj_type is not None and obj_type != key:
-                        raise AttributeError(
-                            "getAnnotationCounts cannot be used with "
-                            "different types of objects")
-                    obj_type = key
-                    obj_ids.append(o.id)
+        for key in objDict:
+            for o in objDict[key]:
+                if obj_type is not None and obj_type != key:
+                    raise AttributeError(
+                        "getAnnotationCounts cannot be used with "
+                        "different types of objects")
+                obj_type = key
+                obj_ids.append(o.id)
 
         if obj_type is None:
             return countAnnotations()
@@ -3510,9 +3505,8 @@ class _BlitzGateway (object):
         params.addIds(obj_ids)
         q = """
             select al from %sAnnotationLink al
-            left outer join al.parent as pa
             left outer join fetch al.child as an
-            where pa.id in (:ids)
+            where al.parent.id in (:ids)
             """ % (obj_type)
 
         ctx = self.SERVICE_OPTS.copy()
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -165,15 +165,13 @@ def countAnnotations(annotationlinks=None):
         omero.model.LongAnnotationI: "LongAnnotation",
         omero.model.MapAnnotationI: "MapAnnotation"}
 
-    uniqueIds = []
+    # Avoid counting the same annotations multiple times
+    uniqueIds = set([al._child._id._val for al in annotationlinks])
+
     total = 0
     regAnnotations = 0
     for al in annotationlinks:
-        # Avoid counting the same annotations multiple times
-        annoId = str(type(al._child))+"_"+str(al._child._id._val)
-        if annoId not in uniqueIds:
-            uniqueIds.append(annoId)
-        else:
+        if al._child._id._val not in uniqueIds:
             continue
 
         total += 1
@@ -1041,7 +1039,7 @@ class BlitzObjectWrapper (object):
             """ % self.OMERO_CLASS
 
         if ns:
-            q = q+" and an.ns in (:ns)"
+            q = q + " and an.ns in (:ns)"
             params.map["ns"] = rstring(ns)
 
         ctx = self._conn.SERVICE_OPTS.copy()
@@ -3505,20 +3503,8 @@ class _BlitzGateway (object):
         if obj_type is None:
             return countAnnotations()
 
-        if obj_type == "project":
-            obj_type = "Project"
-        if obj_type == "dataset":
-            obj_type = "Dataset"
-        if obj_type == "image":
-            obj_type = "Image"
-        if obj_type == "screen":
-            obj_type = "Screen"
-        if obj_type == "plate":
-            obj_type = "Plate"
-        if obj_type == "plateacquisition":
-            obj_type = "PlateAcquisition"
-        if obj_type == "well":
-            obj_type = "Well"
+        obj_type = obj_type.title().replace("Plateacquisition",
+        "PlateAcquisition")
 
         params = omero.sys.ParametersI()
         params.addIds(obj_ids)
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1026,8 +1026,9 @@ class BlitzObjectWrapper (object):
             select sum( case when an.ns = :ratingns
                         and an.class = LongAnnotation
                         then 1 else 0 end),
-                sum( case when an.ns != :ratingns and an.class = LongAnnotation
-                        then 1 else 0 end),
+                sum( case when (an.ns is null or an.ns != :ratingns)
+                               and an.class = LongAnnotation
+                               then 1 else 0 end),
                sum( case when an.class != LongAnnotation
                         then 1 else 0 end ), an.class
                from Annotation an where an.id in
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -141,10 +141,15 @@ def fileread_gen(fin, fsize, bufsize):
     fin.close()
 
 
-def countAnnotations(queryResult=[]):
+def countAnnotations(queryService, ctx, obj_type, obj_ids=[]):
     """
-    Count the different the annotions from the
-    provided projection query result
+    Count the annotions linked to the given object
+
+    :param queryService:  Reference to the query service
+    :param ctx:           The security context
+    :param obj_type:      The type of the object the annotations are linked to
+    :param obj_ids:       List of object ids
+    :return:              Dictionary of annotation counts per annotation type
     """
 
     counts = {
@@ -154,6 +159,33 @@ def countAnnotations(queryResult=[]):
         "LongAnnotation": 0,
         "MapAnnotation": 0,
         "OtherAnnotation": 0}
+
+    if obj_type is None or not obj_ids:
+        return counts
+
+    params = omero.sys.ParametersI()
+    params.addIds(obj_ids)
+    params.add('ratingns',
+               rstring(omero.constants.metadata.NSINSIGHTRATING))
+
+    q = """
+        select sum( case when an.ns = :ratingns
+                    and an.class = LongAnnotation
+                    then 1 else 0 end),
+            sum( case when (an.ns is null or an.ns != :ratingns)
+                           and an.class = LongAnnotation
+                           then 1 else 0 end),
+           sum( case when an.class != LongAnnotation
+                    then 1 else 0 end ), an.class
+           from Annotation an where an.id in
+                (select distinct(ann.id) from %sAnnotationLink ial
+                    join ial.child as ann
+                    join ial.parent as i
+            where i.id in (:ids))
+                group by an.class
+        """ % obj_type
+
+    queryResult = queryService.projection(q, params, ctx)
 
     for r in queryResult:
         ur = unwrap(r)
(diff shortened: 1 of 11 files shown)


components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -201,6 +201,126 @@ def getPlateId(self):
...
+            """ % (objtype)
+            
+        total = 0
+        regAnnotations = 0
+        for al in qs.findAllByQuery(q, params):
+            total += 1
+            if type(al._child) in atypes:
+                annoType = atypes[type(al._child)]
+                if annoType == "LongAnnotation" and\
+                    al._child._ns._val != "openmicroscopy.org/omero/insight/rating":
```
atarkowska: could you please use constant `omero.constants.metadata.NSINSIGHTRATING`

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -201,6 +201,126 @@ def getPlateId(self):
...
+            """ % (obj_type)
+
+        total = 0
+        regAnnotations = 0
+        for al in qs.findAllByQuery(q, params):
+            total += 1
+            if type(al._child) in atypes:
+                annoType = atypes[type(al._child)]
+                if annoType == "LongAnnotation" and\
+                    al._child._ns._val != "openmicroscopy.org/omero/insight/rating":
```
atarkowska: could you please use constant `omero.constants.metadata.NSINSIGHTRATING`

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -201,6 +201,126 @@ def getPlateId(self):
...
+        counts = {"TagAnnotation" : 0,
+            "FileAnnotation" : 0,
+            "CommentAnnotation" : 0,
+            "LongAnnotation" : 0,
+            "MapAnnotation" : 0,
+            "OtherAnnotation" : 0}
+
+        obj_type = None
+        obj_ids = []
+        for object in objects:
```
atarkowska: please do not use `object` as a var, just `obj`

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -201,6 +201,126 @@ def getPlateId(self):
...
+        q = """
+            select al from %sAnnotationLink al
+            left outer join fetch al.parent as pa
+            left outer join fetch al.child as an
+            where pa.id in (:ids)
+            """ % (obj_type)
+
+        total = 0
+        regAnnotations = 0
+        for al in qs.findAllByQuery(q, params):
```
will-moore: You'll need a 3rd argument here so that the group context is right
```findAllByQuery(q, params, self.conn.SERVICE_OPTS)```
otherwise I think it will only query your default group.

File: components/tools/OmeroWeb/omeroweb/webclient/controller/container.py [hunk 5]
```diff
@@ -201,6 +201,126 @@ def getPlateId(self):
...
+            omero.model.LongAnnotationI : "LongAnnotation",
+            omero.model.MapAnnotationI : "MapAnnotation"}
+
+        qs = self.conn.getQueryService()
+
+        params = omero.sys.ParametersI()
+        params.addIds(obj_ids)
+        q = """
+            select al from %sAnnotationLink al
+            left outer join fetch al.parent as pa
```
will-moore: Minor point: I think you shouldn't need the ```fetch``` for parent as you don't need to retrieve parent for anything (just need ```join``` for the query itself).

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/metadata_general_comment.html
```diff
@@ -22,6 +22,11 @@
     <div id="commentsPane">
     <h1 class="can-collapse closed" data-name="comments">
         {{ label|default:"Comments" }}
+        {% for key, value in annotationCounts.items %}
```
atarkowska: no need to loop.  just`({{ annotationCounts.CommentAnnotation }})`

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/metadata_general_file.html
```diff
@@ -22,6 +22,11 @@
     <div id="fileannsPane">
     <h1 class="can-collapse closed" data-name='attachments'>
         {{ label|default:"Attachments" }}
+        {% for key, value in annotationCounts.items %}
```
atarkowska: no need to loop. `({{ annotationCounts.FileAnnotation }})`

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/metadata_general_map.html
```diff
@@ -22,6 +22,11 @@
     <div id="mapAnnsPane">
     <h1 class="can-collapse closed" data-name="keyvaluepairs">
         {{ label|default:"Key-Value Pairs" }}
+        {% for key, value in annotationCounts.items %}
```
atarkowska: no need to loop. `({{ annotationCounts.MapAnnotation }})`

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/metadata_general_other.html
```diff
@@ -22,6 +22,11 @@
     <div id="customAnnotationsPane">
     <h1 class="can-collapse closed" data-name="others">
         {{ label|default:"Others" }}
+        {% for key, value in annotationCounts.items %}
```
atarkowska: no need to loop. `({{ annotationCounts.OtherAnnotation }})`

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/metadata_general_rating.html
```diff
@@ -22,6 +22,11 @@
     <div id="ratingsPane">
     <h1 class="can-collapse closed" data-name="ratings">
         {{ label|default:"Ratings" }}
+        {% for key, value in annotationCounts.items %}
```
atarkowska: no need to loop. `({{ annotationCounts.LongAnnotation }})`

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/metadata_general_tag.html
```diff
@@ -22,6 +22,11 @@
     <div id="tagsPane">
     <h1 class="can-collapse closed" data-name="tags">
         {{ label|default:"Tags" }}
+        {% for key, value in annotationCounts.items %}
```
atarkowska: no need to loop. `({{ annotationCounts.TagAnnotation }})`

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -201,6 +201,126 @@ def getPlateId(self):
...
+        if obj_type == "screen":
+            obj_type = "Screen"
+        if obj_type == "plate":
+            obj_type = "Plate"
+        if obj_type == "plateacquisition":
+            obj_type = "PlateAcquisition"
+        if obj_type == "well":
+            obj_type = "Well"
+
+        atypes = {omero.model.TagAnnotationI : "TagAnnotation",
```
atarkowska: we are repeating the same code, could you move them to helpers please?

will-moore:
Here you're counting annotation 'links' rather than the annotations themselves. E.g. if 2 users add the same Tag, this will show up as ```(2)``` which is different from what Insight is doing.

However, it is consistent with the way webclient displays Tags in the single-object right panel (each tag is shown once for every time it is linked) but not in batch-annotate panel (all links for each tag are collapsed into single tag).
If we want the number to match the number of Tags *shown* in each panel, then we'd need to count them differently for batch vv single object.
Probably better to always count Annotations rather than links, and we can consider changing the display of Tags for single object later. I know @jburel has been advocating this for some time!

will-moore:
A few flake8 errors to fix for travis.

components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/metadata_general_tag.html
```diff
@@ -22,6 +22,11 @@
...
+            {% if key == 'TagAnnotation' %}
+                ({{ value }})
+            {% endif %}
+        {% endfor %}
```
will-moore: You don't need this loop, you can simply do ``` {{ annotationCounts.TagAnnotation }} ```.

atarkowska (changes requested):
Thank for tackling the card. Some comments  in the code.

`BlitzObjectWrapper` offers `listAnnotations` https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/gateway/__init__.py#L949. Why not add `countAnnotations` there? To me looks like it should to to gateway

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -140,6 +140,73 @@ def fileread_gen(fin, fsize, bufsize):
         p += s
     fin.close()
 
+def countAnnotations(annotationlinks):
```
dominikl: I should move these methods somewhere else. Where's a good place for utilility/helper methods @aleksandra-tarkowska ?

will-moore:
Web error reported at 
```/webclient/metadata_details/acquisition/160/```
```
 serverStackTrace = ome.services.query.QueryException: Illegal query
None is not mapped [
           select al from None al
           left outer join al.parent as pa
           left outer join fetch al.child as an
           where pa.id = :id
           ]
```

**Comment by will-moore** [comment]
That error was also causing 'Run' right panel to fail in a bunch of robot tests https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/497/robot/

**Comment by atarkowska** [comment]
could you please add gateway tests?

File: components/tools/OmeroPy/src/omero/gateway/__init__.py [hunk 15]
```diff
@@ -952,6 +1022,28 @@ def getAnnotation(self, ns=None):
             return AnnotationWrapper._wrap(self._conn, rv[0].child, link=rv[0])
         return None
 
+    def getAnnotationCount(self):
```
atarkowska: This could be named `countAnnotations`, but up to you. I would also expect `ns` as a param to match `listAnnotations`

components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
```diff
@@ -201,6 +201,19 @@ def getPlateId(self):
         elif self.acquisition:
             return self.acquisition._obj.plate.id.val
 
+    def getAnnotationCounts(self):
+        """
+        Get the annotion counts for the current object
+        """
+        obj = self._get_object()
```
atarkowska: why not `return self._get_object().getAnnotationCount()`

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3354,6 +3446,36 @@ def listOrphanedAnnotations(self, parent_type, parent_ids, eid=None,
...
 
+    def countAnnotations(self, objects):
+        """
+        Get the annotion counts for the given objects
+        """
+
+        obj_type = None
+        obj_ids = []
+        for obj in objects:
+            objtype = obj.split('=')[0]
```
atarkowska: sorry for being picky, why that has to be a list of strings `type-id`? In the `views.batch_annotate` you have a dict `objs` of real objects https://github.com/dominikl/openmicroscopy/blob/c99b0e5c0e7535326d947cc4f0e04666cf938234/components/tools/OmeroWeb/omeroweb/webclient/views.py#L1969.

example https://github.com/dominikl/openmicroscopy/blob/c99b0e5c0e7535326d947cc4f0e04666cf938234/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py#L238

atarkowska (commented):
thanks for the changes. still few suggestions

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: So you could use `objtype=o.OMERO_CLASS` directly without a need of `getAnnotationLinkTableName`

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -952,6 +1022,28 @@ def getAnnotation(self, ns=None):
...
+
+        params = omero.sys.ParametersI()
+        params.addId(long(self._oid))
+
+        q = """
+            select al from %s al
+            left outer join al.parent as pa
+            left outer join fetch al.child as an
+            where pa.id = :id
+            """ % (getAnnotationLinkTableName(self.OMERO_CLASS))
```
atarkowska: why do you need `getAnnotationLinkTableName`?
How about
```
q = """select al from %sAnnotationLinks al
            left outer join al.parent as pa
            left outer join fetch al.child as an
            where pa.id = :id
    """ % self.OMERO_CLASS
```

components/tools/OmeroPy/src/omero/gateway/__init__.py
dominikl: Couldn't use `countAnnotations` because it's also defined in https://github.com/dominikl/openmicroscopy/blob/c99b0e5c0e7535326d947cc4f0e04666cf938234/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py#L2040 . I'll rename both methods to `getAnnotationCounts` to be consistent.

atarkowska: How about moving https://github.com/dominikl/openmicroscopy/blob/c99b0e5c0e7535326d947cc4f0e04666cf938234/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py#L2040 down to gateway then. no need for 2

atarkowska: So I think we do not want to move `self.annotation_counter` but the method could be unified in the main gateway so we don't need to deprecate anything. Happy to chat

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3446,28 +3426,40 @@ def listOrphanedAnnotations(self, parent_type, parent_ids, eid=None,
         for e in q.findAllByQuery(sql, p, self.SERVICE_OPTS):
             yield AnnotationWrapper._wrap(self, e)
 
-    def countAnnotations(self, objects):
+    def getAnnotationCounts(self, objDict=None):
```
atarkowska: empty dict in python is declared as `mydict = {}` see  https://docs.python.org/2/library/stdtypes.html

This will throw exception in a loop

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,55 @@ def fileread_gen(fin, fsize, bufsize):
...
+        omero.model.TagAnnotationI: "TagAnnotation",
+        omero.model.FileAnnotationI: "FileAnnotation",
+        omero.model.CommentAnnotationI: "CommentAnnotation",
+        omero.model.LongAnnotationI: "LongAnnotation",
+        omero.model.MapAnnotationI: "MapAnnotation"}
+
+    uniqueIds = []
+    total = 0
+    regAnnotations = 0
+    for al in annotationlinks:
```
atarkowska: as Annotation are in the same table I would assume they shouldn't have the same id regardless of type, so unique list could be constructed from the annotationlinks.

Then you could just do: `uniqueIds = set([al._child._id._val for al in annotationlinks])`

if you need per type maybe just use dict `{'type': [ id, ...]}` rather then parse string

just a suggestion

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -952,6 +1001,32 @@ def getAnnotation(self, ns=None):
...
+
+        q = """
+            select al from %sAnnotationLink al
+            left outer join al.parent as pa
+            left outer join fetch al.child as an
+            where pa.id = :id
+            """ % self.OMERO_CLASS
+
+        if ns:
+            q = q+" and an.ns in (:ns)"
```
atarkowska: interesting, is that passing flake8? I think `q = q + " and an.ns in (:ns)"`

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3354,6 +3429,56 @@ def listOrphanedAnnotations(self, parent_type, parent_ids, eid=None,
...
+        if obj_type == "image":
+            obj_type = "Image"
+        if obj_type == "screen":
+            obj_type = "Screen"
+        if obj_type == "plate":
+            obj_type = "Plate"
+        if obj_type == "plateacquisition":
+            obj_type = "PlateAcquisition"
+        if obj_type == "well":
+            obj_type = "Well"
```
atarkowska: how about to replace that long conditional statement with `obj_type.title().replace("Plateacquisition", "PlateAcquisition")`

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,53 @@ def fileread_gen(fin, fsize, bufsize):
     fin.close()
 
 
+def countAnnotations(annotationlinks=None):
```
atarkowska: is that really an optional arg?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,53 @@ def fileread_gen(fin, fsize, bufsize):
...
+
+    counts = {
+        "TagAnnotation": 0,
+        "FileAnnotation": 0,
+        "CommentAnnotation": 0,
+        "LongAnnotation": 0,
+        "MapAnnotation": 0,
+        "OtherAnnotation": 0}
+
+    if annotationlinks is None:
```
atarkowska: if you declare as a list you don't need that bit

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,53 @@ def fileread_gen(fin, fsize, bufsize):
...
+        return counts
+
+    atypes = {
```
aleksandra-tarkowska: this is not needed

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,53 @@ def fileread_gen(fin, fsize, bufsize):
...
+    uniqueIds = set([al._child._id._val for al in annotationlinks])
+
+    total = 0
+    regAnnotations = 0
+    for al in annotationlinks:
+        if al._child._id._val not in uniqueIds:
+            continue
+
+        total += 1
+        if type(al._child) in atypes:
```
atarkowska: isinstance please

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: this is inconvenient, accordion is now configurable https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/settings.py#L663, would you like to take it to the account? How about other way around, building dict of types from annotationlinks

atarkowska:
I am guessing https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/settings.py#L663 could be taken to the account in HQL query as well

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3354,6 +3427,44 @@ def listOrphanedAnnotations(self, parent_type, parent_ids, eid=None,
         for e in q.findAllByQuery(sql, p, self.SERVICE_OPTS):
             yield AnnotationWrapper._wrap(self, e)
 
+    def getAnnotationCounts(self, objDict=None):
```
atarkowska: is this really optional arg?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3354,6 +3427,44 @@ def listOrphanedAnnotations(self, parent_type, parent_ids, eid=None,
...
+        """
+        Get the annotion counts for the given objects
+        """
+        obj_type = None
+        obj_ids = []
+        if objDict is not None:
```
atarkowska: if `def getAnnotationCounts(self, objDict={})` then condition is not needed

will-moore:
@aleksandra-tarkowska I don't think that "omero.web.ui.metadata_panes" can be used to improve the HQL query much because we're not selecting by annotation type in the query. Just loading all annotations in a single query and then count each type subsequently.

atarkowska:
@will-moore I don't understand your comment. 

see https://docs.jboss.org/hibernate/orm/3.5/reference/en/html/queryhql.html#queryhql-where

```
In [15]: q2 = select ial from ImageAnnotationLink ial join fetch ial.child  where ial.child.class in (LongAnnotation) and ial.parent = :id 

In [16]: [ (type(ial.child), ial.child.id.val)  for ial in qs.findAllByQuery(q2,p)]
Out[17]: [(omero.model.LongAnnotationI, 13109L)]
```

as opposite to:

```
In [20]: q = """ select ial from ImageAnnotationLink ial join fetch ial.child  where  ial.parent = id""" 

In [21]: [ (type(ial.child), ial.child.id.val)  for ial in qs.findAllByQuery(q,p)]
Out[22]:
[(omero.model.CommentAnnotationI, 17057L),
 (omero.model.TagAnnotationI, 13602L),
 (omero.model.MapAnnotationI, 15901L),
 (omero.model.FileAnnotationI, 13657L),
 (omero.model.FileAnnotationI, 13656L),
 (omero.model.LongAnnotationI, 13109L),
 (omero.model.FileAnnotationI, 8007L),
 (omero.model.FileAnnotationI, 8008L)]
```

will-moore:
@aleksandra-tarkowska The query in this PR loads all the types of annotations in a single query
```
select al from %sAnnotationLink al left outer join al.parent as pa left outer join fetch al.child as an
where pa.id = :id
```
If we know that e.g. we don't need Comments or Tags because the comments & tags panes are not displayed, we can't easily exclude loading them. And the performance gains would be pretty small.

atarkowska:
@will-moore how do you mean not easy? what is wrong with `where ial.child.class in (LongAnnotations, ...) ` you can build where clause before executing query.

I disagree with the performance, as we are fetching child

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3354,6 +3427,44 @@ def listOrphanedAnnotations(self, parent_type, parent_ids, eid=None,
...
+            return countAnnotations()
+
+        obj_type = obj_type.title().replace("Plateacquisition",
+        "PlateAcquisition")
+
+        params = omero.sys.ParametersI()
+        params.addIds(obj_ids)
+        q = """
+            select al from %sAnnotationLink al
+            left outer join al.parent as pa
```
atarkowska: I think if you don't use parent you don't need this line `id` should be guaranteed , it is enough `where al.parent in (:ids)`

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -952,6 +999,32 @@ def getAnnotation(self, ns=None):
...
+        """
+        Get the annotion counts for the current object
+        """
+
+        params = omero.sys.ParametersI()
+        params.addId(long(self._oid))
+
+        q = """
+            select al from %sAnnotationLink al
+            left outer join al.parent as pa
```
atarkowska: I think if you don't use parent you don't need this line, `id` should be guaranteed , it is enough `where al.parent in (:ids)`

will-moore:
@aleksandra-tarkowska It just seems a lot of work to avoid loading some annotations. I can't imagine many of our users have configured the "metadata panes" to hide some panes, and if they have then those panes probably won't have many annotations on them (otherwise they wouldn't hide them), so the performance gains will be less.
What would be nice is to be able to do all the counts of the different annotation types in the SQL query itself, so we don't need to load any links or annotations.
Or maybe just ``` select al.child.class, al.child.id, al.child.ns from... ``` since that is all you need. (Haven't tried to see if that's possible yet).

atarkowska:
Updated above queries with proper classes.

atarkowska:
Well, if the purpose is to count why do you propose loading data rather then just counting them? Little issue with projection is that you don't get class but discriminator value, that confused me on the beginning.

```
In [137]: q2 = """ select count(an.id), an.class from ImageAnnotationLink ial join ial.child as an join ial.parent as i where i.id = :id group by an.class """                                                                                           
In [138]: for r in qs.projection(q2,p):
  a = unwrap (r)
  print a
   .....: 
[4L, '/type/OriginalFile/']
[1L, '/basic/num/long/']
[1L, '/basic/text/comment/']
[1L, '/basic/text/tag/']
[1L, '/map/']
```

cc @joshmoore

will-moore:
@aleksandra-tarkowska That's certainly getting there, but unfortunately we'd like to count unique '''annotations''' not '''links''' (E.g. if many users add the same tag, we only want to count it once). Also need to use namespaces to distinguish ratings from other Long annotations. So it starts to get a bit tricky. Certainly beyond my SQL knowledge.

will-moore:
Oops - sorry, you are counting ann.ids! ;)

atarkowska:
my example count annotations. I don't understand your comment, could you please get familiar with HQL doc? Have you heard about distinct?

atarkowska:
@joshmoore regarding https://github.com/openmicroscopy/openmicroscopy/pull/5023#issuecomment-272985353. If `In [15]: q2 = select ial from ImageAnnotationLink ial join fetch ial.child  where ial.child.class in (LongAnnotation) and ial.parent = :id ` accept `LongAnnotation` as an arg, is there any way in `select count(an.id), an.class` let `projection` returns class rather then string (string is not very representative)?

joshmoore:
I don't immediately see an alternative to getting the discriminator back. That being said, there is a 1-to-1 mapping between discriminator and class (see https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/model/resources/mappings/annotations.ome.xml#L20) which can be used as a workaround until/unless we find something else.

components/tools/OmeroPy/src/omero/gateway/__init__.py
dominikl: How shall I use `isinstance` to check if the type of the annotation is part of the dictionary `atypes`?

dominikl:
Ok, so maybe a `select count(an.id)...` would work, but then is still the issue, that you have to count LongAnnotations with 'Rating' namespace differently from the others.

dominikl:
I'll leave the HQL query for now like it is. But still I'm stuck on how to reload the annotation counts when you add a new annotation. Any hints?

will-moore:
@dominikl That's going to be a bit more work (updating via JavaScript) since you'll need to load the data as json, then iterate through the panes (or just pick the one that you know is updated) and update the html. You'd need a new views.py method to call your annotation counting code and return the dict as json. I'd be tempted to leave that as a follow-up since this PR is already getting quite long.

dominikl:
Oh wow, ok, I'll leave that for a follow-up PR then!

will-moore:
@dominikl I guess you could maybe go for a simpler JavaScript-only approach: Simply increment the count, except for Tags (and maybe files) where you'd want to check whether the new annotation was already added by someone else. I guess you'd need to do a similar thing when an annotation is removed or deleted. I think it's still a fair bit of work, so best not to try it for this PR.

will-moore:
@dominikl I think you'll need an integration test for this. But this will be similar to the one you've just written, just with a few annotations added.

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: `isinstance(al._child, tuple(atypes.keys()))` https://docs.python.org/2/library/functions.html#isinstance

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -952,6 +996,31 @@ def getAnnotation(self, ns=None):
...
+        params = omero.sys.ParametersI()
+        params.addId(long(self._oid))
+
+        q = """
+            select al from %sAnnotationLink al
+            left outer join fetch al.child as an
+            where al.parent.id = :id
+            """ % self.OMERO_CLASS
+
+        if ns:
```
will-moore: Better to explicitly check here ```if ns is not None:``` since you may actually want to use an empty string etc which would evaluate to False.

atarkowska: if we need to query for empty ns then query has to be `and an.ns is not null`

will-moore: We don't need to check if ns is not null since an empty string is not null.
When I add a Tag with ```ns = rstring("")```
Using ```if ns is not None:``` I get
```
>>> project.getAnnotationCounts(ns="")
{'CommentAnnotation': 0,
 'FileAnnotation': 0,
 'LongAnnotation': 0,
 'MapAnnotation': 0,
 'OtherAnnotation': 0,
 'TagAnnotation': 1}
```
But without that fix I get
```
>>> project.getAnnotationCounts(ns="")
{'CommentAnnotation': 7,
 'FileAnnotation': 6,
 'LongAnnotation': 4,
 'MapAnnotation': 4,
 'OtherAnnotation': 4,
 'TagAnnotation': 8}
```

atarkowska:
@dominikl sorry for a slow response. the simplest what comes to my mind is:

```
select sum( case when an.ns = :ratingns and an.class = LongAnnotation
             then 1 else 0 end),
       sum( case when an.ns != :ratingns and an.class != LongAnnotation
             then 1 else 0 end ), an.class
       from ImageAnnotationLink ial 
             join ial.child as an 
             join ial.parent as i 
       where i.id = :id 
            group by an.class
```

but is up to you

will-moore:
With Dom's help we refined the query to handle LongAnnotations where ns is None.
```
        params.add('ratingns', rstring(omero.constants.metadata.NSINSIGHTRATING))
        
        q = """
            select sum( case when an.ns = :ratingns and an.class = LongAnnotation
                        then 1 else 0 end),
                sum( case when an.class = LongAnnotation
                        then 1 else 0 end),
               sum( case when an.class != LongAnnotation
                        then 1 else 0 end ), an.class
               from %sAnnotationLink ial 
                        join ial.child as an 
                        join ial.parent as i 
                where i.id = :id 
                    group by an.class
            """ % self.OMERO_CLASS
```
This was returning the same counts as for the existing logic.

However, it seems we are still counting the same annotations multiple times.
The code for handling this makes a list of IDs that *includes* the IDs that are duplicated.
```
uniqueIds = set([al._child._id._val for al in annotationlinks])
```
Then, you exclude annotations who's ID is not in this set, but this means that NO annotations get excluded and duplicates are counted
```
    for al in annotationlinks:
        if al._child._id._val not in uniqueIds:
            continue
```

So this means that we need a different query to avoid duplicates, or we just fix that logic and stick with the current HQL?
@aleksandra-tarkowska Any idea if that nice query can get even more complex to avoid duplicate annotations?

atarkowska:
so that was only an example, you could try to replace `sum` with `count` and see if `count` with `distinct` works actually

will-moore:
@dominikl To fix the bug with LongAnnotation and ```ns``` of None, you can use ```unwrap()``` which handles None:
```
if annoType == "LongAnnotation" and unwrap(al._child._ns) !=\
                    omero.constants.metadata.NSINSIGHTRATING:
```

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,50 @@ def fileread_gen(fin, fsize, bufsize):
...
+    regAnnotations = 0
+    for al in annotationlinks:
+        if al._child._id._val not in uniqueIds:
+            continue
+
+        total += 1
+        if type(al._child) in atypes:
+            annoType = atypes[type(al._child)]
+            if annoType == "LongAnnotation" and al._child._ns._val !=\
+                    omero.constants.metadata.NSINSIGHTRATING:
```
atarkowska: do you need extra var? 

```
if isinstance(al._child, omero.model.LongAnnotation) and al._child._ns._val != omero.constants.metadata.NSINSIGHTRATING
```

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: Ideally if you could load atypes from the registry of abstract class. https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/gateway/__init__.py#L4599
Maybe look at this later as an improvement step

will-moore:
OK, I think this is doing what we want now
```
        q = """
            select sum( case when an.ns = :ratingns and an.class = LongAnnotation
                        then 1 else 0 end),
                sum( case when an.class = LongAnnotation
                        then 1 else 0 end),
               sum( case when an.class != LongAnnotation
                        then 1 else 0 end ), an.class
               from Annotation an where an.id in 
                    (select distinct(ann.id) from %sAnnotationLink ial 
                        join ial.child as ann 
                        join ial.parent as i 
                where i.id = :id) 
                    group by an.class
            """ % self.OMERO_CLASS
```

atarkowska:
@dominikl @will-moore https://github.com/openmicroscopy/openmicroscopy/pull/5051 should allow to load annotation class rather then discriminator. Lets review with Monday build.

Thanks @joshmoore

dominikl:
Just found a problem with this PR: Clicking on a Tag in the Tags Pane, will trigger a QueryException. No testing/review needed today until I fixed it.

waxenegger:
perhaps this helps (or not ;-))

```
QueryException: exception ::omero::QueryException
{
    serverStackTrace = ome.services.query.QueryException: Illegal query:
            select sum( case when an.ns = :ratingns
                        and an.class = LongAnnotation
                        then 1 else 0 end),
                sum( case when (an.ns is null or an.ns != :ratingns)
                               and an.class = LongAnnotation
                               then 1 else 0 end),
               sum( case when an.class != LongAnnotation
                        then 1 else 0 end ), an.class
               from Annotation an where an.id in
                    (select distinct(ann.id) from NoneAnnotationLink ial
                        join ial.child as ann
                        join ial.parent as i
                where i.id = :id)
                    group by an.class
            
NoneAnnotationLink is not mapped [
            select sum( case when an.ns = :ratingns
                        and an.class = LongAnnotation
                        then 1 else 0 end),
                sum( case when (an.ns is null or an.ns != :ratingns)
                               and an.class = LongAnnotation
                               then 1 else 0 end),
               sum( case when an.class != LongAnnotation
                        then 1 else 0 end ), an.class
               from ome.model.annotations.Annotation an where an.id in
                    (select distinct(ann.id) from NoneAnnotationLink ial
                        join ial.child as ann
                        join ial.parent as i
                where i.id = :id)
                    group by an.class
            ]
```

not sure if that is a sensible suspicion but it seems that the replacement in the hql leads to NoneAnnotationLink which could be because the obj_type is none, maybe because of title() returning None: https://github.com/dominikl/openmicroscopy/blob/aaeeca9b50eeadb5d27c65deedc562814f86ce1a/components/tools/OmeroPy/src/omero/gateway/__init__.py#L3442

but as I said nothing but a guess, perhaps the syntax is erroneous somewhere else

dominikl:
In that method I already checked that obj_type is not none, but I forgot to check that in the `getAnnotationCount` method which handles single selection (
https://github.com/dominikl/openmicroscopy/blob/aaeeca9b50eeadb5d27c65deedc562814f86ce1a/components/tools/OmeroPy/src/omero/gateway/__init__.py#L1016 ), which caused the error. Fixed with that last commit.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,39 @@ def fileread_gen(fin, fsize, bufsize):
...
+
+    counts = {
+        "TagAnnotation": 0,
+        "FileAnnotation": 0,
+        "CommentAnnotation": 0,
+        "LongAnnotation": 0,
+        "MapAnnotation": 0,
+        "OtherAnnotation": 0}
+
+    for r in queryResult:
```
atarkowska: did you see my comment https://github.com/openmicroscopy/openmicroscopy/pull/5023#issuecomment-274181336 about https://github.com/openmicroscopy/openmicroscopy/pull/5056 ? this code is not needed

dominikl: It doesn't make much of a difference if I get `/basic/num/long/` or `ome.model.annotations.LongAnnotation`, I still have to loop over the result, and check if the LongAnnotation is only a LongAnnotation or a rating, as well as aggregate XML, Boolean, whatever Annotation to "OtherAnnotation".

will-moore:
I noticed that a Long Annotation I have (with no namespace) is not picked up in the Batch Annotate panel (but it is for a Single object) and I see that this is because the queries you have are different.
It would be better not to duplicate this query but instead move the common code to E.g. ```queryAnnotationCounts(omero_class, ids)```

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,39 @@ def fileread_gen(fin, fsize, bufsize):
...
+        elif ur[3] == '/basic/text/comment/':
+            counts['CommentAnnotation'] += int(ur[2])
+        elif ur[3] == '/basic/text/tag/':
+            counts['TagAnnotation'] += int(ur[2])
+        elif ur[3] == '/type/OriginalFile/':
+            counts['FileAnnotation'] += int(ur[2])
+        elif ur[3] == '/map/':
+            counts['MapAnnotation'] += int(ur[2])
+        else:
+            counts['OtherAnnotation'] += int(ur[2])
```
will-moore: Minor point: I don't think you really need all these ```int()``` calls since we already have Longs, unless that's not always the case?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -952,6 +985,42 @@ def getAnnotation(self, ns=None):
             return AnnotationWrapper._wrap(self._conn, rv[0].child, link=rv[0])
         return None
 
+    def getAnnotationCounts(self, ns=None):
```
will-moore: You're not using the ```ns``` parameter here. I don't think we need to support filtering by namespace yet, so this can simply be removed.

atarkowska:
sorry for being picky, where is unit test for countAnnotations?

will-moore:
Should have suggested this earlier, but I think the UI should either display the counts a bit more similar to the jsTree.
I played a bit with this (see commit 55987c6db75e) so just grab that and maybe tweak it if you want?

screen shot 2017-01-26 at 16 43 02

dominikl:
--exclude for now to avoid conflict with #5047

dominikl:
@aleksandra-tarkowska Test is here: https://github.com/openmicroscopy/openmicroscopy/pull/5023/files#diff-82dec98dc083d2b062884eae9d8ddfdeR392

dominikl:
Thanks @will-moore , cherry-picked your commit.

will-moore:
@dominikl Sorry - didn't realise before (just looking at the diffs), but you probably want ```countAnnotations()``` to be in the ```BlitzGateway``` class itself, E.g. like ```getAnnotationLinks()```.
Then you won't have to pass in the queryService or ctx, you can just get them within ```countAnnotations()``` itself. It also makes it easier to reuse elsewhere since you can just do ```conn.countAnnotations("Dataset", [1,2])``` anywhere you have the ```conn``` in hand.

will-moore:
Last commits look good.
I just noticed ```ctx.setOmeroGroup(self.group)```. I'm not sure where this came from (I've not used ```self.group``` much) but it seems that ```self.group``` is a string E.g. 'system' but at this point in the code it's always None.
I think we just want to disable filtering by group (since we have object IDs) so we just need
```ctx.setOmeroGroup(-1)```
or explicitly set it to None
```ctx.setOmeroGroup(None)```.
Don't think there's any functional difference between these? (cc @joshmoore)

**Comment by joshmoore** [comment]
I'm not particularly an expert on `setOmeroGroup` either but a few things:

 * at least for the released version, `self.group` should be a ID and not the group name. (There are periodically error logs to that the effect)
 * `None` removes the value, so doing so explicitly (rather than relying on `self.group` to be `None`) seems clearer (if I follow)
 * Setting to `-1` everywhere could be an issue.

**Comment by will-moore** [comment]
Last commit looks great!
You need to remove the ```exclude``` and merge origin/develop (fixing merge conflicts) so we can do a final test tomorrow.

**Comment by dominikl** [comment]
Thanks @will-moore , forgot about the exclude/merge conflict!

File: components/tools/OmeroPy/src/omero/gateway/__init__.py [hunk 36]
```diff
@@ -3425,7 +3425,7 @@ def countAnnotations(self, obj_type, obj_ids=[]):
             return counts
 
         ctx = self.SERVICE_OPTS.copy()
-        ctx.setOmeroGroup(self.group)
+        ctx.setOmeroGroup(None)
```
will-moore: I found that when you are querying in a group that is NOT your default group, you actually need to use ```ctx.setOmeroGroup(-1)``` here, otherwise you don't see anything:

screen shot 2017-02-28 at 14 56 40

will-moore:
Working fine across different groups now.
Good to merge.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -141,6 +141,95 @@ def fileread_gen(fin, fsize, bufsize):
...
+                    join ial.parent as i
+            where i.id in (:ids))
+                group by an.class
+        """ % obj_type
+
+    queryResult = queryService.projection(q, params, ctx)
+
+    for r in queryResult:
+        ur = unwrap(r)
+        if ur[3] == '/basic/num/long/':
```
atarkowska: https://github.com/openmicroscopy/openmicroscopy/pull/5056 was merged, is there anything stopping you from using `type(an.class)`?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3308,6 +3404,71 @@ def getAnnotationLinks(self, parent_type, parent_ids=None, ann_ids=None,
...
+                        join ial.child as ann
+                        join ial.parent as i
+                where i.id in (:ids))
+                    group by an.class
+            """ % obj_type
+
+        queryResult = self.getQueryService().projection(q, params, ctx)
+
+        for r in queryResult:
+            ur = unwrap(r)
```
jburel, line 3390: same code as above (line 190). A helper method will be better

will-moore: I think @dominikl just forgot to remove the old method when moving it into the Gateway in https://github.com/openmicroscopy/openmicroscopy/pull/5023/commits/b39c5f627be051106cd812936830477b13644f25

dominikl: Thanks. Yes, forgot to remove the old code (line 190+).

dominikl:
--depends-on #5139

jburel:
Thanks for removing the code.
When I add a tag for example, the tag is added but I was expected to be reflected in the count.
This is not the case until I reload the right hand panel
This could be an RFE done in another PR (not 5.3.0) since this PR is getting long

dominikl:
Yes, @will-moore suggested a follow-up PR for that problem, too https://github.com/openmicroscopy/openmicroscopy/pull/5023#issuecomment-273929228

jburel:
@dominikl so I missed it in the long thread. I will add a card

jburel:
card https://trello.com/c/2YHUymUw/18-rfe-increment-count

PR #5025, opened 2017-01-12
Api images

will-moore:
# What this PR does

Adds support for ```/images/``` and ```/images/:image_id/```.

Added integration tests of urls below.
Also added gateway test for ```getObject()``` with 'load_pixels' and 'load_channels' and one for 'orphaned'.

# Testing this PR

- Check tests green.
- Browse from /api/ -> version -> images_url to list all images at e.g. ```/api/v0.1/m/images/```
- Filter by dataset ```/images/?dataset=:id``` or ```datasets/:dataset_id/images/```
- List orphaned images ```/images/?orphaned=true```
- For all above urls, pixels should be loaded for images. json contains "Pixels"
- Show a single image ```/images/:image-id/```
- This should also load Channels data.

# Related reading

https://github.com/openmicroscopy/design/issues/69

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3007,7 +3007,8 @@ class _BlitzGateway (object):
     ###########################
     # Specific Object Getters #
 
-    def getObject(self, obj_type, oid=None, params=None, attributes=None):
+    def getObject(self, obj_type, oid=None, params=None, attributes=None,
+                  opts=None):
         """
         Retrieve single Object by type E.g. "Image" or None if not found.
         If more than one object found, raises ome.conditions.ApiUsageException
@@ -3026,7 +3027,7 @@ class _BlitzGateway (object):
         """
         oids = (oid is not None) and [oid] or None
         query, params, wrapper = self.buildQuery(
-            obj_type, oids, params, attributes)
+            obj_type, oids, params, attributes, opts)
         result = self.getQueryService().findByQuery(
             query, params, self.SERVICE_OPTS)
         if result is not None:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7116,6 +7116,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         Extend base query to handle filtering of Images by Datasets.
         Returns a tuple of (query, clauses, params).
         Supported opts: 'dataset': <dataset_id> to filter by Dataset
+                        'load_pixels': <bool> to load Pixel objects.
 
         :param opts:        Dictionary of optional parameters.
         :return:            Tuple of string, list, ParametersI
@@ -7126,6 +7127,9 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
             query += ' join obj.datasetLinks dlink'
             clauses.append('dlink.parent.id = :did')
             params.add('did', rlong(opts['dataset']))
+        if opts is not None and 'load_pixels' in opts:
+            # We use 'left outer join', since we still want images if no pixels
+            query += ' left outer join fetch obj.pixels pixels'
         return (query, clauses, params)
 
     @classmethod
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7117,10 +7117,10 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         Extend base query to handle filtering of Images by Datasets.
         Returns a tuple of (query, clauses, params).
         Supported opts: 'dataset': <dataset_id> to filter by Dataset
-                        'load_pixels': <bool> to load Pixel objects.
+                        'load_pixels': <bool> to load Pixel objects
                         'load_channels': <bool> to load Channels and
                                                     Logical Channels
-                        'orphaned': <bool>. Images not in Dataset or WellSample
+                        'orphaned': <bool> Images not in Dataset or WellSample
 
         :param opts:        Dictionary of optional parameters.
         :return:            Tuple of string, list, ParametersI
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7141,7 +7141,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         if load_pixels or load_channels:
             # We use 'left outer join', since we still want images if no pixels
             query += ' left outer join fetch obj.pixels pixels' \
-                     ' join fetch pixels.pixelsType'
+                     ' left outer join fetch pixels.pixelsType'
         if load_channels:
             query += ' join fetch pixels.channels as channels' \
                      ' join fetch channels.logicalChannel as logicalChannel' \
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7119,6 +7119,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                         'load_pixels': <bool> to load Pixel objects.
                         'load_channels': <bool> to load Channels and
                                                     Logical Channels
+                        'orphaned': <bool>. Images not in Dataset or WellSample
 
         :param opts:        Dictionary of optional parameters.
         :return:            Tuple of string, list, ParametersI
@@ -7131,15 +7132,34 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
             params.add('did', rlong(opts['dataset']))
         load_pixels = False
         load_channels = False
+        orphaned = False
         if opts is not None:
             load_pixels = 'load_pixels' in opts and opts['load_pixels']
             load_channels = 'load_channels' in opts and opts['load_channels']
+            orphaned = 'orphaned' in opts and opts['orphaned']
         if load_pixels or load_channels:
             # We use 'left outer join', since we still want images if no pixels
             query += ' left outer join fetch obj.pixels pixels'
         if load_channels:
             query += ' join fetch pixels.channels as channels' \
                      ' join fetch channels.logicalChannel'
+        if orphaned:
+            clauses.append(
+                """
+                not exists (
+                    select dilink from DatasetImageLink as dilink
+                    where dilink.child = obj.id
+                )
+                """
+            )
+            clauses.append(
+                """
+                not exists (
+                    select ws from WellSample ws
+                    where ws.image.id = obj.id
+                )
+                """
+            )
         return (query, clauses, params)
 
     @classmethod
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7117,6 +7117,8 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         Returns a tuple of (query, clauses, params).
         Supported opts: 'dataset': <dataset_id> to filter by Dataset
                         'load_pixels': <bool> to load Pixel objects.
+                        'load_channels': <bool> to load Channels and
+                                                    Logical Channels
 
         :param opts:        Dictionary of optional parameters.
         :return:            Tuple of string, list, ParametersI
@@ -7127,9 +7129,17 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
             query += ' join obj.datasetLinks dlink'
             clauses.append('dlink.parent.id = :did')
             params.add('did', rlong(opts['dataset']))
-        if opts is not None and 'load_pixels' in opts:
+        load_pixels = False
+        load_channels = False
+        if opts is not None:
+            load_pixels = 'load_pixels' in opts and opts['load_pixels']
+            load_channels = 'load_channels' in opts and opts['load_channels']
+        if load_pixels or load_channels:
             # We use 'left outer join', since we still want images if no pixels
             query += ' left outer join fetch obj.pixels pixels'
+        if load_channels:
+            query += ' join fetch pixels.channels as channels' \
+                     ' join fetch channels.logicalChannel'
         return (query, clauses, params)
 
     @classmethod
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7111,6 +7111,24 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
     PLANEDEF = omero.romio.XY
 
     @classmethod
+    def _getQueryString(cls, opts=None):
+        """
+        Extend base query to handle filtering of Images by Datasets.
+        Returns a tuple of (query, clauses, params).
+        Supported opts: 'dataset': <dataset_id> to filter by Dataset
+
+        :param opts:        Dictionary of optional parameters.
+        :return:            Tuple of string, list, ParametersI
+        """
+        query, clauses, params = super(
+            _ImageWrapper, cls)._getQueryString(opts)
+        if opts is not None and 'dataset' in opts:
+            query += ' join obj.datasetLinks dlink'
+            clauses.append('dlink.parent.id = :did')
+            params.add('did', rlong(opts['dataset']))
+        return (query, clauses, params)
+
+    @classmethod
     def fromPixelsId(cls, conn, pid):
         """
         Creates a new Image wrapper with the image specified by pixels ID
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7135,15 +7135,20 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         load_channels = False
         orphaned = False
         if opts is not None:
-            load_pixels = 'load_pixels' in opts and opts['load_pixels']
-            load_channels = 'load_channels' in opts and opts['load_channels']
-            orphaned = 'orphaned' in opts and opts['orphaned']
+            load_pixels = opts.get('load_pixels')
+            load_channels = opts.get('load_channels')
+            orphaned = opts.get('orphaned')
         if load_pixels or load_channels:
             # We use 'left outer join', since we still want images if no pixels
-            query += ' left outer join fetch obj.pixels pixels'
+            query += ' left outer join fetch obj.pixels pixels' \
+                     ' join fetch pixels.pixelsType'
         if load_channels:
             query += ' join fetch pixels.channels as channels' \
-                     ' join fetch channels.logicalChannel'
+                     ' join fetch channels.logicalChannel as logicalChannel' \
+                     ' left outer join fetch logicalChannel.photometricInterpretation' \
+                     ' left outer join fetch logicalChannel.illumination' \
+                     ' left outer join fetch logicalChannel.mode' \
+                     ' left outer join fetch logicalChannel.contrastMethod'
         if orphaned:
             clauses.append(
                 """
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
@@ -40,7 +40,7 @@ class TestDeleteObject (object):
         tagId = tag.getId()
         handle = gateway.deleteObjects("Annotation", [tagId])
         gateway._waitOnCmd(handle)
-        assert None == gateway.getObject("Annotation", tagId)
+        assert gateway.getObject("Annotation", tagId) is None
 
     def testDeleteImage(self, gatewaywrapper, author_testimg_generated):
         image = author_testimg_generated
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
@@ -614,8 +614,10 @@ class TestGetObject (object):
         assert len(pr.listParents()) == 0
 
     def testListOrphans(self, gatewaywrapper):
-        gatewaywrapper.loginAsUser()
-        eid = gatewaywrapper.gateway.getUserId()
+        # We login as Author, since they have other non-orphan images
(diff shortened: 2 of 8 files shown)


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7111,6 +7112,58 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
...
+                        'load_channels': <bool> to load Channels and
+                                                    Logical Channels
+                        'orphaned': <bool>. Images not in Dataset or WellSample
+
+        :param opts:        Dictionary of optional parameters.
+        :return:            Tuple of string, list, ParametersI
+        """
+        query, clauses, params = super(
+            _ImageWrapper, cls)._getQueryString(opts)
+        if opts is not None and 'dataset' in opts:
```
jburel, line 7130: Use ``opts.get``

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7111,6 +7112,58 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
...
+            query += ' join obj.datasetLinks dlink'
+            clauses.append('dlink.parent.id = :did')
+            params.add('did', rlong(opts['dataset']))
+        load_pixels = False
+        load_channels = False
+        orphaned = False
+        if opts is not None:
+            load_pixels = 'load_pixels' in opts and opts['load_pixels']
```
jburel: Use ``opts.get``

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7111,6 +7112,58 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
...
+            load_channels = 'load_channels' in opts and opts['load_channels']
```
jburel: Same

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7111,6 +7112,58 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
...
+            orphaned = 'orphaned' in opts and opts['orphaned']
+        if load_pixels or load_channels:
+            # We use 'left outer join', since we still want images if no pixels
+            query += ' left outer join fetch obj.pixels pixels'
+        if load_channels:
+            query += ' join fetch pixels.channels as channels' \
+                     ' join fetch channels.logicalChannel'
```
jburel: If we want to know information about the logical channel, it is also necessary to fetch
``photometricInterpretation, illumination, mode and contrastMethod`` from the logical channel

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -571,6 +571,23 @@ def testGetImage(self, gatewaywrapper, author_testimg_tiny):
...
 
+    def testGetImageLoadPixels(self, gatewaywrapper, author_testimg_tiny):
+        testImage = author_testimg_tiny
+        conn = gatewaywrapper.gateway
+        # By default, don't load pixels
+        image = conn.getObject("Image", testImage.id)
+        assert not image._obj.pixelsLoaded
+        # Load just the pixels
+        image = conn.getObject("Image", testImage.id,
+                               opts={'load_pixels': True})
```
jburel: When the type is added to the query, the value needs to be checked

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -571,6 +571,23 @@ def testGetImage(self, gatewaywrapper, author_testimg_tiny):
...
+        assert image._obj.pixelsLoaded
+        assert not image._obj._pixelsSeq[0].channelsLoaded
+        # Load pixels and channels
+        image = conn.getObject("Image", testImage.id,
+                               opts={'load_channels': True})
+        assert image._obj.pixelsLoaded
+        assert image._obj._pixelsSeq[0].channelsLoaded
```
jburel: The test should be improved to check that the values are correctly loaded.

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -597,8 +614,10 @@ def testTraversal(self, author_testimg_tiny):
         assert len(pr.listParents()) == 0
 
     def testListOrphans(self, gatewaywrapper):
-        gatewaywrapper.loginAsUser()
-        eid = gatewaywrapper.gateway.getUserId()
+        # We login as Author, since they have other non-orphan images
```
jburel: What is author? do you mean owner?

jburel: typo ``orphaned``

components/tools/OmeroWeb/omeroweb/api/urls.py
```diff
@@ -106,6 +106,30 @@
...
+                 name='api_images')
+"""
+GET all images, using omero-marshal to generate json
+"""
+
+api_dataset_images = url(
+    r'^v(?P<api_version>%s)/m/datasets/'
+    '(?P<dataset_id>[0-9]+)/images/$' % versions,
+    views.ImagesView.as_view(),
+    name='api_dataset_imagess')
```
jburel: typo ``api_dataset_imagess``

components/tools/OmeroWeb/omeroweb/api/views.py
```diff
@@ -257,6 +277,27 @@ def get_opts(self, request, **kwargs):
...
+        opts = super(ImagesView, self).get_opts(request, **kwargs)
+        # at /datasets/:dataset_id/images/ we have 'dataset_id' in kwargs
+        if 'dataset_id' in kwargs:
+            opts['dataset'] = long(kwargs['dataset_id'])
+        else:
+            # filter by query /images/?dataset=:id
+            dataset = getIntOrDefault(request, 'dataset', None)
+            if dataset is not None:
+                opts['dataset'] = dataset
+        # When listing images, always load pixels by defualt
```
jburel: typo

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7111,6 +7112,58 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
...
+        load_pixels = False
+        load_channels = False
+        orphaned = False
+        if opts is not None:
+            load_pixels = 'load_pixels' in opts and opts['load_pixels']
+            load_channels = 'load_channels' in opts and opts['load_channels']
+            orphaned = 'orphaned' in opts and opts['orphaned']
+        if load_pixels or load_channels:
+            # We use 'left outer join', since we still want images if no pixels
+            query += ' left outer join fetch obj.pixels pixels'
```
jburel: the pixels type will have to be loaded, you need an extra join ``fetch pixels.pixelsType``

components/tools/OmeroWeb/test/integration/test_api_images.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Create 5 Images in Dataset
+        for i in range(5):
+            img = self.create_test_image(size_x=125, size_y=125,
+                                         session=user1[0].getSession(),
+                                         name="Image%s" % i)
+            img = ImageI(img.id.val, False)
+            dataset.linkImage(img)
+
+        # Import a single orphaned Image "tinyTest.d3d.dv" and get ImageID
```
jburel: Could you avoid using ``tinyTest``? we want to get rid of it!
We can add methods to the testing framework to generate "richer" images.

components/tools/OmeroWeb/test/integration/test_api_images.py
```diff
@@ -0,0 +1,182 @@
...
+        assert_objects(conn, rsp['data'], [orphaned], dtype='Image',
+                       opts={'load_pixels': True})
+
+        # Filter Images by Dataset
+        images.sort(cmp_name_insensitive)
+        payload = {'dataset': dataset.id.val}
+        rsp = _get_response_json(django_client, images_url, payload)
+        # Manual check that Pixels is loaded but Channels are not
+        assert 'Channels' not in rsp['data'][0]['Pixels']
+        assert_objects(conn, rsp['data'], images, dtype='Image',
```
jburel, line 159: Check the values when the queries are reviewed

components/tools/OmeroWeb/test/integration/test_api_images.py
```diff
@@ -0,0 +1,182 @@
...
+                       opts={'load_pixels': True})
+        payload['page'] = 2
+        rsp = _get_response_json(django_client, images_url, payload)
+        assert_objects(conn, rsp['data'], images[limit:limit * 2],
+                       dtype='Image', opts={'load_pixels': True})
+
+        # Show ONLY the orphaned image (channels are loaded by default)
+        img_url = images_url + '%s/' % orphaned.id.val
+        rsp = _get_response_json(django_client, img_url, {})
+        # Manual check that Channels is loaded
```
jburel: are loaded not is

jburel (commented):
First round of review.

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -571,6 +571,23 @@ def testGetImage(self, gatewaywrapper, author_testimg_tiny):
         assert image.getOwnerOmeName == testImage.getOwnerOmeName
         assert image.getThumbVersion() is not None
 
+    def testGetImageLoadPixels(self, gatewaywrapper, author_testimg_tiny):
```
jburel: what is ``author``?

jburel:
General question: Do we want method w/o loading pixels? and the loading of image/pixels is only by default for the methods at the API level

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -597,8 +614,10 @@ def testTraversal(self, author_testimg_tiny):
         assert len(pr.listParents()) == 0
 
     def testListOrphans(self, gatewaywrapper):
-        gatewaywrapper.loginAsUser()
-        eid = gatewaywrapper.gateway.getUserId()
+        # We login as Author, since they have other non-orphan images
+        gatewaywrapper.loginAsAuthor()
```
jburel: Should we deprecate that method and introduce ``loginAsOwner`` (if author means owner), is that a legacy from carlos's work?

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -607,34 +626,35 @@ def testListOrphans(self, gatewaywrapper):
...
+        findImagesInPage = list(conn.listOrphans(
             "Image", eid=eid, params=params, loadPixels=True))
         assert len(findImagesInPage) == 3, \
             "Did not find orphaned images in page"
         for p in findImagesInPage:
             assert p._obj.pixelsLoaded
 
-        # All orphans, no pixels
-        findImages = list(gatewaywrapper.gateway.listOrphans("Image"))
+        # All orphans, without loading pixels
```
jburel: Could the test be split up or parametrised? one test with ``loadPixels=True``, one ``False``

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -607,34 +626,35 @@ def testListOrphans(self, gatewaywrapper):
...
             assert not p._obj.pixelsLoaded
-            client = p._conn
-            handle = client.deleteObjects(
-                'Image', [p.getId()], deleteAnns=True)
-            try:
-                client._waitOnCmd(handle)
-            finally:
-                handle.close()
+
+        # Test getObjects() with 'orphaned' option
```
jburel: Same; break the test in other tests with orphaned option

components/tools/OmeroWeb/test/integration/test_api_containers.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-# Copyright (C) 2016 University of Dundee & Open Microscopy Environment.
+# Copyright (C) 2016-2017 University of Dundee & Open Microscopy Environment.
```
jburel: Change the copyright next time you change the test. Not needed in that PR

components/tools/OmeroWeb/test/integration/test_api_images.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class TestImages(IWebTest):
+    """Tests querying & editing Images."""
+
+    @pytest.fixture()
+    def user1(self):
+        """Return a new user in a read-annotate group."""
+        group = self.new_group(perms='rwra--')
+        user = self.new_client_and_user(group=group)
```
jburel: assignment not needed i.e. ``return self.new_client_and_user(group=group)``

components/tools/OmeroWeb/test/integration/test_api_images.py
```diff
@@ -0,0 +1,182 @@
...
+        limit = 3
+        dataset_images_url = datasets_url + "%s/images/" % dataset.id.val
+        payload = {'dataset': dataset.id.val, 'limit': limit}
+        rsp = _get_response_json(django_client, dataset_images_url, payload)
+        assert_objects(conn, rsp['data'], images[0:limit], dtype='Image',
+                       opts={'load_pixels': True})
+        payload['page'] = 2
+        rsp = _get_response_json(django_client, images_url, payload)
+        assert_objects(conn, rsp['data'], images[limit:limit * 2],
+                       dtype='Image', opts={'load_pixels': True})
```
jburel, line 172: Parametrize the test so the load options are tested with ``True`` and ``False``

components/tools/OmeroWeb/test/integration/test_api_images.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Show ONLY the orphaned image (channels are loaded by default)
+        img_url = images_url + '%s/' % orphaned.id.val
+        rsp = _get_response_json(django_client, img_url, {})
+        # Manual check that Channels is loaded
+        assert len(rsp['Pixels']['Channels']) == 1
+        assert_objects(conn, [rsp], [orphaned], dtype='Image',
+                       opts={'load_channels': True})
```
jburel, line 180: Same: test with ``True`` and ``False``

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: In this case ```opts['dataset']``` dataset is not a boolean so I don't want to check the truthiness of ```opts.get('dataset')```, but just check whether the value is set. E.g. if ```opts.get('dataset')``` is 0 this would be evaluated as ```False```, even though I'd want to use that value.
Unless we decide that we consider ```0``` to be invalid, but that is a change in functionality (not just a stylistic change)?

jburel: Let's not change it. Ignore that comment

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
will-moore: How much "value checking" do we need? We know the query is valid as it's loading the pixels. I'm only testing the new functionality - that the ```load_pixels``` option is being respected. If we're getting an invalid ```sizeX``` etc then something is very wrong (not just my query)! I could also check that ```pixelsType``` is also loaded if that's enough?

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
will-moore: Same question as with ```load_pixels```. How much to check?

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
jburel: Checking that it is loaded will be good for the type
That's should be enough

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
jburel: Checking that the illumination etc. is loaded

will-moore: I have added a ```assert lc.getPhotometricInterpretation().isLoaded()``` but can't check illumination, mode or contrastMethod since they are not set for this test image.

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
will-moore: Added ```pixels.getPixelsType().isLoaded()```

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
jburel: We probably need new method in testing framework allowing to create "richer" images

components/tools/OmeroWeb/test/integration/test_api_images.py
will-moore: Again, I can check that ```['Pixels']['Type']``` is now loaded (with updated query) but for logical channel, none of photometricInterpretation, illumination, mode or contrastMethod are in this test image.

jburel: in that case, we need a richer test image.

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -571,6 +571,33 @@ def testGetImage(self, gatewaywrapper, author_testimg_tiny):
         assert image.getOwnerOmeName == testImage.getOwnerOmeName
         assert image.getThumbVersion() is not None
 
+    @pytest.mark.parametrize("load_pixels", [True, False])
+    @pytest.mark.parametrize("load_channels", [True, False])
+    def testGetImageLoadPixels(self, load_pixels, load_channels,
+                               gatewaywrapper, author_testimg_tiny):
```
will-moore, line 582: @jburel Re: "what is author?". 
author_testimg_tiny is a fixture created by import of 'tinyTest.d3d.dv'

jburel: so we should probably find a better solution since we are trying to get of rid of tinytest

will-moore:
@jburel  Re: https://github.com/openmicroscopy/openmicroscopy/pull/5025#issuecomment-272402028 
I think we decided that ```/images/``` would ALWAYS load pixels since the cost of loading pixels is small and this keeps it simple.

jburel:
@will-moore: my question was more for the "get" method in ``__init__.py`` since we have the option with or w/o pixels.
Do we want that there too? or we are happy with the option being ``false`` by default. The option offers flexibility

will-moore:
@jburel I think at the Blitz Gateway level we probably want to NOT load Pixels by default since that's the behaviour we have had for a long time and Blitz Gateway will lazily load pixels if/when needed.

jburel:
@will-moore: in that case, better not to change the default behaviour in blitz and only have the default to ``true`` at the API level

components/tools/OmeroWeb/test/integration/test_api_images.py
will-moore: In this case I'm not testing the loading of images with/without loading pixels. We don't provide that option for web api (always load pixels). The ```opts={'load_pixels': True}``` is just required here so that the when we ```getObjects('Image', opts={'load_pixels': True})``` to compare objects to json, we are comparing like-for-like (pixels loaded in both cases).

jburel: I was getting a bit ahead of myself here i.e. having the tests for when/if we offer the option at the api level.

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
will-moore: Yes, but that will be a big change to the whole gatewaytest setup since it works in a very different way to our other tests. In fact, I would prefer that we completely get rid of the gateway test setup since it always tries to login as existing user and re-use existing data which fails if the data is not in expected state. It requires you to always try to clean up in every test so as not to corrupt data and you also don't always have the groups & permissions that you want etc.

jburel: That will be another body of work, outside the scope of this PR and even outside what we want to achieve with this work.
Potentially something to change during 5.3.x

components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
```diff
@@ -22,7 +22,7 @@
 from omero.rtypes import wrap
 from omero.testlib import ITest
 from omero.gateway import BlitzGateway, KNOWN_WRAPPERS
-from omero.model import ScreenI, PlateI, WellI, WellSampleI
+from omero.model import ScreenI, PlateI, WellI, WellSampleI, DatasetI, ImageI
```
chris-allan: Just to save yourself some pain later on, you might want to go ahead and have the model object imports one per line in alphabetical order. It'll save on merge conflicts and make it easier to review going forward. I ended up doing it late for omero-marshal, where we have 20+ model objects being imported in `conftest.py`, and I really wished I had done it sooner.

chris-allan (approved):
Structurally good to merge from my perspective.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7111,6 +7112,63 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
     PLANEDEF = omero.romio.XY
 
     @classmethod
+    def _getQueryString(cls, opts=None):
+        """
+        Extend base query to handle filtering of Images by Datasets.
+        Returns a tuple of (query, clauses, params).
+        Supported opts: 'dataset': <dataset_id> to filter by Dataset
+                        'load_pixels': <bool> to load Pixel objects.
```
jburel: minor: extra dot after objects

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7111,6 +7112,63 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
...
+                        'load_channels': <bool> to load Channels and
+                                                    Logical Channels
+                        'orphaned': <bool>. Images not in Dataset or WellSample
```
jburel: dot after ``<bool>``

jburel (commented):
minor typos I missed during the first review.

jburel:
Thanks @will-moore.
I will merge after tagging 5.3.0-m7

PR #5026, opened 2017-01-12
Deprecate scipy usage

jburel:
# What this PR does

Deprecate method introduced to create "map" to be used with the emdatabank viewer
Support for the viewer has been removed in a previous release.

Removing the method in the next release will allow to remove the ``scipy`` dependency

# Testing this PR

No testing required, check that the message is correct

# Related reading

https://trello.com/c/BKRBO167/220-to-deprecate


will-moore:
Looks good. Thanks.

PR #5027, opened 2017-01-12
Deprecate reader

jburel:
# What this PR does

Deprecate the OmeroReader
The reader is not used internally.

Nothing to test.

# Related reading

https://trello.com/c/b8jrapBD/174-omero-reader-out-of-blitz-jar

PR #5029, opened 2017-01-12
Omero reader review

jburel:
# What this PR does

There is some external usage of this reader. In this PR, we  close resources stream/services etc.
The reader has been deprecated  https://github.com/openmicroscopy/openmicroscopy/pull/5027

# Testing this PR

We do not have any internal usage or tests for this reader.
Code review.

# Related reading

https://trello.com/c/b8jrapBD/174-omero-reader-out-of-blitz-jar what this PR assists, fixes, or otherwise affects

### DIFF
--- a/components/blitz/src/loci/ome/io/OmeroReader.java
+++ b/components/blitz/src/loci/ome/io/OmeroReader.java
@@ -469,7 +469,7 @@ public class OmeroReader extends FormatReader {
     public static void main(String[] args) throws Exception {
         try (BufferedReader con = new BufferedReader(
                 new InputStreamReader(System.in, Constants.ENCODING));
-             OmeroReader omeroReader = new OmeroReader()) {
+            OmeroReader omeroReader = new OmeroReader()) {
             // parse OMERO credentials
             System.out.print("Server? ");
             final String server = con.readLine();
--- a/components/blitz/src/loci/ome/io/OmeroReader.java
+++ b/components/blitz/src/loci/ome/io/OmeroReader.java
@@ -104,7 +104,7 @@ public class OmeroReader extends FormatReader {
     private boolean encrypted = true;
 
     private omero.client client;
-    private RawPixelsStorePrx store;
+    private ServiceFactoryPrx serviceFactory;
     private Image img;
     private Pixels pix;
 
@@ -166,11 +166,24 @@ public class OmeroReader extends FormatReader {
         final int[] zct = FormatTools.getZCTCoords(this, no);
 
         final byte[] plane;
+        RawPixelsStorePrx store = null;
         try {
+            store = serviceFactory.createRawPixelsStore();
+            store.setPixelsId(pix.getId().getValue(), false);
             plane = store.getPlane(zct[0], zct[1], zct[2]);
         }
         catch (ServerError e) {
             throw new FormatException(e);
+        } finally {
+            // cannot use try-with-statements
+            // RawPixelsStorePrx does not implement autocloseable
+            if (store != null) {
+                try {
+                    store.close();
+                } catch (Exception ex) {
+                    throw new FormatException(ex);
+                }
+            }
         }
 
         try (RandomAccessInputStream s = new RandomAccessInputStream(plane)) {
@@ -257,7 +270,7 @@ public class OmeroReader extends FormatReader {
             LOGGER.info("Logging in");
 
             client = new omero.client(address, port);
-            ServiceFactoryPrx serviceFactory = null;
+            serviceFactory = null;
             if (user != null && pass != null) {
                 serviceFactory = client.createSession(user, pass);
             }
@@ -305,10 +318,6 @@ public class OmeroReader extends FormatReader {
                 }
             }
 
-            // get raw pixels store and pixels
-
-            store = serviceFactory.createRawPixelsStore();
-
             img = (Image) serviceFactory.getContainerService()
                     .getImages("Image", Arrays.asList(iid), null).get(0);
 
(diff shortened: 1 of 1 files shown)


components/blitz/src/loci/ome/io/OmeroReader.java
```diff
@@ -436,60 +461,48 @@ else if (key.equals("iid")) {
...
-        // construct the OMERO reader
-        final OmeroReader omeroReader = new OmeroReader();
-        omeroReader.setUsername(user);
-        omeroReader.setPassword(pass);
-        omeroReader.setServer(server);
-        omeroReader.setPort(port);
-        omeroReader.setGroupName(group);
-        final String id = "omero:iid=" + imageId;
-        try {
+                OmeroReader omeroReader = new OmeroReader()) {
```
mtbc: some indent issue

components/blitz/src/loci/ome/io/OmeroReader.java
```diff
@@ -103,8 +103,20 @@
     private Long groupID = null;
     private boolean encrypted = true;
 
-    private omero.client client;
-    private RawPixelsStorePrx store;
+    //private omero.client client;
```
mtbc: ok to just remove I'd say

components/blitz/src/loci/ome/io/OmeroReader.java
```diff
@@ -470,7 +469,7 @@ else if (key.equals("iid")) {
     public static void main(String[] args) throws Exception {
         try (BufferedReader con = new BufferedReader(
                 new InputStreamReader(System.in, Constants.ENCODING));
-                OmeroReader omeroReader = new OmeroReader()) {
+             OmeroReader omeroReader = new OmeroReader()) {
```
mtbc: getting there! :smiley:

jburel: I kept the indent for ``InputStreamReader`` since it was initialized as part of ``BufferedReader``
I found it easier to understand
happy to change if you think it does not add any value

mtbc: yes, that is fine, but this line still isn't aligned

dominikl:
Looks alright. Briefly tested to run the `main` method and requested some bytes via `openBytes` method, works fine 👍

components/blitz/src/loci/ome/io/OmeroReader.java
```diff
@@ -166,25 +177,43 @@ public boolean isThisType(String name, boolean open) {
...
 
         return buf;
     }
 
     @Override
     public void close(boolean fileOnly) throws IOException {
         super.close(fileOnly);
-        if (!fileOnly && client != null) {
-            client.closeSession();
+        if (!fileOnly) {
```
chris-allan, line 210: I wouldn't have thought of `__del__()` as being required over `closeSession()`.

/cc @joshmoore

components/blitz/src/loci/ome/io/OmeroReader.java
```diff
@@ -422,7 +446,7 @@ else if (key.equals("iid")) {
                 }
             }
 
-            //            store.setImageID("omero:iid=", (int) img.getId().getValue());
+            //store.setImageID("omero:iid=", (int) img.getId().getValue());
```
chris-allan: If you're going to mess with this line I'd just delete it.

chris-allan (changes requested):

components/blitz/src/loci/ome/io/OmeroReader.java
jburel: I opted to match what is currently in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/Connector.java

components/blitz/src/loci/ome/io/OmeroReader.java
jburel: copy/paste but not delete!! ;-)

components/blitz/src/loci/ome/io/OmeroReader.java
chris-allan: Understood. I thought we established during work on #4817 that now we don't need that?

joshmoore: The only difference between closeSession and __del__ is that the latter shouldn't throw.

jburel: I rolled back the change in case user of the reader is handling the error potentially thrown when closing the reader.

components/blitz/src/loci/ome/io/OmeroReader.java
```diff
@@ -436,60 +459,48 @@ else if (key.equals("iid")) {
...
+            omeroReader.setPassword(pass);
+            omeroReader.setServer(server);
+            omeroReader.setPort(port);
+            omeroReader.setGroupName(group);
             omeroReader.setId(id);
-        }
-        catch (Exception e) {
-            omeroReader.close();
+        } catch (Exception e) {
             throw e;
```
chris-allan, line 502: Are we certain that we no longer need to close the reader?

chris-allan (changes requested):

components/blitz/src/loci/ome/io/OmeroReader.java
jburel: This should be taken care of as part of the "try-with-resources".
Can certainly double-check

chris-allan: Ah, it's just weirdly indented with the new try with resources block. Was difficult to follow with just the diff.

chris-allan (approved):

PR #4706, opened 2016-06-12
Url parser improvement

aleksandra-tarkowska:
This PR:

- allows passing more complex view params to top links
- fixes https://trello.com/c/gTnJCAs0/64-bug-no-reverse-for

Testing this PR
- add link like:

 ```
bin/omero config append omero.web.ui.top_links '["Data", {"viewname": "webindex", "query_string": {"experimenter": -1}}, {"title": "Repo"}]'
```
- go to http://host/webclient/login/ and log in, check if `Reverse for '{}' not found.` no longer appears.
- test default as well

cc @will-moore

PR #5035, opened 2017-01-16
Api urls in json

will-moore:
# What this PR does

Adds urls to the api json in the same style as github api, E.g. https://api.github.com/users/will-moore
E,g, when listing ```/projects/``` each will have links that are absolute urls (clickable in browser)
```
"url:datasets": "http://localhost:4080/api/v0.1/m/projects/11601/datasets/",
"url:project": "http://localhost:4080/api/v0.1/m/projects/11601/",
```

# Testing this PR

1. Urls to expect:
    - /projects/ - each project has link to 'datasets' and 'project' as above.
    - /project/:id/ - has a link to 'datasets' only.
    - /datasets/ - each dataset has link to 'images and 'dataset'.
    - /dataset/ - has link to 'images' only.
    - /images/ - each image has link to 'image' only.
    - Same for Screens -> Plates (no links to Wells - not yet supported)
1. Tests cover all urls in P->D->I in ```test_pdi_urls()``` and S>P in ```test_spw_urls()```
    - I have ignored urls in test_api_projects.py (and now test_api_images)
    - And will have to do the same for future tests.

NB: urls can be configured to be "absolute" vv prefixed using the "omero.web.api.absolute_url" setting.

cc @chris-allan @aleksandra-tarkowska @jburel


atarkowska:
regarding your question about absolute uri, my feeling was that by default we will skip domain from api, but I don't remember exactly what we said over the meeting about `build_absolute_uri` https://github.com/openmicroscopy/openmicroscopy/pull/4708#issuecomment-250018188.


EDITED: So to test urls in unittests did you try using `@override_settings()` see example https://github.com/openmicroscopy/openmicroscopy/pull/4706/files#diff-23c6659b076340831a8d1839cea54cfaR94. For integration tests you would need to decide global config set by `bin/omero` and build_absolute_uri in a test?

atarkowska:
**THAT IS AN EXAMPLE OF BUILDING ABSOLUTE URI in UNITTEST to assert values**
Don't use it in django APP

To call `build_absolute_uri` did you try in your tests:

```
from django.http import HttpRequest

def request_helper(url):
  r = HttpRequest()   
  r.META = {
    "SERVER_NAME": "testserver",
    "SERVER_PORT": "80",
  }
  return r.build_absolute_uri(url)

request_helper(reverse('weblogin'))
```

manics:
Is the absolute URI based on the `Host` header sent by the client?

will-moore:
@manics Are you talking about the Django test framework or 'real world'? I think it's Yes either way.

atarkowska:
@manics yes it is. My understanding was that we will omit using `build_absolute_uri` anywhere unless someone configure it https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/settings.py#L475, but that is no longer true https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/api/views.py#L56.

manics:
@will-moore The real world. We had a load of problems with the IDR due to multi-layered proxies, so just wanted to be sure this would work.

will-moore:
Failing test [test_dataset_images](https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/433/testReport/OmeroWeb.test.integration.test_api_images/TestImages/test_dataset_images/) due to ```image_url``` not being ignored.

```
        assert_objects(conn, rsp['data'], [orphaned], dtype='Image',
>                      opts={'load_pixels': True})

test/integration/test_api_images.py:150: 
```
Need to handle this as in test_api_projects.py

components/tools/OmeroWeb/omeroweb/api/views.py
```diff
@@ -111,39 +111,73 @@ def api_servers(request, api_version, **kwargs):
...
+
+    def add_data(self, marshalled, request, **kwargs):
+        """
+        Post-process marshalled object to add any extra data.
+
+        Used to add urls to marshalled json.
+        Subclasses can configure self.urls to specify urls to add.
+        See ProjectsView urls as example
+        """
+        # if self.urls is None:
```
jburel: remove if no longer needed

components/tools/OmeroWeb/omeroweb/api/views.py
```diff
@@ -251,12 +312,28 @@ def get_opts(self, request, **kwargs):
                 opts['project'] = project
         return opts
 
+    # Urls to add to marshalled object. See ProjectsView for more details
+    urls = {
+        'dataset_url': {'name': 'api_dataset',
```
jburel: maybe keep the same order for the url list in each view
i.e. retrieval of children then object e.g. datasets/project

atarkowska: in Python order in dict is not guaranteed (unless you use https://docs.python.org/2.7/library/collections.html#collections.OrderedDict), why order matters?

jburel: I found it easier to read when reviewing the code
This is mainly for review later on when we get back to it

will-moore:
As discussed in today's web meeting, we'll use the ```url:``` prefix for urls e.g. ```url:datasets``` since this seems compatible with ```omero:``` prefix. Updated in commits above.

components/tools/OmeroWeb/test/integration/test_api_containers.py
```diff
@@ -50,6 +50,16 @@ def cmp_name_insensitive(x, y):
     return cmp(unwrap(x.name).lower(), unwrap(y.name).lower())
 
 
+def build_url(client, url_name, url_kwargs):
+    """Build an absolute url using client response url."""
+    response = client.request()
+    # http://testserver/webclient/
+    webclint_url = response.url
```
jburel: typo: I think you mean ``webclient_url``

will-moore:
@jburel fixed typo, thanks

will-moore:
Last commit should fix test https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/435/testReport/OmeroWeb.test.integration.test_api_projects/TestProjects/test_project_create_read/

will-moore:
With Dom's help we refined the query:
```
        params.add('ratingns', rstring(omero.constants.metadata.NSINSIGHTRATING))
        
        q = """
            select sum( case when an.ns = :ratingns and an.class = LongAnnotation
                        then 1 else 0 end),
                sum( case when an.class = LongAnnotation
                        then 1 else 0 end),
               sum( case when an.class != LongAnnotation
                        then 1 else 0 end ), an.class
               from %sAnnotationLink ial 
                        join ial.child as an 
                        join ial.parent as i 
                where i.id = :id 
                    group by an.class
            """ % self.OMERO_CLASS
```

jburel:
@will-moore I guess your comment is for https://github.com/openmicroscopy/openmicroscopy/pull/5023

jburel:
Thanks @will-moore. Merging

PR #5036, opened 2017-01-16
BfPixelBuffer: fix file leak

joshmoore:
An exception thrown during a call to `BfPixelsBuffer.close()`
inadvertently leaked file handles. The call to `reader()` in
`close()` initiated a call to `BfPixelsWrapper.setId()`. When
this threw an exception, the files were left dangling.

# What this PR does

 * Buffer: Don't call `new BfPixelsWrapper` from `close()`
 * Buffer: Detect unused new wrappers and close them
 * Wrapper: close if an exception is called on `setId`

# Testing this PR

Testing may prove fairly difficult. What's been most prone to produce this error is re-thumbnailing Tara on IDR. If "Too many open files" is hit, then a snowball effect is produced when trying to _close_ files leading to more attempts to open files rather than closing _any_.

# Related reading

 * https://github.com/openmicroscopy/bioformats/pull/2438#r65844693
 * https://trello.com/c/AZNC7S6V/411-formatreader-auto-close-file-on-exception

### DIFF
--- a/components/romio/src/ome/io/bioformats/BfPixelBuffer.java
+++ b/components/romio/src/ome/io/bioformats/BfPixelBuffer.java
@@ -58,16 +58,19 @@ public class BfPixelBuffer implements PixelBuffer, Serializable {
         this.bfReader = bfReader;
     }
 
+    /**
+     * Do not call this method when looking to close the reader.
+     * Instead, call `reader.get()` directly.
+     **/
     protected BfPixelsWrapper reader() {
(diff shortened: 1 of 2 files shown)


joshmoore:
Note: this does not seem to have cleared up all file leaks but should handle the particular case of getting 

```
ERROR [ ome.services.blitz.impl.ThumbnailStoreI] (.Server-40) Error on close, stage1
```

joshmoore:
--rebased-to #5040

PR #5037, opened 2017-01-16
Bugfix: canExportAsJpg

dominikl:
# What this PR does

Fixes a bug when image size x/y cannot be retrieved.

# Testing this PR
On webdev-merge as root, open dataset `dataset-b8bf17d5-7ceb-493d-bec8-13964a8897ed`.  Select the image. No exception should thrown.

# Related reading

https://trello.com/c/069qFkHv/257-bug-canexportasjpg

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/container.py
@@ -213,12 +213,16 @@ class BaseContainer(BaseController):
         except:
             limit = 144000000
         if self.image:
-            if (self.image.getSizeX() * self.image.getSizeY()) > limit:
+            if self.image.getSizeX() is None or\
+                    self.image.getSizeY() is None or\
+                    (self.image.getSizeX() * self.image.getSizeY()) > limit:
                 can = False
         elif objDict is not None:
             if 'image' in objDict:
                 for i in objDict['image']:
-                    if (i.getSizeX() * i.getSizeY()) > limit:
+                    if i.getSizeX() is None or\
+                            i.getSizeY() is None or\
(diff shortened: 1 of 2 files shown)


mtbc:
Should we instead track down why this happens? Every image has a size.

will-moore:
The image that has caused all the recent errors of this kind is a 'test image', probably with no pixels: http://web-dev-merge.openmicroscopy.org/webclient/?show=image-1758 (root).

mtbc:
Aha! That makes sense, thank you @will-moore. (Perhaps someday we will merge `Pixels` into `Image`!)

will-moore:
Being a bit picky, but I'd generally prefer to not call getSizeX() & getSizeY() multiple times. Partly because sometimes Blitz Gateway methods can do more work than you expect (in this case it's not so much - just checking that pixels are loaded - but it's easier to be 'safe' than to look at Blitz Gateway code each time) and partly to improve readability, E.g. this is a little easier on the eye:
```
            sizeX = self.image.getSizeX()
            sizeY = self.image.getSizeY()
            if sizeX is None or sizeY is None or (sizeX * sizeY) > limit:
```

will-moore:
Great, thanks.
Good to merge.

jburel:
a test should be added to check that the value is correctly handled

will-moore:
An OmeroWeb integration test to check that the right panel can load for an image with No pixels would cover a lot of places where we need to handle missing pixels.
@dominikl Let me know if you want some pointers?

dominikl:
Added integration test, thanks @will-moore .

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/469/ is green
merging

PR #4957, opened 2016-11-22
Import into other groups

mtbc:
# What this PR does

This PR allows members of the `system` group to import images into any group regardless of membership.

# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ManagedRepositoryTest/testAdminImportIntoAnotherGroup/ should pass and https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/ should be green.

PR #5048, opened 2017-01-18
New roles

pwalczysko:
# What this PR does

This PR introduces the new roles for facility manager workflows. These workflows are facilitated by creating non-full-admin users (called internatlly light admins) which then can execute tasks like ImportAs (importing for other users) or organizing data of other users or users into groups in any group they like. It goes about the permissions restrictions by having a set of privileges in experimenter_config table in the DB. For each of those privileges (or restrictions, they can be either true or false), there is an expected behaviour. 

The accompanying tests deal with two problematics:

  - unwanted expansion of permissions (e.g. by sudoing as true admin, the light admin will NOT get full admin privileges), these are covered mostly in @mtbc 's tests
  - desired workflows feasibility (e.g. can I, equipped with privileges Sudo, Chown, WriteFile and WriteOwned go through a meaningful workflow of ImporterAs ?), these are covered in my tests

# Testing this PR

Mainly check that the accompanying tests as well as all the other integration tests are passing and make sense.

# Related reading

[design issue](https://github.com/openmicroscopy/design/issues/62)
[trello card](https://trello.com/c/t0nT7KYa/133-new-role)
The .ppt explanation of the tests is at ``/ome/team/pwalczysko/new-roles-tests-explanation/Explanation-Of-TestsNewRole.ppt``
Edit: also the Power Point presentation is now as a [g.doc](https://docs.google.com/presentation/d/18kc75LHiItkHr_dvG1R6jW1qmTNayan8x4ZQC6s7XmI/edit#slide=id.g1b11290c7c_2_96) better accessible.
See here for the [table](https://docs.google.com/spreadsheets/d/1HadUCZlIrJoetDZrWfEFYrralCakqLGWPpjK_hHjaqM/edit#gid=415364306) of features.

The facility managers using omero, as well as persons such as image analysts need a free access to other user's data, and in some degrees freedom to create containers, change ownership of the data etc., without having the full powers and responsibilities of the full administrator of OMERO.

cc @jburel @joshmoore @mtbc @bramalingam 

Not ready for review yet. Will add some .ppt clarifying the tests graphically. Also, @mtbc is still working on new features.

--breaking


jburel:
Few unit test failures (travis) e.g.
``
FAILED: testAddLog on null(ome.server.utests.sec.SecuritySystemTest)
[nomemorytestng] org.jmock.core.DynamicMockError: mockLocalQuery: no match found
[nomemorytestng] Invoked: ome.api.IQuery.findByQuery(<FROM Session s LEFT OUTER JOIN FETCH s.sudoer WHERE s.id = :id>, <PARAMS:id=1 >)
[nomemorytestng] Allowed:
``

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -0,0 +1,1305 @@
...
+        } else {
+            Assert.assertNull(sentProj);
+            Assert.assertNull(sentDat);
+        }
+
+        /* check that after sudo, the light admin is able to ImportAs and target
+         * the import into the just created Dataset.
+         * Check thus that the light admin can import and write the original file
+         * on behalf of the normalUser and into the group of normalUser */
+        client.getImplicitContext().put("omero.group", Long.toString(normalUser.groupId));
```
mtbc: Didn't we already do this above?

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -0,0 +1,1305 @@
...
+        ArrayList <String> permissions = new ArrayList <String>();
+        permissions.add(AdminPrivilegeSudo.value);
+        if (permChgrp) permissions.add(AdminPrivilegeChgrp.value);;
+        if (permWriteOwned) permissions.add(AdminPrivilegeWriteOwned.value);
+        if (permWriteFile) permissions.add(AdminPrivilegeWriteFile.value);
+        final EventContext lightAdmin;
+        lightAdmin = loginNewAdmin(isAdmin, permissions);
+        /* set up the project as the normalUser */
+        loginUser(normalUser);
+        client.getImplicitContext().put("omero.group", Long.toString(normalUser.groupId));
```
mtbc: Isn't the user *already* working in this group without needing this set?

mtbc:
@jburel: Gah, I hate these weird tests of implementation details instead of actual behavior. @pwalczysko: Don't worry about those failures, they're just something tedious I'll have to get around to looking at. In the meantime maybe we can set breaking to ignore travis failures.

pwalczysko:
> In the meantime maybe we can set breaking to ignore travis failures.

Done courtesy of @sbesson - Thank you.

pwalczysko:
The .ppt explanation of the tests is at ``/ome/team/pwalczysko/new-roles-tests-explanation/Explanation-Of-TestsNewRole.ppt`` (still working on it)

mtbc:
@pwalczysko: Happy for you to move this to a different branch corresponding with devspace @aleksandra-tarkowska is kindly setting up for us. (We can tweak branch in its CI config.)

components/blitz/src/omero/cmd/graphs/GraphHelper.java
```diff
@@ -51,15 +56,34 @@
...
+    }
+
+    /**
+     * Check if the current user is an administrator.
+     * @param requiredPrivilege the privilege that the administrator must have if they are a light administrator
+     * @return if the current user is an administrator
+     */
+    public boolean checkIsAdministrator(AdminPrivilege requiredPrivilege) {
+        final EventContext eventContext = helper.getEventContext();
+        if (!eventContext.isCurrentUserAdmin()) {
```
joshmoore, line 80: So any privilege makes a user an admin, or for any privilege to be considered valid, the user must be an admin? To some extent, I wonder if the `ec.isCurrentUserAdmin` method doesn't need to be deprecated in favor of this method (refactored up) since the rest of the code-base will be assuming a binary state as opposed to the new spectrum.

mtbc: Yes, one thing I got from the "restrictions" discussion was that light admins are members of `system`. I did experiment with trying to introduce a string set of privileges in event context but got confused over interactions with sudo and intimidated by fear of getting things wrong. My current thinking is to try to make new roles minimally invasive for 5.3 then once we have more time and experience to fit them in more naturally for 5.4 and your suggestion is a prime candidate for how that needs to be done.

mtbc:
While @pwalczysko's away I might be able to fix test failures via #5054.

components/server/src/ome/logic/AdminImpl.java
```diff
@@ -634,13 +640,34 @@ public long createSystemUser(Experimenter newSystemUser) {
...
+    @Transactional(readOnly = false)
+    public long createLightSystemUser(Experimenter newSystemUser, List<AdminPrivilege> privileges) {
+        assertKnownPrivileges(privileges);
+        final long newSystemUserId = createSystemUser(newSystemUser);
+        newSystemUser = iQuery.get(Experimenter.class, newSystemUserId);
+        final Set<AdminPrivilege> restrictions = new HashSet<>(adminPrivileges.getAllPrivileges());
+        restrictions.removeAll(privileges);
+        final List<NamedValue> userConfig = new ArrayList<NamedValue>(restrictions.size());
+        for (final AdminPrivilege restriction : restrictions) {
+            userConfig.add(new NamedValue(restriction.getValue(), Boolean.toString(false)));
```
joshmoore: What does this currently look like in the DB? I'd imagine a prefix might be useful for long-term upgrades, etc.

components/server/src/ome/logic/AdminImpl.java
```diff
@@ -1442,4 +1557,29 @@ private void adminOrPiOfNonUserGroups(ExperimenterGroup defaultGroup,
...
+    }
+
+    /**
+     * @return the light administrator privileges that apply to the current session
+     */
+    private Set<AdminPrivilege> getCurrentAdminPrivilegesForSession() {
+        if (isAdmin()) {
+            final String hql = "FROM Session s LEFT OUTER JOIN FETCH s.sudoer WHERE s.id = :id";
+            final Parameters params = new Parameters().addId(getEventContext().getCurrentSessionId());
+            final ome.model.meta.Session session = iQuery.findByQuery(hql, params);
```
joshmoore, line 1571: Has there been an performance impact by querying for the privileges? If so, we should likely re-use the `SessionCache` infrastructure to pre-load them.

components/server/src/ome/logic/AdminImpl.java
mtbc: You're thinking the prefix to distinguish from other kinds of user configuration? Maybe even no prefix in `AdminPrivilege.value` but only in `Experimenter.config.name`? Could look into that.

joshmoore:
Few design-level comments made. I haven't yet reviewed the server-side tests.

components/server/src/ome/logic/AdminImpl.java
mtbc: Not that I've noticed but I haven't much checked (can changes in CI timings help?) and the internal security infrastructure, which I figure is where the frequent checks occur, is making use of an explicit cache of them, and where they're not in an OMERO cache I'd guess they're probably soon in a Hibernate one. Could certainly look at preloading sudoer somehow? Maybe even preloading privileges depending on how session cache stuff would handle changes in privileges during long-lived session.

mtbc:
`LightAdminRolesTest` alone takes over an hour to run. Might need to reduce parameter combinations for tests that use expensive operations (e.g., import image, upload script).

mtbc:
@sbesson: If we could now switch "breaking" back to requiring travis passes then this PR will happen to come back in to the merge once @pwalczysko has merged in my latest commits and adjusted for the `Write*`/`Delete*` privilege split.

sbesson:
@mtbc changes done

mtbc:
As noted on the card, this PR includes commits from https://github.com/openmicroscopy/openmicroscopy/pull/4957. This will cause intermittent "is not a member of group" errors in integration tests that do imports so do not worry about surprise failures in the test jobs unless they are consistent.

pwalczysko:
@mtbc : so now that Chown is dragging with itself some delete privileges, in order to be able to perform at all in most situation, do we have a definition about which ones precisely ? In other words, what additional (probably deletion) actions will a light admin equipped with Chown be able to do (without being explicitly given any additional privileges) ? Checked the Experimenter_config table in the DB, even when Chown is true, the Delete rows are still false, so it is not explicitly obvous what additional perms were given (maybe none, which would be greatttt....)

mtbc:
The extra delete power now coming with `Chown` should apply *only* when actually doing a chown, not in other circumstances, so other operations should remain as before.

pwalczysko:
yupeee. thanks @mtbc

pwalczysko:
I have the tests running now with Mark's new permissions setup. Also simplified them as much as possbile in order to spare resources - more on this to come.
It seems we have 9 working workflows (back end tested) as outlined in the .ppt https://github.com/openmicroscopy/openmicroscopy/pull/5048#issuecomment-273841100.
Some are "slick", some more clanky. The most used most clanky workflow would be the linking (put someones image into his dataset and then change the ownership of the link) -> additionally this one does not work in private group at all as already discussed.

ToDo for me:
  - write test for User Organizer (user/group manipulator)
  - try to get more light admin workflows out of the nine above checked using clients
  - write test for WriteScriptRepo (as this is rather a safety feature, maybe I could fall back on a test which was already written by @mtbc ?)

pwalczysko:
The LightAdminRoles test takes now 25 minutes on my mac.

jburel:
@pwalczysko any way we can break it?

pwalczysko:
@jburel : We could split the client-side checking (mainly CLI workflows based on the 9 workflows in the test should be checked). For that, one needs to create the light admin and a user(s) according to the setup and necessary permissions bullet points on each slide in my ppt  and go through the workflow on the CLI. But the person testing this must be on this branch here of course.

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -0,0 +1,1330 @@
...
+     * is and is not using <tt>Sudo</tt> privilege
+     * for this action.
+     * @throws Exception unexpected
+     */
+   @Test(dataProvider = "narrowed combined privileges cases")
+   public void testImporterDelete(boolean isAdmin, boolean isSudoing, boolean permDeleteOwned,
+           String groupPermissions) throws Exception {
+       /* only DeleteOwned permission is truly needed for deletion of links, dataset
+        * and image (with original file) when not sudoing */
+       boolean deletePassing = (!isSudoing && permDeleteOwned) || isSudoing;
```
mtbc: this is simply,
```java
final boolean deletePassing = permDeleteOwned || isSudoing;
```

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -0,0 +1,1330 @@
...
+        * created on behalf of the user */
+       if (deletePassing) {
+           try {
+               doChange(Requests.delete().target(datasetImageLink).build());
+               doChange(Requests.delete().target(projectDatasetLink).build());
+               doChange(Requests.delete().target(image).build());
+               doChange(Requests.delete().target(sentDat).build());
+               doChange(Requests.delete().target(sentProj).build());
+           } catch (ServerError se) {
+           /* not expected */
```
mtbc: If it's not expected then why catch it?
Also, if we must catch it, maybe at least have an `Assert.fail` here then.

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -0,0 +1,1330 @@
...
+     * admins importing data into groups they are not member of will get
+     * merged. For this test, combinations of  <tt>Chown</tt>, <tt>WriteOwned</tt>,
+     *  <tt>WriteFile</tt> and <tt>WriteManagedRepo</tt> privileges will be explored
+     * for the light admin. For this workflow the creation and targeting of a Dataset
+     * is tested too.
+     * @throws Exception unexpected
+     */
+
+    @Test(dataProvider = "6 privileges cases")
+    public void testImporterAsNoSudoChownOnlyWorkflow(boolean isAdmin, boolean permWriteOwned, boolean permWriteManagedRepo, boolean permWriteFile, boolean permChown,
```
mtbc: `permWriteOwned` seems to be called `isSudoing` in the provider!

pwalczysko:
The Power Point presentation is now as a [g.doc](https://docs.google.com/presentation/d/18kc75LHiItkHr_dvG1R6jW1qmTNayan8x4ZQC6s7XmI/edit#slide=id.g1b11290c7c_2_96) better accessible.

pwalczysko:
@mtbc : This should fix the issues you raised. The issue with badly named data provider variables I will deal with once we have all the tests.

pwalczysko:
@mtbc : Would you fly over https://github.com/openmicroscopy/openmicroscopy/pull/5048/commits/eb738dd0e3ec0d561c37081bb0c902940c8d119d briefly please ? I hope it is doing what I say in the comments. Definitely no regression/difference from the full admin behaviour on the Rnd settings front.

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -1249,6 +1253,59 @@ public void testChownAllBelongingToUser(boolean isAdmin, boolean permChgrp, bool
         Assert.assertEquals(retrievedProjectDatasetLinkOtherGroup2.getDetails().getOwner().getId().getValue(), recipient.userId);
     }
 
+    /** Test of light amdin without using Sudo.
```
mtbc: admin :smiley:

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -1249,6 +1253,59 @@ public void testChownAllBelongingToUser(boolean isAdmin, boolean permChgrp, bool
...
+            boolean permWriteOwned, String groupPermissions) throws Exception {
+        /* creation of rendering settings should be always permitted as long as light admin is in System Group
+         * and has WriteOwned permissions. Exception is Private group, where it will
+         * always fail.*/
+        if (!isAdmin) return;
+        boolean isExpectSuccessCreateRndSettings = isAdmin && permWriteOwned && !(groupPermissions == "rw----") ;
+        boolean isExpectSuccessCreateAndChown = isExpectSuccessCreateRndSettings && permChown;
+        final EventContext normalUser = newUserAndGroup(groupPermissions);
+        /* set up the light admin's permissions for this test */
+        ArrayList <String> permissions = new ArrayList <String>();
```
mtbc:
```java
List<String> permissions = new ArrayList<String>();
```
(use most general class for variable declaration, don't precede angle brackets with spaces)

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -1249,6 +1253,59 @@ public void testChownAllBelongingToUser(boolean isAdmin, boolean permChgrp, bool
...
+        if (permChown) permissions.add(AdminPrivilegeChown.value);;
```
mtbc: double semicolon

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -1249,6 +1253,59 @@ public void testChownAllBelongingToUser(boolean isAdmin, boolean permChgrp, bool
...
+        if (permWriteOwned) permissions.add(AdminPrivilegeWriteOwned.value);
+        final EventContext lightAdmin;
+        lightAdmin = loginNewAdmin(isAdmin, permissions);
```
mtbc: could do this after the image is created to avoid switching back between normal user and admin so much

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -1249,6 +1253,59 @@ public void testChownAllBelongingToUser(boolean isAdmin, boolean permChgrp, bool
...
+        /* create an image with pixels as normalUser in a group of the normalUser */
+        loginUser(normalUser);
+        client.getImplicitContext().put("omero.group", Long.toString(normalUser.groupId));
```
mtbc: the normal user should already be in their group

pwalczysko:
The last commit https://github.com/openmicroscopy/openmicroscopy/pull/5048/commits/3d02afb9150473089791990591528f799702b9af adds test for ROIs. We have no difference between root and light admin behaviour there (no regression). One inexpected behaviour detected, but not urgent/blocking, basically chown of ROIs in read-only is not possible per se if the image is not included, see [trello card](https://trello.com/c/7o4q2Tkt/155-fix-graphs-for-mixed-ownership-read-only)

will-moore:
I think it's currently possible for 'root' to add restrictions to themselves which we should probably disallow at the server level (in the same way that we currently prevent 'root' from removing themselves from 'system' group).

mtbc:
Re. https://github.com/openmicroscopy/openmicroscopy/pull/5048#issuecomment-278060296 have tried requesting access a couple of times but with no fruit.

will-moore:
It seems that a 'restricted admin' with 'ModifyUser' privilege can edit themselves OR another user to make them a full Admin!

mtbc:
Yes, "fix privilege elevation routes" is the current checklist item on https://trello.com/c/t0nT7KYa/133-new-role. :smiley:

pwalczysko:
@mtbc : Re https://github.com/openmicroscopy/openmicroscopy/pull/5048#issuecomment-279688557 - sorry, this shoiuld be now fixed, and the .ppt is accessible (albeit it is another .ppt, so please use the link as given in the header of this PR)

pwalczysko:
Testing performance:

On new-role branch:
 - on clean DB locally with freshly restarted laptop running ``.../integration/chown/PermissionsTest.java`` takes 284588 ms = 4.7 minutes (this test has 103 methods, resulting in at least 103 groups and users)
  - importing a plate as root into System  ``bin/omero import /Volumes/ome/data_repo/curated/cellworx/pki/`` results in ``193 files uploaded, 1 fileset, 1 plate created, 384 images imported, 0 errors in 0:02:51.375``
  - moving a plate as root (owner) to a different, newly created group called test (root is a member of this group too)
```
[pwalczysko@ls31619 ~/Work/openmicroscopy/dist]$ time bin/omero chgrp test Plate:10
Using session 5b83ee04-81a6-4bdf-b543-b385e5491614 (root@localhost:4064). Idle timeout: 10 min. Current group: system
omero.cmd.Chgrp2 Plate:10 ok

real	0m5.742s
user	0m0.526s
sys	0m0.117s
```
  - deleting the Plate
```
time bin/omero delete Plate:10
Using session 5b83ee04-81a6-4bdf-b543-b385e5491614 (root@localhost:4064). Idle timeout: 10 min. Current group: system
omero.cmd.Delete2 Plate:10 ok

real	0m3.136s
user	0m0.516s
sys	0m0.120s
```

On develop branch (``latest``)
 - on clean DB locally with freshly restarted laptop running ``.../integration/chown/PermissionsTest.java`` takes 222193 ms = 3.7 minutes (this test has 103 methods, resulting in at least 103 groups and users)
 - mporting a plate as root into System  ``bin/omero import /Volumes/ome/data_repo/curated/cellworx/pki/`` results in``193 files uploaded, 1 fileset, 1 plate created, 384 images imported, 0 errors in 0:01:56.001
``
 - moving the plate into a new group test
```
time bin/omero chgrp test Plate:10
Using session 2867b750-0c6a-4e28-b192-c0baf8fdd343 (root@localhost:4064). Idle timeout: 10 min. Current group: system
omero.cmd.Chgrp2 Plate:10 ok

real	0m5.896s
user	0m0.544s
sys	0m0.122s
```
 - deleting the plate
```
time bin/omero delete Plate:10
Using session 2867b750-0c6a-4e28-b192-c0baf8fdd343 (root@localhost:4064). Idle timeout: 10 min. Current group: system
omero.cmd.Delete2 Plate:10 ok

real	0m3.130s
user	0m0.521s
sys	0m0.116s
```

pwalczysko:
@mtbc : Looks like no difference on move and delete of a real plate with 384 images. Some 28-47 percent slower on import and running integration tests (chown).

Will repeat the import of the plate in the morning maybe there is varying performance of my Mac.

Edit: the morning result on roles branch for import gave
``==> Summary
193 files uploaded, 1 fileset, 1 plate created, 384 images imported, 0 errors in 0:02:41.298``
The variation seems to be some 10 secs,  the worsening 38 percent, I think the 30-47 percent above still stands.

pwalczysko:
My 21 tests now take some 22 minutes. This makes it some ~1 minute per test. In the next step, I intend to cut the time to a half approx, as I will remove all isAdmin loops -> this is being tested already by @mtbc 's tests.

mtbc:
Note that my tests usually assume that the light admin has many privileges: I am focused on security testing that a restriction is disabling. Your tests often come from the other side, that a workflow is possible given only certain privileges.

pwalczysko:
Discussed with @mtbc : It is fine to remove the isAdmin = false cases, which is what I am talking about in https://github.com/openmicroscopy/openmicroscopy/pull/5048#issuecomment-282695999 - these cases are actually testing the isRestricted cases in Mark's tests again.

pwalczysko:
This PR was superseded by https://github.com/openmicroscopy/openmicroscopy/pull/5159. Closing.

PR #5050, opened 2017-01-19
Show platename for well tags

dominikl:
# What this PR does

Show the plate name for tags on wells.
screen shot 2017-01-19 at 15 43 56
Without the PR the tag "WellTag" in the example screenshot will only list the well as "B6", with the PR the name additionally shows the plate name, to which the well belongs.

ToDo:
- If you click on the well, nothing is shown in the center panel. What is expected to be shown in that case?

# Testing this PR

Tag a well. Select "Tags" view in the left hand panel. Make sure you find the tagged well under the tag and that the name of the plate is indicated, too.

# Related reading

[Trello - SPW Follow up](https://trello.com/c/GxqDnD0P/211-spw-follow-up)

/cc @will-moore


waxenegger:
Looks good.

image

PR #5053, opened 2017-01-22
render.py: add 'test' migrated from pix.py

joshmoore:
# What this PR does

On the IDR servers, the `pix.py` script has been used for some time to test whether or not a bfmemo file exists for a given image and if `--force` is passed, to create a memo file. This became necessary with the introduction of the server-wide `fail_if_missing`  (gh-4684). 

# Testing this PR

1. Create an image (`bin/omero import foo.fake`)
2. Test that `render test Image:$ID` shows `ok`
3. Delete the bfmemo file for the image
4. Re-run `test Image:$ID`. `miss` should be shown.
5. Run `test Image:$ID --force`. This time: `fill` should be shown.

# Related reading

 * https://github.com/openmicroscopy/openmicroscopy/pull/4684

----

TODO:

 * [ ] define a better name (and/or move to a separate plugin)

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/render.py
+++ b/components/tools/OmeroPy/src/omero/plugins/render.py
@@ -28,6 +28,7 @@ from omero.gateway import BlitzGateway
 from omero.model import Image
 from omero.model import Plate
 from omero.model import Screen
+from omero.rtypes import rint
 from omero.util import pydict_text_io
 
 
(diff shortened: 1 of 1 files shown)


aleksandra-tarkowska:
--exclude

joshmoore:
Removed @aleksandra-tarkowska's `exclude` with a hope of getting it in for IDR team use even if it doesn't hit the mainline.

PR #5054, opened 2017-01-23
New roles

mtbc:
--exclude new roles work

This PR is a temporary placeholder to fix #5048 and capture new work while @pwalczysko is away.


components/server/src/ome/security/basic/OmeroInterceptor.java
```diff
@@ -91,6 +105,11 @@
 
     private static Logger log = LoggerFactory.getLogger(OmeroInterceptor.class);
 
+    /** thread-safe and may be removed without warning;
+     *  used by {@link omero.cmd.graphs.BaseGraphTraversalProcessor#deleteInstances(String, Collection)} */
+    public static SetMultimap<Session, Long> DELETED_GROUP_LINK_IDS =
+            Multimaps.synchronizedSetMultimap(HashMultimap.<Session, Long>create());
```
joshmoore: I know we're still in the "finding a solution" phase, but a global synchronized map worries me.

mtbc: Me too. Could make private and hide behind a public method that does the `put`s but map would still have to be synchronized. (Could make `preFlush` synchronized instead but that felt even worse.) The architecture of having `Delete2` talk directly to `OmeroInterceptor` which then directly uses `dataSource` is horrifying anyway. It's been tricky to write the interceptor code while working around that most of its methods don't have any sense of which session/transaction they're working in. (Indeed, I'm only indexing by sessions at all here because the current transaction from `Delete2`'s point of view appears not to be the one the interceptor is notified of.)

mtbc:
Opened fresh version as #5141 that eliminates some early bothersome commits like those from #4957 so that the commit history is fairly clean.

@pwalczysko, @will-moore: Once CI is pleased with this then not sure how much of a pain it would be for you to adjust your branches to be just your commits against a new `roles` force-push. Absolutely feel free to push your latest then tell me to push a branch that is your commits atop the new `roles` for you to resume working from: if I am causing undue pain then I am happy to use `rebase -i` and whatnot to fix it from my end.

will-moore:
My branch doesn't include your changes. I just started a clean web branch, then I connect to a server that includes your work.

PR #5055, opened 2017-01-23
Search for wells

dominikl:
# What this PR does

Enables the search for wells.

ToDo: 
- Add/Adjust integration (or robot?) test

# Testing this PR

- ~~Enable the indexing of wells: `./omero config set omero.search.include_types "ome.model.core.Image,ome.model.containers.Project,ome.model.containers.Dataset,ome.model.screen.Plate,ome.model.screen.Screen,ome.model.screen.PlateAcquisition,ome.model.screen.Well"`.~~ Not necessary, with 6d56fdf that's the default now anyway.
- Tag a Well.
- Search for the Tag and check that the well shows up in the search results.

# Related reading

[Trello - Critical: search on wells](https://trello.com/c/23gd5W58/120-critical-search-on-wells)

/cc @will-moore

### DIFF
--- a/components/tests/ui/testcases/web/search.txt
+++ b/components/tests/ui/testcases/web/search.txt
@@ -22,6 +22,8 @@ ${XpathPlate}           xpath=//table[@id='dataTable']//img[@alt='plate']
 ${XpathPlateThumb}      xpath=//table[@id='dataTable']//img[contains(@src,'folder_plate16.png')]
 ${XpathScreen}          xpath=//table[@id='dataTable']//img[@alt='screen']
 ${XpathScreenThumb}     xpath=//table[@id='dataTable']//img[contains(@src,'folder_screen16.png')]
+${XpathWell}            xpath=//table[@id='dataTable']//img[@alt='well']
+${XpathWellThumb}       xpath=//table[@id='dataTable']//img[contains(@src,'well16.png')]
 ${XpathTable}           xpath=//table[@id='dataTable']
 ${XpathTableColumn}     xpath=//table[@id='dataTable']//tr/td[position()=3]
 ${XPathAlertText}       xpath=//div[contains(@class,'ui-dialog')][contains(@style,'display: block')]//p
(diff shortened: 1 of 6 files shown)


mtbc:
The code diff looks good.

dominikl:
Might be worth testing with some other annotations, too, e.g. comments.

mtbc:
This seems to be working pretty well so far. If I look on the tags view for the wells of that tag then I don't see anything for those wells in its central pane but I'm guessing that's outwith this PR?

mtbc:
Yes, it works fine with comments too.

will-moore:
@mtbc The only tagged objects we show in the centre panel are image thumbnails. E.g. don't show P/D/ S/P etc. so we wouldn't expect Wells to show up here currently.

mtbc (approved):

dominikl:
Thanks @mtbc . I just still need to add a robot test, one more commit to come.

dominikl:
Added 'wells' to the robot test, but not sure if that is sufficient to test that the search for wells really works, is it @will-moore ?

joshmoore:
NB: for Well search to be usable on existing data, the upgrade guide, we'll have to suggest re-indexing. If that's the case, it might be worth considering if there's anything else to re-index for 5.3.

will-moore:
There's also the question of whether to provide scripts for moving Image annotations to their parent Wells? The IDR workflow of searching for an Image and seeing the Well Table data in the right panel of search results (along-side the Image) is broken in 5.3. Need to discuss if/how to "fix" this.

mtbc:
The upgrade script could create reindex DB entries for existing wells.

dominikl:
In terms of this PR, is it now good to merge, or still something missing? Upgrade scripts, etc. won't be part of this PR, I guess.

will-moore:
I think this is Good to Merge.

joshmoore:
https://github.com/openmicroscopy/openmicroscopy/pull/5055#issuecomment-276315230 or another proposal should be included somewhere in 5.3.0.

jburel:
@will-moore will look at a script see https://trello.com/c/gasafKWk/291-update-script-wellsample-well-annotation-migration

PR #5051, opened 2017-01-20
Map rclass

joshmoore:
# What this PR does

Make it possible to retrieve the `ome.model` class name for a Hibernate subclass, e.g. `select type(a.class) from Annotation a` might return `ome.model.annotations.CommentAnnotation`. Note: Most HQL functions return the discriminator, including: `a.class` and `type(a)`

# Testing this PR

1. have a database with at least one annotation (`bin/omero obj new CommentAnnotation`)
2. query without this PR: `bin/omero hql -q 'select type(a.class) from Annotation a'`
3. This will fail with an exception from IceMapper

cc: @aleksandra-tarkowska

PR #5056, opened 2017-01-24
Map rclass (rebased onto develop)

snoopycrimecop:
This is the same as gh-5051 but rebased onto develop.

----

# What this PR does

Make it possible to retrieve the `ome.model` class name for a Hibernate subclass, e.g. `select type(a.class) from Annotation a` might return `ome.model.annotations.CommentAnnotation`. Note: Most HQL functions return the discriminator, including: `a.class` and `type(a)`

# Testing this PR

1. have a database with at least one annotation (`bin/omero obj new CommentAnnotation`)
2. query without this PR: `bin/omero hql -q 'select type(a.class) from Annotation a'`
3. This will fail with an exception from IceMapper

cc: @aleksandra-tarkowska

### DIFF
--- a/components/blitz/src/omero/util/IceMapper.java
+++ b/components/blitz/src/omero/util/IceMapper.java
@@ -9,6 +9,7 @@
 package omero.util;
 
 import static omero.rtypes.rbool;
+import static omero.rtypes.rclass;
 import static omero.rtypes.rdouble;
 import static omero.rtypes.rfloat;
 import static omero.rtypes.rint;
@@ -422,6 +423,10 @@ public class IceMapper extends ome.util.ModelMapper implements
             Boolean b = (Boolean) o;
             omero.RBool bool = rbool(b.booleanValue());
             return bool;
+        } else if (o instanceof Class) {
+            Class c = (Class) o;
+            omero.RClass rc = rclass(c.getName());
+            return rc;
         } else if (o instanceof Date) {
             Date date = (Date) o;
             omero.RTime time = rtime(date.getTime());
--- a/components/tools/OmeroPy/test/integration/test_iquery.py
+++ b/components/tools/OmeroPy/test/integration/test_iquery.py
@@ -167,7 +167,8 @@ class TestQuery(ITest):
         """
         params = ParametersI()
         params.addString("uuid", uuid)
-        rv = [x[0] for x in unwrap(self.query.projection(query_string, params))]
+        rv = self.query.projection(query_string, params)
+        rv = [x[0] for x in unwrap(rv)]
         assert len(rv) == 2
         assert "ome.model.annotations.CommentAnnotation" in rv
         assert "ome.model.annotations.TagAnnotation" in rv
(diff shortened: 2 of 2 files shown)


snoopycrimecop:
--rebased-from #5051

components/blitz/src/omero/util/IceMapper.java
```diff
@@ -422,6 +423,10 @@ public RType toRType(Object o) throws omero.ApiUsageException {
             Boolean b = (Boolean) o;
             omero.RBool bool = rbool(b.booleanValue());
             return bool;
+        } else if (o instanceof Class) {
+            Class c = (Class) o;
+            omero.RClass rc = rclass(c.getName());
```
jburel, line 428: could directly do ``return rclass(c.getName());``
The same applies for the other check

components/tools/OmeroPy/test/integration/test_iquery.py
```diff
@@ -150,3 +151,16 @@ def testQueryTaggedUnique(self):
         for idx in range(len(result1)-1):
             # Omit final since == isn't defined for Ice objects.
             assert result1[idx] == result2[idx]
+
+    def testClassType(self):
+        created = []
+        for Ann in (CommentAnnotationI, TagAnnotationI):
```
jburel, line 159: Maybe check other annotation type

components/blitz/src/omero/util/IceMapper.java
joshmoore: Didn't seem the right time to be modifying the class more significantly, so to keep the change here minimal, I copied the style of the rest of the checks.

components/tools/OmeroPy/test/integration/test_iquery.py
joshmoore: This was mostly to show that nothing weird was happening in terms of mapping (statics in `IceMapper` etc.). I don't expect there to be any value in adding more.

components/blitz/src/omero/util/IceMapper.java
jburel: understood, 
can do that in another PR
This was more a note.

joshmoore:
Test fix pushed.

joshmoore:
Test fix pushed.

jburel:
``./components/tools/OmeroPy/test/integration/test_iquery.py:170:80: E501 line too long (80 > 79 characters)``
flake8 issue

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.test_iquery/ is green

jburel:
Without this PR
``
{
    serverStackTrace = 
    serverExceptionClass = 
    message = Unsupported conversion to rtype from java.lang.Class:class ome.model.annotations.CommentAnnotation
}
``
with this PR
``
bin/omero hql -q 'select type(a.class) from Annotation a'
 # | Col1                                    
---+-----------------------------------------
 0 | ome.model.annotations.CommentAnnotation 
 1 | ome.model.annotations.ListAnnotation    
 2 | ome.model.annotations.TermAnnotation 
``
I tested with various annotation types

Tests are green,
merging

PR #5058, opened 2017-01-25
Api child count

will-moore:
# What this PR does

 - Uses "data" wrapper on all json returned from GET, POST & PUT for single object
 - Updates version to ```0```. E.g. ```/api/v0/m/screens/```
 - Uses separate queries for loading childCount
 - Adds childCount support for single objects, e.g.  ```/api/v0/m/projects/:id/```

# Testing this PR

1. Integration tests are updated to handle "data" wrapper and to test childCount for Projects, Datasets and Screens (both single P/D/S and lists).
 2. Manually testing:
    - browse to E.g. ```/api/v0/m/projects/:id/``` and check that project json is within a ```"data": {}``` wrapper.
    - Check that adding ```?childCount=true``` adds an ```omero:childCount``` to the Project json.
    - ```?childCount=true``` should also work for /projects/, /datasets/, /screens/ and for single /datasets/:id and /screens/:id/

# Related reading

This PR removes child_count support from ```conn.buildQuery()``` and instead this is provided by the web api itself in ```api_query.get_child_counts()```. This is partly due to the small performance improvement (see trello card) but also because it makes the  ```conn.buildQuery()``` more robust. Previously if child_count was True, this returned a query that required ```queryService.projection()``` instead of simply ```queryService.findAllByQuery()```.

1. https://trello.com/c/mN819Y4B/263-childcount-performance


components/tools/OmeroWeb/omeroweb/api/api_query.py
```diff
@@ -20,12 +20,36 @@
...
 from api_marshal import marshal_objects
 from copy import deepcopy
 
 
+def get_child_counts(conn, link_class, parent_ids):
+    """
+    Count child links for the specified parent_ids.
+
+    @param conn:        BlitzGateway
+    @param link_class:  E.g. 'ProjectDatasetLink'
```
jburel: e.g.

components/tools/OmeroWeb/test/integration/test_api_errors.py
```diff
@@ -117,8 +117,9 @@ def test_security_violation(self, group_B, user_A):
         payload = {'Name': 'test_security_violation',
                    '@type': OME_SCHEMA_URL + '#Project'}
         save_url_grpA = save_url + '?group=' + str(groupAid)
-        pr_json = _csrf_post_json(django_client, save_url_grpA, payload,
-                                  status_code=201)
+        rsp = _csrf_post_json(django_client, save_url_grpA, payload,
```
jburel: maybe use ``save_url_grp_a`` instead of capital

components/tools/OmeroWeb/test/integration/test_api_login.py
```diff
@@ -71,7 +71,7 @@ def test_base_url_versions_404(self):
         """
         Tests that the base url gives 404 for invalid versions
         """
-        version = '0'
+        version = '0.1'
```
jburel: maybe better to access the api version from ``settings.py`` and add +1 for example so we do not have to be updating both

components/tools/OmeroWeb/test/integration/test_api_projects.py
```diff
@@ -561,11 +566,13 @@ def test_project_update(self, user1):
...
-        assert rsp['Name'] == 'updated name'
-        assert 'Description' not in rsp
+        project_json = rsp['data']
+        assert project_json['@id'] == project.id.val
+        assert project_json['Name'] == 'updated name'
+        assert 'Description' not in project_json
         # Get project again to check update
-        pr_json = _get_response_json(django_client, project_url, {})
+        rsp = _get_response_json(django_client, project_url, {})
+        pr_json = rsp['data']
```
jburel, line 575: maybe add another test starting with no description,  insert description and check if it is correctly updated

jburel:
manual tests passed, ``childCount`` parameter is ignored (as expected) when a plate is loaded

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -3115,7 +3101,7 @@ def buildQuery(self, obj_type, ids=None, params=None, attributes=None,
                             offset, limit and owner for all objects.
                             Also 'order_by': 'obj.name' to order results.
                             Additional opts handled by _getQueryString()
-                            E.g. 'childCount', or filter Dataset by 'project'
+                            E.g. filter Dataset by 'project'
```
jburel: e.g.

jburel:
probably expand the tests to check the plate results even if the ``childCount`` is given in the query

components/tools/OmeroWeb/omeroweb/api/api_query.py
will-moore: This is really the start of a sentence, similar to lines above and below that are all capitalised, and I think it's easier to read like that.

will-moore:
We already test Plate results with & without childCount in the parametrized ```test_datasets_plates()``` see https://github.com/openmicroscopy/openmicroscopy/pull/5058/files#diff-9f0ceee092d1ecc9ac8d4711ff7536c7R261.

components/tools/OmeroWeb/omeroweb/api/api_query.py
jburel: I know it is code but you should never start a sentence if ``e.g.`` anyway
better to have a description then ``e.g.``

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel: This one does not start a sentence

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5999,7 +5985,6 @@ def _getQueryString(cls, opts=None):
                  "left outer join fetch obj.screenLinks spl "
                  "left outer join fetch spl.parent sc")
         # NB: we don't use base _getQueryString.
```
jburel, line 5987: Do you want to remove the comment now that the rest has been removed

components/tools/OmeroWeb/omeroweb/api/api_query.py
jburel: same comment.

components/tools/OmeroWeb/test/integration/test_api_errors.py
```diff
@@ -116,9 +116,10 @@ def test_security_violation(self, group_B, user_A):
...
-        save_url_grpA = save_url + '?group=' + str(groupAid)
-        pr_json = _csrf_post_json(django_client, save_url_grpA, payload,
-                                  status_code=201)
+        save_url_grp_a = save_url + '?group=' + str(groupAid)
+        rsp = _csrf_post_json(django_client, save_url_grp_a, payload,
+                              status_code=201)
+        pr_json = rsp['data']
         projectId = pr_json['@id']
         # Try to save again into group B
         save_url_grpB = save_url + '?group=' + str(groupBid)
```
jburel: sorry missed it ``save_url_grpB`` => ``save_url_grp_b``

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: No. That comment is still valid. It's just that child_count is not supported anywhere now, so the rest of the comment wasn't relevant now.

jburel: understood

jburel:
test passing. Thanks

PR #5060, opened 2017-01-27
deprecate IAdmin.moveToCommonSpace

mtbc:
# What this PR does

Deprecates `IAdmin.moveToCommonSpace` following the pattern of deprecating the other synchronous methods that can now be done otherwise. Also propagates those deprecations to Java annotations.

Will look at reducing internal usage of such once code churn in related classes subsides so as to avoid conflicts.

--exclude because it may not be desirable

# Testing this PR

CI should remain green.

### DIFF
--- a/components/blitz/resources/omero/api/IAdmin.ice
+++ b/components/blitz/resources/omero/api/IAdmin.ice
@@ -1,6 +1,4 @@
 /*
- *   $Id$
- *
  *   Copyright 2010-2014 Glencoe Software, Inc. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  *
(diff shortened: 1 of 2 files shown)


joshmoore:
Can you give an example of doing this otherwise? Note the use of [secureFlush](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/logic/AdminImpl.java#L1056)

mtbc:
Ah, perhaps it can't be? I was expecting we could use `Chgrp2` but maybe not: I've not been able to make sense of the `SecureAction` stuff and perhaps appropriate permissions stuff can't easily be built into `Chgrp2`. `BasicSecuritySystem.doAction` seems key and perhaps tokens too. Maybe it partly depends on [the ticket's](http://trac.openmicroscopy.org/ome/ticket/1794) idea of,

> only admins can add to the common pot

mtbc:
Taking as undesirable until I better understand the existing code.

PR #5063, opened 2017-01-30
Omero properties cleanup

dominikl:
# What this PR does

- Removes deprecated properties  `omero.client.ui.menu.dropdown.leaders` `omero.client.ui.menu.dropdown.everyone` and `omero.client.ui.menu.dropdown.colleagues `
- Add default thumbnail size as client property `omero.client.browser.thumb_default_size`

# Testing this PR

Check that integration test `test_thumbnails.py` still passes.

# Related reading

[Trello - omero properties](https://trello.com/c/Ib4Rk0bw/25-omero-properties)

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1492,6 +1492,8 @@ class _BlitzGateway (object):
         self._defaultOmeroUser = None
         self._maxPlaneSize = None
 
+        self._defaultThumbnailSize = None
+
         self._connected = False
         self._user = None
         self._userid = None
@@ -1547,9 +1549,21 @@ class _BlitzGateway (object):
         if self._maxPlaneSize is None:
             c = self.getConfigService()
             self._maxPlaneSize = (
-                int(c.getConfigValue('omero.pixeldata.max_plane_width')),
-                int(c.getConfigValue('omero.pixeldata.max_plane_height')))
+                int(c.getConfigValue('omero.client.pixeldata.max_plane_width')),
+                int(c.getConfigValue('omero.client.pixeldata.max_plane_height')))
         return self._maxPlaneSize
+
+    def getDefaultThumbnailSize(self):
+        """
+        Returns the default thumbnail size
+        """
+        if self._defaultThumbnailSize is None:
+            try:
+                c = self.getConfigService()
+                self._defaultThumbnailSize = int(c.getConfigValue('omero.client.browser.thumb_default_size'))
+            except:
+                self._defaultThumbnailSize = 96
+        return self._defaultThumbnailSize
 
     def getClientSettings(self):
         """
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -331,14 +331,16 @@ def render_thumbnail(request, iid, w=None, h=None, conn=None, _defcb=None,
     """
     server_id = request.session['connector'].server_id
     direct = True
+    defaultSize = conn.getDefaultThumbnailSize()
+
     if w is None:
-        size = (96,)
+        size = (defaultSize,)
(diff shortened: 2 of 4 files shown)


etc/omero.properties
```diff
@@ -868,6 +862,9 @@ omero.client.viewer.roi_limit=2000
 # sorting them by default client ordering strategy
 omero.client.ui.tree.type_order=tagset,tag,project,dataset,screen,plate,acquisition,image
 
+# The default thumbnail size
+omero.client.browser.thumb_default_size=96
```
atarkowska, line 866: You have to define a bean for new property in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/common/resources/ome/config.xml

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -331,14 +331,16 @@ def render_thumbnail(request, iid, w=None, h=None, conn=None, _defcb=None,
     """
     server_id = request.session['connector'].server_id
     direct = True
+    defaultSize = conn.getDefaultThumbnailSize()
```
atarkowska, line 334: This will slows down thumbnail loading, why do you need that value on each request? that should be set once when connection is created

etc/omero.properties
dominikl: Thanks. Thought, I'm missing something.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
atarkowska: It may be easier to just add it to omeroweb.settings instead, unless there is a need to share it between clients

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -554,8 +556,8 @@ def getConfigValue(key):
             logger.warn("webgateway: get_shape_thumbnail() could not get"
                         " Config-Value for %s" % key)
             pass
-    max_plane_width = getConfigValue("omero.pixeldata.max_plane_width")
-    max_plane_height = getConfigValue("omero.pixeldata.max_plane_height")
+    max_plane_width = getConfigValue("omero.client.pixeldata.max_plane_width")
```
atarkowska, line 559: `omero.client.pixeldata.max_plane*` doesn't exist. why do we change that?

components/tools/OmeroWeb/omeroweb/webgateway/views.py
dominikl: It will just be requested from the server the first time the method is called, the value is held in `_defaultThumbnailSize` variable in the Gateway

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1547,10 +1549,22 @@ def getMaxPlaneSize(self):
...
 
+    def getDefaultThumbnailSize(self):
+        """
+        Returns the default thumbnail size
+        """
+        if self._defaultThumbnailSize is None:
+            try:
+                c = self.getConfigService()
+                self._defaultThumbnailSize = int(c.getConfigValue('omero.client.browser.thumb_default_size'))
+            except:
```
atarkowska, line 1564: could you be specific which exception you are trying to catch?

components/tools/OmeroWeb/omeroweb/webgateway/views.py
atarkowska: sorry didn't read the implementation carefully

components/tools/OmeroWeb/omeroweb/webgateway/views.py
dominikl: Ta, the 'omero.client.pixeldata.*` changes slipped in by accident ,thought I reverted them before committing...

dominikl:
Closing for now, will reopen when I fixed the various issues.

PR #5065, opened 2017-01-31
Omero properties cleanup

dominikl:
# What this PR does
(Replaces  closed PR #5063 )

- Removes deprecated properties  `omero.client.ui.menu.dropdown.leaders` `omero.client.ui.menu.dropdown.everyone` and `omero.client.ui.menu.dropdown.colleagues `
- Add default thumbnail size as client property `omero.client.browser.thumb_default_size`

# Testing this PR

Check that integration test `test_thumbnails.py` still passes.

# Related reading

[Trello - omero properties](https://trello.com/c/Ib4Rk0bw/25-omero-properties)

### DIFF
--- a/components/tools/OmeroWeb/test/integration/test_config.py
+++ b/components/tools/OmeroWeb/test/integration/test_config.py
@@ -131,7 +131,6 @@ class TestConfig(ITest):
         key = "omero.client.ui.menu.dropdown.%s" % prop
         try:
             self.rs.setConfigValue(key, label)
-            print "set %s to %s " % (key, label)
             # test load_server_settings directly
             login_required(default_view).load_server_settings(
                 self.conn, self.r)
--- a/components/tools/OmeroWeb/test/integration/test_config.py
+++ b/components/tools/OmeroWeb/test/integration/test_config.py
@@ -34,6 +34,8 @@ from omeroweb.decorators import login_required
 
 from omero.gateway.utils import propertiesToDict
 
+import pytest
+
 
 def default_view(request):
     pass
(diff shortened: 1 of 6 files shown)


will-moore:
```test_thumbnails.py``` is [passing](https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/444/testReport/OmeroWeb.test.integration.test_thumbnails/).

Do we have any upgrade path (manual or auto upgrade) for users moving from these deprecated configs to their replacement (maybe this was already done prior to 5.3?)

As I understand it, the removed tests cover the "setting" of old values and "getting" of the new values. Do we have other tests that simply set and get the new values? If not then would be best to update the tests to do that.

cc @aleksandra-tarkowska

**Comment by will-moore** [comment]
Test looks good, thanks.
Since we're already using the newer settings ```...leaders.label``` etc in 5.2 we don't need to worry about upgrade for 5.3.
Good to merge.

**Comment by joshmoore** [comment]
However, the various cleanup code here would have redirected user-set values from the old to the new values. So if no user action has upgraded the values, they will now be effectively unset, right?

File: components/tools/OmeroWeb/test/integration/test_config.py [hunk 1]
```diff
@@ -141,16 +122,16 @@ def testDefaultConfigConversion(self):
...
                                       "everyone.label"])
     @pytest.mark.parametrize("label", ["foo"])
     def testUpgradeDropdownMenuConfig(self, prop, label):
-        """ Test if alias loads deprecated property value """
+        """ Test to set and get DropdownMenuConfig """
         d = self.rs.getClientConfigDefaults()
         key = "omero.client.ui.menu.dropdown.%s" % prop
         try:
             self.rs.setConfigValue(key, label)
+            print "set %s to %s " % (key, label)
```
jburel: maybe remove the print, if it was used for debugging

dominikl: Thanks, yes was just used for debugging.

dominikl:
Not quite sure how to proceed here.... @joshmoore , @will-moore ?

will-moore:
@joshmoore Yes, I think you are right. Older users who haven't updated to use newer settings will need to for 5.3. Needs a docs PR for 'upgrade' (rather than a script)?

jburel:
Either we clearly state in the upgrade doc that the properties have been removed or we keep them for 5.3.0 and clearly stipulate that they will be removed in 5.3.1 to give people a chance

jburel:
Discussed yesterday during board review. Indicate in the doc the change.
Those properties have mainly been customised for learning.

jburel:
Option updated https://github.com/openmicroscopy/ome-documentation/pull/1641

jburel:
@aleksandra-tarkowska adjustments might be required in learning when it is upgraded

PR #5079, opened 2017-02-02
Deprecate test

jburel:
# What this PR does

Deprecate the reader test.

# Testing this PR
Nothing to test
# Related reading

https://trello.com/c/b8jrapBD/174-omero-reader-out-of-blitz-jar


dominikl:
Makes sense to deprecate the test as the OMEROReader has been deprecated, good to merge 👍

jburel:
I don't think the test will  be re-used in its current form too.

PR #5084, opened 2017-02-03
Channel processor

jburel:
# What this PR does

This PR sets the default color in RGB in sequence when more than 3 channels
instead of RGBBBB etc.

# Testing this PR

 * Integration test checking that the values are correctly set when using the rendering service
 * unit test for channel processor change 
 * Import one of the images below and check that the channels are RGB RGB
 * Open and image and change the color. Save the settings. Reset the default

# Related reading

See https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8210
and https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8211

cc @sbesson @bramalingam


bramalingam:
Everything looks good! Looks good to merge.

jburel:
@chris-allan: this proposed change will turn the multi-channels images to RGBRGB.. instead of RGBBBBB... if no color found in the file or no metadata in the file found to set the color e.g. Jerry's 30  channels image

PR #5086, opened 2017-02-06
Rename javadoc zip bundle

hflynn:
# What this PR does

Renames the zip bundle from 'OMERO-docs-version.zip' to 'OMERO-javadoc-version.zip' to avoid confusion with the new documentation bundle and match the approach used in BF. 

# Testing this PR

Ask @jburel if you're not sure, I assume something along the lines of `ant release-docs -Domero.release="5.3.0-m6"` works but I can't test locally as I don't have ice etc installed.

# Related reading

See https://trello.com/c/SnUesylG/505-ditching-docs-pdfs

Will need edits made to the downloads page which could be added to https://github.com/openmicroscopy/ome-release/pull/211

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -453,8 +453,8 @@ To get started using Eclipse, execute "ant build-dev" and import the top-level
     </target>
 
     <target name="release-docs" depends="release-javadoc,release-sphinx-api,release-slice2html" description="Generate Docs for all components under dist/docs/api">
-        <zip destfile="${omero.home}/target/OMERO.javadoc-${omero.version}.zip">
-            <zipfileset dir="${dist.dir}/docs" prefix="OMERO.javadoc-${omero.version}"/>
+        <zip destfile="${omero.home}/target/OMERO.apidoc-${omero.version}.zip">
+            <zipfileset dir="${dist.dir}/docs" prefix="OMERO.apidoc-${omero.version}"/>
         </zip>
     </target>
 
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -453,7 +453,7 @@ To get started using Eclipse, execute "ant build-dev" and import the top-level
     </target>
 
     <target name="release-docs" depends="release-javadoc,release-sphinx-api,release-slice2html" description="Generate Docs for all components under dist/docs/api">
-        <zip destfile="${omero.home}/target/OMERO.docs-${omero.version}.zip">
+        <zip destfile="${omero.home}/target/OMERO.javadoc-${omero.version}.zip">
             <zipfileset dir="${dist.dir}/docs" prefix="OMERO.docs-${omero.version}"/>
```
mtbc: the `OMERO.docs` in this line can remain as-is?

hflynn: I'm honestly not sure! I ummed and ahhed about changing that too but given I can't test locally, I've failed to work out if I need to

sbesson: See https://ant.apache.org/manual/Types/zipfileset.html - this variable setss the name of the prefix in the archive, in other terms the name of the folder containing the content that gets created when unarchived. We certainly want to have both the archive name and the prefix matching.

hflynn: Ah thanks! I'll change that too then

hflynn:
Alternatively, we could use 'apidoc' to match the downloads page but no-one has commented to the relevant card to object to 'javadoc'.

hflynn:
I don't *think* the build error is related to this PR

build.xml
jburel: it is not only javadoc that's why the term ``docs`` has been used. Sorry I missed the trello notification

hflynn: What do we want to use then - apidocs? api-docs? something else?

jburel:
apidoc will make sense @hflynn

hflynn:
👍 I'll push that change shortly @jburel

jburel:
tested locally:
``ant release-docs -Domero.release="5.3.0-m8"`` generated ``OMERO.apidoc-5.3.0-m8-ice36-SNAPSHOT.zip``

Merging

PR #5091, opened 2017-02-08
Disable cache

atarkowska:
# What this PR does

disable jquery cache by default

https://trello.com/c/J266hcfA/261-ajaxsetup-cache-false

# Testing this PR

check if ajax request contains timestamp, see  trello card for more description

# Related reading

http://api.jquery.com/jquery.ajax/

cc: @will-moore


waxenegger:
I have done a good 15 minutes of using the web with the chrome dev tools open and checking requests in the network tab. I could not find any xhr request other than POST requests that did not have a timestamp appended, and that's correct and in agreement with what http://api.jquery.com/jquery.ajax/ says.

atarkowska:
--rebased-to #5133

PR #5093, opened 2017-02-08
Rgba test

jburel:
# What this PR does
Introduced unit tests for color encoding/decoding in the pojos

# Testing this PR

Make sure the tests are green
If you want to run them locally
``./build.py -f components/blitz/build.xml test``

# Related reading

https://trello.com/c/sGlYbbvB/213-color-in-shape-settings

cc @dominikl

### DIFF
--- a/components/blitz/test/omero/gateway/model/ShapeSettingsTest.java
+++ b/components/blitz/test/omero/gateway/model/ShapeSettingsTest.java
@@ -27,12 +27,11 @@ import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
-import omero.model.RectangleI;
 import omero.model.Shape;
 
 
 /**
- * 
+ * Tests the encoding/decoding of some properties of ShapeSettingsData
  * @author Jean-Marie Burel &nbsp;&nbsp;&nbsp;&nbsp;
  * <a href="mailto:j.burel@dundee.ac.uk">j.burel@dundee.ac.uk</a>
  * @since 5.3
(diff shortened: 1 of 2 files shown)


components/blitz/test/omero/gateway/model/ShapeSettingsTest.java
```diff
@@ -0,0 +1,107 @@
...
+ * <a href="mailto:j.burel@dundee.ac.uk">j.burel@dundee.ac.uk</a>
+ * @since 5.3
+ */
+@Test(groups = "unit")
+public class ShapeSettingsTest {
+
+    @DataProvider(name = "colors")
+    public static Object[][] colors() {
+       return new Integer[][] {
+               {Integer.valueOf(255), Integer.valueOf(0), Integer.valueOf(0),
```
mtbc: don't just the plain numbers suffice without all the `Integer.valueOf` stuff?

jburel: opted for that way of declaring the value since it will generate warning in Java 9

mtbc: Gah. I don't suppose you recall what the warning is? Just curious.

jburel: deprecated!! ``new Integer()``
but I have not checked if i specify a "int" it will be by default converted using ``Integer.valueOf``

mtbc: Bizarre: sounds like a compiler bug given that creating an Integer array is completely different from calling `Integer`'s constructor!

jburel: I will simplify this code and see what happen when I run it in Java 9 :-)

dominikl:
Looks good, unit tests passed (but didn't test with java9, java8 only). Good to merge 👍

jburel:
@dominikl: ignore Java 9 for now ;-)

PR #5101, opened 2017-02-13
New role webadmin

will-moore:
# What this PR does

Webadmin UI for setting Admin Privileges.

Key Changes:
 - 'Role' field to choose between 'User', 'Administrator' or 'Restricted Admin'
 - 'Role' field is disabled if you are editing yourself (as before: can't remove yourself from being Admin)
 - Can't remove 'System' group from the groups chooser (no X to click). This would select 'User' role and also cause the 'restricted admin' privilege list to disappear which would be a bit disconcerting.
 - When saving as 'Administrator', ALL Privileges will get set to 'true'.
 - When saving as 'User' role, privileges are not wiped clean on server BUT they appear to be all unset when you edit the user again, since ```adminService.getAdminPrivileges()``` returns empty list.
 - 'Write' privilege includes 'WriteOwned', 'WriteFile' & 'WriteManagedRepo'. IF just 1 or 2 of these are set to 'true' via the API, the UI will show 'Write' as unchecked. Save will result in all 3 privileges -> false. Same for 'Delete', and 'Upload Script' (Write/DeleteScriptRepo).

# Testing this PR

 - Create a user with 'Role': 'User', 'Administrator' or 'Restricted Admin'.
    - See that 'system' group is added to 'Groups' chooser when 'Admin' or 'Restricted Admin' chosen.
 - Choose from list of Privileges.
 - Save then 'Edit' user again to see what's been saved. Check 'Role' and 'Privileges' as expected.
 - Edit user with different combinations - again, check what's Saved is preserved.
 - Try setting/getting Admin Privileges via API and checking in UI, especially with partial 'Write' / 'Delete' privileges as described above.

# Related reading

See: https://github.com/openmicroscopy/design/issues/62

screen shot 2017-02-15 at 14 12 04

 TODO:
 - [x] hide the privileges list to only show it alongside 'Role' when 'Restricted Admin' is chosen.
 - [x] Make sure we save changes in correct order so that we don't have Admin with restrictions failed to save.
 - [x] Fix help url to point to docs page (when we know the url)!

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2215,6 +2215,18 @@ class _BlitzGateway (object):
         group = admin_service.getGroup(self.getEventContext().groupId)
         return ExperimenterGroupWrapper(self, group)
 
+    def getAdminPrivileges(self, user_id=None):
+        """
+        Returns list of privileges as strings.
+
+        :return:    List of strings such as ["ModifyUser", "ModifyGroup"]
+        """
+        if user_id is None:
+            user_id = self.getUserId()
+        privileges = self.getAdminService().getAdminPrivileges(
+            omero.model.ExperimenterI(user_id, False))
+        return [unwrap(p.getValue()) for p in privileges]
+
     def isAdmin(self):
         """
         Checks if a user has administration privileges.
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2215,14 +2215,21 @@ class _BlitzGateway (object):
         group = admin_service.getGroup(self.getEventContext().groupId)
         return ExperimenterGroupWrapper(self, group)
 
-    def getAdminPrivileges(self, user_id=None):
+    def getCurrentAdminPrivileges(self):
         """
-        Returns list of privileges as strings.
+        Returns list of Admin Privileges for the current session.
 
         :return:    List of strings such as ["ModifyUser", "ModifyGroup"]
         """
-        if user_id is None:
-            user_id = self.getUserId()
+        privileges = self.getAdminService().getCurrentAdminPrivileges()
+        return [unwrap(p.getValue()) for p in privileges]
+
+    def getAdminPrivileges(self, user_id):
+        """
+        Returns list of Admin Privileges for the specified user.
+
+        :return:    List of strings such as ["ModifyUser", "ModifyGroup"]
+        """
         privileges = self.getAdminService().getAdminPrivileges(
             omero.model.ExperimenterI(user_id, False))
         return [unwrap(p.getValue()) for p in privileges]
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -168,7 +168,7 @@ class ExperimenterForm(NonASCIIForm):
                     required=False,
                     label=role[1],
                     widget=forms.CheckboxInput(attrs={'class': 'privilege',
-                        'disabled': disabled})
+                                                      'disabled': disabled})
                 ))
             )
 
@@ -192,8 +192,7 @@ class ExperimenterForm(NonASCIIForm):
                 reason = "You can't edit your own admin privileges"
             elif experimenter_root:
                 reason = "You can't edit 'root' user's admin privileges"
-            self.fields['role'].widget.attrs['title'] = \
-                "Removal of your own admin rights would be un-doable"
+            self.fields['role'].widget.attrs['title'] = reason
 
     omename = OmeNameField(
         max_length=250,
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -149,13 +149,13 @@ class ExperimenterForm(NonASCIIForm):
 
         roles = [('Sudo', 'Sudo'),
                  # combine WriteFile/MagangedRepo/Owned roles into 'Write'
-                 ('Write', 'Write'),
+                 ('Write', 'Write Data'),
                  # combine DeleteFile/MagangedRepo/Owned roles into 'Delete'
-                 ('Delete', 'Delete'),
+                 ('Delete', 'Delete Data'),
                  ('Chgrp', 'Chgrp'),
                  ('Chown', 'Chown'),
-                 ('ModifyGroup', 'Edit Groups'),
-                 ('ModifyUser', 'Edit Users'),
+                 ('ModifyGroup', 'Create and Edit Groups'),
+                 ('ModifyUser', 'Create and Edit Users'),
                  ('ModifyGroupMembership', 'Add Users to Groups'),
                  ('Script', 'Upload Scripts')]
         for role in roles:
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -197,7 +197,13 @@ class ExperimenterForm(NonASCIIForm):
         widget=forms.TextInput(attrs={'size': 30, 'autocomplete': 'off'}),
         required=False)
 
-    role = forms.ChoiceField(choices=ROLE_CHOICES, widget=forms.RadioSelect)
+    # 'Role' is disabled if experimenter is 'admin' or self,
+    # so required=False to avoid validation error.
+    role = forms.ChoiceField(
+        choices=ROLE_CHOICES,
+        widget=forms.RadioSelect,
+        required=False,
+        initial='user')
     active = forms.BooleanField(required=False)
 
     def clean_confirmation(self):
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -163,7 +163,7 @@ class ExperimenterForm(NonASCIIForm):
                 (role[0], forms.BooleanField(
                     required=False,
                     label=role[1],
-                    widget=forms.CheckboxInput(attrs={'class':'privilege'})
+                    widget=forms.CheckboxInput(attrs={'class': 'privilege'})
                 ))
             )
 
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -138,10 +138,26 @@ class ExperimenterForm(NonASCIIForm):
                 'email', 'institution', 'administrator', 'active',
                 'default_group', 'other_groups']
 
+        ordered_fields = [(k, self.fields[k]) for k in fields_key_order]
+
+        roles = ['Chgrp',
+                 'Chown',
+                 # combine DeleteFile/MagangedRepo/Owned roles into 'Delete'
+                 'Delete',
+                 'DeleteScriptRepo',
+                 'ModifyGroup',
+                 'ModifyGroupMembership',
+                 'ModifyUser',
+                 'ReadSession',
+                 'Sudo',
+                 # combine WriteFile/MagangedRepo/Owned roles into 'Write'
+                 'Write',
+                 'WriteScriptRepo']
+        for role in roles:
+            ordered_fields.append((role, forms.BooleanField(required=False)))
+
         # Django 1.8: Form.fields uses OrderedDict from the collections module.
-        self.fields = OrderedDict(
-            (k, self.fields[k])
-            for k in fields_key_order)
+        self.fields = OrderedDict(ordered_fields)
 
         if experimenter_is_me_or_system:
             self.fields['omename'].widget.attrs['readonly'] = True
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -97,6 +97,7 @@ class ExperimenterForm(NonASCIIForm):
     def __init__(self, name_check=False, email_check=False,
                  experimenter_is_me_or_system=False,
                  experimenter_me=False,
+                 can_modify_user=True,
                  experimenter_root=False,
                  can_edit_role=True, *args, **kwargs):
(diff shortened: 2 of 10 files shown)


will-moore:
@pwalczysko I think this is ready for initial testing now.
The most significant TODO is to only show the privileges list when 'Restricted Admin' is chosen.

pwalczysko:
First suggestions:
Chown --> Change ownership of data
Chgrp --> Move data between groups
Write --> Write Data
Delete --> Delete Data

It is confusing that the new light admin should be in a group. The default, expected situation is that it is just a member of the system group. But maybe that was like that for full admins as well (do I have to choose "system" group into the box in order to be able to create the user ?)

jburel:
``/components/tools/OmeroWeb/omeroweb/webadmin/forms.py:166:62: E231 missing whitespace after ':'``

pwalczysko:
Maybe:
  - Edit users --> Create/Edit users
  - Edit groups --> Create/Edit groups

In fact, both functionalities are covered, it is just a matter of communication whether or not Create action is somehow implicitly contained in Edit (not for me) ?

pwalczysko:
@will-moore : Created a table which maps your checkboxes here with the underlying permissions. Please check that we are on the same page https://docs.google.com/spreadsheets/d/1HadUCZlIrJoetDZrWfEFYrralCakqLGWPpjK_hHjaqM/edit#gid=2044648906

pwalczysko:
Re: The short explanatory text on Write feature:
``'User can create data and assign another user as the owner. Also allows editing of other users\' data',``

Is principally true but

   - ...assign another user as the owner" is not possible in any client we know (Web, Insight, CLI) atm. Possible only in the API (you have to use .setOwner on the object and then iUpdate method.)
    - we shoiuld mention that you can Link (and thus by extension **Annotate**) other user's data
    - also this privilege gives you the right to create data for yourself in the group you are not a member of (binds with the first bullet point, you might find very difficult in some situations to assign another user as the owner, even using chown will be at times cumbersome, so you might decide as analyst to put your attachment as yourself to users image)

Maybe:
``User can create data in groups of which he/she is not a member of. Also allows annotating, adding attachments to and editing of other users data.``

pwalczysko:
Re: ``User can move others' data to a different Owner``

Unless this is the canonical way to formulate it, I would go for 
``User can transfer others' data to a different Owner``
transfer is what Colin coined for the action in the Python test comments.

pwalczysko:
Re: https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-280301697 - confirmed with @sbesson : no way to achieve the immediate ownership assignment in the CLI atm. I would think we should reformulate the hint sentence.

pwalczysko:
Doc PR see https://github.com/openmicroscopy/ome-documentation/pull/1617

pwalczysko:
Happy with last changes, reads good to me.

pwalczysko:
Suggestion:  In the screenshot below, note the alternate usage of the word "user" and "administrator" for the same person (Administrator with... in radiobutton, ...User privileges in the header of checkboxes (but link to administrator privileges), then again in the first Sudo description User (and another "user," and "of that user" - this obviously should be the "user", just trying to point out the slight possible confusion).
screen shot 2017-02-17 at 09 59 23

Suggestion: Stick to "Administrator" in all the comments and headers, or, if too long a word, define an abbreviation in the radio button "Administrator with restricted privileges (AWRP)" and use it in the comments.

pwalczysko:
screen shot 2017-02-17 at 09 57 56

Should probably be ``... User can create and edit....`` (add ``create``)

pwalczysko:
ditto https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-280609555 for groups (create and edit in the comment)

will-moore:
As discussed in Roles meeting now: UI should only allow Full Admins to create users who are Restricted/Full admins. This is a conservative approach that will prevent Restricted Admins with Edit-User or Group-User privileges from creating more users with "clones" of their own permissions.

joshmoore:
> UI should only allow Full Admins to create users

So the server will still allow it?

mtbc:
Yes, server will still allow some light admins to create some light admins, if that's okay.

joshmoore:
I guess the situation I would like to avoid is that the clients don't know how to interpret what's possible when a server-allowed thing happens. So as long as the `can*` methods are such that a client will always assume less is possible if it encounters that there's something it doesn't know about, then things will hopefully work out.

mtbc:
An important concern indeed, thank you for raising it. Technically the client always has the information to calculate manually but, yes, for the `can*` methods we already have it would be great if what they say is correct and right now it may not be. @jburel, @pwalczysko: Some integration testing would be helpful here: wrongly broken ones I'll try to fix. Or I'll try to get to it if you don't first.

will-moore:
The clients don't currently check any of the new Admin Privileges before attempting to do them. We only check if Admin, and if so then you can go ahead and try to Chgrp others' data or Edit Groups/Users etc.
So, Restricted Admins will get security violations etc if they try these.
There may be some simple catches we can add to prevent some of the worst of these, but how to do it right will not be possible for 5.3.0.

pwalczysko:
We just discussed with @mtbc about https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-281399510 - the consequence of this situation are "mild crashes" of the Web UI when something non-allowed is attempted by light admin. We draw the attention to it in the "note on clients" in the https://github.com/openmicroscopy/ome-documentation/pull/1617 doc, but also @mtbc suggested we might put a point about this into What's new @hflynn

mtbc:
Thank you both. Maybe the issue needn't block 5.3.0 at least then.

hflynn:
I still need to add the new role to the what's new page for sysadmins, I can include a warning about this there and in the Help doc.

pwalczysko:
Testing of the functionality of this PR:
Create a series of light admins where each consecutive light admin has one privilege only (one checkbox checked in the UI). Start from the top with ``Sudo``, etc.

Done by now:
   - Sudo only ( -> table has 14 rows, Sudo is not listed, as expected, the rest is false)
   - Write Data only (-> table has 12 rows, WriteOwned, WriteManagedRepo and WriteFile not listed, as expected)
   - Delete Data only (-> table has 12 rows, DeleteOwned, DeleteManagedRepo and DeleteFile not listed, as expected)

Checking done by checking the psql table ``SELECT * FROM Experimenter_Config WHERE experimenter_id=1054;``

Example of the output (Delete Data):
screen shot 2017-02-21 at 20 28 09

@mtbc : I think that we can just trust that when the row is not listed, it means that it is TRUE, right ? Other than that, I will just check all the privileges as listed in the UI in the coming days.

pwalczysko:
ToDo for the testing as per https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-281472022: 
  - Chgrp
  - Chown
  - Create and Edit Groups
  - Create and Edit Users
  - Add Users to Groups
  - Upload Scripts

mtbc:
> when the row is not listed, it means that it is TRUE, right ?

It does.

pwalczysko:
Further testing of the newly created admins:
 
 - Chgrp ( -> table has 14 rows, Chgrp is not listed, as expected, the rest is false, as expected)
 - Chown ( -> table has 14 rows, Chown is not listed, as expected, the rest is false)
 - Create and Edit Groups (-> table has 14 rows, ModifyGroup is not listed, the rest is false, as expected)
 - Create and Edit Users (-> table has 14 rows, ModifyUser is not listed, the rest is false, as expected)
 - Add Users to Groups (-> table has 14 rows, ModifyGroupMembership is not listed, the rest is false, as expected)
 - Upload Scripts (-> table has 13 rows, DeleteScriptRepo and WriteScriptRepo are not listed, the rest is false, as expected)

Test of the 4 workflow light admins
 - Data Viewer (no box ticked) (-> table has 15 rows, all listed, all false, as expected)
 - Importer (was tested already, see above, this is Sudo only option)
 - Analyst (Write File, Chown and Upload Script ticked) (-> table has 9 rows, Chown, WriteManagedRepo, WriteFile, WriteOwned, WriteScriptRepo, DeleteScriptRepo are not listed, the rest is false, as expected)
 - Group and Data Organizer (all ticked except Upload Script and Sudo) (-> table has 4 rows, DeleteScriptRepo, WriteScriptRepo, Sudo and ReadSession are the only listed items, as expected)

pwalczysko:
The https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-281472022 and https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-281655507 complete the light admin creation testing.  All as expected.

ToDo here (as discussed with @will-moore already) : we should try to get a better graphical correspondence between the horizontal level of the checkbox names and the checkboxes themselves

screen shot 2017-02-22 at 12 28 36

Note the checkboxes tend to be a bit lower, which is disturbing.

will-moore:
To test last commits:
 - Only a Full Admin should be able to edit the "Role" of another user (but not themselves).
 - Restricted Admins will see that the Role part of the form is disabled. Also, if they try to add System to the list of user's groups, this will get rejected on Enter.
 - If Restricted Admin is editing a user who is already a Restricted Admin, they will see that the Privileges checkboxes are also disabled.
 - Check that User cannot edit their own Username, Active or Role fields.
 - Also alignment of checkboxes should be fixed.

pwalczysko:
Not included in build. Will test on Monday.

pwalczysko:
Testing of https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-282041754.

  Ad 2. Re
>Also, if they try to add System to the list of user's groups, this will get rejected on Enter. 

This test failed as I have a workaround through which I was still able to smuggle a user through into the System group. See screenshot below. 

Workaround:
  - select System in the dropdown menu of the ``Groups`` box first, and confirm. As @will-moore indicated, this will be "rejected", but... 
  - see that the "System" group now appears on your lower menu, the Default group menu. 
  - Select the System group as the default group and then
  - add it after that also into the ``Groups``, (this action was previously kind of rejected). 
  - But now, as the default is set to System, the rejection fails -> you successfully added a user to System group (confirmed by pressing Save and navigating back to that user from the Users tab -> the user is now listed as an admin

screen shot 2017-02-27 at 17 11 19


screen shot 2017-02-27 at 17 11 19

pwalczysko:
Ad 
>f Restricted Admin is editing a user who is already a Restricted Admin, they will see that the Privileges checkboxes are also disabled.

Yes, this test passed. But a logical extension of this is "...if Restricted Admin is creating a user who they intend to be a Restricted Admin...." -> nothing is disabled then, and a crash of the UI follows after I press Save on the newly created Restricted admin

pwalczysko:
Ad 
>Check that User cannot edit their own Username, Active or Role fields.

Yes, but 
  - non-admin user cannot edit their own name and surname either (and probably anything else either), ended with crash (cannot update Experimenter...)
  - light admin user cannot edit their own name or surname either

Definite regression, not expected at all, compared with eel merge build - all works there for normal users.

pwalczysko:
In Summary:
Bugs: 
  - https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-282786881(not secure enough block for system group addition), 
  - https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-282794309 (regression on normal user's functionality)

RFE: 
  - https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-282790766 - illogical crash on creation is offered instead of greying out options analogically to Edit case

pwalczysko:
Bug: 

  - the security checks in https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-282041754 do work only when the ``Edit` of another user is attempted but
  - there are no UI security checks on ``Create New User`` -> crash ensues when a light admin tries to create another admin

will-moore:
That should fix the issues listed at https://trello.com/c/rKDxVMSK/49-webclient-webadmin-bugs.
In addition:
 - If user can't edit/modify groups then the groups list page won't show the "Add new Group" button and if you edit a group then all fields and the Save button will be disabled.
 - Same if user can't edit/modify users.

pwalczysko:
Save button not active in any case, neither in New user nor in New Group situation.
screen shot 2017-06-08 at 11 08 28

  - log in as root
  - try to create new user (does not have to be light admin) or new group
  - observe Save button disabled as in screenshot

pwalczysko:
Light admin cannot create a new user which would be an admin or light admin (the radio button for selection of admin or light admin is disabled as expected)  - fixes https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-296202375.

Light admin cannot use the "workaround" via default group menu to get the newly created user into system group - the system group never appears in the default group menu now, even when selected previously (and rejected as expected) in the group selection box - fixes https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-282786881

pwalczysko:
Bug:  Downgrading of own privileges (being light admin) on error in web:

  - have a light admin which 
        - has ``Create....Users`` and ``Create....Groups`` but
        - does not have ``Add users to groups``
  - login as this light admin
  - in admin panel, edit yourself
  - add yourself to one additional group
  - click Save, observe error
  - navigate back to the form and see that now you do not have any privileges checkboxes ticked, all your privileges are gone
  - login as root and confirm that the privileges from that light admin were really removed -> not expected

pwalczysko:
Pretty happy with the behavour now.

Two minor points:

   - light admin which has no ``Add users to groups`` but has ``Create groups`` will experience a "mild" error after they try to add more users in already created groups or add any members into a newly created groups (the group will be created with no members in the end, which is fair I guess)
   - light admin which has no ``Add users to groups`` but has ``Create users``, will experience no crash and an apparent success when they try to add or remove groups from other users group membership lists, but the action will in fact not proceed, which is as expected -> the non-expected part is the lack of warning about the action not taking place (if you do not have the background information, you falsely conclude that you successfully modified the group membership of the edited user)


Happy to merge and list the two above points in another place.

will-moore:
I think to handle the different combinations of permissions "properly" with all the user-friendly messages and correct enabled/disabled state on various form inputs would need a bit more scoping and discussion, so a new PR for that would be best. Card created at https://trello.com/c/H5mMBNFk/51-webadmin-create-link-user-group-combinations

pwalczysko:
Thank you @will-moore . Ready to merge FMPOV.

pwalczysko:
This not being merged is giving me testing problems with other Will's PRs

PR #5106, opened 2017-02-15
PixelsService: use Memoizer.failIfMissing (rebased onto develop)

jburel:
This is the same as gh-4684 but rebased onto develop.

----

To prevent DoS'ing a server, this flag can be set which
requires that all files have a valid cached Memoizer. A
next step will be to allow selectively permitting access
based on some flag.

See https://github.com/openmicroscopy/bioformats/pull/2435

NB: The boolean added to PixelsService should likely be refactored
    into a more general "options" class to reduce future method
    clutter.

This prevent the server from being taken down by missing bfmemo files. Note: if `fail_if_missing` is set to true, then sysadmins will need to run a bfmemo generation script after upgrade
cc @hflynn 

Ref https://trello.com/c/cKEf9iDK/240-other-metadata52-rebases

### DIFF
--- a/components/romio/src/ome/io/nio/PixelsService.java
+++ b/components/romio/src/ome/io/nio/PixelsService.java
@@ -764,6 +764,7 @@ public class PixelsService extends AbstractFileSystemService
      * @param filePath Non-null.
      * @param series series to use
      * @param store Min/max store to use with the min/max calculator.
+     * @param raise an exception if no cache file is available
      */
     protected BfPixelBuffer createMinMaxBfPixelBuffer(final String filePath,
                                                       final int series,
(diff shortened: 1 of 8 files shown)


jburel:
--rebased-from #4684

joshmoore:
NB: A CLI tool also exists to check and/or initialize bfmemo files. Barring a good name (currently: `bin/omero render test`) this should be included with this work.

jburel:
Was "render" not going to be a "pip installable" plugin with its own repo?

joshmoore:
Yes, and the check doesn't need to stay in `render.py` but if someone activates this option they _will_ need a tool to initialize bfmemo files.

jburel:
Ok, Let's discuss when you get back

joshmoore:
Sorry, I didn't realize that `render.py` hadn't been backported. I will try to focus on getting that `pip`-installable then independently of this PR.

jburel:
This PR cannot be rebased since it will require some changes in B-F

Discussed with @sbesson, there is no current concrete plan to ``back-port`` changes done in  B-F for IDR on memoizer to the main B-F line
cf. 
https://trello.com/c/eC6Peivz/39-options-api-priorities

PR #4563, opened 2016-04-07
Full viewer button link

will-moore:
See https://trello.com/c/vRHQSsTs/59-full-viewer-button-to-be-link and https://github.com/openmicroscopy/openmicroscopy/issues/4558

This PR makes the thumbnails and the Full Viewer buttons into html ```<a>``` links which means you can right-click and "Open link in new tab".
To test:
 - Right-click on thumbnail -> open link in new tab. Repeat on other thumbnails.
 - Right-click on "Full Viewer" button -> open link in new tab.
 - Test "Full Viewer" button in General tab *and* Preview tab.
 - Check that images have been opened in a series of new tabs.

Also check that other clicks on thumbnails and button are still working OK - E.g. double-click thumb to open image in new window, selecting images, shift-select range of thumbs etc.

Issue #4558, opened 2016-04-05
web-client: open full viewer on a new browser tab


carandraug:
When navigating the web client, I see 2 ways to open the full viewer:

* double click on the image thumbnail;
* click on the "full viewer" button in the right pane.

However, none of them allows to open the full viewer in a separate tab of the browser, it is always on a new window (with minimal browser interface). Since there is no link for the full viewer, I can't see a place where one would be able to do right click for "Open Link in New Tab".

Ideally, the following two things would be nice:

* "Full Viewer" button in the right pane is an actual link for the full viewer (but I guess there's a reason for why using `OME.openPopup('/webclient/img_detail/...` instead of a link straight to it -- I'm guessing that's to use a window with less browser clutter)
* RIght click on the web client thumbnail would bring up a context menu instead of the default browser right click menu. This context menu would include a link to that image full viewer.

---

User case: going through the thumbnails of a dataset, and select a series of them for a closer look and discussion. Currently this opens several new windows which are difficult to navigate. Multiple tabs in a single window have an order which is lost when using multiple windows.

It is possible to workaround this with a tag, but a tag suggest a more permanent classification, while the purpose of this is for the selection to expire at the end of the meeting. Our suggested workflow in the mean time is to tag all the interesting images (we are using "viewer stack"), then use the full viewer (with the "< Prev" and "Next >" buttons) on that tag.

An alternative for this would be to have the "< Prev" and "Next >" buttons on the full viewer go through the currently selected images (instead of all images in the dataset) but that sounds like substantially more work.

will-moore:
Thanks for the suggestion.
It certainly should be possible to make both the "Full Viewer" button and the thumbnail itself into regular links, which you would then be able to right-click > Open Link in new Tab.

For a long time we have liked the idea of providing a custom right-click menu on thumbnails, but have not got around to this yet.
In the meantime it would be quite easy to add "Open in new tab" to the right-click menu for images in the tree.

Having the Prev & Next buttons only go through selected images would be quite easy to do, but I'm not sure the user workflow would be that nice. You'd have to double-click a single image to open the viewer, then go back to the webclient and select multiple images, then return to the image viewer to use Next button. This is not very intuitive.
What might be nicer is to enable a "Full Viewer" button when multiple images are selected. This could open multiple image viewers in separate windows or separate tabs.

will-moore:
[...]

PR #5110, opened 2017-02-17
Open viewer new tab

will-moore:
# What this PR does

This changes the default behaviour when opening images in viewer to open images in a new Tab instead of a popup window.
See https://trello.com/c/KPGIBH9i/270-open-popup-window-url-editable

# Testing this PR

Behaviour has changed in a number of places. Each of these should open a new Tab:

- Double-click on image in jsTree.
- Open with... from jsTree
- Open new image created from script in "Activities" panel
- "Full Viewer" buttons in General and Preview tabs
- Double-click on thumbnail in Dataset
- Double-click on Well in Plate
- Public page: double click in jsTree and thumbnails
- In "Run Script" dialog, click on "Show Script" in footer.

# Related reading

See [Usability Issues](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Usability_issues) with opening new windows. "Generally speaking, it is preferable to avoid resorting to window.open() for several reasons:.."

Previously users have asked to be able to open image viewer in new Tab:
https://github.com/openmicroscopy/openmicroscopy/pull/4563 from https://github.com/openmicroscopy/openmicroscopy/issues/4558

We've also noticed that popup window is too small for iViewer.

cc @waxenegger

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -556,8 +556,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
           "Selected objects are added to the url as ?image=:1&image=2"
           "Objects supported must be specified in options with"
           "E.g. ``{\"supported_objects\":[\"images\"]}`` "
-          "to enable viewer for one or more images, "
-          "``{\"target\":\"_blank\"}`` to open in new tab.")],
+          "to enable viewer for one or more images.")],
 
     # PIPELINE 1.3.20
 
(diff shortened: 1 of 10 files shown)


jburel:
@will-moore: you will have to adjust the robot tests in ``view_image.txt`` since the expectation is to have the image opened in a new window

waxenegger:
@will-moore: out of curiosity ... is it difficult/problematic to maybe append the dataset id to the url (query string) so that any default viewer could make use of that info. I know this is not "open_with" but I'm thinking it would be handy. The motivation for my question is to disambiguate images  for the iviewer (one image - multiples datasets) but even if you wanted to work with a dataset instead of an image it would be useful.

waxenegger:
Working as expected. All the items in the test list open in a new tab, images as well as the script file.

will-moore:
@waxenegger We open the image viewer from a lot of different places, usually just with the Image ID in hand. However, we could create an OME.function for trying to get the jsTree and traverse it to get the datasetId (similar to iViewer's open_with.js) if that would be useful for other viewers cc @chris-allan.

jburel:
This should be an RFE, card created https://trello.com/c/x72GFRg4/15-open-image-new-tab-append-parent-id

PR #5111, opened 2017-02-18
Paths to object hql queries

will-moore:
# What this PR does

Small fixes to hql queries: see https://trello.com/c/VEXIxsT1/237-review-paths-to-objects-hql-queries

# Testing this PR

1. Paths to object should still work (existing tests passing etc).
1. ```/webclient/api/paths_to_object/?image=1``` should work for images in Dataset or Well.

# Related reading

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/show.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/show.py
@@ -862,7 +862,7 @@ def paths_to_well_image(conn, params, well_id=None, image_id=None,
         from WellSample wellsample
         left outer join wellsample.details.owner wsowner
         left outer join wellsample.plateAcquisition acquisition
-        left outer join wellsample.details.owner aowner
+        left outer join acquisition.details.owner aowner
         join wellsample.well well
         left outer join well.plate plate
         left outer join plate.details.owner plowner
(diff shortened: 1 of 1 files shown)


components/tools/OmeroWeb/omeroweb/webclient/show.py
```diff
@@ -860,12 +860,12 @@ def paths_to_well_image(conn, params, well_id=None, image_id=None,
                acquisition.id,
                well.id
         from WellSample wellsample
-        left outer join wellsample.details.owner wsowner
+        join wellsample.details.owner wsowner
         left outer join wellsample.plateAcquisition acquisition
-        left outer join wellsample.details.owner aowner
+        left outer join acquisition.details.owner aowner
```
jburel, line 865: should be a ``join``
Do you need the owner for plate/plate acquisition/well sample?
It is unlikely to be different.

will-moore: @jburel I think if you remove the ```left outer``` from joining acquisition owner then you will not retrieve any Wells that have no PlateAcquisition (since they will also not have any PlateAcquisition owner).

jburel: I will run some tests

jburel: seems right

jburel:
sorry for not looking at it before
tests are green and api check works

PR #5121, opened 2017-02-23
Python testlib

jburel:
# What this PR does

Clean up ``testlib`` for first release.
Reduce the number of "import" methods so we do not overload the testing library
This will now be a "public" tool useful to test external app e.g. mapr

# Testing this PR

 * Check if further cleaning is required
 * Check that the tests are still green

# Related reading

See https://trello.com/c/ECAKLbUr/224-omero-testlib

### DIFF
--- a/components/tools/OmeroPy/src/omero/testlib/__init__.py
+++ b/components/tools/OmeroPy/src/omero/testlib/__init__.py
@@ -440,8 +440,8 @@ class ITest(object):
                     for the_t in range(size_t):
                         plane_2d = fromfunction(f, (size_y, size_x),
                                                 dtype=int16)
-                        script_utils.uploadPlane(raw_pixel_store, plane_2d,
-                                                 the_z, the_c, the_t)
+                        script_utils.upload_plane(raw_pixel_store, plane_2d,
+                                                  the_z, the_c, the_t)
                         min_value = min(min_value, plane_2d.min())
                         max_value = max(max_value, plane_2d.max())
                 pixels_service.setChannelGlobalMinMax(pixels_id, the_c,
@@ -450,10 +450,10 @@ class ITest(object):
                 rgba = None
                 if the_c in colour_map:
                     rgba = colour_map[the_c]
-                script_utils.resetRenderingSettings(rendering_engine,
-                                                    pixels_id, the_c,
-                                                    min_value, max_value,
-                                                    rgba)
+                script_utils.reset_rendering_settings(rendering_engine,
+                                                      pixels_id, the_c,
+                                                      min_value, max_value,
+                                                      rgba)
         finally:
             rendering_engine.close()
             raw_pixel_store.close()
--- a/components/tools/OmeroPy/src/omero/testlib/__init__.py
+++ b/components/tools/OmeroPy/src/omero/testlib/__init__.py
@@ -1241,78 +1241,6 @@ class AbstractRepoTest(ITest):
         finally:
             rfs.close()
 
-    def assertWrite(self, mrepo2, filename, ofile):
-        def _write(rfs):
-            try:
-                rfs.write("bye", 0, 3)
-                assert "bye" == rfs.read(0, 3)
-                # Resetting for other expectations
-                rfs.truncate(2)
-                rfs.write("hi", 0, 2)
-                assert "hi" == rfs.read(0, 2)
-            finally:
-                rfs.close()
-
-        # TODO: fileById is always "r"
-        # rfs = mrepo2.fileById(ofile.id.val)
-        # _write(rfs)
-
-        rfs = mrepo2.file(filename, "rw")
-        _write(rfs)
-
-    def assertNoWrite(self, mrepo2, filename, ofile):
-        def _nowrite(rfs):
-            try:
-                pytest.raises(omero.SecurityViolation,
-                              rfs.write, "bye", 0, 3)
-                assert "hi" == rfs.read(0, 2)
-            finally:
-                rfs.close()
-
-        rfs = mrepo2.fileById(ofile.id.val)
-        _nowrite(rfs)
-
-        rfs = mrepo2.file(filename, "r")
-        _nowrite(rfs)
-
-        # Can't even acquire a writeable-rfs.
-        pytest.raises(omero.SecurityViolation,
-                      mrepo2.file, filename, "rw")
-
-    def assertDirWrite(self, mrepo2, dirname):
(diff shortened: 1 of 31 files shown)


sbesson:
As the semantics of all import methods is covered by the central `import_mif`. this looks fine from my perspective. A couple of additional comments in the perspective of giving this a public exposition:

- `import_mif` do we want to deprecate and rename the  public name to be simply `import_images` or even `import`?
- in the same vein, is there an argument for renaming `series_count` as `image_count` since series only refers to the internal implementation using BF fakes
- what's the use case of `seriesCount=0`? should we change the default to `1` instead as this would simplify most of the default client calls?
- in terms of best practice for the usage of the library, as per the method definition, we might want to pass arguments as key/value pairs rather than using implicit positional arguments

jburel:
@sbesson:
 * I did not yet check if there is a use case for ``seriescount = 0``, for me 1 should be the default
 * probably no need to deprecate since 5.3.0 will be the first public release of it so I will vote for import_images too

jburel:
Actually ``seriesCount = 0`` seems to be used for plate

sbesson:
That makes complete sense as `import_plates` overrides the classical `images/series` semantics of the `FakeReader` as the number of images is imposed by the SPW structure. If this is the major use case, I would definitely vote for upping the default to 1, reduce the amount of hardcoding across tests and specifically override in the case of plates. I will also check how the `FakeReader` handles the mismatch.

Update: after double checking the SPW semantics seems to unconditionally override `seriesCount` so passing a zero value is technically not even required.

jburel:
I think we also need to remove the usage of  ``tinyTest.d3d.dv`` in the library before we release

components/tools/OmeroPy/src/omero/testlib/__init__.py
```diff
@@ -693,15 +693,13 @@ def create_pixels(self, x=10, y=10, z=10, c=3, t=50, pixels_type="int8",
...
     def get_pixels_type(cls, pixels_type):
-        if pixels_type in ["int8", "uint8"]:
-            return 1
         if pixels_type in ["int16", "uint16"]:
             return 2
         if pixels_type in ["int32", "uint132", "float"]:
             return 4
         if pixels_type in ["double"]:
             return 8
-        return -1
```
joshmoore: Perhaps this should throw? I assume that what would have happened if `-1` got passed to the server.

components/tools/OmeroPy/src/omero/testlib/__init__.py
```diff
@@ -1440,10 +1304,95 @@ def fullImport(self, client):
...
+    # Assert methods
     def assertImport(self, client, proc, folder):
         hashes = self.upload_folder(proc, folder)
         handle = proc.verifyUpload(hashes)
         cb = CmdCallbackI(client, handle)
         rsp = self.assertPasses(cb)
         assert 1 == len(rsp.pixels)
         return rsp
+
+    def assertWrite(self, mrepo2, filename, ofile):
```
atarkowska: as we already started to follow pep standards why these doesn't follow naming nomenclature?

components/tools/OmeroPy/test/integration/test_ishare.py
```diff
@@ -23,6 +23,25 @@
 
 class TestIShare(ITest):
 
+    @classmethod
+    def create_share(cls, description="", timeout=None,
```
atarkowska, line 28: if share is deprecated now, how about if we add warning to test helpers as well?

atarkowska (commented):
in general this looks fine to me. What is the best strategy to rebase this library against m52 branch? I had difficulty to do it smoothly.

components/tools/OmeroPy/src/omero/testlib/__init__.py
jburel: it returns ``1`` by default. I pushed a commit to that effect afterwards so it did not change the existing expectation

components/tools/OmeroPy/src/omero/testlib/__init__.py
jburel: Will do that. Still few things to clean up.
Need to stabilise first the tests :-)

components/tools/OmeroPy/test/integration/test_ishare.py
jburel: good idea

components/tools/OmeroPy/src/omero/testlib/__init__.py
```diff
@@ -296,49 +298,17 @@ def import_image(self, filename=None, client=None, extra_args=None,
...
-    Creates a fake file with a seriesCount of images, imports
-    the file and then return the list of images.
-    """
-
-    def import_mif(self, series_count=0, name=None, client=None,
-                   with_companion=False, skip="all", **kwargs):
+    def import_fake_file(self, images_count=1, name=None, client=None,
+                         with_companion=False, skip="all", **kwargs):
+        """
+        Creates a fake file with a seriesCount of images, imports
```
sbesson: `s/seriesCount/images_count`. Otherwise, the name of the new method makes sense and is representative of the called behavior.

will-moore:
Not part of this PR, but this docstring needs a cleanup (still mentions Dataset and project):
```
    def make_file_annotation(self, name=None, binary=None, format=None,
                             client=None, ns=None):
        """
        Creates a new DatasetI instance and returns the persisted object.
        If no name has been provided, a UUID string shall be used.
        :param name: the name of the project
        :param client: The client to use to create the object
        :param ns: The namespace for the annotation
        """
```

jburel:
@will-moore: fix the description of the method and also adjust the code

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/ no test failure due to this PR

will-moore:
Good to merge from me.
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/469/testReport/OmeroWeb.test.integration.test_tree/TestTree/test_marshal_images_thumb_version_thumb0_/ failures 
```AttributeError: 'TestTree' object has no attribute 'createTestImage' ``` need to be fixed in a different PR (probably https://github.com/openmicroscopy/openmicroscopy/pull/5138?) cc @aleksandra-tarkowska

will-moore:
Last commits look good.
All green today at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/471/

PR #5122, opened 2017-02-24
OMERO.Java example: Add example to link table to an image

bramalingam:
# What this PR does

Adds logic to attach the table to an image.

# Testing this PR

Check if the integration tests are passing.
@jburel 

# Related reading

https://trello.com/c/UvRT57ql/551-omero-tables

### DIFF
--- a/examples/Training/java/src/training/HowToUseTables.java
+++ b/examples/Training/java/src/training/HowToUseTables.java
@@ -38,7 +38,6 @@ import omero.model.ImageAnnotationLink;
 import omero.model.ImageAnnotationLinkI;
 import omero.model.OriginalFile;
 import omero.model.OriginalFileI;
-import omero.gateway.facility.BrowseFacility;
 import omero.gateway.facility.DataManagerFacility;
 import omero.gateway.model.ExperimenterData;
 import omero.gateway.model.ImageData;
@@ -64,9 +63,6 @@ public class HowToUseTables
     private static String password = "password";
     //end edit
 
-    /** The id of an image.*/
-    private static long imageID = 1;
-
     /** The image.*/
     private ImageData image;
 
@@ -91,18 +87,6 @@ public class HowToUseTables
         newColumns[1] = new LongColumn("MyLongColumn", "", 
                 new long[rows]);
         return newColumns;
-    }
-
-    /**
-     * Loads the image.
-     * @param imageID The id of the image to load.
-     * @return See above.
-     */
-    private ImageData loadImage(long imageID)
-            throws Exception
-    {
-        BrowseFacility browse = gateway.getFacility(BrowseFacility.class);
-        return browse.getImage(ctx, imageID);
     }
 
 // Create table
@@ -198,8 +182,9 @@ public class HowToUseTables
      * Connects and invokes the various methods.
      *
      * @param args The login credentials.
+     * @param imageId omero Image ID.
      */
-    HowToUseTables(String[] args)
+    HowToUseTables(String[] args, long imageId)
     {
         LoginCredentials cred = new LoginCredentials(args);
         gateway = new Gateway(new SimpleLogger());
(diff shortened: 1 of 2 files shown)


examples/Training/java/src/training/HowToUseTables.java
```diff
@@ -114,6 +132,23 @@ private void createTable()
...
             OriginalFile file = table.getOriginalFile(); // if you need to interact with the table
             file = new OriginalFileI(file.getId(), false);
+
+            DataManagerFacility dm = gateway.getFacility(DataManagerFacility.class);
+
+            FileAnnotation annotation = new FileAnnotationI();
+            annotation.setFile(file);
+            annotation.setNs(omero.rtypes
+                    .rstring(omero.constants.namespaces.NSBULKANNOTATIONS.value));
+            annotation = (FileAnnotation) dm.saveAndReturnObject(ctx, annotation);
```
mtbc: Out of curiosity does it still work without this line?

bramalingam: Good question, adding the link without the annotation on the server? Hmmm, should try it out locally. But was trying to be on the safe side, given that this is an example

mtbc: It also means a possibly unnecessary round-trip to the server which decreases performance on both sides if this is done in bulk: a tricky decision for training materials!

bramalingam: Tested it locally, looks like works even without this line! And it makes sense in a bulk workflow. What do you suggest @jburel, @mtbc's comment on bulk workflows definitely makes sense.! If we are doing it this way should we update our docs page in a similar way as well?

mtbc: It's probably worth an inline comment explaining why it works and is a good idea, drawing attention to the need to remain mindful of what's not yet saved.

jburel:
You need to import ``ImageData``, ``DataManagerFacility`` etc. i.e. add  ``import omero.gateway.model.ImageData;`` etc.
see https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/528/console

jburel:
The constructor needs to be reviewed so we can use the value from the configuration file.
Otherwise the example will now probably fail in the daily build with an image not found error

examples/Training/java/src/training/HowToUseTables.java
```diff
@@ -114,6 +123,25 @@ private void createTable()
...
+                    .rstring(omero.constants.namespaces.NSBULKANNOTATIONS.value));
+            
+            //The following saveAndReturnObject call can be removed in bulk annotation scenarios
+            //The server call is made when a link is created and that handles both the annotation and the link together
+            annotation = (FileAnnotation) dm.saveAndReturnObject(ctx, annotation);
+
+            //now link the image and the annotation
+            ImageAnnotationLink link = new ImageAnnotationLinkI();
+            link.setChild(annotation);
+            link.setParent(image.asImage());
```
dominikl, line 153: This will thrown a NPE, because `image` hasn't been loaded. You could for example load it in the constructor, before calling `createTableandLinkToImage()` after line 193.
```
// load the image
BrowseFacility browse = gateway.getFacility(BrowseFacility.class);
image = browse.getImage(ctx, imageId);
```

jburel: @bramalingam: you need to add again the ``loadImage`` method
what was necessary was to change the constructor so the image ID is the image ID read from the configuration file or the hard coded one if you run using the main method in the class.
i.e.
you need to add 
``
private void loadImage(long imageId) throws Exception {
        BrowseFacility browse = gateway.getFacility(BrowseFacility.class);
        ImageData image = browse.getImage(ctx, imageId);
}
``
and in the constructor invoked the method

examples/Training/java/src/training/HowToUseTables.java
```diff
@@ -154,15 +182,16 @@ private void createTable()
...
-    HowToUseTables(String[] args)
+    HowToUseTables(String[] args, long imageId)
     {
         LoginCredentials cred = new LoginCredentials(args);
         gateway = new Gateway(new SimpleLogger());
         try {
             ExperimenterData user = gateway.connect(cred);
             ctx = new SecurityContext(user.getGroupId());
-            createTable();
+            createTableandLinkToImage();
```
jburel: add ``loadImage(imageId)`` before ``createTableandLinkToImage()``

dominikl:
Thanks. Looks good, training job is passing, too. Good to merge from my point of view 👍

PR #5123, opened 2017-02-24
filesize null

will-moore:
# What this PR does

Fixes display of file annotations in right panel when file.size is None.
See https://trello.com/c/rqVzpUl6/279-bug-loading-attachments

# Testing this PR

1. Create a file-ann without setting size. Attach to object P/D/I etc.
2. Check that this displays correctly in right panel (no error as shown in trello card).

cc @bramalingam


mtbc:
The code changes certainly look sensible.

waxenegger:
That fixes the error. Tested for annotations on Project/Dataset/Image level.

image

image

PR #5126, opened 2017-02-24
finalize 5.3 database version

mtbc:
# What this PR does

SQL scripts to bump the database version to `OMERO5.3__0`.

# Testing this PR

CI in breaking and non-breaking should handle testing but do feel free to review diff and try out the upgrade manually as described in http://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/server-upgrade.html#upgrade-steps.


dominikl:
Tested on osx, upgrading from 5.2.7, no problems 👍 
Interesting to see the two effects (when switching the client from 5.2.7 to 5.3.0): ROI color changes, due to ARGB->RGBA change, and the apparent 'disappearance' of well annotations, because in 5.2.7 the image and not the well was annotated. Are we going to provide update scripts for that, too?

mtbc:
The annotation disappearance is rather disappointing: it's not as if plates' images are now an inappropriate annotation point. (Also consider dataset-to-plate.) We can provide scripts to move them but I don't know anything about that effort. There is an interesting risk of information loss given that *which* of a well's samples was annotated may itself be useful information: it's conceivable that some metadata can be sample-specific.

Yes, color encoding scripts are attached and tested on https://trello.com/c/sGlYbbvB/213-color-in-shape-settings, will need some tidying and documentation for the release though. They will be separate from the main upgrade scripts as their applicability will vary among institutions.

mtbc:
Any vetoes to removing from breaking...?

jburel:
No

jburel:
@dominikl: we discussed providing a script. Nobody yet assigned to yet, something to plan this afternoon since it is a missing piece of the puzzle

jburel:
@will-moore will take care of the "well" annotation script
Thanks all
merging

PR #5081, opened 2017-02-02
fix getThumbnailsByLongestSideSet

atarkowska:
partial rebase of #5062

------------

# What this PR does

fixing getThumbnailsByLongestSideSet in readonly group

https://trello.com/c/3eEBEPG4/7-retrieving-multiple-thumbnails

# Testing this PR

- test thumbnail version like in https://github.com/openmicroscopy/openmicroscopy/pull/3713#issuecomment-95146309

--thumbnails

PR #5062, opened 2017-01-30
Get thumbnails

atarkowska:
# What this PR does

- fix getThumbnailsByLongestSideSet in readonly group https://trello.com/c/3eEBEPG4/7-retrieving-multiple-thumbnails
- allow loading thumbnails in a batches

# Testing this PR

- test thumbnail version like in https://github.com/openmicroscopy/openmicroscopy/pull/3713#issuecomment-95146309
- load plate, all thumbnail should be loaded in a batches
- log in and test if updating rendeting update thumbnail:
  - test centre panel, search and plate grid
- consider any usecases I could miss! :)

# TODO:

- [x]  review pixel vs image in a args (looks like gateway API is not consistent)
- [x] **performance testing** That should happen on real production deployment (with and without nginx micro cache turned on)

cc @pwalczysko 

----
--depends-on https://github.com/openmicroscopy/openmicroscopy/pull/5173

--rebased-to https://github.com/openmicroscopy/openmicroscopy/pull/5081 ~~https://github.com/openmicroscopy/openmicroscopy/pull/5112~~ ~~https://github.com/openmicroscopy/openmicroscopy/pull/5138~~ #5158

PR #5138, opened 2017-02-28
Get thumbnails backend only

atarkowska:
manual and partial rebase of https://github.com/openmicroscopy/openmicroscopy/pull/5062

--------------

# What this PR does

- allow loading thumbnails in a batches

# Testing this PR

- check tests are passing

@jburel @waxenegger @will-moore 

--depends-on https://github.com/openmicroscopy/openmicroscopy/pull/5081
--thumbnails

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4303,25 +4303,21 @@ class _BlitzGateway (object):
             tb = self.createThumbnailStore()
             p = omero.sys.ParametersI().addIds(image_ids)
             sql = """select new map(
-                        i.id as im_id, r.pixels.id as pix_id, r.id as rdef_id
+                        i.id as im_id, r.pixels.id as pix_id
                      )
                      from RenderingDef as r
                          join r.pixels.image as i
                      where i.id in (:ids) """
 
-            rdefs_ids = self.getQueryService().projection(
+            img_pixel_ids = self.getQueryService().projection(
                 sql, p, self.SERVICE_OPTS)
-            pixrdef = dict()
             _temp = dict()
-            for e in rdefs_ids:
+            for e in img_pixel_ids:
                 e = unwrap(e)
                 _temp[e[0]['pix_id']] = e[0]['im_id']
-                pixrdef[e[0]['pix_id']] = e[0]['rdef_id']
 
-            # thumb = tb.getThumbnailByLongestSideSet(
-            #    rint(max_size), image_ids)
-            thumbs_map = tb.getThumbnailByLongestSideSetAndRdef(
-                size=rint(max_size), pixelsRdefMap=pixrdef)
+            thumbs_map = tb.getThumbnailByLongestSideSet(
+                rint(max_size), list(_temp))
             for (pix, thumb) in thumbs_map.items():
                 _resp[_temp[pix]] = thumb
         except Exception:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4303,10 +4303,9 @@ class _BlitzGateway (object):
             tb = self.createThumbnailStore()
             p = omero.sys.ParametersI().addIds(image_ids)
             sql = """select new map(
-                        i.id as im_id, r.pixels.id as pix_id
+                        i.id as im_id, p.id as pix_id
                      )
-                     from RenderingDef as r
-                         join r.pixels.image as i
+                     from Pixels as p join p.image as i
                      where i.id in (:ids) """
 
             img_pixel_ids = self.getQueryService().projection(
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4296,6 +4296,41 @@ class _BlitzGateway (object):
             search.close()
         return rv
 
+    def getThumbnailSet(self, image_ids, max_size=64):
+        tb = None
+        _resp = dict()
+        try:
+            tb = self.createThumbnailStore()
+            p = omero.sys.ParametersI().addIds(image_ids)
+            sql = """select new map(
+                        i.id as im_id, r.pixels.id as pix_id, r.id as rdef_id
+                     )
+                     from RenderingDef as r
+                         join r.pixels.image as i
+                     where i.id in (:ids) """
+
+            rdefs_ids = self.getQueryService().projection(
+                sql, p, self.SERVICE_OPTS)
+            pixrdef = dict()
+            _temp = dict()
+            for e in rdefs_ids:
+                e = unwrap(e)
+                _temp[e[0]['pix_id']] = e[0]['im_id']
+                pixrdef[e[0]['pix_id']] = e[0]['rdef_id']
+
+            # thumb = tb.getThumbnailByLongestSideSet(
+            #    rint(max_size), image_ids)
+            thumbs_map = tb.getThumbnailByLongestSideSetAndRdef(
+                size=rint(max_size), pixelsRdefMap=pixrdef)
+            for (pix, thumb) in thumbs_map.items():
+                _resp[_temp[pix]] = thumb
+        except Exception:
+            logger.error(traceback.format_exc())
+        finally:  # pragma: no cover
+            if tb is not None:
+                tb.close()
+        return _resp
+
 
 class OmeroGatewaySafeCallWrapper(object):  # pragma: no cover
     """
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
@@ -66,6 +66,31 @@ class TestImage (object):
         pthumb = Image.open(ptfile)  # Raises if invalid
         pthumb.verify()  # Raises if invalid
 
+    def testThumbnailSet(self, author_testimg_bad, author_testimg_big):
+        # ordynary and big image (4k x 4k and up)
+        img_ids = [self.image.id, author_testimg_big.id]
+        conn = self.image._conn
+        for (img_id, thumb) in conn.getThumbnailSet(image_ids=img_ids).items():
+            assert img_id in img_ids
+            tfile = StringIO(thumb)
+            thumb = Image.open(tfile)  # Raises if invalid
+            thumb.verify()  # Raises if invalid
+            assert thumb.format == 'JPEG'
+            assert thumb.size == (64, 64)
+
+        thumb = conn.getThumbnailSet(
+            image_ids=[self.image.id], max_size=96)[self.image.id]
+        tfile = StringIO(thumb)
+        thumb = Image.open(tfile)  # Raises if invalid
+        thumb.verify()  # Raises if invalid
+        assert thumb.size == (96, 96)
+
+        badimg_id = author_testimg_bad.id  # no pixels
+        with pytest.raises(KeyError):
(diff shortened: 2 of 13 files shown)


jburel:
``./components/tools/OmeroWeb/omeroweb/webclient/views.py:103:1: F811 redefinition of unused 'warnings' from line 35``

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4296,6 +4296,36 @@ def searchProcessing():
             search.close()
         return rv
 
+    def getThumbnailSet(self, image_ids, max_size=64):
```
will-moore, line 4299: Needs docstring.
Also, we default to using size 96 for thumbs elsewhere so that would be preferable unless there's a reason to use 64?

atarkowska: default in gateway is 64 https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/gateway/__init__.py#L7781

components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
```diff
@@ -66,6 +66,31 @@ def testThumbnail(self, author_testimg_bad, author_testimg_big):
         pthumb = Image.open(ptfile)  # Raises if invalid
         pthumb.verify()  # Raises if invalid
 
+    def testThumbnailSet(self, author_testimg_bad, author_testimg_big):
+        # ordynary and big image (4k x 4k and up)
```
will-moore: ordinary

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel: This is probably a mistake and things did not get unify.
Do we have a reason to keep at 64?

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -632,6 +632,15 @@ def leave_none_unset_int(s):
          int,
          ("Number of images displayed within a dataset or 'orphaned'"
           " container to prevent from loading them all at once.")],
+    "omero.web.thumbnails_batch":
+        ["THUMBNAILS_BATCH",
+         50,
+         int,
+         ("Number of thumbnails retrieved to prevent from loading them"
+          " all at once. Make sure the size is not too big, otherwise"
+          " you may excede limit request line, see"
```
will-moore: exceed

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: as far as I understood is view responsibility to request right size https://github.com/openmicroscopy/openmicroscopy/pull/4687#issuecomment-253154356, so is everyone ok to change to 96 in a gateway now?

atarkowska: Note that people who use gateway only will be forced to rerender thumbnails. I personally wouldn't touch it. Otherwise we have to make it very very clear that they must change implementation

```
getThumbnail() -> getThumbnail(size=(64, 64))
```

jburel: we could be extremely painful at this stage. So let's leave it as it is.
We can revisit when we have better tools to re-render or do that via queues

components/tools/OmeroWeb/omeroweb/webgateway/urls.py
```diff
@@ -263,12 +263,42 @@
...
 
+
+webgateway_get_thumbnails_json = url(
+    r'^get_thumbnails/(?:(?P<w>[0-9]+)/)?$',
+    'webgateway.views.get_thumbnails_json')
+"""
+Returns a set of thumbnail base64 encoded of the OMERO Images,
+optionally scaled to max-width and max-height.
+Params in get_thumbnails/<w>/ are:
+    - iid:  Image ID
```
will-moore: If I'm reading the code right, you request image ids using query string ?id=1&id=2 etc?
This could be explained a bit clearer here.
Also you mention max-height but only the option to set ```max width```.
Presumably ```max width``` is really ```longest side```?

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -1463,6 +1474,60 @@ def get_thumb_url(iid):
...
+    @param iid:         Image ID
+    @param w:           Thumbnail max width. 96 by default
+    @param h:           Thumbnail max height
+    @return:            http response containing base64 encoded thumbnail
+    """
+    jpeg_data = _render_thumbnail(
+        request=request, iid=iid, w=w, h=h,
+        conn=conn, _defcb=_defcb, **kwargs)
+    rv = "data:image/jpeg;base64,%s" % base64.b64encode(jpeg_data)
+    return rv
```
will-moore, line 1526: Since we now use Django's ```JsonResponse``` in ```@jsonp``` this normally expects a dictionary. We have a workaround for old methods that sets the ```safe=False``` for non-dicts (which is why this is working) but it would be nice not to introduce any more like this (to avoid API change if/when we remove the workaround).

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: webgateway was switched to use size 96 by default https://github.com/openmicroscopy/openmicroscopy/pull/4687 which is also what this PR uses.
So I guess this PR is consistent with existing behaviour. The only question is *if* we're moving Blitz Gateway to use size 96 by default, do we want to use that here? If thumbnails of size 96 are cached, then using size 64 will be slower. cc @chris-allan

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -1463,6 +1474,60 @@ def get_thumb_url(iid):
...
+    @param w:           Thumbnail max width. 96 by default
+    @return:            http response containing base64 encoded thumbnails
+    """
+    if w is None:
+        w = 96
+    image_ids = get_longs(request, 'id')
+    logger.debug("Image ids: %r" % image_ids)
+    if len(image_ids) > settings.THUMBNAILS_BATCH:
+        return HttpJavascriptResponseServerError(
+            'Max %s thumbnails at a time.' % settings.THUMBNAILS_BATCH)
```
will-moore, line 1492: Is the primary reason for the THUMBNAILS_BATCH setting to limit the size of the request string?
If so, then by the time we've got to this point, we know that the user has not exceeded this limit.
So is there any reason to reject the request at this stage?
Maybe the behaviour should be more like PAGE_SIZE, which is a default for the browser code to use, but the python code will still support larger page sizes if requested.

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: so to me is bug fixing vs breaking peoples repository, both are very important :) `getThumbnail() -> getThumbnail(size=(64, 64))` is fair but we have to be explicit and make it clear in release notes, docs, etc

chris-allan: I would not conflate BlitzGateway defaults with OMERO.web defaults. While they are obviously quite related, and 5.3.0 is a breaking release, changing the BlitzGateway default behaviour has potentially far reaching consequences. We are unlikely to be able to fully explore them this late in the 5.3.0 development cycle.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
atarkowska: > Is the primary reason for the THUMBNAILS_BATCH setting to limit the size of the request string?

Obviously no is not. THUMBNAILS_BATCH has to be first consumed by the clients to let them use batch size. The error has to be thrown here if client exceeded the batch size.

see https://github.com/openmicroscopy/openmicroscopy/pull/5062/files#diff-e226bce909b5b71ca51f5a26c7db1823R481

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: so how about if we deprecate `getThumbnail(self, size=(64, 64)...)` in fav of `getThumbnail2(self, size=(96, 96)...)` and add warning about size change and explain it there ?

components/tools/OmeroWeb/omeroweb/webgateway/views.py
will-moore: The docs say "Make sure the size is not too big, otherwise you may excede limit request line" which would suggest that it is the main reason for the limit. If other reasons are more important then maybe the docs should be updated?
Why do we have to throw an error here? If a client asks for a page size larger than the PAGE setting, we don't throw an error.

atarkowska: UPDATED: Well the doc says regarding full set of changes from https://github.com/openmicroscopy/openmicroscopy/pull/5062 including client changes. `you may exceed limit request line` is about the length of incoming request, and has nothing to do with that view.
Second thing, we also don't want to let client to freely retrieve as many thumbnails as they like. This is exactly pagination. If we are not happy of that rebase I can close the PR

components/tools/OmeroWeb/omeroweb/webgateway/views.py
atarkowska: All agreed, but I am afraid we have to accept that method as an exception. There are serious issues with compatibility between two branches and I am not happy of more breaking changes while backporting. All can be cleaned up afterwards.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#605](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/605/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/605/consoleText) for more details.
Possible conflicts:
  - PR #5081 aleksandra-tarkowska 'fix getThumbnailsByLongestSideSet'
    - components/tools/OmeroWeb/test/integration/test_tree.py

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#606](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/606/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/606/consoleText) for more details.
Possible conflicts:
  - PR #5081 aleksandra-tarkowska 'fix getThumbnailsByLongestSideSet'
    - components/tools/OmeroWeb/test/integration/test_tree.py

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#607](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/607/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/607/consoleText) for more details.
Possible conflicts:
  - PR #5081 aleksandra-tarkowska 'fix getThumbnailsByLongestSideSet'
    - components/tools/OmeroWeb/test/integration/test_tree.py

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#608](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/608/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/608/consoleText) for more details.
Possible conflicts:
  - PR #5081 aleksandra-tarkowska 'fix getThumbnailsByLongestSideSet'
    - components/tools/OmeroWeb/test/integration/test_tree.py

PR #5139, opened 2017-02-28
travis test failure

jburel:
# What this PR does

Attempt to fix the test timeout failure encountered when building using travis
Increase timeout value

# Testing this PR
Check that the test is green


mtbc:
Might be useful for people to --depends-on before merge. :smiley:

jburel:
First need to check if travis passes :smiley:

jburel:
@mtbc: I have restarted travis to check if we were lucky the first time

mtbc:
Passed again! Maybe it's indeed working.

PR #5143, opened 2017-03-03
prevent container optimistic lock exception on import

mtbc:
# What this PR does

Prevents `OptimisticLockException` when image's container is edited mid-import.

# Testing this PR

Check https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ImportLibraryTest/testImportTargetEditedDuringImport/ and also that the test properly checks that the import occurred as it should have.

One may also want to run the test against a server without this PR to ensure that it fails without the fix.

# Related reading

https://trello.com/c/Ikly5Qpc/113-bug-screen-optimistic-lock-exception-tara

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedImportRequestI.java
@@ -665,6 +665,9 @@ public class ManagedImportRequestI extends ImportRequest implements IRequest {
 
         for (final Image image : imageList) {
             image.unloadAnnotationLinks();
+            image.unloadDatasetLinks();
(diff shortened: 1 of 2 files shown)


dominikl:
I don't really understand the original problem, i.e. why the `OptimisticLockException` occured in first place, but I can confirm that 2ac478e fixes the problem. Ran the integration test 85f39d9 without the bugfix -> `OptimisticLockException`, ran it with the bugfix -> passed (and image is in dataset, like expected); also looks reasonable, the dataset description is modified while the import is in progress. Good to merge 👍

Issue #5102, opened 2017-02-14
Possible bug in code (defect) (CWE-398)


AppChecker:
Hello!

We've checked your sources with [AppChecker static analyzer](https://npo-echelon.ru/en/solutions/appchecker.php) and found suspicious place:

https://github.com/openmicroscopy/openmicroscopy/blob/c4d3e2eca743198739746747eacf7713875c913c/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/actions/CreateTopContainerAction.java#L332

```
                        if (ns != null &&
                                TagAnnotationData.INSIGHT_TAGSET_NS.equals(
                                        ns));
                        withParent = model.canLink(tag);
```

If-statement contains empty block (;), i.e.  (ns != null && TagAnnotationData.INSIGHT_TAGSET_NS.equals(ns)) doesn't effect anything

Possible defect was found by Echelon Team with [AppChecker static analyzer](https://npo-echelon.ru/en/solutions/appchecker.php)

PR #5145, opened 2017-03-04
Fix If-statement containing empty block

joshmoore:
Fixes gh-5102 by removing the `;` character which makes the if statement
pointless. Unclear what the impact of this is on users was previously.


dominikl:
That was indeed a bug, `withParent` flag could have been wrongly set, although this didn't have any further consequences.  This PR fixes the issue, good to merge 👍

PR #5149, opened 2017-03-05
highlight selected menu

atarkowska:
# What this PR does

Temporary fix for top links menu to make sure selected section is highlighted.

firefoxscreensnapz047

# Testing this PR

1. configure top_links and check if the right one is selected while swithcing between them 
2. Limitations, it doesn't work with existing webadmin link so not suitable to be rebased

# Related reading

https://trello.com/c/s0HD7Vi6/173-mapr-i-e-tagexport-py#comment-5764396d81e872aee908a977

> There's no indicator (other than the URL) for which of the sections you're currently looking at.

--no-rebase

PR #5150, opened 2017-03-05
highlight selected menu (rebased onto develop)

atarkowska:
This is the same as gh-5149 but rebased onto develop.

----

# What this PR does

Fix top links menu to make sure selected section is highlighted 

firefoxscreensnapz047

# Testing this PR

1. configure top_links and check if the right one is selected while swithcing between them 

# Related reading

https://trello.com/c/s0HD7Vi6/173-mapr-i-e-tagexport-py#comment-5764396d81e872aee908a977

> There's no indicator (other than the URL) for which of the sections you're currently looking at.


atarkowska:
--rebased-from #5149

PR #5155, opened 2017-03-06
paths_to_object pagesize

will-moore:
# What this PR does

Adds ```?page_size=10``` support to ```/paths_to_object/```.
See https://github.com/ome/omero-iviewer/pull/43/

NB: This ```page_size``` parameter is already testing in ```OmeroWeb/test/integration/test_show.py``` in ```test_project_dataset_images_pagination()```. This PR simply allows the parameter to be set via the request query string.

# Testing this PR

 - Pick an Image in a Dataset with more than 10 images
 - Use the image ID in a URL: ```/webclient/api/paths_to_object/?image=56797&page_size=10```
 - Should see that the path to the Image includes Dataset with pagination info
```
{
  "childIndex": 53,
  "type": "dataset",
  "id": 17752,
  "childPage": 6,
  "childCount": 61
},
```

 # Further reading

The iViewer needs to show 10 thumbnails from the Dataset that includes the current Image, so it needs to know which "page" of the Dataset the current image comes from.
See https://github.com/ome/omero-iviewer/pull/43/
This functionality already exists in paths_to_objects, so that the webclient can open on the correct page of a Dataset to show the specified Image. But previously the page_size defaulted to the currently configured default.

cc @waxenegger

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/show.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/show.py
@@ -573,13 +573,13 @@ def paths_to_object(conn, experimenter_id=None, project_id=None,
                 ds = {
                     'type': 'dataset',
                     'id': datasetId,
+                    'childCount': imgCount,
                 }
                 if imgCount > page_size:
                     # Need to know which page image is on
                     iids = get_image_ids(conn, datasetId)
                     index = iids.index(imageId)
                     page = (index / page_size) + 1  # 1-based index
-                    ds['childCount'] = imgCount
                     ds['childIndex'] = index
                     ds['childPage'] = page
                 path.append(ds)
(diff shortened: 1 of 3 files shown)


waxenegger:
Works fine.

As discussed, it would be great if we could add the image count to the response regardless of the conditional image index meta-info (which gets only returned if paging is necessary - page_size smaller than image count).

will-moore:
@waxenegger Now you should always get the 'childCount', even when no pagination needed.

waxenegger:
Works beautifully.

atarkowska:
This PR broke CI this morning https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/477/testReport/.

will-moore:
Sorry - didn't get around to fixing it yet. Will fix now...

jburel:
PR opened to fix tests
https://github.com/openmicroscopy/openmicroscopy/pull/5168

waxenegger:
I know I wanted to check the integration test ...  but obviously I didn't.

PR #3713, opened 2015-04-14
Set pyramid thumb versions to -1 (Fix #12538)

joshmoore:
Thumbnails that are generated while pyramids are being
created (i.e. the clock symbol) should have a version of
-1 so that clients know to refresh.

There is currently no case where `version == null` is
used or even possible.

cc: @will-moore

PR #5158, opened 2017-03-06
Get thumbnails (rebased onto develop)

atarkowska:
manual rebase of https://github.com/openmicroscopy/openmicroscopy/pull/5062

--------------

# What this PR does

- allow loading thumbnails in a batches

# Testing this PR

- test thumbnail version like in https://github.com/openmicroscopy/openmicroscopy/pull/3713#issuecomment-95146309
- load plate, all thumbnail should be loaded in a batches
- log in and test if updating rendeting update thumbnail:
  - test centre panel, search and plate grid
- consider any usecases I could miss! :)

--depends-on #5207

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4423,25 +4423,21 @@ class _BlitzGateway (object):
             tb = self.createThumbnailStore()
             p = omero.sys.ParametersI().addIds(image_ids)
             sql = """select new map(
-                        i.id as im_id, r.pixels.id as pix_id, r.id as rdef_id
+                        i.id as im_id, r.pixels.id as pix_id
                      )
                      from RenderingDef as r
                          join r.pixels.image as i
                      where i.id in (:ids) """
 
-            rdefs_ids = self.getQueryService().projection(
+            img_pixel_ids = self.getQueryService().projection(
                 sql, p, self.SERVICE_OPTS)
-            pixrdef = dict()
             _temp = dict()
-            for e in rdefs_ids:
+            for e in img_pixel_ids:
                 e = unwrap(e)
                 _temp[e[0]['pix_id']] = e[0]['im_id']
-                pixrdef[e[0]['pix_id']] = e[0]['rdef_id']
 
-            # thumb = tb.getThumbnailByLongestSideSet(
-            #    rint(max_size), image_ids)
-            thumbs_map = tb.getThumbnailByLongestSideSetAndRdef(
-                size=rint(max_size), pixelsRdefMap=pixrdef)
+            thumbs_map = tb.getThumbnailByLongestSideSet(
+                rint(max_size), list(_temp))
             for (pix, thumb) in thumbs_map.items():
                 _resp[_temp[pix]] = thumb
         except Exception:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4423,10 +4423,9 @@ class _BlitzGateway (object):
             tb = self.createThumbnailStore()
             p = omero.sys.ParametersI().addIds(image_ids)
             sql = """select new map(
-                        i.id as im_id, r.pixels.id as pix_id
+                        i.id as im_id, p.id as pix_id
                      )
-                     from RenderingDef as r
-                         join r.pixels.image as i
+                     from Pixels as p join p.image as i
                      where i.id in (:ids) """
 
             img_pixel_ids = self.getQueryService().projection(
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4434,6 +4434,9 @@ class _BlitzGateway (object):
         tb = None
         _resp = dict()
         try:
+            ctx = self.SERVICE_OPTS.copy()
+            if ctx.getOmeroGroup() is None:
+                ctx.setOmeroGroup(-1)
             tb = self.createThumbnailStore()
             p = omero.sys.ParametersI().addIds(image_ids)
             sql = """select new map(
@@ -4443,14 +4446,14 @@ class _BlitzGateway (object):
                      where i.id in (:ids) """
 
             img_pixel_ids = self.getQueryService().projection(
-                sql, p, self.SERVICE_OPTS)
+                sql, p, ctx)
             _temp = dict()
             for e in img_pixel_ids:
                 e = unwrap(e)
                 _temp[e[0]['pix_id']] = e[0]['im_id']
 
             thumbs_map = tb.getThumbnailByLongestSideSet(
-                rint(max_size), list(_temp))
+                rint(max_size), list(_temp), ctx)
             for (pix, thumb) in thumbs_map.items():
                 _resp[_temp[pix]] = thumb
         except Exception:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4417,6 +4417,20 @@ class _BlitzGateway (object):
         return rv
 
     def getThumbnailSet(self, image_ids, max_size=64):
+        """
+        Retrieves a number of thumbnails for image sets. If the Thumbnails
+        exist in the on-disk cache they will be returned directly,
+        otherwise they will be created, for more details
+        see ome.api.ThumbnailStore.getThumbnailByLongestSideSet
+
+        :param image_ids:   A list of image ids
+        :param max_size:    The longest side of the image will be used
+                            to calculate the size for the smaller side
+                            in order to keep the aspect ratio of
+                            the original image.
+        :return:            dictionary of strings holding a rendered JPEG
+                            of the thumbnails.
+        """
         tb = None
         _resp = dict()
         try:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4416,6 +4416,41 @@ class _BlitzGateway (object):
             search.close()
         return rv
 
+    def getThumbnailSet(self, image_ids, max_size=64):
+        tb = None
+        _resp = dict()
+        try:
+            tb = self.createThumbnailStore()
+            p = omero.sys.ParametersI().addIds(image_ids)
+            sql = """select new map(
+                        i.id as im_id, r.pixels.id as pix_id, r.id as rdef_id
+                     )
+                     from RenderingDef as r
+                         join r.pixels.image as i
+                     where i.id in (:ids) """
+
+            rdefs_ids = self.getQueryService().projection(
+                sql, p, self.SERVICE_OPTS)
+            pixrdef = dict()
+            _temp = dict()
+            for e in rdefs_ids:
+                e = unwrap(e)
+                _temp[e[0]['pix_id']] = e[0]['im_id']
+                pixrdef[e[0]['pix_id']] = e[0]['rdef_id']
+
+            # thumb = tb.getThumbnailByLongestSideSet(
+            #    rint(max_size), image_ids)
+            thumbs_map = tb.getThumbnailByLongestSideSetAndRdef(
+                size=rint(max_size), pixelsRdefMap=pixrdef)
+            for (pix, thumb) in thumbs_map.items():
+                _resp[_temp[pix]] = thumb
+        except Exception:
+            logger.error(traceback.format_exc())
+        finally:  # pragma: no cover
+            if tb is not None:
+                tb.close()
+        return _resp
+
 
 class OmeroGatewaySafeCallWrapper(object):  # pragma: no cover
     """
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
@@ -474,10 +474,10 @@ $(function() {
                 // Extra data needed for showing thumbs in centre panel
                 if (node.type === 'dataset' || node.type === 'orphaned' || node.type === 'tag') {
                     payload['sizeXYZ'] = true;
-                    payload['date'] = true;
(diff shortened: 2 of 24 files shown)


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -4416,6 +4416,36 @@ def searchProcessing():
             search.close()
         return rv
 
+    def getThumbnailSet(self, image_ids, max_size=64):
```
will-moore, line 4419: Needs docstring

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
```diff
@@ -474,10 +474,10 @@ $(function() {
...
                 if (node.type === 'dataset' || node.type === 'orphaned' || node.type === 'tag') {
                     payload['sizeXYZ'] = true;
-                    payload['date'] = true;
-                    if (node.type !== 'tag') {
-                        payload['thumbVersion'] = true;
-                    }
+                    //payload['date'] = true;
+                    //if (node.type !== 'tag') {
+                    //    payload['thumbVersion'] = true;
+                    //}
```
will-moore: We need to keep ```payload['date'] = true``` (used in thumbnail list view in center panel).
Rest can be removed.

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
```diff
@@ -382,9 +381,13 @@ OME.refreshThumbnails = function(options) {
...
-            var $this = $(this),
-                base_src = $this.attr('src').split('?')[0];
-            $this.attr('src', base_src + "?_="+rdm);
+            var _t = $(this);
+                OME.load_thumbnail(
+                    options.imageId, options.thumbnail_url,
+                    function(thumb) {
+                        _t.attr('src', thumb);
+                    }
+                );
```
will-moore: As discussed, these thumbnails can be refreshed in a batch instead of 1 call per thumbnail.

components/tools/OmeroPy/src/omero/gateway/__init__.py
atarkowska: forgot to cherry-pick from the other pr

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -1463,6 +1474,60 @@ def get_thumb_url(iid):
...
+    @param iid:         Image ID
+    @param w:           Thumbnail max width. 96 by default
+    @param h:           Thumbnail max height
+    @return:            http response containing base64 encoded thumbnail
+    """
+    jpeg_data = _render_thumbnail(
+        request=request, iid=iid, w=w, h=h,
+        conn=conn, _defcb=_defcb, **kwargs)
+    rv = "data:image/jpeg;base64,%s" % base64.b64encode(jpeg_data)
+    return rv
```
will-moore, line 1531: Continuing the discussion from https://github.com/openmicroscopy/openmicroscopy/pull/5138
You could simply do ``` return {iid: rv}``` (which would be the same format as ```get_thumbnails_json()```) or
```return {'data': rv}```, which is what we do in the /api/ urls. Either is easy to handle in the JavaScript code.

File: components/tools/OmeroWeb/omeroweb/webgateway/views.py [hunk 5]
```diff
@@ -306,9 +308,8 @@ def render_birds_eye_view(request, iid, size=None,
     return render_thumbnail(request, iid, w=size, **kwargs)
 
 
-@login_required()
-def render_thumbnail(request, iid, w=None, h=None, conn=None, _defcb=None,
-                     **kwargs):
+def _render_thumbnail(request, iid, w=None, h=None, conn=None, _defcb=None,
+                      **kwargs):
```
will-moore, line 312: Docstring needs updating ```Returns an HttpResponse...``` no longer true.

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
will-moore: Instead of being passed the full ```thumbnail_url```, ```refreshThumbnails()``` could be passed the ```/webgateway/``` base url and then it could construct the ```/webgateway/get_thumbnails/``` OR ```/webgateway/get_thumbnail/id/``` urls itself, depending on whether it was refreshing ALL thumbs or just a single image.
Alternatively, we could add the ```/webgateway/``` url to the global WEBCLIENT.URLS dictionary, if we know that this is always going to be available, similar to WEBCLIENT.URLS.webindex which we use in ```ome.tree.js```

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.plateview.js
```diff
@@ -160,6 +164,33 @@ jQuery._WeblitzPlateview = function (container, options) {
...
+              if (value === null) {
+                value = opts.defaultThumb;
+              }
+              $("img#"+parentPrefix+"image-"+key).attr("src", value);
+            });
+          }
+          gs_json(thumbnails_url, null, _load_thumbnails);
+          input = input.filter(function(x) {
+              return iids.indexOf(x) < 0;
+          });
```
will-moore: Any reason to filter instead of simply ```input = input.slice(batch, batch * 2)```?

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.plateview.js
```diff
@@ -132,13 +134,15 @@ jQuery._WeblitzPlateview = function (container, options) {
     // NB: don't add other classes here - will get removed on slider change.
     table.addClass('showWellLabel wellSize' + opts.width);
 
+    var imgIds = new Array();
```
will-moore: Much prefer the array literal ``` var imgIds = [];```
E.g. https://www.w3schools.com/js/js_arrays.asp
"There is no need to use new Array(). For simplicity, readability and execution speed, use the array literal method."
Or the airbnb js style guide: https://github.com/airbnb/javascript
```
// bad
const items = new Array();
// good
const items = [];
```

components/tools/OmeroWeb/omeroweb/webgateway/util.py
```diff
@@ -34,6 +34,15 @@ def getIntOrDefault(request, name, default):
     return index
 
 
+def get_longs(request, name):
+
```
will-moore: Move the deleted webclient get_longs docstring here:
"""
Retrieves parameters from the request. If the parameters are not present
an empty list is returned
This does not catch exceptions as it makes sense to throw exceptions if
the arguments provided do not pass basic type validation
"""

will-moore:
I don't see thumbnails working in Shares. The url that is loaded by that page ```/webclient/get_thumbnails/``` doesn't include the share_id (gets empty ```{}```) and if I add it and go to URL ```/webclient/get_thumbnails/238420/?id=56699&id=56700``` now I get nulls  ``` {"56699": null, "56700": null}```

components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
atarkowska: I am not sure we have to complicate, get_thumbnails can be easily called with one or more image ids

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.plateview.js
atarkowska: not really, that is not the same, but I could `input = input.slice(batch, input.length);`

jburel:
see https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/

will-moore:
Sorry - didn't add comment here earlier.... But found intermittent thumbnail failures which @aleksandra-tarkowska has been investigating further.

atarkowska:
last commit drop group -1, should resolve thumbnails loading issue, leaving to server experts to decide 

cc @jburel @mtbc

atarkowska:
test is green https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest.test_image/

components/tools/OmeroWeb/omeroweb/webgateway/views.py
atarkowska: It meant to be inline with render_thumbnail to let people easily switch without much of the code changes. This will have an impact on `refreshThumbnails`, my idea was to reduce number of changes.

jburel:
context: User-3/read-only web-merge
 * Browse a plate
 * Click on a thumbnail in the centre panel
 * The thumbnail is no longer visible
 * The "grid view" thumbnails no longer visible too 

plate_browsing

atarkowska:
Plate loads ok, this happen when you click on well and images on the bottom are populated. @will-moore would you like to review your changes please?

Note that clicking on image updates well in a grid, that is not expected I think
The mentioned issue happend only with this particular plate. other plates are fine

will-moore:
For me, user-3, loading thumbnails for the image IDs in bottom panel,
http://web-dev-merge.openmicroscopy.org/webclient/get_thumbnails/?id=40301&id=40354&id=40355&id=40356&_=1489576785488
I get all nulls.
```
{
"40301": null,
"40354": null,
"40355": null,
"40356": null
}
```

jburel:
The plate loads ok yes
error only occurs when I click on a thumbnail
I noticed that the "spatial view" in the tree does generate a grey image 
I don't have the problem with plate ID ``	2081``

will-moore:
Strangely, when I only select the first image, I get a thumbnail OK at 
http://web-dev-merge.openmicroscopy.org/webclient/get_thumbnails/?id=40301
but all of the other IDs return null, and when combined with that first image, then we get ALL nulls as above.

atarkowska:
@will-moore did you watched web/Blitz logs? It may be that omero.group -1 is pass some how

will-moore:
No I didn't look at the logs. It seems that the images in the first field for that plate are OK, but other fields are not working. E.g. switch to field 2:
It would be strange to get a different group context for different images in the same Well.

screen shot 2017-03-15 at 12 03 17

will-moore:
I am also seeing that sometimes on Save of rendering settings, the thumbnail returned from /webclient/get_thumbnails/?id=54763&id=54763 initially are null (see screenshot) but on reload of the same /get_thumbnails/ url they are now OK.

screen shot 2017-03-15 at 12 16 19

atarkowska:
unsettling group should resolve the problem, it should now guarantee that stateful service ctx will be updated, regardless of the previous context. Would be nice to review that deeply

@will-moore thanks for looking

atarkowska:
@will-moore @jburel  excluding https://github.com/openmicroscopy/openmicroscopy/pull/5172 resolved our favourite orange plate (at least I clicked on most of wells and it works, also updating settings works). although sometime I was getting No Image in (old style `render_thumbnail` url)

firefoxscreensnapz049

Not sure how to explain? There any inconsistency in how thumbnail is restored!

will-moore:
I think, although https://github.com/openmicroscopy/openmicroscopy/pull/5158/commits/5ea88b5246208129c57dbea31dee9bd03c59f233 fixed some issues we were seeing, it has again broken the cross-group thumbnails (getting thumbnails from a group that is not my default group). E.g. user-3 http://web-dev-merge.openmicroscopy.org/webclient/?show=plate-454
So it looks like this is going to need some more work and is not going to be in 5.3.0.
cc @jburel, @pwalczysko

jburel:
excluding for now since it won't be included in 5.3.0

Edit: removed exclude so that this can target metadata53 (Josh, 21/Mar)

atarkowska:
This is going now against m53

atarkowska:
@jburel @sbesson any thoughts about travis?

```
Exception: Unsupported OMERO version: 5.3.1-ice35-SNAPSHOT
```

sbesson:
@aleksandra-tarkowska yes, this is due to an omero-marshal bug which was fixed in https://github.com/openmicroscopy/omero-marshal/pull/38. Once https://github.com/openmicroscopy/openmicroscopy/pull/5209 is merged, the Travis builds should pass again.

atarkowska:
@will-moore ec4e48b is removed now

atarkowska:
--rebased-to #5233

pwalczysko:
Bug:  the thumbs in plates in both default as well as non-default groups are not being returned after Save to All:
  
  - navigate to http://web-dev-merge.openmicroscopy.org/webclient/?show=image-21276, login as user-4 (this is a plate in read-write group which is not default group of user-4)
  - select a field, open Preview
  - check "Greyscale"
  - click "Save to All"
  - observe
screen shot 2017-04-13 at 10 43 49

  - repeat the same with http://web-dev-merge.openmicroscopy.org/webclient/?show=image-21107, which is in default group, the results are as above

atarkowska:
that is what @will-moore notices before. https://github.com/openmicroscopy/openmicroscopy/pull/5235 meant to show that failure. @mtbc any thoughts how to fix it? 

My impression is that we face concurrency issues where thumbnails are requested during saving. When you refresh plate it will load them. We could have a look to make JS call synchronous maybe

mtbc:
No immediate thoughts I'm afraid. Race condition is an interesting idea -- might explain why it's been hard for me to pin down issues at my end -- I guess somehow web client must be doing *something* differently from the test. Certainly open to any tweaks to the integration test that might help trigger failure that the server should handle better.

sbesson:
Marked as `exclude` alongside https://github.com/openmicroscopy/openmicroscopy/pull/5207#issuecomment-293919623 to test the shares fix.

pwalczysko:
Tested together with Mark's PR today again. See https://github.com/openmicroscopy/openmicroscopy/pull/5207#issuecomment-295642459 and https://github.com/openmicroscopy/openmicroscopy/pull/5207#issuecomment-295641248

jburel:
replaced by https://github.com/openmicroscopy/openmicroscopy/pull/5192

PR #5125, opened 2017-02-24
java gateway usage for creating ROIs and add transforms example

bramalingam:
# What this PR does

Update ROIs .java to use the java gateway for creating ROIs and also add example code snippet for handling transforms
@jburel 

# Testing this PR

Check if the integration tests are passing.

PR #5162, opened 2017-03-08
training: close resources

jburel:
# What this PR does

Make sure resources are closed

# Testing this PR

check that the examples are green

# Related reading
https://github.com/openmicroscopy/openmicroscopy/pull/5125

cc @bramalingam

### DIFF
--- a/examples/Training/java/src/training/RawDataAccess.java
+++ b/examples/Training/java/src/training/RawDataAccess.java
@@ -141,7 +141,9 @@ public class RawDataAccess
             }
         } catch (Exception e) {
             throw new Exception("Cannot read the tiles", e);
-        } 
+        } finally {
+            rdf.close();
+        }
     }
 
 // Retrieve stack
--- a/examples/Training/java/src/training/RawDataAccess.java
+++ b/examples/Training/java/src/training/RawDataAccess.java
@@ -141,8 +141,6 @@ public class RawDataAccess
             }
         } catch (Exception e) {
             throw new Exception("Cannot read the tiles", e);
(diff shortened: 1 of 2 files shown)


examples/Training/java/src/training/RawDataAccess.java
```diff
@@ -141,7 +141,9 @@ private void retrieveTile()
             }
         } catch (Exception e) {
             throw new Exception("Cannot read the tiles", e);
-        } 
+        } finally {
+            rdf.close();
```
joshmoore: Does this suggest eventually making the `Facility` interfaces subclass `Closeable`?

mtbc: Relevant to https://trello.com/c/K4d0rFQc/712-try-with-resources-for-blitz-services perhaps.

jburel: @joshmoore @mtbc certainly something to look into
so we can simplify good portion of code

dominikl:
There's actually no `close` method in the `Facility` classes, the services are closed on log out, when the `Connector`s are closed: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/Connector.java#L857

jburel:
That's something we may want to add to close services earlier e.g. rawfilestore
so we could then take advantages of ``try with resources``

jburel:
card https://trello.com/c/Uavp5lOO/30-gateway-facility

dominikl:
Looks good 👍

PR #5163, opened 2017-03-08
Load all `omero/plugins` subdirs on the PYTHONPATH (rebased onto metadata52)

joshmoore:
This is the same as gh-4619 but rebased onto metadata52.

----

During plugin loading, loop over the contents of sys.path
(i.e. PYTHONPATH) and for all directories which contain a
subdirectory `omero/plugin` run the loadpath method as if
it had been passed to `bin/omero` with `--path`.

NB: This currently does not handle plugins packaged as eggs.

See https://github.com/joshmoore/omero-cli-duplicate/pull/1 for an example of how this would be used. Currently excluding as strictly an investigation.

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1370,7 +1370,17 @@ class CLI(cmd.Cmd, Context):
         in the parser
         """
 
-        for plugin_path in self._plugin_paths:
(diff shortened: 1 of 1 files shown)


joshmoore:
--rebased-from #4619

joshmoore:
Merging for eventual testing of decoupled plugins.

PR #5164, opened 2017-03-08
Don't trigger unnecessary field metadata reload (spw)

dominikl:
# What this PR does

Bugfix: When you click on a Field (SPW) which already has been selected, the right hand side metadata panel, reloads but gets stuck in the "Loading..." state. With this PR the selection is checked beforehand before triggering a reload.

# Testing this PR

Perform above mentioned workflow.

# Related reading

https://trello.com/c/JA02JNJf/309-bug-insight-second-click-on-thumb-in-fields-causes-endless-spinner-in-preview


pwalczysko:

PR #5172, opened 2017-03-10
do not attempt to create rendering settings in all-groups context

mtbc:
# What this PR does

Allows the retrieval of thumbnails from multiple groups at once.

--rebased-to #5173 partially

# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ThumbnailStoreTest/testGetThumbnailsMultipleGroups/ should pass.

# Related reading

https://trello.com/c/bJqd3SkJ/316-getthumbnailbylongestsideset-doesn-t-accept-ctx

### DIFF
--- a/components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
+++ b/components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
@@ -245,6 +245,7 @@ public class ThumbnailStoreTest extends AbstractServerTest {
         ThumbnailStorePrx svc = factory.createThumbnailStore();
         svc.setPixelsId(pixelsIdα);
         final byte[] thumbnailα = svc.getThumbnailByLongestSide(null);
+        Assert.assertTrue(thumbnailα.length > 0);
         svc.close();
 
         /* import the image as another user in another group and get its thumbnail */
@@ -253,6 +254,7 @@ public class ThumbnailStoreTest extends AbstractServerTest {
         svc = factory.createThumbnailStore();
         svc.setPixelsId(pixelsIdβ);
         final byte[] thumbnailβ = svc.getThumbnailByLongestSide(null);
+        Assert.assertTrue(thumbnailβ.length > 0);
 
         /* have that other user use all-groups context to fetch both thumbnails at once */
         final List<Long> pixelsIdsαβ = ImmutableList.of(pixelsIdα, pixelsIdβ);
--- a/components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
+++ b/components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
@@ -237,36 +237,66 @@ public class ThumbnailStoreTest extends AbstractServerTest {
      */
     @Test
     public void testGetThumbnailsMultipleGroups() throws Throwable {
+        final byte[] thumbnail;
+        final long pixelsIdα, pixelsIdβ;
+        ThumbnailStorePrx svc = null;
+
         /* create a fake image file */
         final File file = File.createTempFile(getClass().getSimpleName(), ".fake");
+        file.deleteOnExit();
 
-        /* import the image as one user in one group and get its thumbnail */
-        final long pixelsIdα = importFile(importer, file, "fake", false).get(0).getId().getValue();
-        ThumbnailStorePrx svc = factory.createThumbnailStore();
-        svc.setPixelsId(pixelsIdα);
-        final byte[] thumbnailα = svc.getThumbnailByLongestSide(null);
-        Assert.assertTrue(thumbnailα.length > 0);
-        svc.close();
+        try {
+            /* import the image as one user in one group and get its thumbnail */
+            pixelsIdα = importFile(importer, file, "fake", false).get(0).getId().getValue();
+            svc = factory.createThumbnailStore();
+            svc.setPixelsId(pixelsIdα);
+            thumbnail = svc.getThumbnailByLongestSide(null);
+        } finally {
+            if (svc != null) {
+                {
+                    svc.close();
+                    svc = null;
+                }
+            }
+        }
 
-        /* import the image as another user in another group and get its thumbnail */
         setUpNewUserWithImporter();
-        final long pixelsIdβ = importFile(importer, file, "fake", false).get(0).getId().getValue();
-        svc = factory.createThumbnailStore();
-        svc.setPixelsId(pixelsIdβ);
-        final byte[] thumbnailβ = svc.getThumbnailByLongestSide(null);
-        Assert.assertTrue(thumbnailβ.length > 0);
 
-        /* have that other user use all-groups context to fetch both thumbnails at once */
-        final List<Long> pixelsIdsαβ = ImmutableList.of(pixelsIdα, pixelsIdβ);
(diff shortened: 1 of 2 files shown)


aleksandra-tarkowska:
--thumbnails

components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
```diff
@@ -229,4 +230,43 @@ public void testGetThumbnailAfterReset() throws Exception {
...
+     * Test that thumbnails can be retrieved from multiple groups at once.
+     * @throws Throwable unexpected
+     */
+    @Test
+    public void testGetThumbnailsMultipleGroups() throws Throwable {
+        /* create a fake image file */
+        final File file = File.createTempFile(getClass().getSimpleName(), ".fake");
+
+        /* import the image as one user in one group and get its thumbnail */
+        final long pixelsIdα = importFile(importer, file, "fake", false).get(0).getId().getValue();
```
jburel: could the characters create conflict with the other PRs?

components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
```diff
@@ -229,4 +230,43 @@ public void testGetThumbnailAfterReset() throws Exception {
...
+        ThumbnailStorePrx svc = factory.createThumbnailStore();
+        svc.setPixelsId(pixelsIdα);
+        final byte[] thumbnailα = svc.getThumbnailByLongestSide(null);
```
jburel: a ``try/finally`` will be better

components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
```diff
@@ -229,4 +230,43 @@ public void testGetThumbnailAfterReset() throws Exception {
...
+        final List<Long> pixelsIdsαβ = ImmutableList.of(pixelsIdα, pixelsIdβ);
+        final Map<String, String> allGroupsContext = ImmutableMap.of("omero.group", "-1");
+        final Map<Long, byte[]> thumbs = svc.getThumbnailByLongestSideSet(null, pixelsIdsαβ, allGroupsContext);
+
+        /* check that the thumbnails are as before */
+        Assert.assertEquals(thumbnailα, thumbs.get(pixelsIdα));
+        Assert.assertEquals(thumbnailβ, thumbs.get(pixelsIdβ));
+
+        /* clean up */
+        svc.close();
```
jburel: same

components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
```diff
@@ -229,4 +230,73 @@ public void testGetThumbnailAfterReset() throws Exception {
...
+        }
+
+        final Map<Long, byte[]> thumbnails;
+
+        try {
+            /* use all-groups context to fetch both thumbnails at once */
+            final List<Long> pixelsIdsαβ = ImmutableList.of(pixelsIdα, pixelsIdβ);
+            final Map<String, String> allGroupsContext = ImmutableMap.of("omero.group", "-1");
+            svc = factory.createThumbnailStore();
+            thumbnails = svc.getThumbnailByLongestSideSet(null, pixelsIdsαβ, allGroupsContext);
```
atarkowska, line 287: I feel this test woudl be more accurate if is written in isolation. This is what the python test does, it may help revealing the failure. Keep in mind web is stateless

mtbc: As from card,

> PR apparently allows thumbnails to be fetched in an all-groups context for only images that previously had thumbnail fetched otherwise.

the not-all-groups fetches appear to be needed for this step to work. Methods like `ThumbnailCtx.createMissingThumbnailMetadata` might be the key because calls like `updateService.saveAndReturnIds` do not work in a -1 context.

atarkowska: In read only group member won't be able to save anything, right? Could we figure out what exactly is being created then? are tests missing anything from the client workflow?

mtbc: Good question. I don't know if anything else is being created, but `Thumbnail` OMERO model objects are created by the earlier `getThumbnailByLongestSide` calls.

mtbc: The test is happening in a `rwr-r-` group.

atarkowska: but data is accessed by an owner (owner can write still), how about accessing it by a member?

jburel: Clarified the IDR workflow with @joshmoore; 
images are imported using a dedicated user. Thumbnails are generated during the import process so the thumbnail and the settings are owned by the dedicated user. Only when the generation is complete, it is possible to access the data using the ``-1`` context

mtbc: If by a member it still seems to work! I guess the permissions system must give special dispensation for thumbnails or something, cc: @joshmoore.

aleksandra-tarkowska:
--exclude

tmp

atarkowska:
--thumbnails
--exclude

atarkowska:
@mtbc looks like the strange behaviour reported https://github.com/openmicroscopy/openmicroscopy/pull/5158#issuecomment-286711324 comes from this PR. excluded now and rebuilt, seems like no issues. Do you have any thoughts?

mtbc:
I don't, I'm afraid. I wonder if there is some integration test that could trigger the behavior. There wasn't any smoking gun in the logs?

mtbc:
Best not to merge this into 5.3.0 if it causes a strange issue. Then we have more time to try to pin down what triggers it and to look out for other odd behaviors.

atarkowska:
if anyone has any idea feel free to remove  exclude label.

components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
joshmoore: Perhaps

```
/opt/omeA/components/server/src/ome/security $ git grep -n Thumbnail
basic/OmeroInterceptor.java:45:import ome.model.display.Thumbnail;
basic/OmeroInterceptor.java:580:                 Thumbnail.class.isAssignableFrom(changedClass))) {
```

mtbc:
I have no idea how this fix would have caused a problem, rather than trying to figure that out we can probably now hold off until we have a better fix.

components/tools/OmeroJava/test/integration/ThumbnailStoreTest.java
mtbc: `OmeroInterceptor.neededRight` does look like it, thank you. Maybe I was looking in just `BasicACLVoter` before. Interesting to see in that function that a `RenderingDef` can be linked to a `RenderingDef`.

mtbc:
Superseded by #5207.

PR #5175, opened 2017-03-10
TMP python training

jburel:
# What this PR does

Comment out example

# Testing this PR

This is a tmp PR
This should not be merged

# Related reading
CI job has been adjusted
see error https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/552

cc @will-moore

PR #5180, opened 2017-03-13
Training json test

will-moore:
# What this PR does

Allows ci jobs to pass url to training setup script so that the Json Api Login.py example can connect to the correct url.
Otherwise we see test failing Re: https://github.com/openmicroscopy/openmicroscopy/pull/5175

Also added a print statement so we can check that the test is using the URL that it is configured to use.

cc @jburel will look to update the Training ci job to pass in the correct URL.

# Testing this PR

1. If this PR and update to ci job are working, Training tests should be green


jburel:
We may want to add ``WEB_PROTOCOL`` too

will-moore:
Isn't the ```WEB_PROTOCOL``` included in the URL? E.g. ```http://localhost``` vv ```https://localhost```

jburel:
Not for the robot job for example
 we specify the protocol as a separate argument

jburel:
it fails since none of the server name as ``omero`` in the name
it will need to be adjusted to check name and port

jburel:
I have adjusted the job so you server name can be set.
``SERVER_NAME`` is the variable name

jburel:
Due to a bug in another PR this was not run this am, will start another build later today

jburel:
``
CSRF token 4kHM5aYUHtvSxAY4w7n2iKLRKLtFTI2B
Servers:
-id: 1
 name: eel-merge
 host: eel.openmicroscopy.org
 port: 4064
-id: 2
 name: eel-latest
 host: eel.openmicroscopy.org
 port: 14064
-id: 3
 name: eel-integration
 host: eel.openmicroscopy.org
 port: 24064
-id: 4
 name: seabass
 host: seabass.openmicroscopy.org
 port: 24064
``
all good now

PR #5182, opened 2017-03-14
color scripts

jburel:
# What this PR does

Add scripts used to reset the color of shape to match data model

# Testing this PR

Scripts have already been tested but a final check will be good

To test:
 * Using 5.2 server, create shape and set fill color and stroke color
 * Upgrade to 5.3. Check the color, they should not match the ones previously set
 * Run shape_color script to reset the color. They should now match the color created in 5.2

# Related reading

https://trello.com/c/sGlYbbvB/213-color-in-shape-settings

cc @bramalingam @mtbc

### DIFF
--- a/sql/psql/OMERO5.3__0/reverse_shape_color_argb_to_rgba.sql
+++ b/sql/psql/OMERO5.3__0/reverse_shape_color_argb_to_rgba.sql
@@ -23,3 +23,5 @@ UPDATE shape
   WHERE (strokecolor IS NOT NULL OR
          fillcolor   IS NOT NULL) AND
         TRUE;
+
+DROP FUNCTION unfix_color(INTEGER);
--- a/sql/psql/OMERO5.3__0/reverse_shape_color_argb_to_rgba.sql
+++ b/sql/psql/OMERO5.3__0/reverse_shape_color_argb_to_rgba.sql
@@ -13,6 +13,10 @@ BEGIN
 
 END;$$ LANGUAGE plpgsql;
 
+--
+-- Edit TRUE to apply change to a subset of ROIs
+--
+
 UPDATE shape
   SET strokecolor = CASE WHEN strokecolor IS NOT NULL THEN unfix_color(strokecolor) END,
       fillcolor   = CASE WHEN fillcolor   IS NOT NULL THEN unfix_color(fillcolor)   END
--- a/sql/psql/OMERO5.3__0/reverse_shape_color_argb_to_rgba.sql
+++ b/sql/psql/OMERO5.3__0/reverse_shape_color_argb_to_rgba.sql
@@ -1,0 +1,21 @@
+CREATE OR REPLACE FUNCTION unfix_color(color INTEGER) RETURNS INTEGER AS $$
+
+DECLARE
+    least_3_bytes CONSTANT INTEGER := (1 << 24) - 1;
+    least_1_byte CONSTANT INTEGER := (1 << 8) - 1;
+    rgb INTEGER;
+    alpha INTEGER;
+
+BEGIN
+    rgb := color >> 8 & least_3_bytes;
+    alpha := color & least_1_byte;
+    RETURN alpha << 24 | rgb;
+
+END;$$ LANGUAGE plpgsql;
+
+UPDATE shape
+  SET strokecolor = CASE WHEN strokecolor IS NOT NULL THEN unfix_color(strokecolor) END,
+      fillcolor   = CASE WHEN fillcolor   IS NOT NULL THEN unfix_color(fillcolor)   END
+  WHERE (strokecolor IS NOT NULL OR
+         fillcolor   IS NOT NULL) AND
+        TRUE;
--- a/sql/psql/OMERO5.3__0/shape_color_argb_to_rgba.sql
+++ b/sql/psql/OMERO5.3__0/shape_color_argb_to_rgba.sql
@@ -1,0 +1,25 @@
+CREATE OR REPLACE FUNCTION fix_color(color INTEGER) RETURNS INTEGER AS $$
+
+DECLARE
+    least_3_bytes CONSTANT INTEGER := (1 << 24) - 1;
+    least_1_byte CONSTANT INTEGER := (1 << 8) - 1;
+    rgb INTEGER;
+    alpha INTEGER;
+
+BEGIN
+    rgb := color & least_3_bytes;
+    alpha := color >> 24 & least_1_byte;
+    IF alpha = 0 THEN
+        alpha := least_1_byte;
+    END IF;
+	  RETURN rgb << 8 | alpha;
+
+END;$$ LANGUAGE plpgsql;
+
+
+UPDATE shape
+  SET strokecolor = CASE WHEN strokecolor IS NOT NULL THEN fix_color(strokecolor) END,
+      fillcolor   = CASE WHEN fillcolor   IS NOT NULL THEN fix_color(fillcolor)   END
+  WHERE (strokecolor IS NOT NULL OR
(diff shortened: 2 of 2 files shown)


sql/psql/OMERO5.3__0/reverse_shape_color_argb_to_rgba.sql
```diff
@@ -0,0 +1,21 @@
...
+    RETURN alpha << 24 | rgb;
+
+END;$$ LANGUAGE plpgsql;
+
+UPDATE shape
+  SET strokecolor = CASE WHEN strokecolor IS NOT NULL THEN unfix_color(strokecolor) END,
+      fillcolor   = CASE WHEN fillcolor   IS NOT NULL THEN unfix_color(fillcolor)   END
+  WHERE (strokecolor IS NOT NULL OR
+         fillcolor   IS NOT NULL) AND
+        TRUE;
```
mtbc, line 25: could add a comment here saying to adjust `TRUE` as required

sql/psql/OMERO5.3__0/shape_color_argb_to_rgba.sql
```diff
@@ -0,0 +1,25 @@
...
+    rgb INTEGER;
+    alpha INTEGER;
+
+BEGIN
+    rgb := color & least_3_bytes;
+    alpha := color >> 24 & least_1_byte;
+    IF alpha = 0 THEN
+        alpha := least_1_byte;
+    END IF;
+	  RETURN rgb << 8 | alpha;
```
mtbc: something odd happened to indent / spacing

sql/psql/OMERO5.3__0/shape_color_argb_to_rgba.sql
```diff
@@ -0,0 +1,25 @@
...
+
+END;$$ LANGUAGE plpgsql;
+
+
+UPDATE shape
+  SET strokecolor = CASE WHEN strokecolor IS NOT NULL THEN fix_color(strokecolor) END,
+      fillcolor   = CASE WHEN fillcolor   IS NOT NULL THEN fix_color(fillcolor)   END
+  WHERE (strokecolor IS NOT NULL OR
+         fillcolor   IS NOT NULL) AND
+        TRUE;
```
mtbc, line 28: could add a comment here saying to adjust `TRUE` as required

mtbc:
Comments look good to me. :+1:

dominikl:
Hm, doesn't seem to work properly. It modifies the color, but not like expected.

ROIs in 5.2.7:
5 2 7_rois

ROIs after upgrade to 5.3.0:
5 3 0_rois

ROIs after applying the script:
5 3 0_rois_afterscript

dominikl:
That's really strange, OMERO.Web (bottom) is alright, it's just Insight (top):

screen shot 2017-03-16 at 13 39 12

mtbc:
Web looks kind of weird too?

dominikl:
Oh yes, sorry, you're right @mtbc , it should match the top screenshot, which it doesnt...

mtbc:
I gave this a try:

Started from tag `v5.2.7` and drew some ROIs which looked in Insight like this,

insight-5 2 7

Using,

```sql
select discriminator as type, to_hex(strokecolor) as stroke, to_hex(fillcolor) as fill from shape order by id
```

these ROIs appeared in the database as,

```
  type   |  stroke  |   fill   
---------+----------+----------
 rect    | c400ff00 | ffa40000
 ellipse | ffe6b400 | ff0000b4
 line    | c4ff00ff | 40808080
```

I then merged this PR onto https://github.com/openmicroscopy/openmicroscopy/commit/c28b409a413e6f237583e322c420a0a7bb3fa18b and ran the normal database upgrade script.

In Insight the ROIs now looked like,

insight-5 3 0-before

and the database was just as before. After running the `sql/psql/OMERO5.3__0/shape_color_argb_to_rgba.sql` script Insight showed,

insight-5 3 0-after

and web showed,

web-5 3 0-after

and the database contained,

```
  type   |  stroke  |   fill   
---------+----------+----------
 rect    | ff00c4   | a40000ff
 ellipse | e6b400ff | b4ff
 line    | ff00ffc4 | 80808040
```

Now I'll try out the reverse script.

mtbc:
The reverse script moves back to exactly the 5.2.7 database state:

```shell
$ md5sum postgres-* | sort
3dddc8a81597fdc8758a13d24756f96e  postgres-5.2.7.txt
3dddc8a81597fdc8758a13d24756f96e  postgres-5.3.0-A-after-upgrade.txt
3dddc8a81597fdc8758a13d24756f96e  postgres-5.3.0-B-after-insight.txt
3dddc8a81597fdc8758a13d24756f96e  postgres-5.3.0-D-after-fix-unfix-color.txt
d133fc14d211cd180f901e9cc3f367d3  postgres-5.3.0-C-after-fix-color.txt
d133fc14d211cd180f901e9cc3f367d3  postgres-5.3.0-E-after-fix-unfix-fix-color.txt
```

dominikl:
I don't understand it, tested it again. 5.2.7 server and client from the downloads page, 5.3.0-m9-307-1fc4dfc-ice36-b600 server and client from artifacts. The first time I checked out the script from @jburel 's PR, the second time I took the script included in server build. Both times the weird results.

5.2.7:
screen shot 2017-03-16 at 14 49 21

5.3.0 after script:
screen shot 2017-03-16 at 14 46 34

mtbc:
Can you check your before-and-after in the database? That may yield clues.

mtbc:
It's also interesting that you got Insight and Web looking different from each other!

dominikl:
I used a medium alpha/transparency setting of around 128 on each of the colors... 
Yes, that Insight and Web is different is especially weird, that shouldn't be possible. 
Maybe someone else can check it, too and then we make a majority decision, I don't know.

mtbc:
After the `UPDATE`s the scripts probably should drop the functions.

mtbc:
So, we're going ARGB → RGBA. If I try a lower alpha,

```
# select to_hex(fix_color(cast(x'82123456' as int))) as foo;
   foo    
----------
 12345682
```

If you play with the alpha in `psql` you get expected results from the `fix_color` function?

mtbc:
Which version of PostgreSQL are you using? Would be good to make sure there's not an important difference there. I'm on 9.4.10 so probably well behind you Homebrew folks!

dominikl:
9.5.3. I still can't figure it out, looking at the values in the db, the script does exactly the right thing, like for you, but still in the clients I see these strange colors (and I've not taken any drugs :-) ). Anyway, the script works 👍

jburel:
tested with 9.5.6 on ubuntu16.04 
insight 5.2.7
roi_insight-5 2 7

5.3 before patch 
roi_insight-5 3-before-color-patch

5.3 after patch 
roi_insight-5 3-after-color-patch

mtbc:
If we ever find out what's happening on @dominikl's system I'd love to know!

mtbc:
Further to https://github.com/openmicroscopy/openmicroscopy/pull/5182#issuecomment-287085277 finishing the scripts with,

```sql
DROP FUNCTION fix_color(INTEGER);
```

and similar for `unfix_color` for the reverse script would do.

jburel:
testing the latest commit should show
```
CREATE FUNCTION
UPDATE 3
DROP FUNCTION
```
3 is the number of shapes I had so this can be different

dominikl:
Script works fine! I simply ran the wrong script earlier, `reverse_shape_color_argb_to_rgba.sql`. Sorry for all the hassle.

jburel:
@dominikl Thanks
Merging so we can get the doc in too

PR #5191, opened 2017-03-16
Exclude TablesFacility integration test

dominikl:
# What this PR does

Excludes the TablesFacility integration test for now, because it's quite flaky on CI (timeout when running `addTable()` method). Not sure what to do on the long run, strip down the test to only create a small random table or if possible increase resources available to CI, so that it can handle large tables more reliably?

# Testing this PR

Just check the TablesFacility test does *not* run.

### DIFF
--- a/components/tools/OmeroJava/test/gateway.testng.xml
+++ b/components/tools/OmeroJava/test/gateway.testng.xml
@@ -10,7 +10,7 @@
 			<class name="integration.gateway.RawDataFacilityTest" />
 			<class name="integration.gateway.MetadataFacilityTest" />
 			<class name="integration.gateway.ROIFacilityTest" />
-			<class name="integration.gateway.TablesFacilityTest" />
+			<!-- <class name="integration.gateway.TablesFacilityTest" /> excluded because addTable method is flaky on CI -->
 		</classes>
 	</test>
 </suite>


manics:
Was there a specific reason for using 100000 rows?

dominikl:
No, the intention was to create a reasonable sized table to perform some tests on. I just picked 100000 rows by 10 columns, because that creates a table file of size between 2 and 6GB, which I assumed could be reasonable size.

manics:
How about two tests then:
- A small table that shouldn't hit any CI limits e.g. 1001 rows x 2 cols
- The current table test, excluded

dominikl:
Yes, can do that, for 5.3.1, updated card https://trello.com/c/wPBnTfPU/2-flaky-tests . I'll just exclude the whole test for now.

jburel:
Tests seem to have run https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.gateway/TablesFacilityTest/

dominikl:
Looks like the integration test job just runs every test it can find, and ignores `gateway.testng.xml`. Anyway, I leave the tests in, but change the size of the random table which is created, then the test should be less flaky.

dominikl:
Close in favor of #5193

PR #5193, opened 2017-03-17
Reduce table size for integration test

dominikl:
# What this PR does

Reduces the size of the table which is created for the integration tests, due to flakiness of the `addTable()` method/test on the CI.

# Testing this PR

Check that `TablesFacilityTest`s are passing.


jburel:
run 2 builds over the weekend.
Back to green, merging ahead of rc

PR #5199, opened 2017-03-24
Remove the hidden file restrictions

dominikl:
# What this PR does

For some reason Insight/Importer didn't allow to import hidden files (which includes files in hidden directories). This PR removes this restriction.

# Testing this PR

Has to be tested on Windows. Linux and OSX don't show the hidden folders/files in the import dialog in first place, but Windows does if you set the Windows Explorer to do so.

- Create a hidden folder and put some image files into it.
- Import them via Insight and/or Importer.

# Related reading
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2017-March/006399.html


kennethgillen:
👍  @dominikl 

Successfully imported hidden files in a hidden folder on Windows Server 2008 R2, equivalent to Windows 7.

screen shot 2017-03-30 at 11 31 30

dominikl:
Thanks for testing @kennethgillen !

jburel:
Do we want a note in the help
i.e. http://help.openmicroscopy.org/importing-data-5.html

PR #5202, opened 2017-03-28
handle requestline limit

will-moore:
# What this PR does

Improves handling of error when request string is too long.
See https://trello.com/c/iDzN7Dw9/21-request-string-too-long

When many (e.g. >400) images are selected, the right hand panel loads a url that is too long
with all the images as ```/?image=1&image=2&image=3...```.
Now, instead of an error with feedback form, we simply show this error in the right panel.

screen shot 2017-03-28 at 12 27 57

This relies on the response having text "Request Line is too large" in it.

# Testing this PR

1. In order to select a large number of objects you will need lots of images in a Dataset and to configure the ```omero.web.page_size = 500``` OR lots of Datasets in Project (not paginated) OR search for something that gives > 400 results.

2. Select many objects. Check that you see the expected message in right panel (see screenshot).

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/data/includes/right_plugin.general.js.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/data/includes/right_plugin.general.js.html
@@ -76,6 +76,14 @@ $(function () {
                         }
                     }
                     $metadata_general.html(data);
+                },
+                error: function(data) {
+                    // If request string too long. e.g. image=1&image=2&image=3.... etc
+                    // Just show the error message in right panel.
+                    // This error will be ignored by OME.setupAjaxError();
+                    if (data.responseText.indexOf('Request Line is too large') > -1) {
+                        $metadata_general.html(data.responseText);
+                    }
                 }
             });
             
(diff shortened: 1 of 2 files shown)


mtbc:
So far I am seeing messages like "Selected 456 objects" (for images) or "Selected 499 objects" (for datasets); do I just need to bump the numbers up even higher?

mtbc:
Now "Selected 997 objects" for datasets and "Selected 771 objects" for images, right panel looks fine. Is there some reason I need to try to do all this on eel instead?

mtbc:
Ah, it turns out we need web running in nginx for reviewing this PR.

will-moore:
Ah, I see on web-dev-merge, I am getting a different status code for this error.
414 instead of 400:
```
414 Request-URI Too Large
nginx/1.10.2
```

mtbc:
Ah, maybe have to adjust `req.status` check in this PR anyway then!

will-moore:
This seems to be correct code for this error, e.g. see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414

Locally I was getting 400 errors for this, using a different version of nginx: 1.10.3

mtbc:
It seems like they regressed! I wonder if it was intentional.

bramalingam:
Works as expected. Looks good to merge.

PR #5141, opened 2017-03-02
Light administrator permissions

mtbc:
# What this PR does

These commits provide various light administrator permissions.

# Testing this PR

Use accompanying integration tests plus additional test and client work from @pwalczysko and @will-moore. Also see devspace at https://10.0.51.154:8443/.

# Related reading

https://trello.com/b/VkiONV40/new-role

PR #5204, opened 2017-03-29
Light administrator permissions with database triggers

mtbc:
Supplements #5141 with work from https://trello.com/c/caAHqOPa/13-fix-database-trigger-concurrency.

This PR uses database triggers to close the security holes arising from admins using the update service to elevate privileges. Integration tests check these.

This PR also fixes `canEdit`, `canDelete` and adds `canChgrp`, `canChown`. Integration tests are included and @pwalczysko includes further in #5248.

Database patch `OMERO5.4DEV__1` is introduced.


mtbc:
Next to investigate test failures at https://10.0.51.154:8443/job/OMERO-test-integration/lastCompletedBuild/.

mtbc:
The remaining integration test failures predate this PR so it's probably now worth a look, though before actually merging I'd like to investigate those other tests too and add commits to bring us to patch 1 on top of https://github.com/openmicroscopy/openmicroscopy/pull/5211.

mtbc:
On the other hand, the remaining failures predating this PR are OmeroWeb failures which on local investigation are horrific to figure out from my point of view, I'm really not sure what's going on there nor why, though admittedly they *do* pass on https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_api_images/TestImages/. But outwith this PR anyway so I've created https://trello.com/c/nWvUzN5t/24-omeroweb-integration-test-failures accordingly.

Travis should pass after we merge #5209 into `roles` but first I'll also await #5206.

mtbc:
Ha, travis now passes, thank you @sbesson.

pwalczysko:
For what its worth, my tests in https://github.com/openmicroscopy/openmicroscopy/pull/5248 are passing on top of this PR, i.e. the can* methods do what I imagine they should be doing (although https://github.com/openmicroscopy/openmicroscopy/pull/5248 is still work in progress.) FMPOV this PR should be merged as it gives me a better baseline to work with.

dominikl:
~~Do the integration tests really run? E. g. can't find the test result for `LightAdminPrivilegesTest ` in https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ Or is that the wrong place to look for them?~~
Edit: Guess, the integration test results are here: https://10.0.51.154:8443/job/OMERO-test-integration/  But doesn't include the Java integration tests.

mtbc:
https://10.0.51.154:8443/job/OMERO-test-integration/lastCompletedBuild/testngreports/integration/LightAdminPrivilegesTest/

dominikl:
Ah, it's in the TestNG results, of course, thanks.

components/server/src/ome/security/CompositeACLVoter.java
```diff
@@ -100,6 +101,21 @@ public void throwUpdateViolation(IObject object) throws SecurityViolation {
         return choose().restrictions(object);
     }
 
+    @Override
+    public void setPermittedClasses(Map<Integer, Set<Class<? extends IObject>>> objectClassesPermitted) {
+        basic.setPermittedClasses(objectClassesPermitted);
```
joshmoore, line 106: Why is this `basic`?

components/server/src/ome/security/basic/BasicACLVoter.java
```diff
@@ -84,12 +89,27 @@
...
+    /* ignored if empty set */
+    private Set<Class<? extends IObject>> chgrpPermittedClasses = Collections.emptySet();
+
+    /* ignored if empty set */
+    private Set<Class<? extends IObject>> chownPermittedClasses = Collections.emptySet();
+
     private final LightAdminPrivileges adminPrivileges;
 
+    /* preserve permissions past context invalidation through to post-processing */
+    private final ThreadLocal<Set<AdminPrivilege>> adminPrivilegesCache = new ThreadLocal<>();
```
joshmoore, line 101: Not as a must, but if possible I'd suggest concentrating `ThreadLocal` instances in as few places as possible. Here it would most likely be `CurrentDetails`:

```
$ git grep new\ ThreadLocal | grep -v OmeroCpp | grep -v TEST
components/server/src/ome/security/SecurityFilterHolder.java:    protected ThreadLocal<SecurityFilter> current = new ThreadLocal<SecurityFilter>() {
components/server/src/ome/security/SecuritySystemHolder.java:    protected ThreadLocal<SecuritySystem> current = new ThreadLocal<SecuritySystem>() {
components/server/src/ome/security/basic/CurrentDetails.java:    private final ThreadLocal<LinkedList<BasicEventContext>> contexts = new ThreadLocal<LinkedList<BasicEventContext>>();
components/server/src/ome/security/basic/CurrentDetails.java:        = new ThreadLocal<Map<String, String>>();
components/server/src/ome/tools/hibernate/ProxyCleanupFilter.java:        private final ThreadLocal<Integer> depth = new ThreadLocal<Integer>() {
```

components/server/src/ome/security/CompositeACLVoter.java
mtbc: We don't have a particular session context at the time this is called when the graph request factory sets up so `choose()` has no basis to act, and we know it's only the `BasicACLVoter` that we care about here. (Indeed, once we get rid of shares, a bunch of this can be much simplified!)

In general forcing light admin checks into the prior security system has been rather unfortunate in making dependencies messier as we keep needing extra information and half the time it's via some cyclic dependency. When we have the basic functionality working it would be good to look again at the bean architecture here to see where I could clean things up. https://trello.com/c/HRJyHGHh/18-review-implementation-approach notes some of these action items.

mtbc: (In this case we could eventually break out a new permitted-classes configuration parsing bean that is a common dependency of the basic ACL voter and the graph request factory.)

components/server/src/ome/security/basic/BasicACLVoter.java
mtbc: Thank you, noted on https://trello.com/c/HRJyHGHh/18-review-implementation-approach as something to at least come back to. Possibly the privileges should be part of the event context in the first place!

joshmoore: Makes sense to me.

jburel:
Merging
Further improvements in follow-up PRs

PR #5205, opened 2017-03-29
add "mkdir" function to CLI "fs" plugin

mtbc:
# What this PR does

In the managed repository allows administrators create to empty directories that violate the template path.

# Testing this PR

Try out `bin/omero fs mkdir`. The `--parents` option is analogous to regular GNU mkdir' `-p` option.

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ManagedRepositoryTest/ should pass.

# Related reading

https://trello.com/c/aKMV6KJI/24-bug-fs-symlink-strategy-fails

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
@@ -1409,11 +1409,12 @@ public class ManagedRepositoryI extends PublicRepositoryI
         final EventContext ec = IceMapper.convert(effectiveEventContext);
         final FsFile rootOwnedPath = expandTemplateRootOwnedPath(ec, sf);
         final List<CheckedPath> pathsToFix = new ArrayList<CheckedPath>();
-        final List<CheckedPath> pathsForRoot;
+        List<CheckedPath> pathsForRoot;
 
         /* if running as root then the paths must be root-owned */
+        final ome.system.EventContext currentEventContext = adminService.getEventContext();
         final long rootId = adminService.getSecurityRoles().getRootId();
-        if (adminService.getEventContext().getCurrentUserId() == rootId) {
+        if (currentEventContext.getCurrentUserId() == rootId) {
             pathsForRoot = ImmutableList.copyOf(paths);
         } else {
             pathsForRoot = ImmutableList.of();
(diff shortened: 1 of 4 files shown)


will-moore:
Tested locally:
Created a top-level dir without using ```--parents```. OK
Tried to create a nested dir without ```--parents``` which gave me a suitable SecurityViolation.
Running the same command with ```--parents``` worked fine:
```
$ omero fs mkdir test5205_test/temp/test --parents

$ ls -al /OMERO/ManagedRepository/test5205_test/temp
total 0
drwxr-xr-x  3 wmoore  DUNDEE\Domain Users  102 30 Mar 16:07 .
drwxr-xr-x  3 wmoore  DUNDEE\Domain Users  102 30 Mar 16:07 ..
drwxr-xr-x  2 wmoore  DUNDEE\Domain Users   68 30 Mar 16:07 test
```

Looks good to merge AFAIKT.

joshmoore:
Relaunched travis after timeouts. Any thoughts on an integration test for this (attempts to use as different users, etc.)?

mtbc:
I was wondering about that too. Blitz integration tests would need to be able to double-check the filesystem directly, not sure if we have examples of that. Python CLI tests would be a good idea but I am rather unsure of the Python testing environment in general and certainly unfamiliar with how CLI tests work; I would be pretty inefficient if attempting it myself right now. My thinking was that the priority is, descending,

1. fixing this workflow
1. getting the light admin stuff working
1. automating the testing of this workflow

Not sure if @dominikl or @pwalczysko might plausibly get chance to cause testing though.

I would suggest:

1. tests as admin (root and non-root admin)
1. tests as normal user
1. try @will-moore's approach of comparing `--parents` with otherwise
1. check the ownership of the new directories and their parents

An interesting awkwardness is that the code (via API, not CLI) *should* allow normal users to create directories that do match the template path (just as they can during import), except that it would take a template path using, say, `%increment%` instead of `%time%`, to be able to know what that path even is. But then we could test both directories that do match the path and ones that don't.

joshmoore:
https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/ome/services/blitz/repo/RawAccessRequestI.java is quite useful for testing purposes.

manics:
You could have an even higher level of testing, e.g. https://github.com/openmicroscopy/omero-grid-docker/blob/master/test_dropbox.sh

pwalczysko:
I can have a look at it (writing a new CLI test).

mtbc:
@pwalczysko: Will be happy to cherry-pick your commits here if you like. Alternatively I *think* it would work if you had only your commits in a PR with `--depends-on #5205`.

pwalczysko:
@mtbc Started https://github.com/openmicroscopy/openmicroscopy/pull/5254

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#668](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/668/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/668/consoleText) for more details.
Possible conflicts:
  - PR #4957 mtbc 'Import into other groups'
    - components/tools/OmeroJava/test/integration/ManagedRepositoryTest.java

joshmoore:
Code here as well as @pwalczysko's tests in gh-5254 look good. The only two issues that remain (and are somewhat above the scope of the addition):

 * Are there any cases where changing the parents to belong to root will lead to surprises? (I don't have any concretely yet) and if so, can they be detected?
 * As @manics suggested, is there a higher-level workflow that we want to suggest to users and therefore regularly test? For my part, I could almost see having `/disk1` as a standard part of an advanced configuration. (This could be automated in ansible)

mtbc:
I think the cases where parents would be changed to root would tend to be where an admin is creating a directory within a past fileset, which would be a bit odd.

Once this is merged I do plan to document a storage-expansion workflow that I figure we can at least test manually as part of that doc review. I could create a card for the automation of that scenario testing if you think we might approach having the bandwidth to get around to implementing it.

joshmoore:
:+1: for a longer-term fully self-describing test like the dropbox work.

PR #5206, opened 2017-03-29
expose enum constants in generated ome.model classes

mtbc:
# What this PR does

This PR exposes many enumeration values (at least the easier ones!) as string constants in OMERO.server. The idea is to eliminate bugs arising from server code having to copy literal values that then are not updated if the enumeration value is changed.

# Testing this PR

Take a look at how this PR changes the generated Java classes in `components/model/target/generated/src/ome/model/enums/`. This PR includes some changes to server code to illustrate with the `ChecksumAlgorithm` enumeration the use of these new constants. This PR should leave the server's actual behavior unchanged.

# Related reading

http://trac.openmicroscopy.org/ome/ticket/10691

### DIFF
--- a/components/dsl/resources/ome/dsl/object.vm
+++ b/components/dsl/resources/ome/dsl/object.vm
@@ -1369,7 +1369,6 @@ implements java.io.Serializable, IObject
     }
 #end
 
-    @SuppressWarnings("unchecked")
     @Override
     public boolean acceptFilter(ome.util.Filter __filter){
         try {
@@ -1490,7 +1489,6 @@ implements java.io.Serializable, IObject
         }
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public void putAt(String field, Object value)
     {
--- a/components/dsl/resources/ome/dsl/object.vm
+++ b/components/dsl/resources/ome/dsl/object.vm
@@ -1483,7 +1483,7 @@ implements java.io.Serializable, IObject
         } else if (field.equals(${prop.name.toUpperCase()}COUNTPEROWNER)) {
             return get${prop.nameCapped}CountPerOwner();
 #end
-        } else if (field.equals(${prop.nameUpper})) { // TODO use == here first??
+        } else if (field.equals(${prop.nameUpper})) {
             return get${prop.nameCapped}();
 #end
         } else {
--- a/components/dsl/resources/ome/dsl/object.vm
+++ b/components/dsl/resources/ome/dsl/object.vm
@@ -319,6 +319,15 @@ implements java.io.Serializable, IObject
     */
     private static final long serialVersionUID = 0000000030000020302L;
 
+#if($type.isEnum && !$type.shortname.startsWith("Units") && ${type.shortname} != "Binning" && ${type.shortname} != "PixelsType")
+#foreach($prop in $type.properties)
+#if($prop.class.name == "ome.dsl.EntryField")
+#set($enumSymbol = ${prop.name.replaceAll("-","_").replaceAll("[^a-zA-Z0-9_]","")})
+    public final static String VALUE_${enumSymbol} = "${prop.name}";
+#end
+#end
+
+#end
 #if(!$type.global)
    /*
     * Constants naming filters used by the OMERO
@@ -328,8 +337,8 @@ implements java.io.Serializable, IObject
     public final static String GROUP_FILTER = "${gfilter}";
     public final static String EVENT_FILTER = "${efilter}";
     public final static String PERMS_FILTER = "${pfilter}";
-#end
 
+#end
     public ${type.shortname} () {
         this(null, true);
     }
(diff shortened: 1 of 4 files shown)


components/dsl/resources/ome/dsl/object.vm
```diff
@@ -319,6 +319,15 @@ implements java.io.Serializable, IObject
     */
     private static final long serialVersionUID = 0000000030000020302L;
 
+#if($type.isEnum && !$type.shortname.startsWith("Units") && ${type.shortname} != "Binning" && ${type.shortname} != "PixelsType")
```
joshmoore, line 322: Units is excluded I assume due to unicode. Why Binning and PixelsType?

mtbc: They have values that wouldn't make good variable names if we rethought the `VALUE_` prefix. I thought better not to provide them for now if they might be more deeply rethought later, as they can easily be added in the unlikely event of anybody actually missing them. (Their values *are* already available to code on the Blitz code-generated side.)

Yeah, the codegen for the Blitz side does a bit of fiddling with the units, some of which is unnecessary from the Java point of view, and again I was trying to defer tricky decisions that we might later regret and just include the easy stuff while that already provides plenty of value added (at least from my point of view).

jburel: I think we should expose the pixels type too since we have static values in various classes e.g.
``PlaneFactory``.
Maybe in that case drop the ``VALUE``

Also I feel that the prefix should be upper case to match the usual way of introducing static fields and leads to a mix of upper/lower cases constants e.g.

```
public final static String VALUE_Brightfield = "Brightfield";
public final static String VALUE_Phase = "Phase";
public final static String VALUE_DIC = "DIC";
public final static String VALUE_HoffmanModulation = "HoffmanModulation";
public final static String VALUE_ObliqueIllumination = "ObliqueIllumination";
public final static String VALUE_PolarizedLight = "PolarizedLight";
public final static String VALUE_Darkfield = "Darkfield";
public final static String VALUE_Fluorescence = "Fluorescence";
public final static String VALUE_Other = "Other";
public final static String VALUE_Unknown = "Unknown";
```

unless there is a use case of retrieving the static field, dropping the ``VALUE`` and comparing with the type returned from the server

jburel: if we integrate pixels maybe better to have all i.e. include binning

mtbc:
Will have to rethink this.

mtbc:
The overall changes between this PR and its replacement #5242 turn out to be small:

```diff
diff --git a/components/dsl/resources/ome/dsl/object.vm b/components/dsl/resources/ome/dsl/object.vm
index 9205486..248c4be 100644
--- a/components/dsl/resources/ome/dsl/object.vm
+++ b/components/dsl/resources/ome/dsl/object.vm
@@ -319,10 +319,10 @@ implements java.io.Serializable, IObject
     */
     private static final long serialVersionUID = 0000000030000020302L;
 
-#if($type.isEnum && !$type.shortname.startsWith("Units") && ${type.shortname} != "Binning" && ${type.shortname} != "PixelsType")
+#if($type.isEnum && !$type.shortname.startsWith("Units"))
 #foreach($prop in $type.properties)
 #if($prop.class.name == "ome.dsl.EntryField")
-#set($enumSymbol = ${prop.name.replaceAll("-","_").replaceAll("[^a-zA-Z0-9_]","")})
+#set($enumSymbol = ${prop.name.replaceAll("-","_").replaceAll("([a-z])([A-Z])", "$1_$2").toUpperCase().replaceAll("[^A-Z0-9_]","")})
     public final static String VALUE_${enumSymbol} = "${prop.name}";
 #end
 #end
diff --git a/components/server/src/ome/services/RawFileBean.java b/components/server/src/ome/services/RawFileBean.java
index 0193603..2d8663e 100644
--- a/components/server/src/ome/services/RawFileBean.java
+++ b/components/server/src/ome/services/RawFileBean.java
@@ -78,13 +78,13 @@ public class RawFileBean extends AbstractStatefulBean implements RawFileStore {
     /* Map of checksum algorithm names to the corresponding checksum type. */
     private static final ImmutableMap<String, ChecksumType> checksumAlgorithms =
             ImmutableMap.<String, ChecksumType> builder().
-            put(ChecksumAlgorithm.VALUE_Adler_32, ChecksumType.ADLER32).
+            put(ChecksumAlgorithm.VALUE_ADLER_32, ChecksumType.ADLER32).
             put(ChecksumAlgorithm.VALUE_CRC_32, ChecksumType.CRC32).
             put(ChecksumAlgorithm.VALUE_MD5_128, ChecksumType.MD5).
-            put(ChecksumAlgorithm.VALUE_Murmur3_32, ChecksumType.MURMUR32).
-            put(ChecksumAlgorithm.VALUE_Murmur3_128, ChecksumType.MURMUR128).
+            put(ChecksumAlgorithm.VALUE_MURMUR3_32, ChecksumType.MURMUR32).
+            put(ChecksumAlgorithm.VALUE_MURMUR3_128, ChecksumType.MURMUR128).
             put(ChecksumAlgorithm.VALUE_SHA1_160, ChecksumType.SHA1).
-            put(ChecksumAlgorithm.VALUE_File_Size_64, ChecksumType.FILE_SIZE).
+            put(ChecksumAlgorithm.VALUE_FILE_SIZE_64, ChecksumType.FILE_SIZE).
             build();
 
     /** The id of the original files instance. */
```

PR #5210, opened 2017-03-29
slider lut bg reverse

will-moore:
# What this PR does

For channels without LUTs, this gives the slider a gradient from black -> colour.
Also, gradient and LUTs slider background is flipped for Reverse Intensity channels, so that the slider background matches the current rendering (and gives visual indication of which channels are Reverse Intensity).
E.g. in this screenshot it can be clearly seen that the Green channel is Reverse Intensity.

screen shot 2017-03-29 at 23 06 05

# Testing this PR

1. For channels without LUT, slider background should be black -> colour gradient.
2. When Reverse Intensity is chosen, channel slider background is also reversed (with OR without LUTs) so that it matches the current rendering mapping.

# Related reading

https://trello.com/c/weNESvzP/55-reverse-intensity-flips-slider-background


waxenegger:
Looks good. Working as expected, just not in iviewer any more. I have to make a mental note of changing things once this has been merged...

jburel:
Discussed today. Such change has an impact on various repositories i.e. iviewer, figure also insight is affected
We need to be careful when introducing such adjustment

will-moore:
OMERO.figure and OMERO.insight (I presume) don't use the webgateway ```luts_10.png```. They have their own copies of this so they are not affected by this change.
I assume we're not implementing these same changes in Insight now?
If we consider this to be a breaking change, then I can leave the png unchanged and simply add the new one I'm using?

jburel:
true that hey have their own copy, my point was that we need to keep the feature in synch as much as we can so it does not impact the user experience.

waxenegger:
@will-moore I have a PR that should fix my problem with the png. Cannot get it merge-deployed (conflict at the moment) but still, it should do the job regardless (of whether/when the changes become release) so I don't think closing is necessary (from my perspective at least).

will-moore:
To test last commit, it should fix the blurring of adjacent LUTs in the background of the LUT-picker, channel button and channel slider. See before and after 

screen shot 2017-04-26 at 12 39 33

screen shot 2017-04-26 at 12 40 33

waxenegger:
So I could not see that issue any more after you committed the last time. I kind of checked and compared in between web and iviewer quite a bit more recently. I think that's good to be merged.

PR #5208, opened 2017-03-29
Well preview panel

will-moore:
# What this PR does

Adds back support for Preview panel when a Well is selected.
Browsing Wells is much nicer when you can directly preview the image (as you could in 5.2.x), instead of having to select Image first.

Also, we only show the spatial Well layout panel (below tree) when we have Well Sample position info and this panel is hidden when selecting data in tree since we know that non-Well data is being selected.

# Testing this PR

1. Browse Plate, click on Well and see that Preview tab is enabled and should work as normal.
2. Selecting a different field (chooser above plate) and re-selecting a Well should show the current field in Well preview tab.
3. Browsing Wells that have no spatial info for WellSamples will not show the spatial layout panel below tree.
4. When spatial panel is shown (when WellSamples have position data) and then you select data in tree, panel will be hidden again.

# Related reading

https://trello.com/c/kdYC8t7w/54-spw-layout-follow-up

PR #5212, opened 2017-03-30
Insight530 bugs

dominikl:
# What this PR does

Some bugfixes for Insight.

1) Update copyright date in 'About' dialog
2) Avoid Insight crash, when viewing an image with invalid pixel sizes. ()
3) Double click to open full viewer didn't work on field thumbnails.
4) Large image JPEG export from full viewer could crash Insight
5) ~~Preview was disabled for Wells~~

# Testing this PR

1) Open the 'About' dialog (Help), check copyright date.
2) Once nightshade is upgraded to 5.3, open image `3859301`, Insight shouldn't crash.
3) Double click on an field thumbnail, full viewer should open with the field image.
4) Open a large (> 12k x 12k px) image in full viewer. Zoom out, so that the zoomed image is smaller than 12k x 12k px. You should be able to save it as JPEG/PNG/... . Zoom in. Once the zoomed image would be bigger than 12k x 12k px, the "Save as" button should be disabled.
5) ~~Select a well, check that preview shows the selected field for the well.~~

# Related reading
Re 2) https://www.openmicroscopy.org/qa2/qa/feedback/17673/ 
Re 4) http://lists.openmicroscopy.org.uk/pipermail/ome-users/2017-March/006409.html

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BrowserModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BrowserModel.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2016 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2017 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -388,7 +388,8 @@ class BrowserModel
      */
     private double getBarSizeInPx(double ratio)
     {
-        if(unitBarLength == null)
+        if (unitBarLength == null
+                || !Double.isFinite(getPixelsSizeX().getValue()))
             return 1;
         
         try {
@@ -753,6 +754,10 @@ class BrowserModel
      */
     String getUnitBarValue()
     {
+        if (unitBarLength == null
+                || !Double.isFinite(getPixelsSizeX().getValue()))
+            return "1";
+        
     	return UIUtilities.twoDecimalPlaces(unitBarLength.getValue());
     }
     
@@ -761,6 +766,10 @@ class BrowserModel
      * @return See above.
      */
     String getUnitBarUnit() {
+        if (unitBarLength == null
+                || !Double.isFinite(getPixelsSizeX().getValue()))
+            return LengthI.lookupSymbol(UnitsLength.PIXEL);
+        
         return LengthI.lookupSymbol(unitBarLength.getUnit());
     }
     
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BrowserModel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BrowserModel.java
@@ -389,7 +389,8 @@ class BrowserModel
     private double getBarSizeInPx(double ratio)
     {
         if (unitBarLength == null
-                || !Double.isFinite(getPixelsSizeX().getValue()))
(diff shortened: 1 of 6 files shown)


jburel:
see https://github.com/openmicroscopy/openmicroscopy/pull/5208
for enabling the view of a well 

Note that I can also double-click on  a well in centre panel and open the viewer but the ``Full Viewer`` is not enabled

components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/browser/BrowserModel.java
```diff
@@ -388,7 +388,9 @@ private void createGridImages()
      */
     private double getBarSizeInPx(double ratio)
     {
-        if(unitBarLength == null)
+        if (unitBarLength == null
+                || Double.isInfinite(getPixelsSizeX().getValue())
```
jburel, line 392: maybe better to use the same strategy than the one used in 
``EditorUtil.transformPixelsData`` i.e. if  ``BigResult`` is returned then 1

dominikl: I don't think a `BigResult` exception is thrown in that case, but a `NumberFormatException`, see https://www.openmicroscopy.org/qa2/qa/feedback/17673/ , that's why the check for finite.

dominikl:
For enabling the "Full Viewer" button I had the same problem like when I was trying to enable the preview for wells: Need to know the field index; added a static variable for that purpose in the DataBrowserAgent. Having this variable now, I also enabled the preview for wells.
For **testing**, see PR description item 5.

pwalczysko:
All testing steps passed except Nr. 5:

  - select a well, then click onto Preview tab -> works, albeit the wait is considerable
  - with the Preview tab still open, click on another well -> observe endless spinner (waited for minutes, never anything came) -> not expected

screen shot 2017-04-10 at 10 17 42

pwalczysko:
Endless spinner on already opened Preview of a Well bug https://github.com/openmicroscopy/openmicroscopy/pull/5212#issuecomment-292895478 is fixed. Ready to merge FMPOV.

components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/ImViewerAgent.java
```diff
@@ -31,6 +31,7 @@
 import ome.model.units.BigResult;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.openmicroscopy.shoola.agents.dataBrowser.DataBrowserAgent;
```
jburel: problem is that you create a dependency between the 2. 
Why can't you post an event that the viewer will listen to?

dominikl: It's just so much boilerplate code for a single integer value I'd need. But yes, I should do it properly.

pwalczysko:
Double-checked today, the problem with the endless spinner is still solved and all is as expected.

jburel:
@dominikl: to break this PR into 2 so the fix not related to well sample field can be included in 5.3.1

dominikl:
Removed the 'well preview' commits (will open a separate PR for these). Note: That also means that https://github.com/openmicroscopy/openmicroscopy/pull/5212#issuecomment-292189537 is *not* fixed with this PR.

jburel:
nightshade is being upgraded today, so I will check point 2) tonight since we can't download the original one
The rest works as expected.

jburel:
Replaced by https://github.com/openmicroscopy/openmicroscopy/pull/5256

PR #4436, opened 2016-01-24
Search temp fix

aleksandra-tarkowska:
Disable id helper in search and allow search in all groups by default, as tweaked in #4246

PR #5214, opened 2017-04-04
Search temp fix (rebased onto metadata53)

atarkowska:
This is the same as gh-4436 but rebased onto metadata53.

----

Disable id helper in search and allow search in all groups by default, as tweaked in #4246


atarkowska:
--rebased-from #4436

PR #5215, opened 2017-04-04
render.py: new CLI plugin for dealing with rendering settings (manual rebase onto metadata52) (rebased onto metadata53)

atarkowska:
This is the same as gh-4542 but rebased onto metadata53.

----

This was the same as gh-4385 gh-4430 but rebased onto `metadata52`. Now with additional commits.

---

```
```

See https://github.com/IDR/idr-metadata/pull/75 for information on the intended workflow for `omero render edit`.

Perhaps the main discussion point is the use of yaml/json for these settings, and the implementation of `pydict_text_readers.py`. `load` is designed to transparently handle:
- json or yaml (former is stricter on types and is better for generated definitions, the latter is more user friendly if you're crafting it by hand)
- `/local/file/path`s or remote `OriginalFile:Id`s
- If you're happy with this design I'll re-use it in the metadata plugin.
- For the future: could have json/yaml write methods which could tie in with the new `obj`/`tree` plugin https://github.com/openmicroscopy/design/pull/31 in which case `pydict_text_readers.py` should be renamed. 

---

Additional commits implement `omero render edit` to allow rendering settings to be set on an image.
- [x] Editing of  channel names isn't implemented yet, and tests are pending.
- [ ] When testing this plugin I ran into several `Too many open files` errors on the server, and I'm not sure if it's an error in this code, a system misconfiguration, or an OMERO.server bug


atarkowska:
--rebased-from #4542

PR #5216, opened 2017-04-04
Add support for twitter cards (full image viewer only) (rebased onto metadata52) (rebased onto metadata53)

atarkowska:
This is the same as gh-4613 but rebased onto metadata53.

----

This is the same as gh-4260 but rebased onto metadata52.

---


atarkowska:
--rebased-from #4613

joshmoore:
In a follow-up (when @manics is back) we may need to address https://github.com/openmicroscopy/openmicroscopy/pull/4613#issuecomment-241111240

manics:
--rebased-to #5795

PR #5218, opened 2017-04-04
Metadata column column types and datasets (rebased onto metadata53)

atarkowska:
This is the same as gh-4637 but rebased onto metadata53.

----

This PR carries on from work by @emilroz  (PR 120) which makes it possible to specify the column types for OMERO.tables when using populate_metadata.py to parse the csv files and adds support for datasets and projects as the target object for a metadata table.


atarkowska:
--rebased-from #4637

PR #4750, opened 2016-07-14
Change selected to All Users/Memebers

emilroz:
This PR extends https://github.com/openmicroscopy/openmicroscopy/pull/4436 by setting the default to 'All Members' on each group.

PR #5219, opened 2017-04-04
Change selected to All Users/Memebers (rebased onto metadata53)

atarkowska:
This is the same as gh-4750 but rebased onto metadata53.

----

This PR extends https://github.com/openmicroscopy/openmicroscopy/pull/4436 by setting the default to 'All Members' on each group.


atarkowska:
--rebased-from #4750

PR #5220, opened 2017-04-04
populate_metadata.py: add batches to write_to_omero (rebased onto metadata53)

atarkowska:
This is the same as gh-4754 but rebased onto metadata53.

----

# What this PR does

The new `QueryContext` implementations in `populate_metadata.py` were not batching in their `write_to_omero` methods. For very large screens (e.g. idr0016):
- `BulkToMapAnnotationContext` tends to hit MESSAGESIZEMAX limits and
- `DeleteMapAnnotationContext` hits `Caused by: java.io.IOException: Tried to send an out-of-range integer as a 2-byte value: 109734`

Now both are done in batches of 1000.

_Note:_ projections are also causing issues so batching is being added to invocations of `projection` as well.
# Testing this PR
1. A bulk annotation YML configuration is required. This is minimal possible via `test/integration/metadata/test_populate.py`.
2. There should be no change in functionality and no overt slowdown in creation or deletion.
# Related reading
- https://trello.com/c/urQCwi0v/470-error-creating-map-annotations-in-demo2-for-idr0016
- https://trello.com/c/88jfqEem/469-error-deleting-map-annotations-for-idr0001-screena

cc: @eleanorwilliams

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/metadata.py
+++ b/components/tools/OmeroPy/src/omero/plugins/metadata.py
@@ -147,6 +147,11 @@ class MetadataControl(BaseControl):
         populate = parser.add(sub, self.populate)
         populateroi = parser.add(sub, self.populateroi)
 
+        populate.add_argument("--batch",
+                              type=long,
+                              default=1000,
+                              help="Number of objects to save at once")
+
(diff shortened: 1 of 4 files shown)


atarkowska:
--rebased-from #4754

PR #5221, opened 2017-04-04
Revert "Disable metadata.py with OMERO_DEV_PLUGINS" (rebased onto metadata53)

atarkowska:
This is the same as gh-4765 but rebased onto metadata53.

----

# What this PR does

Reverts the requirement to set `OMERO_DEV_PLUGINS` when using the `omero metadata` plugin. This env-var makes sense on the mainline, but since this plugin is in integral part of the work in the `metadata52` development branch it should be enabled by default.
# Testing this PR
1. Check `OMERO_DEV_PLUGINS` isn't set: `$ set | grep OMERO_DEV_PLUGINS`
2. Run `omero metadata ...`, the plugin should be enabled
# Related reading
- Original `develop` PR: https://github.com/openmicroscopy/openmicroscopy/pull/4426

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/metadata.py
+++ b/components/tools/OmeroPy/src/omero/plugins/metadata.py
@@ -500,8 +500,7 @@ class MetadataControl(BaseControl):
                 meas.parse_and_populate()
 
 try:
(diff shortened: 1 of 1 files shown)


atarkowska:
--rebased-from #4765

PR #4852, opened 2016-09-20
populate_metadata: warn on missing image

joshmoore:
# What this PR does

When a well is missing from a plate, a warning is printed. The
same now happens when an image is missing from a dataset. Likely,
a `--strict` argument should be added which will force the existence
of all objects.

# Testing this PR

1. Check all 29 `test/integration/metadata/test_populate.py` tests continue passing.
2. Attempt to annotate idr0023 (cc @eleanorwilliams). All present images should be annotated.

# Related reading

 * https://trello.com/c/q8Fr2ERU/477-in-non-screen-datasets-if-image-file-is-missing-then-bulk-annotation-can-t-be-added-as-gives-error

PR #5224, opened 2017-04-04
populate_metadata: warn on missing image (rebased onto metadata53)

atarkowska:
This is the same as gh-4852 but rebased onto metadata53.

----

# What this PR does

When a well is missing from a plate, a warning is printed. The
same now happens when an image is missing from a dataset. Likely,
a `--strict` argument should be added which will force the existence
of all objects.
# Testing this PR
1. Check all 29 `test/integration/metadata/test_populate.py` tests continue passing.
2. Attempt to annotate idr0023 (cc @eleanorwilliams). All present images should be annotated.
# Related reading
- https://trello.com/c/q8Fr2ERU/477-in-non-screen-datasets-if-image-file-is-missing-then-bulk-annotation-can-t-be-added-as-gives-error

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/populate_metadata.py
+++ b/components/tools/OmeroPy/src/omero/util/populate_metadata.py
@@ -589,17 +589,20 @@ class ParsingContext(object):
                 log.info('Missing well name column, skipping.')
 
             if image_name_column is not None:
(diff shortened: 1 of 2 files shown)


atarkowska:
--rebased-from #4852

PR #4872, opened 2016-09-29
Fix render py

joshmoore:
See https://github.com/jburel/openmicroscopy/pull/11

This should fix the currently failing metadata52 tests.

PR #5225, opened 2017-04-04
Fix render py (rebased onto metadata53)

atarkowska:
This is the same as gh-4872 but rebased onto metadata53.

----

See https://github.com/jburel/openmicroscopy/pull/11

This should fix the currently failing metadata52 tests.


atarkowska:
--rebased-from #4872

atarkowska:
replaced by https://github.com/openmicroscopy/openmicroscopy/pull/5239

PR #5226, opened 2017-04-04
Use jinja2 when populating metadata maps (rebased onto metadata53)

atarkowska:
This is the same as gh-4896 but rebased onto metadata53.

----

# What this PR does

The IDR bulk-annotation -> map-annotations process involves building URLs. Previously these were simple identifiers, but now they may include phrases including characters which need to be escaped when used in a URL. This PR now uses the Jinja templating module if it's installed, so standard Jinja2 filters such as `{{ value | urlencode }}` can be used.
# Testing this PR

Current use case: Mapr @aleksandra-tarkowska which creates mapr URLs with spaces.
# Potential issues

This adds a new optional dependency, `jinja2`, to OMERO.py. Should it be made mandatory along with yaml in order to keep things clearer?

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/metadata_utils.py
+++ b/components/tools/OmeroPy/src/omero/util/metadata_utils.py
@@ -30,6 +30,13 @@ from collections import deque
 from omero.constants import namespaces
 import re
 
+# TODO: Make jinja2 mandatory?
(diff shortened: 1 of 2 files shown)


atarkowska:
--rebased-from #4896

PR #4915, opened 2016-10-31
Create annotation links in batches if there are too many links

manics:
# What this PR does

#4775 added grouped-batch creation of MapAnnotations and links. This was required so that if a single MapAnnotation had multiple parents the MapAnnotation would not be incorrectly duplicated. However a small number of MapAnnotations have a very large number of parents which results in the Ice maximum message size being exceeded. This implements an additional method which creates the MapAnnotation and the links in separate steps if a batch size is exceeded.

# Testing this PR

Run `omero metadata populate` on a dataset which involves the creation of a `MapAnnotation` with an extremely large number of parents.

# Related reading

- https://trello.com/c/76es10Kg/23-rfe-populate-metadata-is-broken

PR #5227, opened 2017-04-04
Create annotation links in batches if there are too many links (rebased onto metadata53)

atarkowska:
This is the same as gh-4915 but rebased onto metadata53.

----

# What this PR does

#4775 added grouped-batch creation of MapAnnotations and links. This was required so that if a single MapAnnotation had multiple parents the MapAnnotation would not be incorrectly duplicated. However a small number of MapAnnotations have a very large number of parents which results in the Ice maximum message size being exceeded. This implements an additional method which creates the MapAnnotation and the links in separate steps if a batch size is exceeded.

# Testing this PR

Run `omero metadata populate` on a dataset which involves the creation of a `MapAnnotation` with an extremely large number of parents.

# Related reading

- https://trello.com/c/76es10Kg/23-rfe-populate-metadata-is-broken

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/populate_metadata.py
+++ b/components/tools/OmeroPy/src/omero/util/populate_metadata.py
@@ -647,7 +647,40 @@ class _QueryContext(object):
     def __init__(self, client):
         self.client = client
 
-    def projection(self, q, ids, ns=None):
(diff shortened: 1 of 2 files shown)


atarkowska:
--rebased-from #4915

PR #5228, opened 2017-04-04
Limit search results to the first 500 (rebased onto metadata53)

atarkowska:
This is the same as gh-4961 but rebased onto metadata53.

----

In order to protected the server from unnecessary
overhead, limit the number of search results to 500.
This can be improved by adding a per-call setting
and/or an overriding server-wide configuration.

cc: @aleksandra-tarkowska


atarkowska:
--rebased-from #4961

PR #4989, opened 2016-12-09
IDR: use new logo

joshmoore:
Place the logo in the repository to prevent it being lost between upgrades.

cc: @manics

PR #5230, opened 2017-04-04
IDR: use new logo (rebased onto metadata53)

atarkowska:
This is the same as gh-4989 but rebased onto metadata53.

----

Place the logo in the repository to prevent it being lost between upgrades.

cc: @manics


atarkowska:
--rebased-from #4989

manics:
--rebased-to #5763

PR #5207, opened 2017-03-29
allow creation of thumbnails in all-groups context

mtbc:
# What this PR does

Adjusts the heart of the server's thumbnail generation so that in a `omero.group: -1` context thumbnails can still be generated, by grouping the pixels objects by group and creating their thumbnails in the corresponding group context.

# Testing this PR

Experiment with the thumbnail service to ensure that this PR does not cause any regressions.

Check if the new https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ThumbnailStoreTest/testGetThumbnailsMultipleGroups/ test passes and looks to relate to what this PR does.

# Related reading

https://trello.com/c/bJqd3SkJ/39-getthumbnailbylongestsideset-doesn-t-accept-ctx

PR #5235, opened 2017-04-05
getThumbnails_test

will-moore:
# What this PR does

Adds a test based on testing of https://github.com/openmicroscopy/openmicroscopy/pull/5207

As seen in webclient and reproduced in Python, after 'Save to All' getting thumbnail set WITH
```omero.group: -1``` fails until we get a thumbnail set WITHOUT ```omero.group: -1``` 

Test reproduces this but uses ```with pytest.raises(omero.SecurityViolation):``` to wrap the failing line.
Should be removed when the bug is fixed.

# Testing this PR

1. Test should be green.

cc @mtbc


mtbc:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.test_thumbnailPerms/TestThumbnailPerms/testGetThumbnailSetAfterApplySettings/

(currently "Failed: DID NOT RAISE")

aleksandra-tarkowska:
--exclude (until Will is back)

will-moore:
@aleksandra-tarkowska Fixed test. Should be able to include this now, thanks.

jburel:
Removed the exclude flag

mtbc:
... and it's now passing. :+1:

mtbc:
--exclude for testing of #5249, #5250.

mtbc:
Want to try adding an integration test to reproduce issue at https://github.com/openmicroscopy/openmicroscopy/pull/5207#issuecomment-295653183 ?

will-moore:
@mtbc As far as I can see, getThumbnailSet() was failing after 'save to all' but the test I wrote for that is now passing. I'm not sure how to reproduce the latest failures, except maybe to simply add a large number of images to that same test? But if it's failing because "you can't getThumbnailSet() while thumbnails are being re-created", then we don't have a fix for that, right? So this would likely stay as a failing test which might not be very useful?
Also I could try add a failing test for shares if that would help fix the issue?

mtbc:
Yeah, doesn't sound like it's worth it at this time, does it? Thank you for your thoughts! If there aren't any *regressions* with shares then I think we're probably good unless IDR depends on them. At least, I'm unlikely in the near future to get to investigating handling in-progress thumbnails!

PR #5236, opened 2017-04-05
Shape Data

jburel:
# What this PR does

A comma was added after each pair of coordinate.
This is not compatible the ome-model see https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome.html

# Testing this PR
Units tests have been added to handle list of points with comma between each pair and list of points w/o comma after each pair
Make sure the tests are green

Save a polyline or polygon, check that the list of points is something lik ``1,2 4,5``

# Related reading

https://trello.com/c/8lGnMoIc/67-polyline-and-polygon-insight

### DIFF
--- a/components/blitz/src/omero/gateway/model/ShapeData.java
+++ b/components/blitz/src/omero/gateway/model/ShapeData.java
@@ -154,7 +154,7 @@ public abstract class ShapeData
         {
             if (i != 0)
             {
-                buf.append(", ");
+                buf.append(" ");
             }
             buf.append(toNumber(points[i].x));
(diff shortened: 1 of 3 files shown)


dominikl:
Looks good, unit test is passing, too.

joshmoore:
Is there a chance that such objects exist in the DB and if so, what will happen with them?

jburel:
They will exist in the DB when manipulated they will be fixed in the "correct" format, similar to what we have done previously.
Clients handle the legacy and "correct" format.

PR #5239, opened 2017-04-10
Fix render py (rebased onto metadata53) 

atarkowska:
This is the same as gh-4872 but rebased onto metadata53 and #4988 #5053 #5163

----

See https://github.com/jburel/openmicroscopy/pull/11

This should fix the currently failing metadata52 tests.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7613,7 +7613,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                 self._re = self._prepareRE(rdid=rdid)
             except omero.ValidationException:
                 logger.debug('on _prepareRE()', exc_info=True)
-                self._re = None
+                self._closeRE()
         return self._re is not None
 
     def resetRDefs(self):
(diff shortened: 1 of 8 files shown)


atarkowska:
--rebased-from #4872

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8192,8 +8243,9 @@ def getZoomLevelScaling(self):
             rv[i] = float(level)/sizeXList[0]
         return rv
 
+    @assert_re()
     def setActiveChannels(self, channels, windows=None, colors=None,
-                          reverseMaps=None):
+                          reverseMaps=None, noRE=False):
```
atarkowska, line 8248: @joshmoore conflict were on `noRE` as `reverseMaps` were added in develop. This is something to review

joshmoore:
Note: this minimally leaves one very common situation where`self.c` is nulled before `self.__del__` is called:

```
[jamoore@test45-omeroreadonly-1 log]$ grep _assert_unregistered OMEROweb.log | head -n 1
2017-12-14 17:58:45,963 WARNI [                           omero.gateway] (proc.24164) _assert_unregistered():1596 No services could be found: No session available
[jamoore@test45-omeroreadonly-1 log]$ grep _assert_unregistered OMEROweb.log | wc
  98762 1777623 19652522
```

PR #5242, opened 2017-04-10
expose enum constants in generated ome.model classes (second try)

mtbc:
# What this PR does

This PR exposes many enumeration values (at least the easier ones!) as string constants in OMERO.server. The idea is to eliminate bugs arising from server code having to copy literal values that then are not updated if the enumeration value is changed.

# Testing this PR

Take a look at how this PR changes the generated Java classes in `components/model/target/generated/src/ome/model/enums/`. This PR includes some changes to server code to illustrate with the `ChecksumAlgorithm` enumeration the use of these new constants. This PR should leave the server's actual behavior unchanged.

# Related reading

http://trac.openmicroscopy.org/ome/ticket/10691
supersedes #5206

### DIFF
--- a/components/dsl/resources/ome/dsl/object.vm
+++ b/components/dsl/resources/ome/dsl/object.vm
@@ -328,8 +328,8 @@ implements java.io.Serializable, IObject
     public final static String GROUP_FILTER = "${gfilter}";
     public final static String EVENT_FILTER = "${efilter}";
     public final static String PERMS_FILTER = "${pfilter}";
-#end
 
+#end
     public ${type.shortname} () {
         this(null, true);
     }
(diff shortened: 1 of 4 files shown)


jburel:
This time the enumerations are correct
e.g.
``
public final static String VALUE_BRIGHTFIELD = "Brightfield";
public final static String VALUE_PHASE = "Phase";
public final static String VALUE_DIC = "DIC";
public final static String VALUE_HOFFMAN_MODULATION = "HoffmanModulation";
``

PixelsType and Binning are now included. This is useful.
We will need a card so we do the "cleaning" in follow-up PRs

mtbc:
Created https://trello.com/c/w9GnPkQ7/72-tidy-up-enum-string-constants.

jburel:
Thanks Merging

PR #5252, opened 2017-04-17
Cors headers

will-moore:
# What this PR does

Adds CORS support to OMERO.web. See https://trello.com/c/HHVTF2l8/20-integration-cors-etc
This installs https://github.com/ottoyiu/django-cors-headers (added to requirements.txt).

We allow setting of origins that are authorised to make CORS requests:

```
$ bin/omero config append omero.web.cors_origin_whitelist '"localhost:8000"'
```
OR to allow ALL origins
```
$ bin/omero config set omero.web.cors_origin_allow_all True
```
By default these configs are empty/False, so that cors functionality is disabled by default.

# Testing this PR

1. Should be built and deployed by devspace at http://10.0.51.146/web/api/

2. Go to https://github.com/will-moore/openmicroscopy/blob/cors_headers/examples/Training/javascript/index.html save the raw html file locally then
```
cd path/to/dir/
```
or if you've checked out this branch:
```
$ cd openmicroscopy/examples/Training/javascript/
```
And then:
```
python -m SimpleHTTPServer
```
3. Go to http://localhost:8000/, open devtools and follow the Network tab to see what is happening.

4. Enter the url http://10.0.51.146/web/ and click "Connect" (this devspace deploy has been configured to accept ```"localhost:8000"``` as shown above)

5. This will GET available servers, CSRF token and login url and show a login screen.

6. Login uses POST (should see the pre-flight OPTIONS request) and then lists Projects.

7. Connecting from origins other than ```"localhost:8000"``` should fail.

screen shot 2017-04-20 at 16 27 58

screen shot 2017-04-20 at 16 28 58

# Related reading

This is built by devspace including this PR (and ignoring the exclude flag) using 
```
scc merge develop -Dnone -Icors --no-ask --reset --update-gitmodules
```
As things stand in this PR, we require install of ```django-cors-headers``` and it's configured in ```INSTALLED_APPS``` and in ```MIDDLEWARE_CLASSES```.
To enable it, you simply have to configure the allowed origins (see above).
An alternative is that we don't specify it in requirements.txt and users have to install it as needed.
This would also mean that they'd also have to configure ```INSTALLED_APPS``` and ```MIDDLEWARE_CLASSES``` (see @manics PR) but we'd have to be sure that ```'corsheaders.middleware.CorsMiddleware'``` comes *before* ```'django.middleware.csrf.CsrfViewMiddleware',``` (and others?) and ```'corsheaders.middleware.CorsPostCsrfMiddleware'``` comes at the end.

Thoughts?
cc @chris-allan @aleksandra-tarkowska 

Would be nice to support ```https``` at ```https://10.0.51.146/web/``` so that we can host the example on gh-pages which uses ```https```.
Could maybe use the same sslcert that is used by https://10.0.51.146:8443/ See:
https://github.com/openmicroscopy/devspace/blob/master/sslcert

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -692,6 +692,22 @@ CUSTOM_SETTINGS_MAPPINGS = {
           " 'webtest/webclient_plugins/center_plugin.overlay.js.html',"
           " 'channel_overlay_panel']``. "
           "The javascript loads data into ``$('#div_id')``.")],
+
+    # CORS
+    "omero.web.cors_origin_whitelist":
+        ["CORS_ORIGIN_WHITELIST",
+         '[]',
+         json.loads,
+         ("A list of origin hostnames that are authorized to make cross-site "
+          "HTTP requests."
+          "Used by the django-cors-headers app as described at "
+          "https://github.com/ottoyiu/django-cors-headers")],
+    "omero.web.cors_origin_allow_all":
+        ["CORS_ORIGIN_ALLOW_ALL",
+         "false",
+         parse_boolean,
+         ("If True, cors_origin_whitelist will not be used and all origins "
+          "will be authorized to make cross-site HTTP requests.")],
 }
 
 DEPRECATED_SETTINGS_MAPPINGS = {
@@ -1194,17 +1210,13 @@ MANAGERS = ADMINS  # from CUSTOM_SETTINGS_MAPPINGS  # noqa
 # omeroweb.connector.Connector object
 SESSION_SERIALIZER = 'django.contrib.sessions.serializers.PickleSerializer'
 
-# Or use CORS_ORIGIN_WHITELIST below
-CORS_ORIGIN_ALLOW_ALL = True
-
+# Configuration for django-cors-headers app
+# See https://github.com/ottoyiu/django-cors-headers
+# Configration of allowed origins is handled by custom settings above
 CORS_ALLOW_CREDENTIALS = True
-
 # Needed for Django <1.9 since CSRF_TRUSTED_ORIGINS not supported
 CORS_REPLACE_HTTPS_REFERER = True
 
-# CORS_ORIGIN_WHITELIST = (
-#     'localhost:8000',
-# )
 
 # Load server list and freeze
 from connector import Server
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -971,6 +971,7 @@ MIDDLEWARE_CLASSES = (
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
+    'corsheaders.middleware.CorsPostCsrfMiddleware',
 )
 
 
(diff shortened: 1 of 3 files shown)


manics:
FYI @aleksandra-tarkowska and I had a discussion about making `MIDDLEWARE_CLASSES` configurable, e.g. as a dictionary of `{"middleware.class.name": integer-priority}`

waxenegger:
The devspace install looks good. 

I tried running the example on a local apache and with 4 browsers and the (cors) request flow worked smoothly (incl. pre-flight for POST):
cors-debug-success

Next I changed my web server's port to standard 80 which was not in the whitelist and I received an error:
cors-failure

This is unrelated to the workings of your PR but I'm hoping that someone can explain to me the necessity of additionally setting X-CSRFToken. Now of course I know what the intention is (duh protection from cross domain forgery) but I fail to see what it adds security-wise. Sure it's not always sent like the cookie, yet It doesn't exactly fall into the category: https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29_Prevention_Cheat_Sheet#Double_Submit_Cookie either, especially not since it's kept in memory and reused which allows for better access than a httponly cookie actually. Maybe I'm missing something ...

will-moore:
I think the key is that the X-CSRFToken is not available to other scripts that your browser might be tricked into running. Scripts from other domains shouldn't be able to access the CSRF token (although requests that they make will have the csrf token and sessionId in the cookie data) so they will be able to GET but not POST.

chris-allan:
Additionally setting `X-CSRFToken`? Can you explain what you mean in addition to, @waxenegger?

The CSRF token needs to be passed by the requester somehow. Otherwise this defeats the purpose of CSRF protection entirely. This is usually either done via URL encoded arguments, post data or request header. As JavaScript framework libraries like jQuery usually have an infrastructure to set a request header on all requests that tends to be the preferred option. Some like Angular even have automatic tools for doing this. The cookie and corresponding request header that Angular uses is `XSRF-TOKEN` and `X-XSRF-TOKEN` respectively.

For reference:
 * https://docs.angularjs.org/api/ng/service/$http

waxenegger:
Well, I'm clear on the intentions of the token, I'm not so clear about achieving them. for that double submit mechanism to be tight it seems that the x-csrf header that is sent should be different with every post and not be reused. 

also there is an interesting comment here in regards to mima: 
https://docs.djangoproject.com/fr/1.11/_modules/django/middleware/csrf/ (after the line: if request.is_secure():)

either way, this is more of a general discussion and shouldn't be in the PR I think.

will-moore:
Since this seems to be OK so far, and tests are passing on devspace, I'll remove the ```exclude``` label.

manics:
Is the plan to definitely merge this into develop? If so I'll rebase it to metadata53, and rebase my middleware PR #5257 on top of this PR on both branches to avoid future merge conflicts.

will-moore:
It will be merged into develop in some form, but the question remains about whether we include the ```django-cors-headers``` app in the release (and in the middleware config) as in this PR or if we simply document the install and config steps (which will use the extra settings in this PR). If we don't include the app in the release then we won't need to change the middleware settings in this app, so there shouldn't be any conflict.
@chris-allan @jburel Any opinions?

jburel:
I will be more inclined to only document it in the 5.3 line and provide an example.

will-moore:
@manics Go ahead and rebase your PR - I can exclude this until that's merged and then we can see how painful / easy it is to install cors app (and middleware) using your new setting.

 --exclude

PR #5255, opened 2017-04-18
Tests ignore usergroup images

will-moore:
# What this PR does

Fixes python integration tests that fail when globally readable images are unexpectedly returned from cross-group queries.
See https://trello.com/c/nWvUzN5t/24-omeroweb-integration-test-failures

--roles

# Testing this PR

1. Tests linked from trello card above should be green.

cc @mtbc


mtbc:
Maybe still an issue: https://10.0.51.154:8443/job/OMERO-test-integration/26/testReport/OmeroWeb.test.integration.test_api_images/TestImages/test_dataset_images/

will-moore:
@mtbc Hope that fixes it?

mtbc:
:+1: Now the only failing test is the expected make movie script.

mtbc (approved):

jburel:
@will-moore: Does it mean that the API call could lead to some "strange" behaviour?

mtbc:
I think it's expected behavior: an "omero.group: -1" finds one everything one's allowed to read, and the confusion was that the tests were working with a new user in a new group and putting some data there then being surprised that an all-groups context query was also returning data from outside that new group. Specifying the group ID for the query does limit the results as one would have expected.

jburel:
Thanks for the clarification

will-moore:
We generally assume that there are no images in the "user" group (accessible to all users) so the test wasn't expecting to find these extra images for a new user.
Also in the real world, users would probably be confused if there were images in the "user" group and they showed up under ```/api/images/``` with ```group = -1```. But it's probably very rare for anyone to put images in "user" group (or otherwise make them globally readable) and also rare to query all images without filtering by group or Dataset, so I think we're OK.

joshmoore:
The under-specification of `user` is likely something we should at least come back to. Once something is available, someone somewhere will likely make use of it, and if that is breaking assumptions, we should be aware of it and perhaps warn.

mtbc:
It's not just the `user` group, it's world-readable groups in general, of which the CLI happily allows creation:

```
Added group public (id=53) with permissions rwr-r-
```

joshmoore:
Then add that as a second topic which needs rationalization.

PR #5257, opened 2017-04-19
Configure django middleware

manics:
# What this PR does

Makes the Django middleware classes configurable using a new property `omero.web.middleware`

# Testing this PR

This PR should have no change on existing configurations.

To test the new property add a new middleware class. For example, you can serve static files directly in Django using [whitenoise](https://devcenter.heroku.com/articles/django-assets#whitenoise) which should be production quality in contrast to the built-in development server:

    pip install whitenoise
    OMERO.web/bin/omero config append omero.web.middleware '{"index":0, "class":"whitenoise.middleware.WhiteNoiseMiddleware"}'
    OMERO.web/bin/omero web start

OMERO.web should be running on http://localhost:4080

# Related reading

- https://trello.com/c/cWcsQClq/42-public-url-filter-vv-post-urls
- https://github.com/openmicroscopy/openmicroscopy/pull/4876 for a PR with a similar configuration property
- openmicroscopy docker slack channel for discussion of another use case (horizontally scaling OMERO.web docker)
- this will conflict with https://github.com/openmicroscopy/openmicroscopy/pull/5252 if it is rebased @will-moore

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/utils.py
+++ b/components/tools/OmeroWeb/omeroweb/utils.py
@@ -43,3 +43,8 @@ def reverse_with_params(*args, **kwargs):
     if qs:
         url += '?' + urlencode(qs)
     return url
+
+
+def sort_properties_to_tuple(input_list, index="index", element="class"):
+    return tuple(e[element] for e in sorted(
+                 input_list, key=lambda k: k[index]))
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -373,6 +373,23 @@ CUSTOM_SETTINGS_MAPPINGS = {
         ["APPLICATION_SERVER_MAX_REQUESTS", 0, int,
          ("The maximum number of requests a worker will process before "
           "restarting.")],
+    "omero.web.middleware":
+        ["MIDDLEWARE_CLASSES_LIST",
+         ('['
+          '{"index": 1, '
+          '"class": "django.middleware.common.BrokenLinkEmailsMiddleware"},'
+          '{"index": 2, '
+          '"class": "django.middleware.common.CommonMiddleware"},'
+          '{"index": 3, '
+          '"class": "django.contrib.sessions.middleware.SessionMiddleware"},'
+          '{"index": 4, '
+          '"class": "django.middleware.csrf.CsrfViewMiddleware"},'
+          '{"index": 5, '
+          '"class": "django.contrib.messages.middleware.MessageMiddleware"}'
+          ']'),
+         json.loads,
+         ("The maximum number of requests a worker will process before "
+          "restarting.")],
     "omero.web.prefix":
         ["FORCE_SCRIPT_NAME",
          None,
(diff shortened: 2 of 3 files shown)


joshmoore:
Would a plugin ever need to add something to this list?

will-moore:
Code looks good.
Tested locally with the 3 commits rebased onto develop:
This worked fine:
```
$ omero config append omero.web.middleware '{"index": 0.5, "class": "corsheaders.middleware.CorsMiddleware"}'
```
resulting in MIDDLEWARE_CLASSES:
```
(u'django.middleware.common.BrokenLinkEmailsMiddleware', 
u'corsheaders.middleware.CorsMiddleware', 
u'django.middleware.common.CommonMiddleware', u'django.contrib.sessions.middleware.SessionMiddleware', 
u'django.middleware.csrf.CsrfViewMiddleware', 
u'django.contrib.messages.middleware.MessageMiddleware')
```
Duplicate indexes are possible, but are handled fine.
Good to merge.

components/tools/OmeroWeb/omeroweb/utils.py
```diff
@@ -43,3 +43,8 @@ def reverse_with_params(*args, **kwargs):
     if qs:
         url += '?' + urlencode(qs)
     return url
+
+
+def sort_properties_to_tuple(input_list, index="index", element="class"):
+    return tuple(e[element] for e in sorted(
```
atarkowska, line 49: `str(e[element])` ?

joshmoore:
And I'd re-iterate my question about [plugins](https://github.com/openmicroscopy/openmicroscopy/pull/5257#issuecomment-295269962) -- i.e. to what extent does this need to be extensible so that plugins needing middleware can take part in this mechanism without needing to rewrite the entire list (a la top_links)?

components/tools/OmeroWeb/omeroweb/utils.py
manics: It looks like unicode works fine.

manics:
@joshmoore What functionality would that require that isn't handled by `index: FLOAT`?

joshmoore:
Sorry, I get it now. It was the use of append that I was phasing out on.

What happens if the same middle-ware gets added twice at different priorities?

Would rough guidelines on what various ranges of floats might mean be possible with the existing middleware that's out there. e.g. priorities like `VERY_HIGH`, `HIGH`, ... `LOW`, `VERY_LOW`? Does anyone know how sensitive the ordering is?

atarkowska:
@manics maybe just lets do `set`, first come first win

manics:
I think ordering is important for some modules (maybe critical if it's security related?).

will-moore:
I'm not sure we should try to over-engineer this to try and catch all kinds of invalid inputs (duplicates, invalid index etc) or try to juggle the order ourselves (assign meaning to index ranges - what's the difference between ```HIGH``` and ```VERY HIGH```.
If a user is following instructions such as https://github.com/ottoyiu/django-cors-headers where the order is critical, they just need to do ```$ bin/omero config get``` to see what's already there, then add the middleware accordingly.
There's a ton of ways to screw up this config so that the middleware doesn't work or django won't even start (e.g. Import error if you didn't install the app that you add) and we can't try to catch them all.

joshmoore:
I guess I'm most concerned with a plugin author stating in his/her README:

```
OMERO.web/bin/omero config append omero.web.middleware '{"index":0, "class":"...something..."}'
```

And having that break a production system. If we knew that a plugin should *never* be able to inject any middleware above/below some other middleware, then I think encoding (or at least *explaining*) that would be worthwhile.

manics:
I imagine any piece of middleware has the potential to break a system regardless of position.

joshmoore:
Certainly, but I think we have to provide what assistance we can, i.e. if possible to design this such that it's safer, I think it's worth considering.

manics:
I think the advice for plugin authors should be don't touch middleware, it's for sys-admins only.

manics:
--rebased-to #5269

joshmoore:
@manics : do you want this merged as is, or deal with the conflicts before?

manics:
My understanding is that the rebased version on develop won't conflict with https://github.com/openmicroscopy/openmicroscopy/pull/5252#issuecomment-298319006, and the other conflict with https://github.com/openmicroscopy/openmicroscopy/pull/5158 can only be dealt with after it's merged. OTOH I'm not going to be using this functionality on `metadata53` for the next couple of weeks so if you prefer to hold off and handle all PRs together that's fine with me.

joshmoore:
Ok. If this isn't targeted for 3.7, then I'll leave ON_HOLD and we can address for either 0.3.8 or 0.4.0

joshmoore:
Merging and we can get this into the next 0.4.0-* builds.

PR #5259, opened 2017-04-24
for light admin restrictions, clean up server code and improve performance

mtbc:
# What this PR does

Adjusts the server component to improve performance without changing behavior. Brings sudoer and current light admin privileges into `EventContext` and `CurrentDetails`.

# Testing this PR

Check with https://10.0.51.154:8443/job/OMERO-test-integration/ that there are no new test failures and that the test running time is improved.

# Related reading

https://trello.com/c/ToMbDClf/26-improve-performance

### DIFF
--- a/components/blitz/resources/omero/System.ice
+++ b/components/blitz/resources/omero/System.ice
@@ -6,9 +6,8 @@
 #ifndef OMERO_SYSTEM_ICE
 #define OMERO_SYSTEM_ICE
 
-#include <omero/RTypes.ice>
 #include <omero/ModelF.ice>
-#include <Ice/BuiltinSequences.ice>
+#include <omero/SystemF.ice>
 
 /*
  * The omero::system module combines the ome.system and ome.parameters
@@ -17,40 +16,6 @@
  */
 module omero {
   module sys {
-
-    // START: TRANSFERRED FROM COLLECTIONS
-
-    /*
-     * Some collections were initially defined under omero::sys
-     */
-
-        ["java:type:java.util.ArrayList<Long>:java.util.List<Long>"]
-            sequence<long> LongList;
-
-        ["java:type:java.util.ArrayList<Integer>:java.util.List<Integer>"]
-            sequence<int> IntList;
-
-        ["java:type:java.util.ArrayList<String>:java.util.List<String>"]
-            sequence<string> StringSet;
-
-        ["java:type:java.util.HashMap<Long,Long>:java.util.Map<Long,Long>"]
-            dictionary<long, long> CountMap;
-
-        /**
-         * ParamMap replaces the ome.parameters.QueryParam
-         * type, since the use of varargs is not possible.
-         **/
-        ["java:type:java.util.HashMap"]
-            dictionary<string,omero::RType> ParamMap;
-
-        /**
-         * IdByteMap is used by the ThumbnailService for the multiple thumbnail
-         * retrieval methods.
-         **/
-        ["java:type:java.util.HashMap"]
-            dictionary<long,Ice::ByteSeq> IdByteMap;
-
-    // END: TRANSFERRED FROM COLLECTIONS
 
     /**
      * Maps the ome.system.EventContext interface. Represents the
--- a/components/blitz/resources/omero/System.ice
+++ b/components/blitz/resources/omero/System.ice
@@ -8,6 +8,7 @@
 
 #include <omero/ModelF.ice>
 #include <omero/SystemF.ice>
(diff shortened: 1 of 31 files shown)


mtbc:
Ha, well so far we may have a speed-up in excess of 1%!

mtbc:
Performance improved but broke `test_ishare`, will investigate.

mtbc:
The improvement isn't much, admittedly.

components/blitz/resources/omero/System.ice
```diff
@@ -33,6 +30,9 @@ module omero {
         ["java:type:java.util.ArrayList<Integer>:java.util.List<Integer>"]
             sequence<int> IntList;
 
+        ["java:type:java.util.ArrayList<String>:java.util.List<String>"]
+            sequence<string> StringSet;
```
joshmoore: `omero/Collections.ice:            sequence<string> StringSet;` is inappropriate here?

mtbc: I think so: there's a circular import, where Collections uses System.

mtbc:
To do: do *not* propagate light admin privileges into sudo session.

components/blitz/resources/omero/System.ice
joshmoore: That's unfortunate. Looks like I made some improvement on this once:

```
commit 830b799130acde2f92a908f59526cb8a84a07918
Author: jmoore <josh@glencoesoftware.com>
Date:   Fri Nov 18 15:27:53 2011 +0100

    Re-arrange System/Collections Ice imports

    3.4 for some reason is more strict on the order of imports.
    This is likely caused by package-based import cycles, but it's
    unclear.
```

Looks like the current cause is the need to have `EventContext` in Collections. One option would be a `SystemF.ice` like `ModelF.ice`. I can try to open a clean up PR on your branch after this is merged if you'd prefer.

joshmoore:
The changes look good to me, esp. in the sense that fields are in the places that I would have naturally gone looking for them. It also generally looks like LOC is decreasing. I _haven't_ yet gone back and reviewed the state of this PR against `5_3`. It would be good if you could suggest a sensible time point for doing that, @mtbc.

Taking back my ":+1:" since

> To do: do not propagate light admin privileges into sudo session.

came in while I was editing.

mtbc:
Thank you! Arose from a conversation I just had with @pwalczysko. At least it's only the following commits that will require review, shouldn't be much. :smiley:

components/blitz/resources/omero/System.ice
mtbc: That would be great, thank you. (Or if before merging then I'm happy to cherry-pick.)

joshmoore: See 40af1cc4777903951b9a05aa0c5105defb4360f1:

```
commit 40af1cc4777903951b9a05aa0c5105defb4360f1
Author: jmoore <josh@glencoesoftware.com>
Date:   Mon May 1 16:39:45 2017 +0200

    Collections.ice: refactor import of System
```

from my roles branch if you'd like.

joshmoore:
Sorry, @mtbc, I was unclear (and incomplete). This would require:

```
commit 5ec92824f142a6ba08f013d0deec9ca4e296e6b7
Author: jmoore <josh@glencoesoftware.com>
Date:   Tue May 2 11:40:09 2017 +0200

    Remove duplicate StringSet
```

to completely revert.

pwalczysko:
testChown and testChgrp are passing now. https://10.0.51.154:8443/job/OMERO-test-integration/39/testngreports/integration/LightAdminRolesTest/

pwalczysko:
Checked manually that Delete Data permission given in Ui is not pervasive enough to allow deletion of a third-party owned data after sudoed in as a user.

pwalczysko:
Upload Script permission does not pervade through Sudo either.

pwalczysko:
WriteData given does not seem to be pervasive through Sudo, as Editing of thirs-party dataset after sudoing as a user fails.

pwalczysko:
I will see how to put the tests https://github.com/openmicroscopy/openmicroscopy/pull/5259#issuecomment-298605568, https://github.com/openmicroscopy/openmicroscopy/pull/5259#issuecomment-298608086, and https://github.com/openmicroscopy/openmicroscopy/pull/5259#issuecomment-298608772 into the back-end java tests. 

But as they passed, I am happy with the state of sudo here. 
Ready to merge fmpov.

joshmoore:
WFM. Merging.

PR #5260, opened 2017-04-25
Pixels enum

jburel:
# What this PR does

Use the enum value

# Testing this PR

Check that the build is green and the tests remain green

# Related reading

https://trello.com/c/w9GnPkQ7/72-tidy-up-enum-string-constants

cc @mtbc 

first round of cleaning

### DIFF
--- a/components/blitz/src/loci/ome/io/OmeroReader.java
+++ b/components/blitz/src/loci/ome/io/OmeroReader.java
@@ -369,7 +369,7 @@ public class OmeroReader extends FormatReader {
             m.sizeT = sizeT;
             m.rgb = false;
             m.littleEndian = false;
-            m.dimensionOrder = ome.model.enums.DimensionOrder.VALUE_XYZCT;
+            m.dimensionOrder = omero.model.enums.DimensionOrderXYZCT.value;
             m.imageCount = sizeZ * sizeC * sizeT;
             m.pixelType = FormatTools.pixelTypeFromString(pixelType);
 
--- a/components/blitz/src/loci/ome/io/OmeroReader.java
+++ b/components/blitz/src/loci/ome/io/OmeroReader.java
@@ -369,7 +369,7 @@ public class OmeroReader extends FormatReader {
             m.sizeT = sizeT;
             m.rgb = false;
             m.littleEndian = false;
-            m.dimensionOrder = "XYZCT";
+            m.dimensionOrder = ome.model.enums.DimensionOrder.VALUE_XYZCT;
             m.imageCount = sizeZ * sizeC * sizeT;
             m.pixelType = FormatTools.pixelTypeFromString(pixelType);
 
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -5695,7 +5695,7 @@ public class OMEROMetadataStoreClient
         // We're always the same dimension order in the server; force it to
         // "XYZCT" (ticket:3124, ticket:3718, ticket:3668)
         o.setDimensionOrder((DimensionOrder) getEnumeration(
-            DimensionOrder.class, ome.model.enums.DimensionOrder.VALUE_XYZCT));
+            DimensionOrder.class, omero.model.enums.DimensionOrderXYZCT.value));
     }
 
     /* (non-Javadoc)
--- a/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
+++ b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
@@ -5695,7 +5695,7 @@ public class OMEROMetadataStoreClient
         // We're always the same dimension order in the server; force it to
         // "XYZCT" (ticket:3124, ticket:3718, ticket:3668)
         o.setDimensionOrder((DimensionOrder) getEnumeration(
-            DimensionOrder.class, "XYZCT"));
+            DimensionOrder.class, ome.model.enums.DimensionOrder.VALUE_XYZCT));
     }
 
     /* (non-Javadoc)
--- a/components/blitz/src/ome/formats/importer/ImportSize.java
+++ b/components/blitz/src/ome/formats/importer/ImportSize.java
@@ -1,6 +1,4 @@
 /*
- *   $Id$
- *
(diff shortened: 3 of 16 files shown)


joshmoore:
I did a first quick read-through looking for copy-n-paste errors without seeing any.

components/server/test/ome/server/itests/pixeldata/PersistentEventLogLoaderTest.java
```diff
@@ -273,9 +274,9 @@ protected Pixels pixels() {
         pixels.setSizeZ(1);
         pixels.setSizeT(1);
         pixels.setSizeC(1);
-        pixels.setSha1("UNKNOWN");
-        pixels.setPixelsType(new PixelsType("int8"));
-        pixels.setDimensionOrder(new DimensionOrder("XYZCT"));
+        pixels.setSha1(ChecksumAlgorithm.VALUE_ADLER_32);
```
joshmoore: Is this what you intended? `UNKNOWN` --> `Adler-32`?

jburel: it is, I assume that it was similar to "pending..." we have in various locations
Happy to revert if we want to test the insert of such value

joshmoore: It is like "pending..." (and therefore needs to be cleaned up / rationalized) but I think changing it to another, unexpected value ("Adler-32") is probably out of scope of this PR.

jburel: The adler-32 is due to code completion in eclipse :-) I will pick another value

components/tools/OmeroJava/test/integration/ModelMockFactory.java
```diff
@@ -86,16 +86,16 @@
     static int SIZE_T = 10;
 
     /** The dimension order for the pixels type. */
-    static String XYZCT = "XYZCT";
+    static String XYZCT = ome.model.enums.DimensionOrder.VALUE_XYZCT;
```
mtbc: These classes should already have access to constants like `omero.model.enums.DimensionOrderXYZCT.value`.

jburel:
@mtbc should be sorted now

components/blitz/src/loci/ome/io/OmeroReader.java
```diff
@@ -369,7 +369,7 @@ else if (key.equals("iid")) {
             m.sizeT = sizeT;
             m.rgb = false;
             m.littleEndian = false;
-            m.dimensionOrder = "XYZCT";
+            m.dimensionOrder = ome.model.enums.DimensionOrder.VALUE_XYZCT;
```
mtbc: should be able to use `omero.model.*` here

components/blitz/src/ome/formats/OMEROMetadataStoreClient.java
```diff
@@ -5695,7 +5695,7 @@ public void setPixelsDimensionOrder(
         // We're always the same dimension order in the server; force it to
         // "XYZCT" (ticket:3124, ticket:3718, ticket:3668)
         o.setDimensionOrder((DimensionOrder) getEnumeration(
-            DimensionOrder.class, "XYZCT"));
+            DimensionOrder.class, ome.model.enums.DimensionOrder.VALUE_XYZCT));
```
mtbc: should be able to use `omero.model.*` here

components/blitz/src/ome/formats/importer/ImportSize.java
```diff
@@ -81,17 +81,17 @@ public ImportSize(String fileName, Pixels pixels, String dimOrder) {
     }
 
     private int getSequenceNumber(String dimOrder) {
-        if (dimOrder.equals("XYZTC"))
+        if (ome.model.enums.DimensionOrder.VALUE_XYZTC.equals(dimOrder))
```
mtbc: should be able to use `omero.model.*` here

mtbc:
Looking through, I think there are still a fair few places where the more user-familiar `omero.model.*` are available, even beyond where I commented.

components/rendering/src/omeis/providers/re/Renderer.java
```diff
@@ -82,18 +83,17 @@
...
+    public static final String MODEL_RGB = RenderingModel.VALUE_RGB;
 
     /** Identifies the type used to store model values. */
     public static final String MODEL_HSB = MODEL_RGB;
 
     /** Identifies the type used to store photometric interpretation values. */
-    public static final String PHOTOMETRIC_MONOCHROME = "Monochrome";
-
-    //static final String RGB_COLOR_DOMAIN = "RGB";
+    @Deprecated
```
mtbc, line 95: :+1: for deprecations

components/server/test/ome/server/itests/pixeldata/PersistentEventLogLoaderTest.java
```diff
@@ -273,9 +274,9 @@ protected Pixels pixels() {
         pixels.setSizeZ(1);
         pixels.setSizeT(1);
         pixels.setSizeC(1);
-        pixels.setSha1("UNKNOWN");
-        pixels.setPixelsType(new PixelsType("int8"));
-        pixels.setDimensionOrder(new DimensionOrder("XYZCT"));
+        pixels.setSha1(ChecksumAlgorithm.VALUE_SHA1_160);
```
mtbc: might want to leave as unknown unless the hash matches SHA1?

jburel:
updated

mtbc:
Looks good though still curious about https://github.com/openmicroscopy/openmicroscopy/pull/5260#discussion_r113399024.

components/server/test/ome/server/itests/pixeldata/PersistentEventLogLoaderTest.java
jburel: it is only used here as an entry in the DB

joshmoore: Ah, ok. But if so, why not keep it as "UNKNOWN" or "TestValue" if you prefer?

mtbc (approved):

PR #5261, opened 2017-04-25
Enum python

jburel:
# What this PR does

Use constants from enum

# Testing this PR

Check that the build is green and the tests are green

# Related reading

https://trello.com/c/w9GnPkQ7/72-tidy-up-enum-string-constants
Related to https://github.com/openmicroscopy/openmicroscopy/pull/5260

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/script_utils.py
+++ b/components/tools/OmeroPy/src/omero/util/script_utils.py
@@ -34,6 +34,7 @@ from os.path import exists
 import omero.clients
 from omero.rtypes import unwrap
 from omero.model.enums import PixelsTypeint8, PixelsTypeuint8
+from omero.model.enums import PixelsTypefloat
 import omero.util.pixelstypetopython as pixelstypetopython
 
 try:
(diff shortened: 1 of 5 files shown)


mtbc:
Should `omero.model.enums.*` be used where the Blitz API is available?

jburel:
probably
still few things to check, wanted to check the full test suite

jburel:
synchonized value for sha1 with https://github.com/openmicroscopy/openmicroscopy/pull/5260

components/tools/OmeroPy/src/omero/util/script_utils.py
```diff
@@ -30,8 +30,11 @@
 from numpy import add, array, asarray, fromstring, reshape, zeros
 from os.path import exists
 
+
```
mtbc: These *double*-blanks are interesting, are they now a standard way we separate our-code imports?

jburel: this might have been introduced when moving import around
I will remove it

mtbc: :+1: I think there was at least another double-blank introduced. Apart from those, this looks good, we'll see when CI is green.

mtbc (approved):

PR #5263, opened 2017-04-26
Import into other groups

mtbc:
# What this PR does

This PR allows full administrators, or sufficiently endowed light administrators, to import images into any group regardless of membership. More precisely, it stabilizes the test suite from the previous attempt which caused various intermittent errors like, `ome.conditions.SecurityViolation: User 318 is not a member of group 276 and cannot login`.

# Testing this PR

https://10.0.51.154:8443/job/OMERO-test-integration/ should be fairly blue. Note the included integration test and that, with the same server commit in #4957 and #5258, CI on either side has now not shown that error for some time.

# Related reading

https://trello.com/c/RQLsMkEV/14-fix-import-into-other-groups

PR #5268, opened 2017-04-29
add deprecation notes for coming 5.4 changes

mtbc:
# What this PR does

Deprecates methods that are removed in #5141 and #5263.

# Testing this PR

CI should remain green.

# Related reading

https://github.com/openmicroscopy/openmicroscopy/pull/5263#issuecomment-298025198

### DIFF
--- a/components/blitz/src/ome/services/blitz/fire/Registry.java
+++ b/components/blitz/src/ome/services/blitz/fire/Registry.java
@@ -62,7 +62,7 @@ public interface Registry {
      * interval     := Seconds between retries
      * client_uuid  := Uuid of the client which should be used
      * </pre>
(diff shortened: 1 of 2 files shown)


components/blitz/src/ome/services/blitz/fire/Registry.java
```diff
@@ -62,7 +62,9 @@
      * interval     := Seconds between retries
      * client_uuid  := Uuid of the client which should be used
      * </pre>
+     * @deprecated in OMERO 5.4 uses group ID instead of group name
```
jburel: the way it is displayed can be a bit confusing. "deprecated in OMERO 5.4".
Maybe better to have ``In`` upper case

jburel:
Thanks for the adjustment

PR #5272, opened 2017-05-03
remove admin privileges bean from classes that no longer use it

mtbc:
# What this PR does

Cleans up passing beans around: some classes no longer need the admin privileges bean so we can remove it from their code.

# Testing this PR

CI should remain unchanged.

# Related reading

https://trello.com/c/6TYMHVJj


dominikl:
Looks alright, no CI issues reported. Always good to remove unnecessary code 👍

PR #5274, opened 2017-05-03
histogram_json 404

will-moore:
# What this PR does

Correctly returns 404 if image is not found for webgateway/histogram_json/ instead of an exception.
See https://trello.com/c/DCzm4guz/74-histogram-json-handle-image-404

# Testing this PR

1.Go to ```/webgateway/histogram_json/{image_id}/channel/0/``` for a valid and invalid ```image_id```.

2. Should get 404 for invalid image.


waxenegger:
what are the odds of typing in a 5 digit random number (logged in as user-1) and hitting a valid image id....I should play lotto .... next I try 1 and lo and behold ... there is no such image id.

to stay on topic: it works and I get the 404

PR #5277, opened 2017-05-03
canAnnotate comment fix

will-moore:
# What this PR does

Fixes use of 'canAnnotate()' for Comments in right panel.
Should not allow group owner to annotate others' data in a private group.

# Testing this PR
Test that: 
 - Group owner can't add Comments to others' data in private group
 - This should be true when single object selected OR multiple (batch annotate).


waxenegger:
Works for both single and batch selection: 
commen-can-annotate

PR #5278, opened 2017-05-04
Update Slack section to notify only when the success status is changed

sbesson:
See https://docs.travis-ci.com/user/notifications/#Configuring-slack-notifications
this should reduce the level of notifications (currently set to always)
/cc @joshmoore @jburel


joshmoore:
Merging as green

sbesson:
--rebased-to #5291

PR #5281, opened 2017-05-05
cache OMERO sessions in level 2 Hibernate cache

mtbc:
# What this PR does

Adjusts HQL queries of sessions in performance bottlenecks to enable caching.

# Testing this PR

The server, especially integration tests, should in general run a little faster.

# Related reading

https://trello.com/c/tYg2svfX/

### DIFF
--- a/components/server/src/ome/api/local/LocalQuery.java
+++ b/components/server/src/ome/api/local/LocalQuery.java
@@ -1,7 +1,5 @@
 /*
- * ome.api.local.LocalQuery
- *
- *   Copyright 2006 University of Dundee. All rights reserved.
+ *   Copyright 2006-2017 University of Dundee. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  */
 
@@ -9,6 +7,11 @@ package ome.api.local;
 
 import org.springframework.orm.hibernate3.HibernateCallback;
 
+import ome.annotations.NotNull;
+import ome.conditions.ValidationException;
+import ome.model.IObject;
+import ome.parameters.Parameters;
+import ome.parameters.QueryParameter;
 import ome.services.query.Query;
 
 /**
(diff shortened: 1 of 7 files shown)


components/server/src/ome/security/SecuritySystem.java
```diff
@@ -119,10 +114,8 @@
     /**
      * Prepares the current {@link EventContext} instance with the current
      * {@link Principal}. An exception is thrown if there is none.
-     * 
-     * @param isReadOnly
      */
-    void loadEventContext(boolean isReadOnly);
```
mtbc: If merged then this would be a 5.3 deprecation.

jburel: Do you want to open a 5.3 PR and mark it as deprecated?
so we can/cannot remove it later on depending on outcome

joshmoore:
I may have to go back and review the reasoning, but I'm remembering there being a reason (around stateful sessions or filters, etc) that we hid Hibernate's `Session` instance. There's also the `LocalQuery` service for nice server-internal items. Would it be an option to have a `findByCachedQuery` (or similar) there?

mtbc:
Add an API method to `LocalQuery` specifically? I could give it a try.

joshmoore:
This looks nice and simple! One research-y question (if you happened to look into it while preparing this): is there anything we should take into account when it comes to this API in terms permitting external users to _also_ enable the cache for HQL queries? My hope was always that we could enable _all_ queries (but I don't know what the impact of that would be)

mtbc:
Good question @joshmoore. I'd be hesitant to offer clients the choice of caching: the impression from my reading is that the query cache often does more harm than good (overhead in checking, invalidating, etc.), especially for data that might change, so the choice probably demands more subtlety and testing than we might expect from people writing scripts, etc., especially as this could impact other users of their server. I'm not even wholly convinced about this case here: VisualVM certainly suggests that this PR is a clear win in the expected places, but in my more real-world testing I don't see anything compelling either way emerging from the noise of varying times. In short: I doubt it's promising enough to be worth investigation / complication.

Could well be worth revisiting once we're using Hibernate 5: of course I've largely been reading Hibernate 3 stuff.

PR #5284, opened 2017-05-09
more carefully prevent update/delete for user group

mtbc:
# What this PR does

Makes the ACL checks for update/delete wrt the `user` group clearer and safer.

# Testing this PR

https://10.0.51.154:8443/job/OMERO-test-integration/testngreports/ should have no skips or failures: @pwalczysko, this PR should fix *all* cases of `testOfficialScriptDeleteNoSudo` if you want to try again removing the skip. (Can be tested locally if need be.)

The test in the description of https://github.com/openmicroscopy/omero_private/issues/13 should fail. (Source may need a bit of tweaking for current branch, but not much: `PYTHON_MIMETYPE == "text/x-python"` and `roles` is now already a field in the superclass.)

### DIFF
--- a/components/server/src/ome/security/basic/BasicACLVoter.java
+++ b/components/server/src/ome/security/basic/BasicACLVoter.java
@@ -458,15 +458,25 @@ public class BasicACLVoter implements ACLVoter {
             return rv; // EARLY EXIT!
         }
 
-        Permissions grpPermissions = c.getCurrentGroupPermissions();
-        if (grpPermissions == null || grpPermissions == Permissions.DUMMY) {
-            if (d.getGroup() != null) {
-                Long gid = d.getGroup().getId();
+        Permissions grpPermissions = null;
+        if (d.getGroup() != null) {
+            /* got a group set so review its permissions */
+            final Long gid = d.getGroup().getId();
+            if (roles.getUserGroupId() == gid) {
+                /* special handling for user group permissions */
+                grpPermissions = new Permissions(Permissions.EMPTY);
+            } else {
+                /* not user group so use group's permissions */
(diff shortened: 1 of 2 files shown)


joshmoore:
In general, this code block has long caused issues. I'm slightly nervous about having something as specific as `"Directory"` encoded here but don't have any immediate suggestions on how to do it differently.

mtbc:
It might be that I should create a DevX card for simpllifying this block, seeing which tests break, and justifying to ourselves that those tests aren't at fault. As things are, as with the graphs work, a facet of my initial approach to implementing light admin restrictions is to try to avoid unnecessary changes in behavior, so I am trying not to tweak existing tests.

Speculating wildly: it is also possible that there is actually some `canAnnotate` of related objects or something that this directory-checking is a poor emulation of, cf. `RepositoryDaoImpl.canWriteParentDirectory`, i.e. that the check in the ACL voter should ultimately be approaching the asked question quite differently.

pwalczysko:
tested locally  with this PR  in (scc merge roles). 

Testing the test in header of https://github.com/openmicroscopy/omero_private/issues/13 ended with failure as expected.

Adjusting the test according to https://github.com/openmicroscopy/omero_private/pull/15#issuecomment-283307652 fixed the failing test into a passing test.

The removal of the Skp lines in https://github.com/pwalczysko/openmicroscopy/blob/new-role-cleaned/components/tools/OmeroJava/test/integration/LightAdminRolesTest.java#L1577 . The skipped tests passed as well as expected.

Ready to merge RMPOV

mtbc:
Created https://trello.com/c/PAg0H0wB/756-review-group-permissions-in-server to note ongoing uncertainty.

PR #5286, opened 2017-05-09
Remove scc travis-merge step from the Travis build (rebased onto roles)

sbesson:
This is the same as gh-5285 but rebased onto roles.

----

Following https://blog.travis-ci.com/2017-05-08-security-advisory, the public token used to merge dependent PRs has been revoked. To fix builds in the meantime, this commit removes the set up and usage of scc travis-merge which implies every PR will now be tested in isolation via Travis CI.
/cc @mtbc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,11 +32,7 @@ env:
     - BUILD="build-java"
 
 before_install:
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #5285

PR #5291, opened 2017-05-10
Update Slack section to notify only when the success status is changed (rebased onto roles)

sbesson:
This is the same as gh-5278 but rebased onto roles.

----

See https://docs.travis-ci.com/user/notifications/#Configuring-slack-notifications
this should reduce the level of notifications (currently set to always)
/cc @joshmoore @jburel


sbesson:
--rebased-from #5278

joshmoore:
Green (though no reason to expect otherwise). Merging.

PR #5292, opened 2017-05-10
Fix class cast exception

jburel:
# What this PR does

Fix class cast exception

# Testing this PR
Go to the images or file tab
And refresh

# Related reading
https://trello.com/c/AV11o7Tu/70-insight-bugs

cc @dominikl


dominikl:
Thanks @jburel . That certainly fixes the `ClassCastException`, I just can't replicate the original bug. Tried with latest build, went to Images/File tabs, refresh, didn't get the exception.

jburel:
I struggled to see how it happens.
Since it is maintenance mode, I went for the "obvious" fix.

dominikl:
👍  Good to merge anyway!

PR #5295, opened 2017-05-12
Make Job no longer a system type for OMERO

mtbc:
# What this PR does

Has the security system no longer regard `Job` as being a system type.

# Testing this PR

CI should pass.

# Related reading

https://trello.com/c/aC6uk5Vy/489-job-as-system-type

### DIFF
--- a/components/server/src/ome/security/SystemTypes.java
+++ b/components/server/src/ome/security/SystemTypes.java
@@ -1,6 +1,4 @@
 /*
- *   $Id$
- *
  *   Copyright 2008 Glencoe Software, Inc. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  */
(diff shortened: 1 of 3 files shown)


joshmoore:
To summarize / refresh my memory: [jobs.ome.xml](https://github.com/mtbc/openmicroscopy/blob/89a082005ee8072265647b83a167e3af02511e1e/components/model/resources/mappings/jobs.ome.xml#L24 doesn't list jobs as a systems type but the ACL tries to treat it as one. 

It looks like this might have been some share-related workaround: https://github.com/mtbc/openmicroscopy/commit/26c97105e894599478228c0572c4dcf0735a56aa -- that being said, much has changed so still :thumbsup:

The only real consequence of this change is that users can now modify their own `Job` records, which was never intended. That being said, I don't know of anything security-related which depends on the integrity of the `Job` rows. It is more the principle of protecting provenance. My assumption, however, is that we are more likely to drop `Job` completely in favor of something more flexible and so the point is likely moot.

Assuming the above doesn't raise any flags for anyone, ok to merge.

mtbc:
Thank you! Relisted to have someone test that shares still seem to work as, judging by the need for 5.3.1, our automated testing of them is light.

pwalczysko:
Shares work fine. Tried also inverted and LUT. All as expected.

mtbc:
Thank you @pwalczysko, I appreciate that extra reassurance!

PR #5296, opened 2017-05-15
provide helper for Boolean test case combinations

mtbc:
# What this PR does

Adds to the codebase a convenient helper function that I use in initially developing parameterized Java tests under TestNG. Easily extended to provide for desired number of arguments. Could be consumed by other providers that filter out unwanted cases or add extra non-boolean arguments.

It would be possible to write something more automatic and featureful -- reflecting on the target method arguments and filling in boxed Booleans with a null too, enumerations with all values, etc. -- but probably not worth the effort at this time.

# Testing this PR

Could try using one of the new providers in one's own code. For example, add to an OmeroJava test class a new method,
```java
    @Test(dataProvider = "test cases using three Boolean arguments")
    public void test(boolean x, boolean y, boolean z) {
        System.err.println(Joiner.on(", ").join(x, y, z));
    }
```
and see if all argument combinations are exercised.


pwalczysko:

PR #5298, opened 2017-05-16
bump Ehcache version to 2.6.2

mtbc:
# What this PR does

Bumps Ehcache version up to 2.6.2.

# Testing this PR

Make sure that the rendering control in Insight still works and that it does not log new scary errors.

# Related reading

https://trello.com/c/4MM44jrz/593-ehcache


jburel:
FYI: insight does not use the cache anymore but no need to remove it

mtbc:
Ta. Guess it can go with the rest of Insight someday. :smiley:

mtbc:
(I glanced through `config/ehcache.xml` and didn't notice anything that obviously needed updating.)

joshmoore:
Apologies for not having commented on this yesterday. A few minor points:

 * In general, I don't consider a minor bump to a dependency to necessarily be breaking. _Certainly_ one that doesn't feature prominently feature in our API.
 * Our previous use of ehcache was quite limited (and in memory!) so I don't suspect there will be any issues.
 * *If* anyone is likely to have problems though, it would be Hibernate and Spring.

mtbc:
https://github.com/openmicroscopy/openmicroscopy/blob/f81147e0/components/insight/SRC/org/openmicroscopy/shoola/env/rnd/RenderingControlProxy.java#L593 is why I thought it worth giving the rendering control a quick check.

mtbc:
@jburel: Anything to check before merging?

jburel:
most of the code is commented out and the init is in a try/catch block

dominikl:
Didn't notice any errors in Insight, looks good from my point of view, too.

PR #5303, opened 2017-05-16
BrowseFacility load wells

dominikl:
# What this PR does

Adds a method to load wells by id to the gateway's `BrowseFacility`.

If you want to access the images for specific wells, you need to have properly initialized `WellData` objects.  The existing `findObject` methods don't do that, therefore this new method was necessary. 

# Testing this PR

Code review.

~~Unfortunately I don't have an integration for this method yet (would require some `fake` screen import for setup). But it will be tested indirectly by the `romero.gateway` R package, where this method will be needed.~~

Check `BrowseFacilityTest.testGetWells` integration test passes.

### DIFF
--- a/components/blitz/src/omero/gateway/facility/BrowseFacility.java
+++ b/components/blitz/src/omero/gateway/facility/BrowseFacility.java
@@ -30,6 +30,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


jburel:
You don't need to import ``fake`` files
you could mimic what is done in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroJava/test/integration/ModelMockFactory.java

PR #4916, opened 2016-11-01
Read only tables

joshmoore:
# What this PR does

HDF5 tables that are copied into place, read-only, should still be accessible via the OMERO.tables API.

# Testing this PR

1. Create an OMERO.tables file (e.g. with populate_metadata)
2. `chmod a-w` on the file under `/OMERO/Files`
3. Try to open the file via `openTable`
4. Without this PR, the open will fail with an `IOError`. With this PR, the file will be opened but not be writable.

# Related reading

* https://trello.com/c/H0tV3G7h/5-read-only-omero-tables

PR #5306, opened 2017-05-17
tables: if lock fails, fall back to read-only

joshmoore:
# What this PR does

Workaround portalocker.py being broken on NFS.

# Testing this PR

1. Mount /OMERO read-only from NFS (like on the orcas)
2. Try to open an HDF5 file. `Bad file descriptor` will be printed in Tables-0.log
3. With this patch, the file should be opened read-only

# Related reading

1. See gh-4916

### DIFF
--- a/components/tools/OmeroPy/src/omero/tables.py
+++ b/components/tools/OmeroPy/src/omero/tables.py
@@ -8,6 +8,7 @@
 
 import Ice
 import time
+import errno
 import numpy
 import logging
 import threading
@@ -137,7 +138,7 @@ class HdfList(object):
                 None, None,
                 "Cannot acquire exclusive lock on: %s" % hdfpath, 0)
         except IOError, ie:
-            if ie.errno == 9:  # Bad File Descriptor (NFS)
+            if ie.errno == errno.EBADF:  # Bad File Descriptor (NFS)
                 hdffile.close()
                 hdffile = hdfstorage.openfile("r")
                 fileno = hdffile.fileno()
--- a/components/tools/OmeroPy/src/omero/tables.py
+++ b/components/tools/OmeroPy/src/omero/tables.py
@@ -136,6 +136,14 @@ class HdfList(object):
             raise omero.LockTimeout(
                 None, None,
                 "Cannot acquire exclusive lock on: %s" % hdfpath, 0)
+        except IOError, ie:
+            if ie.errno == 9:  # Bad File Descriptor (NFS)
+                hdffile.close()
+                hdffile = hdfstorage.openfile("r")
+                fileno = hdffile.fileno()
+            else:
+                hdffile.close()
+                raise
         except:
             hdffile.close()
             raise


components/tools/OmeroPy/src/omero/tables.py
```diff
@@ -136,6 +136,14 @@ def addOrThrow(self, hdfpath, hdfstorage):
             raise omero.LockTimeout(
                 None, None,
                 "Cannot acquire exclusive lock on: %s" % hdfpath, 0)
+        except IOError, ie:
+            if ie.errno == 9:  # Bad File Descriptor (NFS)
+                hdffile.close()
+                hdffile = hdfstorage.openfile("r")
+                fileno = hdffile.fileno()
```
mtbc: If this throws does the hdffile get closed?

joshmoore: Considering the previous run with less expectations ("r" < "w") passed, I don't expect so, but you're definitely right that something weird could occur. Thanks.

joshmoore: Actually, the original code has this same issue in lines 128-129. Do you agree?

mtbc: I do: https://github.com/openmicroscopy/openmicroscopy/blob/6a7279afffab7b3af9fe73d40e4b89e006dff204/components/tools/OmeroPy/src/omero/tables.py#L128

components/tools/OmeroPy/src/omero/tables.py
```diff
@@ -136,6 +136,14 @@ def addOrThrow(self, hdfpath, hdfstorage):
             raise omero.LockTimeout(
                 None, None,
                 "Cannot acquire exclusive lock on: %s" % hdfpath, 0)
+        except IOError, ie:
```
manics, line 140: Since this is new code can you use `as` instead? https://stackoverflow.com/a/2535770

components/tools/OmeroPy/src/omero/tables.py
```diff
@@ -136,6 +136,14 @@ def addOrThrow(self, hdfpath, hdfstorage):
...
+            if ie.errno == 9:  # Bad File Descriptor (NFS)
```
manics: https://github.com/openmicroscopy/openmicroscopy/blob/v5.3.2/components/tools/OmeroPy/src/omero/cli.py#L504 Uses a named constant, do you want to rationalize?

manics (approved):
Tested in vagrant with
```
- hosts: all

  pre_tasks:
    - name: Create Directories
      become: yes
      file:
        path: "{{ item }}"
        state: directory
      with_items:
      - /srv/nfs-data
      - /mnt/nfs-ro

  roles:
    - role: openmicroscopy.postgresql
      postgresql_users_databases:
      - user: omero
        password: omero
        databases: [omero]
    - role: openmicroscopy.omero-server
      omero_server_upgrade: True
      omero_server_omego_additional_args: --ci https://HOSTNAME:8443 --branch OMERO-build

    - role: openmicroscopy.nfs-share
      nfs_shares:
        /srv/nfs-data:
        - host: "*"
          options: ro,root_squash
    - role: openmicroscopy.nfs-mount
      nfs_share_mounts:
      - path: /mnt/nfs-ro
        location: localhost:/srv/nfs-data

  environment:
    OMEGO_SSL_NO_VERIFY: 1

  vars:
    ice_python_wheel: http://users.openmicroscopy.org.uk/~spli/centos-zeroc-ice-py/dist/zeroc_ice-3.6.3-cp27-cp27mu-linux_x86_64.whl
    postgresql_version: "9.6"
```
Good to merge (code comments are optional)

For reference I created a table using `omero shell login`, then:
```
mv /OMERO/Files/75 /srv/nfs-data/
ln -s /mnt/nfs-ro/75 /OMERO/Files/75
```

components/tools/OmeroPy/src/omero/tables.py
joshmoore: Holding off on a styling change at this point.

PR #5307, opened 2017-05-18
make GraphPathBean.PropertyDetails private

mtbc:
Makes `GraphPathBean.PropertyDetails` a private class. It was previously exposed to package/default visibility for `GraphPathReport`'s copy-and-paste code to reuse. It was not public but seems safest to open against roles rather than mainline.


joshmoore:
:+1:

PR #5308, opened 2017-05-22
Gateway login improvements

dominikl:
# What this PR does

Fixes two issues:
- If a server doesn't have a `guest` account, the connection failed because `ServerError` was thrown (and not handled) on the session id check. Noticed during workshop prep for the romero.gateway notebook.
- Generally if something failed on the session creation and/or login level the client got a very generic `DSOutOfServiceException`. With this PR the client at least will get some more information about the reason. 'Host not reachable', 'Login credentials invalid', 'Version mismatch', etc. When extracting the Gateway code from Insight somewhere on the way the version check got lost. Can't imagine that was intentional, therefore added version check again (otherwise the client get rather cryptic marshal exceptions).

# Testing this PR

- Try to connect to a server which doesn't have the `guest` account. Is there still a server like this running somewhere @joshmoore ?
- Try various error conditions, wrong credentials, incompatible server version, etc. Check that the message of the `DSOutOfServiceException` makes sense.
- Try a valid condition, and make sure you can still connect to the server with correct version and login credentials.
- Check integration test

Here's a simple example test 'client' code: https://gist.github.com/dominikl/b48a029cb4977bfafa7ddb3f027d6d85 or test with Insight and check its log file.

### DIFF
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -263,7 +263,7 @@ public class Gateway {
         } catch (PermissionDeniedException e) {
             throw new DSOutOfServiceException("Login credentials not valid", e);
         } catch (ServerError e) {
-            throw new DSOutOfServiceException("Server error", e);
+            throw new DSOutOfServiceException(e.getMessage(), e);
         } catch (SocketException e) {
             throw new DSOutOfServiceException(e.getMessage(), e);
         } catch (DNSException e) {
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -85,6 +85,7 @@ import omero.gateway.model.GroupData;
 import omero.gateway.util.PojoMapper;
 import Glacier2.CannotCreateSessionException;
 import Glacier2.PermissionDeniedException;
+import Ice.ConnectionRefusedException;
 import Ice.DNSException;
 import Ice.SocketException;
 
@@ -264,6 +265,8 @@ public class Gateway {
         } catch (ServerError e) {
             throw new DSOutOfServiceException("Server error", e);
         } catch (SocketException e) {
+            if (e instanceof ConnectionRefusedException)
+                throw new DSOutOfServiceException("Connection refused", e);
             throw new DSOutOfServiceException("Host unreachable", e);
         } catch (DNSException e) {
             throw new DSOutOfServiceException("Can't resolve hostname "
@@ -1055,7 +1058,8 @@ public class Gateway {
                             .getUsername(), c.getUser().getPassword());
                 }
             }
-        } catch (Exception e1) {
+        } 
+        catch (Exception e1) {
             // close the session again before passing on the exception
             secureClient.closeSession();
             throw e1;
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -265,9 +265,7 @@ public class Gateway {
         } catch (ServerError e) {
             throw new DSOutOfServiceException("Server error", e);
         } catch (SocketException e) {
-            if (e instanceof ConnectionRefusedException)
-                throw new DSOutOfServiceException("Connection refused", e);
-            throw new DSOutOfServiceException("Host unreachable", e);
+            throw new DSOutOfServiceException(e.getMessage(), e);
         } catch (DNSException e) {
             throw new DSOutOfServiceException("Can't resolve hostname "
                     + c.getServer().getHostname(), e);
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -1013,8 +1013,9 @@ public class Gateway {
                         "guest", "guest");
                 this.pcs.firePropertyChange(PROP_SESSION_CREATED, null, secureClient.getSessionId());
                 guestSession.getSessionService().getSession(username);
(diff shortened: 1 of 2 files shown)


mtbc:
* Correct login connected fine.
* Bad password gave me `DSOutOfServiceException: Login credentials not valid`.
* Bad host gave me `DSOutOfServiceException: Can't resolve hostname ...`

but,

* Trying incompatible server version *also* gives me, `DSOutOfServiceException: Login credentials not valid`.
* I can run a local server but I don't know how to disable the `guest` user nor from the `TestConnect` side how to tweak the `LoginCredentials` or what to expect having done so.
* At least some of these conditions could have integration test coverage?

mtbc:
If the server is not running on a host then I get `DSOutOfServiceException: Host unreachable` even though the host is perfectly reachable and simply rejecting the connection to that port.

mtbc:
:+1: for example test code, much appreciated.

dominikl:
Which server did you use for the incompatible version test? I tried demo server, which gave me `omero.gateway.exception.DSOutOfServiceException: Client version 5.3.3-ice36-SNAPSHOT is not compatible with server version 5.2.7` as expected.

mtbc:
I used 10.0.51.154 (roles/5.4).

dominikl:
The server returns version 5.3.2  (`IConfigPrx.getVersion()`)

sbesson:
@mtbc @dominikl: as we never created a starting tag e.g.`5.4.0-m0` tag on the roles branch and since the OMERO versioning is tag-based, the roles server version is effectively `5.3.x` as confirmed https://github.com/openmicroscopy/openmicroscopy/pull/5308#issuecomment-304001477 and escapes the version comparison logic. Is that an incentive to creating a 5.4.x tag asap at least to clarify the expectations?

mtbc:
Woah, I had no idea that the config service returns something generated via git, now I wonder what happens if I just build from a source zip rather than a git repo? Anyhow, happy to defer to @pwalczysko regarding tagging.

jburel:
@sbesson there is a 5.4.0-m0 tag

mtbc:
(How come `v5.4.0-m0` didn't suffice?)

dominikl:
Certainly not necessary to create a tag just for testing this PR, unless you'd need it otherwise, too.

sbesson:
Ah apologies @mtbc and @jburel . The issue actually comes from the fact the last 5.3.x releases where merged into the roles branch including the corresponding `v5.3.x` tag. Thus the output of `git describe` uses the last available tag which is now `v5.3.2` on the roles branch

mtbc:
Ah, if the current devspace-for-new-major-version workflow is one we plan to repeat then this is an issue we'll have to fix.

components/blitz/src/omero/gateway/Gateway.java
```diff
@@ -264,6 +265,8 @@ public ExperimenterData connect(LoginCredentials c)
         } catch (ServerError e) {
             throw new DSOutOfServiceException("Server error", e);
         } catch (SocketException e) {
+            if (e instanceof ConnectionRefusedException)
+                throw new DSOutOfServiceException("Connection refused", e);
             throw new DSOutOfServiceException("Host unreachable", e);
```
mtbc: does `SocketException` happen to already come with a useful message that can just be repeated?

mtbc:
Outwith this PR, but apparently we can push some tag like `v5.4.0-m0p1` and whatnot after 5.3 merges.

dominikl:
Oh dear... argh, the exception handling is quite a mess in the gateway. Wasn't a good idea to wrap nearly everything into a `DSOutofSeriviceException` in first place, but now I have to stick to it, to not introduce breaking changes. Anyway, will add a commit and pass on the `ServerError` message at least.

mtbc:
(travis needs a kick, maybe tables tests?)

dominikl:
Finally, travis green. Ready for another review now. But I still don't know how to test the 'guest' account problem, no idea how to remove the guest account from a server.

mtbc:
Tried connecting to `www.ksu.edu` (!) and got a,

```
omero.ClientError: Obtained null object proxy
```

joshmoore:
> omero.ClientError: Obtained null object proxy

Expected if ports are not open, etc.

dominikl:
Hm... should that be transformed into a more meaningful error message then?

joshmoore:
https://github.com/dominikl/openmicroscopy/blob/620df26c511f92d26e9d4c6e722fab1da7e64113/components/blitz/src/omero/client.java#L750 which is repeated in each language will need to be modified.

mtbc:
With `demo.openmicroscopy.org`,

> Client version 5.3.3-ice35-SNAPSHOT is not compatible with server version 5.2.7

:+1:

mtbc:
This PR is **good to merge** for me if you prefer further work to follow in subsequent PRs. :+1:

dominikl:
Thanks @mtbc . Yes, I'll leave it to that, if necessary I can always open another new PR as follow-up.

jburel:
@dominikl do you have a card to capture the next adjustments?

dominikl:
No. I'm actually a bit lost at the moment, where to add things (also with respect to the R gateway). OMERO 5.3.3 board, 5.4.0-m1, 5.3.x, etc. ?

mtbc:
https://trello.com/b/SiqOu2Bl/omero-540 maybe?

jburel:
5.4.0 board please

jburel:
Card created https://trello.com/c/aXTsTn5U/14-gateway

PR #5309, opened 2017-05-23
remove unused arguments and fields from graphs code

mtbc:
# What this PR does

Changes in the graph traversal code means that some arguments and fields (session, system types) are no longer required. This PR reduces the clutter in the code without changing behavior.

# Testing this PR

CI should show no regressions.


joshmoore:
Assuming this is rolling back changes made on this branch, then LGTM. If not, the usual deprecation heads up. :+1:

mtbc:
Thank you. Probably will indeed need to open a deprecation PR on mainline. *Update:* See #5310.

PR #5311, opened 2017-05-25
have "git describe" set omero.version using --first-parent

mtbc:
# What this PR does

Ignores merged-in branches when deciding which git tag to use in naming the build. This means that we don't have to make a 5.4 tag for a new-version devspace every time we merge in mainline. However, it does mean that once the devspace *becomes* mainline, *if* that is effected by merging the devspace into mainline then 5.4 would need to be tagged there.

# Testing this PR

Check out `roles`, try merging `scc/develop/merge/daily` into it, don't worry about merge conflicts, just try building and when it finishes or errors out see that `target/omero.version` mentions 5.4, not 5.3.

--exclude for now

# Related reading

Discussion on #5308.

### DIFF
--- a/components/antlib/resources/version.xml
+++ b/components/antlib/resources/version.xml
@@ -70,6 +70,7 @@
                     <arg value="--match"/>
                     <arg value="${gitmatch}"/>
                     <arg line="--dirty"/>
+                    <arg line="--first-parent"/>
                 </args>
             </git>
 
--- a/components/blitz/src/omero/gateway/facility/BrowseFacility.java
+++ b/components/blitz/src/omero/gateway/facility/BrowseFacility.java
@@ -30,6 +30,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.commons.collections.CollectionUtils;
+
 import omero.api.IContainerPrx;
 import omero.api.IQueryPrx;
 import omero.api.IScriptPrx;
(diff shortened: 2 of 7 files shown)


sbesson:
While the strategy looks reasonable, it is worth noting this would bump the requirement on the minimum version of Git since the `--first-parent` feature was introduced in Git 1.8.4. Practically,tThis requirement is effectively not satisfied on our current CI build node where the Git version if older than that and caused today's OMERO artifacts to be (un)versioned as `UNKNOWN` affecting the functionality of components like `omero-marshal` /cc @will-moore . If builds are required today, this PR should likely be excluded before restarting the daily jobs. /cc @jburel

mtbc:
Absolutely feel free to exclude this PR and retrigger CI. `eel` seems to have git 1.7.1 which is now seven years old. (!)

mtbc:
Switched to `roles` because for OMERO 5.4 we drop CentOS 6 so we may rely on a newer `git`. Thank you to @sbesson for diagnosing the problem.

mtbc:
Gah, in `roles` devspace this is causing,

> ome.conditions.InternalException: Bad version format:UNKNOWN-ice36
at ome.logic.ConfigImpl.getVersion(ConfigImpl.java:385)

preventing CLI imports, etc.

mtbc:
Also, bizarrely,

> ome.system.UpgradeCheck - UPGRADE AVAILABLE:Please upgrade to 5.2.4.

... 5.2.4?

joshmoore:
The QA setting hasn't been updated since Bio-Formats versions broke with OMERO.

sbesson:
Updated.

mtbc:
Any ideas for https://github.com/openmicroscopy/openmicroscopy/pull/5311#issuecomment-304590551? Otherwise maybe best if I just close, not sure what else to try.

PR #5313, opened 2017-05-28
Don't recursively load feedback on ajaxError

will-moore:
# What this PR does

This fixes the recursive loading of ```/feedback/feedback``` when loading this generates 500 errors.
Several times recently, qa has received thousands of e-mails starting:
```
Internal Server Error: /feedback/feedback/
```
which are caused by:
 - webclient page loads OK
 - subsequent ajax call fails (```500``` error)
 - this is handled by ajaxError trying to load ```/feedback/feedback/``` which also fails
 - then we enter a recursive loop, generating an e-mail to qa each time!

# Testing this PR

Can be tested locally:
- load webclient page - open devtools to monitor Network requests
- edit an OMERO.web python file to create an error e.g. bogus import statement
- trigger ajax call from webclient (e.g. click on Project)
- you should get a 500 response for the original ajax call and a subsequent call to ```feedback``` which also fails with 500 error.
- shouldn't see any more recursive calls to ```feedback```. 

cc @kennethgillen @manics

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.popup.js
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.popup.js
@@ -514,6 +514,10 @@ OME.setupAjaxError = function(feedbackUrl){
         } else if (req.status == 500) {
             // Our 500 handler returns only the stack-trace if request.is_json()
             error = req.responseText;
+            // If the failed request was loading feedback, prevent recursive loading of feedback!
+            if (settings.url.startsWith(feedbackUrl)) {
+                return;
+            }
             OME.feedback_dialog(error, feedbackUrl);
         } else if (req.status == 400) {
             if (req.responseText.indexOf('Request Line is too large') > -1) {


joshmoore:
Good catch, @will-moore . Two questions: 1) What caused this to finally bottom out? (i.e. why didn't it go on continously) and 2) is there a more robust way of catching loops that we should look into (after this PR goes in)?

will-moore:
I don't know what caused this to 'bottom out' on ```web-dev-merge```. When this happens locally during development, it just keeps going till I shut it down.
I'm not sure there's any way of catching loops. What's the definition of a loop? The 'keepalive ping' is a loop (setInterval) and so is e.g. polling for script completion. But we wouldn't want to kill those.
I'd consider the current failure to be a "one off", or at least we'd need to see more of this kind of bug in order to know how to generalise. E.g. do you try to tackle the problem in JavaScript or in Django?

joshmoore:
> When this happens locally during development, it just keeps going till I shut it down.

Definitely scary.

> I'm not sure there's any way of catching loops. What's the definition of a loop?

Something where the call stack is very, very deep.  Not that this is something we can use in production, but the rough idea would be similar to https://stackoverflow.com/questions/7462075/call-stack-size-within-javascript

> The 'keepalive ping' is a loop (setInterval) and so is e.g. polling for script completion. But we wouldn't want to kill those.

Agreed. I mispoke. "Loop" should have been "non-terminating recursive invocation".

will-moore:
The issue here is not JavaScript recursively calling itself (very deep stack) but JavaScript making an Ajax call with gives a 500 error (from Python) which is handled by JavaScript etc.
If I use ```console.trace()``` within the ```ajaxError``` then it's the same (small) size for every recursive call
```
ome.popup.js?_5.3.1-ice36-SNAPSHOT:508 ajaxError
(anonymous)	@	ome.popup.js?_5.3.1-ice36-SNAPSHOT:508
dispatch	@	jquery-1.11.1.js:4641
elemData.handle	@	jquery-1.11.1.js:4309
trigger	@	jquery-1.11.1.js:4550
done	@	jquery-1.11.1.js:9285
callback	@	jquery-1.11.1.js:9685
```

joshmoore:
Thanks for digging, @will-moore. I see some "literature" on 3xx loops, but nothing that includes a 500. Merging for the health of CI.

PR #5315, opened 2017-05-30
Api experimenters groups

will-moore:
# What this PR does

Adds initial support for ```/experimenters/``` and ```/experimentergroups/``` to the JSON API.

This includes updating the ```getQueryString()``` for the BlitzGateway ExperimenterWrapper and ExperimenterGroupWrapper classes, so that by default they load experimenters without groups and load groups without experimenters.

Need to use the opts to load experimenters / groups (same as for Image 'load_pixels' etc).
```
conn.getObjects("ExperimenterGroup", opts={load_experimenters=True})
conn.getObjects("Experimenter", opts={load_groups=True})
```
In several places in webadmin we are using ```conn.getObjects("ExperimenterGroup")``` to list groups (e.g. in the experimenter/new/ form) when we don't need to ```load_experimenters```. Now that these are not being loaded, the loading time for that page is ~2 x faster for me.

I've also updated the ```group.copyGroupExperimenterMap()``` and ```experimenter.copyGroupExperimenterMap()``` to lazy load if not previously loaded.
This means that doing the following is slower because of lazy loading, but at least it won't fail
```
for e in conn.getObjects("Experimenter"):
    e.copyGroupExperimenterMap()
    # or this also needs to load groups
    e.isAdmin()
```

TODOs:

 - [x] Support for single Experimenter or Group
 - [x] Tests
 - [ ] Loading experimenterGroupMap etc?? (see below)
 - [ ] Query experimenters or groups by name? E.g. to look up your own experimenter ID from omeName or lastName? E.g. ```/experimenters/?omeName=will```

# Testing this PR

1. Browse to ```/api/v0/m/experimenters/``` to list all.
1. filter with ```?group=3``` or with  ```/api/v0/m/groups/3/experimenters```
1. Browse to ```/api/v0/m/experimentergroups/``` to list all.
1. filter groups with ```?experimenter=3``` or with  ```/api/v0/m/experimenters/3/experimentergroups```
1. Follow links to see single experimenter or group: ```/api/v0/m/experimenters/3/``` or ```/api/v0/m/groups/3/```
1. Need to test webclient and webadmin in places where we are loading experimenters / groups to check that we don't get unloaded exceptions:
    - Main /webclient/ page - Groups/Users menu
    - Search page (groups select option)
    - webadmin listing and editing groups
    - webadmin listing and editing experimenters
    - User settings in webclient (including for group leaders)

# Related reading

To discuss:
 - Do we expose the full name ```experimenterGroups``` anywhere (urls, filters etc) or do we simply stick with ```groups```?
 - Currently omero-marshal doesn't support loading groupExperimenterMap for Experimenters OR Groups. This means you can't choose to list Groups with Experimenters loaded (or list Experimenters with Groups loaded). Also, can't see which members are Owners of a Group.

cc @chris-allan @jburel


will-moore:
Closing until after #POTN and more complete...

jburel:
```

./components/tools/OmeroWeb/omeroweb/api/urls.py:327:24: E127 continuation line over-indented for visual indent
./components/tools/OmeroWeb/omeroweb/webadmin/views.py:102:80: E501 line too long (82 > 79 characters)
```

mtbc:
excluding to let #5349 through now #5101 is in

will-moore:
@jburel Could you remove the "exclude" label from this PR? Thanks

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -2886,35 +2890,15 @@ def listColleagues(self):
...
+            "ExperimenterGroup", self.getEventContext().groupId,
+            opts={'load_experimenters': True})
         if not default.isPrivate() or self.isLeader():
             for d in default.copyGroupExperimenterMap():
                 if d is None:
                     continue
                 if d.child.id.val != self.getUserId():
                     yield ExperimenterWrapper(self, d.child)
 
-    def groupSummary(self, gid=None, exclude_self=False):
```
will-moore: Seems that webtagging is still using this.
So I will leave it until webtagging has updated.

joshmoore: Agreed, but removal of any method from `omero.gateway` should include a deprecation so if you foresee dropping it, perhaps add the deprecation now?

will-moore: It has been marked as deprecated for some time already. That's why I decided to remove it.

joshmoore: Apologies, you're of course right. I had confused this with https://github.com/openmicroscopy/openmicroscopy/pull/5402/files#diff-5586c69038567a9c1960ae4f6d731278L509 where I convinced myself that the removal was less critical since it was in `webclient`

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5648,15 +5653,42 @@ def _getQueryString(cls, opts=None):
         Returns string for building queries, loading Experimenters only.
 
         Returns a tuple of (query, clauses, params).
+        Supported opts: 'group': <group_id> to filter by ExperimenterGroup
+                        'load_groups': <bool> to load ExperimenterGroups'
```
jburel: extra ``'``

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5883,15 +5915,42 @@ def _getQueryString(cls, opts=None):
         Returns string for building queries, loading Experimenters for each
         group.
         Returns a tuple of (query, clauses, params).
+        Supported opts: 'experimenter': <experimenter_id> to filter by
+                                        ExperimenterGroup
```
jburel: filter by Experimenter

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -5883,15 +5915,42 @@ def _getQueryString(cls, opts=None):
...
+                        'load_experimenters': <bool> to load Experimenters'
```
jburel: extra ``'``

rgozim:
API's tested and working well. 

Only concern, as noted by @will-moore, might be the ability of a _public_ user being able to view experimenter and experimenter group data via the API (no authentication required to query the API).

jburel:
That is a serious concern

will-moore:
Hmmm - let's discuss the ```public``` user issue. If you've configured the ```public url regex``` then chances are that ```/api/``` urls are not public unless desired. E.g. they're not public on IDR.

joshmoore:
It would be useful if a API definition itself (whether this one, or one introduced by third parties) could have some "awareness" of its critical nature so that the burden isn't externalized to the sysadmin to always have a regex set correctly.

will-moore:
Re: discussion of ```guest``` user.
Logging in as guest, then using the session ID to access the JSON API gives the same exception as trying to load data via the CLI:
```
$ bin/omero user list
Using session 2960bf65-c0b4-4748-8b47-40eac3fdc22f (guest@localhost:4064). Idle timeout: 10 min. Current group: guest 
...
  File "/Users/wmoore/Desktop/OMERO/openmicroscopy/dist/lib/python/omero_api_IAdmin_ice.py", line 864, in lookupExperimenters
    return _M_omero.api.IAdmin._op_lookupExperimenters.invoke(self, ((), _ctx))
omero.SecurityViolation: exception ::omero::SecurityViolation
message = No matching roles found in [guest] for session 2960bf65-c0b4-4748-8b47-40eac3fdc22f (allowed: [user])
```
also seen at:
```/api/v0/m/projects/?bsession=2960bf65-c0b4-4748-8b47-40eac3fdc22f&server=1```

will-moore:
Need more discussion of experimenters/groups visibility for public & other users.
No need to get this into 5.4.0, closing for now...

PR #5316, opened 2017-06-05
Add new luts

jburel:
# What this PR does

Add 2 new luts

# Testing this PR

check that the luts are available and that they can be used

# Related reading

https://twitter.com/subflux/status/870760499746623488

cc @kennethgillen 
If this PR is merged, we will need to upgrade "lut.png" since we do not have yet a way to automatically generate it
cc @will-moore


mtbc:
They don't look pretty but they do seem to work and PUP_NR looks to match what @kennethgillen's tweet showed. Two small notes:

* they're the only ones named in upper instead of lower case
* I don't see their pretty strips in the chooser, is that the described `lut.png` issue?

luts

jburel:
@mtbc: for  the pretty strips, the ``lut.png`` will need to be updated.
I will adjust the names to match the rest

jburel:
@mtbc I had to force push to change name

will-moore:
Updated luts preview image in https://github.com/openmicroscopy/openmicroscopy/pull/5318

mtbc (approved):

jburel:
Excluding for tomorrow's build
I would like to see if we have issue in iviewer for example if we are missing the luts
--exclude

PR #5318, opened 2017-06-06
luts_janelia

will-moore:
# What this PR does

Updates the luts preview with luts added in https://github.com/openmicroscopy/openmicroscopy/pull/5316

# Testing this PR

1. Listing of LUTs in webclient should show preview for pup_br.lut and pup_nr.lut.
2. Picking of those luts should correspond to rendering that matches the lut previews.
3. Check that reverse intensity for regular colours still works (uses lut_10.png).


mtbc:
We should find a way to automate these menial aspects of adding LUTs! (PNG, height.)

mtbc (approved):

jburel:
@will-moore @waxenegger this will need to be propagated to figure and iviewer

will-moore:
@waxenegger I think I will remove the black->transparent gradient from the bottom of the ```luts_10.png```, mostly because this makes it harder to regenerate it when the LUTs list changes but also because it's more tricky to position this dynamically.
Also, when we do generate this dynamically on the server, it won't include the gradient either.

Another issue: With the JSON data above (even including 'png_index') we don't know for sure what the size of the ```luts_10.png``` is (or how many LUTs it contains). If some LUTs are removed from the server by a user, the returned JSON list will be shorter, even though the size of the png is the same. Maybe for now we assume that this isn't going to happen and that the highest `png_index` tells us the number of LUTs and therefore the size of the png (so we can scale it 3x height, e.g. ```background-size: 100% 1200px```. Makes sense?

mtbc:
(Outwith this PR, does the Blitz API need to be able to return a dynamic accurately suitable PNG from the actually available LUTs? If so, may be worth a card describing exactly what's required.)

waxenegger:
@will-moore sorry, I'm not a 100% sure I can follow. 

If we use the hard-coded list in the webgateway for what is supposed to be in the png (for now and for this more 'immediate' solution - any future solution aside)  then we do know if a lut is missing and we can put a placeholder (e.g.-1)  instead an index. I used this method and incremented the index only if the lut exists and it should work for both missing as well as newly added luts, right?

the overall height I get from the image, the last row (gradient) is the height - row-height. I'm not claiming it's the most straight-forward approach but I think it's a correct approach nonetheless.

jburel:
@mtbc we do have a card I think, I will look for it

will-moore:
@waxenegger You get the overall height of the png from the image itself? Do you load it into an into an img with an img.onload event to get the height? I guess I could do this too, but it would be nicer if the LUTs JSON gave me that. If I remove a bunch of luts from ```dist/lib/python/scripts/luts/``` then the LUTs JSON I get is a much reduced list. E.g. the highest ```png_index``` may be 10 even though all 38 of the LUTs are still in the static ```luts_10.png```. So, this can't be used to know the height of the png if we want to allow users to delete LUTs.

waxenegger:
@will-moore sure, it would be nicer if I got the height back with the json, then I'd save myself loading the image. If you iterate over the luts there are and increment only if the lut is in the hard-coded list, the index should not over/undershoot. There'd be -1 for gaps (whether removed lut or added lut) and the index should be corresponding.

will-moore:
@waxenegger I added the full list of ```png_luts``` to the LUTS JSON so you can use the length of that list to know the size of the ```luts_10.png```.
I have not yet removed the transparent gradient from the ```luts_10.png``` yet (as discussed above), but in OMERO.figure I have started using a separate gradient.png so I'm not relying on the gradient in ```luts_10.png``` since I would like to remove it at some point.
If you do the same in iviewer it would make it more future-proof.

waxenegger:
@will-moore see https://github.com/ome/omero-iviewer/pull/103

dominikl:
Just noticed, I should add this to Insight too...

mtbc (approved):

PR #5321, opened 2017-06-07
add a training example for bulk shape access

mtbc:
# What this PR does

Adds a Python training example showing that shapes can be retrieved from a ROI using batch querying for scalability. Batched writing can follow once https://trello.com/c/Q499M0qm/758-cannot-write-rois-at-scale is solved. Cc: @polarise.

# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/ should remain green. Also, give it a try locally:

```shell
$ python python/Bulk_Shapes.py 
created 20 shapes
retrieved 20 shapes in 4 batches
```

### DIFF
--- a/examples/Training/python/Bulk_Shapes.py
+++ b/examples/Training/python/Bulk_Shapes.py
@@ -45,7 +45,7 @@ for shape in range(total_shape_count):
     roi.addShape(point)
 
 roi = updateService.saveAndReturnObject(roi)
-print "created {} shapes".format(total_shape_count)
+print "created {0} shapes".format(total_shape_count)
 
 
 # Query Shapes in batches
@@ -68,8 +68,8 @@ while True:
         shapes_queried += len(shapes)
         shape_batch_count += 1
 
-print "retrieved {} shapes in {} batches".format(shapes_queried,
-                                                 shape_batch_count)
+print "retrieved {0} shapes in {1} batches".format(shapes_queried,
+                                                   shape_batch_count)
 
 
 # Delete ROI with Shapes
--- a/examples/Training/python/Bulk_Shapes.py
+++ b/examples/Training/python/Bulk_Shapes.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+#
+# Copyright (C) 2014-2017 University of Dundee & Open Microscopy Environment.
+#                    All Rights Reserved.
+# Use is subject to license terms supplied in LICENSE.txt
+#
+
+"""
+FOR TRAINING PURPOSES ONLY!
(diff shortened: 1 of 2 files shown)


jburel:
@mtbc you will need to add the new example to https://github.com/openmicroscopy/openmicroscopy/blob/develop/examples/Training/python/__main__.py otherwise it will not be run

mtbc:
Thank you @jburel.

jburel:
Today's output
```
Running training suite
  File "/usr/lib64/python2.6/runpy.py", line 122, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.6/runpy.py", line 34, in _run_code
    exec code in run_globals
  File "/opt/hudson/workspace/OMERO-DEV-merge-training/src/examples/Training/python/__main__.py", line 15, in <module>
    execfile(os.path.join(training_dir, 'Bulk_Shapes.py'))
  File "/opt/hudson/workspace/OMERO-DEV-merge-training/src/examples/Training/python/Bulk_Shapes.py", line 48, in <module>
    print "created {} shapes".format(total_shape_count)
ValueError: zero length field name in format
```

mtbc:
I think that's a Python 2.6 -ism. As OMERO 5.3 requires Python 2.7 I don't think that's a problem with this PR.

mtbc:
(Or did I misunderstand http://www.openmicroscopy.org/site/support/omero5.3/sysadmins/version-requirements.html#python? Now I look again its meaning is ambiguous, cc: @hflynn, @sbesson.)

sbesson:
@mtbc: agreed we might need to clarify this page. My understanding is the following in terms of minimal requirements:
- OMERO.py 5.3: Python 2.6 minimum (hence OMERO.server 5.3: Python 2.6 minimum)
- OMERO.web 5.3: Python 2.7 minimum

From a client perspective, I would expect Python 2.6 code to be supported for the lifetime of OMERO 5.3 although Python 2.7 would be recommended. In the case of this training example addition, it might be easier for now to fall back to Python 2.6 semantics if that's not too costly also as we need to schedule the work for upgrading our OMERO CI infrastructure to Python 2.7. /cc @jburel

jburel:
@sbesson is correct.
I will open a PR to clarify the doc
Thanks for pointing that out

jburel:
see https://github.com/openmicroscopy/ome-documentation/pull/1698

mtbc:
Ha, thank you for the docs fix! For this PR the corresponding fix is easy.

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/646/ is green and the example has been run

PR #5322, opened 2017-06-12
render.py active channels

jburel:
# What this PR does

Copy settings even if the channel is off

# Testing this PR

Copy settings across images with some channels off

# Related reading
https://trello.com/c/6GvOjNN0/26-bug-rendering-settings-not-applied-to-channels-that-also-have-the-tag-active-false

cc @joshmoore

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/render.py
+++ b/components/tools/OmeroPy/src/omero/plugins/render.py
@@ -457,8 +457,6 @@ class RenderControl(BaseControl):
         for (i, c) in newchannels.iteritems():
             if c.label:
                 namedict[i] = c.label
-            if not c.active:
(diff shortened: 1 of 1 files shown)


jburel:
I just realised that there is a limited test coverage for the plugin 
I will add a test for the change

jburel:
We should probably migrate ``render.py`` to ``ome``. see https://trello.com/c/qMrZfcoU/78-external-cli-plugins
In that case, I will close this PR and fix against the error against the new repo

joshmoore:
https://github.com/ome/omero-cli-render created. Note: we may want to add a page to the main documentation so that the URL on that repository can point to "CLI Plugins" (etc).

PR #5324, opened 2017-06-13
zeroc-ice for ice 3.5

jburel:
# What this PR does

Do not install zeroc-ice for ice 3.5

# Testing this PR

Use omero-install with OMEROVER=OMERO-DEV-merge-build ICEVER=ice35

# Related reading

Error showed up when investigating https://github.com/ome/omero-install/pull/154

see https://github.com/ome/omero-install/pull/156

### DIFF
--- a/components/tools/OmeroWeb/requirements-py27-all-ice35.txt
+++ b/components/tools/OmeroWeb/requirements-py27-all-ice35.txt
@@ -1,0 +1,9 @@
+# Python installation requirements for OMERO.web
+# ==============================================
+#
+#     pip install -r requirements-py27-all-ice35.txt
+#
+
+numpy>=1.9
+Pillow<3.4
+-r requirements-py27-ice35.txt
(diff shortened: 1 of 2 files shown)


joshmoore:
@jburel : did you look into handling this with [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers)? I remember Ola having an example but I haven't (yet) found which repository it is in.

My guess is that something like:

```
zeroc-ice>3.5,<3.7; python_version > '2.6'
```

would suffice to reduce the redundancy (and therefore increase the usability) of these files.

jburel:
@joshmoore I have not looked at it yet. I discovered the problem late last night :-) So I went for a "safe" solution

jburel:
It will have to be the ice version. I will see how I can get hold of it something like ``slice2py --version``

dominikl:
'omero-install' on docker went fine at least 👍

jburel:
unfortunately I have not found a solution using the marker in the requirements file
so we will probably have to go with the multi-files option for now

jburel:
This is still going to be an issue with 2 versions of ice to support
the check for ``zeroc-ice`` version will not work for 3.7 for example

PR #5327, opened 2017-06-14
have graph traversal automatically review "linking" model objects

mtbc:
# What this PR does

Simplifies the graph policy rules specified in `components/blitz/resources/ome/services/graph-rules/` so that the null state changes to `[-]` are largely no longer needed. Some remain for `Chmod2`: though no longer even necessary there, they linger to keep that ruleset clearer, and are made more specific so that every object in the group need not be visited.

Corresponding graph policy code changes automatically guess where `[-]` rules should be and act accordingly.

Also adjusts a few rules to better fit the pattern by which others are expressed and addresses a minor ambiguity in the rule language that affected `OTF` which was used as both term name and class name.

# Testing this PR

https://10.0.51.154:8443/job/OMERO-test-integration/ should suffice. Graph operations should continue to work as before.

# Related reading

https://trello.com/c/jL8kYW5V/47-automatically-recheck-multi-match-graph-rules

### DIFF
--- a/components/blitz/resources/ome/services/graph-rules/blitz-chgrp-rules.xml
+++ b/components/blitz/resources/ome/services/graph-rules/blitz-chgrp-rules.xml
@@ -682,15 +682,15 @@
         <!-- Move a reagent only if also moving screens and wells that use it. -->
 
         <bean parent="graphPolicyRule" p:matches="Screen[I].reagents = R:[E]{i}" p:changes="R:{r}"/>
-        <bean parent="graphPolicyRule" p:matches="L:WellReagentLink[!D].parent = [I], L.child = R:[E]{i}" p:changes="R:{r}"/>
+        <bean parent="graphPolicyRule" p:matches="L:WellReagentLink[EI].parent = [I], L.child = R:[E]{i}" p:changes="R:{r}"/>
         <bean parent="graphPolicyRule" p:matches="WellReagentLink[I].child = R:[E]{i}" p:changes="R:{r}"/>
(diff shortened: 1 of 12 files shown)


mtbc:
Moved from taking 4h0 to 4h3 so a negligible slowdown, excellent.

(Test failures on https://10.0.51.154:8443/job/OMERO-test-integration/77/ I do have a local approach for fixing but I couldn't easily pin down and reproduce to double-check; at least some are intermittent. At least now I know where to focus.)

mtbc:
The integration tests now run a little faster; perhaps `Chmod2` got sped up a little by https://github.com/openmicroscopy/openmicroscopy/pull/5327/commits/2bd02414.

dominikl:
The integration tests passed (just one unrelated failed integration test). Didn't notice any problems (just used Insight to perform some delete and change group operations). But if necessary someone else who is more familiar with the Graph issue should review the code.

mtbc:
Thank you @dominikl. Unfortunately nobody else is much familiar with the code. Still, I'll not merge this PR today in case someone else does want to jump in for a look.

joshmoore:
I've glanced over the code and wasn't horrified. (Certainly the reduction in rules is nice)

mtbc:
@pwalczysko's JVM stack blew in a large chown due to https://github.com/openmicroscopy/openmicroscopy/pull/5327/files#diff-273b19a35d4a2949ed5ddbe425ac40a1R1230. This should be rewritten non-recursively.

mtbc:
Fixed in #5457.

mtbc:
> the reduction in rules

Especially, not having to notice when such rules might be needed.

PR #5328, opened 2017-06-14
remove checks for Ice version < 3.6

mtbc:
# What this PR does

Removes code paths that apply only to Ice < 3.6.

# Testing this PR

CI should remain green. Also check that can still use 5.4 devspace web client and Insight client to connect.

# Related reading

http://www.openmicroscopy.org/site/support/omero5/sysadmins/version-requirements.html#ice

### DIFF
--- a/components/blitz/src/omero/client.java
+++ b/components/blitz/src/omero/client.java
@@ -331,14 +331,10 @@ public class client {
 
         // Strictly necessary for this class to work
         optionallySetProperty(id, "Ice.ImplicitContext", "Shared");
-        if (Ice.Util.intVersion() >= 30600) {
-            optionallySetProperty(id, "Ice.ACM.Client.Timeout",
-                    ""+omero.constants.ACMCLIENTTIMEOUT.value);
-            optionallySetProperty(id, "Ice.ACM.Client.Heartbeat", ""+
-                    omero.constants.ACMCLIENTHEARTBEAT.value);
-        } else {
-            optionallySetProperty(id, "Ice.ACM.Client", "0");
-        }
+        optionallySetProperty(id, "Ice.ACM.Client.Timeout",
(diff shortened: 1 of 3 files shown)


jburel:
This might be better to be included post 5.4.0-m1

components/blitz/src/omero/client.java
```diff
@@ -350,11 +346,8 @@ private void init(Ice.InitializationData id) {
         optionallySetProperty(id, "omero.block_size", Integer
             .toString(omero.constants.DEFAULTBLOCKSIZE.value));
 
-        // Set the default encoding if this is Ice 3.5 or later
-        // and none is set.
-        if (Ice.Util.intVersion() >= 30500) {
-            optionallySetProperty(id, "Ice.Default.EncodingVersion", "1.0");
-        }
+        // Set the default encoding if none is set.
+        optionallySetProperty(id, "Ice.Default.EncodingVersion", "1.0");
```
rleigh-codelibre, line 350: With Ice 3.6 and above being the only supported versions, we could also look at increasing the encoding version here as well.  The same applies to other instances of the encoding version in this PR and the codebase in general.  May *possibly* have some impact on shares, but I'm not certain of that.

jburel:
omero.properties still has 3.5
As indicated in standup, other sections will need to be reviewed at the same time
It will be preferable to have a card with a list of actions e.g. ci, repo to check so it is "less random" before opening a PR that will break the daily build

jburel:
and we might need similar checks with 3.7 so this could go in a "3.7"support effort

mtbc:
I don't believe that this PR would break the daily build unless it's relying on deploying the Ice 3.5 artifacts?

As discussed in person, better to make the various Ice changes all in one PR so some are not forgotten. I don't know what all those are and am not planning to investigate right now so closing this PR.

PR #5333, opened 2017-06-16
luts_janelia

dominikl:
# What this PR does

Update LUT icons to include pup_br.lut and pup_nr.lut

# Testing this PR

Check that pup_br and pup_nr LUTs have icons in the colorpicker.

# Related reading

Related Web PR: #5318


mtbc:
If in the preview pane I pick a channel and pick a LUT and hit "Preview" then I see the change but if I then hit "Cancel" instead of "Accept" the LUT effects are still shown. Is that a known bug?

mtbc:
Huh. And if I pick a LUT, hit "Preview", then choose one of the solid (non-LUT?) colors near the top of the list and hit "Preview" again, I don't see the preview image update.

mtbc:
Outwith this PR, as it's the same on web, but I'm curious about how we get a black section where none of the LUT looks black:

screenshot

mtbc (approved):
Some peculiar issues made testing tricky but within the scope of this PR I think that all looks fine.

dominikl:
That's true. The cancel and preview behaviour is quite buggy, especially when previewing and switching between LUTs multiple times... Can add that to the Insight bugs trello card.
RE the black section, I think both LUTs contain black values, at least the icon looks like it at the far right side (can't confirm, I don't know anything about these LUTs).

pwalczysko:
Not sure about 

>... as it's the same on web, 

No Preview button in Web. How would you repeat your workflow in web then ?
I agree that this is probably not the fault of this PR. 

Edit: Got it, only the black in the image is the same on Web.

mtbc:
On closer inspection the LUT *does* turn out to have a thin black stripe on the far right. Perhaps in my viewing 16-bit images there was some "color interpolation" happening between that LUT stripe and the adjacent one for the next-lowest 8-bit pixel value so that is how I got dark color shades from a bright-looking LUT. (Also with interpolation turned off in the client so not from the adjacent pixels unless JPEG compression is a culprit.)

pwalczysko:
Double-checked the black stripe in freshly downloaded luts from the janelia paper and opening them in Fiji. There indeed is a black bar in the lut at the max value in Fiji.

PR #5335, opened 2017-06-19
License

jburel:
# What this PR does

 * Update the README with list of dependencies
 * Remove dependencies in insight

# Testing this PR

Check that the build is green

cc @chris-allan @jrswedlow

### DIFF
--- a/lib/licenses/README
+++ b/lib/licenses/README
@@ -16,7 +16,7 @@
   Within this document, the term 'third-party software' shall refer to the set
   of all the files which are contained in the source repository and which are
   listed in each 'Artifacts' entry within the 'Third-Party Software Licenses'
-  section of this document.  The term 'OME' or, equivalently, 'our software'
+  section of this document. The term 'OME' or, equivalently, 'our software'
   shall refer to any other file contained in the repository.
 
 
@@ -25,7 +25,7 @@
   -------
 
   This document states which licenses apply to which files in the
-  source repository.  It also gives credit to other projects and
+  source repository. It also gives credit to other projects and
   developers that provided us with their valuable software.
 
 
@@ -47,9 +47,9 @@
   files. This section lists all the third-party software that is used
   by OMERO.server using the following format:
 
-  Artifacts:    The files our software makes use of.  The wildcard character * can
+  Artifacts:    The files our software makes use of. The wildcard character * can
                 be appended to a sequence of characters S denoting a file name.
-                This shall mean 'all the files whose name starts with S'.  For example,
+                This shall mean 'all the files whose name starts with S'. For example,
                 'ant-*' means all the files whose name starts with 'ant-'.
                 Unless otherwise specified, all artifacts are to be found under
                 lib/repository. If specified directory pathnames are given relative to
@@ -597,7 +597,7 @@
 
   Artifacts:    lwf-stubs-*.jar
   Project:      Stub of proprietary Java API to handle Luratech LWF compression.
-                Required to compile Bio-Formats's support for Luratech LWF compression for the Opera Flex format.
+                Required to compile Bio-Formats' support for Luratech LWF compression for the Opera Flex format.
   Web Site:     http://www.luratech.com/
   License:      BSD License
   License File: bsd.txt
@@ -608,8 +608,15 @@
   License File: cddl-v1.0.txt
 
 
-  Artifacts:    mipav-stubs-*.jar
+  Artifacts:    metakit-*.jar
   Project:      A library for reading Metakit database files.
+  Web Site:     http://www.openmicroscopy.org/site/products/bio-formats
+  License:      General Public License (GPL), Version 2
+  License File: gpl-v2.0.txt
+
+
+  Artifacts:    mipav-stubs-*.jar
+  Project:      Stubs of MIPAV Java classes required to compile Bio-Formats' MIPAV plugin.
   Web Site:     http://mipav.cit.nih.gov
   License:      BSD License
   License File: bsd.txt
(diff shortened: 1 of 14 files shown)


jburel:
--exclude

lib/licenses/README
```diff
@@ -329,10 +577,54 @@
...
+  Project:      Minimal overhead Java logging
+  Web Site:     http://code.google.com/p/minlog/
+  License:      New BSD License
+  License File: bsd-new.txt
+
+
+  Artifacts:    native-lib-loader-*.jar
+  Project:      A library for loading native libraries.
+  Web Site:     http://github.com/scijava/native-lib-loader
+  License:      Free BSD License
```
mtbc: usually "FreeBSD" single word

lib/licenses/README
```diff
@@ -445,26 +776,66 @@
...
-  Web Site:     http://www.jgoodies.com/freeware/forms/index.html
-  License:      BSD
+  Artifacts:    xml-apis-*.jar
+  Project:      xml-commons provides an Apache-hosted set of DOM, SAX, and 
+                JAXP interfaces for use in other xml-based projects. Our hope is that we 
+                can standardize on both a common version and packaging scheme for these 
+                critical XML standards interfaces to make the lives of both our developers 
+                and users easier. The External Components portion of xml-commons contains 
+                interfaces that are defined by external standards organizations. For DOM, 
+                that's the W3C;
```
mtbc: odd to end on a semicolon

mtbc:
We seem to have a couple of libraries with no license noted. Have we at least a Trello card for investigating further?

jburel:
@mtbc @will-moore is looking at the ones I left

will-moore:
SpinBtn is long dead. No license info and the only url I can find is 404 http://www.softwareunity.com/jquery/JQuerySpinBtn/
Most jQuery stuff uses MIT license

jburel:
I will add to  https://trello.com/c/zNVVhCR2/28-javascript-and-python-dependencies-review so we can look at removing it

lib/licenses/README
```diff
@@ -564,162 +961,241 @@
...
+  License:      MIT License
+  License File: mit.txt
 
 
-  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/raphael/raphael-min.js
-  Project:      Raphael javascript library
-  Web Site:     http://raphaeljs.com/
-  License:      MIT license.
-  License File: jquery.txt
+  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/ie8_pngfix/*
```
will-moore: I can't see that we use this anywhere.

lib/licenses/README
```diff
@@ -564,162 +961,241 @@
...
+  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/jquery.quicksearch-1.0.js
+  Web Site:     https://github.com/riklomas/quicksearch
+  License:      MIT License
+  License File: mit.txt
+
+
+  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/jquery.selectboxes-2.2.6.js
+  Web Site:     
+  License:      MIT License or GNU GENERAL PUBLIC LICENSE
+  License File: mit.txt, gpl-v3.0.txt
```
will-moore, line 1242: Can't see that we use this anywhere.

lib/licenses/README
```diff
@@ -564,162 +961,241 @@
...
+  License:      MIT License
+  License File: mit.txt
 
 
-  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/d3*
-  Project:      d3.js
-  Web Site:     http://underscorejs.org/
-  License:      BSD
-  License File: bsd.txt
+  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/aops-1.3.js
```
will-moore, line 1203: We don't use this aops-1.3.js library anywhere.

lib/licenses/README
```diff
@@ -564,162 +961,241 @@
...
+
+  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/panojs-2.0.0/*
+  Project:      An interactive JavaScript widget for panning and zooming a panoramic image
+                stitched together dynamically from smaller tiles. 
+  Web Site:     http://www.dimin.net/software/panojs/
+  License:      MIT License
+  License File: mit.txt
+
+
+  Artifacts:    components/tools/OmeroWeb/omeroweb/webgateway/static/3rdparty/pie-1.0beta5/*
```
will-moore, line 1184: I'm sure we can remove this now - only for old IE.

jburel:
@will-moore I will not remove them from the code in the point release. 
I have added them to the card mentioned above.
They have similar licenses that some already listed

sbesson:
@jburel: one thought while scanning `lib/licenses/README` is that we are not listing the Bio-Formats components (`formats-gpl`...). Is that a deliberate choice for our own dependencies?

jburel:
I did not list any of our own dependencies.
Happy to do so if we think this is required

chris-allan:
I would think it safe to link to the Bio-Formats dependencies rather than potentially introduce copy and paste errors and additional maintenance burden by also including them in `lib/licenses/README`.

sbesson:
As we are independently maintaining the list Bio-Formats dependencies, I am definitely in favor of linking rather than trying to maintain in two places. I will review this list and flag all the low-level dependencies worth discussing.

lib/licenses/README
```diff
@@ -658,6 +665,47 @@
...
+                memory efficiency improvements.
+                Used by Bio-Formats for OLE support (cxd, ipw, oib, zvi).
+                Used by VisBio overlays logic for XLS export feature.
+  Web Site:     http://jakarta.apache.org/poi/
+  License:      Apache License, Version 2.0
+  License File: apache-v2.0.txt
+
+
+  Artifacts:    ome-xml-*.jar
+  Project:      Java API to handle Microsoft MDB format (Access).
```
sbesson: That's the same description as `ome-mdbtools`

sbesson (commented):
Compared the content of the license README with the dependency report generated by Ivy and bundled under `share/{server,client}` in OMERO.server. 

The main missing dependencies I could find are:

- `formats-api`, `formats-gpl`, `formats-bsd`
- `bcprov-jdk14`, `bcmail-jdk14`
- `jts`
- `jaxrpc-api`
- `axis-saaj` should be `axis2-saaj`

jburel:
@sbesson the list should now be complete

sbesson:
The last commit addresses my comments from https://github.com/openmicroscopy/openmicroscopy/pull/5335#pullrequestreview-45710189

PR #5336, opened 2017-06-20
have periodic light admin privileges cleanup use quartz

mtbc:
# What this PR does

Adjusts the scheduling of the `_current_admin_privileges` database table cleanup to be done via the server's existing Quartz machinery instead of via a separate thread pool.

# Testing this PR

Keep an eye on the `_current_admin_privileges` table. If the server isn't doing anything then after ten seconds or so the table should be pretty much empty, old rows certainly shouldn't linger. Also can add to `etc/logback.xml` a line,

```xml
<logger name="ome.security.basic.LightAdminPrivilegesCleanup" level="DEBUG"/>
```

then in `var/log/Blitz-0.log` an entry should appear every ten seconds,

`running periodic cleanup of _current_admin_privileges table`

# Related reading

https://trello.com/c/12q3mmA3/24-scheduler-for-currentadminprivileges-table-cleanup

### DIFF
--- a/components/server/resources/ome/services/sec-system.xml
+++ b/components/server/resources/ome/services/sec-system.xml
@@ -98,9 +98,20 @@
     <constructor-arg ref="scriptRepoUuids"/>
   </bean>
 
-  <bean id="adminPrivilegesCleanup" class="ome.security.basic.LightAdminPrivilegesCleanup" destroy-method="close">
+  <bean id="adminPrivilegesCleanup" class="ome.security.basic.LightAdminPrivilegesCleanup">
     <constructor-arg ref="simpleSqlAction"/>
-    <constructor-arg value="10"></constructor-arg>  <!-- seconds -->
+    <constructor-arg value="5"></constructor-arg>  <!-- seconds -->
+  </bean>
+
+  <bean id="adminPrivilegesCleanupTrigger" class="org.springframework.scheduling.quartz.CronTriggerBean">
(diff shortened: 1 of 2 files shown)


mtbc:
Two things to look for in `psql`:

```sql
SELECT count(*) FROM _current_admin_privileges;
```

Because server activity causes new entries in the table, during much activity (e.g., run integration tests) this number should get quite large, but after the server has been quiet for many seconds then the number should drop.

```sql
SELECT transaction FROM _current_admin_privileges ORDER BY transaction LIMIT 1;
```

says the smallest transaction number in the table. When it exists at all then this should keep increasing: unless you are doing something like moving a *really* large plate, the same number should not hang around for minutes.

pwalczysko:
The behavour of psql table is as expected.

pwalczysko:
The periodic cleanup output in Blitz.log is happening too.

Ready to merge FMPOV.

mtbc:
@joshmoore: Is the implementation sufficiently improved?

joshmoore:
:+1: for no new thread pool and no blocking behavior. I'm still worried I'm missing some other option for you, but perhaps we just make sure this is part of the requirements for the backend queue.

mtbc:
I think it's just that Quartz isn't a good fit for delay-based scheduling.

I don't know if there's a Message Queues card/issue somewhere where we can add this as a "come back to" so that even this solution remains just temporary? I'm happy to note among MQ requirements but not sure where they are. :smiley:

joshmoore:
I've added to the design doc.

PR #5337, opened 2017-06-20
connection logging improvements

will-moore:
# What this PR does

As part of debugging connection failures reported in https://trello.com/c/YMv4rVNs/52-web-login-failures, we want to know why ```connector = session.get('connector', None)``` is failing in ```loging_required at omeroweb/decorators.py```.
Since the default session is ```omeroweb.filesessionstore``` (our modified version of django's file session store) I've added more logging to that class to see what errors may be preventing retrieval of a connector.

However, with public user enabled, failure to retrieve an existing connector should not cause failure to login since we can always create a new connection from username and password in settings.

Probably don't want to include this in 5.3.3 release?

cc @joshmoore @chris-allan @jburel

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/filesessionstore.py
+++ b/components/tools/OmeroWeb/omeroweb/filesessionstore.py
@@ -1,7 +1,6 @@
 import datetime
 import errno
 import logging
-import traceback
 import os
 import shutil
 import tempfile
@@ -117,7 +116,7 @@ class SessionStore(SessionBase):
                 logger.debug("No file_data for session: %s" %
                              self._key_to_file())
         except (IOError, SuspiciousOperation):
-            logger.debug(traceback.format_exc())
+            logger.debug("Failed to load session data", exc_info=True)
             self.create()
         return session_data
 
@@ -195,7 +194,7 @@ class SessionStore(SessionBase):
                     os.unlink(output_file_name)
 
         except (OSError, IOError, EOFError):
-            logger.debug(traceback.format_exc())
+            pass
 
     def exists(self, session_key):
         return os.path.exists(self._key_to_file(session_key))
--- a/components/tools/OmeroWeb/omeroweb/filesessionstore.py
+++ b/components/tools/OmeroWeb/omeroweb/filesessionstore.py
@@ -1,6 +1,7 @@
 import datetime
 import errno
 import logging
+import traceback
 import os
 import shutil
 import tempfile
@@ -17,6 +18,8 @@ from django.contrib.sessions.exceptions import InvalidSessionKey
 # Aleksandra Tarkowska:
 # This is temporary solution to fix clearout of expiered sessions
 # See: https://code.djangoproject.com/ticket/22938
+
+logger = logging.getLogger(__name__)
 
 
 class SessionStore(SessionBase):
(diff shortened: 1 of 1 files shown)


components/tools/OmeroWeb/omeroweb/filesessionstore.py
```diff
@@ -97,9 +100,9 @@ def load(self):
                     session_data = self.decode(file_data)
                 except (EOFError, SuspiciousOperation) as e:
                     if isinstance(e, SuspiciousOperation):
-                        logger = logging.getLogger(
+                        l = logging.getLogger(
```
joshmoore, line 102: Good catch!

joshmoore:
Just a heads up that though this looks sensible, it will only help to track down the issue in the demo case as opposed to the IDR case (where the redis session store is used)

components/tools/OmeroWeb/omeroweb/filesessionstore.py
```diff
@@ -110,7 +113,11 @@ def load(self):
                     session_data = {}
                     self.delete()
                     self.create()
+            else:
+                logger.debug("No file_data for session: %s" %
+                             self._key_to_file())
         except (IOError, SuspiciousOperation):
+            logger.debug(traceback.format_exc())
```
chris-allan: The `exc_info` attribute is always available to the logger so there is no real need to import `traceback` and call `format_exc()` here.

components/tools/OmeroWeb/omeroweb/filesessionstore.py
```diff
@@ -184,7 +195,7 @@ def save(self, must_create=False):
                     os.unlink(output_file_name)
 
         except (OSError, IOError, EOFError):
-            pass
+            logger.debug(traceback.format_exc())
```
chris-allan: The `exc_info` attribute is always available to the logger so there is no real need to import `traceback` and call `format_exc()` here.

jburel:
on return @will-moore could you go back to this PR?
re-activate, close?

will-moore:
Not sure how to test this PR, unless we have a system that we know is giving errors in reading from session.

chris-allan:
Do you currently have a way of reproducing the errors locally, @will-moore? Is it sufficient to deliberately corrupt files in `/tmp`?

chris-allan:
Are you sure you want to silently ignore `OSError`, `IOError`, and `EOFError` now?

will-moore:
Added logging back for ```save()``` too.

will-moore:
No, I don't have a way of reproducing these errors locally.

chris-allan:
Okay, so if you delete your current sessions from `/tmp`, log in, actively corrupt the session file in `/tmp`, and then attempt another request can you reproduce it?

will-moore:
On login in web, OMEROweb.log shows
```
2017-08-10 11:50:17,486 DEBUG [               omeroweb.filesessionstore] (proc.86948) save():143 Save session to file with session_file_name: /var/folders/b5/_dnzdy750_s53w_21t6lvf61cn9pqx/T/sessionidqna2x6nluc7zyzjk5jxaka4sd6gmddt3
```

If I delete this file, subsequent requests are not able to get the session data (use is logged out) and I see:
```
2017-08-10 11:50:54,152 DEBUG [               omeroweb.filesessionstore] (proc.86948) load():119 Failed to load session data
  File "/Users/wmoore/Desktop/OMERO/openmicroscopy/dist/lib/python/omeroweb/filesessionstore.py", line 93, in load
    with open(self._key_to_file(), "rb") as session_file:
IOError: [Errno 2] No such file or directory: '/var/folders/b5/_dnzdy750_s53w_21t6lvf61cn9pqx/T/sessionidqna2x6nluc7zyzjk5jxaka4sd6gmddt3'
```

So this seems to be working fine.

manics:
Testing with https://github.com/openmicroscopy/omero-web-docker
```
docker build --build-arg OMERO_VERSION=OMERO-DEV-merge-build -t test-omero-web-docker .
docker run -it --rm -p 4080:4080 --add-host omero:<ip-of-omero-server> -e CONFIG_omero_web_debug=true test-omero-web-docker
```
Works as described.

chris-allan (approved):

PR #4956, opened 2016-11-21
make search GET only

atarkowska:
Partial rebase of https://github.com/openmicroscopy/openmicroscopy/pull/4900 (one commit only as this commit was removed from https://github.com/openmicroscopy/openmicroscopy/pull/4923)

---

# What this PR does

This PR make general search GET only

--exclude (breaking robot tests)

PR #5338, opened 2017-06-21
Search get only

will-moore:
# What this PR does

Uses ```GET``` for webclient searches, allowing search pages to be bookmarked, cached etc.
This PR is based on Ola's: #4956.

# Testing this PR

1. Robot tests cover most search scenarios (updated for changes in this PR).

 2. Try searching with the top Search field in main webclient pages: the search query will be shown in the url.

 3. Searching with the search form will load the results using a url that includes the search query (allowing it to be cached). Check that search works as expected and you can use the browser devtools "Network" tab to check the url itself if desired.

# Related reading

See https://github.com/openmicroscopy/openmicroscopy/pull/4956.
cc @joshmoore


will-moore:
I'm assuming this won't go into 5.3.3 (too late).
Should I close or exclude till after the release? @jburel.

jburel:
no need to exclude it unless it is going to break everything :-)

jburel:
The query is not in the url when searching from the search page
it is only when I search from the top Search field

will-moore:
@jburel Ah, yes - I think it's my description that needs modifying there. The url doesn't change when you're searching in the search form, so I guess you can't bookmark those searches. So the only benefit we have of using GET in that case is for caching of the search results, since the query is now in the url.
I'll update the description...

jburel:
The search robot tests are passing 
The 2 failures are not related to this PR

joshmoore:
Cheers @will-moore Pointed https://trello.com/c/nCgYaayI/361-linkable-get-url-for-search to follow-on tasks on https://trello.com/c/io5JdIAO/79-critical-link-to-search-results

manics:
--rebased-to #5378

PR #5341, opened 2017-06-23
offer query caching to clients

mtbc:
# What this PR does

Offers via the `Parameters` object a `cache()` method for enabling caching of queries by Hibernate. Applies to the query service's `findByQuery`, `findAllByQuery`, `projection` methods.

This is an *experimental* offering that may be included in OMERO 5.4 but removed in OMERO 5.5.

# Testing this PR

Requires repeated performance testing of challenging queries both with and without using `cache()` for the parameters object in client scripts.

# Related reading

https://trello.com/c/7YxTcHyA/23-expose-query-caching

### DIFF
--- a/components/blitz/resources/omero/System.ice
+++ b/components/blitz/resources/omero/System.ice
@@ -95,12 +95,16 @@ module omero {
      *  acquisition...:= whether or not acquisitionData (objectives, etc.)
      *                  should be loaded
      *
+     * cacheable      := whether or not the query is cacheable by Hibernate
+     *                   (use with caution: caching may be counterproductive)
      **/
     class Options
     {
       omero::RBool  leaves;
       omero::RBool  orphan;
       omero::RBool  acquisitionData;
+      ["deprecated:experimental: may be wholly removed in next major version"]
+      omero::RBool  cacheable;
     };
 
     /**
--- a/components/blitz/src/omero/sys/ParametersI.java
+++ b/components/blitz/src/omero/sys/ParametersI.java
@@ -1,9 +1,6 @@
 /*
- *   $Id$
- *
  *   Copyright 2008 Glencoe Software, Inc. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
- *
  */
 
 package omero.sys;
@@ -358,7 +355,7 @@ public class ParametersI extends omero.sys.Parameters {
         return null;
     }
 
-    // ~ Parameters.theOption.leaves, orphan, acquisitionData
+    // ~ Parameters.theOption.leaves, orphan, acquisitionData, cacheable
     // =========================================================================
 
     /**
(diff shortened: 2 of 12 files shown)


pwalczysko:
@bramalingam will deliver his script/query and I can connect to cytohack server to test this later.

pwalczysko:
As this PR needs to have for its testing both server and client 5.4. versions, it would be better to wait for ``roles`` branch to become ``develop`` which should be imminent.

After this, take ``team/pwalczysko/script-from-Balaji`` script and edit the server and login credentials.
Then run the script, probably against eel, with and without the ``cache`` bit.

joshmoore:
First thing noticed while testing is that it would be nice to have `cache()` in ParametersI.py as well.

joshmoore:
Still trying to find a query where this made a big difference...

```
Cached: [452.55174112319946, 343.43770313262939, 342.61567807197571]
Uncached: [372.41474103927612, 352.2955219745636, 346.74935388565063]
```

```
#!/usr/bin/env python

import omero
import omero.all
from omero.cli import cli_login
from omero.rtypes import unwrap
from time import time

q = (
   "select count(*) from Annotation a1, Annotation a2, Annotation a3 "
   "where a1.textValue = a2.textValue and a1.textValue = a3.textValue"
)
p = omero.sys.ParametersI()
p.theOptions = omero.sys.Options()
m = {"omero.group": "-1"}

res = {True: [], False: []}
with cli_login() as cli:
    c = cli.get_client()
    s = c.sf.getQueryService()
    for x in range(3):
        for y in (True, False):
            p.theOptions.cacheable = omero.rtypes.rbool(y)
            start = time()
            rv = unwrap(s.projection(q, p, m))
            print rv
            stop = time()
            res[y].append(stop-start)
print "Cached:", res[True]
print "Uncached:", res[False]
```

mtbc:
@bramalingam has his big map annotation query in one of his scripts.

mtbc:
It's nice to see that it's having any useful effect on repeated queries. :smiley:

pwalczysko:
@bramalingam 's script and setup is the only one which could help here apparently - but the debugging and ``transfer`` from IDR (5.3.x, IDR specific annotations setup) to eel is far from trivial. 
@bramalingam kindly suggested to take over the setup and the testing of this PR from now on

bramalingam:
- Simple Query: "select distinct new map( mv.value as value) from Annotation as a join a.mapValue as mv where lower(mv.value) = :value and mv.name = :name"

        - Cached:True
        - Time Taken:0.0105600357056 

        - Cached:False 
        - Time Taken:0.00676918029785 

- Complex Query: "select distinct new map(mv.name as name, mv.value as value, mv2.name as name2, mv2.value as value2) from Image as i join i.annotationLinks as ial join i.annotationLinks as ial2 join ial.child as a join a.mapValue as mv join ial2.child as a2 join a2.mapValue as mv2 where a.ns = :ns and a2.ns = :ns2 and mv.value in (:values) and mv.name = :name and i in (select image from WellSample where well.plate in (select child from ScreenPlateLink where parent.id = 1251))"

       - Cached:True 
       - Time Taken:0.0135080814362 
      
       - Cached:False 
       - Time Taken:0.0132989883423 


Matlab code used for setting up database and Python script used to test the queries. 
Looks like the performance improves in simple queries which return a lot of data, doesn't make a lot of a difference for complex queries which return minimal data.

@mtbc @pwalczysko @joshmoore

mtbc:
It looks like it took longer for the simple query? Are these times for initial or repeat queries? I'm a bit confused but thank you for experimenting!

bramalingam:
reformatting it to the same format as josh (3 iterations for each query, cached first and then uncached next):

```
Simple Query Cached: [0.012197017669677734, 0.01100015640258789, 0.008594036102294922]
Simple Query Uncached: [0.00928807258605957, 0.007967948913574219, 0.00954294204711914]
Complex Query Cached: [0.018158912658691406, 0.017222881317138672, 0.017548084259033203]
Complex Query Uncached: [0.01773691177368164, 0.017360925674438477, 0.0191648006439209]
```

sbesson:
Discussed with @pwalczysko @mtbc @dominikl  and @joshmoore. Agreement is that there is no harm in merging this although we might have to come up with a different strategy to substantially improve large queries in which case this might have to be deprecated.

PR #5344, opened 2017-06-27
merge v5.3.3 into roles

mtbc:
# What this PR does

Merges `v5.3.3` in ready for tagging `v5.4.0-m1`.


mtbc:
Merging after discussion with @jburel. Will then tag and open against `develop` with `--breaking`.

PR #5348, opened 2017-06-28
Some bugfixes for the colorpicker

dominikl:
# What this PR does

Some bugfixes for the ColorPicker

# Testing this PR

Check that 'Preview', 'Revert', 'Cancel' and 'Accept' actions work like expected.

screen shot 2017-06-28 at 11 27 16

# Related reading

https://www.openmicroscopy.org/qa2/qa/feedback/17770/

https://trello.com/c/kOxAdxvG/13-insight

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/ui/colourpicker/RGBControl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/ui/colourpicker/RGBControl.java
@@ -392,7 +392,7 @@ class RGBControl
 	
     /** Fires Changed event to all listeners stating the model has changed. */
     void fireChangeEvent() {
-        if(CommonsLangUtils.isNotEmpty(model.getLUT()))
+        if (CommonsLangUtils.isNotEmpty(model.getLUT()))
             fireChangeEvent(PaintPotUI.LUT_PROPERTY);
         else
             fireChangeEvent(PaintPotUI.COLOUR_CHANGED_PROPERTY);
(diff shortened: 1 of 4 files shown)


components/insight/SRC/org/openmicroscopy/shoola/util/ui/colourpicker/RGBControl.java
```diff
@@ -392,7 +392,10 @@ boolean isOriginalRevInt() {
 	
     /** Fires Changed event to all listeners stating the model has changed. */
     void fireChangeEvent() {
-        fireChangeEvent(PaintPotUI.COLOUR_CHANGED_PROPERTY);
+        if(CommonsLangUtils.isNotEmpty(model.getLUT()))
```
mtbc: space after "if"

pwalczysko:
@dominikl : We identified still a lingering problem.

  - the Revert button should bring you to the state before you opened the color picker
  - the channel you are on has blue color and is not reverted
  - open the color picker on 3channel image (only the channel you are viewing is on)
  - check the Reverse  checkbox and click on a LUT
  - click Preview -> the display in Preview changes as expected
  - click Reverse -> the display in Preview does not go back to the "blue" state -> not expected
  - click Cancel -> the color picker disappears, but the Preview goes to blue state (belatedly)

Starting from a LUT before you open the color picker for the first time:
 
   - be on "union jack" LUT
   - open color picker, select "Orange hot" LUT and 
   - check Reverse checkbox
   - click Preview -> all as expected
   - click Revert -> reverts to "union jack" as expected
   - click Cancel -> reverts to "Orange hot" -> not expected, Cancel should do nothing and the end state is not the state before you opened the color picker

jburel:
@dominikl do you think that will be a quick fix? I do not want us to spend too much time on insight unless it is a blocker

dominikl:
Was quite a bit trickier than I thought initially, but I think I got it now. Just about to push another commit. If there are still problems, I'll leave it.

dominikl:
Ok, I close this PR, just can't get it working... :-/

PR #5349, opened 2017-06-28
merge roles into mainline develop for OMERO 5.4 series

mtbc:
# What this PR does

Merges commits back from `roles` into `develop`.

# Testing this PR

CI jobs should not show regressions.

PR #5350, opened 2017-06-29
temporarily allow PostgreSQL 9.3

mtbc:
Required to test on trout and eel before seabass is ready, see https://github.com/openmicroscopy/openmicroscopy/pull/5349#issuecomment-311964435.

### DIFF
--- a/components/dsl/resources/ome/dsl/psql-header.vm
+++ b/components/dsl/resources/ome/dsl/psql-header.vm
@@ -53,7 +53,7 @@ BEGIN
 
 END;$$ LANGUAGE plpgsql;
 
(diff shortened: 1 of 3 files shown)


sbesson (approved):
Thanks @mtbc. Given the requirements of the imminent 5.4.0-m1 i.e. merge the roles features into develop and the state of our CI, this revert seems like the easiest way forward for the short term. In the mid-term, we will have to come back to this as we tackle the infrastructure work for OMERO 5.4.

mtbc:
We certainly will. Created blocker card at https://trello.com/c/67TJXVuC/53-revert-postgres-version-leniency.

PR #5352, opened 2017-06-29
privileges canChgrp

will-moore:
# What this PR does

Fixes bug of Light Admin without ```canChgrp``` privilege still being able to do chgrp in webcilent.
See "crash on Move in not-your-group" checkbox in https://trello.com/c/LUzfIhaU/46-crashes-in-webclient 

# Testing this PR

1.  Create a Light Admin user without ```Chgrp``` privilege
1. User shouldn't be able to chgrp other users' data (right-click menu item "Move to Group" should be disabled.
1. Full Admin or Light Admin with ```Chgrp``` privilege should still be able to move others' data.
1. Users or Light Admins without ```Chgrp``` should still be able to move their own data.

cc @pwalczysko @mtbc


mtbc:
It's probably outwith this PR but FYI when one has the actual object's details in hand we now have a couple more can* methods that should always work: `canChgrp()` and `canChown()` at http://www.openmicroscopy.org/site/support/omero5.3-staging/developers/Server/Permissions.html#permissions-object.

will-moore:
@mtbc - It would be nice to user server logic for ```canChgrp``` but I don't see that flag on that permissions object.
This permissions object is returned from the "magic" mapping of ```project_details_permissions``` in this query:
```
        select new map(project.id as id,
               project.name as name,
               project.details.owner.id as ownerId,
               project as project_details_permissions,
               (select count(id) from ProjectDatasetLink pdl
                where pdl.parent = project.id) as childCount)
        from Project project
```
When I print it, it's just a dict:
```
{'canDelete': True, 'canAnnotate': True, 'canEdit': True, 'perm': 'rwr---', 'canLink': True}
```

mtbc:
Huh, that's interesting. Your `omero_model_Permissions_ice.py` has `canChgrp`? I wonder if there's some Pythonic thing somewhere on the way to the web client that I didn't know about. `canChgrp()` is working fine from the Java integration tests.

pwalczysko:
travis failing.

will-moore:
Ah, I didn't realise that @mtbc's PR wasn't getting included in the merged build yet.
So, currently this PR is causing a crash when you try to load another user's data in the tree.
I need to exclude this or open against roles. Currently impossible to test my other PRs.

mtbc:
My PR's in breaking: feel free to add `--breaking` and use https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-trigger/.

will-moore:
--breaking

pwalczysko:
All four tests from the header of this PR passed. All works as expected.
Although the ``canChgrp`` missing question is still to be solved, I suggest we merge this PR as it fixes the crashes in appropriate manner.

jburel:
not included in today's build
Removing breaking label so it goes along @mtbc "migration" PR

mtbc:
With #5358 that kind of query now gives me results like `{'canDelete': True, 'canAnnotate': False, 'perm': 'rw----', 'canChown': False, 'canChgrp': False, 'canLink': False, 'canEdit': True}`. (*Not* a blocker for this PR though.)

components/tools/OmeroWeb/omeroweb/webclient/tree.py
```diff
@@ -59,7 +59,8 @@ def parse_permissions_css(permissions, ownerid, conn):
     permissionsCss = [r for r in restrictions if permissions.get(r)]
     if ownerid == conn.getUserId():
         permissionsCss.append("isOwned")
-    if ownerid == conn.getUserId() or conn.isAdmin():
+    if (ownerid == conn.getUserId() or
+            "Chgrp" in conn.getEventContext().adminPrivileges):
```
sbesson: Looks good. For unification of our codebase, can this be adjusted to use the enums as in https://github.com/openmicroscopy/openmicroscopy/pull/5376/files#diff-4c592a4ec58d6f4830c88b81f8aa77a3R66

sbesson (changes requested):
One stylistic request, then this is good to merge.

will-moore:
Use canChgrp from #5358 now.
Needs retest.

pwalczysko:
The UI  retest indicated in https://github.com/openmicroscopy/openmicroscopy/pull/5352#issuecomment-319220794 and started according to the header of this PR fails on:

  - full admin has no "Move" menu item enabled on any objects of another user (user-4, who is a member of at least 2 groups)
  - the same for light admin  with ``Chgrp`` privilege

will-moore:
For some reason, the functionality added in #5358 has stopped working and the ```project_details_permissions``` map queries return a permissions dict that is missing the ```canChgrp``` and ```canChown``` flags.
This is why @pwalczysko's retest failed, but we also see Robot test failures etc.

mtbc:
I haven't yet managed to reproduce the *missing* flags locally (via `bin/omero hql --all ...`) but I'll keep investigating.

mtbc:
So, while I *am* always getting those map entries, they are wrongly `false` even for `root` so they may not be of much help anyway. This fell through the cracks because I had no idea that `_details_permissions` was possible. I created https://trello.com/c/QbdB9M8k/115-canchgrp-canchown-bug and am not yet easily seeing a fix.

will-moore:
@mtbc Before this PR, there was no check that the permissions loaded like this:
```
perms2 = unwrap(reader.projection((
                "select new map(p.id as id, p as p_details_permissions) "
                "from Project p where p.id = :id"),
                ParametersI().addId(project.id.val)))[0][0]
```
were correct, partly because the ```ProjectionFixtures``` at https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/testlib/__init__.py#L1132 didn't have the ```canChgrp``` and ```canChown``` added to them, and because the ```def assertPerms(perms, fixture)``` used in the test didn't check for ```canChgrp``` and ```canChown```. Adding these checks on the 1st August caused the tests to start failing.

will-moore:
@mtbc I'll close this until we have canChgrp flag working in ```_details_permissions``` dict, so we don't have so many tests failing... cc @joshmoore

joshmoore:
Ooops. Forgot to point out that this can likely be re-opened with https://github.com/openmicroscopy/openmicroscopy/pull/5437 though at this point we can await the status of this mornings integration tests: https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration/748/

will-moore:
Reopening this to test against #5437.

will-moore:
Test permissions now passing with #5437 https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/630/testReport/OmeroPy.test.integration.test_permissions/

components/tools/OmeroWeb/test/unit/test_tree.py
```diff
@@ -185,7 +187,7 @@ def test_parse_permissions_css(
             # Test with matching owner_ids, which means
             # isOwned and canChgrp is True
             expected.append('isOwned')
-            expected.append('canChgrp')
+            # expected.append('canChgrp')
```
joshmoore: Can this be removed?

joshmoore:
@sbesson : I assume you're happy with this?

sbesson (approved):

mtbc (approved):
Great to see all the tests.

PR #5354, opened 2017-07-02
Chown all

pwalczysko:
# What this PR does

Implements the feature of ``Chown2`` command which allows to transfer ownership of all data owned by user(s) to another user in one step on the CLI.

# Testing this PR
This is just a suggestion. 

Will write integration test when this approach is okayed. 

The things to test are

``bin/omero chown 6 Image:3,4 Dataset:6,9 Project:1-4,6`` should work as previously, i.e. the named objects will be chowned to user 6.
``bin/omero chown 6 Experimenter:1,2,5-8`` shoiuld chown all what users 1,2,5,6,7,8 own, but run ``dry-run`` by default, which has to be overcome by ``--force`` (I think similar logic is on ``delete`` and it makes sense for me to have it here too)
``bin/omero chown 6 Experimenter:1,2`` shoiuld chown all what users 1,2 own and execute immeditately (no need for ``--force``
``bin/omero chown 6  Image:3,4 Dataset:6,9 Project:1-4,6 Experimenter:4`` should give an output ``Cannot combine commands`` and no chown shoiuld proceed at all.

All the integration tests should still pass.

Note that this command is a bit "out of the range" with respect of CLI structure, because it 
1. needs the "obj" machinery to work as it was up till now
2. needs some "user" machinery to cater for the fact that here, users are being passed in place of objects like P/D/Is

I hope I made the best compromise without too much hacking.

# Related reading

https://trello.com/c/eaTcjBfW/194-cli-chown-hand-over-all-users-data

@joshmoore @jburel @mtbc 

Superseded by https://github.com/openmicroscopy/openmicroscopy/pull/5357

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1580,6 +1580,7 @@ class GraphArg(object):
     def __call__(self, arg):
         cmd = self.cmd_type()
         targetObjects = dict()
+        targetUsers = []
         try:
             parts = arg.split(":", 1)
             assert len(parts) == 2
@@ -1597,8 +1598,17 @@ class GraphArg(object):
                     ids.extend(range(low, high+1))
                 else:
                     ids.append(long(id))
-            targetObjects[graph[0]] = ids
+            # Allow also experimenters to be processed by this method
+            # Note thoiugh that Expermienter cannot be combined with
+            # other graph objects in the combine_commands method.
+            #.Either only Experimenter is used in the arguments or
+            # other objects (like Datasets, Images etc.)
+            if "Experimenter" in graph:
+                targetUsers = ids
+            else:
+                targetObjects[graph[0]] = ids
             cmd.targetObjects = targetObjects
+            cmd.targetUsers = targetUsers
             if len(graph) > 1:
                 skiphead = omero.cmd.SkipHead()
                 skiphead.request = cmd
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/src/omero/cli.py
```diff
@@ -1597,8 +1598,17 @@ def __call__(self, arg):
                     ids.extend(range(low, high+1))
                 else:
                     ids.append(long(id))
-            targetObjects[graph[0]] = ids
+            # Allow also experimenters to be processed by this method
+            # Note thoiugh that Expermienter cannot be combined with
+            # other graph objects in the combine_commands method.
+            # Either only Experimenter is used in the arguments or
+            # other objects (like Datasets, Images etc.)
```
mtbc, line 1605: Why make it either/or? Too tricky to implement otherwise? (The server code can cope with a combination.)

pwalczysko:
``bin/omero chown 4 Expermenter:1 Experimenter:2``
and
``bin/omero chown 4 Image:1 Image:2 Dataset:3,4``
must be tested as well

pwalczysko:
The ``bad-target`` reporting of chgrp and delete was brokent by this PR when 

``bin/omero chgrp 2 Experimenter:52`` is passed, it should report ``bad-target`` but this PR causes that the command apparently "passes" with OK output

Will have to reconsider.

mtbc:
So, I'm confused about how best to implement this:

* I think that the current hacking of things before `combine_commands` will break if we specify `Experimenter:{1,2}` (note the braces).

* Your checking suggests that it wrongly allows `Experimenter:1` to now be passed to the `chgrp` plugin.

Here's an implementation suggestion: revert your `cli.py` changes and instead, given that in the `main_method` we do,

```python
        for command_check in commands:
            self._check_command(command_check)
```

then in `ChownControl` override the `_check_command` to,

1. Call its overridden method from the superclass, then,
1. Move any IDs for `"Experimenter"` in `command_check.targetObjects` *out* of that dictionary and into `targetUsers` instead.

Alternative suggestions from someone more Python-competent (@joshmoore, @sbesson, @jburel?) are most welcome as I am just guessing here.

pwalczysko:
@mtbc 
>iven that in the main_method we do,

You mean the main_method of class GraphControl

joshmoore:
Another alternative would be to not try to "overload" the term `chown` and instead use something like `transfer`.

pwalczysko:
Superseded by https://github.com/openmicroscopy/openmicroscopy/pull/5357

PR #5357, opened 2017-07-03
Enable chown-all-belonging-to-user on CLI in and/or way with objects

pwalczysko:
# What this PR does

Implements the feature of ``Chown2`` command which allows to transfer ownership of all data owned by user(s) to another user in one step on the CLI.

# Testing this PR
This is just a suggestion. 

Will write integration test when this approach is okayed. 

The things to test are

``bin/omero chown 6 Image:3,4 Dataset:6,9 Project:1-4,6`` should work as previously, i.e. the named objects will be chowned to user 6.
``bin/omero chown 6 Experimenter:1,2,5-8`` shoiuld chown all what users 1,2,5,6,7,8 own, but run ``dry-run`` by default, which has to be overcome by ``--force`` (I think similar logic is on ``delete`` and it makes sense for me to have it here too)
``bin/omero chown 6 Experimenter:1,2`` shoiuld chown all what users 1,2 own and execute immeditately (no need for ``--force``
``bin/omero chown 6  Image:3,4 Dataset:6,9 Project:1,4,6 Experimenter:4`` should chown the data as specified
``bin/omero chown Experimenter:0 Experimenter:7`` should chown the data of user 7 to user 0
``bin/omero chown Image:1000`` where image 1000 does not exist should fail
``bin/omero delete Experimenter:4`` should fail
``bin/omero chown 5 Experimenter:4 Expermenter:19 --ordered`` should chown the data of users 4 and 19 to user 5 in two separate commands
``bin/omero chown 5 Experimenter:17 Dataset:1,5 Expermienter:19`` shoiuld chown the data of users 17, 19 and Datasets 1,5 to user5

All the integration tests should still pass.

# Related reading

https://trello.com/c/eaTcjBfW/194-cli-chown-hand-over-all-users-data

@joshmoore @jburel @mtbc 

This supersedes https://github.com/openmicroscopy/openmicroscopy/pull/5354 - all the [suggestions](https://github.com/openmicroscopy/openmicroscopy/pull/5354#issuecomment-312654309) of @mtbc are implemented in this PR

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/chown.py
+++ b/components/tools/OmeroPy/src/omero/plugins/chown.py
@@ -87,7 +87,7 @@ class ChownControl(GraphControl):
                  " and/or user(s) to transfer all data from in the"
                  " form Experimenter:<Id>")
 
-    def populate_targetUsers(self, command_check):
+    def populate_target_users(self, command_check):
         """
         Move the Experimenters whose data are to be
         transferred from targetObjects
@@ -107,7 +107,7 @@ class ChownControl(GraphControl):
         as well as the possibility to adjust the command.
         """
         super(ChownControl, self)._check_command(command_check)
-        self.populate_targetUsers(command_check)
+        self.populate_target_users(command_check)
 
     def _process_request(self, req, args, client):
         # Retrieve user id
--- a/components/tools/OmeroPy/src/omero/plugins/chown.py
+++ b/components/tools/OmeroPy/src/omero/plugins/chown.py
@@ -14,7 +14,6 @@
 """
 
 from omero.cli import CLI, GraphControl, ExperimenterArg, GraphArg
-from omero_ext.argparse import SUPPRESS
 import sys
 
 HELP = """Transfer ownership of data between users. Entire graphs of data,
@@ -75,46 +74,18 @@ class ChownControl(GraphControl):
         import omero.all
         return omero.cmd.Chown2
 
-    def _configure(self, parser):
-        parser.set_defaults(func=self.main_method)
-        self._add_wait(parser, default=-1)
-        parser.add_argument(
-            "--include",
-            help="Modifies the given option by including a list of objects")
-        parser.add_argument(
-            "--exclude",
-            help="Modifies the given option by excluding a list of objects")
-        parser.add_argument(
-            "--ordered", action="store_true",
-            help=("Pass multiple objects to commands strictly in the order "
-                  "given, otherwise group into as few commands as possible."))
-        parser.add_argument(
-            "--list", action="store_true",
-            help="Print a list of all available graph specs")
-        parser.add_argument(
-            "--list-details", action="store_true", help=SUPPRESS)
-        parser.add_argument(
-            "--report", action="store_true",
-            help="Print more detailed report of each action")
-        parser.add_argument(
-            "--dry-run", action="store_true",
-            help=("Do a dry run of the command, providing a "
-                  "report of what would have been done"))
-        parser.add_argument(
-            "--force", action="store_true",
-            help=("Force an action that otherwise defaults to a dry run"))
-        self._pre_objects(parser)
-        parser.add_argument(
-            "obj", nargs="*", type=GraphArg(self.cmd_type()),
-            help="objects to be processed in the form <Class>:<Id>"
-                 " and/or user(s) to transfer all data from in the"
-                 " form Experimenter:<Id>")
-
     def _pre_objects(self, parser):
         parser.add_argument(
             "usr", nargs="?", type=ExperimenterArg,
             help="user to transfer ownership of specified objects and/or all"
                  " objects owned by specified user(s) to")
+
+    def _objects(self, parser):
+        parser.add_argument(
+            "obj", nargs="*", type=GraphArg(self.cmd_type()),
+            help="objects to be processed in the form <Class>:<Id>"
+                 " and/or user(s) to transfer all data from in the"
+                 " form Experimenter:<Id>")
 
     def populate_targetUsers(self, command_check):
         # Move the Experimenters whose data are to be
@@ -122,12 +93,11 @@ class ChownControl(GraphControl):
         # to targetUsers. The rewritten _check_command
         # method checked these Experiments when they were
         # in targetObjects, but Chown2 needs them in targetUsers
-        for k, v in command_check.targetObjects.items():
-            if k == "Experimenter":
-                if not v:
-                    return
-                command_check.targetUsers = v
-                del command_check.targetObjects["Experimenter"]
+        try:
+            command_check.targetUsers = command_check.targetObjects.pop(
+                "Experimenter")
+        except KeyError:
+            pass
 
     def _check_command(self, command_check):
         # Rewrite higher class method to have command check
--- a/components/tools/OmeroPy/src/omero/plugins/chown.py
+++ b/components/tools/OmeroPy/src/omero/plugins/chown.py
@@ -23,6 +23,9 @@ There are two ways to use this command, either specify the data to be
 transferred or specify users from whom the ownership of all their data
 will be transferred. These two usage ways can be combined.
 
+The usage with specified users has to be considered as advanced usage
+and might potentially be slow.
+
 This command can only be used by OMERO administrators and group owners.
 
 Group owners can only transfer ownership between members of the owned group.
@@ -52,8 +55,10 @@ Examples:
     # Transfer all images contained under two projects
     omero chown 101 Project/Image:201,202
 
+    # Advanced usage & potentially slow:
     # Transfer all data owned by user 111 and one image to user 4
     omero chown 4 Experimenter:111 Image:17
+    # Advanced usage & potentially slow:
     # Transfer all data of users 1,3 and 7 to user 10
     omero chown 10 Experimenter:1,3,7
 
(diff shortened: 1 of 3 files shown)


components/tools/OmeroPy/src/omero/plugins/chown.py
```diff
@@ -65,10 +75,65 @@ def cmd_type(self):
...
+            help="user to transfer ownership of specified objects and/or all"
+                 " objects owned by specified user(s) to")
+
+    def populate_targetUsers(self, command_check):
+        # Move the Experimenters whose data are to be
+        # transferred from targetObjects
+        # to targetUsers. The rewritten _check_command
+        # method checked these Experiments when they were
+        # in targetObjects, but Chown2 needs them in targetUsers
+        for k, v in command_check.targetObjects.items():
```
mtbc: There *has* to be an easier way to do this in Python, something like,

```python
if "Experimenter" in command_check.targetObjects:
    command_check.targetUsers = command_check.targetObjects["Experimenter"]
    del command_check.targetObjects["Experimenter"]
```

or somesuch. (The above may not be it, but the correct thing will be correspondingly simple, no loop.)

components/tools/OmeroPy/src/omero/plugins/chown.py
```diff
@@ -65,10 +75,65 @@ def cmd_type(self):
...
+            "--dry-run", action="store_true",
+            help=("Do a dry run of the command, providing a "
+                  "report of what would have been done"))
+        parser.add_argument(
+            "--force", action="store_true",
+            help=("Force an action that otherwise defaults to a dry run"))
+        self._pre_objects(parser)
+        parser.add_argument(
+            "obj", nargs="*", type=GraphArg(self.cmd_type()),
+            help="objects to be processed in the form <Class>:<Id>"
```
mtbc: It's really unfortunate that you have to copy and paste all the code above just to change a little bit of the help text. Does argparse let you have the superclass one run but then *change* the help text for a given parser argument? Or (just like with `_pre_objects`) could you refactor to just override just a small new helper method that `GraphControl._configure` calls?

components/tools/OmeroPy/src/omero/plugins/chown.py
simleo: Try this:

```python
try:
    command_check.targetUsers = command_check.targetObjects.pop("Experimenter")
except KeyError:
    pass
```

pwalczysko: This is the ``req`` - see below, we were wondering what type of object that (req.targetUsers) is. 
```
bin/omero chown 2 Image:1,51 Dataset:105,106 Experimenter:52,2,102 --report --dry-run
Using session fdca147e-2f7b-4a52-8294-5a28eff7d38c (root@localhost:4064). Idle timeout: 10 min. Current group: system
object #0 (::omero::cmd::Chown2)
{
    targetObjects = 
    {
        key = Image
        value = 
        {
            [0] = 1
            [1] = 51
        }
        
        key = Dataset
        value = 
        {
            [0] = 105
            [1] = 106
        }
    }
    childOptions = {}
    dryRun = True
    userId = 2
    targetUsers = 
    {
        [0] = 52
        [1] = 2
        [2] = 102
    }
}
```

pwalczysko:
This shoiuld address comments of @mtbc @simleo . Thank you very much.

joshmoore:
First kudos on the method overriding. A few more code-review comments:

 1. For comments, of a method, rather than using `#` use `"""..."""` which will be much more user-friendly. i.e. A user who wants to learn that

```
def my_method():
    """
   does something
    """
```
can use `help(my_method)`.

 2. When using underscore naming, don't introduce a capital, i.e. `populate_target_users` rather than `populate_targetUsers`

 3. Finally, I have the feeling that GraphArg should allow you to do what you want more easily (e.g. `GraphArg( ..., limit=Experimenter.class)` but that will need more investigation and can easily wait until after this PR.

From an integration testing point-of-view, my primary question would be concerning expected run times of these methods. Have any numbers been gathered?

jburel:
is it something we want to target for 5.4.0 considering the limited amount of time we have?

pwalczysko:
@jburel : I do not see anything worrying in @joshmoore 's comments - all should be done by tomorrow I guess except the performance test - but I understand that the performance test is meant to test performance of the back-end ``Chown2``  command, which we released already, albeit only in API ?

pwalczysko:
My immediate idea was to use the roles devspace for the performance tests, but I hear that @sbesson is using it/going to use it ? 
@joshmoore : The answer to whether numbers have been gathered is no (see above).

joshmoore:
> I understand that the performance test is meant to test performance of the back-end Chown2 command, which we released already, albeit only in API 

Agreed, though I think the `Experimenter` target has not really been exposed to date. One option of course would just be to put a clear warning as either "Experimental" or "Advanced usage & potentially slow".

mtbc:
Probably good to also extend `components/tools/OmeroPy/test/integration/clitest/test_chown.py` for this new functionality but the integration testing part could be in a separate 5.4.1 PR if we are sure to create the card now.

pwalczysko:
Discussing with @mtbc we came up with following performance test plan:

   - set up a new user on eel and chown all the data of user-4 to this user (really solid big graph of data, plates, P/D/Is, annotations, ROIs....)
   - set up two new users on eel, create a ``mixed ownership`` graph and watch how long does that take to process by Chown2 (when targetng the user)
   - set up two new users on eel, import the plate with 14000 ROIs to this user, then chown to the other user

       - this plate
       - all that this user owns (this user owns just that big plate, thus we get a comparable numbers about what takes longer)

It does not seem that the roles devspace would be necessary (although maybe partially, for some tests it would be a "nice to have")

Note: we have also the https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroJava/test/integration/chown/PermissionsTest.java#L441 which is testing happily the "chownAll (targetUser)", albeit this is hardly a huge graph of course which is being transferred.

pwalczysko:
Created https://trello.com/c/Acn8hbEZ/11-performance-testing-of-chownallbelongingtouser for the performance testing.

Pushed two commits which should fix the two first points @joshmoore made.
Created https://trello.com/c/CWixi6UU/12-improve-structure-of-the-chown-all-on-cli-code for the third (more in-depth code improvement request by @joshmoore )

For the non-performance integration test, I will try to write it during tomorrow and add to this PR.

pwalczysko:
@joshmoore @mtbc : https://github.com/openmicroscopy/openmicroscopy/pull/5357/commits/19de9d13b9f087ead5b947fa30e1119867c818c8 adds integration test. Hopefully this should suffice at this stage, but happy to adjust or expand.

https://github.com/openmicroscopy/openmicroscopy/pull/5357/commits/3d9dab55106bdb39263202a7eb4a461470949150 adds the slowness warnings as requested by @joshmoore

mtbc:
The new integration tests pass. :+1: FYI they take about ten seconds per run (parameter).

components/tools/OmeroPy/test/integration/clitest/test_chown.py
```diff
@@ -89,7 +88,7 @@ def testChownBasicUsageTargetUser(self, simpleHierarchy, user_prefix):
         # and trasfer all owned by user1 first step to user2 and
```
mtbc: transfer

components/tools/OmeroPy/src/omero/plugins/chown.py
```diff
@@ -48,6 +55,13 @@
     # Transfer all images contained under two projects
     omero chown 101 Project/Image:201,202
 
+    # Advanced usage & potentially slow:
+    # Transfer all data owned by user 111 and one image to user 4
+    omero chown 4 Experimenter:111 Image:17
+    # Advanced usage & potentially slow:
+    # Transfer all data of users 1,3 and 7 to user 10
```
mtbc: space after comma

components/tools/OmeroPy/test/integration/clitest/test_chown.py
```diff
@@ -58,6 +58,56 @@ def testChownBasicUsageWithId(self, object_type, user_prefix):
         assert obj.id.val == oid
         assert obj.details.owner.id.val == user.id.val
 
+    def testChownBasicUsageTargetUser(self, simpleHierarchy):
+        proj, dset, img = simpleHierarchy
+        argument = "Experimenter"
+
+        # create a new user  in the same group
```
mtbc: double space

components/tools/OmeroPy/test/integration/clitest/test_chown.py
```diff
@@ -58,6 +58,56 @@ def testChownBasicUsageWithId(self, object_type, user_prefix):
...
+        obj = self.query.get('Image', img.id.val, all_grps)
+        assert obj.id.val == img.id.val
+        assert obj.details.owner.id.val == user1.id.val
+
+        # create dataset
+        object_name = "Dataset"
+        oid = self.create_object(object_name)
+
+        # now create a second user (user2) in the same group
+        # and trasfer all owned by user1 first step to user2 and
```
mtbc: transfer

mtbc:
Functionally seems to be working fine and code refactoring is certainly an improvement.

pwalczysko:
These should fix Mark's comments.

mtbc (approved):

sbesson:
Thumbs up from everyone. Only follow-up might be to document the new features in the SPhinx documentation as part of https://trello.com/c/JWI2G9rm/35-document-chown-for-cli

mtbc:
Re. https://github.com/openmicroscopy/openmicroscopy/pull/5357#issuecomment-313106351 see https://trello.com/c/Acn8hbEZ/.

PR #5358, opened 2017-07-04
Further support for canChgrp, canChown

mtbc:
# What this PR does

Further supports the new `canChgrp`, `canChown` methods.

# Testing this PR

Code review and CI suffice.

# Related reading

#5204 and https://trello.com/c/ENmwK2ru/16-can-methods-on-client-side-object-permissions


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -671,7 +671,14 @@ def canChgrp(self):
         group. Web client will only allow this for the data Owner. Admin CAN
         move other user's data, but we don't support this in Web yet.
         """
-        return self.isOwned() or self._conn.isAdmin()   # See #8974
+        return self.getDetails().getPermissions().canChgrp()
```
joshmoore, line 674: This is probably the only change here that requires significant testing.

joshmoore (commented):
Assuming all integration tests are sufficiently happy with the one noted change in BlitzGateway, the changes here look good to me.

components/tools/OmeroPy/src/omero/gateway/__init__.py
mtbc: You're thinking, try out "move" under different permissions situations in webclient, see if it's grayed out when it should be and works properly when not?

joshmoore: Guess I'm mostly worried if `canChgrp` ever got used as a proxy for something else that now differs from the server-side logic.

mtbc: Interesting point. The previous condition is,
```python
self.isOwned() or self._conn.isAdmin()
```
and I've documented `Chgrp2` as, "The user must be either an administrator, or the owner of the objects and a member of the target group."

While `canChgrp` will return `false` for things like `FilterSet` or `DatasetImageLink` (wants the `Dataset` or `Image` selected) it should return exactly as the previous condition for the kinds of things that are selectable by the user in the web (I think this is used mostly for the left-hand tree view). So seeing as CI looks fairly happy I'd guess we're probably safe; I wonder if there's anything else specific we should test.

mtbc: cc: @will-moore so he can come back and raise any issues he realizes this may affect.

joshmoore:
:+1: Mentioned in call. No blocker from me, mostly just a heads up for further release testing.

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel: Ref to this PR in https://trello.com/c/N0SpXfr8/54-canchgrp-canchown-for-web

jburel:
Merging before tagging

PR #5359, opened 2017-07-05
CLI: require stdin.isatty before using cli.input

joshmoore:
The cli.input method currently blindly calls either
raw_input or getpass both of which hang if stdin is
piped into the bin/omero command. A further fix here
would be to teach the input method itself how to check
stdin, but this require a more further review.

See
https://trello.com/c/mUAfr6kf/200-bin-omero-should-fail-if-not-logged-in-and-no-stdin

# What this PR does

fails fast if `echo foo | bin/omero login ...` is used.

# Testing this PR

The following should now all fail with a useful comment:

 * `echo foo | bin/omero login`
 * `echo foo | bin/omero login localhost`
 * `echo foo | bin/omero login root@localhost`

but *not*

 * `echo foo | bin/omero -w mypassword login root@localhost`

# Related reading

* https://trello.com/c/mUAfr6kf/200-bin-omero-should-fail-if-not-logged-in-and-no-stdin

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -448,6 +448,9 @@ class SessionsControl(BaseControl):
                 action = "Reconnected to"
 
         if not rv:
+
+            self._require_tty("cannot request password")
+
             tries = 3
             while True:
                 try:
@@ -883,6 +886,7 @@ class SessionsControl(BaseControl):
         defserver = store.last_host()
         if not port:
             port = str(omero.constants.GLACIER2PORT)
+        self._require_tty("cannot request server")
         rv = self.ctx.input("Server: [%s:%s]" % (defserver, port))
         if not rv:
             return defserver, name, port
@@ -892,11 +896,17 @@ class SessionsControl(BaseControl):
     def _get_username(self, defuser):
         if defuser is None:
             defuser = get_user("root")
+        self._require_tty("cannot request username")
         rv = self.ctx.input("Username: [%s]" % defuser)
         if not rv:
             return defuser
         else:
             return rv
+
+    def _require_tty(self, msg):
+        if sys.stdin.isatty():
+            return
+        self.ctx.die(564, "stdin is not a terminal: %s" % msg)
 
 
 try:
--- a/components/tools/OmeroPy/src/omero/plugins/sessions.py
+++ b/components/tools/OmeroPy/src/omero/plugins/sessions.py
@@ -449,7 +449,10 @@ class SessionsControl(BaseControl):
 
         if not rv:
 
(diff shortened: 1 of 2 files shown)


sbesson:
Generally looks sensible. The following plugins are using `omero.cli.CLI.input()` at the moment: `admin`, `db`, `hql`, `obj`, `user`, `sessions`, `script`, `tag`.
Moving the `_require_tty()` check at the `CLI.input()` method rather than at each plugin level sounds like the proper way to go but as mentioned in the description this might require a precise assessment and review of the entire set of affected commands.

Another question is whether `self.ctx.isquiet` should also be taken into account in this check. In `hql`, the `-q` option fails interactive commands while in `omero sessions` it is limited to reducing the logging verbosity so far.

joshmoore:
> require a precise assessment and review of the entire set of affected commands.

If/when we do that, it might then also be worth investing in conditional logic to read from stdin in a non-blocking way so that _if_ there's enough info, it gets used. (This doesn't look to be trivially cross-platform.)

> Another question is whether self.ctx.isquiet should also be taken into account

i.e. potentially only fail on `quiet and not isatty`? That would amount to not changing the breaking behavior unless users add the `-q`. Could well make sense for some of the other plugins, but as long as this is limited to login, it's likely not a huge win.

components/tools/OmeroPy/src/omero/plugins/sessions.py
```diff
@@ -892,12 +899,18 @@ def _get_server(self, store, name, port):
...
         rv = self.ctx.input("Username: [%s]" % defuser)
         if not rv:
             return defuser
         else:
             return rv
 
+    def _require_tty(self, msg):
+        if sys.stdin.isatty():
+            return
+        self.ctx.die(564, "stdin is not a terminal: %s" % msg)
```
manics, line 912: FYI 564 results in exit-code 52

manics (approved):
Looks good to me

components/tools/OmeroPy/src/omero/plugins/sessions.py
joshmoore: Are you depending on the exit code for anything, @manics?

manics: No, though in future it'd be nice if these were defined somewhere so we could use them.

sbesson:
Merging that in. There is a more global review of the strategy but as `sessions` is the plugin called by almost anything, having this fixed now is valuable. Main limitation/potential breaking impact is that this does not account for people having modified their stdin. Way to handle that would be to define an environment variable (see comment in https://trello.com/c/mUAfr6kf/60-bin-omero-should-fail-if-not-logged-in-and-no-stdin). @mtbc mentions this should be captured in what's new.

PR #5363, opened 2017-07-05
OMERO.web backend SSL is controlled by OMERO.web administrator

manics:
# What this PR does

- Removes the padlock checkbox from the OMERO.web login page.
- Adds a new property `omero.web.secure` that controls whether OMERO.web uses encrypted connections to OMERO.server, defaults to `False` (unencrypted) to keep the current default behaviour.

# Testing this PR

1. Setup a default OMERO.server, block port 4063 (e.g. using a firewall).
2. Setup OMERO.web on a different host, with `omero.web.server_list=[["omero-server-address", 4064, "omero"]]`
3. Attempt to login to OMERO.web. It should fail with an unhelpful error message- this is the existing behaviour (after creating a session it tries to use unencrypted port 4063 which is inaccessible).
4. Set `omero.web.secure=True` and restart OMERO.web
5. You should now be able to login

# Related reading

- https://trello.com/c/sc79fkLw/61-drop-lock-icon-in-web

PR #5364, opened 2017-07-05
OMERO.web backend SSL is controlled by OMERO.web administrator (rebased onto metadata53)

manics:
This is the same as gh-5363 but rebased onto metadata53.

----

# What this PR does

- Removes the padlock checkbox from the OMERO.web login page.
- Adds a new property `omero.web.secure` that controls whether OMERO.web uses encrypted connections to OMERO.server, defaults to `False` (unencrypted) to keep the current default behaviour.

# Testing this PR

1. Setup a default OMERO.server, block port 4063 (e.g. using a firewall).
2. Setup OMERO.web on a different host, with `omero.web.server_list=[["omero-server-address", 4064, "omero"]]`
3. Attempt to login to OMERO.web. It should fail with an unhelpful error message- this is the existing behaviour (after creating a session it tries to use unencrypted port 4063 which is inaccessible).
4. Set `omero.web.secure=True` and restart OMERO.web
5. You should now be able to login

# Related reading

- https://trello.com/c/sc79fkLw/61-drop-lock-icon-in-web

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/decorators.py
@@ -297,7 +297,9 @@ class login_required(object):
                 server_id = settings.PUBLIC_SERVER_ID
             username = settings.PUBLIC_USER
             password = settings.PUBLIC_PASSWORD
-            is_secure = request.GET.get('ssl', False)
+            is_secure = settings.SECURE
+            if not is_secure:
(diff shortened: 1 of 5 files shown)


manics:
--rebased-from #5363

joshmoore:
Merging based on https://github.com/openmicroscopy/openmicroscopy/pull/5363#issuecomment-313621236 (tested on prod40/idr-testing)

PR #5368, opened 2017-07-06
Disable Owners and Members fields if not sufficient permissions

dominikl:
# What this PR does

If the light admin has only create/edit group permission, creating a group will throw an error when webclient_gateway attempts to add a group owner and/or member (the user needs 'add users to group' permission for that). With this PR these fields are simply ~~hidden~~ disabled, when the user lacks the appropriate permisson.

# Testing this PR

Create a light admin user which only has 'Create and Edit Group' permission. Check that ~~there are no~~ Owners and Members fields are disabled. Create a group. 
Grant the 'Add Users to Groups' permission to the light admin user. Check that now ~~there are~~ the Owners and Members fields are enabled. Create a group.

# Related reading

https://trello.com/c/LUzfIhaU/46-crashes-in-webclient

# Remarks
- ~~Disabling the fields would be nicer than hiding them... Any Javascript experts here to give me a hint?~~
- ~~I think this PR is quite 'hacky', can't imagine @will-moore would be happy to merge this...~~
- Generally, "Create and Edit Groups" permission doesn't make much sense without the "Add Users to Group" permission, does it?

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/forms.py
@@ -278,7 +278,7 @@ PERMISSION_CHOICES = (
 class GroupForm(NonASCIIForm):
 
     def __init__(self, name_check=False, group_is_current_or_system=False,
-                 can_modify_group=True, *args, **kwargs):
+                 can_modify_group=True, can_add_member=True, *args, **kwargs):
         super(GroupForm, self).__init__(*args, **kwargs)
         self.name_check = name_check
         try:
@@ -317,6 +317,11 @@ class GroupForm(NonASCIIForm):
         if not can_modify_group:
             for field in self.fields.values():
                 field.widget.attrs['disabled'] = True
+
+        # If we can't add members, disable owners and members fields
+        if not can_add_member:
+            self.fields['owners'].widget.attrs['disabled'] = True
+            self.fields['members'].widget.attrs['disabled'] = True
 
     name = forms.CharField(
         max_length=100,
--- a/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form.html
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form.html
@@ -93,7 +93,13 @@
                         "Follow the 'OMERO permissions' link below for full details.",
                         null, "WARNING", ['OK'], null, 180);
                 });
-
+                
+                if ('{{ can_add_member }}' == 'False') {
+                    $("label[for='id_owners']").hide();
+                    $("#id_owners_chosen").hide();
+                    $("label[for='id_members']").hide();
+                    $("#id_members_chosen").hide();
+                } 
         })
     </script>
 
--- a/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form.html
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form.html
@@ -93,13 +93,6 @@
                         "Follow the 'OMERO permissions' link below for full details.",
                         null, "WARNING", ['OK'], null, 180);
                 });
-                
-                if ('{{ can_add_member }}' == 'False') {
-                    $("label[for='id_owners']").hide();
-                    $("#id_owners_chosen").hide();
-                    $("label[for='id_members']").hide();
-                    $("#id_members_chosen").hide();
-                } 
         })
     </script>
 
--- a/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form.html
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/group_form.html
@@ -93,6 +93,7 @@
                         "Follow the 'OMERO permissions' link below for full details.",
                         null, "WARNING", ['OK'], null, 180);
                 });
+
         })
     </script>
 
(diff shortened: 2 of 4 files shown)


will-moore:
The 'right' way to do this (at least the way we've been doing it so far) is with Django's forms. As an example, see the usage of ```can_modify_group``` in webadmin/views.py and in ``` GroupForm ``` in webadmin/forms.py. However, the number of these parameters and the logic of which fields should be enabled/disabled with them is getting a bit crazy. The JavaScript approach in this PR actually moves some of that complexity elsewhere (which is nice) but now we have similar logic in different places...

dominikl:
Thanks @will-moore . I'll have a look at views.py/GroupForm and will implement it accordingly.

pwalczysko:
The first tes

 ``Create a light admin user which only has 'Create and Edit Group' permission. Check that there are no Owners and Members fields are disabled. Create a group.`` 

passes.

But when I try to edit the just created group (adding a description), then I get a red warning about not being an owner of that group. The action itself (the edit) seems to proceed as expected.

screen shot 2017-07-10 at 17 55 05

pwalczysko:
The second test 

``Grant the 'Add Users to Groups' permission to the light admin user. Check that now there are the Owners and Members fields are enabled. Create a group.``

fully passes.

pwalczysko:
This PR improves the behaviour. Happy for merge and deal with the problem https://github.com/openmicroscopy/openmicroscopy/pull/5368#issuecomment-314168409 in subsequent PR or creating a card for it.

dominikl:
Thanks @pwalczysko . I think I can fix that here, should be fairly easy.

dominikl:
Sorry, once again I'm stuck, I have no idea where this message https://github.com/openmicroscopy/openmicroscopy/pull/5368#issuecomment-314168409 comes from, can't find it anywhere in the code :-/ 
But this can be done in a follow up PR, if we don't want to wait with merging until @will-moore is back.

pwalczysko:
@dominikl : I found that the message is inside file ``components/server/src/ome/logic/AdminImpl.java`` on line 1524
On github this is https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/logic/AdminImpl.java#L1524
Sorry if you knew this :), I realize this might still not lead anywhere.

Happy for this to merge and create a Trello card for follow-up PR.

dominikl:
Thanks @pwalczysko . I was only looking in web code! As it's on the server side, it would make sense to fix that in a separate PR.

pwalczysko:
Created a new checklist point on https://trello.com/c/LUzfIhaU/46-crashes-in-webclient - is that a good idea ? Or does it need a fully new card @mtbc ?

mtbc:
The checklist is fine but could you be clearer about the situation? What are the users and privileges and memberships and stuff and what is being attempted via the API that wrongly fails? I can try to write an integration test that reproduces the issue - or is there an existing one that is easily adapted? Is this a blocker for m2?

dominikl:
I updated the description on the card for how to replicate the warning.

mtbc:
:+1: Do you know what web's actually doing under the hood here though? I'll try reproducing locally and figuring out from the logs and DB. :smiley:

*Update:* Nothing obvious from the logs.

will-moore:
Could this be coming from ```updatePermissions()```  https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py#L1541 ?

dominikl:
That might be the case, thanks @will-moore . Will check again, unfortunately will be too late for 'm2' build.

dominikl:
Yes, there were still some 'update permissions' and 'change group members' actions triggered when editing groups, fixed now with the last commit.

mtbc:
'update permissions' *might* still work, up to you if you want to try leaving that in, I can't remember what limits that

mtbc:
Doh, sorry, that probably separately depends on `Chmod` so ignore me. :smiley:

dominikl:
No, tried that, without https://github.com/openmicroscopy/openmicroscopy/pull/5368/commits/66865d8d1fe25e03f9c6bb256c6496ba4d8aff00#diff-7bcff1e41a9e76739c322db1ca40c091R1484 'Current user is neither...' error message.

mtbc:
Thank you for indulging my curiosity. :smiley:

dominikl:
But it doesn't make sense... what permission does one need to change a group's permissions?

mtbc:
I too am puzzled and have thus created https://trello.com/c/HmEm2FLz/81-modifygroup-enough-for-chmod. @pwalczysko, is this something you have tried yet?

dominikl:
Sorry @mtbc , forget about my comment. Commented out these two lines about permissions and didn't get the error message. Pretty sure I tested that before and I got the exception, but maybe that was some caching issue, etc. But I think it makes sense anyway, if the permissions haven't been changed there's no need to make a updatePermissions call.

mtbc:
Ah, so it looks like there's nothing further on this we need investigate?

dominikl:
Yes, seems to be fine. You can edit permissions with 'create/edit group' permission.
But with respect to this PR, might be good if @pwalczysko could re-review it, sorry for the hassle.

pwalczysko:
All works as expected (full retest). The bug https://github.com/openmicroscopy/openmicroscopy/pull/5368#issuecomment-314168409 is fixed too.

pwalczysko:
Ready to merge.

PR #5343, opened 2017-06-27
Add login_required to open_with_options

manics:
# What this PR does

See https://trello.com/c/gl7lsY2C/45-openwith-jsonp-without-loginrequired

PR #5371, opened 2017-07-07
Add login_required to open_with_options (rebased onto develop)

snoopycrimecop:
This is the same as gh-5343 but rebased onto develop.

----

# What this PR does

See https://trello.com/c/gl7lsY2C/45-openwith-jsonp-without-loginrequired

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -1681,6 +1681,7 @@ def projectDetail_json(request, pid, conn=None, **kwargs):
     return rv
 
 
+@login_required()
 @jsonp
 def open_with_options(request, **kwargs):
     """


snoopycrimecop:
--rebased-from #5343

sbesson:
@will-moore: can you take a look at this on return? Might require some input from @manics on the ways to reproduce the issue.

manics:
@sbesson this is the fix @will-moore told me to implement after he tracked down the problem

will-moore:
My original implementation of ``open_with_options()`` didn't include the ```@login_required``` decorator simply because the code doesn't require a connection to OMERO. There is no use case for needing the ```open-with``` config when you're not logged in.
However, what we're seeing here is that the ```@jsonp``` decorator expects there to always be a ```connector``` on the web session and if you are not already logged in then this won't be there.
This PR simply ensures that there is the ```connector``` on the session by using the  ```@login_required``` decorator.

IF we needed to allow access to the ```open-with``` config when you're not logged in, then we could avoid using the ```@jsonp``` decorator and instead do:
```
return JsonResponse({'open_with_options': viewers})
```
but then we don't have the same error handling or ```jsonp``` functionality.

Probably a better alternative fix is to remove the ```server_id``` functionality from the ```jsonp``` decorator since it seems to serve no purpose now (unless I'm missing something, the ```kwargs['server_id']``` is never used). Remove these lines:
```
            server_id = kwargs.get('server_id', None)
            if server_id is None:
                server_id = request.session['connector'].server_id
            kwargs['server_id'] = server_id
```

jburel:
@will-moore are you looking at your alternative suggestion in another branch?

will-moore:
No. I could do, or could be done in this PR?

jburel:
easier if you cherry-pick the commit and work on your own branch

jburel:
Investigation of suggested option is ongoing

will-moore:
This is now fixed in https://github.com/openmicroscopy/openmicroscopy/pull/5433 so we can close this PR.

PR #5373, opened 2017-07-07
have roles tests use file path in finding new image imports

mtbc:
# What this PR does

Uses import location to find imported file instead of ID counting hack.

# Testing this PR

CI suffices.

# Related reading

https://trello.com/c/B8gDofmV/70-make-roles-tests-more-thread-safe

### DIFF
--- a/components/tools/OmeroJava/test/integration/LightAdminPrivilegesTest.java
+++ b/components/tools/OmeroJava/test/integration/LightAdminPrivilegesTest.java
@@ -48,6 +48,7 @@ import omero.cmd.HandlePrx;
 import omero.cmd.graphs.ChildOption;
 import omero.gateway.util.Requests;
 import omero.gateway.util.Requests.Delete2Builder;
+import omero.grid.ImportLocation;
 import omero.grid.ManagedRepositoryPrx;
 import omero.grid.ManagedRepositoryPrxHelper;
 import omero.grid.RepositoryMap;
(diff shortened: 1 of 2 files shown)


bramalingam:
CI tests are passing for the LightAdminPrivilegesTest in today's build:
[LightAdminPriviligesTest](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/625/testngreports/integration/LightAdminPrivilegesTest/)
(4 tests alone have been skipped, not sure if these changes are affecting them)

CI tests for RolesTest have all been skipped in today's build:
[LightAdminRolesTest](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/625/testngreports/integration/LightAdminRolesTest/)
Have all been skipped, not sure if this is because of the changes.
@mtbc : could you please clarify the reason for the skips?


A small suggestion:

given the number of places the logic is used, is there a case for creating a method like , getImportedFilePath or something? (Might help in managing the edits in future?)

mtbc:
They all pass in https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/626/ which should be available later this afternoon. This morning's build was a general disaster.

In the longer term the helper method that's already used should be returning something useful from the import response instead of ignoring it so that we don't need to do this weird query at all: we shouldn't really need the file path. This PR is just focusing on making the code more thread-safe and at least already reduces the size of the repeats.

bramalingam:
Thank you for the explanation. Looks good to merge, according to me,.

sbesson:
More tests, more sensible usage. :+1:

PR #5375, opened 2017-07-10
batch original files query in repository DAO

mtbc:
# What this PR does

Batches the server's internal querying of open files by ID. May ease the use of `bin/omero admin cleanse` with ginormous directories if the query results can be passed via Ice.

# Testing this PR

Difficult: open to ideas! #5205 could be of help. `df -i` may reassure about inode exhaustion. There may be an efficient way to construct a single fileset (plate?) of very many tiny files.

# Related reading

http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-May/005987.html
http://trac.openmicroscopy.org/ome/ticket/13225

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/RepositoryDaoImpl.java
+++ b/components/blitz/src/ome/services/blitz/repo/RepositoryDaoImpl.java
@@ -492,9 +492,15 @@ public class RepositoryDaoImpl implements RepositoryDao {
             if (CollectionUtils.isEmpty(ids)) {
                 return Collections.emptyList();
             }
-            Parameters p = new Parameters();
-            p.addIds(ids);
-            return q.findAllByQuery(LOAD_ORIGINAL_FILE+"f.id in (:ids)", p);
+            final List<ome.model.core.OriginalFile> files = new ArrayList<>(ids.size());
+            final String filesById = LOAD_ORIGINAL_FILE + "f.id in (:ids)";
+            for (final List<Long> idsBatch : Iterables.partition(ids, 1024)) {
+                final Parameters params = new Parameters().addIds(idsBatch);
+                for (final IObject file : q.findAllByQuery(filesById, params)) {
+                    files.add((ome.model.core.OriginalFile) file);
+                }
+            }
+            return files;
 
     }
 
--- a/components/blitz/src/ome/services/blitz/repo/RepositoryDaoImpl.java
+++ b/components/blitz/src/ome/services/blitz/repo/RepositoryDaoImpl.java
@@ -104,6 +104,9 @@ public class RepositoryDaoImpl implements RepositoryDao {
     private static final String LOAD_USER_INSTITUTION =
             "SELECT institution FROM " + Experimenter.class.getName() + " WHERE id = :id";
 
+    /* for large database queries */
+    private static final int BATCH_SIZE = 1024;
+
     private final static Logger log = LoggerFactory.getLogger(RepositoryDaoImpl.class);
 
     private final IceMapper mapper = new IceMapper();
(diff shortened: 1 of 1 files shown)


joshmoore:
cc: @c0c0n3 -- to some extent it depends on what exception was being seen. If it's a PG integer overflow, then this will work. If it's an Ice MessageSizeMax, then we'll have to dig deeper (i.e. include paging at the method level).

c0c0n3:
Have tested your fix and as far as I can tell it works, well at least in my test scenario, see below.

### Test environment
Installed a full stack OMERO 5.3.2 server in a Linux container. (i.e. everything: Postgress, nginx, omero, etc.) Hardware: quad-core w/ 24G RAM.

### Test setup
I used the script below to create a directory with a fileset of 50,000 empty files.

    #!/usr/bin/env bash

    TARGET_DIR=$1
    FILES_N=$2

    # Generate a test fileset with empty files that can still be imported
    # into OMERO.
    # see:
    # - https://www.openmicroscopy.org/site/support/bio-formats5.4/formats/pattern-file.html
    #
    # Usage:
    # ./mk-fake-fileset.sh <target-dir> <how-many-files>
    #

    mkdir "$TARGET_DIR"
    cd "$TARGET_DIR"
    echo "test_T<1-$FILES_N>.fake" >> test.pattern
    for ((k = 1; k <= $FILES_N; ++k));
    do
        touch test_T"$k".fake
    done

After creating a regular OMERO user and a dataset for him, I imported the fileset into that dataset using: `omero import`. Note: the import took about 1hr on my box, so be prepared to wait if you want to rerun the tests yourself :-)

### Making OMERO break
At this point, I ran

    omero admin cleanse --dry-run -u root /OMERO

Note the script, after reconciling files, tries to delete empty directories but to determine if a directory can be considered empty, it calls `listFiles` which eventually results in a call to `getOriginalFiles` to list the files in the ginormous fileset imported earlier. The script fails because of a server-side exception in `getOriginalFiles` caused by us trying to stick way too many parameters (ids) in the "original file" query, in fact more than the PG driver can handle---see [this issue][pg-driver-issue]. I have attached a zip file with the script's output (`cleanse.before.out`) as well as the blitz log where you can actually see the exception in full---line 4966 "java.io.IOException: Tried to send an out-of-range integer as a 2-byte value: 50007".

### Testing your fix
I checked out openmicroscopy at tag `v5.3.2`, back-ported your fix, ran `python build.py`, and copied
`dist/lib/{server, client}/blitz.jar` to their respective directories in my server installation. Then I ran again

    omero admin cleanse --dry-run -u root /OMERO

which succeeded---output captured in `cleanse.after.out` in the attached zip file.





[pg-driver-issue]: https://github.com/pgjdbc/pgjdbc/issues/90
[cleanse-debug.zip](https://github.com/openmicroscopy/openmicroscopy/files/1139430/cleanse-debug.zip)

joshmoore:
As an aside: I wonder if `Parameters.addIds` should throw if `len(ids) > 32K`

c0c0n3:
@joshmoore throwing your own exception would make it easier for sys admins to diagnose the problem in the future, as they won't have to wade through the nasty Java exception chain. In fact, looking at the [bug report][bug-report] submitted by Olivier, I can't see no trace of the IO exception that actually caused the failure---and in fact if you look at the log I've attached it's not easy to tell at a glance what the problem is.

Regardless, the real issue here is that we should try avoiding prepared statements with a large number of parameters if at all possible. Batching avoids the exception, but doesn't help too much with performance. Ideally those parameters should end up in a db table and then the select should join from that table. Obviously as we all know, that's easier said than done... 

[bug-report]: http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-May/005992.html

joshmoore:
@mtbc was this implementation a result of `repo` not being mapped, i.e. is it now a non-issue?

mtbc:
> I wonder if `Parameters.addIds` should throw if `len(ids) > 32K`

What determines this? Could look at adding such a limit but would be good to either,

* Make sure it does ongoingly match the actual limit. Maybe make it configurable? For example, with a later version of the PostgreSQL JDBC driver, or a switch to the Oracle one, perhaps it is much different.

* Choose a common-denominator lower limit and make sure our own scripts batch client-side.

Alternatively, simply catching and reporting the error better may be a good compromise.

> a result of `repo` not being mapped

I don't believe so but that suggested https://trello.com/c/8UHZzADL/765-do-not-query-repo-via-sql to me, thank you.

c0c0n3:
Oh, I didn't realise you guys support multiple back-ends: Postgres, Oracle, etc. Then I'm not sure throwing our own exception is actually a good idea as it can get quite tricky to do it in a driver-independent way as @mtbc pointed out. The extra dev effort is not worth, IMHO, the little return in better error messages. I initially thought a one-liner would do, e.g. `if ids.size > 32K then throw`. So ignore my comment above, I reckon at this stage it's better *not* to throw the exception ourselves? Sorry for the confusion.

joshmoore:
@c0c0n3 I think a valid value could be found despite multiple backends (especially since non-Postgres is for all intents and purposes non-extant), and throwing earlier rather than later seems like a generally good idea (though not necessarily as part of this PR).

PR #5379, opened 2017-07-12
no longer use Django utils library "functional"

mtbc:
# What this PR does

Replaces Django compatibility library with standard one.

# Testing this PR

CI suffices.


joshmoore:
Wrong target branch?

mtbc:
Want to make sure it doesn't completely break Python testing or web client before I bring it into a merge build right before I ask you to merge a bunch of other PRs for m2. :smiley:

chris-allan (changes requested):
Just in case anyone else is using `omero_ext.functional`, even though I'm not aware of anyone who is, it would probably be a good idea to put a backwards compatibility package in place and then use the built in Python infrastructure to deprecate it.

Such a compatibility `omero_ex.functional` module could just effectively be `from functools import wraps` and a deprecation annotation I believe.

chris-allan (commented):
Just need our standard Python header and we should be good to go I think.

chris-allan (approved):

PR #5381, opened 2017-07-12
Generate Nginx includes for advanced configuration (e.g. SSL)

manics:
# What this PR does

This creates a minimal nginx configuration file that can be included in a fixed file created by a sysadmin. This allows custom nginx options to be defined once (e.g. SSL options), instead of having to modify the generated web config after every upgrade, and means any Nginx server options can be used.

# Testing this PR

1. Install a production OMERO.web in the standard way
2. Install Nginx.
3. Create a omero-web nginx wrapper file with an `include` statement for the generated omero-web config. This file is exclusively managed by the sysadmin e.g. `/etc/nginx/conf.d/omero-web-wrapper.conf`:
```
server {
    listen 80;
    server_name $hostname;

    # SSL configuration ...

    sendfile on;
    client_max_body_size 0;

    # Include generated file from omero web config nginx-location:
    include /opt/omero/web/omero-web-location.include;
}
```
4. Generate the minimal omero-web configuration in the location specified by your `include` statement. The expectation is that this would be routinely regenerated on every OMERO.web upgrade. e.g.
```
omero web config nginx-location > /opt/omero/web/omero-web-location.include
```
5. Start nginx
6. Start OMERO.web: `omero web start`
7. OMERO.web should work as normal

# Related reading
- https://trello.com/c/igcRoPcu/11-configure-ssl-via-bin-omero-web-nginx

# Notes
- `omero web config nginx-location` generates the `location` blocks only, other server options are now left for the sysadmin to manage e.g. `sendfile on;`, `client_max_body_size 0;`.
- I've removed the separate `upstream` block since there was only one server which means it's redundant, and it would be impossible to auto-generate a load-balanced upstream configuration since you'd need to know the addresses of all OMERO.web backends.

### DIFF
--- a/etc/templates/web/nginx-location.conf.template
+++ b/etc/templates/web/nginx-location.conf.template
@@ -19,7 +19,7 @@ location @proxy_to_app%(PREFIX_NAME)s {
     proxy_redirect off;
     proxy_buffering off;
 
-    proxy_pass http://%(FASTCGI_EXTERNAL)s fail_timeout=0;
+    proxy_pass http://%(FASTCGI_EXTERNAL)s;
 }
 
 location %(FORCE_SCRIPT_NAME)s {
--- a/etc/templates/web/nginx-location.conf.template
+++ b/etc/templates/web/nginx-location.conf.template
@@ -1,5 +1,7 @@
+# You are strongly recommended to include the following in your enclosing
+# server configuration:
 #sendfile on;
-#client_max_body_size ;
+#client_max_body_size 0;
 
 # maintenance page serve from here
 location @maintenance%(PREFIX_NAME)s {
--- a/etc/templates/web/nginx-location.conf.template
+++ b/etc/templates/web/nginx-location.conf.template
@@ -1,7 +1,18 @@
-# You are strongly recommended to include the following in your enclosing
-# server configuration:
-#sendfile on;
-#client_max_body_size 0;
+# These location blocks should be included in your server configuration
+# For example:
+#
+##server {
+##    listen 80;
+##    server_name $hostname;
+##
+##    # SSL configuration ...
+##
+##    sendfile on;
+##    client_max_body_size 0;
+##
+##    # Include generated file from omero web config nginx-location:
+##    include /opt/omero/web/omero-web-location.include;
+##}
 
 # maintenance page serve from here
 location @maintenance%(PREFIX_NAME)s {
--- a/etc/templates/web/nginx-location.conf.template
+++ b/etc/templates/web/nginx-location.conf.template
@@ -1,0 +1,30 @@
+#sendfile on;
+#client_max_body_size ;
+
+# maintenance page serve from here
+location @maintenance%(PREFIX_NAME)s {
+    root %(ROOT)s/etc/templates/error;
+    try_files $uri /maintainance.html =502;
+}
+
+# weblitz django apps serve media from here
+location %(STATIC_URL)s {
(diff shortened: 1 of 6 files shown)


bramalingam:
Followed the steps as suggested, OMERO.web works as expected. Looks good to merge (to me)

etc/templates/web/nginx-location.conf.template
```diff
@@ -0,0 +1,30 @@
+#sendfile on;
```
joshmoore: Perhaps either remove these or extend this comment block to say that such properties should be set by the sysadmin. Perhaps include a full working example?

manics: I'll add a comment. Can you explain what you mean by a full working example?

joshmoore:
```
server {
    listen 80;
    server_name $hostname;

    # SSL configuration ...

    sendfile on;
    client_max_body_size 0;

    # Include generated file from omero web config nginx-location:
    include /opt/omero/web/omero-web-location.include;
}
```

from the description of this PR.

manics: OK, I get you. Just worried that it's another piece of commented code that will never be tested. It's essentially `omero web config nginx` with the `location` blocks replaced by a single `include /opt/omero/web/omero-web-location.include;` so if we do modify the existing `nginx` templates we should also modify this comment.

manics:
I've updated the comment, and added a test to ensure the comment in `nginx-location` matches the generated `nginx` config

manics:
--rebased-to #5387

kennethgillen:
Can confirm webclient is working with these changes - big 👍 💯  - I'd need guidance on how to run the tests, or let someone else try the test.

joshmoore:
Using omero-web-docker:

```
$ perl -i -pe 's/ARG OMERO_VERSION=latest/ARG OMERO_VERSION=OMERO-DEV-merge-build/' Dockerfile
$ docker build -t josh-web .
$ docker run --rm -ti --entrypoint /opt/omero/web/venv/bin/python josh-web /opt/omero/web/OMERO.web/bin/omero web config nginx-location > omero.conf
$ docker run -d --name josh-web -p 4080:4080 josh-web
$ docker run -d --name josh-nginx --link josh-web:web -p 9090:80 -v `pwd`/nginx.conf:/etc/nginx/nginx.conf:ro -v `pwd`/omero.conf:/omero.conf:ro nginx
```

with nginx.conf:

```
user  nginx;
worker_processes  1;

error_log  /var/log/nginx/error.log warn;
pid        /var/run/nginx.pid;


events {
    worker_connections  1024;
}


http {
    include       /etc/nginx/mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    access_log  /var/log/nginx/access.log  main;

    sendfile        on;
    #tcp_nopush     on;

    keepalive_timeout  65;

    #gzip  on;

    include /etc/nginx/conf.d/*.conf;

    server {
      listen 80;
      server_name $hostname;

      sendfile on;
      client_max_body_size 0;

      # Include generated file from omero web config nginx-location:
      include /omero.conf;
    }
}
```

for a passing docker test (barring statics which is under discussion) :+1:

Merging and updating the card for the necessary docs.

PR #5382, opened 2017-07-13
Webui field display

dominikl:
# What this PR does

Fixes some issues with the field display

- Lower left box of field positions: Add label 'Field positions in well'
- Lower centre panel showing the images from each field: Add label 'Images from each field'
- "The plate view in the centre panel for some screens looks quite messy": Set the background of the cell displaying the well to white to match the placeholder (empty) well color.

# Testing this PR

Check the points mentioned above.

# Related reading

https://trello.com/c/MvTUcbkj/18-webui-display-of-fields

Note: I can't replicate first on the card, would need an example plate showing that behaviour.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
@@ -2183,10 +2183,11 @@ div.paging input.button_pagination {
     position: relative;
 }
 #spw table td.well {
-    background: #eee;
+    background: #fff;
 }
 #spw table td.placeholder {
     vertical-align: top;
+    background: #fff;
 }
 
 #spw table td.ui-selecting { border: 1px dashed #555; padding: 1px; background-color:#87ABD2;}
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/data/containers.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/data/containers.html
@@ -848,6 +848,7 @@
 
     <!-- Panel hidden unless needed for showing spatial birds eye view of Wells - see ome.spwgridview.js -->
     <div id="left_panel_bottom" class="left_panel_inner" style="top: auto; height: 0; border-top: solid 1px hsl(210,10%,85%);">
+        <label style="margin-top: 4px; margin-left: 4px;">Field positions in well</label>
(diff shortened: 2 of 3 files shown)


pwalczysko:
``Images from each field`` -> hmm, doesn't it somehow imply there could be more images per field ? But in fact, afaik, there is typically just one image per field. Maybe ``fields`` or ``fields from each well`` would be better ? cc @eleanorwilliams

pwalczysko:
If possible (but only if possible) , could the label be more central on the box and a bit more away from the top rim ?
screen shot 2017-07-18 at 14 37 57

eleanorwilliams:
@pwalczysko @dominikl Yes, I think Petr is right, and maybe "fields from well" would be better than "Images from each field" .  I am thinking "fields from well" rather than "fields from each well"  because a single well will be clicked on to show this view.

pwalczysko:
Re: ``"The plate view in the centre panel for some screens looks quite messy": Set the background of the cell displaying the well to white to match the placeholder (empty) well color.``

Discussing with @eleanorwilliams, she likes the indicated change by @dominikl here.
But we have one problem (to fix by @dominikl ) and one question (probably for @will-moore or @jburel ).

The problem to fix in this PR is:

  - the black grid stripes around the empty wells disappeared due to this fix, return them back please
    
The question (probably nothing to do with this PR) for @will-moore and @jburel is : 

  - Why were there all wells without the "lip" (a bit of background stripe under the image thumb itself) for each and every well in any screen in 5.2 whereas now (both 5.3 nightshade and 5.3 idr as well as 5.4 eel merge and latest) we have a mixture of wells with "lip" and without it ? 

For example of well with a "lip" see http://web-dev-merge.openmicroscopy.org/webclient/?show=well-334506  login as user-4
Well without a "lip" see http://web-dev-merge.openmicroscopy.org/webclient/?show=well-338079 login as user-4

pwalczysko:
In summary, two RFEs here https://github.com/openmicroscopy/openmicroscopy/pull/5382#issuecomment-316066375 and  https://github.com/openmicroscopy/openmicroscopy/pull/5382#issuecomment-316081200, one bug https://github.com/openmicroscopy/openmicroscopy/pull/5382#issuecomment-316092312 and one question https://github.com/openmicroscopy/openmicroscopy/pull/5382#issuecomment-316092312 .

jburel:
Closing in favour of gh-5403

PR #5386, opened 2017-07-20
Increase OMEROweb log-size to 500MB

manics:
# What this PR does

OMERO.web logs are currently rotated at 5MB, in contrast to the Java servers such as Blitz which are rotated at 500MB. This PR sets it to 500MB, matching the `LOGSIZE` set for the Python servers https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.0-m2/components/tools/OmeroPy/src/omero/util/__init__.py#L33

# Testing this PR

Install OMERO.web. Interact with the webclient until `OMEROweb.log` is larger than 5MB. Setting `omero.web.debug=false` will help fill the logs up.

This PR is required for debugging issues on the IDR regarding large numbers of queries/sessions.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -85,6 +85,8 @@ FULL_REQUEST_LOGFORMAT = (
     ' HTTP %(status_code)d %(request)s')
 
 LOGGING_CLASS = 'omero_ext.cloghandler.ConcurrentRotatingFileHandler'
+LOGSIZE = 500000000
+
 
 LOGGING = {
     'version': 1,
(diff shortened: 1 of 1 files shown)


joshmoore:
LGTM. I do wonder if we don't want to import the value from OmeroPy rather than redefining the value. My guess is also that this is something we should target for the mainline.

joshmoore:
Merging for 0.4.3. When we go through PRs to rebase on 5.3, we'll need to come back to the above discussion.

kennethgillen:
🙏 mainline rebase?

joshmoore:
Give it a try, @kennethgillen ? `scc rebase 5386 develop`

kennethgillen:
--rebased-to #5681

kennethgillen:
Wow! Thanks @joshmoore

joshmoore:
Kudos!

PR #5388, opened 2017-07-20
Generate Nginx includes for advanced configuration (e.g. SSL) (rebased onto metadata53) (rebased onto develop)

manics:
This is the same as gh-5387 but rebased onto develop.

----

This is the same as gh-5381 but rebased onto metadata53.

----

# What this PR does

This creates a minimal nginx configuration file that can be included in a fixed file created by a sysadmin. This allows custom nginx options to be defined once (e.g. SSL options), instead of having to modify the generated web config after every upgrade, and means any Nginx server options can be used.

# Testing this PR

1. Install a production OMERO.web in the standard way
2. Install Nginx.
3. Create a omero-web nginx wrapper file with an `include` statement for the generated omero-web config. This file is exclusively managed by the sysadmin e.g. `/etc/nginx/conf.d/omero-web-wrapper.conf`:
```
server {
    listen 80;
    server_name $hostname;

    # SSL configuration ...

    sendfile on;
    client_max_body_size 0;

    # Include generated file from omero web config nginx-location:
    include /opt/omero/web/omero-web-location.include;
}
```
4. Generate the minimal omero-web configuration in the location specified by your `include` statement. The expectation is that this would be routinely regenerated on every OMERO.web upgrade. e.g.
```
omero web config nginx-location > /opt/omero/web/omero-web-location.include
```
5. Start nginx
6. Start OMERO.web: `omero web start`
7. OMERO.web should work as normal

# Related reading
- https://trello.com/c/igcRoPcu/11-configure-ssl-via-bin-omero-web-nginx

# Notes
- `omero web config nginx-location` generates the `location` blocks only, other server options are now left for the sysadmin to manage e.g. `sendfile on;`, `client_max_body_size 0;`.
- I've removed the separate `upstream` block since there was only one server which means it's redundant, and it would be impossible to auto-generate a load-balanced upstream configuration since you'd need to know the addresses of all OMERO.web backends.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/web.py
+++ b/components/tools/OmeroPy/src/omero/plugins/web.py
@@ -97,8 +97,10 @@ def assert_config_argtype(func):
                 mismatch = True
             if settings.APPLICATION_SERVER in (settings.WSGITCP,):
                 if argtype not in (
-                        "nginx", "nginx-development",
-                        "nginx-location", "nginx-upstream"):
+                    "nginx",
+                    "nginx-development",
+                    "nginx-location",
+                ):
                     mismatch = True
             if (settings.APPLICATION_SERVER in (settings.WSGI,) and
                     argtype not in ("apache22", "apache24", "apache")):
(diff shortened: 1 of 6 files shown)


manics:
--rebased-from #5387

manics:
Somehow this ended up as a duplicate rebase

PR #5389, opened 2017-07-21
adjust OMERO_STARTUP_WAIT delay from 10000 to 80000

mtbc:
# What this PR does

Sets `OMERO_STARTUP_WAIT` to 80 seconds.

# Testing this PR

Tables tests may again pass.

# Related reading

https://trello.com/c/wPBnTfPU/58-flaky-tests

### DIFF
--- a/components/blitz/src/ome/services/blitz/Entry.java
+++ b/components/blitz/src/ome/services/blitz/Entry.java
@@ -52,7 +52,7 @@ public class Entry {
     volatile Ice.Communicator ic = null;
 
     private static void waitOnStartup() {
-        int ms = 10000; // 10 seconds by default
+        int ms = 80000; // 80 seconds by default
         try {
             String prop = System.getenv("OMERO_STARTUP_WAIT");
             ms = Integer.valueOf(prop);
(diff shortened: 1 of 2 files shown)


manics:
If this works are you proposing it as a production change too? Or leaving it at `10000` and setting `OMERO_STARTUP_WAIT`?

mtbc:
No, not proposing as a production change, more as a hope to get tables tests passing in merge CI until someone like @joshmoore can render a more authoritative opinion on the issue. The increase from 10s to 80s seems alarmingly large to me.

mtbc:
Well, tables seem good today.

joshmoore:
Note that the wait time is absolute. _Nothing_ is done until the timer finishes and in fact, shutdown is blocked as well. Unless we can make have this periodically check, then I wouldn't suggest upping it across the board substantially.

mtbc:
Let's see if we can live without this now OMERO-DEV-merge-integration is mysteriously faster again.

mtbc:
Looks like we've had failures all this week anyway.

PR #5390, opened 2017-07-21
Group owner test

pwalczysko:
# What this PR does
Adds a test for group owner who is at the same time a light admin with no privileges or with DeleteOwned privileges.
The light admin is a mamber and owner of one group, and should be able to delete data there whether he/she has the DeleteOwned privilege or not. In another group, where the light admin/owner is not a member and not an owner, he/she shoudl be able to delete data only when he/she has the DeleteOwned privilege.

# Testing this PR

CI suffices

@mtbc


mtbc:
Reran Travis which turned green. Relisting for tomorrow to check Jenkins.

mtbc (approved):

PR #5391, opened 2017-07-21
adjust some CLI admin checking to respect restrictions

mtbc:
# What this PR does

OMERO.cli has various subcommands that check if the user is an administrator. This PR modifies some of those to check specifically that the user has certain light admin privileges. It does not add checks to subcommands that did not already have administrator checks.

# Testing this PR

Subcommands that used to report "SecurityViolation: Admins only!" will now, in cases where a certain light admin privilege is always required for that subcommand and the light admin does not have it, report a different "SecurityViolation" message that lists the missing privilege(s). Affected subcommands are:

* `bin/omero fs mkdir` which now requires `WriteManagedRepo` and `Chown`
* `bin/omero fs rename` which now requires `WriteOwned`, `WriteManagedRepo`, `DeleteOwned`, `DeleteManagedRepo`
* `bin/omero group add` which now requires `ModifyGroup`
* `bin/omero user add` which now requires `ModifyUser` and `ModifyGroupMembership`
* `bin/omero ldap create` which now requires `ModifyUser` and `ModifyGroupMembership`
* `bin/omero ldap setdn` which now requires `ModifyUser` or `ModifyGroup` depending on the arguments.

Web admin allows adjusting privileges but for fine-grained tests one may need to consult our sysadmin CLI docs for 5.4. Further, note that the server may take up to a minute to apply privilege changes to existing sessions.

# Related reading

https://trello.com/c/VyOwZwyr/40-admin-only-chgrp-chown-in-clients

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -34,6 +34,8 @@ from omero.cli import NonZeroReturnCode
 from omero.cli import VERSION
 from omero.cli import UserGroupControl
 
+from omero.model.enums import AdminPrivilegeReadSession
+
 from omero.plugins.prefs import \
     WriteableConfigControl, with_config, with_rw_config
 from omero.install.windows_warning import windows_warning, WINDOWS_WARNING
@@ -1789,7 +1791,7 @@ OMERO Diagnostics %s
             " regenerated. Use the omero.ports.xxx configuration properties"
             " instead.")
 
-    @admin_only()
+    @admin_only(AdminPrivilegeReadSession)
     def cleanse(self, args):
         self.check_access()
         from omero.util.cleanse import cleanse
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -957,7 +957,7 @@ present, the user will enter a console""")
         else:
             self.ctx.call(command)
 
-    @admin_only
+    @admin_only()
     @with_config
     def fixpyramids(self, args, config):
         self.check_access()
@@ -1789,7 +1789,7 @@ OMERO Diagnostics %s
             " regenerated. Use the omero.ports.xxx configuration properties"
             " instead.")
 
-    @admin_only
+    @admin_only()
     def cleanse(self, args):
         self.check_access()
         from omero.util.cleanse import cleanse
--- a/components/tools/OmeroPy/src/omero/plugins/ldap.py
+++ b/components/tools/OmeroPy/src/omero/plugins/ldap.py
@@ -92,7 +92,7 @@ to users.""")
         for x in (active, list, getdn, setdn, discover, create):
             x.add_login_arguments()
 
-    @admin_only
+    @admin_only()
     def active(self, args):
         c = self.ctx.conn(args)
         ildap = c.sf.getLdapService()
@@ -102,7 +102,7 @@ to users.""")
         else:
             self.ctx.die(1, "No")
 
-    @admin_only
+    @admin_only()
     def list(self, args):
         c = self.ctx.conn(args)
         iadmin = c.sf.getAdminService()
@@ -129,7 +129,7 @@ to users.""")
                 count += 1
         self.ctx.out(str(tb.build()))
 
-    @admin_only
+    @admin_only()
     def getdn(self, args):
         c = self.ctx.conn(args)
         iadmin = c.sf.getAdminService()
@@ -163,7 +163,7 @@ to users.""")
         if dn is not None and dn.strip():
             self.ctx.out(dn)
 
-    @admin_only
+    @admin_only()
     def setdn(self, args):
         c = self.ctx.conn(args)
         iupdate = c.sf.getUpdateService()
@@ -188,7 +188,7 @@ to users.""")
                               in ("yes", "true", "t", "1")))
             iupdate.saveObject(obj)
 
-    @admin_only
+    @admin_only()
     def discover(self, args):
         c = self.ctx.conn(args)
         ildap = c.sf.getLdapService()
@@ -223,7 +223,7 @@ to users.""")
                                      % (e.getId().getValue(),
                                         e.getOmeName().getValue()))
 
-    @admin_only
+    @admin_only()
     def create(self, args):
         c = self.ctx.conn(args)
         ildap = c.sf.getLdapService()
--- a/components/tools/OmeroPy/src/omero/plugins/ldap.py
+++ b/components/tools/OmeroPy/src/omero/plugins/ldap.py
@@ -11,6 +11,9 @@
 import sys
 
 from omero.cli import CLI, ExceptionHandler, admin_only, UserGroupControl
+from omero.model.enums import (
+    AdminPrivilegeModifyGroup, AdminPrivilegeModifyGroupMembership,
+    AdminPrivilegeModifyUser)
 from omero.rtypes import rbool
 
 HELP = """Administrative support for managing users' LDAP settings
@@ -163,6 +166,14 @@ to users.""")
         if dn is not None and dn.strip():
             self.ctx.out(dn)
 
+    @admin_only(AdminPrivilegeModifyGroup)
+    def update_group(self, iupdate, group):
+        iupdate.saveObject(group)
+
+    @admin_only(AdminPrivilegeModifyUser)
+    def update_user(self, iupdate, user):
+        iupdate.saveObject(user)
+
     @admin_only()
     def setdn(self, args):
         c = self.ctx.conn(args)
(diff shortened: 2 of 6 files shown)


pwalczysko:
Why are the ``chgrp`` and ``chown`` commands not on your list of affected commands (i.e. why not to implement the same logic to them) ? Maybe I am slightly missing the point here.

pwalczysko:
Will test tomorrow.

mtbc:
> It does not add checks to subcommands that did not already have administrator checks.

* The `chgrp` plugin was handled in #5376. It is not easily converted to use decorators because its admin check is not a simple gateway into a method so the current code is probably optimally clear from that point of view.
* The `chown` plugin has no administrator checks: it never used to report "SecurityViolation: Admins only!" because it leaves the whole non-administrator, "an owner of the group that the data is in *if* the target user is a member of the group" check to the server. This PR is basically just about fixing the existing client-side admin checks to accommodate the new restrictions.

Arguably part of reviewing this PR might be checking that my omissions are correct. Fortunately, for `chown` the check is easy: something so broad as `grep -i admin components/tools/OmeroPy/src/omero/plugins/chown.py` finds only comment/help lines. The current server code does at least give useful error messages ("user X is not an owner of group Y", etc.) so I figure that improving client-side checking here is outwith the restricted admins work and can wait until we have more developer time available for OMERO.cli.

pwalczysko:
Positive (doer has right permissions) and negative (doer has not right permissions) tests of ``bin/omero fs mkdir huhly --parents`` passed. The message of the negative test output was as expected.

Will continue with the other commands tomorrow.

pwalczysko:
``bin/omero fs rename`` went fine with full admin and simple user (got warning ``Admins only``) and with light admin I got ``SecurityViolation: Admin restrictions: DeleteManagedRepo, DeleteOwned, WriteManagedRepo, WriteOwned``  which is as expected.

``bin/omero group add`` passes for full and light admin as expected, with appropriate warnings each.
``bin/omero user add `` passes for full and light admin as expected, with appropriate warnings each.
``bin/omero ldap create`` passes for full and light admin as expected, with appropriate warnings each.

pwalczysko:
``bin/omero ldap setdn --user-name user-1 true`` will deliver the expected message for light admin and full admin. The same when replacing ``true`` with ``false``. When replacing the ``--user-name`` with ``--group-name``, the message will complain about ``ModifyGroup`` for light admin.
All as expected here too.

pwalczysko:
For ``bin/omero group add`` and ``bin/omero user add`` checked also that a light admin with the appropriate premissions is able to execute those actions. The messages about the missing permissions were as appropriate even when the permissions were added incrementally to the light admin.

pwalczysko:
No problems detected. All tests are passing for me.

The only thing is that I cannot check the functionality of the ``ldap setdn`` and ```ldap create``` commands, as not really sure what these are supposed to do, and mainly, the setup of the LDAP seems locally impossible for me.
I do not have any alarming indications, and from the perspective of the bits changed in this PR these commands certainly work too. Maybe somebody who knows more about LDAP could have a look before merge here if something did not escape our attention ?

mtbc:
Kenny's LDAP settings in https://trello.com/c/tVjXXmoP/59-resurrect-dev-ldap-service#comment-5979b6cca29e5d94e30156bf are working for me.

pwalczysko:
Also set up the LDAP as per https://trello.com/c/tVjXXmoP/59-resurrect-dev-ldap-service#comment-5979b6cca29e5d94e30156bf and the commands
``bin/omero ldap create <username>`` and
``bin/omero ldap setdn <username> false`` were working as expected.

Ready to merge FMPOV

components/tools/OmeroPy/src/omero/plugins/ldap.py
```diff
@@ -163,7 +166,15 @@ def getdn(self, args):
...
-    @admin_only
+    @admin_only(AdminPrivilegeModifyGroup)
+    def update_group(self, iupdate, group):
+        iupdate.saveObject(group)
+
+    @admin_only(AdminPrivilegeModifyUser)
+    def update_user(self, iupdate, user):
+        iupdate.saveObject(user)
+
+    @admin_only()
```
joshmoore, line 177: Should this be `AdminPrivilegeModifyUser, AdminPrivilegeModifyGroup`?

joshmoore: I found the trello conversation about this. If you move the `_check_admin` function out of the `admin_only` function definition then you can remove the `admin_only` decorator here in favor of an in-line invocation of `_check_admin` with the appropriate parameters to match the user's arguments.

mtbc: I thought we were keen to use the decorator though? Otherwise (renamed) `_check_admin` becomes more API we commit to offering to consumers. Can make this change though if desired. I also wonder if there's an easy way to use `admin_only` inline, maybe not.

joshmoore: Few things:

 * I don't see a particular concern about adding API for `check_admin` as long as it's documented, re-usable, etc
 * Not sure that we're necessarily _keen_ to use the decorator. Probably the biggest downside of _not_ having it, is that it's easier to miss the what's going on. But unless we want to add OR-semantics to `admin_only`, this particular method seems to not match what the decorator provides.
 * All that being said: we may not need to change _anything_ since the internal use of `update_group` and `update_user` will _also_ be wrapped by the decorator usages on those methods.

mtbc: :+1: Maybe just depends if we want to handle these conditional checks via decorated internal methods or by the API also offering `check_admin`.

joshmoore: In chatting with @mtbc, it turns out I had this backwards: I assume `@admin_only` would require a _full_ admin, but it doesn't. It simply defines "in system group" and therefore even a light admin with no privileges will be able to attempt the given call. Since `update_group` or `upgrade_user` then impose stricter requirements, the user will still get a sensible message.

joshmoore:
Building on https://github.com/openmicroscopy/openmicroscopy/pull/5391#discussion_r133151709, a result of having `@admin_only()` accept any light admin implies that if we ever add a privilege for "read-others data" then `bin/omero admin cleanse` will once against have the potential of deleting everyone's data since low ranking admin may not be allowed to read everything.

components/tools/OmeroPy/src/omero/plugins/ldap.py
mtbc: This PR should improve things from where they were without this PR.

Still, outwith this PR but in the context of that distinction/confusion it should thus be noted that I may have missed that some of the remaining `@admin_only()` may still benefit from specific privileges being required. More severely, one may worry about that a light admin running a CLI subcommand may cause,

1. First step of subcommand happens.
1. Second step of subcommand fails (needs different privilege).
1. No useful rollback of the first, as previously it seemed implausible that the first would succeed yet not the second.

I didn't find such in my CLI review but the possibility ought still be considered.

joshmoore: Loading of `getConfigValue`s occurs to me as another possible permission.

components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -1789,7 +1791,7 @@ def ports(self, args, config):
             " regenerated. Use the omero.ports.xxx configuration properties"
             " instead.")
 
-    @admin_only()
+    @admin_only(AdminPrivilegeReadSession)
```
joshmoore: Possession of `ReadSession` counts as the highest level of light admin'ing?

mtbc: No light admin should have `ReadSession` and webadmin provides no way to give it to a light admin without simply making them a full admin. See warning boxes on https://docs.openmicroscopy.org/omero/5.4.0-m3/developers/Server/LightAdmins.html and https://docs.openmicroscopy.org/omero/5.4.0-m3/sysadmins/cli/LightAdmins.html.

joshmoore: I appreciate that this represents the current state of 5.4.x but I'm still worried that A) this is confusing to users and developers (or at least to me) and B) that this isn't future-proof unless we overload `ReadSession` to be equivalent with `FullAdmin`. I've added a card for us to further discuss the definition of `FullAdmin`: https://trello.com/c/4dvUYSjS/126-definition-of-ultimate-admin

joshmoore:
Merging as an improvement to the current CLI. (Thanks for bearing with me)

components/tools/OmeroPy/src/omero/plugins/admin.py
joshmoore: see https://github.com/will-moore/openmicroscopy/pull/58

PR #5394, opened 2017-07-24
light admin creates light admin

will-moore:
# What this PR does

This removes the client-side restriction on Light Admins creating other (Light) Admins.
However, we now allow Light Admin to choose role of ```Full Admin``` or *any* privileges for the new / edited user even though the server will prevent any privilege escalation.

So, we need to add back restrictions based on the Light Admin's own privileges.

cc @pwalczysko @mtbc 

# Testing this PR

 - Needs full testing of:
    - Create new user/ Admin / Light Admin
    - Edit yourself / user / Admin / Light Admin, 
 - By Admin / Light Admin with various privileges

TODO:
 - [x] Add test for new Blitz Gateway method
 - [x] Add web integration tests (NB: only test for experimenter form is currently in test_csrf.py)


mtbc:
A light admin can't create a full system user *then* add restrictions as the first step is blocked. The admin service now offers `createLightSystemUser` as a convenience method which should work for you.

pwalczysko:
Tests in https://github.com/openmicroscopy/openmicroscopy/pull/5397 show that when the method mentioned in [Mark's comment](https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-317779369) is used, no unexpected behaviour of light-creates-light workflow occurs in the four tested cases.

will-moore:
As discussed with @mtbc, we really need a ```createLightSystemUser``` to have a ```withPassword``` option since I was previously using ```admin.createExperimenterWithPassword()```.

mtbc:
Absolutely: https://trello.com/c/KPI5h2y3/90-admin-service-has-light-admins#comment-59789198f62fdd7fde522c51.

will-moore:
Now, when a Light-Admin (or Admin) is creating/editing another Experimenter OR themselves, we hide any privileges that they don't have. This prevents the Light-Admin from trying to add these (privilege escalation) but also means they can't remove privileges they don't have. NB: I hide the checkboxes instead of disable them because if they are disabled then they are not included in the submitted form data (as if they are un-checked) and will all be removed from the saved user's privileges.

Also, I haven't disabled the 'Administrator' or 'User' Role options when a Light-Admin or Admin is editing themselves (not sure if this is possible in Django's form - might have to use JavaScript instead if needed).
Question: Can an Admin / Light-Admin remove all of their own Privileges? The server prevents them from removing themselves from the System group. Should they also be prevented from losing all their privileges?

mtbc:
> Can an Admin / Light-Admin remove all of their own Privileges?

Probably worth testing to be sure but I'd expect so. I've tried to make everything exactly what it says on the tin, with the exceptions that `root` is always a full administrator and, later, that privilege escalation must be prevented. Though, for this particular matter part of me wonders: if they are so careless as to remove their own privileges accidentally, probably best that they no longer have them!

pwalczysko:
Will do tests according to https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-307067330 and https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-307354854 and https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-306791636 and https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-307072678

Although the expected behaviour in this PR might be slightly different, these are the workflows to watch.

pwalczysko:
Bug1 (blocking): LIght admin with only ``Create and edit users`` cannot deactivate another user.
   - create light admin as above (user-6, ie. another admin creates)
   - login as that light admin and edit a user
   - the user is originally active, uncheck the ``Active`` checkbox and Save
   - observe normal behaviour of the UI -> all seems successful
   - again edit the user you have apparently deactivated
   - observe the checkbox is checked, and the user appears active

mtbc:
I suppose (de)activation requires `ModifyGroupMembership`.

pwalczysko:
>I suppose (de)activation requires ModifyGroupMembership.

Do not dispute you might be right. But definitely not something explainable or desirable. Not intuitive at the least.

pwalczysko:
Bug2:  (blocking): LIght admin with only Create and edit users can inadvertently create another light admin without even wanting to. Also any other light admin can only create light admins, the creation of full admins is not prevented, but results in other light admins being created. The same goes if light admin attempts to create a normal user -> light admin is created instead with no warning.
   - create light admin as above (user-6, ie. another admin creates)
   - login as that light admin and create a new (normal, non-admin) user
   - add the just created user to the group in the creation form (this is compulsory for user creation and nothing is stopping you in the UI to do that)
   - click Save
   - light admin must have the impression he/she succesfully created a normal user
   - in fact, the new user is a light admin, is a member of System group -> not expected

pwalczysko:
[Bug1](https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-318642739) and [Bug2](https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-318648888) actually strongly suggests that ``Create and Edit Users`` being given in isolation makes not much sense - in the first place, you cannot create a normal user, because you cannot add them to the group. Maybe @mtbc 's https://trello.com/c/UHagkHfL/94-guide-toward-sane-restriction-combinations gives the lead - there is some guidance in the doc, yes, but I think maybe even some UI modification might be needed ?

pwalczysko:
Light admin with ``Create and Edit users`` and ``Edit group Membership`` can deactivate another user as expected. Similarly as described in https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-318648888, light admin cannot create any other user than a light admin (they cannot create normal users, which is a blocker, and when trying to create a full admin, they inadvertently create a new light admin, which is not desirable either.

pwalczysko:
Checked also behaviours for light admin - all as expected. For other types of light admins, all expected except the blocker https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-318648888. Will retest after this behaviour is changed.

pwalczysko:
As discussed, we will deal with [Bug1](https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-318642739) on the trello card https://trello.com/c/UHagkHfL/94-guide-toward-sane-restriction-combinations. This is above the scope of this PR.

will-moore:
@pwalczysko Fixed the ```bug2``` above.
NB: As discussed with @mtbc, I found that light-admins with ONLY ```CreateUser``` can create users and add them to groups in one shot via ```admin.createExperimenterWithPassword(exp, password, defaultGroup, listOfGroups)```, so probably we should also support this for creating light-admins?

mtbc:
Definitely worth a card, probably not a big job. Happy for others to decide when it should be done.

pwalczysko:
@will-moore 
>so probably we should also support this for creating light-admins?

Clear yes from me.

will-moore:
Fixed disabling on 'Administrator' option if Light Admin is editing or creating another user (so that Light Admin can't try to create a Full Admin).
Also, if you are editing *yourself*, or if you are editing the *root* user then all the Role options will be disabled (but need to test that this doesn't affect the role or privileges).

pwalczysko:
Suspected bug: Although the behaviour itself is not undesirable, the following should be treated as inexpected behaviour. 
   - light admin edits him/herself
   - light admin unchecks some boxes (removes some of hteir privileges)
   - light admin clicks Save
   - light admin goes and edits themselves again
   - the unchecked boxes are checked again, light admin was not able to remove privileges from themselves -> not expected.

pwalczysko:
Bug3 (blocking) : a variation of the [suspected bug](https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-319389212), this time with editing of others - the editing is allowed in expected manner, but after ``Save`` is clicked, the changes do not take place.

  - light admin with "Crete and Edit Users" privillege only edits another user (non-admin)
  - light admin tries to make a light admin out of this user with the same privileges as themselves
  - the action is proceeding and the UI complies -> expected
  - after "Save" is clicked, and the edited user is navigated to again, observation is that the edition was not successful -> not expected

pwalczysko:
Bug4 (blocking): 

  - light admin with "Crete and Edit Users" privillege only edits another user (non-admin)
  - light admin changes the surname of the user
  - an error appears after "Save" is clicked

pwalczysko:
Expected tested behaviour:

 - light admin with "Crete and Edit Users" privillege only can create another user and add them to one group (during creation step)
 - light admin with "Crete and Edit Users" privillege only can create another user and add them to multiple groups (3 groups tested, during creation step)
 - light admin with "Crete and Edit Users" privillege only can create another light admin and add them to system group (during creation step)
 - light admin with "Crete and Edit Users" privillege only can edit and successfully change others' passwords
 - light admin with "Crete and Edit Users" privillege only cannot edit group info and cannot create new groups

pwalczysko:
Not expected behaviour:

  - light admin with "Crete and Edit Users" privillege only cannot disable others (possibly not for this PR, and maybe the https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-319394272 is in play here, but it would be nice to have)
  - light admin with "Crete and Edit Users" privillege only cannot create a light admin and add them to multiple groups at creation step, probably connected to https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-319202895

pwalczysko:
Bug5: light admin elicits a crash when trying to downgrade another light admin to be a user.

   - create a ight admin with "Crete and Edit Users" privillege only
   - logged in as this light admin, edit another light admin (also one with "Create and Edit Users" privilege only)
   - click the radio button "User" (the membership of System group goes automatically away, as expected)
   - add the new user to any group
   - click Save and observe crash

Edit: The bug can be connected to the addition to group step. Although it is possible to this light admin to create a new user and add them to any number of groups at creation, when downgrading a light admin to a user and adding them to groups, this does not seem to follow the same logic. Nevertheless, when a light admin with "Create...users" and "Add users to groups" performs the workflow, it succeeds.

Edt2: The light admin with "Create ...Users" privilege cannot upgrade a user to light admin either. This seems again to be connected to their inability to add users to groups, because a light admin with "Create...users" and "add users..." privileges can upgrade a user to light admin succesfully.

pwalczysko:
Other tested features:

 - light admin with "Crete and Edit Users" privillege only can edit and successfully change others' passwords
 - light admin with "Crete and Edit Users" privillege only can edit and successfully change others' lastname and save
 - light admin with "Crete and Edit Users" privillege only cannot edit group info and cannot create new groups
 - light admin with "Crete and Edit Users" privillege only can edit and successfully change other light admin's password
 - light admin with "Crete and Edit Users" privillege only cannot downgrade a full admin to light admin (fails silently)
 - light admin with "Crete and Edit Users" privillege only cannot downgrade a full admin to user (fails silently)
 - light admin with "Crete and Edit Users" privillege only cannot upgrade a user to light admin (fails silently) -> not expected
 - light admin with "Crete and Edit Users" privillege only can edit and successfully change full admin's password
 - light admin with "Crete and Edit Users" privillege only cannot deactivate a user
 - light admin with "Crete and Edit Users" privillege only can add or remove privileges form another  light admin within the permitted margin
 - light admin with "Create and Edit Users" pivilege only can edit themselves and remove the privilege. They can click save and still be
able to re-edit themselves and add the privilege back. If they logout and login after they removed the privilege from themselves, 
they are no longer able to Create and Edit Users.
 - light admin with "Create and Edit Users" privilege only cannot downgrade themselves to User or upgrade themselves to full admin
 - light admin with "Create and Edit Users" privilege only can create a new user and put them to multiple groups at creation
 - light admin with "Create and Edit Users" privilege only can create a new light admin with the same privileges as themselves and put them into System group at creation
 - light admin with "Create and Edit Users" privilege only cannot add users to groups (neither users nor light admins, fails silently)
 - light admin with "Create and Edit Users" privilege only cannot create a light admin and add them to more groups than System group at creation
 - full admin can create and edit light admins
 - light admin with "Create and Edit Users" and "Add users to groups" privileges cannot create and edit groups
 - light admin with "Create and Edit Users" and "Add users to groups" privileges can remove users from groups (via admin users menu edit)
 - light admin with "Create and Edit Users" and "Add users to groups" privileges can add users from groups (via admin users menu edit)
 - light admin with "Create and Edit Users" and "Add users to groups" privileges can activate and deactivate users
 - light admin with "Create and Edit Users" and "Add users to groups" privileges can remove privileges from other, more privileged light admins, but only those privileges which they themselves have. The other checkboxes are disabled in the UI. When the changed form is saved, the edited light admin is able to execute his privileges which the editing light admin could not change.
 - light admin with "Create and Edit Users" and "Add users to groups" privileges can downgrade a light admin to user and add them to a group and save successfully
 - light admin with "Create and Edit Users" and "Add users to groups" privileges can downgrade a full admin to user
 - light admin with "Create and Edit Users" and "Add users to groups" privileges can upgrade a user to light admin (but not full admin)
 - light admin with "Create and Edit Users" and "Create and Edit Groups" privileges can create a new group
 - light admin with "Create and Edit Users" and "Create and Edit Groups" privileges can edit existing group (including downgrade/upgrade)
 - light admin with "Create and Edit Users" and "Create and Edit Groups" privileges cannot add users to the newly created group
 - light admin with "Create and Edit Users" and "Create and Edit Groups" privileges cannot add users to edited, existing group
 - light admin with "Create and Edit Users" and "Create and Edit Groups" privileges cannot remove from or add existing users to groups via Users forms
 - light admin with all possible privileges can create another light admin with all possible privileges
 - light admin with all possible privileges can remove or add all possible privileges from existing light admins
 - full admin can upgrade an existing light admin to be a full admin

pwalczysko:
Bug (preexisting?) 

  - as light admin or  full admin, edit another full or light admin and downgrade them to user
  - htis will leave the Default group dropdown not populated with a group
  - click Save
  - observe a red warning as expected about Default group being a compulsory entry
  - try to rectify by adding the default group -> observe you cannot, you have to leave the form and come back to that user again, because Save does not become enabled after you rectified the mistake

Actually, there are 2 problems here, one is the failure to populate the default group automatically after downgrade to a user, two is the failure to re-enable the Save button.

pwalczysko:
In summary, I can see [Bug5](https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-319636802) and the [preexisting? bug](https://github.com/openmicroscopy/openmicroscopy/pull/5394#issuecomment-319667407).

The "Deactivate user" issue (does not work with "Create ...Users" privilege only) should probably be carded.

Other than that, happy with the behaviour.

pwalczysko:
Created https://trello.com/c/N9kCsCNL/97-deactivate-user for the adtivate/deactivate users issue.

will-moore:
Wow Petr, that's a LOT of testing! Thanks.
I think bug5 is simply the fact that the method we're using to create a regular User also allows adding of that User to multiple groups (even if the Light-Admin doesn't have this privilege) whereas the new method for creating a Light-Admin *doesn't* support adding the new user to groups.
So, a Light-Admin with ONLY Create/Edit User can create a regular user in a number of groups, but then can't remove that user from those groups.
Also, a Light-Admin with ONLY Create/Edit can't remove another Light-Admin from the system group to make them a regular user.

To fix bug5, I'd need to test whether the current Light-Admin user has "Modify Group-User" and if not then disable the "Role" field so that they can't click on "User".
I'll look at this, but the logic is getting a bit crazy now and we risk introducing major bugs while trying to fix very edge-case bugs.

This PR is a lot of work in the interests of trying to align the client and server!

will-moore:
To test last commit, check that this is now passing:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/617/testReport/OmeroWeb.test.integration.test_csrf/TestCsrf/test_create_user/

pwalczysko:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/618/testReport/OmeroWeb.test.integration.test_csrf/TestCsrf/test_create_user/
reports:  "FIXED"

will-moore:
TODO: hand dissabled Save on form error, update info for Create/Edit user and when user clicks this, auto-select to Modify Group/User too.

pwalczysko:
Works as expected. The simultaneous checkbox selection, the enabling of Save button as well as the new text of the explanatory snippet. 

Ready to merge.

will-moore:
@jburel Want me to handle the TODOs in description in a separate PR?
If so, anything else needed for this PR?

jburel:
Tests should minimally be added to this PR

joshmoore:
New method and tests look good. I _do_ wonder if the code @will-moore's written for adding/removing from the privileges doesn't suggest a server-side method to reduce potential race conditions, e.g., by loading the object server-side and only making the necessary changes rather than calling `setPrivileges`, but that's independent of this PR.

Recorded under https://trello.com/c/veYNrjiw/11-upsert-to-save-new-model-objects
cc: @manics @mtbc

PR #5395, opened 2017-07-25
BF bump to 5.6.0-m1

dgault:
# What this PR does

Bump bio-formats version to 5.6.0-m1

# Testing this PR

- Ensure builds and tests are green
- Test an import of any sample file which requires the JAI library


mtbc:
Which formats require JAI?

mtbc (approved):
Checked with various J2K and ND2 files.

PR #5397, opened 2017-07-25
Add testModifyUserCreateLight and provider

pwalczysko:
# What this PR does

Adds a test where light admin creates another light admin. The creator and created have always the same set of permissions. 4 sets of permissions are tested, which are matching the light admin doc and the permissions are easily read from https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-281472022 and https://github.com/openmicroscopy/openmicroscopy/pull/5101#issuecomment-281655507

# Testing this PR

CI suffices

# Related reading

This PR is showing that the lightAdmin creates lightAdmin workflow has no apparent blocker.
See https://github.com/openmicroscopy/openmicroscopy/pull/5394, cc @will-moore 

Ready to review

PR #3605, opened 2015-03-11
Blitz getattr with units

will-moore:
In 5.1 the API for methods that return units is unchanged from 5.0. E.g.
```
# omero.model.pixelsI - this works in 5.0 and 5.1
pixels.getPhysicalSizeX().getValue()
```
However, in the Blitz Gateway this is now broken
```
# pixelsWrapper
pixels.getPhysicalSizeX().getValue()
AttributeError: 'UnitsLength' object has no attribute 'getValue'
```
We get this error in Make_Movie script (although this used to work until recently)?
This PR updates Blitz Gateway ``` __getattr__ ``` to handle units and includes a test:

```
$ cd OmeroPy
$ ./setup.py test -t test/integration/gatewaytest/test_image.py -k testUnitsGetValue
```

Check that tests are passing at https://ci.openmicroscopy.org/view/Failing/job/OMERO-5.1-merge-integration/ and Make_Movie script works.

--no-rebase

PR #5400, opened 2017-07-27
Fix training print encode error

will-moore:
# What this PR does

Fixes failing Training examples test at
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/690/console

This was failing because the BlitzGateway BlitzObjectWrapper ```__getattr__``` which handles access to underlying ```omero.model``` objects decodes strings to unicode. If you then try to cast this to a string (for example to print to console) then we get a UnicodeEncodeError.
Test started failing because "Köhler" was set as a user name in a previous test added at
https://github.com/openmicroscopy/openmicroscopy/pull/5397/files#diff-d91232ef1e245a87e311c4a79a2a5f77R1692

For example:
```
>>> from omero.gateway import BlitzGateway
>>> conn = BlitzGateway(client_obj=client)
>>> exp = conn.getUser()

# This uses __getattr__ to return function that unwraps rstring and decodes to unicode
>>> exp.getLastName()
    u'K\xf6hler'

# Here we call getLastName on the simply unwrap the rstring to return a string
>>> exp._obj.getLastName().getValue()
    'K\xc3\xb6hler'
```

This PR removes the decoding to unicode in ```__getattr__``` so that e.g. ```BlitzGatewayObject.getLastName()``` returns a string.

This also makes the BlitzGatewayObject more consistent since previously, ```experimeterWrapper.getName()``` returns unicode but ```imageWrapper.getName()``` returns a string.

```
>>> exp.getName()
    u'will'
>>> i = conn.getObject("Image", 9551)
>>> i.getName()
    '438_01_R3D_D3D\xc3\xb6.dv'
```

# Testing this PR

 - Check that test above is now passing.
 - Test ``` bin/omero shell --login ``` examples above to see that we always get a String for ```exp.getName(), exp.getLastName()``` etc.
 - Test display of unicode characters in webclient. E.g. use "Köhler" as a user name, image name, tag name etc and check that this is displayed correctly everywhere.

# Related reading

TODO: Add to release notes - potentially breaking API change for Blitz Gateway in 5.4.0.

cc @chris-allan, @knabar, @joshmoore Any idea why this we used ```rv.val.decode('utf8')``` originally?

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1285,7 +1285,7 @@ class BlitzObjectWrapper (object):
                     rv = getattr(self._obj, attrName)
                     if hasattr(rv, 'val'):
                         if isinstance(rv.val, StringType):
-                            return rv.val.decode('utf8')
+                            return rv.val
                         # E.g. pixels.getPhysicalSizeX()
                         if hasattr(rv, "_unit"):
                             return rv
@@ -1305,8 +1305,7 @@ class BlitzObjectWrapper (object):
                 # If this is a _unit, then we ignore val
                 # since it's not an rtype to unwrap.
                 if not hasattr(rv, "_unit"):
-                    return (isinstance(rv.val, StringType) and
-                            rv.val.decode('utf8') or rv.val)
+                    return rv.val
             return rv
         raise AttributeError(
             "'%s' object has no attribute '%s'"
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1285,7 +1285,7 @@ class BlitzObjectWrapper (object):
                     rv = getattr(self._obj, attrName)
                     if hasattr(rv, 'val'):
                         if isinstance(rv.val, StringType):
-                            return rv.val
+                            return rv.val.decode('utf8')
                         # E.g. pixels.getPhysicalSizeX()
                         if hasattr(rv, "_unit"):
                             return rv
@@ -1305,7 +1305,8 @@ class BlitzObjectWrapper (object):
                 # If this is a _unit, then we ignore val
                 # since it's not an rtype to unwrap.
                 if not hasattr(rv, "_unit"):
-                    return rv.val
+                    return (isinstance(rv.val, StringType) and
+                            rv.val.decode('utf8') or rv.val)
             return rv
         raise AttributeError(
             "'%s' object has no attribute '%s'"
--- a/examples/Training/python/Connect_To_OMERO.py
+++ b/examples/Training/python/Connect_To_OMERO.py
@@ -84,13 +84,8 @@ if __name__ == '__main__':
     # New in OMERO 5
     print "Admins:"
     for exp in conn.getAdministrators():
-        fullName = exp.getFullName()
-        print "fullName", type(fullName)
-        print "Full Name...", fullName
-        print "str", str(fullName)
-        print "Full Name: %s" % fullName
         print "   ID: %s %s Name: %s" % (
-            exp.getId(), exp.getOmeName(), fullName)
+            exp.getId(), exp.getOmeName(), exp.getFullName())
 
     # The 'context' of our current session
     ctx = conn.getEventContext()
--- a/examples/Training/python/Connect_To_OMERO.py
+++ b/examples/Training/python/Connect_To_OMERO.py
@@ -84,8 +84,13 @@ if __name__ == '__main__':
     # New in OMERO 5
     print "Admins:"
     for exp in conn.getAdministrators():
+        fullName = exp.getFullName()
+        print "fullName", type(fullName)
+        print "Full Name...", fullName
+        print "str", str(fullName)
+        print "Full Name: %s" % fullName
         print "   ID: %s %s Name: %s" % (
-            exp.getId(), exp.getOmeName(), exp.getFullName())
+            exp.getId(), exp.getOmeName(), fullName)
 
     # The 'context' of our current session
     ctx = conn.getEventContext()
--- a/examples/Training/python/Connect_To_OMERO.py
+++ b/examples/Training/python/Connect_To_OMERO.py
@@ -62,7 +62,7 @@ if __name__ == '__main__':
 
     print "Member of:"
     for g in conn.getGroupsMemberOf():
-        print "   ID:", g.getName(), " Name:", g.getId()
+        print "   ID:", g.getId(), " Name:", g.getName()
     group = conn.getGroupFromContext()
     print "Current group: ", group.getName()
 
@@ -79,7 +79,7 @@ if __name__ == '__main__':
 
     print "Owner of:"
     for g in conn.listOwnedGroups():
-        print "   ID: ", g.getName(), " Name:", g.getId()
+        print "   ID: ", g.getId(), " Name:", g.getName()
 
     # New in OMERO 5
     print "Admins:"
--- a/examples/Training/python/Connect_To_OMERO.py
+++ b/examples/Training/python/Connect_To_OMERO.py
@@ -70,11 +70,11 @@ if __name__ == '__main__':
     owners, members = group.groupSummary()
     print "   Group owners:"
     for o in owners:
-        print "     ID: %s %s Name: %s" % (
+        print "     ID: %s UserName: %s Name: %s" % (
             o.getId(), o.getOmeName(), o.getFullName())
     print "   Group members:"
     for m in members:
-        print "     ID: %s %s Name: %s" % (
+        print "     ID: %s UserName: %s Name: %s" % (
             m.getId(), m.getOmeName(), m.getFullName())
 
     print "Owner of:"
@@ -84,7 +84,7 @@ if __name__ == '__main__':
     # New in OMERO 5
     print "Admins:"
     for exp in conn.getAdministrators():
-        print "   ID: %s %s Name: %s" % (
+        print "   ID: %s UserName: %s Name: %s" % (
             exp.getId(), exp.getOmeName(), exp.getFullName())
 
     # The 'context' of our current session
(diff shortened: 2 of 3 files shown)


jburel:
ready for a first round of review?

jburel:
Maybe also update 
``` print "     ID: %s %s Name: %s" % (
            m.getId(), m.getOmeName(), m.getFullName())
```
to ``ID, Username, Name``

will-moore:
Test passing at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/699/console
(search console log for "August KÃ¶hler")

joshmoore:
Note: still looking at this, but the call to `decode()` was introduced in gh-3605

joshmoore:
Old behavior:
```
t1
<type 'str'> string
<type 'unicode'> string
t2
<type 'str'> ünicode prefix
<type 'unicode'> ünicode prefix
t3
<type 'unicode'> ünicode prefix
<type 'unicode'> ünicode prefix
```

New behavior:
```
t1
<type 'str'> string
<type 'str'> string
t2
<type 'str'> ünicode prefix
<type 'str'> ünicode prefix
t3
<type 'unicode'> ünicode prefix
<type 'unicode'> ünicode prefix
```

cF:
```
from o_g_patched import BlitzObjectWrapper as BOW

class T1(object):
    val = "string"

class T2(object):
    val = "ünicode prefix"

class T3(object):
    val = u"ünicode prefix"

class Mock(object):
    def __init__(self):
        self.t1 = T1()
        self.t2 = T2()
        self.t3 = T3()

x = BOW(obj=Mock())

print "t1"
print type(x.t1), x.t1
print type(x.getT1()), x.getT1()
print "t2"
print type(x.t2), x.t2
print type(x.getT2()), x.getT2()
print "t3"
print type(x.t3), x.t3
print type(x.getT3()), x.getT3()
```

i.e. for minimally anything that's LIKE-A `RType` in that it has a `val` field, the return type of `getX` is going to change.

will-moore:
@joshmoore https://github.com/openmicroscopy/openmicroscopy/pull/3605 didn't introduce the ```decode()```, it just refactored to handle Units, but the String handling was unchanged.

joshmoore:
Noted, thanks. Tracking further back:

```
commit 8ff935dce649518e95daaf7f1ed159764eb9b1eb
Author: cneves-x <cneves-x@05709c45-44f0-0310-885b-81a1db45b4a6>
Date:   Thu Apr 8 15:34:16 2010 +0000
```

but that raises even further the question of how we gauge the breaking-ness of something like this.

chris-allan:
While the inconsistency is something we should definitely resolve I don't think this is a good change. In my eye a `unicode` return value is correct. Anyone who was properly handling the encoding previously will now be broken and will have to review every instance where they call **any** getter method via the gateway and associated wrapper objects.

The reason why `decode()` was used in the first place was to establish a uniform return value (`unicode`) where encoding is not implicitly enforced. Ice requires that we encode strings when passing them to the server. The Ice Python bindings do not support `unicode` and will not perform any sort of encoding for us. We use UTF-8 for that although we don't explicitly enforce this.

There are many things involved in coercing a `unicode` value for printing to STDOUT. These include how the Python version was compiled, the default encoding, the locale setting, and many others. Have we established why this was failing? It cannot be just because there is a `unicode` value being returned.

My system for reference:

```
In [1]: import sys

In [2]: a = u"ünicode prefix"

In [3]: type(a)
Out[3]: unicode

In [4]: a
Out[4]: u'\xfcnicode prefix'

In [5]: print a
ünicode prefix

In [6]: print 'foo ', a
foo  ünicode prefix

In [7]: sys.getdefaultencoding()
Out[7]: 'ascii'

In [8]: print 'foo %s' % a
foo %s ünicode prefix

In [9]: b = a.encode('UTF-8')

In [10]: b
Out[10]: '\xc3\xbcnicode prefix'

In [11]: print b
ünicode prefix

In [12]: print 'foo ', b
foo  ünicode prefix

In [13]: print 'foo %s' % b
foo ünicode prefix
```

will-moore:
@chris-allan Given the lack of support for unicode in Ice and our various degrees of object wrapping and method overwriting in BlitzGateway, I don't see any way of ensuring we get unicode returned everywhere. So the easiest way to resolve the inconsistency is to return strings everywhere.
I don't know if we're "properly" handling the return values from Blitz Gateway get*() methods, but this PR hasn't caused any obvious breakages in OMERO.py tests or OMERO.web.

This is failing in the test above, but I don't know what the python setup is there (except for ```omego download python --branch OMERO-DEV-merge-build --ice 3.6```).
```
print "   ID: %s %s Name: %s" % (exp.getId(), exp.getOmeName(), exp.getFullName())
```
I think it was also failing for me locally when I was originally testing this, but then I set PYTHONIOENCODING (or something similar) and now I can't reproduce locally. (can't reset).

joshmoore:
> Given the lack of support for unicode in Ice and our various degrees of object wrapping and method overwriting in BlitzGateway, I don't see any way of ensuring we get unicode returned everywhere.

My feeling after seeing the output from my test script was certainly that we're in a very ambiguous, confusing and really dangerous place.

> So the easiest way to resolve the inconsistency is to return strings everywhere.

I'd definitely suggest we not take the easiest route if we're going to touch this code at all. I could see though making this code configurable so that the consumer can choose on of a couple of strategies:

 * LEGACY: the current behavior
 * STRINGS: what you're proposing
 * UNICODE: the reverse
 * MIXED: for the behavior of this PR if you can see a real use
 * etc.

One way of achieving that would be to make use of the `codecs` module (https://pymotw.com/2/codecs/ etc) internally, and letting the user set any of the relevant options on the BlitzGateway itself.

> I don't know if we're "properly" handling the return values from Blitz Gateway get*() methods, but this PR hasn't caused any obvious breakages in OMERO.py tests or OMERO.web.

But to some extent, that's not a sufficient test. Changing the return type to the user is a significant breaking change.

will-moore:
Added 
```
import sys
reload(sys)
sys.setdefaultencoding('utf-8')
```
as suggested by Josh: https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7810#p15910
Some say this is dangerous, but probably OK for testing:
https://stackoverflow.com/questions/11741574/how-to-print-utf-8-encoded-text-to-the-console-in-python-3/11742928#11742928

will-moore:
Tests now passing. See 
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/724/console 
printing Köhler as:
```
ID: 2885 UserName: d05275b2-3422-47f9-bbe8-3f825efdf593 Name: August KÃ¶hler
```

joshmoore:
To clarify: I wasn't pointing at the use of `setdefaultencoding` as an infile solution but at something more system or process wide. Considering what the file is intended to be, I think we should really be suggesting a plausible workaround.

That being said, the printing here still looks broken, so we likely still need to determine the underlying cause (locale, LANG, etc) and determine the various possible states.

will-moore:
Removing 'infile' solution. Needs adding to ci machine. cc @sbesson

will-moore:
As discussed: opened Docs PR to explain Unicode in BlitzGateway. https://github.com/openmicroscopy/ome-documentation/pull/1749

Test is failing again at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/729/console
Need to discuss best solution for that.

will-moore:
I have used a different approach for printing of unicode.
However, this may break printing of regular strings, since they will be converted to unicode and back to strings which can cause errors as discussed https://pythonhosted.org/kitchen/unicode-frustrations.html#frustration-4-now-it-doesn-t-take-byte-strings
I could use the ```kitchen``` library discussed there, but it would be nice not to have to add a dependency just for this.
It seems like my previous commit https://github.com/openmicroscopy/openmicroscopy/pull/5400/commits/787413312fe361275dd468a09ea56d519c204c3a which fixed the tests is probably the best way to go. But @joshmoore didn't like it previously. What's the preferred solution?

chris-allan:
Using `sys.setdefaultencoding()` should always be avoided. If you want to read more about why it's a bad idea this is a good reference:

 * https://stackoverflow.com/questions/3828723/why-should-we-not-use-sys-setdefaultencodingutf-8-in-a-py-script

I personally don't have a problem with the byte string issue. If these fields legitimately have bytes in them then we have even bigger issues to deal with. I would not expect the current `BlitzGateway` implementation to work with such strings in the first place anyway.

will-moore:
@chris-allan Re: "If these fields legitimately have bytes in them then we have even bigger issues to deal with."
This is exactly the issue we have to deal with, since e.g. ```ObjectWrapper.getName()``` returns a string. As do a bunch of other methods listed at https://github.com/openmicroscopy/ome-documentation/pull/1749

```
In [5]: import sys
In [6]: import codecs
In [7]: sys.stdout = codecs.getwriter('utf8')(sys.stdout)

In [8]: p = conn.getObject("Project", 13578)
In [9]: name = p.getName()
In [10]: name
Out[10]: 'API TEST foo K\xc3\xb6hler'
In [11]: print name
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-11-2e5da873a229> in <module>()
----> 1 print name

/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/codecs.pyc in write(self, object)
    355         """ Writes the object's contents encoded to self.stream.
    356         """
--> 357         data, consumed = self.encode(object, self.errors)
    358         self.stream.write(data)

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 14: ordinal not in range(128)
```

will-moore:
Test is passing again: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/732/
```
ID: 2821 UserName: bab43208-e525-4a33-b0be-0140fd975085 Name: August KÃ¶hler
```

will-moore:
Discussed: since this is now green, let's merge this and add examples in a new PR that cause failures for ```print project.getName()``` etc.

PR #5404, opened 2017-08-01
Cleanup locks if plpgsql ome_nextval fails

manics:
# What this PR does

`ome_nextval` calls:
1. `pg_advisory_lock(1, Lid)` before calling
2. `nextval(seq)`
3. `pg_advisory_unlock(1, Lid)`

If the call to `nextval(seq)` fails `pg_advisory_unlock` is never called, leading to a potential deadlock. This PR replaces `pg_advisory_lock`/`pg_advisory_unlock` with a single call to `pg_advisory_xact_lock` which automatically unlocks at the end of the transaction.

In addition this removes the `BEGIN`-`EXCEPTION`-`END` wrapper since the [lock functions exist in PostgreSQL 9.1](https://www.postgresql.org/docs/9.1/static/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS) which is less than the minimum OMERO requirement.

This also includes an SQL file for manually testing the IDR.

# Testing this PR

See https://trello.com/c/1DGiYzfM/61-plpgsql-omenextval-may-deadlock-with-a-readonly-user

It's not reliably reproducible

### DIFF
--- a/components/dsl/resources/ome/dsl/psql-footer.vm
+++ b/components/dsl/resources/ome/dsl/psql-footer.vm
@@ -139,9 +139,18 @@ BEGIN
           INSERT INTO _lock_ids (id, name) VALUES (Lid, seq);
       END IF;
 
-      PERFORM pg_advisory_xact_lock(1, Lid);
-      PERFORM nextval(seq) FROM generate_series(1, increment);
-      SELECT currval(seq) INTO nv;
+      PERFORM pg_advisory_lock(1, Lid);
+
+      BEGIN
+          PERFORM nextval(seq) FROM generate_series(1, increment);
+          SELECT currval(seq) INTO nv;
+      EXCEPTION
+          WHEN OTHERS THEN
+              PERFORM pg_advisory_unlock(1, Lid);
+          RAISE;
+      END;
+
+      PERFORM pg_advisory_unlock(1, Lid);
 
       RETURN nv;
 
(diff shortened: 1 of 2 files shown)


manics:
--rebased-to #5405

manics:
--exclude

manics:
I've manually applied `sql/psql/patches/psql-footer.sql` to the test IDR database:

    psql -h localhost -U omero idr < psql-footer.sql

This ***must not*** be applied to production until it is fully reviewed.

manics:
It turns out `pg_advisory_xact_lock` doesn't work, presumably because `ome_nextval` is called multiple times in the same transaction? Instead revert to the old `pg_advisory_lock` `pg_advisory_unlock` functions, but with a `BEGIN EXCEPTION END` block to ensure `pg_advisory_unlock` is called if an error occurs in `nextval` or `currval`.

I've manually applied the updated sql/psql/patches/psql-footer.sql to the test IDR database again.

manics:
The patch is now on idr-next too following a lockup over the weekend:
`psql < /opt/omero/server/OMERO.server/sql/psql/patches/psql-footer.sql`

joshmoore:
I don't foresee us performing an upgrade until we move to 5.4 where this has been handled more thoroughly. I'd suggest we close since the patch has already been manually applied.

manics:
👍

PR #5405, opened 2017-08-01
Cleanup locks if plpgsql ome_nextval fails

manics:
# What this PR does

`ome_nextval` calls:
1. `pg_advisory_lock(1, Lid)`
2. `nextval(seq)`
3. `pg_advisory_unlock(1, Lid)`

If the call to `nextval(seq)` fails `pg_advisory_unlock` is never called, leading to a potential deadlock. This PR adds a `BEGIN`-`EXCEPTION`-`END` wrapper to ensure `pg_advisory_unlock` is called when an error occurs.

In addition this removes the `BEGIN`-`EXCEPTION`-`END` wrapper pg_advisory_{lock,unlock}` around since the [lock functions exist in PostgreSQL 9.0](https://www.postgresql.org/docs/9.0/static/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS) which is less than the minimum OMERO requirement.

This is a modified version of `metadata53` gh-5404.

# Testing this PR

See https://trello.com/c/1DGiYzfM/61-plpgsql-omenextval-may-deadlock-with-a-readonly-user

It's not reliably reproducible


manics:
--rebased-from #5404

manics:
--exclude

manics:
Rebased to include #5411

manics:
Added the schema change following https://docs.openmicroscopy.org/contributing/schema-changes.html

--breaking

jburel:
breaking job is currently disabled (This was running on trout)
Do you need to have the PR build somewhere?

manics:
If if passes code review it could probably go straight into mainline since it's changing a function, no tables are affected.

mtbc:
The code changes do look good to me: your thinking is agreeably clear. There are some minor outstanding to-do's re. the usual DB patch PR workflow before merge; if that's a surprise, I'm happy to enumerate those I notice but most obviously upgrade scripts.

This has been getting live testing for a little while on a `metadata53` deployment?

https://trello.com/c/1DGiYzfM/95-plpgsql-omenextval-may-deadlock-with-a-readonly-user#comment-598c562f15324c293eb58a1d presumably requires work on OMERO.server. However, if we *could* obviate the need for the DB-level check then that would be nice as in PostgreSQL `EXCEPTION` isn't cheap.

manics:
It was in use on the idr-testing server, but not on production.

manics:
There was a lockup over the weekend so I've just patched our new production DB too https://github.com/openmicroscopy/openmicroscopy/pull/5404#issuecomment-322154347

manics:
With respect to upgrade scripts, isn't `psql-footer.sql` unconditionally run, so would overwrite the old definition?

mtbc:
I wasn't aware that it is.

joshmoore:
`footer.sql` is *not* run on upgrade, but it's contents (where necessary) are copied to the upgrade.

manics:
Updated

mtbc:
* We need a `OMERO5.4DEV__3/OMERO5.4DEV__2.sql`.
* `OMERO5.4DEV__3/OMERO5.3__0.sql` needs bumping from patch 2 to 3 at the start (grep it for `OMERO5.4DEV.*2`).

jburel:
Tested locally: upgraded from patch 2
```
psql -h localhost -U omero omero < sql/psql/OMERO5.4DEV__3/OMERO5.4DEV__2.sql 
BEGIN
CREATE FUNCTION
 omero_assert_db_version 
-------------------------
 
(1 row)

DROP FUNCTION
INSERT 0 1
CREATE FUNCTION
UPDATE 1
                                 status                                 
------------------------------------------------------------------------
                                                                       +
                                                                       +
                                                                       +
 YOU HAVE SUCCESSFULLY UPGRADED YOUR DATABASE TO VERSION OMERO5.4DEV__3+
                                                                       +
                                                                       +
 
(1 row)

COMMIT

```
Server restarted and no problem to log in

jburel:
@mtbc are you happy with the latest change?
If so we could merge this PR so @dominikl can move with another DB patch cf. https://github.com/openmicroscopy/openmicroscopy/pull/5436

mtbc:
I'll list for Monday in case I don't get to this today.

mtbc (approved):

PR #5408, opened 2017-08-01
No color set

jburel:
# What this PR does

Handle possible NPE when color is not set

# Testing this PR

Code review should be enough
I have not been able to reproduce the issue with the image I have used

# Related reading

https://www.openmicroscopy.org/qa2/qa/feedback/17808


waxenegger:
Looking at the code the null checks are a reasonable addition. If I didn't make a mistake following the stack trace the old color would have to have been null.

Also I played a bit with the most recent insight artifact and color changes. Seems to not give me an exception, also not when I revert.

PR #5409, opened 2017-08-01
Travis update

jburel:
# What this PR does

Use trusty to build 
This is necessary to build with newer version of pytables
and newer version of ice

 * building using ``ant`` fails all the time, need to use ``./build.py``.
 * installation of python dependencies needs to be explicitly done via
``sudo apt-get`` during ``before_install:`` otherwise the dependencies will not be available to the container. This means that the python tests will fail with for example ``No module Ice``. Change ``sudo`` option to ``required``. I did not notice an increase in the build time with that change.  See https://github.com/travis-ci/travis-ci/issues/8048
 * update ``numexpr`` via pip to avoid some table test failure  e.g.
``test/unit/tablestest/test_servants.py:349:`` will fail with ``ValueError: Using `oa_ndim == 0` when `op_axes` is NULL. Use `oa_ndim == -1` or the MultiNew iterator for NumPy <1.8 compatibility``

# Testing this PR
Check that travis is green

See https://trello.com/c/gVsUkLH8/100-travis-upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jdk: oraclejdk8
 # This (sudo: false) is needed to "run on container-based infrastructure" on
 # which cache: is available
 # http://docs.travis-ci.com/user/workers/container-based-infrastructure/
-sudo: false
+sudo: required
 
 # http://docs.travis-ci.com/user/caching/#Arbitrary-directories
 cache:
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
Looks generally OK.

I'm uncertain that sudo is required.  Every package you are installing is in the trusty whitelist (https://github.com/travis-ci/apt-package-whitelist) because I requested them months ago and they have been available for a while now.

Did you find what build.py is doing which makes a difference?  It's not related to sudo builds being run on different VMs?

jburel:
without ``sudo: required`` it is not possible to install the python dependencies in ``before_install`` step.
When https://github.com/travis-ci/travis-ci/issues/8048 is resolved we will probably have to revisit. Note that using ``group:edge`` as suggested in issue did not work

jburel:
@rleigh-codelibre works now with ant. Still unclear why it was not before (without explicit error it is tricky to figure error)

jburel:
@sbesson comment added to this PR too

jburel:
--rebased-to #5447

PR #5410, opened 2017-08-02
projectPixels: t range overshoots (unless full range is selected)

waxenegger:
Piggy-backing on the method projectPixels (ProjectionService) to save a projection as a new image (for iviewer) I noticed that I got errors when setting the range of Ts not to the full range or 0:0.

Cross-checking with insight,I got the same error: 

`
ServerExceptionClass = "ome.conditions.ValidationException"
    message = "C=0 or T=22 out of range for Pixels Id 104: T '22' greater than sizeT '22'."
`

This can be reproduced with images that have more than 1 T and Z. Change the T range in the projection dialog to something that is not the entire range and you should get the error.

### DIFF
--- a/components/tools/OmeroJava/test/integration/ProjectionServiceTest.java
+++ b/components/tools/OmeroJava/test/integration/ProjectionServiceTest.java
@@ -235,6 +235,19 @@ public class ProjectionServiceTest extends AbstractServerTest
     }
 
     /**
+     * Test the projection with a restricted but valid timepoint range.
+     *
+     * @throws Exception Thrown if an error occurred.
+     */
+    @Test
+    public void testTimepointIntervalNotFullRange() throws Exception {
+        Pixels pixels = importImage();
+        List<Integer> channels = Arrays.asList(0);
+        projectImage(pixels, 2, 4, 1, 3, 1,
+                ProjectionType.MAXIMUMINTENSITY, null, channels);
+    }
+    
+    /**
      * Test the projection with an invalid timepoint range.
      *
      * @throws Exception Thrown if an error occurred.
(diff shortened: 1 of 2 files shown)


jburel:
Thanks for finding that one.
Strange that it was not captured before. Could you add some tests, for example, to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroJava/test/integration/ProjectionServiceTest.java?

waxenegger:
no problem, I'll add a test

jburel:
@waxenegger https://github.com/openmicroscopy/openmicroscopy/pull/5411 has been merged so you will have to merge develop this should fix the travis issue for now

components/tools/OmeroJava/test/integration/ProjectionServiceTest.java
```diff
@@ -235,6 +235,19 @@ public void testSameTimepointInterval() throws Exception {
...
     /**
+     * Test the projection with a restricted but valid timepoint range.
+     *
+     * @throws Exception Thrown if an error occurred.
+     */
+    @Test
+    public void testTimepointIntervalNotFullRange() throws Exception {
+        Pixels pixels = importImage();
+        List<Integer> channels = Arrays.asList(0);
+        projectImage(pixels, 2, 4, 1, 3, 1,
```
jburel, line 246: I will check that the projected image has the correct number of timepoints

waxenegger: Feel free, I tested it locally. More importantly, the test has an assert for the number of t points in the new image: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroJava/test/integration/ProjectionServiceTest.java#L121

jburel: I will need to think of  the best way of comparing
Might need to look at combining stuff I did for LUT.
Not the goal of the PR
Thanks for the test

PR #5411, opened 2017-08-02
Use precise

jburel:
# What this PR does

specify dist

# Testing this PR

Nothing to test

PR #5413, opened 2017-08-03
Use precise (rebased onto metadata53)

manics:
This is the same as gh-5411 but rebased onto metadata53.

----

# What this PR does

specify dist

# Testing this PR

Nothing to test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: java
-
+dist: precise
 jdk: oraclejdk8
 
 # This (sudo: false) is needed to "run on container-based infrastructure" on


manics:
--rebased-from #5411

joshmoore:
Hopefully we can go green again!

jburel:
Since that patch, the mainline now uses trusty see https://github.com/openmicroscopy/openmicroscopy/pull/5409
That changes was not ported to metadata53
I can open a PR if it is useful

joshmoore:
If you are happy with the state of the mainline, yes, a PR to synchronize would be useful.

jburel:
it works fine on develop. We might have to change in few months when the problem is fixed but that should be easy.
I will open a PR today

jburel:
This will not be possible metadata 53 is missing a requirement file specific for ice35 
this will lead to some ice version mismatch

manics:
What are the missing requirements?

jburel:
For 5.3.3 I had to introduce an extra file https://github.com/openmicroscopy/openmicroscopy/pull/5324 (not the best! but did not want to change everything in a point release)
otherwise web will not work with ice 3.5 (depending on installation).
and I use ``requirements-py27-all-ice35.txt`` in travis.yml
The ``requirements-py27-all.txt`` will install ``zeroc-ice`` for ice 3.6

joshmoore:
Any reason to not also introduce the file on metadata53?

PR #5414, opened 2017-08-03
Allow child-options to be set when deleting MapAnnotations

manics:
# What this PR does

Allows child-options to be passed to the `Delete2` command, as suggested by @mtbc to workaround the performance issues described in https://trello.com/c/kRPgBR36/57-linked-annotation-deletion-is-very-slow-due-to-53-graph-deletion-rules#comment-597c44d077236b6fb2e8fc40

# Testing this PR

Child options in can be passed using `--localcfg`. For example:

    omero metadata populate --context deletemap 
        --localcfg '{"ns":"openmicroscopy.org/omero/bulk_annotations",
        "childoptions":[{"excludeType":["Annotation"]}]}'
        Well:125 --report

will delete `WellAnnotationLink`s to `MapAnnotation`s in namespace `openmicroscopy.org/omero/bulk_annotations` for `Well:125`, but will never delete the `MapAnnotation` even if it is orphaned since this introduces a significant performance degradation.

## Updated 20170823

--depends-on #5464

This should be a faster version of the above: `--localcfg '{"ns":"openmicroscopy.org/omero/bulk_annotations", "typestoignore":["Annotation"]}'`

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/populate_metadata.py
+++ b/components/tools/OmeroPy/src/omero/util/populate_metadata.py
@@ -36,6 +36,7 @@ from collections import defaultdict
 
 import omero.clients
 from omero import CmdError
+from omero.cmd.graphs import ChildOption
 from omero.rtypes import rlist, rstring, unwrap
 from omero.model import DatasetAnnotationLinkI, DatasetI, FileAnnotationI
 from omero.model import OriginalFileI, PlateI, PlateAnnotationLinkI, ScreenI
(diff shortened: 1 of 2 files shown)


mtbc:
Sorry it didn't help, thanks for trying it.

joshmoore:
@manics: do you see benefit in this regardless of the performance issue(s)?

manics:
Potentially. Are you happy with the concept of using `--localcfg` for additional arguments, or do you consider this "development only" and all args should be configured with a CLI parameter?

joshmoore:
Wouldn't consider it development only (since we want to _reduce_ the number of commits on the metadata branches). If it's something that's needed, then we should finish designing the API and get it available everywhere.

manics:
I meant do you see the use of `--localcfg` as an easy way to get additional parameters into the metadata plugin for the IDR, but adding an explicit CLI option (e.g. `--delete-child-args` if we decide the work we've done is production quality), or are you happy with `--localcfg` going onto mainline too? It's a bit like `python_function(arg1, arg2, arg3)` (explicit arguments) vs `python_function(**kwargs)` (pass in anything on the CLI and let the function deal with it)

manics:
Additional options are needed to support https://github.com/openmicroscopy/openmicroscopy/pull/5464

manics:
I've added an option for `typestoignore`, and some tests for this PR.

Note that I had to define the list of fixtures as a list of types and construct each fixture on each method call `fixture = fixture(), fixture.init(self)` unlike in the other tests in order for the new tests to pass. I've got no idea why.

joshmoore:
Perhaps what we should actually be considering is moving from:
`omero metadata populate --context deletemap `
to
` omero metadata delete`
etc.

manics:
Added to https://trello.com/c/iFe4XS0z/226-refactor-rewrite-populatemetadatapy

joshmoore:
This is merged into the build currently released as prod43. Merging it, but further review for the mainline may be necessary.

PR #5416, opened 2017-08-03
state_copy_paste

rgozim:
Added a simple boolean property to WEBCLIENT object (in base_container.html) which acts as a flag to enable / disable the paste option in the 'rendersettings' context menu.

### Testing this PR

Please ensure you have logged out and logged in before testing

1. Right click on image in left tree and check if "Rendering Settings" > "Paste and save" is disabled

2. Copy an image from the left hand tree and you should be able to select another image to paste on ( "Paste and save"  no longer disabled)

3 . Click to copy an image from the right hand "Preview" tab. Click on an image on the left tree and the option to  "Paste and save" should be available.

Trello card:
https://trello.com/c/GEekfV57/71-rnd-settings


pwalczysko:
the enabling and disabling of the menu works as expected. 
One problem detected which was not caused by this PR (repeatable on eel latest). Card created https://trello.com/c/W5tqjH0s/101-rnd-copy-and-paste-from-others.

Ready to merge.

PR #5426, opened 2017-08-09
Add description for pixeldata max_plane_{width,height} properties

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2017-July/006561.html

Replaces https://github.com/openmicroscopy/ome-documentation/pull/1718, this PR adds a minimal description of the plane size cutoff properties for generating pyramids server-side.

Once merged, this change should be propagated to `sysadmin/config.txt` in ome-documentation via the autogeneration script.


jburel:
Do we want to add that the values will be ignored for float and double pixels type?

sbesson:
Can certainly add these limitations as we have been short in explaining them. Following https://github.com/openmicroscopy/bioformats/pull/2915#issuecomment-321554891, once we have a dedicated page for the current OMERO pyramid format we can also refer to it in these properties.

joshmoore:
LGTM.

PR #5427, opened 2017-08-10
Allow the local Maven repository to be overriden 

sbesson:
## What this PR does

This PR primarily adds the ability to control the parameter defining the local Maven repository used for retrieving and storing the JARs during the build of the Java components. An [Ivy property](http://ant.apache.org/ivy/history/latest-milestone/settings/property.html) had been previously defined but was not externally settable without modifying `etc/ivysettings.xml`.

7220b43 sets the `override` attribute to `true` so that the `local-maven2-dir` property can be overriden in command-line or via `etc/local.properties`. 4dd2745 and 9053637 additionally deprecate this property and introduce a new property, `maven.repo.local` in agreement with the Maven naming scheme, allowing to control the local Maven repository.

### Without this PR

Prior to testing, rename your local Maven repository as `~/.m2.bak`. Without this PR, any of the command below should lead to the Java artifacts retrieved under the default `~/.m2/repository`.

- `./build.py -Dlocal-maven2-dir=/tmp/.m2/repository build-dev`
- `echo "local-maven2-dir=/tmp/.m2/repository2" > etc/local.properties && ./build.py build-dev`

### With this PR:

Run all or part of the following commands:

- `./build.py -Dlocal-maven2-dir=/tmp/.m2/repository1 build-dev`
- `./build.py -Dmaven.repo.local=/tmp/.m2/repository2 build-dev`
- `echo "local-maven2-dir=/tmp/.m2/repository3" > etc/local.properties && ./build.py build-dev`
- `echo "maven.repo.local=/tmp/.m2/repository4" > etc/local.properties && ./build.py build-dev`

In all cases above, the Java artifacts should be stored in the correct temporary location.

## Related reading

Under some conditions, non-defaut Maven repository needs to be specified for building OMERO. See https://github.com/ome/homebrew-alt/pull/136 for a use case.

### DIFF
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -16,7 +16,8 @@
   <properties file="${ivy.settings.dir}/build.properties"/>
   <properties file="${ivy.settings.dir}/local.properties"/>
   <properties file="${ivy.settings.dir}/../target/omero.version"/>
-  <property name="local-maven2-dir" value="${user.home}/.m2/repository/"/>
+  <property name="local-maven2-dir" value="${user.home}/.m2/repository/"
+      override="false"/>
 
   <settings defaultResolver="${omero.resolver}"/>
 
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -16,7 +16,9 @@
   <properties file="${ivy.settings.dir}/build.properties"/>
   <properties file="${ivy.settings.dir}/local.properties"/>
   <properties file="${ivy.settings.dir}/../target/omero.version"/>
-  <property name="maven.repo.local" value="${user.home}/.m2/repository/"
+  <property name="local-maven2-dir" value="${user.home}/.m2/repository/"
+      override="false"/> <!-- deprecated - use maven.repo.local instead -->
+  <property name="maven.repo.local" value="${local-maven2-dir}"
       override="false"/>
 
   <settings defaultResolver="${omero.resolver}"/>
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -16,7 +16,7 @@
   <properties file="${ivy.settings.dir}/build.properties"/>
   <properties file="${ivy.settings.dir}/local.properties"/>
   <properties file="${ivy.settings.dir}/../target/omero.version"/>
-  <property name="local-maven2-dir" value="${user.home}/.m2/repository/"
+  <property name="maven.repo.local" value="${user.home}/.m2/repository/"
       override="false"/>
 
   <settings defaultResolver="${omero.resolver}"/>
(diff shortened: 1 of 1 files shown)


sbesson:
The main driver for conducting this work might be deprecated for OMERO 5.4.0 - see https://trello.com/c/avYucEd4/118-disable-omero-homebrew-job. That being said, being able to define/override the dependency directory might be useful under other circumstances e.g. to avoid concurrent jobs to overwrite the same SNAPSHOTS or to be sure OMERO is built from a fresh Maven without needing to completely remove `~/.m2`. What are your thoughts @jburel @mtbc ? Happy either to list this PR in the context of 5.4.0 or close it and defer to a further minor release.

mtbc:
I think it's useful also because sometimes people have .m2 somewhere else, e.g., to where their maven-aware IDE is downloading, so as not to duplicate caches.

sbesson:
Given the conflict with #5462, I propose we review this PR so that conflicts can be fixed. Being able to specify a custom Maven repository would probably help the testing of #5462

rleigh-codelibre:
The change looks fine and builds fine for me using the override.  Please merge.

sbesson:
The only final consideration here with regard to compatibility is that it is possible that some downstream consumer was using the internal `maven-local2-dir` property after importing `etc/ivysettings.xml`. Rather than directly dropping the property, I could push a commit aliasing (and deprecating) the old property as:

```
<property name="maven-local2-dir" value="${maven.repo.local}"/>
```

/cc @joshmoore @jburel

joshmoore:
Ha. I had thought about this. :) I would have suggested the reverse those:

```
<property name="maven.repo.local" value="${maven-local2-dir}"/>
```

so that someone using the old value gets the new behavior.

mtbc:
Also deprecate it?

joshmoore:
Via comment, sure. Unfortunately, no explicit way to do so.

sbesson:
Yep http://ant.apache.org/ivy/history/latest-milestone/settings/property.html does not give any deprecation key but more than happy to add a comment.

mtbc:
Could also go into release notes.

sbesson:
Pushed 9053637 and updated the description to match the last change. Happy to capture this in the release notes if there is a staging file or add to a card.

mtbc:
So testing should now include use of the internal `maven-local2-dir` property?

sbesson:
;( I had rewritten the PR description but apparently failed to save it. Updating now.

sbesson:
@mtbc that's the description updated with the exhaustive list of scenarios.

mtbc:
`./build.py -Dmaven-local2-dir=/tmp/.m2/repository1 build-dev` leaves `/tmp/.m2` empty and fills `~/.m2/repository`.

sbesson:
@mtbc is the status of https://github.com/openmicroscopy/openmicroscopy/pull/5427#issuecomment-324892999 with or without this PR?

mtbc:
With the PR.

sbesson:
Just realized that since https://github.com/openmicroscopy/openmicroscopy/pull/5427#issuecomment-324566755, I introduced a typo in the variable name which I propagated to the description: `maven-local2-dir` should be `local-maven2-dir`. Updated the description to use the correct name and will relist for Monday.

mtbc (approved):

PR #5429, opened 2017-08-12
FindIce: Add support for Ice 3.7.0

rleigh-codelibre:
# Context

See [upstream merge request](https://gitlab.kitware.com/cmake/cmake/merge_requests/1137) and [mailing list thread](http://public.kitware.com/pipermail/cmake/2017-August/066028.html) from the University of Swansea.

# Purpose

This PR will enable future use of Ice 3.7 on Linux, MacOS X and Windows with the officially provided Ice 3.7 builds and distribution-provided builds.  In the present, the purpose of the PR is to ensure that the Ice 3.5 and 3.6 support we use today will not regress as a result of the addition of Ice 3.7 support.  Local testing worked; the PR is to ensure the changes don't negatively affect the CI infrastructure and to keep us in sync with the upstream logic.

# Changes

- Search for NuGet packages on Windows, which is the new way the official binary builds are distributed for 3.7.0
- Search for C++11 versions of libraries, with `++11` suffix, taking Windows naming inconsistencies into account (Windows libraries have a `37` suffix, but this is placed *before* the `++11`.
- Add new Ice 3.7 programs
- Document new Ice 3.7 components
- Add debug output for search suffixes
- Simplify debug logic

# Testing

Check builds remain green.  No manual building is required.

### DIFF
--- a/components/tools/OmeroCpp/cmake/FindIce.cmake
+++ b/components/tools/OmeroCpp/cmake/FindIce.cmake
@@ -10,8 +10,16 @@
 #
 # This module supports multiple components.
 # Components can include any of: ``Freeze``, ``Glacier2``, ``Ice``,
-# ``IceBox``, ``IceDB``, ``IceGrid``, ``IcePatch``, ``IceSSL``,
-# ``IceStorm``, ``IceUtil``, ``IceXML``, or ``Slice``.
+# ``IceBox``, ``IceDB``, ``IceDiscovery``, ``IceGrid``,
+# ``IceLocatorDiscovery``, ``IcePatch``, ``IceSSL``, ``IceStorm``,
+# ``IceUtil``, ``IceXML``, or ``Slice``.
+#
+# Ice 3.7 and later also include C++11-specific components:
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
Green in https://ci.openmicroscopy.org/job/OMERO-DEV-merge-cpp/648/

joshmoore:
Without this PR, would the build progress on an Ice 3.7 system or fail early?

rleigh-codelibre:
@joshmoore Without the PR with Ice 3.7: On Unix it will find Ice while on Windows it will fail to find it due to the packaging and naming differences.  But on Unix it will fail due to the IceUtil component being removed and not found.

With the PR and Ice 3.7: It will find Ice on all platforms, but will continue to fail due to IceUtil being not found.

With or without the PR and Ice 3.5 and 3.6: There should be no change in behaviour; Ice and all its components will be discovered on all platforms.

jburel:
Discussed today with @rleigh-codelibre, as indicated in the description the aim of this PR was to check potential impact of intro of Ice 3.7.
This will not be included in 5.4, closing to avoid confusion.
To be re-opened when we start work on Ice 3.7 support

rleigh-codelibre:
Merged upstream.

PR #5435, opened 2017-08-14
Deprecate ReverseIntensityMap/Context

dominikl:
# What this PR does

Deprecate 'ReverseIntensityMap' and 'ReverseIntensityContext'; added warning that these classes will be renamed in future

# Related reading

https://trello.com/c/AN7eXyM3/19-rename-in-viewers-reverse-intensity

Is that everything or are there more classes/methods to be renamend @jburel ?

### DIFF
--- a/components/common/src/omeis/providers/re/codomain/ReverseIntensityContext.java
+++ b/components/common/src/omeis/providers/re/codomain/ReverseIntensityContext.java
@@ -18,7 +18,7 @@ package omeis.providers.re.codomain;
  * @version 2.2
  * @since OME2.2
  * 
- * @deprecated This class will be renamed to InvertIntensityContext (or similar) in
+ * @deprecated This class will be renamed to InverseIntensityContext (or similar) in
  *             future versions
  */
 public class ReverseIntensityContext extends CodomainMapContext {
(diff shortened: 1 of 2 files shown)


mtbc:
Could it make sense for them to vacuously extend correctly named ones to allow immediate code transition?

dominikl:
Was wondering about that, too. Deprecating without offering an alternative isn't very nice.

jburel:
You could introduce a ``InvertCodomainMap``

mtbc:
"Invert" or "inverse" or "inverted"?

dominikl:
I tried to rename the classes to 'Inverse...' but keep (and deprecate) the 'Reverse...' ones in #5436 . But for some reason that breaks the rendering settings (therefore excluded #5436 for now)

Edit: Closed the other PR again. Don't understand the Ice/Model stuff enough...

jburel:
@dominikl sorry I dropped of my radar. i will do that asap

jburel:
Comments added to the original PR https://github.com/openmicroscopy/openmicroscopy/pull/5436
Introducing Invert context is still doable
It will require a DB patch

jburel:
@dominikl will you re-activate the previous PR and close this one?

dominikl:
Ah yes, I'll try to reactive #5436 again with your suggested db-patch.

mtbc:
Maybe also `@Deprecated` annotations on the classes?

will-moore:
In the UI "Reverse Intensity" -> "Invert", so at API, ```ReverseIntensityMap``` -> ```InvertMap```?

mtbc:
Probably fine to leave `ReverseIntensityAction` as Insight is increasingly maintenance-only.

From my point of view, there are suggestions in comments but this PR is already good to merge.

jburel:
For insight only change the UI not the underline class name or method

dominikl:
Um, what!? - Can't really make sense of the last few comments, I'll just leave the PR as it is :-)

PR #5445, opened 2017-08-17
32bit minmax

dominikl:
# What this PR does

Adds a new method `findMinMax` to the RawPixelsStore, which gets the minimum and maximum pixel value of an image. This method is used ~~for certain pixel types (e.g. 32bit float)~~ when the `StatsInfo` is not available (e.g. for 32bit float images), to set the channel window start and end to reasonable values. It's called from the `computeLocationStats` method of the RenderingService, so that channel window start/end values are set on import of these kind of images. And for already imported images the clients can trigger it by clicking on the 'Imported' rendering setting button.

# Testing this PR

- Get the example file from https://www.openmicroscopy.org/qa2/qa/feedback/17841/
- Import it and check that a reasonable thumbnail is generated (not just a uniformly gray one) and the image looks ok with the default rendering settings.
- Log in as user-3, pick an image from dataset 'Pre-PR5445' (that's the same images but imported prior to this PR). Reset the rendering settings to 'Imported'. Check again that the proper thumbnail is generated and the image looks ok with these new rendering settings. **Note: Accidentaly uploaded the images to eel latest, so not possible to test that now**

# Related reading

https://trello.com/c/r4cINtil/105-32-bit-images-issue

### DIFF
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -337,7 +337,9 @@ module omero {
                 idempotent IntegerIntegerArrayMap getHistogram(IntegerArray channels, int binCount, bool globalRange, omero::romio::PlaneDef plane) throws ServerError;
                 
                 /**
-                 * Find the minimum and maximum pixel values for the specified channels. This method can currently only handle non-pyramid images,
+                 * Find the minimum and maximum pixel values for the specified channels by iterating over a full plane.
+                 * In case of multi-z/t images only the 'middle' plane with index maxZ/2 ,respectively maxT/2 is taken into account.
+                 * Note: This method can currently only handle non-pyramid images,
                  * in which case an empty map will be returned.
                  * @param channels the channels
                  * @return See above.
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -338,7 +338,7 @@ module omero {
                 
                 /**
                  * Find the minimum and maximum pixel values for the specified channels by iterating over a full plane.
-                 * In case of multi-z/t images only the 'middle' plane with index maxZ/2 ,respectively maxT/2 is taken into account.
+                 * In case of multi-z/t images only the 'middle' plane with index maxZ/2, respectively maxT/2 is taken into account.
                  * Note: This method can currently only handle non-pyramid images,
                  * in which case an empty map will be returned.
                  * @param channels the channels
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -337,6 +337,14 @@ module omero {
                 idempotent IntegerIntegerArrayMap getHistogram(IntegerArray channels, int binCount, bool globalRange, omero::romio::PlaneDef plane) throws ServerError;
                 
                 /**
+                 * Find the minimum and maximum pixel values for the specified channels. This method can currently only handle non-pyramid images,
+                 * in which case an empty map will be returned.
+                 * @param channels the channels
+                 * @return See above.
+                 **/
+                idempotent IntegerDoubleArrayMap findMinMax(IntegerArray channels) throws ServerError;
+                
+                /**
                  * Returns the byte width for the pixel store.
                  * @return See above.
                  **/
--- a/components/blitz/resources/omero/api/RawPixelsStore.ice
+++ b/components/blitz/resources/omero/api/RawPixelsStore.ice
@@ -339,8 +339,7 @@ module omero {
                 /**
                  * Find the minimum and maximum pixel values for the specified channels by iterating over a full plane.
                  * In case of multi-z/t images only the 'middle' plane with index maxZ/2, respectively maxT/2 is taken into account.
-                 * Note: This method can currently only handle non-pyramid images,
-                 * in which case an empty map will be returned.
+                 * Note: This method can currently only handle non-pyramid images, otherwise an empty map will be returned.
                  * @param channels the channels
                  * @return See above.
                  **/
--- a/components/server/src/ome/logic/RenderingSettingsImpl.java
+++ b/components/server/src/ome/logic/RenderingSettingsImpl.java
@@ -98,12 +98,6 @@ public class RenderingSettingsImpl extends AbstractLevel2Service implements
     /** Reference to the raw pixels store. */
     private RawPixelsStore rawPixStore;
     
-    /** Pixel types we can't use the global min/max as channel start/end window **/
-    private static final Set<String> minMaxPixelTypes = new HashSet<String>(
-            Arrays.<String> asList(new String[] { PixelsType.VALUE_FLOAT,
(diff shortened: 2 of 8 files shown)


components/blitz/resources/omero/api/RawPixelsStore.ice
```diff
@@ -337,6 +337,16 @@ module omero {
                 idempotent IntegerIntegerArrayMap getHistogram(IntegerArray channels, int binCount, bool globalRange, omero::romio::PlaneDef plane) throws ServerError;
                 
                 /**
+                 * Find the minimum and maximum pixel values for the specified channels by iterating over a full plane.
+                 * In case of multi-z/t images only the 'middle' plane with index maxZ/2 ,respectively maxT/2 is taken into account.
```
jburel: space in wrong location ``maxZ/2 ,respectively``

components/server/src/ome/logic/RenderingSettingsImpl.java
```diff
@@ -93,6 +95,15 @@
     /** Reference to the service used to retrieve the pixels metadata. */
     protected transient IPixels pixelsMetadata;
  
+    /** Reference to the raw pixels store. */
+    private RawPixelsStore rawPixStore;
+    
+    /** Pixel types we can't use the global min/max as channel start/end window **/
```
jburel: the same will be true for any pixels type if no stats have been calculated so maybe we want to support all pixel types.

components/server/src/ome/logic/RenderingSettingsImpl.java
```diff
@@ -917,17 +942,26 @@ private void computeLocationStats(Pixels pixels,
...
-        		min = qs.getPixelsTypeMin();
-        		max = qs.getPixelsTypeMax();
-        	}
-            cb.setInputStart(new Double(min));
-            cb.setInputEnd(new Double(max));
+            
+            if (realMinMax == null || realMinMax.isEmpty()) {
+                // use global min/max according to pixeltype
+                sf.computeLocationStats(pixels, buf, planeDef, w);
+                cb.setNoiseReduction(sf.isNoiseReduction());
```
jburel: noiseReduction can be ignored we no longer use it

components/server/src/ome/services/RawPixelsBean.java
```diff
@@ -745,6 +745,30 @@ public synchronized void setTimepoint(byte[] arg0, int arg1) {
         return result;
     }
 
+    @RolesAllowed("user")
+    public synchronized Map<Integer, double[]> findMinMax(int[] channels) {
```
jburel, line 749: Throw if the plane is too big?

components/blitz/resources/omero/api/RawPixelsStore.ice
```diff
@@ -337,6 +337,16 @@ module omero {
                 idempotent IntegerIntegerArrayMap getHistogram(IntegerArray channels, int binCount, bool globalRange, omero::romio::PlaneDef plane) throws ServerError;
                 
                 /**
+                 * Find the minimum and maximum pixel values for the specified channels by iterating over a full plane.
+                 * In case of multi-z/t images only the 'middle' plane with index maxZ/2, respectively maxT/2 is taken into account.
+                 * Note: This method can currently only handle non-pyramid images,
+                 * in which case an empty map will be returned.
```
dominikl: Just noticed that doesn't make sense. Of course in case of pyramid images an empty map is returned. Will fix that in another commit after PR testing today.

components/server/src/ome/logic/RenderingSettingsImpl.java
```diff
@@ -917,17 +938,25 @@ private void computeLocationStats(Pixels pixels,
...
+                    min = qs.getPixelsTypeMin();
+                    max = qs.getPixelsTypeMax();
+                }
+                cb.setInputStart(new Double(min));
+                cb.setInputEnd(new Double(max));
+            } else {
+                // use real min/max value of the pixels
+                cb.setInputStart(realMinMax.get(w)[0]);
+                cb.setInputEnd(realMinMax.get(w)[1]);
+                cb.setNoiseReduction(false);
```
dominikl: Will remove this line, too.

components/server/src/ome/logic/RenderingSettingsImpl.java
```diff
@@ -1159,6 +1188,17 @@ public void setPixelsData(PixelsService dataService) {
...
         pixelsData = dataService;
     }
+    
+    /**
+     * Sets injector. For use during configuration. Can only be called once.
+     * 
+     * @param rawPixStore
+     *            The value to set.
+     */
+    public void setRawPixStore(RawPixelsStore rawPixStore) {
```
mtbc: maybe call fully `rawPixelsStore` where API-visible

components/server/src/ome/logic/RenderingSettingsImpl.java
```diff
@@ -917,17 +938,25 @@ private void computeLocationStats(Pixels pixels,
...
+                sf.computeLocationStats(pixels, buf, planeDef, w);
+                min = sf.getInputStart();
+                max = sf.getInputEnd();
+                if (Math.abs(min - max) < EPSILON) {
+                    qs = quantumFactory.getStrategy(qDef, pixels);
+                    min = qs.getPixelsTypeMin();
+                    max = qs.getPixelsTypeMax();
+                }
+                cb.setInputStart(new Double(min));
+                cb.setInputEnd(new Double(max));
```
mtbc: good opportunity to get rid of this `new Double`

jburel:
pr_5445


Changes look good: left merge, right latest (no change)

dominikl:
Does web-dev-merge not connect to eel 4064? Uploaded 5 images yesterday in dataset 'Pre PR5445' to test the reset to 'Imported' on already imported images. Can't see the dataset when using OMERO.web.

jburel:
you imported them to latest
http://web-dev-latest.openmicroscopy.org/webclient/
I can see them, you might have set the wrong port in insight

dominikl:
Oh no, you're right, I uploaded them to 14064. So point 3 is now impossible to test. Could only exclude PR for Monday upload the images again and include the PR again for Tuesday.

jburel:
w/o the stats info, this seems to be an happy medium.
i.e. the image can be viewed. The controls still offer the full range e.g. sliders but I am no sure we can do  much w/o making several assumptions
On-going discussion on the card

jburel:
Clicking the ``imported`` button works in web and insight but it does not in iviewer (todo: check issue)
i.e. the changes are applied and the image is "visible" again
Min/max behaviour is different between and insight. Insight goes for full range, this is expected
web does not.
This is outside the scope of the PR

components/server/src/ome/logic/RenderingSettingsImpl.java
dominikl: Sorry, missed that comment. I'll add another small commit for the rename of the variable to 'rawPixelsStore'.

dominikl:
Sorry, missed Marks comment on Friday, so last commit is not in today's build, but it's a simple rename of a variable, code review 's probably sufficient.

jburel:
Comment about clients capture https://trello.com/c/F7zMjEbh/134-clients-no-stats-controls-check

PR #5447, opened 2017-08-18
Travis update (rebased onto metadata53)

jburel:
This is the same as gh-5409 but rebased onto metadata53.

----

# What this PR does

Use trusty to build 
This is necessary to build with newer version of pytables
and newer version of ice

 * building using ``ant`` fails all the time, need to use ``./build.py``.
 * installation of python dependencies needs to be explicitly done via
``sudo apt-get`` during ``before_install:`` otherwise the dependencies will not be available to the container. This means that the python tests will fail with for example ``No module Ice``. Change ``sudo`` option to ``required``. I did not notice an increase in the build time with that change.  See https://github.com/travis-ci/travis-ci/issues/8048
 * update ``numexpr`` via pip to avoid some table test failure  e.g.
``test/unit/tablestest/test_servants.py:349:`` will fail with ``ValueError: Using `oa_ndim == 0` when `op_axes` is NULL. Use `oa_ndim == -1` or the MultiNew iterator for NumPy <1.8 compatibility``

# Testing this PR
Check that travis is green

See https://trello.com/c/gVsUkLH8/100-travis-upgrade

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ omero_version.py
 eclipse.log
 velocity.log*
 .idea
+.vscode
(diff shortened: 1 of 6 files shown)


jburel:
--rebased-from #5409

jburel:
Closing something did not work when using scc

PR #5448, opened 2017-08-18
Travis trusty

jburel:
# What this PR does

Same as #5409 
Use trusty instead of precise

# Testing this PR
Check that travis is green

cc @joshmoore


jburel:
Actually this is not going to work
since the metadata53 does not have the web requirement file for 35
this implies that ``slice2py`` will be for Ice 3.6 and it will lead to an ice mistmatch

It was working previously b/c we are using our own ice 35 but the current setup in the travis file implies that 3 different versions of Ice are partially installed 3.4, 3.5 and 3.6 (we are lucky it was working!)

PR #5450, opened 2017-08-18
Rendering engine bulk settings update

chris-allan:
# What this PR does

The rendering engine was designed with stateful use in mind. Unfortunately, this means that many remote calls are required to establish initial state. When the rendering engine is used with stateless environments like OMERO.web these remote calls generate excessive overhead on each and every rendering operation.

The above is of particular importance when the number of channels is high (>10).

This PR allows the remote caller to update a series of settings by passing a single populated RenderingDef via one remote method invocation rather than making a large number of them to achieve the same result.

During testing several performance related improvements were also made to linear quantization. `Math.pow()` is very expensive, especially when `k` is fractional. Attempts have been made to avoid it entirely when `k=1` and other similar situations where `x` is being raised to the power of `1`.

# Testing this PR

1. required setup

One or more imported images.

2. actions to perform

Make method calls both singularly and with the bulk `updateSettings()` and compare the results.

Check that the integration tests are green
3. expected observations

Identical behaviour to current methods.

# Related reading

N/A

/cc @will-moore, @waxenegger, @dominikl

### DIFF
--- a/components/common/src/omeis/providers/re/RenderingEngine.java
+++ b/components/common/src/omeis/providers/re/RenderingEngine.java
@@ -11,6 +11,7 @@ import ome.api.StatefulServiceInterface;
 import ome.conditions.ValidationException;
 import ome.model.core.Pixels;
 import ome.model.display.QuantumDef;
+import ome.model.display.RenderingDef;
 import ome.model.enums.Family;
 import ome.model.enums.RenderingModel;
 import omeis.providers.re.codomain.CodomainMapContext;
@@ -506,6 +507,15 @@ public interface RenderingEngine extends StatefulServiceInterface {
      * @see #addCodomainMapToChannel(CodomainMapContext, int)
      */
     public void removeCodomainMapFromChannel(CodomainMapContext mapCtx, int w);
+
+    /**
+     * Updates the current rendering settings based on a provided rendering
+     * definition and associated sub-objects.
+     * @param settings Rendering definition to copy from. Each sub-object
+     * will be processed as though the specific method was called with
+     * related attributes provided as arguments.
+     */
+    public void updateSettings(RenderingDef settings);
 
     /** Saves the current rendering settings in the database. */
     public void saveCurrentSettings();
--- a/components/common/src/omeis/providers/re/RenderingEngine.java
+++ b/components/common/src/omeis/providers/re/RenderingEngine.java
@@ -529,9 +529,10 @@ public interface RenderingEngine extends StatefulServiceInterface {
      * If one or more attributes that apply to a particular method are
      * <code>null</code> it will be <b>skipped</b> in its entirety. The
      * underlying Renderer is not able to handle partial field updates.
-     * Furthermore, {@link ChannelBinding}s that are <code>null</code> and
-     * indexes in the {@link RenderingDef#WAVERENDERING} array greater than the
-     * currently looked up {@link Pixels#SIZEC} will be skipped.
+     * Furthermore, {@link ome.model.display.ChannelBinding} references that are
+     * <code>null</code> and indexes in the {@link RenderingDef#WAVERENDERING}
+     * array greater than the currently looked up {@link Pixels#SIZEC} will be
+     * skipped.
      */
     public void updateSettings(RenderingDef settings);
 
--- a/components/common/src/omeis/providers/re/RenderingEngine.java
+++ b/components/common/src/omeis/providers/re/RenderingEngine.java
@@ -527,10 +527,11 @@ public interface RenderingEngine extends StatefulServiceInterface {
      * <li>{@link RenderingEngine#setChannelLookupTable(int, String)}</li>
      * </ul>
      * If one or more attributes that apply to a particular method are
-     * <code>null</code> it will be <b>skipped</b>. The underlying Renderer is
-     * not able to handle partial field updates.
-     *
-     * <b>NOTE:</b> Codomain mappings are ignored.
+     * <code>null</code> it will be <b>skipped</b> in its entirety. The
+     * underlying Renderer is not able to handle partial field updates.
+     * Furthermore, {@link ChannelBinding}s that are <code>null</code> and
+     * indexes in the {@link RenderingDef#WAVERENDERING} array greater than the
+     * currently looked up {@link Pixels#SIZEC} will be skipped.
      */
     public void updateSettings(RenderingDef settings);
 
--- a/components/common/src/omeis/providers/re/RenderingEngine.java
+++ b/components/common/src/omeis/providers/re/RenderingEngine.java
@@ -513,8 +513,24 @@ public interface RenderingEngine extends StatefulServiceInterface {
      * definition and associated sub-objects.
      * @param settings Rendering definition to copy from. Each sub-object
      * will be processed as though the specific method was called with
-     * related attributes provided as arguments. <b>NOTE:</b> Codomain
-     * mappings are ignored.
+     * related attributes provided as arguments. The following methods are
+     * called underneath: <ul>
+     * <li>{@link RenderingEngine#setModel(RenderingModel)}</li>
+     * <li>{@link RenderingEngine#setDefaultZ(int)}</li>
+     * <li>{@link RenderingEngine#setDefaultT(int)}</li>
+     * <li>{@link RenderingEngine#setQuantumStrategy(int)}</li>
+     * <li>{@link RenderingEngine#setCodomainInterval(int, int)}</li>
+     * <li>{@link RenderingEngine#setActive(int, boolean)}</li>
+     * <li>{@link RenderingEngine#setChannelWindow(int, double, double)}</li>
+     * <li>{@link RenderingEngine#setQuantizationMap(int, Family, double, boolean)}</li>
+     * <li>{@link RenderingEngine#setRGBA(int, int, int, int, int)}</li>
+     * <li>{@link RenderingEngine#setChannelLookupTable(int, String)}</li>
+     * </ul>
+     * If one or more attributes that apply to a particular method are
+     * <code>null</code> it will be <b>skipped</b>. The underlying Renderer is
+     * not able to handle partial field updates.
+     *
+     * <b>NOTE:</b> Codomain mappings are ignored.
      */
     public void updateSettings(RenderingDef settings);
 
--- a/components/common/src/omeis/providers/re/RenderingEngine.java
+++ b/components/common/src/omeis/providers/re/RenderingEngine.java
@@ -513,7 +513,8 @@ public interface RenderingEngine extends StatefulServiceInterface {
      * definition and associated sub-objects.
      * @param settings Rendering definition to copy from. Each sub-object
      * will be processed as though the specific method was called with
-     * related attributes provided as arguments.
+     * related attributes provided as arguments. <b>NOTE:</b> Codomain
+     * mappings are ignored.
      */
     public void updateSettings(RenderingDef settings);
 
--- a/components/server/src/ome/services/RenderingBean.java
+++ b/components/server/src/ome/services/RenderingBean.java
@@ -1216,7 +1216,7 @@ public class RenderingBean implements RenderingEngine, Serializable {
     @RolesAllowed("user")
     public void setChannelLookupTable(int w, String lookup) {
         StopWatch t0 = new Slf4JStopWatch(
-                "RenderingBean.setChannelLookupTable");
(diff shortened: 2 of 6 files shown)


jburel:
@dominikl insight has a similar method so we could easily replace the code by this method.
the only difference is that insight has  a boolean flag to update or not the z and t

components/server/src/ome/services/RenderingBean.java
```diff
@@ -774,6 +776,77 @@ public long saveAsNewSettings() {
...
+            // Emulate setChannelWindow(int, double, double)
+            Double start = cb.getInputStart();
+            Double end = cb.getInputEnd();
+            if (start != null && end != null) {
+                setChannelWindow(w, start, end);
+            }
+            // Emulate setQuantizationMap(int, Family, double, boolean)
+            Family family = cb.getFamily();
+            Double coefficient = cb.getCoefficient();
+            Boolean noiseReduction = cb.getNoiseReduction();
```
jburel, line 842: noise reduction is no longer used when we do the mapping of the pixels intensity
So having it null or not null will not really matter

components/server/src/ome/services/RenderingBean.java
```diff
@@ -774,6 +776,77 @@ public long saveAsNewSettings() {
 
     /**
      * Implemented as specified by the {@link RenderingEngine} interface.
+     *
+     * @see RenderingEngine#updateSettings()
```
mtbc: Javadoc reports,

> can't find updateSettings() in omeis.providers.re.RenderingEngine

jburel:
Added missing integration tests  to https://github.com/chris-allan/openmicroscopy/pull/10

jburel:
One point to discuss:
The method allows to copy rendering settings between images with incompatible pixels type.
This is not what is currently in place when applying settings across images see https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/logic/RenderingSettingsImpl.java#L1107

This should be synchronized to avoid major confusion since we have 2 sections of the code doing the same thing 

Either we relax the check when applying the settings or we introduce a check in the RE

cc @joshmoore

will-moore:
@jburel I'm not sure this will be confused with copying settings between images. This is more similar to ```conn.setActiveChannels(self, channels, windows=None, colors=None, reverseMaps=None)``` (which should be updated to use this new bulk settings).

Probably if I take the rendering settings from a 4-channel image and try to save them to a 1-channel image, I should get an Exception, just as ```re.setActive(3, True)``` will give me an exception.

jburel:
To clarify:
The method when applying the settings using the ``RenderingSettingsService`` only works if the number of channels matches e.g. I cannot copy the settings from a 2-channels image to a 3-channels image
But using the method introduced in the RE I can do that.
This is the part that we need to agree on in my opinion

chris-allan:
As @will-moore has correctly surmised the target of this method is absolutely not copying settings between images although it could be used to do so. It is designed specifically with the optimisation of `conn.setActiveChannels()`, etc. in mind and aiding viewer implementations.

In fact, I would not even suggest that a "real" rendering settings object with an `id` be used with this method in the first place. Since a `RenderingDef` has all the relevant properties and sub-objects already implemented it was used for convenience.

jburel:
I understand the initial idea but as soon as we expose it, any rnd settings could be specified. The 2 methods mentioned are very similar, actually they are the same excluding the check.

The actual implementation i.e. allowing settings from images with less channels is probably what we should use when applying settings in the ``RenderingSettingService``(i remember having a conversation with @will-moore a while back about it)
To offer even more flexibility in the RE, you could argue that we could use in the RE the channels_number =min (number_of_channels_from_rnd_def, number_of_channels_of_re)

chris-allan:
I disagree. They are certainly similar but their use cases, implementation, prototypes, and associated service have completely different implications. The rendering settings service requires a source and destination, and saves the result. `updateSettings()` does not save, the rendering settings can come from anywhere, and are specified explicitly.

If you want to make restrictions to `updateSettings()` that's fine, and logical, but those restrictions should be informed by what people want to do with the rendering engine and preventing them from shooting themselves in the foot, not by what the rendering settings service does or does not do.

jburel:
The method does not the save, that's true but ``updateSettings`` then ``saveSettings`` will give a different result from ``applySettings``. What is used a basis is still the ``rndSettings``, the destination here being the pixels id and its associated settings.
The logical bit for me, as mentioned previously, will be to relax the very strict check in ``applySettings``
Actually the methods of the renderingSettings service do not offer the flexibility required
i.e. it is all settings or nothing. What is actual needed is partial copy/paste. (not in this PR)

jburel:
@chris-allan: I have opened a PR adding support for codomain ctx and also added extra null
Tested have been added too

See https://github.com/chris-allan/openmicroscopy/pull/11

chris-allan:
@jburel: Go that included here now as well. There is no need to create additional overhead for yourself by opening a PR or create new branches every time as I'm just fast-forward merging your commits into my branch and pushing. Ping me here when you want some commits included and I'll happily bring them in.

jburel:
@chris-allan will do but I don't think we need more for now
Possible other changes e.g. insight using new method should be done in another PR

chris-allan:
I would argue that `updateSettings()` is **always** going to give you a different result than `saveSettings()`. Hence my position. They're really not the same and should not have similar semantics. `updateSettings()` is inherently optimised to **only** contain the minimal amount of rendering settting information required to get the desired result from `renderCompressed()` or `renderAsPackedInt()`.

This goal has potentially been blurred a bit with the latest changes to the null checking in 3c4ab0f691b47fa9e4d47aa2ee3bbb30ef8d66ec. I'll comment on that separately.

jburel:
We should probably, in that case, throw an exception indicating the minimum requirement for the method if not met
e.g. ``model, boolean cdmin/cdmax`` did not have any ``null`` check.

components/server/src/ome/services/RenderingBean.java
```diff
@@ -774,6 +776,90 @@ public long saveAsNewSettings() {
...
+    @RolesAllowed("user")
+    public void updateSettings(RenderingDef settings) {
+        if (settings == null) {
+            return;
+        }
+
+        // Emulate setModel(RenderingModel)
+        RenderingModel model = settings.getModel();
+        if (model != null) {
+            setModel(model);
```
chris-allan, line 791: It's probably not a semantic that we have ever explicitly defined and I don't know what would happen but technically `null` can be passed through to `RenderingBean#setModel(RenderingModel)` now. We should be sure that this is still adhering to those semantics and that people aren't calling `setModel(null)` and expecting a specific behaviour.

components/server/src/ome/services/RenderingBean.java
```diff
@@ -774,6 +776,90 @@ public long saveAsNewSettings() {
...
+        if (quantumDef != null) {
+            // Emulate setQuantumStrategy(int)
+            Integer bitResolution = quantumDef.getBitResolution();
+            if (bitResolution != null) {
+                setQuantumStrategy(bitResolution);
+            }
+            // Emulate setCodmainInterval(int, int)
+            Integer start = quantumDef.getCdStart();
+            Integer end = quantumDef.getCdEnd();
+            if (start != null && end != null) {
```
chris-allan, line 813: I didn't have a concrete use case for this so didn't do it. It's not really possible now and I don't know under what conditions you would pass a `QuantumDef` instance without these filled out. Any ideas on what the behaviour should be? Should we really allow just setting the bit resolution or a `QuantumDef` with no fields populated?

components/server/src/ome/services/RenderingBean.java
```diff
@@ -774,6 +776,90 @@ public long saveAsNewSettings() {
...
+                setCodomainInterval(start, end);
+            }
+        }
+        for (int w = 0; w < settings.sizeOfWaveRendering(); w++) {
+            ChannelBinding cb = settings.getChannelBinding(w);
+            // Emulate setActive(boolean)
+            Boolean active = cb.getActive();
+            if (active != null) {
+                setActive(w, active);
```
chris-allan: I semi-deliberately set this as minimum requirement and didn't have a use case for it to be left out. Any further ideas on what the behaviour should be? Do we really want to allow the "activeness" or a channel to be unspecified?

chris-allan:
Perfectly reasonable approach. I deliberately did not do that because of the inherent semantic differences between the `RenderingDef` and the actual underlying `RenderingBean` concrete method implementations. The methods are inherently pretty fine grained right now so I'll take your lead there.

Since these just throw `ServerError` and we probably want a "nice" error message I'm not 100% on the best way to do that from the `server` component and have the desired `blitz` mapping. @mtbc and/or @joshmoore can probably help there. I experimented with this a bit and just ended up with the rather useless `omero.ApiUsageException: null` on the client side.

Also, someone at some point has gone through the rather painstaking work of moving some Javadoc into `RenderingEngine.ice`. Is all Javadoc allowed there? If so where does that output end up? Can we use `{@link ...}`, etc. with impunity in the Slice definitions?

/cc @sbesson

jburel:
We have some usage of ``{@link }`` in some slices e.g. ``IConfig.ice``, the problem we had with ice 3.6.2 was related to quote so we had to introduce numerous ``""`` which was not nice

chris-allan:
Had a review today with @jburel and we are going to establish the precedent of `null` dictating everywhere that the implementation should skip the assignment of that particular setting. In addition, we will support `null` values within the `ChannelBinding` array signifying that all the setting assignment for that particular channel should be skipped. The array processing itself will be bounded to the number of channels in the currently looked up image This will allow the channel binding array size to exceed `sizeC` where indexes `> sizeC` will be skipped.

chris-allan:
Changes made in 55d079c24e44766dcd449dd3e2044ef463bc4565 as agreed above with @jburel.

jburel:
@chris-allan 3 commits on top of this branch to test the latest changes
branch https://github.com/jburel/openmicroscopy/tree/chris/bulk-re-update-develop

chris-allan:
Pulled those ones in @jburel.

components/common/src/omeis/providers/re/RenderingEngine.java
```diff
@@ -507,6 +508,33 @@ public void setQuantizationMap(int w, Family family, double coefficient,
...
+     * <li>{@link RenderingEngine#setActive(int, boolean)}</li>
+     * <li>{@link RenderingEngine#setChannelWindow(int, double, double)}</li>
+     * <li>{@link RenderingEngine#setQuantizationMap(int, Family, double, boolean)}</li>
+     * <li>{@link RenderingEngine#setRGBA(int, int, int, int, int)}</li>
+     * <li>{@link RenderingEngine#setChannelLookupTable(int, String)}</li>
+     * </ul>
+     * If one or more attributes that apply to a particular method are
+     * <code>null</code> it will be <b>skipped</b> in its entirety. The
+     * underlying Renderer is not able to handle partial field updates.
+     * Furthermore, {@link ChannelBinding}s that are <code>null</code> and
```
mtbc:
```
Tag @link: reference not found: ChannelBinding
```
Probably just needs package name prepending.

jburel:
I have just noticed that the rendering engine is never closed in all the tests
When this is merged, I will open a new PR fixing the tests including the ones added in this PR

joshmoore:
> I have just noticed that the rendering engine is never closed in all the tests

@jburel : perhaps try to first make the test fail on TearDown due to the rendering engine not being closed? That may be something to add to the base class.

joshmoore:
I pushed a very minor typo fix (in case you want to blow it away). Otherwise, the API (incl. null discussion) makes sense to me. My only minor wish is a prefix for the Slf4JStopWatch. I think `omero.` is used in other related locations. Integration tests are currently green. Happy for this to get in.

jburel:
We could unify that in the followup PR, i.e. closing RE if you are ok with that @joshmoore

chris-allan:
Have a look at e593fdc67815f7a388742a3100c9091b2fac825b and see if that's more or less what you were thinking, @joshmoore.

jburel:
there is no mention of the ``bean`` in the ThumbnailBean for example
usual it is ``omero.methodName``

chris-allan:
I'm happy to strip the `rendering_engine` if it's a problem. I just didn't want to have collisions with other places where we might be doing rendering engine timings.

jburel:
I don't think it is a problem, it is clearer the way you have changed it in my opinion

joshmoore:
No objections to the longer value. Can add `omero.a_b` to the other uses over time.

PR #4995, opened 2016-12-12
Web marshal RGBA int

dominikl:
# What this PR does

With this PR color integers will be interpreted as RGBA (instead of ARGB).

# Testing this PR

Set a specific fill and line color for an ROI in Insight. Check that the color displayed is the same for Web and Insight. ~Ignore the opacity for now.~

# Related reading
- [Trello - Shape colour read as RGBA](https://trello.com/c/YLP2lElr/248-shape-colour-read-as-rgba)
- Insight PR: #4948

~Apparently Web does not take the alpha value into account; it is read correctly, but the color is always shown as full opaque. Is that expected @will-moore ?~

PR #5451, opened 2017-08-18
Fixed rgb_int2css and rgb_int2rgba

dominikl:
Both methods calculated alpha as fraction of 256 whereas
it should have been 255 ( alpha int value [0, 255] )

(was introduced by #4995 , thanks for spotting this issue!)

/cc @jburel @chris-allan

### DIFF
--- a/components/tools/OmeroWeb/test/unit/test_marshal.py
+++ b/components/tools/OmeroWeb/test/unit/test_marshal.py
@@ -41,7 +41,7 @@ def basic_line(default_id):
     shape.y1 = rdouble(1.0)
     shape.x2 = rdouble(2.0)
     shape.y2 = rdouble(3.0)
-    shape.strokeColor = rint(ctypes.c_int(0x112233FF).value) # r=17,g=34,b=51,a=255
+    shape.strokeColor = rint(ctypes.c_int(0x112233FF).value)
     return shape
 
 
@@ -64,7 +64,7 @@ def basic_polyline(request, default_id):
     shape = omero.model.PolylineI()
     shape.id = rlong(default_id)
     shape.points = rstring(points)
-    shape.strokeColor = rint(ctypes.c_int(0x11223300).value) # r=17,g=34,b=51,a=0
+    shape.strokeColor = rint(ctypes.c_int(0x11223300).value)
     return shape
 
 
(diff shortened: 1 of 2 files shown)


components/tools/OmeroWeb/test/unit/test_marshal.py
```diff
@@ -40,6 +40,7 @@ def basic_line(default_id):
     shape.y1 = rdouble(1.0)
     shape.x2 = rdouble(2.0)
     shape.y2 = rdouble(3.0)
+    shape.strokeColor = rlong(287454207)     # 0x112233FF,r=17,g=34,b=51,a=255
```
chris-allan: This value is a 32-bit integer (`rint`) not an `rlong`.

components/tools/OmeroWeb/test/unit/test_marshal.py
```diff
@@ -62,6 +63,7 @@ def basic_polyline(request, default_id):
     shape = omero.model.PolylineI()
     shape.id = rlong(default_id)
     shape.points = rstring(points)
+    shape.strokeColor = rlong(287453952)     # 0x11223300,r=17,g=34,b=51,a=0
```
chris-allan: This value is a 32-bit integer (`rint`) not an `rlong`.

components/tools/OmeroWeb/test/unit/test_marshal.py
```diff
@@ -186,15 +188,25 @@ def test_ellipse_marshal(self, basic_ellipse):
         assert 1.0 == marshaled['radiusX']
         assert 0.5 == marshaled['radiusY']
 
-    def test_rgba(self, basic_ellipse):
+    def test_rgba(self, basic_ellipse, basic_line, basic_polyline):
```
chris-allan: Probably a good idea to include some values < 0. This has really tripped us up in the past. These are instances where the red component is > 127 (`0x7f`). Can be easily achieved by using the `ctypes` module:

```
In [1]: import ctypes

In [2]: ctypes.c_int(0x800000ff).value
Out[2]: -2147483393
```

/cc @will-moore

chris-allan (changes requested):

chris-allan (approved):

mtbc:
Need to run flake8 on `test_marshal.py`.

components/tools/OmeroWeb/test/unit/test_marshal.py
```diff
@@ -41,7 +41,7 @@ def basic_line(default_id):
     shape.y1 = rdouble(1.0)
     shape.x2 = rdouble(2.0)
     shape.y2 = rdouble(3.0)
-    shape.strokeColor = rint(ctypes.c_int(0x112233FF).value) # r=17,g=34,b=51,a=255
+    shape.strokeColor = rint(ctypes.c_int(0x112233FF).value)
```
jburel: still useful to have
so maybe
```
# r=17,g=34,b=51,a=255
shape.strokeColor = rint(ctypes.c_int(0x112233FF).value)
```

jburel:
Thanks @dominikl

PR #5452, opened 2017-08-18
Cleanup thirdparty libs

rgozim:
### What this PR does

Removed old or unused libraries from OmeroWeb


jburel:
Any specific areas to look at?

will-moore:
Removal of the SpinButton from the zoom field has changed the layout slightly:

screen shot 2017-08-23 at 15 09 44

The functionality of the up/down buttons in the zoom field (in chrome) is less predictable than the SpinButton too. Sometimes the handling of change doesn't happen immediately or until you move the mouse away? 
I don't imagine that there's an easy fix for this (see a similar thing for the Well "images per row" input).
This is pretty minor functionality (don't use this in the Preview viewer) so I guess it's a question of "How much do we want to remove the SpinButton plugin"?

jburel:
I will keep it in until we can "replace" the viewer with iviewer
it won't affect anything at the license level

jburel:
@rgozim could you roll back the ``spinbutton`` change?

jburel:
spinner is back 
Thanks 
Merging

PR #5436, opened 2017-08-14
Rename and deprecated ReverseIntensityMap/Context

dominikl:
# What this PR does

Deprecates 'ReverseIntensityMap' and ReverseIntensityContext' and adds the equivalents 'InverseIntensityMap'  and 'InverseIntensityContext'.
Replaces #5435 

# Testing this PR

# Related reading
https://trello.com/c/AN7eXyM3/19-rename-in-viewers-reverse-intensity

--exclude  **!! Does not work !!**

For some reason this breaks imports and rendering settings. Any ideas why @mtbc / @jburel ?
E. g. import:
```
java.lang.RuntimeException: Failure response on import!
Category: ::omero::grid::ImportRequest
Name: import-request-failure
Parameters: {stacktrace=java.lang.RuntimeException: omero.ApiUsageException
[...]

PR #5459, opened 2017-08-21
Deprecate ReverseintensityMap/Context

dominikl:
# What this PR does

Deprecates ReverseIntensityMap/Context and introduces InverseIntensityMap/Context, so that the former can be removed later.

Couldn't reopen #5436 any longer.

# Related reading

https://trello.com/c/AN7eXyM3/19-rename-in-viewers-reverse-intensity

--exclude I don't think this works properly yet


dominikl:
I guess at some later point a db upgrade script is needed, which copies the previous values from `reverseintensitycontext` table to the new `inverseintensitycontext` table and then deletes the former, right?

mtbc:
That's a very large diff, I wonder if something went wrong with the `git mv` stuff at https://docs.openmicroscopy.org/contributing/schema-changes.html#updating-the-schema-and-the-sql-scripts or if we need to further document some interim commit step, cc: @joshmoore.

mtbc:
I'm happy to help with the later upgrade / table merging script when the time comes.

dominikl:
Ah, I might have missed `git mv`... let me try again.

dominikl:
I'm not sure any longer, that this is a good strategy. With deprecating ReverseIntensityContext and adding InverseIntensityContext, we have to support both through the full stack and in the clients, otherwise one client could use InverseIntensityContext while another client still uses ReverseIntensityContext and show different things. Maybe just deprecating now, and simply rename later, would really be better. How did you handle cases like that in the past? @mtbc @jburel

jburel:
Introduced only at the DB level? otherwise we won't be able to integrate it before next major release
We have do that in the past for LUT. No support in clients/server only entry in DB

dominikl:
I just can't get my head around it... the DB patch will now create a new table `inverseintensitycontext`, which means there are now two tables and two objects, reverse and inverse, and clients could theoretically use both. That's not a good idea. But just creating the a new table `inverseintensitycontext` without adding it to server/blitz doesn't make sense either. 
Ideally the `reverseintensitycontext` table should be renamed to `inverseintensitycontext` and then create a legacy `ReverseIntensityContext` class which maps to the `inverseintensitycontext` table. Is that possible?

joshmoore:
> Is that possible?

Perhaps, but we've not done it before.

dominikl:
I should just leave it and reopen  #5435 again (simply just add deprecate comments).

jburel:
Let's discuss that today
there is a limited number of places where it is used so handling both for 5.4 should not be too bad

dominikl:
Only add deprecation note for now  #5435

PR #5461, opened 2017-08-21
remove_ns_from_conn.createOriginalFile

will-moore:
# What this PR does

Removes the ```ns``` (namespace) argument from ```conn.createOriginalFileFromFileObj()``` and ```conn. createOriginalFileFromLocalFile()``` where it is ignored - probably got left over from a previous refactor.

# Testing this PR

1. Can be tested by saving an OMERO.figure since with this current PR https://github.com/ome/omero-figure/pull/239 figure now uses the ```conn.createOriginalFileFromFileObj()```.

# Related reading

Noticed this when looking into https://trello.com/c/4unMpwmq/128-createoriginalfilefromfileobj and finding that the method I wanted was already there!

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3956,9 +3956,9 @@ class _BlitzGateway (object):
             origFilePathAndName = localPath
         path, name = os.path.split(origFilePathAndName)
         fileSize = os.path.getsize(localPath)
-        with open(localPath, 'rb') as fileHandle
-            return self.createOriginalFileFromFileObj(
-                fileHandle, path, name, fileSize, mimetype)
+        with open(localPath, 'rb') as fileHandle:
+            return self.createOriginalFileFromFileObj(fileHandle, path, name,
+                                                      fileSize, mimetype)
 
     def createFileAnnfromLocalFile(self, localPath, origFilePathAndName=None,
                                    mimetype=None, ns=None, desc=None):
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -3862,7 +3862,7 @@ class _BlitzGateway (object):
         return {'imageCount': len(imageIds), 'updateCount': updateCount}
 
     def createOriginalFileFromFileObj(self, fo, path, name, fileSize,
-                                      mimetype=None):
+                                      mimetype=None, ns=None):
         """
         Creates a :class:`OriginalFileWrapper` from a local file.
         File is uploaded to create an omero.model.OriginalFileI.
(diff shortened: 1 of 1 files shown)


mtbc:
```
./components/tools/OmeroPy/src/omero/gateway/__init__.py:3949:61: F821 undefined name 'ns'
```

joshmoore:
Can you follow up on what "probably leftover" means in practical terms? What kind of notification were users given? Is this going into the upgrade notes as a breaking change? Etc.

will-moore:
I don't know the full history, but I think that 
```conn.createFileAnnfromLocalFile(localPath, origFilePathAndName=None, mimetype=None, ns=None, desc=None)```
originally contained all the code for
```createOriginalFileFromLocalFile(localPath, origFilePathAndName=None, mimetype=None, ns=None)()```
and then the latter method got split out, or there was some copy/paste error from one to the other.
I could try to inspect the history in more detail (tried already a bit) but it's tricky because of lots of ```flake8``` refactoring etc.

This doesn't remove any functionality (not breaking API from that point of view) but it's possible someone was passing a value into the ```ns``` parameter (which was being ignored) and now, trying to do this would cause an Exception, so it should certainly be included in the release notes.

However, the fix is so trivial (remove the ```ns``` parameter) and this is a major release so I think this is OK to include in 5.4.0

**Comment by jburel** [comment]
Since it is possible that someone was passing the parameter ``ns``, we should deprecate the method
and introduce a new one w/o ``ns``.
There is no reason to have potential users of the method reacting negatively on the lists, forum etc.

**Comment by will-moore** [comment]
We don't need to deprecate the method.
We can simply issue a deprecation warning if someone passes the ```ns``` parameter.
Or we could just add a ```**kwargs``` in place of the ```ns``` parameter, which would mean that any additional parameters could be passed without breaking.

**Comment by jburel** [comment]
Better to go for the non-breaking/no exception option
and add a warning indicating that ``ns`` was not used if specified and can now be ignored

**Comment by joshmoore** [comment]
NB: `ns=` argument was originally copied over in https://github.com/openmicroscopy/openmicroscopy/pull/133/files#diff-e4feecea30b2ee097639cbb7c9171fedR2732

**Comment by will-moore** [comment]
How long do we need to leave warnings before we can remove deprecated methods?
The problem I see with adding a warning and leaving the method as 
```
conn.createOriginalFileFromFileObj(self, fileHandle, path, name, fileSize, mimetype, ns)
```
is that this is more likely to make users USE ```ns``` instead of remove it since the warning is elsewhere in the code and easily missed.
I wonder if any of our users ever see any of the warnings we add?
Seems you don't see them by default in normal usage.
Probably most users first know that something has changed when it breaks (or possibly in the release notes if they read them).
So maybe best solution is just to announce it in the release notes and remove it for 5.4.0.

**Comment by will-moore** [comment]
Added back the ```ns=None``` parameter with deprecation warnings.
But, would still like to discuss usage of warnings and when we can finally remove ```ns```.

**Comment by jburel** [comment]
Thanks for adding it back

File: components/tools/OmeroPy/src/omero/gateway/__init__.py [hunk 1]
```diff
@@ -3938,17 +3944,22 @@ def createOriginalFileFromLocalFile(self, localPath,
...
+            warnings.warn(
+                "Deprecated in 5.4.0. The ns parameter was added in error"
+                " and has always been ignored",
+                DeprecationWarning)
         if origFilePathAndName is None:
             origFilePathAndName = localPath
         path, name = os.path.split(origFilePathAndName)
         fileSize = os.path.getsize(localPath)
         fileHandle = open(localPath, 'rb')
         try:
```
jburel: since you are modifying that code you may also want to use ``with open`` instead of the ``try/finally``

jburel: @will-moore will you adjust the code in this PR?

will-moore: Fixed.

jburel:
Tested with OMERO.figure
Works as expected.
Thanks

PR #5463, opened 2017-08-22
Base template script injection

will-moore:
# What this PR does

Allows an html template to be included in the base ```core_html``` template that is used for all webclient pages. See https://trello.com/c/mDE65WQK/132-inject-scripts-ga-etc
This can be used for adding ```<script>```, ```<link/>``` etc into the ```<head>``` of all pages.

# Testing this PR

Create an html file e.g.  ```base_include.html``` and save it to e.g. ```omero_web_templates``` dir.
```
<script>
alert("Hello World");
</script>
```
Then:
```
$ bin/omero config append omero.web.template_dirs '"/Users/wmoore/Desktop/omero_web_templates"'
$ bin/omero config set omero.web.base_include_template 'base_include.html'
```
Restart web....

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -636,7 +636,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
         ["BASE_INCLUDE_TEMPLATE",
          None,
          identity,
-         ("Template to be included in every page, at the bottom of the <body>")],
+         ("Template to be included in every page, at the end of the <body>")],
     "omero.web.login_redirect":
         ["LOGIN_REDIRECT",
          '{}',
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -1061,6 +1061,7 @@ TEMPLATES = [
                 'django.template.context_processors.tz',
                 'django.contrib.messages.context_processors.messages',
                 'omeroweb.custom_context_processor.url_suffix',
+                'omeroweb.custom_context_processor.base_include_template',
             ],
         },
     },
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -632,6 +632,11 @@ CUSTOM_SETTINGS_MAPPINGS = {
          ("Define template used as an index page ``http://your_host/omero/``."
           "If None user is automatically redirected to the login page."
           "For example use 'webclient/index.html'. ")],
+    "omero.web.base_include_template":
+        ["BASE_INCLUDE_TEMPLATE",
+         None,
+         identity,
+         ("Template to be included in the header of every page.")],
     "omero.web.login_redirect":
         ["LOGIN_REDIRECT",
          '{}',
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -636,7 +636,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
         ["BASE_INCLUDE_TEMPLATE",
          None,
          identity,
-         ("Template to be included in the header of every page.")],
+         ("Template to be included in every page, at the bottom of the <body>")],
     "omero.web.login_redirect":
         ["LOGIN_REDIRECT",
          '{}',
--- a/components/tools/OmeroWeb/omeroweb/custom_context_processor.py
+++ b/components/tools/OmeroWeb/omeroweb/custom_context_processor.py
@@ -8,4 +8,4 @@ def url_suffix(request):
 
 
 def base_include_template(request):
-	return {'base_include_template': settings.BASE_INCLUDE_TEMPLATE}
+    return {'base_include_template': settings.BASE_INCLUDE_TEMPLATE}
--- a/components/tools/OmeroWeb/omeroweb/custom_context_processor.py
+++ b/components/tools/OmeroWeb/omeroweb/custom_context_processor.py
@@ -1,6 +1,11 @@
 from omero_version import omero_version
+from django.conf import settings
 
 
 def url_suffix(request):
     suffix = u"?_%s" % omero_version
     return {'url_suffix': suffix}
+
+
+def base_include_template(request):
+	return {'base_include_template': settings.BASE_INCLUDE_TEMPLATE}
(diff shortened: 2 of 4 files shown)


jburel:
```
./components/tools/OmeroWeb/omeroweb/custom_context_processor.py:11:1: E101 indentation contains mixed spaces and tabs
./components/tools/OmeroWeb/omeroweb/custom_context_processor.py:11:1: W191 indentation contains tabs
```

manics:
Would it be better to include `_header_` somewhere in the property name, in case we want to add other includes in future e.g. a footer?

will-moore:
@manics I can't see that we'd need to add another "include" within a specific place within the html since   the base template is extended in so many different ways that there's no consistent place.
However, it might make sense to move the include block to just before the ```</body>``` tag so that all the DOM is loaded before the script.  This would also allow the include to specify a DOM element if it needed to add one into the page.

joshmoore:
Perhaps these are related to @manics' questions but:

 * what other templates would there be to need to configure an entire dir?
 * assuming only one `base_include_template` is possible, can it further include other templates?

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -632,6 +632,11 @@ def leave_none_unset_int(s):
          ("Define template used as an index page ``http://your_host/omero/``."
           "If None user is automatically redirected to the login page."
           "For example use 'webclient/index.html'. ")],
+    "omero.web.base_include_template":
+        ["BASE_INCLUDE_TEMPLATE",
+         None,
+         identity,
+         ("Template to be included in the header of every page.")],
```
manics: Change this comment since it's now included at the end of the page?

will-moore:
Nested templates works fine:

/Users/wmoore/Desktop/omero_web_templates/base_include.html
```
<script>
	alert("Hello {% include 'nested_template.html' %} ")
</script>
```

/Users/wmoore/Desktop/omero_web_templates/nested_template.html 
```
 Nested template message!
```
screen shot 2017-08-23 at 21 29 27

The ```omero.web.template_dirs``` setting already exists and is used for a bunch of other template settings.

joshmoore:
A few more things:

 * "Nested templates works fine:" :+1:
 * "omero.web.template_dirs setting already exists" :+1:
   - do you have a pointer to a current usage?
   - what's the default value if none is set?
 * is there any existing (if less well named) way of doing what you're doing here? can `extra_js` be set externally?
 * can you explain why the property is "base_include" but it's in "core_html"?
 * based on @manics comment https://github.com/openmicroscopy/openmicroscopy/pull/5463#issuecomment-324569768, is there anything we should do regarding updating our templating configuration?

joshmoore:
re: travis status

```
./components/tools/OmeroWeb/omeroweb/settings.py:639:80: E501 line too long (81 > 79 characters)

```

manics:
`omero.web.template_dirs` is the core django property `TEMPLATE_DIRS`: https://docs.djangoproject.com/en/1.8/ref/settings/#template-dirs

will-moore:
Most of the templates that we extend are called ```base...```. It just so happens that the very lowest one is named ```core```. 
"Base" implies it's going to be extended.

joshmoore:
Testing on the breaking queue.

jburel:
What was the outcome of your tests? @joshmoore

joshmoore:
Apologies. The configuration of a `<script>...</script>` element in a `base_template.html` which is included on the CI servers definitely allows injecting into every page. I tried to convince myself that there was another backwards-compatible way to do this but failed, so this looks to just be a necessary extension point.

Follow-ups could include:
 * docs (which I think are in progress)
 * longer-term clean up of the hierarchy
 * cleanup of the template configuration due to changes in Django since 1.8
 * perhaps some good example uses like `OMERO.analytics`

Thanks, @will-moore

will-moore:
Looking at the hierarchy of templates from webgateway, webclient, webadmin and feedback apps:

Probably the biggest "bug" I see here is that the ```batch_annotate.html``` and ```annotations_share.html``` (which extend the ```webgateway/core_html.html```) are loaded into the right panel of webclient, so you could get a Google analytics hit by loading right panel instead of when loading a new page, which is probably undesirable.

```
- webgateway/base_site.html
- webgateway/core_html.html
	- webgateway/base_frame.html
	- webclient/annotations/annotations_share.html 		*Right panel share    NB: metadata_general DOES NOT extend core_html.html
	- webclient/annotations/batch_annotate.html 		*Right panel
	- webclient/annotations/download_placeholder.html 	Dialog for original files download
	- webclient/scripts/no_processor.html
	- webclient/scripts/script_ui.html
	- webgateway/viewport/omero_image.html
	- webadmin/forgotten_password.html
	- webclient/login.html
		- webclient/index.html 							Example template
	- feedback/base_error.html
		- feedback/404.html
		- feedback/500.html
		- feedback/comment.html
	- feedback/error.html
	- feedback/thanks.html
	- webgateway/base/base_header.html
		- webclient/scripts/base_custom_dialog.html
			- webclient/scripts/make_movie.html
			- webclient/scripts/split_view_figure.html
			- webclient/scripts/thumbnail_figure.html
		- webgateway/base/base_main.html
			- webgateway/base/container2.html
				- webclient/base/base.html
					- webclient/index/index.html 		Not used
			- webgateway/base/container3.html
				- webclient/base/base_container.html 	Main webclient home page
					- webclient/data/containers.html
					- webclient/history/history.html
					- webclient/public/public.html
					- webclient/search/search.html
			- webadmin/base.html
				- webadmin/admin_base.html
					- webamdin/email.html
					- webadmin/experimenters.html
					- webadmin/groups.html
					- webadmin/statistics.html
				- webadmin/avatar.html
				- webadmin/experimenter_form.html
				- webadmin/group_form.html
				- webadmin/group_form_owner.html
				- webadmin/myaccount.html
				- webadmin/noemail.html
```

jburel:
Do you want to record that somewhere?
It will probably require some attention

will-moore:
Yes, but I'm not sure I want to put it in our public docs since then it becomes our "published API" and any changes are "breaking".
I'll put it on a card...
I could fix the extending of ```core_html.html``` by the right panel content for 5.4.0?

jburel:
I meant on a card not in the doc
How long will it take to fix the issue with ``core_html`` (PR freeze date is passed) but it is a bug!

will-moore:
Card created: https://trello.com/c/BQyJ0Ay7/47-django-templates-cleanup

PR #5460, opened 2017-08-21
allow model types to be ignored during object deletion

mtbc:
# What this PR does

This PR adds a `Delete2.typesToIgnore` property that can be used to stall graph traversal at nodes of those types. This is a strictly *experimental* feature that when misused has the potential to leave the model graph in an unfortunate state or to produce strange error messages. It is however a useful interim feature for IDR workflows until we upgrade Hibernate and adjust its query caching.

# Testing this PR

Set graphs logging to DEBUG using https://docs.openmicroscopy.org/omero/5.4.0-m3/developers/Server/ObjectGraphs.html#changing-the-log-level. (Alternatively, @manics has means to instrument PostgreSQL queries.)

Link a map annotation to multiple images. Using a Python script delete one of the links via `Delete2(..., typesToIgnore=['Annotation'])`. Observe that with the extra `typesToIgnore` a lot less processing than otherwise happens with the annotation's *other* links.

Also see https://github.com/openmicroscopy/openmicroscopy/pull/5460#issuecomment-324338672.

Integration tests at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.delete/AnnotationDeleteTest/.

# Related reading

https://trello.com/c/mfW0N1an/89-poor-performance-when-deleting-many-one-annotation-links

PR #5464, opened 2017-08-22
allow model types to be ignored during object deletion (rebased onto metadata53)

manics:
This is the same as gh-5460 but rebased onto metadata53.

----

# What this PR does

This PR adds a `Delete2.typesToIgnore` property that can be used to stall graph traversal at nodes of those types. This is a strictly *experimental* feature that when misused has the potential to leave the model graph in an unfortunate state or to produce strange error messages. It is however a useful interim feature for IDR workflows until we upgrade Hibernate and adjust its query caching.

# Testing this PR

Set graphs logging to DEBUG using https://docs.openmicroscopy.org/omero/5.4.0-m3/developers/Server/ObjectGraphs.html#changing-the-log-level. (Alternatively, @manics has means to instrument PostgreSQL queries.)

Link a map annotation to multiple images. Using a Python script delete one of the links via `Delete2(..., typesToIgnore=['Annotation'])`. Observe that with the extra `typesToIgnore` a lot less processing than otherwise happens with the annotation's *other* links.

# Related reading

https://trello.com/c/mfW0N1an/89-poor-performance-when-deleting-many-one-annotation-links

### DIFF
--- a/components/blitz/resources/omero/cmd/Graphs.ice
+++ b/components/blitz/resources/omero/cmd/Graphs.ice
@@ -202,6 +202,11 @@ module omero {
          * Delete model objects.
          **/
         class Delete2 extends GraphModify2 {
+                /**
+                 * Ignore in the operation all objects of these types.
+                 **/
+                ["deprecated:experimental: may be wholly removed in next major version"]
+                omero::api::StringSet typesToIgnore;
         };
 
         /**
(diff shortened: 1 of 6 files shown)


manics:
--rebased-from #5460

joshmoore:
This has been merged to the mainline. I've cherry-picked the tests that Mark added there, but since this is already deployed as part of a merge build for prod43, I'm merging now for tagging.

joshmoore:
Cherry-picking https://github.com/openmicroscopy/openmicroscopy/pull/5464/commits/a2e1683443c6dcd38c6fe352340418ebff216c88 was a mistake.

PR #5465, opened 2017-08-23
Shape stats minimal

waxenegger:
With clients other than insight potentially wanting to query some basic stats regarding intensity under rois there was a need to turn to the already deprecated RoiService in order to not reinvent the wheel.

There was a good deal there but some things needed fixing/changig:
- A new rois service method was created: 'getShapeStatsRestricted'
- the original sql in the old getShapeStats referenced a column that was not there.
- not all SmartShapes implemented the method 'public void areaPoints(PointCallback cb)' needed for point iteration/stats accumulation but threw an error instead.
- transforms were not integrated in the calculation
- the old point iteration requested the same row of data a countless number of times and looped inefficiently
- the old method allowed querying of big, tiled source which it could not cope with, the new one disallows it for now (perhaps tiled iteration is an option later)
- the old method used the entire z/t range if shapes were not linked to a particular z/t, the new one forces 'fallback' z/t as parameters (the idea is that the z/t that one is looking at is used instead of going through all z*t combinations.
- the new method restricts to using only rois on the same image in one call and groups rois by common z/t/c before looping so that that common combination is not requested from the reader more than once.

TEST: check that the integration test (testStatsBasicUsageAndInputChecks) in https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/RoiServiceTest/ passes.

### DIFF
--- a/components/blitz/src/ome/services/roi/GeomTool.java
+++ b/components/blitz/src/ome/services/roi/GeomTool.java
@@ -381,7 +381,10 @@ public class GeomTool {
                "join fetch r.image i join fetch i.pixels p " +
                "where s.id in (:ids)").
            setParameterList("ids", shapeIds).list();
-       
+       if (results.size() != shapeIds.size()) {
+           throw new ApiUsageException("Given hape id(s) invalid");
+       }
+
        ome.model.core.Image image = null;
        ome.model.core.Pixels pixels = null;
        for (final Object r : results) {
@@ -411,7 +414,6 @@ public class GeomTool {
            }
            lookupValue.add(shape);
        }
-       if (zt_lookup.size() == 0) return null;
 
        // any point iteration in a tiled image is a lost cause
        if (data.needsPyramid(pixels)) 
@@ -420,8 +422,11 @@ public class GeomTool {
       Set<Integer> validChannels = null;
        if (channels != null && channels.length > 0) {
            validChannels = new HashSet<Integer>(channels.length);
-           for (int ch : channels)
-               if (ch >= 0 && ch < pixels.getSizeC()) validChannels.add(ch);
+           for (int ch : channels) {
+               if (ch < 0 || ch >= pixels.getSizeC())
+                   throw new ApiUsageException("Given channel(s) out of bounds.");
+               validChannels.add(ch);
+           }
        }
        // common info for all shapes 
        final long pixelId = pixels.getId();
--- a/components/blitz/src/ome/services/roi/GeomTool.java
+++ b/components/blitz/src/ome/services/roi/GeomTool.java
@@ -382,7 +382,7 @@ public class GeomTool {
                "where s.id in (:ids)").
            setParameterList("ids", shapeIds).list();
        if (results.size() != shapeIds.size()) {
-           throw new ApiUsageException("Given hape id(s) invalid");
+           throw new ApiUsageException("Given shape id(s) invalid");
        }
 
        ome.model.core.Image image = null;
--- a/components/blitz/src/ome/services/roi/GeomTool.java
+++ b/components/blitz/src/ome/services/roi/GeomTool.java
@@ -32,6 +32,7 @@ import ome.util.SqlAction;
 import omero.api.RoiStats;
 import omero.api.ShapePoints;
 import omero.api.ShapeStats;
+import omero.model.AffineTransformI;
 import omero.model.Ellipse;
 import omero.model.Line;
 import omero.model.Point;
@@ -359,11 +360,10 @@ public class GeomTool {
 
     public ShapeStats [] getStatsRestricted(
             List<Long> shapeIds, 
-            int z_for_unattached, int t_for_unattached,
+            int zForUnattached, int tForUnattached,
             int[] channels) {
-        if (shapeIds == null || shapeIds.isEmpty()) {
-            return null;
-        }
+        if (shapeIds == null || shapeIds.isEmpty()) 
+            throw new ApiUsageException("Provide a non empty list of shape ids.");
 
         final Session session = factory.getSession();
         final List<ShapeStats> shapeStats = 
@@ -376,25 +376,33 @@ public class GeomTool {
        // fetch shapes from db and perform some basic checks
        List results = 
            session.createQuery(
-               "select distinct s from Shape s " + 
-               "join fetch s.roi r join fetch r.image i " + 
-               "join fetch i.pixels p join fetch p.channels c " + 
-               "join fetch c.logicalChannel lc where s.id in (:ids)").
+               "select distinct s from Shape s " +
+               "left join fetch s.transform t left join fetch s.roi r " +
+               "join fetch r.image i join fetch i.pixels p " +
+               "where s.id in (:ids)").
            setParameterList("ids", shapeIds).list();
        
        ome.model.core.Image image = null;
+       ome.model.core.Pixels pixels = null;
        for (final Object r : results) {
            final ome.model.roi.Shape shape = (ome.model.roi.Shape) r;
            final ome.model.roi.Roi roi = shape.getRoi();
            final ome.model.core.Image img = roi.getImage();
            
-           // check if all shapes come fro the same image
-           if (image == null) image = img;
-           else if (image.getId() != img.getId())
-               throw new ApiUsageException("all shapes have to be from the same image");
+           // check if all shapes come from the same image
+           if (image == null) {
+               image = img;
+               pixels = image.getPrimaryPixels();
+           } else if (image.getId() != img.getId())
+               throw new ApiUsageException("All shapes have to be from the same image");
+           // check if z/t unattached fallback values are not out of bounds
+           if (zForUnattached < 0 || zForUnattached >= pixels.getSizeZ() ||
+               tForUnattached < 0 || tForUnattached >= pixels.getSizeT())
+               throw new ApiUsageException(
+                   "Fallback value(s) for unattached z/t shapes are out of bounds");
            // if we have unattached z/t we use the unattached z,t fallback
-           final int theZ = shape.getTheZ() != null ? shape.getTheZ() : z_for_unattached;
-           final int theT = shape.getTheT() != null ? shape.getTheT() : t_for_unattached;
+           final int theZ = shape.getTheZ() != null ? shape.getTheZ() : zForUnattached;
+           final int theT = shape.getTheT() != null ? shape.getTheT() : tForUnattached;
            String lookupKey = theZ + "/" + theT;
            List<ome.model.roi.Shape> lookupValue = zt_lookup.get(lookupKey);
            if (lookupValue == null) {
@@ -405,13 +413,9 @@ public class GeomTool {
        }
        if (zt_lookup.size() == 0) return null;
 
-       // common info for all shapes 
-       final ome.model.core.Pixels pixels = image.getPrimaryPixels();
        // any point iteration in a tiled image is a lost cause
        if (data.needsPyramid(pixels)) 
            throw new ApiUsageException("This method can not handle tiled images yet.");
-       final long pixelId = pixels.getId();
-       final int sizeX = pixels.getSizeX();
        // check for channels filter
       Set<Integer> validChannels = null;
        if (channels != null && channels.length > 0) {
@@ -419,6 +423,10 @@ public class GeomTool {
            for (int ch : channels)
                if (ch >= 0 && ch < pixels.getSizeC()) validChannels.add(ch);
        }
+       // common info for all shapes 
+       final long pixelId = pixels.getId();
+       final int sizeX = pixels.getSizeX();
+       final int sizeY = pixels.getSizeY();
        // loop over shapes (grouped by z/t planes)
        for (final String key : zt_lookup.keySet()) {
            final String [] keyTokens = key.split("/");
@@ -445,6 +453,8 @@ public class GeomTool {
                        smartShape.areaPoints(
                            new SmartShape.PointCallback() {
                                public void handle(int x, int y) {
+                                   // we won't use pixels outside of the image 
+                                   if (x < 0 || y < 0 || x >= sizeX || y >= sizeY) return;
                                    stats.pointsCount[w]++;
                                    double value = pd.getPixelValue(sizeX * y + x);
                                    stats.min[w] = Math.min(value, stats.min[w]);
@@ -577,11 +587,17 @@ public class GeomTool {
                 IObject iobj = (IObject) source;
                 omero.model.IObject robj = (omero.model.IObject) o;
                 robj.setId(rlong(iobj.getId()));
-                robj.unload();
+                // this seems to me the least unintrusive way to remedy
+                // the copying of the affine tranform
+                // the other being the .combined files
+                // which as is just cast for the transform 
+                // (effectively losing info)
+                if (robj instanceof AffineTransformI)
+                    ((AffineTransformI) robj).copyObject(source, this);
+                else robj.unload();
             }
 
             return source;
-
         }
 
     }
--- a/components/blitz/src/ome/services/roi/GeomTool.java
+++ b/components/blitz/src/ome/services/roi/GeomTool.java
@@ -412,13 +412,16 @@ public class GeomTool {
        if (data.requiresPixelsPyramid(pixels)) 
            throw new ApiUsageException("This method cannot handle tiled images yet.");
        // check for channels filter
-       Set<Integer> validChannels = null;
+       Set<Integer> validChannels = new HashSet<Integer>();
        if (channels != null && channels.length > 0) {
-           validChannels = new HashSet<Integer>(channels.length);
            for (int ch : channels) {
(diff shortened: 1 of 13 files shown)


components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+       for (final Object r : results) {
+           final ome.model.roi.Shape shape = (ome.model.roi.Shape) r;
+           final ome.model.roi.Roi roi = shape.getRoi();
+           final ome.model.core.Image img = roi.getImage();
+           
+           // check if all shapes come from the same image
+           if (image == null) {
+               image = img;
+               pixels = image.getPrimaryPixels();
+           } else if (image.getId() != img.getId())
```
mtbc: use `equals` to compare `Long` IDs

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+               zt_lookup.put(lookupKey, lookupValue);
+           }
+           lookupValue.add(shape);
+       }
+
+       // any point iteration in a tiled image is a lost cause
+       if (data.needsPyramid(pixels)) 
+           throw new ApiUsageException("This method can not handle tiled images yet.");
+       // check for channels filter
+      Set<Integer> validChannels = null;
```
mtbc: indent

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+               validChannels.add(ch);
+           }
+       }
+       // common info for all shapes 
+       final long pixelId = pixels.getId();
+       final int sizeX = pixels.getSizeX();
+       final int sizeY = pixels.getSizeY();
+       // loop over shapes (grouped by z/t planes)
+       for (final String key : zt_lookup.keySet()) {
+           final String [] keyTokens = key.split("/");
```
mtbc: no space before `[]`

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+               final ShapeStats stats = makeStats(size_stats);
+               stats.shapeId = shape.getId();
+               final double[] sumOfSquares = new double[size_stats];
+
+               final PixelBuffer buf = data.getBuffer(pixelId);
+               try {
+            	   int i = 0;
+                   for (int c = 0; c < pixels.getSizeC(); c++) {
+                       if (validChannels != null && 
+                           !validChannels.contains(new Integer(c))) continue;
```
mtbc: shouldn't need this `new Integer`?

components/blitz/src/ome/services/roi/PixelData.java
```diff
@@ -60,4 +61,20 @@ public double get(PixelBuffer buf, int x, int y, int z, int c, int t) {
         }
     }
 
+    public  ome.util.PixelData getPlane(PixelBuffer buf, int z, int c, int t) {
+        try {
+            return  buf.getPlane(z, c, t);
```
mtbc: double space after `return`

components/blitz/src/ome/services/roi/PixelData.java
```diff
@@ -60,4 +61,20 @@ public double get(PixelBuffer buf, int x, int y, int z, int c, int t) {
...
+        } catch (IOException e) {
+            throw new ResourceError("IOException: " + e);
+        } catch (DimensionsOutOfBoundsException e) {
+            throw new ApiUsageException("DimensionsOutOfBounds: " + e);
+        } catch (IndexOutOfBoundsException iobe) {
+            throw new ValidationException("IndexOutOfBounds: " + iobe);
+        }
+    }
+
+    public boolean needsPyramid(Pixels pix) {
```
mtbc: Maybe match the wrapped `requiresPixelsPyramid` method name?

components/blitz/src/omero/model/SmartPointI.java
```diff
@@ -18,7 +20,17 @@
...
+            double point_x = x.getValue();
+            double point_y = y.getValue();
+            if (transform != null) {
+                final AffineTransform t = Util.getAwtTransform(transform);
+                
+                final Point2D p = 
+                    t.transform(new Point2D.Double(point_x, point_y), null);
+                point_x = p.getX();
+                point_y = p.getY();
+        	 }
```
mtbc: reduce indent

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+           if (image == null) {
+               image = img;
+               pixels = image.getPrimaryPixels();
+           } else if (!image.getId().equals(img.getId()))
+               throw new ApiUsageException("All shapes have to be from the same image");
+           // check if z/t unattached fallback values are not out of bounds
+           if (zForUnattached < 0 || zForUnattached >= pixels.getSizeZ() ||
+               tForUnattached < 0 || tForUnattached >= pixels.getSizeT())
+               throw new ApiUsageException(
+                   "Fallback value(s) for unattached z/t shapes are out of bounds");
```
jburel, line 406: should we go for something less aggressive and set the value to ``0 if < 0``
or to ``pixels.getSizeZ/T-1`` if >=pixels.getSizeZ/T``
since we support ``null`` array of channels?

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+            int[] channels) {
+        if (shapeIds == null || shapeIds.isEmpty()) 
+            throw new ApiUsageException("Provide a non empty list of shape ids.");
+
+        final Session session = factory.getSession();
+        final List<ShapeStats> shapeStats = 
+            new ArrayList<ShapeStats>(shapeIds.size());
+        // we lump together shapes that belong to the same z/t
+        // in order to not have to read the same z/t twice
+        final Map<String, List<ome.model.roi.Shape>> zt_lookup =
```
jburel: guava multimap could be used in this case

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+           if (lookupValue == null) {
+               lookupValue = new ArrayList<ome.model.roi.Shape>();
+               zt_lookup.put(lookupKey, lookupValue);
+           }
+           lookupValue.add(shape);
+       }
+
+       // any point iteration in a tiled image is a lost cause
+       if (data.requiresPixelsPyramid(pixels)) 
+           throw new ApiUsageException("This method can not handle tiled images yet.");
```
jburel: cannot?

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+       // check for channels filter
+       Set<Integer> validChannels = null;
+       if (channels != null && channels.length > 0) {
+           validChannels = new HashSet<Integer>(channels.length);
+           for (int ch : channels) {
+               if (ch < 0 || ch >= pixels.getSizeC())
+                   throw new ApiUsageException("Given channel(s) out of bounds.");
+               validChannels.add(ch);
+           }
```
jburel, line 424: if channels is ``null`` it might be easier to easier to always create ``validChannels``
and iterate through that list
so we can avoid check in 2 locations below i.e.
```
validChannels != null ? validChannels.size() : pixels.getSizeC();
```
and
```
if (validChannels != null &&  !validChannels.contains(c)) continue;
```

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+           for (ome.model.roi.Shape shape : zt_lookup.get(key)) {
+               final SmartShape smartShape = (SmartShape) new ShapeMapper().map(shape);
+               final int size_stats = 
+                   validChannels != null ? validChannels.size() : pixels.getSizeC();
+               final ShapeStats stats = makeStats(size_stats);
+               stats.shapeId = shape.getId();
+               final double[] sumOfSquares = new double[size_stats];
+
+               final PixelBuffer buf = data.getBuffer(pixelId);
+               try {
```
jburel: try with resources could be used to simplify the code

components/blitz/src/ome/services/roi/GeomTool.java
```diff
@@ -354,6 +358,143 @@ public void handle(int x, int y) {
...
+                   }
+               } finally {
+                   try {
+                       buf.close();
+                   } catch (IOException e) {
+                       log.error("Error closing " + buf, e);
+                   }
+               }
+               for (int w = 0; w < size_stats; w++) {
+                   stats.mean[w] = stats.sum[w] / stats.pointsCount[w];
```
jburel: ``stats.pointsCount[w]`` be ``0``?

components/blitz/src/ome/services/roi/PixelData.java
```diff
@@ -60,4 +61,20 @@ public double get(PixelBuffer buf, int x, int y, int z, int c, int t) {
         }
     }
 
+    public  ome.util.PixelData getPlane(PixelBuffer buf, int z, int c, int t) {
```
jburel: Add doc

components/blitz/src/ome/services/roi/PixelData.java
```diff
@@ -60,4 +61,20 @@ public double get(PixelBuffer buf, int x, int y, int z, int c, int t) {
...
+        } catch (IOException e) {
+            throw new ResourceError("IOException: " + e);
+        } catch (DimensionsOutOfBoundsException e) {
+            throw new ApiUsageException("DimensionsOutOfBounds: " + e);
+        } catch (IndexOutOfBoundsException iobe) {
+            throw new ValidationException("IndexOutOfBounds: " + iobe);
+        }
+    }
+
+    public boolean requiresPixelsPyramid(Pixels pix) {
```
jburel: Javadoc

components/blitz/src/omero/model/SmartPolylineI.java
```diff
@@ -10,15 +10,34 @@
...
 public class SmartPolylineI extends omero.model.PolylineI implements SmartShape {
 
     public void areaPoints(PointCallback cb) {
-        throw new UnsupportedOperationException();
+        Shape s = asAwtShape();
+        if (s == null) {
+            return;
+        }
+        final PathIterator it = s.getPathIterator(Util.getAwtTransform(transform));
+        double [] vals = new double[] {0,0,0,0,0,0};
```
jburel: space after ``,``

components/blitz/src/omero/model/SmartShape.java
```diff
@@ -214,7 +218,50 @@ public static void pointsByBoundingBox(Shape s, Rectangle2D r,
...
+        }
+        
+        public static Set<Point2D> getQuantizedLinePoints(Line2D line, Set<Point2D> points) {
+            if (line == null) return null;
+            
+            Point2D start = line.getP1();
+            Point2D end = line.getP2();
+            Point2D m = new Point2D.Double(
+                end.getX()-start.getX(), end.getY()-start.getY());
+            double lengthM = (Math.sqrt(m.getX()*m.getX()+m.getY()*m.getY()));
```
jburel, line 253: if ``start == end == (0, 0) then lengthM == 0``

components/tools/OmeroJava/test/integration/RoiServiceTest.java
```diff
@@ -518,4 +538,213 @@ private Roi createRoi(Image img, int z, int t) throws ServerError {
...
+            Arrays.asList(new Long[] {new Long (-1)}), new Integer(0), new Integer(0),
+            new int[] {}, "erroneous id" });
+        inputs.add(new Object[] { ids, new Integer(10), new Integer(10), new int[] {},
+            "wrong fallback z" });
+        inputs.add(new Object[] { ids, new Integer(0), new Integer(-10), new int[] {},
+            "wrong fallback t" });
+        inputs.add(new Object[] { ids, new Integer(0), new Integer(0), new int[] {200},
+            "wrong channel" });
+
+        // all these inputs should trigger an ApiUssageException
```
jburel: extra ``s``

components/blitz/src/ome/services/roi/GeomTool.java
waxenegger: Is this a dependency already? Otherwise I'd ask myself what we are gaining given. Performance wise nothing I's suspect since the number of entries are low for a hashmap anyhow.

Update: I have seen already that it's used in numerous places. Adjusting the code should be easy in that case.

components/blitz/src/ome/services/roi/GeomTool.java
waxenegger: @jburel sorry I miss the context, how would a channel null impact here. the z/tForUnattached are meant to be fallback values if a roi is unattached so that we don't iterate over, say, 100 * 60 * plane_width * plane_height x channels) pixels. the channels come in as a list and have been checked for bounds as well. are you asking if we want to allow unattached (at the cost of potentially doing all the pixel scanning over the entire t/z range)?

components/blitz/src/ome/services/roi/GeomTool.java
waxenegger: The first check is a bounds check and targets the given input parameter which is a chance to limit the channels to only a few. The second check is whether the specific channel is in that (now bound-checked) list.
I don't see how I can do away with it if I want the caller to restrict the iteration to just a few channels. At a minimum I have to check if it is empty when it iterates over all channels if I created an empty list. That's much of a muchness.
Maybe you are saying that the default behavior for an empty input of channels is not ok because in that case all channels are used. If I changed that then we could loop over all channels in the list after bound checking. But that makes the channel input list mandatory from a caller's perspective.
Or did you want me to fill the list with the channel numbers in the case of no channel input/restrictions handed in? But even there the number of operations is pretty much similar.

components/blitz/src/ome/services/roi/GeomTool.java
waxenegger: unfortunately PixelBuffer's close can throw an IOException and that doesn't fare well with the try with resources paradigm. I'm afraid I'll need to catch this one since adding a throws to the method declaration to break the workflow in case of such an exception is not an option either...

components/blitz/src/omero/model/SmartShape.java
waxenegger: I have addressed that degenerate case of a line being a point really with my latest commit. not sure why github doesn't mark it as outdated...

components/blitz/src/ome/services/roi/GeomTool.java
jburel: I am suggesting to always populate a list of channels index
if int[] channels is not null then we use the specified channels and keep the check
otherwise populate the ``validchannels`` list using value c = 0 to  pixels.getSizeC()-1;

You then iterate through ``validchannels`` for the stats

waxenegger: ok, then I read your comment the right way I believe. the very last commit should do that.

components/blitz/src/ome/services/roi/GeomTool.java
jburel: ``int[] channels`` can be ``null`` and in that case stats for all channels are calculated
We now starts to have images with a large number of channels
i was suggesting that if the value of z is out of bounds then z ==0 or z =max-1 (same for t)
to follow the logic for channels
or we say that "unattached" channels (i.e. null) is not supported

components/blitz/src/ome/services/roi/GeomTool.java
jburel: The comment above depends on what we decide to do for "unattached" channels

components/blitz/src/ome/services/roi/GeomTool.java
jburel: we already have a guava dependency but this is not a blocker just a suggestion

components/blitz/src/ome/services/roi/GeomTool.java
waxenegger: I think we need to discuss this. I still see 2 things here, the channel number and the unattached z/t.

If we are concerned about the default behavior for missing channel input (it being filled up with all available channels) and images with a big number of channels we should make that parameter mandatory I believe. Then the caller has to list the channels they need.

As for the fallback of unattached we can default to either 0 or max if the supplied value makes no sense, or anything in between for that matter, if throwing an exception is too radical.

jburel:
the stats added to this PR are the stats currently in insight to help the migration from insight to web viewer

PR #5468, opened 2017-08-24
Disable UI when no sudo/upload permission

dominikl:
# What this PR does

Disables the specific UI elements if a light admin lacks the upload scripts and sudo permissions.

# Testing this PR

- Light admin (in group with other users but **not** group owner) without 'sudo' permission: Make sure you don't see the 'Import For' box.
- Same light admin but with 'sudo' permission: Now you should see the 'Import For' element.
- Normal user but group owner (of group with other users): Now you should also see the 'Import For' element.
screen shot 2017-08-24 at 15 11 02

- Light admin without 'Upload scripts' permission: You shouldn't see the upload scripts button.
- Light admin with 'Upload scripts' permission: Now you should see the upload scripts button.
screen shot 2017-08-24 at 14 51 31

# Related reading

https://trello.com/c/4dvUYSjS/126-definition-of-ultimate-admin

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/TreeViewerAgent.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/TreeViewerAgent.java
@@ -147,7 +147,7 @@ public class TreeViewerAgent
 	}
 	
     /**
-     * Returns <code>true</code> if the currently logged in user is is allowed
+     * Returns <code>true</code> if the currently logged in user is allowed
      * to move objects to/from groups, <code>false</code> otherwise.
      * 
      * @return See above.
@@ -161,7 +161,7 @@ public class TreeViewerAgent
     }
 
     /**
-     * Returns <code>true</code> if the currently logged in user is is allowed
+     * Returns <code>true</code> if the currently logged in user is allowed
      * to edit groups, <code>false</code> otherwise.
      * 
      * @return See above.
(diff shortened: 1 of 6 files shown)


jburel:
Do you have lightadmin already setup on eel?

dominikl:
I always just used a local server for testing and switched between Web (setting the light admin flags) and Insight. I think @pwalczysko has set up light admins an eel with all possible flavors. But there are so many users now, it's not easy to find them.

pwalczysko:
@dominikl @jburel : the light admins are easily found by going to webadmin and then filtering the users for the string "light"

jburel:
Tested import with:
* full-light: The option is available as expected
* light02Aug3: The option is not available as expected

components/insight/SRC/org/openmicroscopy/shoola/agents/treeviewer/TreeViewerAgent.java
```diff
@@ -202,6 +202,20 @@ public static boolean isEditUser() {
         return isAdministrator() && b.booleanValue();
     }
     
+    /**
+     * Returns <code>true</code> if the currently logged in user is is allowed
```
jburel: double ``is is``

jburel:
Upload button is not available for full-light despite having the ability to upload script

dominikl:
Which user was that? Just checked with 'light-can-all' and upload script button is there.

jburel:
I used ``full-light``

jburel:
I found the source of the problem
I first logged in as user-1 who does not have the permission then used "switch user"
and logged in as full-light
In that case the option is not available

The switch user is mainly a dev feature so I will leave it as it is

jburel:
small typo in the doc otherwise looks good

PR #5470, opened 2017-08-24
Reverse invert webclient

will-moore:
# What this PR does

Renames "reverse" to "inverted" in the url ```map``` query string parameter.
```
?maps=[{"inverted":{"enabled":false}},{"inverted":{"enabled":true}}]
```
Also adds "inverted" to json for the webgateway/imgData/ json, but kept the "reverseIntensity" boolean too, so that ```iviewer``` etc aren't broken. 
```
{ "reverseIntensity": true
  "inverted": true,
```
same for ```conn. getAllRenderingDefs()``` ( -> webgateway/get_image_rdefs_json/<id>).

# Testing this PR
This affects all rendering settings functionality in webclient.

1. Choose "Invert" channel
1. Test copy, paste of settings between images
1. Pick rendering settings from owner's thumbnails
1. Undo, redo etc.
1. Open Full viewer from Preview panel - url should have ```?maps=[{"inverted":{"enabled":false}},{"inverted":{"enabled":true}}]``` and image should respect same invert settings as in Preview panel.
1. Copy image link in Full viewer. Should have "inverted" in url. Pasting into new window,"inverted" should be respected.
1. iviewer should behave as normal
1. /webgateway/imgData/<id> should show "inverted" flag in channels as well as "reverseIntensity" (duplicated)


jburel:
cc @waxenegger

waxenegger:
I don't think we are doing ourselves any favors by letting visual representations/renames propagate to query strings and data formats that other tools rely on. In the case of reverse intensity this is the second such change already. Not sure this is a good strategy to maintain.

mtbc:
The alternative is having an ongoing situation wherein developers must discover and remember that the code and UI name the same things differently, making it more confusing to get to grips with. I confuse Petr sometimes by using the "under the hood" terminology when he is thinking in terms of the UI he sees. Certainly agreed there's a cost there though: we should try hard to name things well in the first place. https://xkcd.com/910/ :smiley:

jburel:
The term has been around for a decade :-) and use in the UI for a short time a while ago
We have not been able to change the backend
If we keep reverse in the url but we will have to keep it for a long time and probably confuse some users too since the controls  to manipulate rendering settings

rgozim:
Followed testing instructions and everything is working as expected.

waxenegger:
@mtbc I hear you and I understand your reasoning. But for any web app using the resource, this is effectively like breaking interfaces. Not a big deal in this case maybe (only that many internal apps to change) but it's a good decision that the web api uses versioning.

ah, I just noticed, I failed to see that the old calling/json is still supported so never mind my ramblings...

joshmoore:
@waxenegger : barring:

 * https://github.com/openmicroscopy/openmicroscopy/pull/5470/files#diff-52d80c8ae2043af9119b666a1fb1998bR606 and
 * https://github.com/openmicroscopy/openmicroscopy/pull/5470/files#diff-e4feecea30b2ee097639cbb7c9171fedR8368

would you agree that this is backwards compatible? And if so, would that alleviate your concerns?

waxenegger:
@joshmoore yes, I failed to see initially that the old json/query string was still in the code, not sure what I was looking at but ok, my bad...

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8360,7 +8365,7 @@ def getZoomLevelScaling(self):
         return rv
 
     def setActiveChannels(self, channels, windows=None, colors=None,
-                          reverseMaps=None):
+                          invertMaps=None):
```
joshmoore: This change counts as a breaking change. Since the rest of this PR attempts to be backwards compatible, it makes sense to do so here as well. `if reverseMaps is not None` could issue a warning so that it's effectively deprecated.

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.viewport.js
```diff
@@ -603,9 +603,9 @@ jQuery._WeblitzViewport = function (container, server, options) {
     this.setChannelActive(idx, !_this.loadedImg.channels[idx].active);
   };
 
-  this.setChannelReverseIntensity = function (idx, reverse, noreload) {
```
joshmoore: Removal of this function is another breaking change. I assume the cost of having it as an alias is low enough that it would be worth it.

```
this. setChannelReverseIntensity = this. setChannelInverted
```

joshmoore (changes requested):
A few remaining breakages here.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8384,17 +8389,22 @@ def setActiveChannels(self, channels, windows=None, colors=None,
                             Must be list for each channel
         :param colors:      List of colors. ['F00', None, '00FF00'].
                             Must be item for each channel
-        :param reverseMaps: List of boolean (or None). If True/False then
+        :param invertMaps:  List of boolean (or None). If True/False then
                             set/remove reverseIntensityMap on channel
         """
+        if reverseMaps is not None:
```
joshmoore, line 8401: If this is set and invertMaps is _not_ set, then invertMaps should be updated to make use of this value (even if the warning should still be printed)

joshmoore (changes requested):
One more small issue but otherwise this is a nice example of renaming a model object via deprecation that we can use in the future. :+1:

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -811,19 +811,26 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
...
         return
-    reverses = _get_maps_enabled(r, 'reverse', img.getSizeC())
+    invert_flags = None
+    if 'maps' in r:
+        reverses = _get_maps_enabled(r, 'reverse', img.getSizeC())
+        # 'reverse' is now deprecated (5.4.0). Also check for 'invert'
+        invert_flags = _get_maps_enabled(r, 'inverted', img.getSizeC())
+        # invert is True if 'invert' OR 'reverse' is enabled
+        invert_flags = [z[0] if z[0] is not None else z[1] for z in
+                        zip(invert_flags, reverses)]
```
will-moore: File "/opt/hudson/workspace/WEB-DEV-merge-deploy/OMERO.py/lib/python/omeroweb/webgateway/views.py", line 821, in _get_prepared_image
   zip(invert_flags, reverses)]
TypeError: zip argument #1 must support iteration

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: Fixed in 23fbd02ac69e342a775fb81545867e06b164d2d3

will-moore:
To test last commit, right-click on the main image in Preview panel -> open in new tab. Then edit the url query ```&maps=[]``` to some invalid JSON, e.g. ```&maps=[InvalidJSON}]``` and enter. Shouldn't see the error from my comment above: https://github.com/openmicroscopy/openmicroscopy/pull/5470#pullrequestreview-60234925

jburel:
No error notice
I tried:
*  a combination of ``inverted`` and ``reverse``: works as expected
* invalid json: works no error

jburel:
@joshmoore still marked as "requested changes"
The changes have been made, ok to approve?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8674,20 +8686,25 @@ def isGreyscaleRenderingModel(self):
 
     @assert_re()
     def setReverseIntensity(self, channelIndex, reverse=True):
+        """Deprecated in 5.4.0. Use setChannelInverted()."""
```
joshmoore, line 8695: A warning could also be raised here.

components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.viewport.js
```diff
@@ -1159,8 +1165,15 @@ jQuery._WeblitzViewport = function (container, server, options) {
     }
     if (query.maps) {
       query.maps.map(function(m, idx){
-        if (m.reverse) {
-          this.setChannelReverseIntensity(idx, m.reverse.enabled, true);
+        // Handle deprecated 'reverse' OR 'invert' codomain map
+        if (m.reverse || m.inverted) {
+          var enabled;
+          if (m.reverse) {
```
joshmoore: This says to me that `reverse` has priority over `inverted` whereas above in `setActiveChannels` it's the other way around. Likely not a big issue, but if this is to be an example of deprecating a term, it might should be consistent.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -811,19 +811,27 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
...
     if img is None:
         return
-    reverses = _get_maps_enabled(r, 'reverse', img.getSizeC())
+    invert_flags = None
+    if 'maps' in r:
+        reverses = _get_maps_enabled(r, 'reverse', img.getSizeC())
+        # 'reverse' is now deprecated (5.4.0). Also check for 'invert'
+        invert_flags = _get_maps_enabled(r, 'inverted', img.getSizeC())
+        # invert is True if 'invert' OR 'reverse' is enabled
+        if reverses is not None and invert_flags is not None:
```
joshmoore, line 820: Should this be `and invert_flags is None` to match the logic of `setActiveChannels`?

joshmoore (approved):

joshmoore:
The general thrust of the "changes requested" have all be addressed so I've clicked on "approve" just now. However, there are 3 minor issues that I commented on while re-reading

will-moore:
Few robot tests failing: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/702/robot/

components/tools/OmeroWeb/omeroweb/webgateway/views.py
will-moore: No, I need to check if both are None here since ```zip(invert_flags , reverses)``` will fail if either is ```None```. If there's no ```"reverse"``` in the request `maps` query string then ```reverses = [None, None]``` here. The following line ```z[0] if z[0] is not None else z[1]``` is where I'm checking the "invert_flags" first, then using "reverse" if invert is not used.

joshmoore: Thank you for the explanation and I'm certainly splitting hairs at this point, but I'm not sure what would need to happen for this to occur. I'd think if `invert_flags` is not None, then no further processing is needed.

will-moore: In the case where url is
```
...&maps=[{"inverted":{"enabled":true}},{"reverse":{"enabled":true}}]
```
then we want to take into account both the 'inverted' and 'reverse' maps.
```
invert_flags = [True, None]
reverses = [None, True]
```
So, we cannot ignore ```reverses``` even if ```invert_flags is not None```.

Or in this case, where we are using the old url:
```
...&maps=[{"reverse":{"enabled":true}},{"reverse":{"enabled":true}}]
```
then
```
invert_flags = [None, None]
reverses = [True, True]
```
So even now, ```invert_flags is not None```, but we are zipping and comparing with reverses for each channel in turn.

jburel: The combination of inverted/reverse works.
 tested yesterday
Maybe a test for the combination?

pwalczysko:
The functionality retested and looks good. 

The robotframework has no results yet nevertheless, because the builds (703, 704) are failing prior to any robot tests runs.

will-moore:
Test LUTs Inverted passing:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/705/robot/Web%20&%20Web/Web_1/Rdef%20Test/Test%20LUTs%20Inverted/

will-moore:
@jburel If happy with this it would be good to get it merged since it conflicts with https://github.com/openmicroscopy/openmicroscopy/pull/5484 which hasn't been reviewed yet.

jburel:
if you could add a warning as indicated https://github.com/openmicroscopy/openmicroscopy/pull/5470#pullrequestreview-60812895
to match the rest then we should be good.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8680,20 +8692,27 @@ def isGreyscaleRenderingModel(self):
...
+        warnings.warn("Deprecated in 5.4.0. Use setChannelInverted()",
+                      DeprecationWarning)
```
will-moore, line 8697: @jburel Warning was already added here, but it was the line below Josh's comment so it didn't register.

jburel: thanks

PR #5392, opened 2017-07-21
revert workaround for IcePy communicator leak

mtbc:
# What this PR does

Removes a prior workaround for Ice leaking Communicator objects which may now be fixed in Ice 3.6 which OMERO 5.4 requires. These days we are seeing "communicator not destroyed" errors from OMERO.cli that may be caused by the workaround.

# Testing this PR

Review the related reading to see if either the original problem or the worrying messages occur.

# Related reading

https://forums.zeroc.com/discussion/4969/python-seems-to-leak-communicator-objects
https://trello.com/c/DPv5q2Ec/77-command-line-intermittent-error

PR #5471, opened 2017-08-25
Connection cleanup on 404

will-moore:
# What this PR does

Try to fix the increase in number of processes seen with 404 pages: https://trello.com/c/2fI9hSVQ/127-demo-internal-server-error-briefly

The ```@login_required``` decorator normally closes all BlitzGateway services but if an exception is raised by the wrapped method then this wasn't happening. This included ```Http404()``` which is raised if e.g. ```conn.getObject("Image", id)``` returns None in many webgateway views methods.

We now wrap the calling of the wrapped method with a try/finally to make sure we always try to close services.

cc @chris-allan 

# Testing this PR

1. Previously, every request to a ```webgateway``` view that returns 404 resulted in an increase of 4 processes. E.g. on web-dev-latest:
```
[centos@spider-1 OMERO.py]$ ps -efT | grep gunicorn | wc
    122    2915   36623
# Then refresh  webgateway/img_detail/1
[centos@spider-1 OMERO.py]$ ps -efT | grep gunicorn | wc
    126    3011   37831
```
This should be fixed - Number of processes should not increase.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/decorators.py
@@ -475,7 +475,7 @@ class login_required(object):
             try:
                 retval = f(request, *args, **kwargs)
             finally:
-                # If f() raised Exception, e.g. Http404() we still need to cleanup services
+                # If f() raised Exception, e.g. Http404() we must still cleanup
                 try:
                     logger.debug(
                         'Doing connection cleanup? %s' % doConnectionCleanup)
@@ -485,7 +485,7 @@ class login_required(object):
                                 v.close()
                             conn.c.closeSession()
                 except:
-                    logger.warn('Failed to clean up connection.', exc_info=True)
+                    logger.warn('Failed to clean up connection', exc_info=True)
             return retval
         return update_wrapper(wrapped, f)
 
--- a/components/tools/OmeroWeb/omeroweb/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/decorators.py
@@ -421,7 +421,7 @@ class login_required(object):
             return None
 
         session['connector'] = connector
-        return connection
+        return
 
     def __call__(ctx, f):
         """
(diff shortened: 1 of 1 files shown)


jburel:
```
./components/tools/OmeroWeb/omeroweb/decorators.py:478:80: E501 line too long (91 > 79 characters)
./components/tools/OmeroWeb/omeroweb/decorators.py:488:80: E501 line too long (80 > 79 characters)

```
flake8 issue

components/tools/OmeroWeb/omeroweb/decorators.py
```diff
@@ -421,7 +421,7 @@ def get_authenticated_connection(self, server_id, request):
             return None
 
         session['connector'] = connector
-        return connection
+        return
```
joshmoore, line 424: What does it mean for this return value to be `None`?

will-moore:
Initial testing today seems to have made web-dev-merge a lot more robust....

components/tools/OmeroWeb/omeroweb/decorators.py
will-moore: Returning None from ```get_authenticated_connection()``` simply means we are not logged in.
I made this change because ```PyCharm``` told me that this was an "Unresolved reference", and inspection of the code confirms that if you reach this point, ```connection``` would not be defined.

joshmoore: Cheers. Comparing with the previous block, I partially wonder if this should be `del session['connector']` then to match the `return None` behavior. Since (I think) `connector` must be `None` at this point, it likely doesn't matter too much. Which is all a long way of saying that at some point we should probably clean-up / simplify this block.

joshmoore:
Big :+1: for invariably cleaning up blitz resources! I'm unsure why we've only begun seeing this in 5.4, but it's certainly a candidate for any potential patches to 5.3.

mtbc:
Maybe revealed by #5392?

components/tools/OmeroWeb/omeroweb/decorators.py
chris-allan: Just for consistency with the style of the other conditions and the expected return value let's `return None` here or switch the other locations to just `return`.

chris-allan:
I wouldn't have thought so @mtbc but it's possible.

joshmoore:
If *so*, that would be quite the indictment of gh-5392.

PR #5474, opened 2017-08-28
Web diagnostics

dominikl:
# What this PR does

`./omero admin diagnostics` also checks status and logs of OMERO.web, but as OMERO.web could be run independantly, it falsely reports an error in that case.
This PR removes the 'web' part from 'admin diagnostics' and puts it into the web module as 'web diagnostics'

# Testing this PR

- Run `./omero admin diagnostics` and check that you don't see any web related output.
- Run `./omero web diagnostics` and check that you see all the web related output.

# Related reading

https://trello.com/c/XdIDdBNS/137-omero-admin-diagnostics-web-status-is-misleading

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -956,7 +956,7 @@ class DiagnosticsControl(BaseControl):
         diagnostics = parser.add(
             sub, self.diagnostics,
             "Run a set of checks on the current, "
-            "preferably active ")
+            "preferably active server")
         diagnostics.add_argument(
             "--no-logs", action="store_true",
             help="Skip log parsing")
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -943,6 +943,32 @@ class BaseControl(object):
             out += ","
         return out.rstrip(",")
 
+
+class DiagnosticsControl(BaseControl):
+    """
+    Superclass (and SPI-interface) for any control commands that would
+    like to provide a "diagnostics" method, like bin/omero admin diagnostics
+    and bin/omero web diagnostics. The top-level diagnostics command then
+    can find each such plugin and iterate over it.
+    """
+
+    def _add_diagnostics(self, parser, sub):
+        diagnostics = parser.add(
+            sub, self.diagnostics,
+            "Run a set of checks on the current, "
+            "preferably active ")
+        diagnostics.add_argument(
+            "--no-logs", action="store_true",
+            help="Skip log parsing")
+
+    def _diagnostics_banner(self, control_name):
+
+        self.ctx.out("""
+%s
+OMERO Diagnostics (%s) %s
+%s
+        """ % ("="*80, control_name, VERSION, "="*80))
+
     def _sz_str(self, sz):
         for x in ["KB", "MB", "GB"]:
             sz /= 1000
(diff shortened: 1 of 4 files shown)


dominikl:
I did quite a lot copy/pasting (methods like `sz_str`, `item`) from admin.py to web.py. Guess there's a better way to do this, worth it?

components/tools/OmeroPy/src/omero/plugins/web.py
```diff
@@ -686,6 +694,80 @@ def set_environ(self, ice_config=None):
...
+                        elif re.match(err_regex, lcl):
+                            err += 1
+                    msg = ""
+                    if warn or err:
+                        msg = " errors=%-4s warnings=%-4s" % (err, warn)
+                    self.ctx.out("%-12s %s" % (sz_str(p.size), msg))
+
+        # OMERO.web diagnostics
+        self.ctx.out("")
+        from omero.plugins.web import WebControl
```
sbesson: Since this is now a method in `Webcontrol`, this import is unecessary and `status(args)` can be used directly 2 lines below

sbesson (commented):
Agreed re  code duplication. Thinking of the ongoing micro-services, I could imagine some of this logic re-use more generically e.g. for an `bin/omero thumbnails-ms diagnostics`.

An immediate way to refactor would be to derive a new subclass of `BaseControl` with the helper methods in `cli.py` and make `AdminControl` and `WebControl`  inherit it.

joshmoore:
Something for a follow-on:

> An immediate way to refactor would be to derive a new subclass of BaseControl with the helper methods in cli.py and make AdminControl and WebControl inherit it.

`ctx` provides access to all `Control` objects. Perhaps instead we make an interface `DiagnosticsControl` and use multiple inheritance. Then `bin/omero diagnostics` could loop through all such controls.

dominikl:
Moved the common methods to `BaseControl`. Is that better? By the way, I'm getting bit out of my python comfort zone here... ;-) Edit: ~~And it didn't work apparently.~~ Was just flake8

joshmoore:
Pushed 11413f02d380b94c590b02a3a7b441a6dbfc115f for discussion. To some extent, this _reintroduces_ the confusion of `bin/omero admin diagnostics`, so we'd need to decide if this "advanced usage" is worth it. Regardless, it does introduce a more formal "diagnostics API" which may be worthwhile.

cc: @jburel @sbesson

joshmoore:
Opened https://github.com/dominikl/openmicroscopy/pull/9 *without* the `bin/omero diagnostics` command (which I'll add to github.com/joshmoore/omero-cli-diagnostics)

jburel:
@dominikl can we retest this PR?

dominikl:
Yes, I think so. Sorry should have listed it on standup.

components/tools/OmeroPy/src/omero/cli.py
```diff
@@ -944,6 +944,75 @@ def _order_and_range_ids(self, ids):
...
+    like to provide a "diagnostics" method, like bin/omero admin diagnostics
+    and bin/omero web diagnostics. The top-level diagnostics command then
+    can find each such plugin and iterate over it.
+    """
+
+    def _add_diagnostics(self, parser, sub):
+        diagnostics = parser.add(
+            sub, self.diagnostics,
+            "Run a set of checks on the current, "
+            "preferably active ")
```
manics: preferably active... `server`? `application`?
```
$ omero admin --help
...
  <subcommand>
    diagnostics       Run a set of checks on the current, preferably active
    start             Start icegridnode daemon and waits for required components to come up, i.e. status == 0
...
```

manics (changes requested):
Looks fine apart from the help text.

manics:
Good to merge

PR #5475, opened 2017-08-28
test_api_images filter ALL by owner

will-moore:
# What this PR does

Fixes ```test_api_images.test_dataset_images``` listing "ALL" images, we now filter by owner so that the test doesn't fail when globally visible images are present (e.g. in ```user``` group).

cc @jburel


jburel:
Looking at the diff with 5.3
I probably only need to filter by group
This is already the case in develop but not in 5.3
So this PR is not needed 
Thanks for checking, closing

PR #5476, opened 2017-08-28
min/max float default

dominikl:
# What this PR does

If a float image doesn't have statsinfo, use `-2147483648` and `2147483647` as default pixel range.

# Testing this PR

Web / iViewer:
Use the same image as for testing #5445 . In the image preview click on 'Min/Max'. Check that above values are used as pixel window (should also match Insight).

# Related reading

https://trello.com/c/F7zMjEbh/134-clients-no-stats-controls-check

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7412,8 +7412,8 @@ class _ChannelWrapper (BlitzObjectWrapper):
                            PixelsTypeuint16: 0,
                            PixelsTypeint32: -32768,
                            PixelsTypeuint32: 0,
-                           PixelsTypefloat: -32768,
-                           PixelsTypedouble: -32768}
+                           PixelsTypefloat: -2147483648,
+                           PixelsTypedouble: -2147483648}
                 pixtype = self._obj.getPixels(
                     ).getPixelsType().getValue().getValue()
                 return minVals[pixtype]
@@ -7438,8 +7438,8 @@ class _ChannelWrapper (BlitzObjectWrapper):
                            PixelsTypeuint16: 65535,
                            PixelsTypeint32: 32767,
                            PixelsTypeuint32: 65535,
-                           PixelsTypefloat: 32767,
-                           PixelsTypedouble: 32767}
+                           PixelsTypefloat: 2147483647,
+                           PixelsTypedouble: 2147483647}
                 pixtype = self._obj.getPixels(
                     ).getPixelsType().getValue().getValue()
                 return maxVals[pixtype]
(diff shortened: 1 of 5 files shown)


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7412,8 +7412,8 @@ def getWindowMin(self):
                            PixelsTypeuint16: 0,
                            PixelsTypeint32: -32768,
                            PixelsTypeuint32: 0,
-                           PixelsTypefloat: -32768,
-                           PixelsTypedouble: -32768}
+                           PixelsTypefloat: -2147483648,
```
jburel: Still a problem for pixelstype signed/unsigned 32
The value should not be hardcoded in the file but we should use the methods ``getPixelsTypeLowerBound/UpperBound`` from the RE to determine the min/max value
since the values are determined by taken the "Significants bit if the information is available
see https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/rendering/src/omeis/providers/re/metadata/StatsFactory.java#L222

jburel:
Web and insight clients are now in synch
Need to adjust iviewer
Could you add tests to ``components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py``

dominikl:
Yes, I'll add a test. Doesn't iViewer use this method, too? Seems to work.

jburel:
sorry iviewer is fixed for min/max
it does not behave as the other clients for imported that will need to be updated
nothing to do with this PR

jburel:
card for iviewer https://trello.com/c/xOvfDC49/25-imported-full-range-button

dominikl:
Hm, I think I'm a bit stuck here. How can I import a float image which can be used in `test_rdefs.py` ? /cc @will-moore ?

will-moore:
Something like:
```
self.import_fake_file(name="test&pixelType=float.fake")
```
See
https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/testlib/__init__.py#L309
and
https://docs.openmicroscopy.org/bio-formats/5.6.0/developers/generating-test-images.html

dominikl:
Thanks @will-moore . Found a slightly different way, but yes, with `.fake` image.

components/tools/OmeroPy/src/omero/gateway/scripts/dbhelpers.py
```diff
@@ -406,15 +406,19 @@ def create(self):
...
+                # If it's a .fake file, simply create it
+                os.close(os.open(fpath, os.O_CREAT | os.O_EXCL))
+            else:
+                # First try to download the image
+                try:
+                    # print "Trying to get test image from " + TESTIMG_URL +
+                    # self.filename
+                    sys.stderr.write('<')
+                    f = urllib2.urlopen(TESTIMG_URL + self.filename)
+                    open(fpath, 'wb').write(f.read())
```
jburel: you should use 
```
with open(fpath, 'wb') as f:
   f.read()
```
to make sure resources are closed

jburel:
test/integration/gatewaytest/test_rdefs.py::TestRDefs::testFloatDefaultMinMax PASSED is passing
Thanks

PR #5479, opened 2017-08-30
Fix double-loading of right panel

will-moore:
# What this PR does

Fixes the double-loading of the right panel when Dataset thumbnails are clicked.

# Testing this PR

1. To see the bug, e.g. in idr /nightshade / web-dev-latest,  open dev-tools (right-click on page and "Inspect Element") and view the Network tab. Clicking on a Dataset thumbnail will show 2 requests for /metadata_details/image/id/ 
2. Repeat with this PR web-dev-merge and see that only a single request is made to load the right panel.
3. Check that other selection changes are still working OK. Single/multi-select of various objects and check they are loaded in right panel correctly.


dominikl:
Good to merge 👍

PR #5482, opened 2017-08-31
isFullAdmin Java gateway

dominikl:
# What this PR does

Adds another method `isFullAdmin` to the Java gateway, which checks if the user is an administrator with full admin privileges.

# Testing this PR

Check that `testAdminPrivileges` integration test still passes.

# Related reading

https://trello.com/c/FjayBnvM/39-gateway-support-for-roles

PR #5483, opened 2017-08-31
Isfulladmin Python gateway

dominikl:
# What this PR does

Adds another method isFullAdmin to the Python gateway, which checks if the user is an administrator with full admin privileges. 

# Testing this PR

Check that `test_full_admin_privileges` integration test still passes.
Which does not at the moment, therefore
--exclude

# Related reading

https://trello.com/c/FjayBnvM/39-gateway-support-for-roles

Same for Java gateway: #5482

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_permissions.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_permissions.py
@@ -43,3 +43,25 @@ class TestPrivileges(ITest):
 
         expected = ['Sudo', 'ModifyGroup', 'ModifyUser']
         assert set(expected) == set(conn.getAdminPrivileges(exp.id.val))
+
+    def test_full_admin_privileges(self):
+        """Test full admin privileges check."""
+        conn = BlitzGateway(client_obj=self.root, try_super=True)
+        allPrivs = []
+        for p in conn.getEnumerationEntries('AdminPrivilege'):
+            allPrivs.append(p.getValue())
+        conn.updateAdminPrivileges(self.user.id.val, add=allPrivs)
+        test = conn.getCurrentAdminPrivileges()
+        assert set(allPrivs) == set(test)
+        assert conn.isFullAdmin()
+
+        conn.updateAdminPrivileges(self.user.id.val, remove=['Chown'])
+
+        privs = set(conn.getCurrentAdminPrivileges())
+        print "privs:"
+        print privs
+
+        print "all:"
+        print allPrivs
+
+        assert not conn.isFullAdmin()
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2324,6 +2324,21 @@ class _BlitzGateway (object):
 
         return self.getEventContext().isAdmin
 
+    def isFullAdmin(self):
+        """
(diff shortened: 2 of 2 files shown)


components/tools/OmeroPy/test/integration/gatewaytest/test_permissions.py
```diff
@@ -43,3 +43,25 @@ def test_update_admin_privileges(self):
...
+        for p in conn.getEnumerationEntries('AdminPrivilege'):
+            allPrivs.append(p.getValue())
+        conn.updateAdminPrivileges(self.user.id.val, add=allPrivs)
+        test = conn.getCurrentAdminPrivileges()
+        assert set(allPrivs) == set(test)
+        assert conn.isFullAdmin()
+
+        conn.updateAdminPrivileges(self.user.id.val, remove=['Chown'])
+
+        privs = set(conn.getCurrentAdminPrivileges())
```
dominikl, line 60: User still has all privileges, including 'Chown'. Does anyone know why?

mtbc: In the same session you have to wait a minute for the cache to expire. If they have a new session you should see the change immediately.

joshmoore: Perhaps because https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/logic/AdminImpl.java#L1622 doesn't refresh the event context? If so, then after https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/logic/AdminImpl.java#L1359 we should do so.

You can test the hypothesis by adding a call to [IAdmin.synchronizeLoginCache](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/logic/AdminImpl.java#L415) between your `update` and your `get`

joshmoore: :-1: for introducing a `sleep()`

mtbc: Should also work fine to just re-login, whatever the Python test equivalent is of `loginUser(eventContext)` is.

joshmoore: Agreed, but it'd be nice to solve this server side so that it's not necessary for clients to jump through hoops.

dominikl: I tried both, wait and call `synchronizeLoginCache` neither did work. I'll check if I can somehow re-login in the test.

joshmoore: @dominikl : and an explicit call to `getAdminService().getEventContext()`?

dominikl: Ok, I guess one problem is that I create the `conn` with `client_obj=self.root`...  I just remove the test, maybe someone else can add one. Can't get my head around this `self.*` stuff, `self.root`, `self.client`, `self.user`, ... Makes me really appreciate real OO typed languages.

dominikl:
I'm closing this. Can't make any progress.

PR #5486, opened 2017-09-01
share_add_no_images

will-moore:
# What this PR does

Fixes error at ```webclient/action/add/share/``` if no images are specified with ```?image=1```.
Not sure how user got this error, since webclient only loads the create-share form when Images are selected in the UI.

# Testing this PR

1. Go to ```webclient/action/add/share/``` and check you get a message instead of an error.

2. Try creating a share as normal in webclient to check that this still works OK.


mtbc:
Got "Server Error. (500)" but https://ci.openmicroscopy.org/job/OMERO-DEV-merge-build/ICE=3.6,jdk=8_LATEST,label=octopus/ didn't run so maybe that confirms the problem that this PR fixes.

mtbc (approved):
All working today with this fix merged in.

PR #5490, opened 2017-09-04
Add update tables method

dominikl:
# What this PR does

At the moment there are only `add` and `get` methods in the `TablesFacility`. This PR adds an `update` method. It is quite limited, i.e. one can only modify values, but not add or delete rows/columns (that's because the OMERO tables interface doesn't have methods for addition/deleting of rows/column, afaik).

# Testing this PR

Not really much to test, check the `testUpdateTable` integration test makes sense and passes, should be sufficient.

# Related reading

https://trello.com/c/aXTsTn5U/221-gateway
https://github.com/imagej/imagej-omero/pull/65#issuecomment-321252703

### DIFF
--- a/components/blitz/src/omero/gateway/facility/TablesFacilityHelper.java
+++ b/components/blitz/src/omero/gateway/facility/TablesFacilityHelper.java
@@ -26,6 +26,7 @@ import omero.gateway.model.PlateData;
 import omero.gateway.model.ROIData;
 import omero.gateway.model.TableData;
 import omero.gateway.model.TableDataColumn;
+import omero.gateway.model.WellData;
 import omero.gateway.model.WellSampleData;
 import omero.grid.BoolColumn;
 import omero.grid.Column;
@@ -225,6 +226,8 @@ public class TablesFacilityHelper {
                 for (int j = 0; j < nRows; j++) {
                     MaskData md = new MaskData(mc.x[j], mc.y[j], mc.w[j],
                             mc.h[j], mc.bytes[j]);
+                    md.setZ(mc.theZ[j]);
+                    md.setT(mc.theT[j]);
                     rowData[j] = md;
                 }
                 dataArray[i] = rowData;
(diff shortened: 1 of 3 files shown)


components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -276,5 +281,46 @@ public void testThreshold() throws Exception {
...
+    public void testUpdateTable() throws Exception {
+        // modify values for row 10 to 20, columns 5, 6 and 7
+        TableData td = tablesFacility.getTable(rootCtx,
+                original.getOriginalFileId(), 10, 20, new int[] { 5, 6, 7 });
 
+        for (int c = 0; c < td.getColumns().length; c++) {
+            Class<?> type = td.getColumns()[c].getType();
+            for (int r = 0; r < td.getData()[0].length; r++) {
+                if (type.equals(String.class)) {
+                    td.getData()[c][r] = "newValue";
```
bramalingam, line 295: Not part of the scope for this PR, but getData method is used for setting data here (which could be confusing for the consumers of the api). I would have expected something of the order of:
 `newValue = td.getData()[c][r]; 
   td.setData('newValue');`

components/blitz/src/omero/gateway/facility/TablesFacilityHelper.java
```diff
@@ -467,4 +470,198 @@ int getNRows() {
...
+     * Note:
+     * - Size and data types must match!
+     * - Size of Double/Float/Long arrays can't be changed!
+     * 
+     * @param toUpdate
+     *            The omero.grid.Data object to update
+     * @param data
+     *            The new data
+     */
+    static void updateData(Data toUpdate, TableData data) {
```
bramalingam, line 484: Maybe this method could take an optional argument (rows and columns to update) as well, and potentially this could improve performance in scattered updates of huge tables.

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -115,7 +119,7 @@ public void testGetTableInfo() throws Exception {
         Assert.assertEquals(info.getColumns(), original.getColumns());
     }
 
-    @Test(dependsOnMethods = { "testAddTable" })
+    @Test(dependsOnMethods = { "testGetTableInfo" })
```
rgozim, line 122: This needs to be reverted. This test only relies on a table existing.

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -137,7 +141,7 @@ public void testSearch() throws Exception {
             }
     }
 
-    @Test(dependsOnMethods = { "testAddTable" })
+    @Test(dependsOnMethods = { "testSearch" })
```
rgozim, line 144: This needs to be reverted. This test only relies on a table existing.

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -171,6 +175,7 @@ public void testInvalidParams() throws Exception {
         }
     }
 
+    @Test(dependsOnMethods = { "testInvalidParams" })
```
rgozim, line 178: This should be reverted. This test doesn't actually have any method dependencies.

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -201,7 +206,7 @@ public void testObjectColumnType() throws Exception {
         Assert.assertEquals(td2.getNumberOfRows(), 1);
     }
 
-    @Test(dependsOnMethods = { "testAddTable" })
+    @Test(dependsOnMethods = { "testObjectColumnType" })
```
rgozim, line 209: This needs to be reverted. Running individual tests shows correct dependencies.

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -210,7 +215,7 @@ public void testGetAvailableTables() throws Exception {
                 original.getOriginalFileId());
     }
 
-    @Test(dependsOnMethods = { "testAddTable" }, invocationCount = 10)
+    @Test(dependsOnMethods = { "testGetAvailableTables" }, invocationCount = 10)
```
rgozim, line 218: This needs to be reverted. This test only relies on a table existing.

components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -262,7 +267,7 @@ public void testReadTable() throws Exception {
         }
     }
 
-    @Test(dependsOnMethods = { "testAddTable" })
+    @Test(dependsOnMethods = { "testReadTable" })
```
rgozim, line 270: This needs to be reverted. This test only relies on a table existing.

jburel:
@dominikl what is the status of this PR?

components/blitz/src/omero/gateway/facility/TablesFacilityHelper.java
dominikl: Not sure that's necessary. `data` doesn't have to be the whole table, `TableData` has internal offsets which specify which particular section of the whole table the `TableData` object represents. My assumption was that the user loads a particular section, modifies the data and then writes that chunk back to the table.

dominikl:
I think it's good to merge. The tests have to be run in the specified order, they do not only rely on the table existing. For example if the testUpdateTable runs before the testReadTable testReadTable will fail because it compares the data with the original data. I know ideally the tests should be self contained. But that would require a complete reimplementation of the test class.

joshmoore:
Re: test ordering -- understood. We can always come back to that in the future.

Re: https://github.com/openmicroscopy/openmicroscopy/pull/5490/files/6b99631b5c959846f7097a7283830a5015492eaf#diff-da1df632c1cb56804742faeb754d2b60 and https://github.com/openmicroscopy/openmicroscopy/pull/5490/files/6b99631b5c959846f7097a7283830a5015492eaf#diff-32a67f881554e8186a3a9818db66c19a -- these may need review as we start to use this more in anger. 

See https://trello.com/c/GJMSXRbk/44-pre-merge-review-updatedata

PR #5492, opened 2017-09-04
Unify style with other `return` statements

chris-allan:
# What this PR does

Unify style with other `return` statements in the `get_authenticated_connection()` block. Follow up from comment on #5471.

# Testing this PR

N/A; purely stylistic.

# Related reading

* #5471


will-moore:
Apologies for not getting back to this.
Looks good, thanks.

PR #5493, opened 2017-09-04
BF bump to 5.7.0

dgault:
# What this PR does

Bump bio-formats version to 5.7.0

# Testing this PR

Make sure the build is green

Test import of QA-17824 and verify that each plane has a unique deltaT value rather than each channel having the same value (test comes from PR: https://github.com/openmicroscopy/bioformats/pull/2922)


dominikl:
👍  Good to merge.

PR #5495, opened 2017-09-07
Matlab deleteObjects use Requests util

dominikl:
# What this PR does

- Updates Matlab's `deleteObjects` method to use the Java `Requests` utils.
- Updates the Delete example to use the `deleteObjects` method.

# Testing this PR

Run the `examples/Training/matlab/DeleteData.m` example.

# Related reading
 
Follow up for #5460

### DIFF
--- a/components/tools/OmeroM/src/libs
+++ b/components/tools/OmeroM/src/libs
@@ -1,0 +1,1 @@
+../target/libs
--- a/components/tools/OmeroM/src/libs
+++ b/components/tools/OmeroM/src/libs
@@ -1,1 +1,0 @@
-../target/libs
--- a/components/tools/OmeroM/src/annotations/linkAnnotation.m
+++ b/components/tools/OmeroM/src/annotations/linkAnnotation.m
@@ -39,6 +39,10 @@ ip.addRequired('parentId', @(x) isempty(x) || (isvector(x) && isnumeric(x)));
 ip.parse(session, annotation, parentType, parentId);
 objectType = objectTypes(strcmp(parentType, objectNames));
 
+% Throw exception if type is roi
+assert(~strcmp(objectType.class, 'omero.model.Roi'),...
+    ['Cannot link annotations to rois. ']);
+
 % Get the parent object
 if isnumeric(parentId),
     parent = getObjects(session, parentType, parentId);
--- a/components/tools/OmeroM/src/delete/deleteObjects.m
+++ b/components/tools/OmeroM/src/delete/deleteObjects.m
@@ -48,7 +48,7 @@ for i = 1 : numel(ids)
 end
 targetObject = java.util.Hashtable;
 targetObject.put(objectType.delete2,idlist);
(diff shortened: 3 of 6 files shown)


components/tools/OmeroM/src/libs
```diff
@@ -0,0 +1 @@
+../target/libs
```
sbesson: what is this for?

dominikl: Argh, accidentally commited, will remove again (symlink to libs, needed locally to get the matlab stuff running).

mtbc:
The diff looks pretty good to me, thank you @dominikl.

sbesson:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/733 looks good with this PR included.

 Overall,
- :+1: for using the new `omero.gateway` methods and the requests builders
- :+1: for having `DeleteData.m` used `deleteObjects` directly

I think the main concern is with the generic addition of `Roi` to `getObjectTypes` as this (arguably poorly named) function is used as a validator in a couple of other places. For instance `getObjects(session, 'roi', id);` would now be allowed but probably fail in the implementation.

dominikl:
Sorry, was away, couldn't fix your remark @sbesson . Is it too late now for 5.4.0? Exclude PR?

jburel:
@dominikl if you could fix that today that will be great

sbesson:
Discussed earlier today with @dominikl, our proposal would be to keep the current changes but be defensive and add some additional check in `createObject` and `getObjects` similar to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroM/src/io/getObjects.m#L68 to fail fast if these functions are called with `roi` as the object type at the moment.

dominikl:
Checked the methods which use `getObjectTypes`: Only `getObject` and `linkAnnotation` are a problem, because they trigger a call to `loadContainerHierarchy` (`createObject` works fine also with Roi). Added the 'Throw exception' clause @sbesson mentioned to these two methods.

sbesson:
https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/741 passed as expected with this PR included, testing the deletion workflow:

```
Created session for user training_user (id: 4214) using group training_group (id: 3194)
Image 8198 created
Create rectangular and point shape
Created ROI 655
Reading ROIs attached to image 8198
Found 1 ROI(s)
  Removing 2 shapes
Deleting ROI 655
Deleting image 8198
Image 8198 deleted
```

Manually tested the behavior of the other affected methods:

```
>> getObjects(s, 'roi', 1)
Error using getObjects (line 74)
Rois have to be loaded via roi service.

>> t=omero.model.TagAnnotationI();
>> linkAnnotation(s,t,'roi',1)
Error using linkAnnotation (line 43)
Cannot link annotations to rois.
 
>> r=createObject(s,'roi','this is a roi');
>> r.getName().getValue()
 
ans =
 
this is a roi
```

:+1: Good to merge.

sbesson (approved):

PR #5496, opened 2017-09-07
isFullAdmin training example

will-moore:
# What this PR does

Add a couple of new BlitzGateway methods to Training examples.
See https://trello.com/c/fgRo78J1/143-blitz-gateway-and-other-api-changes

No breaking changes to BlitzGateway for 5.4.0 as far as I can see.

# Testing this PR

1. Training job should be green


dominikl:
Looks good, works like expected (tested locally). 👍

PR #5497, opened 2017-09-07
Redis

jburel:
# What this PR does

Install dependency required to use redis
Including this requirement file will not change any workflow when installing web dependencies

# Testing this PR

Check that the dependency is installed

# Related reading

https://trello.com/c/mPY2uYY6/124-include-django-redis-in-the-default-requirementstxt-for-omeroweb

Doc PR, omero-install and omeroweb-install PRs are also needed

cc @sbesson @joshmoore

### DIFF
--- a/components/tools/OmeroWeb/requirements-py27-ice35.txt
+++ b/components/tools/OmeroWeb/requirements-py27-ice35.txt
@@ -9,4 +9,4 @@ django-pipeline==1.3.20
 gunicorn>=19.3
 
 omero-marshal==0.5.1
--r requirements-redis.txt
+django-redis>=4.4
--- a/components/tools/OmeroWeb/requirements-py27-ice35.txt
+++ b/components/tools/OmeroWeb/requirements-py27-ice35.txt
@@ -9,3 +9,4 @@ django-pipeline==1.3.20
 gunicorn>=19.3
 
 omero-marshal==0.5.1
+-r requirements-redis.txt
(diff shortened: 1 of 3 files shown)


joshmoore:
:+1: for the general change. However, I think the naming of these files is becoming less and less easier to follow. Might be worth scheduling when a renaming could happen.

NB: perhaps for the simplicity of the patch, force rebase?

jburel:
The name of the files is a problem we have the legacy nginx, redis etc.
We will need to have a clear picture of what is used where
omeroweb, devspace etc.

jburel:
I remember already having a discussion for 5.3 about those names and we did not come to anything meaningful

joshmoore:
Naming's always fun. But minimally, having a file named "redis" with only "redis" in it, when redis is now recommended seems overkill.

jburel:
I agree!! 
but it is still used in devspace for example so we will need to synch a release of devspace at the same time if we remove it
I could mark the redis file deprecated

will-moore:
Minor point: Do we want to start using ```requirements-redis.txt``` in 2 places at the same time as deprecating it?
But not a blocker for me. So, Good to merge.

jburel:
I can replace its usage directly by installing redis in the relevant files
I was thinking to do that when the requirement will be removed.
Happy to do that in this PR

jburel:
@will-moore no longer use the file ``requirements-redis`` as suggested

jburel:
@manics is happy if everybody is happy
@will-moore any more comments?

will-moore:
Looks good. 👍

PR #5501, opened 2017-09-12
Rendering settings exposure - quantization maps

waxenegger:
Certain rendering options such as quantization maps have only been used in insight.

To enable web clients to use these options -since they rely on the render methods in the webgateway-
 the most sensible approach seems to me to add a parameter to the query string.

TEST: use the render methods/requests to fetch an image, adding a "quantization" object to the json that is handed of maps parameter (which already contains invert). "quantization" can contain properties family and coefficient, e.g.
```
...&maps=[{"reverse":{"enabled":true},"quantization":{"family":"logarithmic","coefficient":0.2}},{"reverse":{"enabled":true},"quantization":{"family":"exponential","coefficient":1.0}}]
```

Check that the tests are green: https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest.test_rdefs/

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7475,6 +7475,34 @@ class _ChannelWrapper (BlitzObjectWrapper):
                 reverse = True
         return reverse
 
+    def getFamily(self):
+        """
+        Returns the channel family
+
+        :return:    the channel family
+        :rtype:     String
+        """
+        if self._re is None:
+            return None
+
+        f = self._re.getChannelFamily(self._idx)
+        if f is None:
+            return f
+
+        return f.getValue()
+
+    def getCoefficient(self):
+        """
+        Returns the channel coefficient
+
+        :return:    the channel coefficient
+        :rtype:     double
+        """
+        if self._re is None:
+            return None
+
+        return self._re.getChannelCurveCoefficient(self._idx)
+
 ChannelWrapper = _ChannelWrapper
 
 
@@ -7550,6 +7578,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
     _re = None
     _pd = None
     _rm = {}
+    _qf = {}
     _pixels = None
     _archivedFileCount = None
     _filesetFileCount = None
@@ -8714,6 +8743,46 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         # If we want to invert, add it to the channel (again)
         if inverted:
             self._re.addCodomainMapToChannel(r, channelIndex)
+
+    @assert_re()
+    def getFamilies(self):
+        """
+        Gets a list of available families.
+
+        :return:    Families
+        :rtype:     List of :class:`BlitzObjectWrapper`
+        """
+        if not len(self._qf):
+            for f in [BlitzObjectWrapper(self._conn, f)
+                      for f in self._re.getAvailableFamilies()]:
+                self._qf[f.value.lower()] = f
+        return self._qf.values()
+
+    @assert_re()
+    def setQuantizationMap(self, channelIndex, family, coefficient):
+        """
+        Sets the quantization strategy according to the given family
+
+        :param channelIndex:    The index of channel (int)
+        :param family:          The family (string)
+        :param coefficient:     The coefficient (float)
+        """
+        self.getFamilies()
+        f = self._qf.get("linear")
+        try:
+            f = self._qf.get(str(family).lower(), f)
+        except:
+            pass
+
+        c = 1.0
+        try:
+            c = float(coefficient)
+            if c < 0 or c > 1.0:
+                c = 1.0
+        except:
+            pass
+
+        self._re.setQuantizationMap(channelIndex, f._obj, c, False)
 
     @assert_re(ignoreExceptions=(omero.ConcurrencyException))
     def getRenderingDefId(self):
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -7496,7 +7496,7 @@ class _ChannelWrapper (BlitzObjectWrapper):
         Returns the channel coefficient
 
         :return:    the channel coefficient
-        :rtype:     double
+        :rtype:     float
         """
         if self._re is None:
             return None
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8761,12 +8761,16 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
     @assert_re()
     def setQuantizationMap(self, channelIndex, family, coefficient):
         """
-        Sets the quantization strategy according to the given family
+        Sets the quantization strategy to the given family
+        and coefficient
 
         :param channelIndex:    The index of channel (int)
         :param family:          The family (string)
         :param coefficient:     The coefficient (float)
         """
+        if channelIndex < 0 or channelIndex >= self.getSizeC():
+            return
+
         f = self.getFamilies().get("linear")
         try:
             f = self._qf.get(family.lower(), f)
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8771,9 +8771,10 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         if channelIndex < 0 or channelIndex >= self.getSizeC():
             return
 
-        f = self.getFamilies().get("linear")
+        families = self.getFamilies()
+        f = families.get("linear")
         try:
-            f = self._qf.get(family.lower(), f)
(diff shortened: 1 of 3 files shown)


waxenegger:
--exclude

joshmoore:
Tests will be needed for additions to `omero.gateway` to protect against future breaking changes

waxenegger:
yes, that was on my list. 

I briefly talked with @jburel about it. maybe @will-moore can give me tips as well what kind of test/framework might work best. I think robot-tests were used for inverse.

joshmoore:
`OmeroPy/src/omero/gateway` changes will also need `OmeroPy/test/integration/gatewaytest/`

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -762,6 +762,42 @@ def _get_signature_from_request(request):
...
+    if img is None:
+        return
+
+    try:
+        if isinstance(quant_json, (unicode, str)):
+            quant_json = json.loads(quant_json)
+
+        size_c = len(quant_json)
+        if size_c is None:
+            return
```
will-moore: ```len(quant_json)``` wouldn't return ```None```?
```if size_c == 0:``` ?

File: components/tools/OmeroWeb/omeroweb/webgateway/views.py [hunk 2]
```diff
@@ -762,6 +762,42 @@ def _get_signature_from_request(request):
...
+
+        channels = img.getChannels()
+        if (size_c > len(channels)):
+            size_c = len(channels)
+
+        i = 0
+        for c in range(size_c):
+            family = quant_json[i]['family']
+            if family is None:
+                continue
```
will-moore: Do you want to increment ```i``` here? Otherwise ```quant_json[i]['family']``` will be None in the next iteration too?

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -762,6 +762,42 @@ def _get_signature_from_request(request):
...
+            if img.getFamilies().get(family.lower(), None) is None:
+                continue
+            coeff = quant_json[i].get('coefficient', None)
```
will-moore: ```coeff = quant_json[i].get('coefficient', 1.0)```
would save you checking for None below

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -852,6 +895,7 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
 
     img.setInvertedAxis(bool(r.get('ia', "0") == "1"))
     compress_quality = r.get('q', None)
+
```
will-moore: Nicer not to add single blank lines (and above).

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -762,6 +762,42 @@ def _get_signature_from_request(request):
     return rv
 
 
+def _set_quantization_maps(quant_json, img):
```
will-moore: I think it would be better to have this functionality in the BlitzGateway so it can be tested independently of web and can be used by other Python clients.
Tests could be added under ```components/tools/OmeroWeb/test/integration/gatewaytest/test_rdefs.py``` although for "inverted" I added to ```components/tools/OmeroWeb/test/integration/test_rendering.py```

waxenegger: Thanks for pointing me to the test locations.

the method you referred to above essentially just gathers the parameters from the json. most other clients using python would much rather want to call setQuantizationMap directly (without the json detour), right? there is also the wrapper on the image that the method above calls. Granted it is applicable to only single channels but still.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8715,6 +8744,69 @@ def setChannelInverted(self, channelIndex, inverted=True):
...
+        :param coefficient:     The coefficient (float)
+        """
+        if channelIndex < 0 or channelIndex >= self.getSizeC():
+            return
+
+        f = self.getFamilies().get("linear")
+        try:
+            f = self._qf.get(family.lower(), f)
+        except:
+            pass
```
will-moore: Minor point, how about
```
families = self.getFamilies()
f = families.get("linear")
try:
    f = families.get(family.lower(), f)
...
```
Getting the same dict of families in 2 different ways is slightly more confusing, and could lead to errors if someone were to try accessing ```self._qf``` without first calling ```self.getFamilies()```.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8715,6 +8744,69 @@ def setChannelInverted(self, channelIndex, inverted=True):
...
+        of mapping information (for each entry, i.e. channel)
+        e.g. [{'family': 'linear', coefficient: 1.0}]
+
+        :param maps:     list of quantization settings
+        """
+        if not isinstance(maps, list):
+            return
+
+        i = 0
+        for m in maps:
```
will-moore: Could do this to avoid a couple of lines to set and update i...
```
for i, m in enumerate(maps):
```

components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py
```diff
@@ -353,3 +353,106 @@ def testResetDefaults(self, gatewaywrapper):
...
+            "family": channels[0].getFamily().getValue(),
+            "coefficient": channels[0].getCoefficient()
+        }
+        # input checks for coefficient (checking against initial values)
+        self.image.setQuantizationMap(1, "linear", "coefficient")
+        self.image.setQuantizationMap(1, "linear", 100)
+        assert settings[1] == {
+            "family": channels[1].getFamily().getValue(),
+            "coefficient": channels[1].getCoefficient()
+        }
```
will-moore, line 392: This ```assert``` and the 2 above it still pass even if I remove all 6 calls to ```self.image.setQuantizationMap()```, so I'm not sure what this is testing, other than ```setQuantizationMap()``` doesn't throw an Exception with invalid inputs?

components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py
```diff
@@ -353,3 +353,106 @@ def testResetDefaults(self, gatewaywrapper):
...
+            }
+            i += 0
+        self.image.setQuantizationMaps(["nonsense", 8])
+        i = 0
+        for chan in channels:
+            assert settings[i] == {
+                "family": channels[i].getFamily().getValue(),
+                "coefficient": channels[i].getCoefficient()
+            }
+            i += 0
```
will-moore: Here too, this would save a few lines in this and the other test:
```
for i, chan in enumerate(channels):
```

components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py
waxenegger: yes, that is the check. that the original values have not been altered. 
see comment: # input checks for family (checking against initial values)

will-moore:
We should probably use the ```maps``` query string for quantization too? E.g.
```
&maps=[{"inverted":{"enabled":false},"quantization":{"family":"exponential","coefficient":0.3}}]
```
instead of current:
```
&maps=[{"inverted":{"enabled":false}}]&quant_maps=[{"family":"exponential","coefficient":0.3}]
```
Then
```
        qm = [m['quantization'] for m in json.loads(r['maps'])]
        img.setQuantizationMaps(qm)
```

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -820,12 +820,10 @@ def _get_prepared_image(request, iid, server_id=None, conn=None,
...
             invert_flags = [z[0] if z[0] is not None else z[1] for z in
                             zip(invert_flags, reverses)]
-
-    # more rendering settings per channel: quantization maps
-    # just applied, not saved at the moment
-    if 'quant_maps' in r:
         try:
-            img.setQuantizationMaps(json.loads(r['quant_maps']))
+            # quantization maps (just applied, not saved at the moment)
+            qm = [m['quantization'] for m in json.loads(r['maps'])]
```
will-moore: Final tweak (sorry)...
Use ``` m.get('quantization') ``` instead of ```m['quantization']``` since we can't assume that it's there.
If it's missing for a single channel, we don't want it to fail for All other channels. E.g. support this:
```
&maps=[{"inverted":{"enabled":false}},{"quantization":{"family":"linear","coefficient":0.1},"inverted":{"enabled":false}}]
```

will-moore:
Thanks. Good to merge.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7475,6 +7475,34 @@ def isInverted(self):
...
+            return f
+
+        return f.getValue()
+
+    def getCoefficient(self):
+        """
+        Returns the channel coefficient
+
+        :return:    the channel coefficient
+        :rtype:     double
```
jburel: we have ``double`` here and ``float`` below for the same parameter

jburel:
Thanks

PR #5502, opened 2017-09-12
Add tests for light admin

jburel:
# What this PR does

Tests the manipulation of enumeration
Context:
* Create a user in the system group with no privileges.
* User is allowed to delete/reset enumeration
From a discussion with @mtbc, this should not be allowed

### DIFF
--- a/components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
+++ b/components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
@@ -2507,25 +2507,33 @@ public class LightAdminRolesTest extends RolesTests {
     }
 
     /**
-     * Resets the enumerations
+     * Tests if deleted enumeration can be reset
      * @throws Exception
      */
     @Test
     public void testResetEnumerationsbyRestrictedSystemUser() throws Exception {
         logNewAdminWithoutPrivileges();
-        ContrastMethod ho = new ContrastMethodI();
-        ho.setValue(omero.rtypes.rstring("testResetEnumerationsbyRestrictedSystemUser"));
         final ITypesPrx types_svc = factory.getTypesService();
         List<IObject> types = types_svc.allEnumerations(ContrastMethod.class.getName());
-        //original number of enumerations
         int n = types.size();
-        ho = (ContrastMethod) types_svc.createEnumeration(ho);
+        Iterator<IObject> i = types.iterator();
+        int count = 0;
+        while (i.hasNext()) {
+            try {
+                types_svc.deleteEnumeration(i.next());
+                count++;
+            } catch (Exception e) {
+                //Cannot delete the enumeration since it is used
+            }
+        }
+        //not all enum are used so we should have deleted at least one
+        Assert.assertTrue(count > 0);
         types = types_svc.allEnumerations(ContrastMethod.class.getName());
-        //check that an enumeration has been added
-        Assert.assertEquals(types.size(), (n+1));
-        types_svc.deleteEnumeration(ho);
+        Assert.assertEquals(types.size(), (n-count));
+        //reset the deleted enumerations
         types_svc.resetEnumerations(ContrastMethod.class.getName());
         types = types_svc.allEnumerations(ContrastMethod.class.getName());
+        //We should be back to the original list
         Assert.assertEquals(types.size(), n);
     }
 
@@ -2576,7 +2584,7 @@ public class LightAdminRolesTest extends RolesTests {
     }
 
     /**
-     * Tests f it possible to modify a script using the upload method from
+     * Tests if it possible to modify a script using the upload method from
      * omero.client
      * @throws Exception
      */
--- a/components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
+++ b/components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
@@ -28,7 +28,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.UUID;
 
-import ome.model.acquisition.Detector;
+import ome.model.enums.DetectorType;
 import ome.model.enums.EventType;
 import omero.RLong;
 import omero.RString;
@@ -112,7 +112,7 @@ public class LightAdminRolesTest extends RolesTests {
     public void tearDown() throws Exception {
         final ITypesPrx svc = root.getSession().getTypesService();
         svc.resetEnumerations(ContrastMethod.class.getName());
-        svc.resetEnumerations(Detector.class.getName());
+        svc.resetEnumerations(DetectorType.class.getName());
         super.tearDown();
     }
 
@@ -2516,7 +2516,7 @@ public class LightAdminRolesTest extends RolesTests {
     public void testResetEnumerationsbyRestrictedSystemUser() throws Exception {
         logNewAdminWithoutPrivileges();
         final ITypesPrx types_svc = factory.getTypesService();
-        List<IObject> types = types_svc.allEnumerations(Detector.class.getName());
+        List<IObject> types = types_svc.allEnumerations(DetectorType.class.getName());
         int n = types.size();
         Iterator<IObject> i = types.iterator();
         int count = 0;
@@ -2530,11 +2530,11 @@ public class LightAdminRolesTest extends RolesTests {
         }
         //not all enum are used so we should have deleted at least one
         Assert.assertTrue(count > 0);
-        types = types_svc.allEnumerations(Detector.class.getName());
+        types = types_svc.allEnumerations(DetectorType.class.getName());
         Assert.assertEquals(types.size(), (n-count));
         //reset the deleted enumerations
-        types_svc.resetEnumerations(Detector.class.getName());
-        types = types_svc.allEnumerations(Detector.class.getName());
+        types_svc.resetEnumerations(DetectorType.class.getName());
+        types = types_svc.allEnumerations(DetectorType.class.getName());
         //We should be back to at the original list. Other enum might have been
         //added by other tests.
         Assert.assertTrue(types.size() >= n);
--- a/components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
+++ b/components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 
+import ome.model.enums.EventType;
 import omero.RLong;
 import omero.RString;
 import omero.SecurityViolation;
@@ -2445,6 +2446,12 @@ public class LightAdminRolesTest extends RolesTests {
         return testCases.toArray(new Object[testCases.size()][]);
     }
 
+    /**
+     * Tests if an enumeration not used can be deleted.
+     * This should pass but it can be restored using the method
+     * <code>resetEnumerations</code>.
+     * @throws Exception
+     */
     @Test
     public void testDeleteEnumerationsbyRestrictedSystemUser() throws Exception {
         final IAdminPrx svc = root.getSession().getAdminService();
@@ -2461,14 +2468,21 @@ public class LightAdminRolesTest extends RolesTests {
         //try to delete an enum type. The type is not important
         final ITypesPrx types_svc = factory.getTypesService();
         List<IObject> types = types_svc.allEnumerations(ContrastMethod.class.getName());
+        int n = types.size();
+        Assert.assertTrue(n > 0);
         Iterator<IObject> i = types.iterator();
         while (i.hasNext()) {
             types_svc.deleteEnumeration(i.next());
         }
     }
 
-    @Test
-    public void testResetEnumerationsbyRestrictedSystemUser() throws Exception {
+    /**
+     * Tests if an enumeration already used can be deleted.
+     * This should fail.
+     * @throws Exception
+     */
+    @Test(expectedExceptions = omero.ValidationException.class)
+    public void testDeleteUsedEnumerationsbyRestrictedSystemUser() throws Exception {
         final IAdminPrx svc = root.getSession().getAdminService();
         final List<AdminPrivilege> expectedPrivileges = new ArrayList<AdminPrivilege>();
         final String lightAdminName = UUID.randomUUID().toString();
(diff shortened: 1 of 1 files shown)


mtbc:
> this should not be allowed

It does not violate any of the restrictions. This is an aspect of why a restricted admin is indeed best thought of as a kind of admin: they can do anything that is not explicitly restricted and why my initial instinct in implementing light admins was to instead add powers to normal users. (`WriteOwned` does not apply because nobody owns an enumeration.)

We could add `WriteEnum` and `DeleteEnum` restrictions if desired, that would probably not take very long. (I did initially experiment with broader data restrictions but ran into issues with creating events and suchlike. It should be fine if limited to enums though.)

jburel:
@mtbc not as serious as  I initially thought
A "delete" enum can be easily restored via the ``resetEnumerations`` method. 
Enum already used like ``EventType`` cannot be deleted.

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2437,4 +2447,160 @@ public void testModifyGroupEdit(boolean isPrivileged,
...
+     * This should pass but it can be restored using the method
+     * <code>resetEnumerations</code>.
+     * @throws Exception
+     */
+    @Test
+    public void testDeleteEnumerationsbyRestrictedSystemUser() throws Exception {
+        final IAdminPrx svc = root.getSession().getAdminService();
+        final List<AdminPrivilege> expectedPrivileges = new ArrayList<AdminPrivilege>();
+        final String lightAdminName = UUID.randomUUID().toString();
+        final String lightAdminPassword = UUID.randomUUID().toString();
```
mtbc: Would using an existing method like `loginNewAdmin` make this setup easier?

jburel: I started to write more tests that I initially anticipated. :-)
I will refactor that part since it starts to be too much "copy/paste"

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2437,4 +2447,160 @@ public void testModifyGroupEdit(boolean isPrivileged,
...
+        Iterator<IObject> i = types.iterator();
+        while (i.hasNext()) {
+            types_svc.deleteEnumeration(i.next());
+        }
+        types = types_svc.allEnumerations(ContrastMethod.class.getName());
+        Assert.assertEquals(types.size(), 0);
+        types_svc.resetEnumerations(ContrastMethod.class.getName());
+        types = types_svc.allEnumerations(ContrastMethod.class.getName());
+        Assert.assertEquals(types.size(), n);
+        types_svc.resetEnumerations(ContrastMethod.class.getName());
```
mtbc: This reset might also be good for `root` to do in an `@AfterClass` depending on in what state these test methods may have left the server.

jburel: True but here I wanted to check that any member of system could roll back potential mistakes and not only root

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2437,4 +2447,160 @@ public void testModifyGroupEdit(boolean isPrivileged,
...
+        final String lightAdminName = UUID.randomUUID().toString();
+        final String lightAdminPassword = UUID.randomUUID().toString();
+        Experimenter lightAdmin = createExperimenterI(lightAdminName, "test", "user");
+        final long lightAdminId = svc.createRestrictedSystemUserWithPassword(lightAdmin,
+                expectedPrivileges, omero.rtypes.rstring(lightAdminPassword));
+        lightAdmin = svc.getExperimenter(lightAdminId);
+        omero.client client = newOmeroClient();
+        client.createSession(lightAdminName, lightAdminPassword);
+        init(client);
+        //try to delete an enum type. The type is not important
```
mtbc: didn't try to delete it?

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -96,6 +106,22 @@
...
+        svc.resetEnumerations(ContrastMethod.class.getName());
+        super.tearDown();
+    }
+
+    /**
+     * Creates a new administrator without any privileges
+     * and create a new {@link omero.client}.
+     */
+    protected EventContext logNewAdminWithoutPrivileges() throws Exception {
+        return loginNewAdmin(true, new ArrayList<String>());
```
mtbc: `Collections.<String>emptyList()` suffices, no need for allocation

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2437,4 +2453,173 @@ public void testModifyGroupEdit(boolean isPrivileged,
...
+        final ITypesPrx ts = root.getSession().getTypesService();
+        List<IObject> types = ts.allEnumerations(ContrastMethod.class.getName());
+        int n = types.size();
+        ho = (ContrastMethod) ts.createEnumeration(ho);
+        types = ts.allEnumerations(ContrastMethod.class.getName());
+        Assert.assertEquals(types.size(), (n+1));
+        logNewAdminWithoutPrivileges();
+        final ITypesPrx types_svc = factory.getTypesService();
+        types = types_svc.allEnumerations(ContrastMethod.class.getName());
+        int m = types.size();
```
mtbc: not sure what the thinking is where where we use `m` sometimes and `n+1` othertimes

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2437,4 +2453,173 @@ public void testModifyGroupEdit(boolean isPrivileged,
...
+        final ITypesPrx types_svc = factory.getTypesService();
+        List<IObject> types = types_svc.allEnumerations(EventType.class.getName());
+        Iterator<IObject> i = types.iterator();
+        while (i.hasNext()) {
+            types_svc.deleteEnumeration(i.next());
+        }
+    }
+
+    /**
+     * Tests if a deleted enumeration can be restored.
```
mtbc: Does it?

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
jburel: n is the original size of the list before a new enum is added
m is the size of the list after adding an enum. that's why we have check m and n+1

mtbc:
@pwalczysko: You may want to look through this as you know the test class well.

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
pwalczysko: I am probably a bit confused too here, just like @mtbc . I read this test as a sequence:
   - root creates an enum, raising the number of enums from n to n+1
   - light admin double-checks the number of enums, and for them it is m
   - m is equal to n+1
   - light admin resets the enums (this is probably the restoring step)
   - the size of the enums is m after the restoration, which is not surprising

I am missing the [deletion step](https://github.com/openmicroscopy/openmicroscopy/pull/5502/files#diff-d91232ef1e245a87e311c4a79a2a5f77R2484) which would precede the restoration step ?

jburel: The test has now been reviewed.

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2437,4 +2453,182 @@ public void testModifyGroupEdit(boolean isPrivileged,
...
+        //original number of enumerations
+        int n = types.size();
+        ho = (ContrastMethod) types_svc.createEnumeration(ho);
+        types = types_svc.allEnumerations(ContrastMethod.class.getName());
+        //check that an enumeration has been added
+        Assert.assertEquals(types.size(), (n+1));
+        types_svc.deleteEnumeration(ho);
+        types_svc.resetEnumerations(ContrastMethod.class.getName());
+        types = types_svc.allEnumerations(ContrastMethod.class.getName());
+        Assert.assertEquals(types.size(), n);
```
pwalczysko: And the size after delete + restore is again "n" ? I would expect "n+1" if the reset really restores the deleted enum, as indicated in the comment https://github.com/openmicroscopy/openmicroscopy/pull/5502/files#diff-d91232ef1e245a87e311c4a79a2a5f77R2459 in testDeleteEnum....

jburel: it resets to the original as far as I know so it should be n
I might be wrong

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2437,4 +2453,182 @@ public void testModifyGroupEdit(boolean isPrivileged,
...
+                        found = true;
+                    }
+                }
+            }
+        }
+        Assert.assertTrue(found);
+    }
+
+    /**
+     * Tests f it possible to modify a script using the upload method from
```
pwalczysko: "Tests f" -> "Tests if it is"  (missing is, missing if)

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
pwalczysko: No, you are right, the test passes for me locally as it is. Probably still needs some redesign as discussed.

jburel:
@pwalczysko your comments have been addressed

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2576,7 +2584,7 @@ public void testIndexObjectbyRestrictedSystemUser() throws Exception {
     }
 
     /**
-     * Tests f it possible to modify a script using the upload method from
+     * Tests if it possible to modify a script using the upload method from
```
pwalczysko: Sorry, still missing "is" (or any verb in the sentence).
Suggestion 1:
``Test if it is possible to modify a script using the upload method from omero.client.``
Suggestion 2:
``Test if a script can be modified using the upload method from omero.client.``

pwalczysko: Suggestion 3: 
``Test that a script cannot be modified using the upload method from omero.client.``

pwalczysko:
Yes, the last change https://github.com/openmicroscopy/openmicroscopy/pull/5502/commits/8a7401295b5fcdc49d10ea0d721c4743127187f5 makes perfect sense. 

Found one persisting typo https://github.com/openmicroscopy/openmicroscopy/pull/5502#discussion_r139913284.

Will checkout your branch and run the tests locally now.

pwalczysko:
The line https://github.com/openmicroscopy/openmicroscopy/pull/5502/commits/8a7401295b5fcdc49d10ea0d721c4743127187f5#diff-d91232ef1e245a87e311c4a79a2a5f77L2529 fails with.
```
FAILED: testResetEnumerationsbyRestrictedSystemUser
java.lang.AssertionError: expected [11] but found [10]
	at org.testng.Assert.fail(Assert.java:94)
	at org.testng.Assert.failNotEquals(Assert.java:494)
	at org.testng.Assert.assertEquals(Assert.java:123)
	at org.testng.Assert.assertEquals(Assert.java:370)
	at org.testng.Assert.assertEquals(Assert.java:380)
	at integration.LightAdminRolesTest.testResetEnumerationsbyRestrictedSystemUser(LightAdminRolesTest.java:2537)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:80)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:714)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:901)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1231)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:127)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:111)
	at org.testng.TestRunner.privateRun(TestRunner.java:767)
	at org.testng.TestRunner.run(TestRunner.java:617)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:334)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:329)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:291)
	at org.testng.SuiteRunner.run(SuiteRunner.java:240)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1198)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1123)
	at org.testng.TestNG.run(TestNG.java:1031)
	at org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:126)
	at org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:137)
	at org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:58)
```

pwalczysko:
The [problem](https://github.com/openmicroscopy/openmicroscopy/pull/5502#issuecomment-330795107) is in the order in which one runs your tests. If the ``testResetEnumerationsbyRestrictedSystemUser`` is run repeatedly, all the repeats pass.
If ``testResetEnumerationsbyRestrictedSystemUser`` is run just after ``testUpdateEnumerationsbyRestrictedSystemUser()``, then the ``testResetEnumerationsbyRestrictedSystemUser`` fails as indicated in my [previous comment](https://github.com/openmicroscopy/openmicroscopy/pull/5502#issuecomment-330795107)

jburel:
I need to use a different enum for that test so it does not clash with the other tests

pwalczysko:
In summary, one [bug](https://github.com/openmicroscopy/openmicroscopy/pull/5502#issuecomment-330795107) and one [typo](https://github.com/openmicroscopy/openmicroscopy/pull/5502#discussion_r139913284). All the other tests were passing.

mtbc:
Maybe `testResetEnumerationsbyRestrictedSystemUser` needs to work differently: as it is it assumes that no previous test inserts a new enum. Perhaps it even suffices just to test that the number increased?

pwalczysko:
Unfortunately even after addition of https://github.com/openmicroscopy/openmicroscopy/pull/5502/commits/62433c28e7de4c442c9269b32953965b1163fb48, the ``testResetEnumerationsbyRestrictedSystemUser`` still fails when run after ``testUpdateEnumerationsbyRestrictedSystemUser``.
What is happening is:
``testUpdateEnumerationsbyRestrictedSystemUser`` adds an enum, raising the ``n`` from 10 to 11
``testResetEnumerationsbyRestrictedSystemUser`` reports the n to be all the time 11, before delete, after delete, and after reset, the n is all the time 11
```testResetEnumerationsbyRestrictedSystemUser`` types.size() is after delete 0, but after reset only 10 (not 11)
the test fails in such case. This shows that there is some immutable number of enums (10 in my case) to which the reset method always sets things back. If one additional enum is added by another test, the reset method will not restore it.

**Comment by jburel** [comment]
thanks for the investigation @pwalczysko 
Easiest option will be to use different enumeration type to avoid unexpected side effect

**Comment by pwalczysko** [comment]
@jburel : What if in ``testResetEnumerationsbyRestrictedSystemUser`` you first reset, then count (and put it to n), then delete, then reset again, then assert that types.size() is equal to n ?

**Comment by pwalczysko** [comment]
>Easiest option will be to use different enumeration type to avoid unexpected side effect

aha, good, this seems cleaner than my suggestion

**Comment by mtbc** [comment]
Future-proofing is indeed a challenge: want to delete an enumeration value that,

* nothing is using, except some future test might
* reset would restore, which means some new unique one won't do.

I can't suggest anything better off-hand I'm afraid.

**Comment by jburel** [comment]
I think all types of enumerations are used. 
the count in place is already not future-proofed,
I put it there as safe-guard so the test fails if we use all enum of a given type

File: components/tools/OmeroJava/test/integration/LightAdminRolesTest.java [hunk 10]
```diff
@@ -2530,11 +2530,11 @@ public void testResetEnumerationsbyRestrictedSystemUser() throws Exception {
...
         Assert.assertTrue(count > 0);
-        types = types_svc.allEnumerations(Detector.class.getName());
+        types = types_svc.allEnumerations(DetectorType.class.getName());
         Assert.assertEquals(types.size(), (n-count));
         //reset the deleted enumerations
-        types_svc.resetEnumerations(Detector.class.getName());
-        types = types_svc.allEnumerations(Detector.class.getName());
+        types_svc.resetEnumerations(DetectorType.class.getName());
+        types = types_svc.allEnumerations(DetectorType.class.getName());
         //We should be back to at the original list. Other enum might have been
```
pwalczysko: ``We should be back to at the original list`` -> either ``at`` or ``to`` - I think both ``at`` and ``to`` is not correct - @mtbc ?

pwalczysko:
Looks good now for the test passing - checked locally, the problem with the failing in different order of tests run is solved. 

Typo https://github.com/openmicroscopy/openmicroscopy/pull/5502#pullrequestreview-63909519 is persisting. I think https://github.com/openmicroscopy/openmicroscopy/pull/5502#discussion_r140203726 needs some attention too, but these are just typos.

Hopefully CI will be happy too.

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
mtbc: Indeed.

jburel:
@pwalczysko text should now be fixed

components/tools/OmeroJava/test/integration/LightAdminRolesTest.java
```diff
@@ -2616,8 +2615,8 @@ public void testModifyScriptUsingUploadFromClientbyRestrictedSystemUser() throws
     }
 
     /**
-     * Tests if it possible to upload a file owned by another user using
-     * the upload method from omero.client
+     * Tests if a file owned by another usert can be uploaded using the upload
```
pwalczysko: ``usert`` -> ``user`` (one ``t`` too much)

pwalczysko:
the typo fix is okay, the tests were skipped in the CI build though

pwalczysko:
All the six new tests are passing here. The other tests are passing too. https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/LightAdminRolesTest/

Ready to merge.

jburel:
Merging so I can review the other tests elsewhere

PR #5507, opened 2017-09-14
core_html_not_for_right_panel

will-moore:
# What this PR does

As noted in https://trello.com/c/BQyJ0Ay7/47-django-templates-cleanup, the "batch annotate" and "shares" in right panel are extending the ```webgateway/core_html.html``` template, so that anything added there such as base_template include from #5463 is loaded in the right panel.

Although this doesn't "break" anything, it's probably not expected. E.g. for Google analytics, we don't want to count "hits" when the right panel is loaded (although we will get hits when the image viewer is opened etc).

# Testing this PR

1. Quick check that right panel looks OK and functionality is not broken (E.g. add annotations etc).
2. Since Google analytics is deployed on eel, can check that script is not loaded for "batch annotate" panel when multiple objects are selected, or when "share" is loaded in right panel in Shares page.
2. Open dev-tools (e.g. by right-click on page -> Inspect element) then choose Network tab and JS to only show scripts.
3. Shouldn't see script loaded when right panel loads. Currently we see:

screen shot 2017-09-14 at 11 54 43

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/annotations_share.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/annotations_share.html
@@ -1,4 +1,3 @@
-{% extends "webgateway/core_html.html" %}
 {% load i18n %}
 {% load common_tags %}
 {% load common_filters %}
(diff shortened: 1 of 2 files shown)


joshmoore:
Does this mean that previously we were getting 2 hits for batch/shares and now 1, or now zero? What's the functional reason these inherited from core previously?

will-moore:
We should now be getting 1 'hit' when the main page loads and no additional hits when the right panel loads.
I don't know why those templates were extending core_html when all the others for the right panel weren't. I don't see any functional difference.

jburel:
Anything else to investigate?

will-moore:
I think this still needs testing as described above.

dominikl:
screen shot 2017-09-20 at 10 53 08

Seems to work, ``analytics.js`` only loaded once. Didn't notice any problems with the right  panel either. Good to merge 👍

PR #5510, opened 2017-09-15
Metadata plugin testtables for robot

will-moore:
# What this PR does

This attempts to prevent failure of the ```robot_setup.py``` script when tables are not available
or working correctly, since this means that NO robot tests are run.

Adds a new command ```testtables``` to the metadata plugin.
Can be used like this to only run ```populate``` if ```testtables``` doesn't fail.

```bin/omero metadata testtables && bin/omero metadata populate...```

# Testing this PR

1. Shouldn't see complete failures of robot tests with setup failing due to tables. E.g. today: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/712/

1. Instead, robot tests that actually test the tables functionality will fail, in the same way that they do for integration tests.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/metadata.py
+++ b/components/tools/OmeroPy/src/omero/plugins/metadata.py
@@ -143,7 +143,7 @@ class MetadataControl(BaseControl):
         bulkanns = parser.add(sub, self.bulkanns)
         mapanns = parser.add(sub, self.mapanns)
         allanns = parser.add(sub, self.allanns)
-        testtables = parser.add(sub, self.testtables)
+        parser.add(sub, self.testtables)
         rois = parser.add(sub, self.rois)
         populate = parser.add(sub, self.populate)
         populateroi = parser.add(sub, self.populateroi)
--- a/components/tools/OmeroPy/src/omero/plugins/metadata.py
+++ b/components/tools/OmeroPy/src/omero/plugins/metadata.py
@@ -426,7 +426,6 @@ class MetadataControl(BaseControl):
         if not initialized:
             self.ctx.die(100, "Failed to initialize Table")
 
-
     # WRITE
 
     def populate(self, args):
--- a/components/tools/OmeroPy/src/omero/plugins/metadata.py
+++ b/components/tools/OmeroPy/src/omero/plugins/metadata.py
@@ -143,6 +143,7 @@ class MetadataControl(BaseControl):
         bulkanns = parser.add(sub, self.bulkanns)
         mapanns = parser.add(sub, self.mapanns)
         allanns = parser.add(sub, self.allanns)
+        testtables = parser.add(sub, self.testtables)
         rois = parser.add(sub, self.rois)
         populate = parser.add(sub, self.populate)
         populateroi = parser.add(sub, self.populateroi)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by joshmoore** [comment]
NB: This reminds me that we need to look in the behavior of trying to override one of the shipped plugins by using `pip`.

**Comment by mtbc** [comment]
If we manage to create the table, do we want to delete it again before we finish?

**Comment by bramalingam** [comment]
@mtbc : If you are suggesting a delete workflow to keep the database as clean as possible, I have seen it being useful in cases where the viewport gets cluttered (like creating new projects,datasets and screens) during execution of a few robot tests, and this has led to flakiness and failures of other Robot tests. But I don't the think the annotations will have the same effect (when deleted, in the robot tests context), but if this is intended as a delete test for the tables, then it sounds like a good idea.

**Comment by mtbc** [comment]
This adds a new command to the metadata plugin so it may well be used by external users' clients well outside the context of our test infrastructure.

**Comment by rgozim** [comment]
I just tested the Develop branch using the robot_setup.sh mock data uploading. It fails and never completes. 

With @will-moore's PR, the upload error doesn't stop the entire operation and completes successfully. 

However, any failed uploads do not get added to the console output so any occurring issues could be missed.

File: components/tools/OmeroPy/src/omero/plugins/metadata.py [hunk 1]
```diff
@@ -395,6 +396,20 @@ def get_anns(md):
             indent = 0
         self._output_ann(md, get_anns, args.parents, indent)
 
+    def testtables(self, args):
+        "Tests whether tables can be created and initialized"
+        client = self.ctx.conn(args)
+
+        sf = client.getSession()
+        sr = sf.sharedResources()
+        table = sr.newTable(1, 'testtables')
```
jburel: table is never closed

will-moore:
Today it seems that tables failed but setup script was able to complete and all tests ran, with failures for tables tests: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/718/robot/ as expected.

components/tools/OmeroPy/src/omero/plugins/metadata.py
joshmoore: Agreed. `TablePrx` instances should be closed in try/finally blocks and in this case a further finally should delete the file that was created.

joshmoore (changes requested):
Generally :+1: for making this more user friendly. I do wonder if the method would live better somewhere else. i.e. when we extract the `metadata.py` plugin for pip-installation, should we consider a `tables.py` plugin as well, which could help with creating, finding, etc. tables independent of their use for populating metadata, or is that too much?

will-moore:
After ```table = sr.newTable(1, 'testtables')``` if ```table``` is ```None```, then presumably we don't need to call close or delete the file?
Testing my commit above, locally I have ```table.initialize([])``` failing and the tables are subsequently getting deleted OK.

joshmoore:
> After table = sr.newTable(1, 'testtables') if table is None, then presumably we don't need to call close or delete the file?

Correct.

will-moore:
I think a "tables" plugin is a good idea. Certainly if you were looking to work with tables via CLI you wouldn't think to look under "metadata".
But probably not in this PR.

joshmoore:
> I think a "tables" plugin is a good idea. Certainly if you were looking to work with tables via CLI you wouldn't think to look under "metadata".

Agreed. Another benefit might be that we could choose to package `omero/plugins/tables.py` not just as `omero-cli-tables` but as `omero-tables` which would bring in PyTables dependencies. Or getting even wackier, as different implementations of the tables API to get around some of the limitations of HDF5.

> But probably not in this PR.

Definitely not. Sorry for dragging my feet here, it's just been difficult trying to figure out where place to put the CLI plugins.

Merging under the assumption that this is still in development-mode and we can come back to the questions during the migration.

joshmoore (approved):

PR #5513, opened 2017-09-18
Java gateway bugfixes

dominikl:
# What this PR does
- Guest login was used to determine if a username is actually a session id. With this PR a `joinSession` attempt is made, if the username matches UUID format.
- The `Facility.handleException()` method silently 'swallowed' exceptions caused by connection issues. With this PR a `DSOutofServiceException` is thrown in these cases.
- Adds `JoinSessionCredentials` to provide a specific class for session id logins.
- Automatically remove stale connectors from the groupConnectorMap. That way the Gateway can continue working after a server restart without explicit dis- and reconnect.
- Makes the `Gateway` `AutoCloseable` so that it can be used in try-with-resources blocks.
- In various Facility methods, check arguments and return immediately when arguments not reasonable (Ids not provided, etc.).

# Testing this PR

- Check Java gateway integration tests.
- Test that login via session id still works
- Test that reconnect after connection loss still works. E. g. open Insight. Disconnect network. Try to browse a dataset. A connection warning should appear after a while. Reconnect the network. Insight should reconnect again, too, without crashing.
- Continuously request some data from the server using the same Gateway connection (catch all exceptions). Stop and restart the server in between. While the server is not running you will get some exceptions. But when the server is ready again the Gateway should continue working as usual. See below for an example test client (*).

# Related reading

- https://trello.com/c/XKg3FisH/42-java-gateway
- https://trello.com/c/JF3VjeWk/694-refactor-logincredentials
- Context for `Facility.handleException()` and server restart issue:  https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8367

--

*) Example client to use for testing server restart
```
LoginCredentials l = new LoginCredentials( ... );
Gateway g = new Gateway(new SimpleLogger());

ExperimenterData exp = g.connect(l);
SecurityContext ctx = new SecurityContext(exp.getGroupId());

while (true) {
    try {
        IObject obj = g.getFacility(BrowseFacility.class).findIObject(
                ctx, Image.class.getSimpleName(), 1);
        System.out.println(obj.getClass().getSimpleName() + " "
                + obj.getId().getValue());
    } catch (Throwable e) {
        e.printStackTrace();
    } finally {
        try {
            Thread.sleep(5000);
        } catch (InterruptedException e) {
        }
    }
}
```

### DIFF
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.ResourceBundle;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -140,9 +141,6 @@ public class Gateway {
 
     /** Property to indicate that a stateless service has been created */
     public static final String PROP_STATELESS_SERVICE_CREATED = "PROP_STATELESS_SERVICE_CREATED";
-
-    /** Regex matching ICE Session ID */
-    public static final String ICE_SESSION_ID_REGEX = "[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}";
     
     /** Reference to a {@link Logger} */
     private Logger log;
@@ -1107,7 +1105,12 @@ public class Gateway {
      *         otherwise.
      */
     private boolean isSessionID(String s) {
-        return s != null && s.matches(ICE_SESSION_ID_REGEX);
+        try {
+            UUID.fromString(s);
+        } catch (Exception e) {
+            return false;
+        }
+        return true;
     }
     
     /**
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -142,7 +142,7 @@ public class Gateway {
     public static final String PROP_STATELESS_SERVICE_CREATED = "PROP_STATELESS_SERVICE_CREATED";
 
     /** Regex matching ICE Session ID */
-    private static final String ICE_SESSION_ID_REGEX = "[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}";
+    public static final String ICE_SESSION_ID_REGEX = "[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}";
     
     /** Reference to a {@link Logger} */
     private Logger log;
@@ -402,15 +402,15 @@ public class Gateway {
      * @param user
      *            The user to get the session ID for
      * @return See above
+     * @throws DSOutOfServiceException
+     *             If the connection is broken, or not logged in
      */
-    public String getSessionId(ExperimenterData user) {
-        try {
-            Connector c = getConnector(new SecurityContext(user.getGroupId()),
-                    false, false);
-            if (c != null) {
-                return c.getClient().getSessionId();
-            }
-        } catch (DSOutOfServiceException e) {
+    public String getSessionId(ExperimenterData user)
+            throws DSOutOfServiceException {
+        Connector c = getConnector(new SecurityContext(user.getGroupId()),
+                false, false);
+        if (c != null) {
+            return c.getClient().getSessionId();
         }
         return null;
     }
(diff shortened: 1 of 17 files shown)


components/blitz/src/omero/gateway/Gateway.java
```diff
@@ -142,7 +142,7 @@
     public static final String PROP_STATELESS_SERVICE_CREATED = "PROP_STATELESS_SERVICE_CREATED";
 
     /** Regex matching ICE Session ID */
-    private static final String ICE_SESSION_ID_REGEX = "[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}";
+    public static final String ICE_SESSION_ID_REGEX = "[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}";
```
joshmoore: Interesting. Might be worth moving this to an Ice constant and using this server-side as well. To data, I've always treated the session IDs as opaque, i.e. no assumptions could be made.

components/tools/OmeroJava/test/integration/gateway/GatewayUsageTest.java
```diff
@@ -73,6 +74,31 @@ public void testLoginWithArgs()
...
+        String sessionId = gw.getSessionId(root);
+
+        Gateway gw2 = new Gateway(new SimpleLogger());
+        JoinSessionCredentials c2 = new JoinSessionCredentials(sessionId,
+                client.getProperty("omero.host"), Integer.parseInt(client
+                        .getProperty("omero.port")));
+        ExperimenterData root2 = gw2.connect(c2);
+        Assert.assertNotNull(root2);
+        Assert.assertEquals(gw2.getSessionId(root2), sessionId);
+        gw2.disconnect();
```
joshmoore: If something throws re: `gw2`, `gw1` will not be closed which could hang the test.

components/blitz/src/omero/gateway/Gateway.java
dominikl: The documentation says that each Ice object has an UUID and so does the session ID look like, although I didn't find anything in the documentation which explicitely says that the session ID is an UUID, too.

dominikl: Hm, actually I could also just use Java's builtin `UUID.fromString()` method.

joshmoore: > although I didn't find anything in the documentation which explicitely says that the session ID is an UUID, too.

Exactly. It's not *yet* part of the contract.

components/tools/OmeroJava/test/integration/gateway/GatewayUsageTest.java
dominikl: Could make the Gateway AutoCloseable ( c31e719 ), would make things easier. Any objections?

dominikl:
Just for reference, already noticed the problem with the 'swallowed' exceptions in 5.2 https://github.com/openmicroscopy/openmicroscopy/pull/4344#issuecomment-161947619 and Josh's idea of an 'ExceptionHandler', but forgot about it again. General exception handling overhaul would be a breaking change, but probably necessary, maybe for 5.5 then...

components/blitz/src/omero/gateway/facility/BrowseFacility.java
```diff
@@ -130,7 +130,7 @@
     public Collection<DataObject> getHierarchy(SecurityContext ctx, Class rootType,
             List<Long> rootIDs, Parameters options)
             throws DSOutOfServiceException, DSAccessException {
-        if (rootType == null || CollectionUtils.isEmpty(rootIDs))
+        if (rootType == null)
```
mtbc: The `rootIDs.isEmpty()` case will go okay?

dominikl: Yes, same behaviour as roodIDs == null (get all).

mtbc: Aha, maybe also mention in Javadoc, just add "or empty" or something.

components/tools/OmeroJava/test/integration/gateway/GatewayUsageTest.java
mtbc: No objections. :smiley:

dominikl:
Not quite sure, ready to merge or relist for standup?

mtbc:
Has it yet had any functional testing?

dominikl:
Nope, doesn't look like it. I'll relist and maybe @pwalczysko can have a look when he can spare some time.

pwalczysko:
All the four bullet points from the "Testing this PR" header passed.
Ready to merge FMPOV

joshmoore:
Thanks, @dominikl. At some point, it would be interesting to review your changes here against the Python Gateway (i.e. properly specifying the string format for the session ids, etc.) Merging into the `clisplit` branch for further testing.

PR #5515, opened 2017-09-19
Search plugin

jburel:
# What this PR does

Only admin can index object
This PR adds a check in the search method if ``index`` is true

# Testing this PR
Check that the new test is green

# Related reading
Noticed while working on https://trello.com/c/FjayBnvM/39-gateway-support-for-roles

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/search.py
+++ b/components/tools/OmeroPy/src/omero/plugins/search.py
@@ -95,7 +95,7 @@ class SearchControl(HqlControl):
 
         if args.index:
             if not self.ctx.get_event_context().isAdmin:
-                raise Exception("Only admin can index object")
+                self.ctx.die(432, "Only admin can index object")
 
             try:
                 parts = args.type.split(":")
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/src/omero/plugins/search.py
```diff
@@ -94,6 +94,9 @@ def search(self, args):
         import omero.all
 
         if args.index:
+            if not self.ctx.get_event_context().isAdmin:
+                raise Exception("Only admin can index object")
```
joshmoore: Sorry, I was rushing earlier in my pseudo-code. To match the style of the plugins, this should use `self.ctx.die`

jburel: ok I will adjust accordingly

jburel:
@joshmoore adjusted

joshmoore:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/691/testReport/OmeroPy.test.integration.clitest.test_search/TestSearch/test_search_index_by_user/ was green this morning. :+1:

PR #5518, opened 2017-09-25
s/his.size/this.size

kennethgillen:
This PR fixes what looks like a typo.

It certainly looks like a value wasn't being assigned that was intended to be.

# Testing this PR

1. required setup
Ability to build OMERO.web, or manually edit the `.js` file.

2. actions to perform
Identify the expected behaviour.

3. expected observations
Not sure - I'd check with @will-moore and @rgozim.

# Related reading

n/a

n/a

Not sure - viewport size? Again I'd check with @will-moore and @rgozim.

Perhaps this fix should just be included in someone else's open PR if they understand what this code's doing.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.viewport.js
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.viewport.js
@@ -73,7 +73,7 @@ var Metadata = function () {
       if (this.rdefs.invertAxis) {
         var t = this.size.t;
         this.size.t = this.size.z;
-        his.size.z = t;
+        this.size.z = t;
       }
     this.current.zoom = 100;
   };


components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.viewport.js
```diff
@@ -73,7 +73,7 @@ var Metadata = function () {
       if (this.rdefs.invertAxis) {
         var t = this.size.t;
         this.size.t = this.size.z;
-        his.size.z = t;
+        this.size.z = t;
```
jburel, line 76: good catch

will-moore:
Looks good, thanks.
I think we've never seen this bug because we don't use the ```invertAxis``` functionality now.

PR #5522, opened 2017-09-28
Ignore unhandled annotation targets

dominikl:
# What this PR does

If an ome.tiff contains annotations of e. g. PlaneInfo objects the import crashes with a RuntimeException. With this PR no exception will be thrown, just a warning logged (like it's already done if the annotation target is null, see line 83).

# Testing this PR

Import the file from https://www.openmicroscopy.org/qa2/qa/feedback/17938/ 

# Related reading

https://trello.com/c/1sX3tep3/62-bug-creating-annotations-on-plane

### DIFF
--- a/components/blitz/src/ome/formats/model/ReferenceProcessor.java
+++ b/components/blitz/src/ome/formats/model/ReferenceProcessor.java
@@ -114,8 +114,8 @@ public class ReferenceProcessor implements ModelProcessor {
               indexes.put(Index.IMAGE_INDEX, indexArray[0]);
               indexes.put(Index.CHANNEL_INDEX, indexArray[1]);
             } else {
-                log.warn("Ignoring target class " + target+", not handled yet.");
-                continue;
+              log.warn("Ignoring target class " + target + ", not handled yet.");
+              continue;
             }
             container = store.getIObjectContainer(targetClass, indexes);
           }
--- a/components/blitz/src/ome/formats/model/ReferenceProcessor.java
+++ b/components/blitz/src/ome/formats/model/ReferenceProcessor.java
@@ -114,16 +114,16 @@ public class ReferenceProcessor implements ModelProcessor {
               indexes.put(Index.IMAGE_INDEX, indexArray[0]);
               indexes.put(Index.CHANNEL_INDEX, indexArray[1]);
             } else {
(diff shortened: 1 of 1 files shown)


components/blitz/src/ome/formats/model/ReferenceProcessor.java
```diff
@@ -114,16 +114,16 @@ public void process(IObjectContainerStore store) throws ModelException {
               indexes.put(Index.IMAGE_INDEX, indexArray[0]);
               indexes.put(Index.CHANNEL_INDEX, indexArray[1]);
             } else {
-              throw new RuntimeException(
-                  String.format("Unable to synchronize reference %s --> %s",
-                                reference, target));
+                log.warn("Ignoring target class " + target+", not handled yet.");
```
mtbc: spaces around `+`

rleigh-codelibre:
Looks fine to me, and importing the image shows success with the warnings in the import log.

dominikl:
Thanks @rleigh-codelibre . Just also fixed the formatting issue noticed by @mtbc (no re-testing needed imho).

chris-allan (changes requested):
The reason why this exception was originally thrown was to catch logic errors in Bio-Formats readers. Especially cases where mappings were missing or needed to be written. Fundamentally, I would not remove it.

Has the root cause of the original error been located?

dominikl:
Ok, then I close this PR again. So basically the file is invalid? Not sure if that's been made very clear in @rleigh-codelibre 's forum reply: https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8371

rleigh-codelibre:
From the point of view of annotations and annotation references, the file is valid and can be processed to add these annotations and references to the metadata store.  OMERO isn't handling these valid references.  There are some classes which are annotatable in the data model which aren't annotatable with OMERO.

The example in the forum had some odd (but valid) *content* inside the annotation, but the annotations and annotation references themselves were fine.

chris-allan:
In cases of mismatch between the models we have to do the work to either perform the correct mapping or ignore cases where we have references to objects that are not annotatable in OMERO. That's a fine solution to this issue that doesn't result in removing a safety check from the reference handling.

PR #5524, opened 2017-09-29
stabilize integration tests

mtbc:
# What this PR does

Rewrites Java integration tests to behave the same but without creating so many users and service proxies. Also adjusts usage of Ice implicit context.

# Testing this PR

[OMERO-DEV-merge-integration-java](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/) should pass. Code changes should not invalidate any tests. Specifically, while a `ImplicitGroupContext` is open then `this.client` should not change session/login. It may be easiest to ignore whitespace changes via https://github.com/openmicroscopy/openmicroscopy/pull/5524/files?w=1.

### DIFF
--- a/components/tools/OmeroJava/test/integration/AbstractServerTest.java
+++ b/components/tools/OmeroJava/test/integration/AbstractServerTest.java
@@ -2369,7 +2369,17 @@ public class AbstractServerTest extends AbstractTest {
         return provideEveryBooleanCombination(4);
     }
 
+    /**
+     * Override the Ice implicit call context with a group ID.
+     * Removes the override upon closing.
+     * @author m.t.b.carroll@ixod.org
+     * @since 5.4.0
+     */
     protected class ImplicitGroupContext implements AutoCloseable {
+        /**
+         * Set the implicit group context to the given group.
+         * @param groupId a group ID
+         */
         ImplicitGroupContext(long groupId) {
             if (client.getImplicitContext().containsKey(Login.OMERO_GROUP)) {
                 throw new IllegalStateException("group context already set");
@@ -2377,8 +2387,12 @@ public class AbstractServerTest extends AbstractTest {
             client.getImplicitContext().put(Login.OMERO_GROUP, Long.toString(groupId));
         }
 
-        ImplicitGroupContext(RLong id) {
-            this(id.getValue());
+        /**
+         * Set the implicit group context.
+         * @param groupId a group ID
+         */
+        ImplicitGroupContext(RLong groupId) {
+            this(groupId.getValue());
         }
 
         @Override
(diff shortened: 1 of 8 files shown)


components/tools/OmeroJava/test/integration/ManagedRepositoryTest.java
```diff
@@ -95,7 +95,7 @@
 
     @BeforeMethod
     public void setupNewUser() throws Exception {
-        newUserAndGroup("rw----");
+        //newUserAndGroup("rw----");
```
joshmoore: I assume this is a temporary commenting out?

joshmoore:
In general, :+1:. Only one discussion that might be worth having is how this `ImplicitContext` strategy merges with that used in `omero/gateway/__init__.py`. I.e. is there any general policy that needs reviewing unifying? Or, it may just be that Java and Python vary.

components/tools/OmeroJava/test/integration/ManagedRepositoryTest.java
jburel: sorry that was my initial commit while testing
@mtbc could you remove the line?
This is not needed

components/tools/OmeroJava/test/integration/AbstractServerTest.java
```diff
@@ -1306,6 +1330,9 @@ protected void assertNoneExist(String className, Collection<Long> ids) throws Se
     protected List<Pixels> importFile(OMEROMetadataStoreClient importer,
             File file, String format, boolean metadata, IObject target)
             throws Throwable {
+        if (importer == null) {
```
dominikl, line 1318: Is this necessary? `createImporter()` will be called in line 1354 anyway.

components/tools/OmeroJava/test/integration/ManagedRepositoryTest.java
mtbc: Sure, though I'll hold off for now as you may also have other commits arising from comments.

components/tools/OmeroJava/test/integration/AbstractServerTest.java
jburel: actually line 1354 ``importer`` should be used instead ``createImporter`` since an importer created by another class might be specified.
This was done while trying to narrow down the issue
@mtbc could you change it?

mtbc: I think it's likely I'll guess wrongly about what change is needed. Could you give me something to cherry-pick or just tell me exactly what which line should read instead and then I'll do it?

mtbc: Ohh, I think I understand now, I can do that.

jburel:
@joshmoore we have seen similar error ``user x in group y`` in python (to be investigated in another PR). It could be due to the usage of the ``implicitContext``

joshmoore:
> we have seen similar error user x in group y in python

Understood. The only caveat there is that use of that group is public contract for BlitzGateway in Python.

Changes here all look good to me.

mtbc:
@joshmoore: Instead of `AbstractServerTest.client.getImplicitContext().put...` ought I simply be using `AbstractServerTest.factory.setSecurityContext`? From the docs I wasn't sure that I may use any group of which I am a member.

joshmoore:
The former changes it for the client-side `omero.client` instance the latter for the server-side `Session` instance. But for a single client with a single thread the results would be the same.

mtbc:
Interesting! I should check what happens if I do both at once inconsistently. :smiley:

PR #5525, opened 2017-09-29
Insight Java 9 compatibility 

dominikl:
--exclude not for 5.4.0 and not ready yet anyway (OSX build still has to be adjusted)

# What this PR does

For Java 9 Insight has to be launched with `--add-modules java.activation` argument.
This PR modifies the Linux launch scripts, as well as Windows and Mac builds accordingly.

Note: This might need an `exe4j` update on the build system. I had to update my local `exe4j` because the `<vmoptions>` attribute wasn't supported by the version I had before (no exception, just ignored).

# Testing this PR

A bit tricky to test: Ideally test that Importer and Insight still work with Java 1.8, then test again if both still work with Java 9, on Linux, OSX and Windows.

# Related reading

https://trello.com/c/Eh7p5387/57-to-watch-java-19

### DIFF
--- a/components/insight/build/Importer_template.exe4j
+++ b/components/insight/build/Importer_template.exe4j
@@ -27,7 +27,7 @@
         fontSize="8" fontColor="0,0,0" fontWeight="500" />
     </text>
   </splashScreen>
-  <java mainClass="@MAIN_CLASS@" vmParameters="@VM_PARAMETERS@"
+  <java mainClass="@MAIN_CLASS@"
     arguments="containerImporter.xml" allowVMPassthroughParameters="true" preferredVM=""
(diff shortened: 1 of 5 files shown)


rleigh-codelibre:
I haven't tried this yet, but would using `module-info.java` with `requires java.activation;` make this work transparently, without needing to add anything to the launch scripts?

dominikl:
Thought about that too initially. But that would mean you have to build Insight with Java 9 in first place. And I guess we're not going to update the build systems to Java 9 soon, are we?

jburel:
not on top of our todo list at the moment.

dominikl:
Just realised, much easier is to simply add `javax.activation` as dependency to ivy. Will open a PR for that later after 5.4.0 release.

dominikl:
Replaced by #5534

PR #5526, opened 2017-10-04
remove password from guest user

mtbc:
# What this PR does

Provides a utility script for removing any password for the guest user.

# Testing this PR

1. Check that you can log in as the guest user with any password.

1. Use `bin/omero user password guest` to set a password on the guest user.

1. Check that you *cannot* log in as the guest user with any password but the one you set.

1. Use `psql` with the `allow-guest-user-without-password.sql` script to remove the password from the guest user.

1. Check that you can log in as the guest user with any password.

1. Check that you still need the correct password to log in as other users.


dominikl:
Works like expected 👍

PR #5528, opened 2017-10-04
Merge 5.3.5 release into metadata53

manics:
# What this PR does

Merge 5.3.5 release into metadata53.

This seems larger than expected, so maybe we should drop it.


manics:
This hasn't had a proper round of testing so should not be deployed to production

sbesson:
Closing for now as this is not in scope for the upcoming IDR 0.4.5 release and work will start separately into looking at porting the IDR work to 5.4.

sbesson:
Note this PR was [reopened](https://github.com/openmicroscopy/openmicroscopy/pull/5528#event-1443589462) to confirm the fix included #5638 on the testing server since this PR had been previously included and the DB has been patched to `5.3__1`. In the context of the next [production IDR](https://trello.com/b/3hRu3hfw/idr-045) release, given the other priorities and availabilities, I do not feel confident this was included and tested enough to include in the next-production server. Closing this and we can re-evaluate whether we want to invest in bringing the changes from 5.3.5 for one of the upcoming releases or whether out time is more wisely spent only looking at the upgrade of IDR to OMERO 5.4.

PR #5527, opened 2017-10-04
Fix 2017-SV4

sbesson:
Summary of the changes:

- the first commit is a forward port of the security fix in 5.3.5
- the second commit adjusts the server unit tests to prevent the usage of `userId` 1 (guest) for the tests
- the third commit is a cherry-pick of https://github.com/openmicroscopy/openmicroscopy/pull/5529 to restore the guest usage of `ResetPassword`

/cc @jburel @mtbc

PR #5529, opened 2017-10-05
allow guest users to use the reset password command

mtbc:
# What this PR does

Allows the `guest` user to use the reset password command.

# Testing this PR

Test the "forgotten password" functionality in web.

# Related reading

https://github.com/openmicroscopy/openmicroscopy/pull/5527

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,10 @@
 language: java
-
 jdk: oraclejdk8
 
 # This (sudo: false) is needed to "run on container-based infrastructure" on
 # which cache: is available
 # http://docs.travis-ci.com/user/workers/container-based-infrastructure/
-sudo: false
+sudo: required
 
 # http://docs.travis-ci.com/user/caching/#Arbitrary-directories
 cache:
@@ -19,11 +18,6 @@ matrix:
 addons:
   apt_packages:
     - git
-    - zeroc-ice34
-    - python-imaging
-    - python-numpy
-    - python-tables
-    - python-yaml
     - cmake
     - libgtest-dev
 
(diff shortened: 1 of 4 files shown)


jburel:
@mtbc the changes to``travis.yml`` done in develop will need to be ported. 
so it builds with ice 3.5 otherwise it will always fails


 using ``dist:precise`` might be enough..

pwalczysko:
I do not think this PR made it into the build which was triggered yesterday lunchtime, see https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/821/console .

Nevertheless, as reported by @jburel in General and confirmed by me, the Forgotten Password functionality works fine on web-dev-merge even without this PR.

jburel:
@pwalczysko This PR is for 5.3 that's why it is not included in the build
The relevant commit is included in https://github.com/openmicroscopy/openmicroscopy/pull/5527 (develop) which is included in the build. That's why it works

pwalczysko:
Test: (locally run)

  - checkout v5.3.5 tag 
  - merge to it this branch with this PR
  - ./build.py build-dev
  - fresh db
  - start server and web
  - login to web as root -> went okay
  - create a new user in web, give them the email address of myself
  - logout
  - login as the newly created user to double-check all is okay -> went okay
  - logout
  - click onto the Forgotten password link
  - fill in the new user's login name and my email address
  - no crash with stacktrace or Internal server error message observed

```
Mail server connection failed; nested exception is javax.mail.MessagingException: Could not connect to SMTP host: localhost, port: 25; nested exception is: java.net.ConnectException: Connection refused
```
Is the message in the splashscreen of the webclient, which I take as the test passing, and failing only at the stage of my local server not being configured to send emails.

jburel:
We can probably close this PR 
@sbesson what do you think?

mtbc:
Will probably wrap this into any further 5.3 security releases until 5.5/6.0 is out? Is there somewhere to thus capture this PR as an inclusion candidate?

sbesson:
Agreed. There is no active plan to have another 5.3.x patch release as of today. Some of the restrictions introduced by the OMERO 5.3.5 security fix have already been lifted in OMERO 5.4.0 as mentioned in https://www.openmicroscopy.org/security/advisories/2017-SV4-guest-user/. If we have to carry out a OMERO 5.3.6, this fix can certainly be reviewed for inclusion at that time.

PR #5537, opened 2017-10-09
add initial unit tests for graph policy rules

mtbc:
# What this PR does

Adds some initial overdue unit testing coverage for the `GraphPolicyRule` class. There is certainly scope for plenty more.

Incidentally, this work prototypes an initial non-public `DataProviderBuilder` class for supplying TestNG `@DataProvider`-based tests that in due course we may wish to think about and more greatly utilize.

# Testing this PR

Travis suffices.

# Related reading

https://trello.com/c/q9NrmK9X/16-graphpolicyrule-unit-tests

### DIFF
--- a/components/server/test/ome/services/graphs/DataProviderBuilder.java
+++ b/components/server/test/ome/services/graphs/DataProviderBuilder.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2017 University of Dundee & Open Microscopy Environment.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package ome.services.graphs;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


will-moore:
Travis is green.
Needs code review too?

mtbc:
As the PR only introduces test code then I'd guess code review is hardly critical but still welcome.

dominikl:
Code looks alright (but I don't know enough about the graph stuff to fully understand it!). Integration test job still running, will check it later.

dominikl:
Are the `GraphPolicyRuleTest`s actually run by the CI?

mtbc:
They're run by Travis but this can also be checked locally with,

```
./build.py -f components/server/build.xml test
```

dominikl:
Ah alright. Then it's good to merge from my point of view. Unless someone else with a deeper knowledge about the graphs than me wants to check the code again.

joshmoore:
The list of all classes registered in Hibernate can be automated. Is there anything _missing_ from this list?

mtbc:
I don't think there is. How to get hold of Hibernate stuff from a unit test?

joshmoore:
Ah, admittedly I wasn't taking the unit-test-ness into account. In the integration case, `ExtendedMetadata` will do it for you. Here, you'd need to parse `components/model/target/classes/hibernate.cfg.xml` directly, which is likely not worth the burden. That being said, I'd very much prefer if we _don't_ have to start updating this test on each model change.

mtbc:
It should be fine so long as the model changes don't affect any of the policy rules used in the tests. I included a currently complete list just so I could then write tests freely without worrying about that coverage.

joshmoore:
Understood. In way of a mental note, though, let's do please auto-generate this list the next time it's touched.

mtbc:
Actually, scratch that, it's even better: the unit tests should keep on working fine regardless of how far the "real" OMERO model drifts. I use the classes as named in OMERO model so the rules in the tests can look more like rules with which the reader (including me) is more likely to already be familiar.

mtbc:
Created https://trello.com/c/5LEOcpx4/775-autogenerate-class-list-for-tests.

mtbc:
@pwalczysko: The `DataProviderBuilder` introduced by this class *may* be a useful basis for developing `@DataProvider` methods in general. Something to be aware of for future test-writing anyway.

PR #5539, opened 2017-10-10
for the service factory checking for guests, refresh context

mtbc:
# What this PR does

Adjust the Blitz service factory check for guest users to refresh the event context.

# Testing this PR

Flakiness in integration tests may be reduced.

# Related reading

https://trello.com/c/qUkOm9ld/59-group-membership-flakiness


jburel:
If I remember correctly @joshmoore mentioned that such change could have an impact on web connection speed.

mtbc:
I too recall mention of probable performance impact. Further, I also tried a similar kind of fix to no good effect.

mtbc:
#5541 suffices.

PR #5540, opened 2017-10-10
Prevent 'endless' countTiles loop

dominikl:
# What this PR does

When an image reports ridiculous pixel sizes the `countTiles` method basically goes into an endless loop, which hangs the import process and creates unnecessary high server load. With this PR the pixel size is checked beforehand and if unreasonable high values are reported a default 'backoff' value is returned. This method is only used for constructing an Exception object anyway, therefore general implications of this PR should be very limited.

# Testing this PR

- Import the image from https://www.openmicroscopy.org/qa2/qa/feedback/17685 . You need to import the zipped file. Import of the file directly will be rejected. Make sure the import process finishes and the server load goes back to normal. Note: A thumbnail won't be generated, as this flawed image claims to have a size x,y,z of 892411973 x 1684497696 x 25971 pixels.
Alternatively you can also try to import something like `crazytestfile&pixelType=uint8&sizeX=892411973&sizeY=1684497696&sizeZ=25971.fake` instead.
- Check that `ImporterTest` integration test passes.

# Related reading

https://trello.com/c/GpPJ1Kcu/187-image-with-invalid-metadata-kills-server

### DIFF
--- a/components/romio/src/ome/io/nio/SimpleBackOff.java
+++ b/components/romio/src/ome/io/nio/SimpleBackOff.java
@@ -56,6 +56,10 @@ public class SimpleBackOff implements BackOff {
         this(new ConfiguredTileSizes(), 1000, 1000000);
     }
 
+    public SimpleBackOff(TileSizes sizes) {
+        this(sizes, 1000, 1000000);
+    }
+
     public SimpleBackOff(TileSizes sizes, long defaultValue, long maxPixels) {
         this.sizes = sizes;
         this.count = 10;
--- a/components/romio/src/ome/io/nio/SimpleBackOff.java
+++ b/components/romio/src/ome/io/nio/SimpleBackOff.java
@@ -48,13 +48,19 @@ public class SimpleBackOff implements BackOff {
 
     protected final TileSizes sizes;
 
+    protected final long maxPixels;
+    
+    protected final long defaultValue;
+    
     public SimpleBackOff() {
-        this(new ConfiguredTileSizes());
+        this(new ConfiguredTileSizes(), 1000, 1000000);
     }
 
-    public SimpleBackOff(TileSizes sizes) {
+    public SimpleBackOff(TileSizes sizes, long defaultValue, long maxPixels) {
         this.sizes = sizes;
         this.count = 10;
+        this.defaultValue = defaultValue;
+        this.maxPixels = maxPixels;
         try {
             ServiceFactory sf = new ServiceFactory();
             service = sf.getInstance(JAIIIOService.class);
@@ -85,10 +91,9 @@ public class SimpleBackOff implements BackOff {
 
     protected long calculate(Pixels pixels) {
         // only count tiles if pixels report reasonable size values
-        final long cutOff = 1000000;
-        return (pixels.getSizeC() > cutOff || pixels.getSizeT() > cutOff
-                || pixels.getSizeX() > cutOff || pixels.getSizeY() > cutOff || pixels
-                .getSizeZ() > cutOff) ? 1000
+        return (pixels.getSizeC() > maxPixels || pixels.getSizeT() > maxPixels
+                || pixels.getSizeX() > maxPixels || pixels.getSizeY() > maxPixels || pixels
+                .getSizeZ() > maxPixels) ? defaultValue
                 : (long) (scalingFactor * countTiles(pixels));
     }
 
--- a/components/romio/src/ome/io/nio/SimpleBackOff.java
+++ b/components/romio/src/ome/io/nio/SimpleBackOff.java
@@ -84,7 +84,12 @@ public class SimpleBackOff implements BackOff {
     }
 
     protected long calculate(Pixels pixels) {
-        return (long) (scalingFactor * countTiles(pixels));
+        // only count tiles if pixels report reasonable size values
+        final long cutOff = 1000000;
+        return (pixels.getSizeC() > cutOff || pixels.getSizeT() > cutOff
+                || pixels.getSizeX() > cutOff || pixels.getSizeY() > cutOff || pixels
+                .getSizeZ() > cutOff) ? 1000
+                : (long) (scalingFactor * countTiles(pixels));
     }
 
     protected int countTiles(Pixels pixels) {
--- a/components/server/resources/ome/services/service-ome.io.nio.PixelsService.xml
+++ b/components/server/resources/ome/services/service-ome.io.nio.PixelsService.xml
@@ -41,7 +41,9 @@
   </bean>
 
   <bean id="backOff" class="${omero.pixeldata.backoff}">
-    <constructor-arg ref="tileSizes"/>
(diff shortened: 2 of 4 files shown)


jburel:
You could have a test using a fake image with a crazy tile size

dominikl:
That's true, a `*.fake` file works too. I'll add a test for that.

components/romio/src/ome/io/nio/SimpleBackOff.java
```diff
@@ -84,7 +84,12 @@ public void throwMissingPyramidException(String msg, Pixels pixels) {
     }
 
     protected long calculate(Pixels pixels) {
-        return (long) (scalingFactor * countTiles(pixels));
+        // only count tiles if pixels report reasonable size values
+        final long cutOff = 1000000;
```
jburel: maybe  move the value to the configuration file

dominikl:
I just still have to find a proper name for the test, but apart from that ready to review.

components/romio/src/ome/io/nio/SimpleBackOff.java
```diff
@@ -84,7 +90,11 @@ public void throwMissingPyramidException(String msg, Pixels pixels) {
     }
 
     protected long calculate(Pixels pixels) {
-        return (long) (scalingFactor * countTiles(pixels));
+        // only count tiles if pixels report reasonable size values
+        return (pixels.getSizeC() > maxPixels || pixels.getSizeT() > maxPixels
+                || pixels.getSizeX() > maxPixels || pixels.getSizeY() > maxPixels || pixels
+                .getSizeZ() > maxPixels) ? defaultValue
+                : (long) (scalingFactor * countTiles(pixels));
```
mtbc, line 101: Does this `long` cast have to be *within* the following parentheses to have useful effect?

etc/omero.properties
```diff
@@ -390,6 +390,11 @@ omero.pixeldata.repetitions=1
 # to calculate the backoff (in ms) that users
 # should wait for an image to be ready to view.
 omero.pixeldata.backoff=ome.io.nio.SimpleBackOff
+# A default value for the backoff time
```
mtbc: maybe end with period, given those around

components/tools/OmeroJava/test/integration/ImporterTest.java
```diff
@@ -638,6 +638,27 @@ public void testImportGraphicsImages() throws Exception {
         Assert.assertEquals(0, failures.size());
     }
 
+    @Test(timeOut = 5000)
+    public void testImportInsaneImage() throws Exception {
+        // Simulates QA 17685 (an image with unreasonably huge pixel sizes stuck the import process
+        // in a basically endless loop when trying to throw an exception)
+        File f = new File(
+                System.getProperty("java.io.tmpdir"),
```
mtbc: Maybe `File.createTempFile` or similar?

mtbc:
I appreciate that this situation preexists this PR but seeing as you've probably recently had to think about it: Would it be possible to add a bit of explanatory documentation to `SimpleBackOff`? E.g., for your two new fields and maybe for what `calculate` is actually supposed to calculate. It's hard to look at the code and think "ah yes, that makes sense" when the meaning of the terms is ambiguous.

components/romio/src/ome/io/nio/SimpleBackOff.java
mtbc: Yes, some testing confirms that where it is it's essentially useless.

mtbc:
Test results at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/ImporterTest/.

components/romio/src/ome/io/nio/SimpleBackOff.java
dominikl: `scalingFactor` is a double, so the result of the multiplication has to be cast to a long, if not it'll cause a compile error.

mtbc: Ah! Thank you, I missed it. That makes more sense, sorry for the noise.

components/tools/OmeroJava/test/integration/ImporterTest.java
dominikl: Doesn't work, `File.createTempFile` adds random characters to the file name, but for the 'fake' import it to work it needs an exact file name.

mtbc: Ah, of course, and this class doesn't seem to have an instance of `TempFileManager` handy. Fair enough, no need to right now refactor the whole importer test class to use one.

dominikl:
I don't know anything about the reasoning behind it, imho this whole `Backoff` thing looks to me like hugely complicated construction for a rather simple thing (throwing an exception) anyway, specific omero.properties, counting tiles, calculating warmup factors and scaling factors,...

mtbc:
Fair enough, at least the test passes! Perhaps @joshmoore can add clarity at some later date. :smiley:

components/romio/src/ome/io/nio/SimpleBackOff.java
dominikl: No problem. I know it's a bit messy, double by int multiplation, casting to long... but it compiles ;-)

mtbc:
No blockers from me at least, I appreciate this isn't an easy PR and I certainly appreciate your taking the card. It's good to make the server harder to accidentally kill. @jburel, looks good to you?

components/romio/src/ome/io/nio/SimpleBackOff.java
```diff
@@ -48,13 +48,19 @@
...
+    
+    protected final long defaultValue;
+    
     public SimpleBackOff() {
-        this(new ConfiguredTileSizes());
+        this(new ConfiguredTileSizes(), 1000, 1000000);
     }
 
-    public SimpleBackOff(TileSizes sizes) {
+    public SimpleBackOff(TileSizes sizes, long defaultValue, long maxPixels) {
```
jburel: keep the method ``public SimpleBackOff(TileSizes sizes)`` so it does not break anything and introduce this one

components/server/resources/ome/services/service-ome.io.nio.PixelsService.xml
```diff
@@ -41,7 +41,9 @@
   </bean>
 
   <bean id="backOff" class="${omero.pixeldata.backoff}">
-    <constructor-arg ref="tileSizes"/>
+    <constructor-arg index="0" ref="tileSizes"/>
```
jburel, line 44: this is a change in the constructor. 
so we will need 2 constructors

otherwise the configuration will have to be introduced in a major release

mtbc (approved):

dominikl:
Last commit deletes the image at the end of the integration tests again to avoid potential problems of subsequent tests.

mtbc (approved):

PR #5542, opened 2017-10-11
webgateway - render_image_region: respect tile size parameters

waxenegger:
The web resource _webgateway/render_image_region_ is used by the default viewer and iviewer to get image tiles. Both bother handing in a tile width and height (which incidentally is the default tile size queried before) which is then replaced with the default size regardless.

It is, however, very useful to be able to adjust that tile size. In particular it would be one of the few quick and easy options to lower the tile requests and reduce strain on the server. The native top resolution of a tile pyramid is a good example. If this level happens to be fairly large still, i.e. necessitates a lot of tiles to cover it, and the view is all zoomed out to see the large image all of these tiles have to be retrieved in order to be able to display the image. If the tile size is 256 square for instance and that size is doubled we reduce the the number of requests to 1/4 of what it was before. 

To not reach a point to where this increase can be detrimental this PR sets an upper limit of 1024 pixels side length.


joshmoore:
@waxenegger : has this been through any testing? Do you have any more work coming?

jburel:
@joshmoore I have changed its target.

waxenegger:
@joshmoore I struggle to find a sensible way to test this. My initial idea was to extend: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/test/integration/test_rendering.py#L218

If you look at the tests you'll see that they only check for exceptions thrown in the case of missing query parameters/values. In other words, they don't follow through all the way to checking the returned image in the success case. Without that I don't see how I can really test the code bits that I added since they don't raise exceptions.

Now you might ask why didn't I start a test that returns a successfully rendered tile and, yes, I tried, only to find out that the test images created are all backed by an instance of RomioPixelBuffer which will throw an UnsupportedOperationExceptiom on setResolutionLevel which happens to be one of the code lines that is executed along the way. Not sure if there would be other hurdles further down the roadbut as it is this is already problematic enough.

Incidentally I could not find any integration test that will test the successful rendering of an image on the webgateway level (as in all the way to the binary return), and the only integration test there is this one which tests on a much lower level and is therefore not useful: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/test_render.py

Maybe you know of another way to test this because I'm out of ideas...

waxenegger:
@joshmoore good news.

I found an existing import method to create my test image which satisfies the conditions for https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/romio/src/ome/io/nio/PixelsService.java#L506 to give me back a BfPixelBuffer on which I can run my test cases.

Also I changed the code for the tile dimension bound checking ever so slightly: default tile size is used given value <= 0, max is clamped to 1024.

joshmoore:
@waxenegger : so this is ready for the `old/new|new/old|new/new` testing?

waxenegger:
@joshmoore the test case tuples you mentioned are iviewer/webgateway in relation to this [PR](https://github.com/ome/omero-iviewer/pull/130l), right? 

Because for the default viewer/webclient such a test would not make sense since they are coupled anyhow. That's what the added test should address.

waxenegger:
hmm. Either I'm missing something, I'm blind or looking in the wrong place. Shouldn't I be seeing the test https://github.com/openmicroscopy/openmicroscopy/pull/5542/files#diff-521930d801f9664bc545924aa8dce91eR306 in https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_rendering/

sbesson:
@waxenegger this is expected as per https://github.com/openmicroscopy/openmicroscopy/pull/5542#event-1383371526 i.e. the base branch of this PR is not `develop` and it is not included in the ci-master merge build. I would check with @jburel if there is a devspace integrating the `workflows` development branch.

jburel:
devspace information https://trello.com/c/hdHcpvXc/82-build-info

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -904,11 +904,19 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
...
+            # if tile size is given respect it
+            if len(zxyt) > 4:
+                tile_size = [int(zxyt[3]), int(zxyt[4])]
+                tile_defaults = [w, h]
+                for i, tile_length in enumerate(tile_size):
+                    # use default tile size if <= 0
+                    if tile_length <= 0:
+                        tile_size[i] = tile_defaults[i]
+                    # allow no bigger than 1024
+                    if tile_length > 1024:
```
jburel: maybe a value to configure?

jburel:
* Tested "default" viewer with this PR, no regression
* Tested daily iviewer with and without this PR

components/tools/OmeroWeb/omeroweb/webgateway/views.py
waxenegger: would an explicit declaration do, and, if so, where would be the recommended way to put it: within the method or file (after imports)?

jburel: something similar to ``omero.pixeldata.tile_height=256`` in ``etc/omero.properties`` (it needs to be handled in components/common/resources/ome/config.xml too)
that can then accessed like https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/views.py#L569

components/common/resources/ome/config.xml
```diff
@@ -2,9 +2,9 @@
 <!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
 <!--
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-# 
+#
```
jburel: you can get rid of the ``$Id``

components/common/resources/ome/config.xml
```diff
@@ -249,6 +249,11 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
+    <bean class="ome.system.Preference" id="omero.webgateway.max_tile_length">
```
jburel: name it ``omero.pixeldata.max_tile_length`` to be in synch with the rest
It might not only be used in webgateway

jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_rendering/TestRenderImageRegion/ is green

PR #5544, opened 2017-10-12
abort pyramid-making when tile fails

mtbc:
# What this PR does

Prevents filling logs with "Error during tile population" by having `PixelsService.performWrite` abort further tiles when building a pyramid from one throws an exception.

# Testing this PR

Review code and logs. There should still be *some* error reported for any failing cases of making pyramids for imported big images, just not crazily many.

# Related reading

https://trello.com/c/xUMP9iuF/54-tile-loop-logging-bug

### DIFF
--- a/components/romio/src/ome/io/nio/PixelsService.java
+++ b/components/romio/src/ome/io/nio/PixelsService.java
@@ -439,7 +439,7 @@ public class PixelsService extends AbstractFileSystemService
                         log.warn("Error clearing empty or incomplete pixel " +
                                  "buffer.", e2);
                     }
-                    throw new Utils.FailedTileLoopException();
+                    throw new FailedTileLoopException();
                 }
             }
             }, source, (int) tileSize.getWidth(), (int) tileSize.getHeight());
--- a/components/romio/src/ome/io/nio/PixelsService.java
+++ b/components/romio/src/ome/io/nio/PixelsService.java
@@ -1,7 +1,7 @@
 /*
  * ome.io.nio.PixelsService
  *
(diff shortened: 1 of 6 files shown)


mtbc:
@kennethgillen: Would it be possible to obtain a file relating to https://trello.com/c/WeQn1lY3/752-ns-zyue-file-not-found for testing this PR?

kennethgillen:
@mtbc  - 3x airyscan files on OME share "squig" for you. Certainly not for inclusion into data_repo etc without further discussion.

components/romio/src/ome/io/nio/PixelsService.java
```diff
@@ -438,13 +439,15 @@ public void run(int z, int c, int t, int x, int y, int w,
                         log.warn("Error clearing empty or incomplete pixel " +
                                  "buffer.", e2);
                     }
-                    return;
+                    throw new Utils.FailedTileLoopException();
```
joshmoore: Drop prefix?

joshmoore:
This is marked `breaking` but if we wanted the fix for 5.4.x a `RuntimeException` could be thrown and then a move to the caught `Exception` in 5.5+.

dominikl:
Code looks alright. But wasn't able to test the PR.

joshmoore:
@dominikl : the same holds for me.  I think the biggest question here is that we need to agree on the `RuntimeException --> Exception` migration.

jburel:
All happy with the ``warning`` in the doc and the exception thrown?

PR #5545, opened 2017-10-13
Gateway methods

jburel:
# What this PR does

* Port commits currently in metadata 53. The Commits were used by the cli plugin render
* Add tests for new methodss
* Introduce ``set_active_channels`` and deprecate ``setActiveChannels``
* Review tests and make sure that all services are closed

# Related reading

see https://github.com/ome/omero-cli-render/pull/1

# To test
* Check that the tests are green

cc @joshmoore

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1551,6 +1551,7 @@ class _BlitzGateway (object):
         self._user = None
         self._userid = None
         self._proxies = NoProxies()
+        self._tracked_services = dict()
         if self.c is None:
             self._resetOmeroClient()
         else:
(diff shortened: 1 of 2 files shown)


dominikl:
Tests are passing and make sense. Given that really all services are registered via the`_register_service` method. I can't really work it out how that works. Would it be possible to add a test for that, too? Because in case they wouldn't get registered in first place, `assert not g._assert_unregistered(...)` in the tests would always pass.

jburel:
All services are registered, what is important is to make sure that they are all closed.
This was not the case before the changes i.e. 17 services were left opened
This is why ``assert not g._assert_unregistered`` has been added at the end of all the tests

joshmoore:
A bit concerned by how invasive some of the changes are (i.e. it will be easy to forget to follow the patterns when writing new tests). 2 questions:

 * Can the test assertion be done at a higher level? e.g. during tear down. We track all `omero.client` creations; we could do the same for gateways.
 * Are the remaining open services indicative of a code problem as opposed to a test problem? Does this mean that OMERO.web is also leaving them open?

jburel:
This will require discussion since I do not know what was the initial plan on the metadata branch
The strategy in the tests is the one used in the metadata branch for render i.e. check at the end of each test
It certainly helps tracking down open service in the tests.
OMERO.web is probably leaving a good number of service open but I think it is mainly a test problem. I did not review that as part of this PR.
The main goal is to make the structure accessible to the plugins

joshmoore:
This has been in for some time. Merging to take a look at the omero-cli-render repo. Note: created https://trello.com/c/Qc6QEQS0/40-remove-need-for-imgclosere to come back to the changes here and try to remove the invasiveness.

PR #5546, opened 2017-10-13
in codegen add checks for legal property values

mtbc:
# What this PR does

For server-side model objects adds checks for legal property values for types from `psql-types.properties`:

```ini
PositiveInteger=positive_int
NonNegativeInteger=nonnegative_int
PositiveFloat=positive_float
PercentFraction=percent_fraction
```

This enables errors to be more promptly detected: in the code that sets the value, rather than later on when we save the objects and hit the constraints from `psql-header.sql`:

```sql
CREATE DOMAIN nonnegative_int AS INTEGER CHECK (VALUE >= 0);
CREATE DOMAIN nonnegative_float AS DOUBLE PRECISION CHECK (VALUE >= 0);
CREATE DOMAIN positive_int AS INTEGER CHECK (VALUE > 0);
CREATE DOMAIN positive_float AS DOUBLE PRECISION CHECK (VALUE > 0);
CREATE DOMAIN percent_fraction AS DOUBLE PRECISION CHECK (VALUE >= 0 AND VALUE <= 1);
```

which causes an undesirable PostgreSQL error in the server log.

Prompt detection of the violation means that the stack trace points more directly at the actual culprit rather than something rather later in the code path.

Also adds some related comments to our SQL scripts.

# Testing this PR

`PropertyConstraintTest` does some testing of this PR via Travis. You could try writing code to create `ome.model` objects (*not* `omero.model`) and set bad values in them according to the corresponding constraints from http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome.html (see the "simple types" and look for property values that use them): you should get an immediate exception without needing to try to save to the database. It might also be interesting to find one of these images for which somebody got a related "ConstraintViolationException: could not insert" upon importing to see how the stack trace is now changed. I could try to hack a suitable fake together with OME-XML if desired.

# Related reading

https://trello.com/c/oqiU0S38/73-check-constraints-in-model-property-setters

### DIFF
--- a/components/dsl/resources/ome/dsl/object.vm
+++ b/components/dsl/resources/ome/dsl/object.vm
@@ -1286,6 +1286,16 @@ implements java.io.Serializable, IObject
 
     public void set${prop.nameCapped}(${prop.fieldType} ${prop.name}) {
         $errorIfUnloaded
+## See table column type alterations in psql-footer.vm.
+#if(($type.shortname == "Laser" && ($prop.name == "wavelength")) || ($type.shortname == "LightSettings" && ($prop.name == "wavelength")) || ($type.shortname == "LogicalChannel" && ($prop.name == "emissionWave" || $prop.name == "excitationWave")) || ($type.shortname == "Pixels" && ($prop.name == "physicalSizeX" || $prop.name == "physicalSizeY" || $prop.name == "physicalSizeZ")) || ($type.shortname == "TransmittanceRange" && ($prop.name == "cutIn" || $prop.name == "cutOut")))
+        if (${prop.name} != null && ${prop.name}.getValue() <= 0) {
+            throw new IllegalArgumentException("values of ${type.shortname}.${prop.name} must be strictly positive");
+        }
+#elseif(($type.shortname == "TransmittanceRange" && ($prop.name == "cutInTolerance" || $prop.name == "cutOutTolerance")))
+        if (${prop.name} != null && ${prop.name}.getValue() < 0) {
+            throw new IllegalArgumentException("values of ${type.shortname}.${prop.name} must not be negative");
+        }
+#end
         this.${prop.name} = ${prop.name};
     }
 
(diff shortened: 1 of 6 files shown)


mtbc:
The changes in our model code caused by this PR are 
[model-changes.diff.gz](https://github.com/openmicroscopy/openmicroscopy/files/1381860/model-changes.diff.gz).

components/dsl/resources/ome/dsl/object.vm
```diff
@@ -1286,6 +1286,16 @@ implements java.io.Serializable, IObject
 
     public void set${prop.nameCapped}(${prop.fieldType} ${prop.name}) {
         $errorIfUnloaded
+## See table column type alterations in psql-footer.vm.
+#if(($type.shortname == "Laser" && ($prop.name == "wavelength")) || ($type.shortname == "LightSettings" && ($prop.name == "wavelength")) || ($type.shortname == "LogicalChannel" && ($prop.name == "emissionWave" || $prop.name == "excitationWave")) || ($type.shortname == "Pixels" && ($prop.name == "physicalSizeX" || $prop.name == "physicalSizeY" || $prop.name == "physicalSizeZ")) || ($type.shortname == "TransmittanceRange" && ($prop.name == "cutIn" || $prop.name == "cutOut")))
```
joshmoore, line 1290: Why did the field names need to be hard-coded? Do the mapping files not capture the types appropriately?

mtbc: They indeed do not capture range constraints on dimensioned values.

joshmoore: Aha. Further dragging cleanup from the units migration. Half-tempted to say we should do this correctly but putting the necessary metadata into the mapping files.

mtbc: I certainly agree though we're unlikely to effect this anytime soon. Needs some design: for instance, perhaps best to break existing range constraints wholly out of `type` into a new attribute? Could create a card.

jburel: Do you want to move the card indicated in the description to a new board to capture the follow-up work?

mtbc: This PR satisfies the current card. Created https://trello.com/c/s1XwJTmD/ accordingly.

PR #5547, opened 2017-10-13
increase graphs logging level to DEBUG

mtbc:
Causes DEBUG-level logging for graphs code. Assists investigation of https://trello.com/c/DaJKgLoY/82-chown-all-scriptjob-failures. *Not* for merging.


mtbc:
Captured the log and attached to card.

PR #5549, opened 2017-10-16
enforce canonical path representation for repositories

mtbc:
# What this PR does

Complementing the current expression index `originalfile_repo_path_index` in the database this PR adds checks in Java code for how file paths in repositories should be represented. If the server catches these errors earlier then fewer PostgreSQL errors are logged.

# Testing this PR

CI should remain green.

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/CheckedPath.java
+++ b/components/blitz/src/ome/services/blitz/repo/CheckedPath.java
@@ -316,7 +316,11 @@ public class CheckedPath {
      * slash.
      */
     protected String getDirname() {
-        return this.fsFile.toString() + FsFile.separatorChar;
+        if (FsFile.emptyPath.equals(this.fsFile)) {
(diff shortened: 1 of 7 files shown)


mtbc:
Still need to update some Python integration tests.

mtbc:
This morning's `LutProviderImpl.initLutReaders` exception makes me wonder if the problem was that the LUTScripts registration was occurring *before* `Ring.init` so the `_protect_originalfile_repo_*` database triggers had no valid `node.uuid` to check as a secret key.

mtbc:
`PublicRepositoryI.assertFindDir` fails on import for users who have directories in the "old" form. Could consider workaround or simply rely on database upgrade to fix existing file paths.

mtbc:
Will need a database upgrade anyway to avoid subversion of `_fs_dir_delete` so not for 5.4.

PR #5550, opened 2017-10-17
Switch to lookup table provider service

chris-allan:
# What this PR does

The previous lookup table loading implementation both leaves the task to
the individual service and has potentially severe performance
implications.  Two different loading mechanisms were used in the
thumbnail and rendering services due to the session state and service
availability.  Furthermore, the actual loading itself was deferred deep
within the renderer infrastructure.

This PR institutes a single `LutProvider` interface which can be
passed into services which desire access to the currently available
lookup tables.  A corresponding `LutProviderImpl` Spring Bean has also
been added, allowing for the implementation to be swapped in and out as
required.

Due to session containment differences between the thumbnail and
rendering services LUT loading was taken all the way.  Loading now
happens at server start time.  The only perceptible downside to this is
that adding a new LUT requires a server restart.  However, the
implementation was written in such a way that server runtime updates to
the available LUTs could be easily added if desired.

# Testing this PR

Use of the thumbnail and rendering services while both under and devoid of the influence of lookup tables should be unchanged.

# Related reading

1. https://trello.com/c/9iAe3Sg9/95-lut-support

2. glencoesoftware/omero-ms-image-region@8a95d980ea025168615cf5009d8349638b5c5e1d


will-moore:
What's the method to use for adding LUTs to the server? Use existing API to add Original Files of ```mimetype = 'text/x-lut'```? In the user group? Or is the existing workflow of adding them to ```dist/lib/scripts/luts/``` still valid?

chris-allan:
@mtbc and @joshmoore will want to take a close look at 859cb6fd88eb73b7702bf231911b1432e481bddb and be sure we are happy with both the sequence and result of this change.

chris-allan:
There's no change at all the semantics of adding or removing LUTs, @will-moore. The only caveat is that we are not refreshing the list on every call to the `ThumbnailBean` or `RenderingBean`. A server restart is now required to get the current list of LUTs sync'd up with what is in the script repository.

will-moore:
Several scripts test failures, due to this PR? https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-breaking-integration-java/855/

chris-allan:
One thing that had occurred to me is that #5273 does have the side effect of subtly breaking the API of `ScriptRepoHelper.loadAll(*)`. The return value of that method will now have the secret key inserted into the `name` attribute. We use the method so sparingly that I don't see this being a problem but it's probably something we should double check, @mtbc.

mtbc:
This morning's `LutProviderImpl.initLutReaders` exception makes me wonder if the problem was that the LUTScripts registration was occurring *before* `Ring.init` so the `_protect_originalfile_repo_*` database triggers had no valid `node.uuid` to check as a secret key.

chris-allan:
Certainly possible. There are so many potential races with all of this that I think the deferred initialization strategy is definitely the one to pursue. Should have something pushed with that change here shortly.

mtbc:
Might also suffice to add an explicit bean dependency in the Spring config but whichever works. :smiley:

chris-allan:
Yeah, I went down that road. Unfortunately, it quickly descends into bringing all these dependencies into the "blitz" component as a result. Which I did try but it felt really dirty. Especially as I would really like to be able to use these beans with the microservices and not have to initialize all of Blitz in order to do so. Furthermore, because the scripts are actually registered in the database by reacting to the Spring Application Context refresh events I'm not even 100% that it would work consistently. We already have enough problem with that as it is.

mtbc:
It's thus possible that relying on the Node UUIDs for allowing the updates to the file repo property was a design error that will need revisiting. In either case I concede that it's ugly.

chris-allan:
4e7ef8707fd6134a47839a83aa147467800c150e strips out all the weird stuff from 859cb6fd88eb73b7702bf231911b1432e481bddb and just goes directly to the filesystem in a `ScriptRepoHelper` compatible way. This gets rid of all of the concurrency issues, loads everything at startup time, and utilizes much of the same logic.

We have to go to the filesystem directly to load the LUTs anyway so I can't see a downside to this approach.

/cc @jburel

components/server/src/ome/logic/LutProviderImpl.java
```diff
@@ -0,0 +1,103 @@
...
+     * @see omeis.providers.re.lut.LutProvider#getLutReaders(ome.model.display.ChannelBinding[])
+     */
+    public List<LutReader> getLutReaders(ChannelBinding[] channelBindings) {
+        log.debug("Looking up LUT readers for {} channels from {} LUTs",
+                channelBindings.length, lutReaders.size());
+        List<LutReader> toReturn = new ArrayList<LutReader>();
+        for (ChannelBinding cb : channelBindings) {
+            if (cb.getActive()) {
+                LutReader lut = lutReaders.get(cb.getLookupTable());
+                if (lut != null) {
```
mtbc: Seem to do the same thing in both cases?

joshmoore:
Comments from an initial read with Chris on Zoom:

 * one concern that may belong in this new `Provider` interface is the generation of the overview PNG
 * otherwise, looks like the default implementation could be replaced by one to use redis, which is nice
 * long-term, I'd like to see all the scripts cruft stripped out in favor of something more truly asynchronous

joshmoore:
@chris-allan points out that this could be a candidate for a milestone release due to the performance benefits, but that would require a general agreement on the behavior change related to reloading of LUTs.

cc: @sbesson @jburel

mtbc:
This PR causes https://trello.com/c/NSSdxI5n/64-bug-blocking-save-to-all-thumbs.

mtbc:
`org.hibernate.NonUniqueObjectException` exceptions attached to card.

joshmoore:
Excluding from develop to have 5.4.2 tested for a while without. Happy for this to be re-included on an integration branch.

mtbc:
That works excellently, thank you. Whatever the underlying issue is, the latest commit appears to work around it. :+1:

I wasted some time with the web client's "Save to All" saying "Cannot apply settings to other images". Turns out I had to move the images into a dataset -- pity the error gives no hint of that! Outwith this PR though.

jburel:
Discussed with @chris-allan, we will look at the underline issue post 5.4.4
Merging tomorrow unless any vetoes

chris-allan:
Sounds good to me!

PR #5551, opened 2017-10-17
testtables_initialize_columns

will-moore:
# What this PR does

Fixes the CLI  ```metadata tablestest``` plugin to not use an empty list of Columns when testing
```table.initialize([])``` since this now fails with
```
{
    serverStackTrace = 
    serverExceptionClass = 
    message = No columns provided
}
```
Instead we provide a single column, so this should work if tables are functioning correctly.
Robot framework setup.py uses ```$ bin/omero metadata tablestest``` to only attempt populate_metadata if it passes.

Also, we were using ```bulk_annotations``` (created by populate metadata) as a second File Annotation in some tests that were not testing Tables functionality. These were failing when that file was missing. Now we create a second file annotation so these tests no-longer rely on ```bulk_annotations```.

# Testing
 - Recently several tests of "Bulk Annotations" were failing because populate_metadata was not getting run (even though tables were working on that machine). https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/739/robot/

 - These tests should now be passing (if tables are working).

 - Other tests such as https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/739/robot/Web%20&%20Web/Web_1/Annotate%20Test/Test%20File%20Annotations/ should no longer fail with e.g. ```NoSuchElementException: Message: Could not locate element with visible text: bulk_annotations```

### DIFF
--- a/components/tests/ui/robot_setup.sh
+++ b/components/tests/ui/robot_setup.sh
@@ -146,7 +146,7 @@ do
   bin/omero import $TINY_IMAGE_NAME --debug ERROR
 done
 
-# Uplodad files and create FileAnnotations
+# Upload files and create FileAnnotations
 ofile=$(bin/omero upload $FILE_ANNOTATION)
 bin/omero obj new FileAnnotation file=$ofile
 ofile2=$(bin/omero upload $FILE_ANNOTATION2)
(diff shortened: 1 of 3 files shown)


will-moore:
The setup script is failing with ```bin/omero: error: argument <subcommand>: invalid choice: 'metadata'``` at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/742/console since we need to set ```OMERO_DEV_PLUGINS=1``` for both invocations of ```bin/omero metadata..``` on the same line.
Fixed by ```export OMERO_DEV_PLUGINS=1``` in commit above.

rgozim:
Local test passing and passing in CI.

joshmoore:
:+1: Note: I pushed an extra typo fix.

PR #5553, opened 2017-10-19
fix_guest_username_check

will-moore:
# What this PR does

Fixes checking of "guest" user, e.g. in webadmin Users table.
See https://trello.com/c/U47AiD1R/48-weird-guest-login

To test, check that a user with e.g. username "g" is still editable at /webadmin/experimenters/ (and "guest" user is not).


dominikl:
Users (also User 'g') are still editable. But so is the guest account. Edit dialog works also for guest user, save button is enabled. Would have to set a group to perform a save action (didn't try that, don't wanna break the web-dev server). 
Tested logged in with root user on web-dev-merge.

will-moore:
@dominikl Thanks. My mistake. Fixed now!

dominikl:
Works now, thanks @will-moore . Good to merge 👍 
(Tested with local build of the PR, should be enough, but can test again on web-dev-merge tomorrow if needed).

PR #5555, opened 2017-10-19
Fix filter_by_name refresh

will-moore:
# What this PR does

Fixes the page refresh when you "submit" the thumbnail filter form.
See https://trello.com/c/Znp0U4Mw/32-web-client-filter-by-name-bug
We don't use the ```<form>``` element here at-all, but the default form behaviour of "submit" when you hit enter in the form was causing page to refresh.
We now just use a ```<div>``` instead.
See https://trello.com/c/Znp0U4Mw/32-web-client-filter-by-name-bug

# Testing this PR

1. Browse to Dataset of thumbnails
1. Filter by name
2. In the filter field, hit enter (and wait)
3. Page should not refresh (go to /webclient/?#)


dominikl:
Works like expected, good to merge 👍

PR #5562, opened 2017-11-02
null-protect interceptor setting admin privs in database

mtbc:
# What this PR does

Adds a null check around OMERO interceptor's post-flush activity noting current administrator privileges in the database.

# Testing this PR

CI should show no regressions. Small code diff should make sense.

# Related reading

https://trello.com/c/kSKvNpjw/101-bug-ns-errors-null-pointer-on-session-close

### DIFF
--- a/components/server/src/ome/security/basic/OmeroInterceptor.java
+++ b/components/server/src/ome/security/basic/OmeroInterceptor.java
@@ -15,6 +15,7 @@ import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.ObjectUtils;
 import org.hibernate.CallbackException;
 import org.hibernate.EmptyInterceptor;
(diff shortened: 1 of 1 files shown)


joshmoore:
Certainly fine with either these changes, or with making the `insert` method more cleanly handle empties. (There are a number of SqlAction methods that need to do that.)

With your being the most familiar with this code, can you say anything to _why_ the field would empty though, @mtbc?

mtbc:
An empty list is plausible when the light admin simply does not have any permissions (e.g., no boxes checked in webadmin).

I don't know how the `null` happened -- I've not seen it in CI -- but my best guess so far is that before a transaction was finished the `CurrentDetails.current()` got set to `CloseOnNoSessionContext`; perhaps a session timed out?

joshmoore:
Understood. :+1:

joshmoore:
Merging. (It occurs to me to wonder though why this is done in 2 JDBC calls)

mtbc:
We want to remove old rows and add new ones. Maybe there's some kind of "replace" we can do even with the WHERE or we can somehow combine both SQL into one call. (Could also used a stored procedure to help.)

PR #5564, opened 2017-11-07
CLI - Hide passwords by default

dominikl:
# What this PR does

Hide passwords by default when running `./omero config get`. Deprecate `--hide-password` option and replace with `--show-password`.

# Testing this PR

- Run `./omero config get` and check that passwords are replaced by `********`.
- Run `./omero config get --hide-password` and check you get the same output.
- Run `./omero config get --show-password` and check that passwords are displayed.

# Related reading

https://trello.com/c/9rHSC7aB/54-hide-passwords-in-config-get

Is there a certain format to use when deprecating arguments? I just added `[Deprecated ...]` in front of the argument's help text for now.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -26,6 +26,7 @@ from omero.util import edit_path, get_omero_userdir
 from omero.util.decorators import wraps
 from omero.util.upgrade_check import UpgradeCheck
 from omero_ext import portalocker
+from omero_ext.argparse import SUPPRESS
 
 import omero.java
 
@@ -148,14 +149,13 @@ class PrefsControl(WriteableConfigControl):
         get.set_defaults(func=self.get)
         get.add_argument(
             "KEY", nargs="*", help="Names of keys in the current profile")
-        get.add_argument(
-            "--hide-password", action="store_true",
-            help="[Deprecated; this is the default now] Hide values of "
-                 "password keys in the current profile")
-        get.add_argument(
-            "--show-secrets", action="store_true",
-            help="Show values of sensitive keys (passwords, tokens, etc.) "
-                 "in the current profile")
+
+        secrets = get.add_mutually_exclusive_group()
+        secrets.add_argument("--show-secrets", action="store_true",
+                             help="Show values of sensitive keys (passwords, tokens, "
+                                  "etc.) in the current profile")
+        secrets.add_argument("--hide-password", action="store_false", dest="show_secrets",
+                             help=SUPPRESS)
 
         set = parser.add(
             sub, self.set,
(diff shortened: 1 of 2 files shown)


manics:
Do you think it's worth making the flag more generic, e.g. `--show-secrets`, in case in future we want to hide tokens or other secret config?

components/tools/OmeroPy/src/omero/plugins/prefs.py
```diff
@@ -150,7 +150,12 @@ def _configure(self, parser):
             "KEY", nargs="*", help="Names of keys in the current profile")
         get.add_argument(
             "--hide-password", action="store_true",
-            help="Hide values of password keys in the current profile")
+            help="[Deprecated; this is the default now] Hide values of "
```
joshmoore: You can also use `SUPPRESS` if that's preferred: https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.0/components/tools/OmeroPy/src/omero/plugins/import.py#L325

components/tools/OmeroPy/src/omero/plugins/prefs.py
```diff
@@ -150,7 +150,12 @@ def _configure(self, parser):
...
+                 "password keys in the current profile")
+        get.add_argument(
+            "--show-secrets", action="store_true",
```
joshmoore: You can also have `--show-secrets` store into the same variable (`hide_secrets`) by using `store_false` with the `target` argument. This would keep the rest of the logic the same. See https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.0/components/tools/OmeroPy/src/omero/plugins/metadata.py#L171

That example also uses `add_mutually_exclusive_group()` which may be of interested.

joshmoore (commented):
A few general suggestions about our use of the `argparse` API added. My only real concern is how _breaking_ we consider the change. If we consider the previous behavior a security risk, though, it's still worth getting in.

dominikl:
The PR only makes sense when the change is *breaking* right? There's no point in adding a `--show-secrets` flag when the default is to show passwords anyway. If it shouldn't be breaking then I'd say just close the PR and the respective trello card. Looks like @mtbc added the card initially.

mtbc:
I consider the previous behavior a security risk but can wait for 5.5/6.0 if too breaking to change.

components/tools/OmeroPy/src/omero/plugins/prefs.py
```diff
@@ -148,9 +149,13 @@ def _configure(self, parser):
         get.set_defaults(func=self.get)
         get.add_argument(
             "KEY", nargs="*", help="Names of keys in the current profile")
-        get.add_argument(
-            "--hide-password", action="store_true",
-            help="Hide values of password keys in the current profile")
+
+        secrets = get.add_mutually_exclusive_group()
+        secrets.add_argument("--show-secrets", action="store_true",
```
joshmoore: Rename `secrets` to `passswords` to match previous expectations.

joshmoore (changes requested):
Sorry for not having sooner. As soon as I tried to use this on the CLI, I memory-typed `--show-passwords`. Unless there's a motivation, for having 2 different values, I'd suggest unifying.

re: breaking -- I'm not suggesting this should be closed. I just want to make sure we all agree on the implications, the impact, and the implementation. (..._couldn't think of any other "imp" words_...)

joshmoore (approved):

joshmoore:
Renames all look nice. Thanks, @dominikl

PR #5568, opened 2017-11-14
Filter by tag

will-moore:
# What this PR does

Adds an important feature from Insight to webclient: Filter Dataset thumbnails by Tag.
See https://trello.com/c/PNcVcXOt/26-filtering-by-tags-in-webclient

# Testing this PR

1. Tag some images in a Dataset
1. Add Filter -> Tag
1. List of Tags used in the Dataset should be listed in a Select element
1. Choosing Tags adds them to the filter and images without those Tags should be hidden
1. Clicking ```-``` on Tags removes them from the filtering
1. Combine Filtering by Rating and/or Name
1. Add Filter -> Remove all filters  should work as expected

screen shot 2017-11-14 at 22 46 01

TODO: Add robot framework tests.


pwalczysko:
Feature is very helpful, and usage is intuitive. 
I would have 1 RFE and one 2 Bugs.

RFE: Do not re-offer tags which are already being filtered for. This means, have tags A, B, C in the dataset. Select the tag A to be filtered for. Select a dropdown again, but now, have only tags B and C to be selected as second tags (at present, A can be re-selected again, and although this does not crash the app, it makes no sense)

Bug1: See screenshot below, if you select enough tags, the layout will distort so that it disables the functionality

screen shot 2017-11-16 at 11 12 50

Bug2: Filter for some tags. Then, click on "Remove all filters" . Filter for tags again. The starting point is now coming from the tags you already pre-selected in the first round. This is not expected. "Remove all filters" should really remove the whole list of tags when the tagging filter is revisited.

will-moore:
RFE and Bug2 fixed.
Bug1 is harder to fix. I'm not sure what it should look like, and redoing the layout when the list gets too long is tricky. Do we need to support filtering by so many Tags?

rgozim:
Works well, however not a big fan of the UI. Could we maybe use a drop down with checkboxes (see screenshot)?

screen shot 2017-11-23 at 10 58 28

will-moore:
I briefly explored using the Chosen plugin for picking Tags, but had issues with the drop-down menu being hidden behind the lower panel.
However, I quite like the current behaviour of showing Tags in the same way as they appear in the Tags tab, which allows them to be seen when the drop-down menu is collapsed (so you know what you're filtering by) and removed with a single click (instead of having to open the drop-down menu again).

jburel:
robot PR merged, you can now add more tests

jburel:
Will using the Chosen plugin fix Bug 1 reported by @pwalczysko?

will-moore:
@jburel No. For chosen.js you can set the width, e.g. ```500px``` and it will start off this wide even if you don't choose many tags. If you do choose lots of tags, you still get the same problem:

screen shot 2018-02-01 at 12 28 52

I have this locally as e58c4287cf just in case we want to pursue this, but it's not looking good.

jburel:
ok thanks for checking
I find the location of the tags very strange
having the selection box pushed to the right each time I add a tag is not very nice

Also I have no way to remove a specific filter, it is all or nothing
This will need to be taken into consideration

jburel:
Discussed today: 
* Remove ``Filter``
* Keep ``Add Filter`` selection box on left-hand side
* Introduce new row for ``Choose Tag`` selection box
* Keep the  ``Choose Tag`` selection box on left
* Add option to remove the "Tag" row

will-moore:
Only thing remaining is to add option to remove the Tag row. Any suggestions about where this should go?
Where can this be tested (devspace)?
screen shot 2018-02-01 at 16 58 53

jburel:
We can test it tomorrow on devspace  http://10.0.51.135:32775/web

maybe a cross in the on the left of the selection box

pwalczysko:
@jburel How can I get to the jenkins UI of that devspace ? I would like to change the config so that it does not produce the empty thumbnails and we can accumulate some data on this devspace over days. https://github.com/openmicroscopy/devspace/pull/90

jburel:
card with info https://trello.com/c/hdHcpvXc/82-build-info 
purge data is off

jburel:
2 robot tests failures
https://10.0.51.143:32866/job/OMERO-robot/24/

will-moore:
NB: The css changes in this PR use flex-box which is not supported on IE9.
I don't think we've defined anywhere what browsers we support or have tested on?
@pwalczysko?
IE9 had 0.9% market share in 2015 https://www.w3schools.com/browsers/browsers_explorer.asp (last year for which it's listed) so I think we are probably safe.
cc @waxenegger Does Aurelia still support IE9? Not sure if OMERO.figure works in IE9.

pwalczysko:
Tested on workflows devspace (user-3 on user-5's images in RA) and user-6 on user-4's images in read-only. 

I find the ease of use and the logic good now. All points from https://github.com/openmicroscopy/openmicroscopy/pull/5568#issuecomment-362331979 seem to be implemented.
Ready to merge FMPOV once the robot test problem https://github.com/openmicroscopy/openmicroscopy/pull/5568#issuecomment-363028848 is fixed.

pwalczysko:
@will-moore I am testing on IE11 on Win7 laptop. IE8 does really sound ancient. The majority of users in SLS have Win7 + Firefox combo. Actually I always just test on the most recent IE version.

will-moore:
@jburel Pushed a couple of commits that hopefully will help the robot tests.

jburel:
I have redeployed the devspace so we are sure it will not hang tomorrow
I will update the card shortly

waxenegger:
in regards to IE versions: https://www.microsoft.com/en-au/windowsforbusiness/end-of-ie-support

jburel:
https://10.0.51.143:32875/job/OMERO-robot/25/ one failure related to this PR

will-moore:
Robot tests green today https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/798/ 🎉

jburel:
Thanks @will-moore for the adjustments it is better now

PR #5569, opened 2017-11-15
fetch artifacts from repos over HTTPS, not HTTP

mtbc:
# What this PR does

Has Ivy fetch third-party artifacts over HTTPS rather than simple HTTP.

# Testing this PR

OMERO should build fine even *without* a local maven cache already populated.

# Related reading

https://github.com/openmicroscopy/bioformats/pull/2992
--rebased-to #5619


jburel:
Tested in devspace and with this PR it was possible to build. It was failing without
I built locally OMERO without this PR and I did not have any problems to build so the problem might not affect as many people as we thought initially
Merging

PR #5572, opened 2017-11-17
Gitignore additions

rgozim:
### What this PR does
Adds files extensions (.iml) to root project .gitignore file and removes restriction of preventing other .gitignore files from being added to subfolders of the project.

No testing required, just your opinion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,11 +46,14 @@ temp-testng-customsuite.xml
 .settings
 .springBeans
 /components/tools/OmeroWeb/omeroweb/media/webtest
+/components/tools/OmeroWeb/omeroweb/static/
+/components/var
 # Scons output files
 .sconsign.dblite
 *.dylib
 *.egg
 *.os
+*.iml
 omero_version.py
 eclipse.log
 velocity.log*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -46,11 +45,14 @@ temp-testng-customsuite.xml
 .settings
 .springBeans
 /components/tools/OmeroWeb/omeroweb/media/webtest
+/components/tools/OmeroWeb/omeroweb/static/
+/components/var
```
joshmoore, line 49: I'm not sure where `components/var` would be coming from.

.gitignore
```diff
@@ -46,11 +45,14 @@ temp-testng-customsuite.xml
 .settings
 .springBeans
 /components/tools/OmeroWeb/omeroweb/media/webtest
+/components/tools/OmeroWeb/omeroweb/static/
```
joshmoore, line 48: I could see moving these down to OmeroWeb/.gitignore (or one deeper). Especially when/if OmeroWeb is extracted.

joshmoore (commented):
Few minor comments, but this works for me.

PR #5573, opened 2017-11-17
Test scripts

jburel:
# What this PR does

Similar to CLI testing
This PR moves ``script.py`` to testlib so scripts can be tested in their own repo

see https://github.com/ome/scripts/pull/136

cc @joshmoore


joshmoore:
Cheers, @jburel. Note: looking into how to test `omero-cli-render` and `OmeroPy/scripts` from within devspace, we may need to standardize a bit how things are built and tested.

Merging into clisplit.

PR #5577, opened 2017-11-17
Error message change on OMERO import

rgozim:
Changed error message from `Directory exists but is not registered` to `OMERO data directory and database out of sync. Please re-create your OMERO data directory` when a user attempts to import data into OMERO but hasn't re-created their data directory.

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/PublicRepositoryI.java
+++ b/components/blitz/src/ome/services/blitz/repo/PublicRepositoryI.java
@@ -855,7 +855,7 @@ public class PublicRepositoryI implements _RepositoryOperations, ApplicationCont
         if (null == repositoryDao.findRepoFile(sf, sql, repoUuid, checked, null)) {
             omero.ResourceError re = new omero.ResourceError();
             IceMapper.fillServerError(re, new RuntimeException(
-                    "Directory exists but is not registered: " + checked));
+                    "OMERO data directory and database out of sync. Please re-create your OMERO data directory."));
             throw re;
         }
     }


mtbc:
Recreating the data directory may cause more problems than it solves (data loss, duplicate repos, etc.) if the underlying issue is not the developer-centric workflow of missing out a step when trying to start out with a fresh DB. I am not sure if this condition could occur after, for instance, restoring a backup of a production server that was running when the backup was taken, or after an OOM killer took out either Blitz or PostgreSQL and the server is restarted after. I'd definitely invite @joshmoore's comment on this PR.

joshmoore:
Thanks for opening this, @rgozim. To some extent, I was looking exactly for how you'd put this, i.e. "OMERO data directory and database out of sync." I'd _minimally_ suggest that we integrate that into the previous message to make this clearer.

So, how far can we take this in terms of helping devs & sysadmins at the same time? Options I can think of:

 * print a longer statement with help for both groups
 * point to documentation with the above
 * make the re-creation suggestion if a dev build is detected

Other changes that may also be discussible:

 * don't raise the exception and only warn (what will this hide?)
 * print a statement about recreating the directory manually (could get annoying if N recreations are needed)
 * provide a fs command (`fsck`) to find all the directories and optionally fix them

manics:
I think the focus should be on sysadmins since there's a risk of data loss on a production system. If you can cover development errors in the same message _without adding confusion_ then that's great.

joshmoore:
@rgozim : you want to finish this one off for 5.4.2?

rgozim:
We can probably hold off on this until a later release. I really like @joshmoore's suggestion of using build detection. 

On Android it's possible to determine if a jar was built as a particular flavour by generating a `BuildConfig` class at compile time. In theory, I think it should be possible to do something similar when someone runs `build.py build-dev` for our system. However, I am not fully clued up on how are build system is setup, or it's intricacies, so it might be more complicated than following [something like this](https://stackoverflow.com/a/13361409)?

joshmoore:
Ok. Let's discuss options for detection next week and assume this won't make it into 5.4.2. Thanks, @rgozim

mtbc:
> if a dev build is detected

... actually scares me a little: it feels dangerous to me to have OMERO behaving differently for me as a developer from what it does for our users unless I can stay very clear about what those differences are, given that my job is ultimately about crafting our users' experience. Of course such differences are necessary to some extent, just not lightly.

joshmoore:
I would think in this case we'd be _appending_ in the developer case something along the lines of "you may need to re-initialize your repository", something which we are unlikely to want to tell a sysadmin.

manics:
Out of all the other issues devs run into whilst setting up omero does this justify special attention?

joshmoore:
I agree it's not a must to have this information, but on the flip-side: if we had the infrastructure in place, is there any reason to not give extra clues to developers about what might be the problem?

manics:
Only what @mtbc alludes to, i.e. any developer-specific logic risks hiding/causing a bug

jburel:
Probably good to move that the discussion to a card.
Make sure we have a clear picture before changing the message
Happy to capture the discussion and close the PR

jburel:
Card created https://trello.com/c/p5a06pLw/348-error-message-change-on-omero-import
Closing this PR for now

PR #5580, opened 2017-11-20
extend in-group helper methods for Java integration tests

mtbc:
Extend `AbstractServerTest.assertInGroup` to accept the various parameter type combinations.


jburel:
Useful changes 
Merging so we can start using the new method in the tests

PR #5584, opened 2017-11-27
Re-raise exception from integration tests

joshmoore:
Without this change, only the fact that a connection cannot
be established is raised, hiding the original cause. In this
case, that was use of "localhost" rather than "omero" from
etc/ice.config

# Testing this PR
1. Run integration tests with a bad hostname
2. Ensure that a useful error message is printed
3. On ci-master, all tests should remain green


dominikl:
Works like expected, with the PR one can see the details of the failed connection:
```
E       DNSException: exception ::Ice::DNSException
E       {
E           error = 8
E           host = blup
E       }
```
Tests are still working, when a valid host is specified (tested locally for now, will check CI when the merge-build job is finished, too).

dominikl:
`OMERO-DEV-merge-integration-python` looks fine, too. Good to merge from my point of view 👍

PR #5585, opened 2017-11-27
Rename tools/test_setup.py to OmeroPy/src/omero_setup.py

joshmoore:
So that downstream repositories can make use of the PyTest
class defined here, the module will need to be packaged as
part of lib/python.

# What this PR does

Adds `omero_setup` to the release package. Note: this module was not put under an `omero` subdirectory since that requires Ice to be available.

# Testing this PR

1. All builds should continue functioning as normally
2. Additionally, `from omero_setup import PyTest` should work when `OMERO_DIST/lib/python` is on the PYTHONPATH

# Related reading

* https://github.com/ome/omero-cli-render/pull/1


mtbc:
> merge ... into openmicroscopy:clisplit

> All builds should continue functioning as normally

... at which builds ought I be looking? Is there now a devspace?

mtbc (approved):
based on my local testing

PR #5588, opened 2017-11-29
TEMP disable saving changes to User and Group

will-moore:
# What this PR does

This is an investigation into failing robot tests.
I noticed from e.g. https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-robotframework/773/console that after the last test run on Firefox, where the user edits themselves and group, ALL login attempts fail for that user in the subsequent Chrome tests.
I can't reproduce this locally (edit user doesn't break login) so I'm not sure whether this will have any effect on the CI job, but it's worth a try...


jburel:
excluding to see if it makes a difference
more tests passing now

--exclude

PR #5591, opened 2017-11-30
Update swingx to 1.6.1

dominikl:
# What this PR does

Update the version of the swingx library which Insight uses to 1.6.1. This version doesn't have the
`ColumnHeaderRenderer` any longer, so I just removed its usage. The `OMETreeTable` is actually the only place where the `swingx` lib is used. The only implication of this is, that the table headers in Insight are now left aligned instead of centred. In my opinion that's not a big deal. But on the other hand will make the plugin compatible with Fiji again.

# Testing this PR

Replicate the error:
In **Fiji** (must be Fiji, ImageJ works) open OMERO plugin, open a dataset and in the centre panel choose the table view (or open a dataset with more than 100 images, then the table view will be the default). Plugin will crash with ClassCastException.

Replace the OMERO plugin in Fiji with the the merge build. Repeat the workflow. Plugin should work as expected.

Briefly check some other tables in Insight and/or plugin, e. g. the ROI/Folder table and make sure they still work as before.

# Related reading

https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8412
https://trello.com/c/YDyYAUIO/153-outstanding-insight-bugs - Inbox checklist

/cc @pwalczysko @jburel

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/util/ui/treetable/OMETreeTable.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/util/ui/treetable/OMETreeTable.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.util.ui.treetable.TreeTable 
  *
   *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2007 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2017 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
@@ -29,6 +29,7 @@ import java.awt.event.MouseListener;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+
 import javax.swing.DefaultCellEditor;
 import javax.swing.Icon;
 import javax.swing.JCheckBox;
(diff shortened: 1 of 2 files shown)


jburel:
Without this PR when you click on the button "View as list" in the centre panel
The following error is thrown
```
no update needed
Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: sun.swing.table.DefaultTableCellHeaderRenderer cannot be cast to org.jdesktop.swingx.table.ColumnHeaderRenderer
	at org.openmicroscopy.shoola.util.ui.treetable.OMETreeTable.initialize(OMETreeTable.java:118)
	at org.openmicroscopy.shoola.util.ui.treetable.OMETreeTable.<init>(OMETreeTable.java:126)
	at org.openmicroscopy.shoola.agents.dataBrowser.view.ImageTable.<init>(ImageTable.java:403)
	at org.openmicroscopy.shoola.agents.dataBrowser.view.ImageTableView.initComponents(ImageTableView.java:87)
	at org.openmicroscopy.shoola.agents.dataBrowser.view.ImageTableView.<init>(ImageTableView.java:110)

```
With this PR, the problem is fixed and the data can be viewed as a a list

PR #5592, opened 2017-12-06
Quantization follow up

waxenegger:
# What this PR does

Follow-up for: https://github.com/openmicroscopy/openmicroscopy/pull/5501
Having looked into using the above PR in iviewer I noticed that some bits were missing to integrate it seamlessly with the other rendering settings that are already used:
- image_marshal does not include the quantization info, hence the information is missing in the json that's requested
- copy and paste, as well as the rendering settings request did also not include quantization in its logic

# Testing this PR
Check that the tests are green:
- https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_marshal/TestImgDetail/
- https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_rendering/TestRendering/

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8875,6 +8875,8 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                     # 'reverseIntensity' is deprecated. Use 'inverted'
                     'inverted': reverse,
                     'reverseIntensity': reverse,
+                    'family': unwrap(w.getFamily().getValue()),
+                    'coefficient': unwrap(w.getCoefficient()),
                     'rgb': {'red': w.getRed().val,
                             'green': w.getGreen().val,
                             'blue': w.getBlue().val}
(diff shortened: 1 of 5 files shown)


chris-allan:
We're also interested in this for different types of data. Might try and add this API to omero-ms-image-region ASAP for testing.

/cc @emilroz, @knabar

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8875,6 +8875,8 @@ def getAllRenderingDefs(self, eid=-1):
                     # 'reverseIntensity' is deprecated. Use 'inverted'
                     'inverted': reverse,
                     'reverseIntensity': reverse,
+                    'family': unwrap(w.getFamily().getValue()),
+                    'coefficient': unwrap(w.getCoefficient()),
```
jburel, line 8879: shouldn't be ``w.getCoefficient().val``?

chris-allan: That will throw an exception if the value is `None`. Using `unwrap()` with `w.getCoefficient()` is correct.

jburel: thanks
coefficient cannot be null.

jburel:
no failure in devspace, changing target
Will merge tomorrow if everything is green

jburel:
testing passing merging

PR #5595, opened 2017-12-07
Bump omero-marshal to 0.5.2

sbesson:
This release adds support for Masks - see https://github.com/openmicroscopy/omero-marshal/blob/v0.5.2/CHANGELOG.md

To test this PR:
- import a file with masks into a server, e.g `test&masks=15fake` and use the JSON API to confirm the associate mask shapes are encoded
- check the expanded JSON API integration test is still passing

### DIFF
--- a/components/tools/OmeroWeb/requirements-py27-ice35.txt
+++ b/components/tools/OmeroWeb/requirements-py27-ice35.txt
@@ -8,5 +8,5 @@ Django>=1.8,<1.9
 django-pipeline==1.3.20
 gunicorn>=19.3
 
-omero-marshal==0.5.1
+omero-marshal==0.5.2
 django-redis>=4.4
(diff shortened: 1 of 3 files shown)


sbesson:
78131ff extends the `test_api_rois` integration test to include a Mask shape in the ROI. This test should fail with `omero-marshal` 0.5.1 or lower but pass with `omero-marshal` 0.5.2 or later with to the addition of the specific encoders/decoders.

NB: in terms of CI, this test addition will require the environment of the tests to be upgraded accordingly . In devspace, this is done via the job-specific virtualenv https://github.com/openmicroscopy/devspace/blob/0.6.0/home/jobs/OMERO-test-integration/config.xml#L102 but on ci-master this might require manual intervention.

joshmoore:
Merging into next_infra to test on ci-master via gh-5616

PR #5596, opened 2017-12-08
 Projection save

jburel:
# What this PR does

The user can create a projected image if he/she has the correct permissions
This can only be done if ``canEdit`` is true and not ``canAnnotate``
The problem was reported via QA

# Testing this PR

Test 1 insight only
* log as user-3
* Select the read-annotate group
* Add another user from the group
* Open an image with multiple z
* Click on the projection tab
* Check that the ``Project...`` button is greyed out

Test 2 insight only
* log as user-4
* Select the read-write group
* Add another user from the group
* Open an image with multiple z
* Click on the projection tab
* Check that the ``Project...`` button is not greyed out. Click on the button
* Check that a projected image is created

In both cases, the user should still be able to create a projection preview

# Related reading

see https://trello.com/c/2U0ekjwL/58-save-projected-image-rwra

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/actions/ProjectionProjectAction.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/actions/ProjectionProjectAction.java
@@ -63,9 +63,10 @@ public class ProjectionProjectAction
     	if (model.getState() == ImViewer.READY && !model.isBigImage()) {
     		if (model.getSelectedIndex() != ImViewer.PROJECTION_INDEX)
     			setEnabled(false);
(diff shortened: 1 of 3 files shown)


bramalingam:
Tested the workflows as suggested. Everything works as expected. 
Looks good to merge.

On a side note: I tried the workflows on a dv image and the resultant "projected" image from Test2 seems to be in a dv format as well (Atleast going by the name). Is this for real or is it a little misleading to keep the dv extension on this file?

dominikl (commented):
👍 Looks good. As Balaji mentioned, might be worth to strip the file extension for projections, but not necessarily a matter of this PR.

jburel:
The file extension should be fixed in another PR.

PR #5598, opened 2017-12-11
password script param

will-moore:
# What this PR does

If a script ```String``` parameter is named 'password' or 'Password', we use ```type="password"``` in the web script UI to hide input:

See http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2017-December/004093.html

screen shot 2017-12-11 at 13 09 11

# Testing this PR

1. Upload a script with String parameter named "Password" or "password".

1. Should see hidden characters in UI.

### DIFF
--- a/components/tools/OmeroPy/src/omero/testlib/__init__.py
+++ b/components/tools/OmeroPy/src/omero/testlib/__init__.py
@@ -121,8 +121,9 @@ class ITest(object):
             cls.root.setAgent("OMERO.py.root_test")
             cls.root.createSession("root", rootpass)
             cls.root.getSession().keepAlive(None)
-        except:
-            raise Exception("Could not initiate a root connection")
+        except Exception:
+            cls.log.error("Could not initiate a root connection")
+            raise
 
         cls.group = cls.new_group(perms=cls.DEFAULT_PERMS)
         cls.user = cls.new_user(group=cls.group,
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/scripts/include_param.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/scripts/include_param.html
@@ -81,7 +81,8 @@
                 {% if i.number %}
                     <input tabIndex="1" type="text" name="{{ i.key }}" class="{{ i.number }}" {% ifnotequal i.default None %} value="{{ i.default }}" {% endifnotequal %} />
                 {% else %}
-                    <input tabIndex="1" type="text" name="{{ i.key }}" {% ifnotequal i.default None %} value="{{ i.default }}" {% endifnotequal %} />
+                    <input tabIndex="1" name="{{ i.key }}" {% ifnotequal i.default None %} value="{{ i.default }}" {% endifnotequal %}
+                        {% ifequal i.name|lower 'password' %} type="password" {% else %} type="text" {% endifequal %} />
                 {% endif %}
             {% endif %}
         {% endif %}
--- a/history.txt
+++ b/history.txt
@@ -12,7 +12,7 @@ This is a bug-fix release.
 
 Improvements include:
 
--  labelled zoom slider bars in the UI to differentiate from horizontal
+-  labeled zoom slider bars in the UI to differentiate from horizontal
    scrollbars and make clear thumbnails can be zoomed (including Plate and
    Well thumbnails)
 -  fixes for installation walkthrough documentation - installation of script
@@ -36,7 +36,7 @@ Improvements include:
    * fixed the CLI metadata tablestest plugin to not use an empty list of
      Columns
 
-This release also upgrades the version of Bio-Formats which OMERO uses to
+This release also upgrades the version of Bio-Formats that OMERO uses to
 5.7.2.
 
 5.4.0 (October 2017)


jburel:
@will-moore we will need a small example script with the field uploaded to the server to test the UI

jburel:
This PR and #5599 will be tested in a devspace this might have some security issues
I will prepare that in the coming days

not realistic to have them included for 5.4.2 (scheduled for January) due to possible side-effects

will-moore:
Uploaded a script to eel: Util Scripts -> Test Password

joshmoore:
My primary concern is that this implies to the user that the password entry is safe when it is not (or at least not known to be)

will-moore:
@joshmoore I was trying to establish yesterday whether there were any reasons not to go ahead with this and my impression was that there weren't any blockers. Is that not the case?

manics:
Things to consider:
- DEBUG logging on Processor/Blitz/Web, in both normal and error conditions
- Behaviour with insecure 4063 port in the default config
- Misleading behaviour of scripts on older OMERO versions- admins may install this script on an older OMERO and be misled into thinking the password field/namespace is always hidden when it's a feature of new servers only. Could you make this more explicit, e.g. add a `hidden` parameter to `scripts.String()` so that it should fail on older versions, or add a new type `scripts.Password` instead of adding special handling for an existing type?

will-moore:
Hmm - I actually seem to have found more logging of Password script parameter before: https://www.openmicroscopy.org/community/viewtopic.php?t=8094&p=17362#p17368

jburel:
ok so not really a route we can take

will-moore:
Just tested again, and the logging I saw at https://www.openmicroscopy.org/community/viewtopic.php?t=8094&p=17362#p17368 does not happen if you have the default setting:
```
bin/omero config set omero.web.debug False
```
But it would be worth testing if you know you're using passwords, that log directory doesn't contain these anywhere. E.g. enter a password into a script field and then search for it in logs:
```
$ grep my_password -r dist/var/log/
```
Also need to check files created (stdout & stderr).
```
grep my_password -r path/to/OMERO.data.dir/Files/
```
NB: even entering your password into your command line means it will get logged/recorded in history, so best to test this with a fake password, or change your password afterwards to be sure.

jburel:
This will require more scoping we cannot offer such option as part of the main release.
Closing this PR for now
One approach suggested by @mtbc 
in general

joshmoore:
One path would be using NOPASSWD sudo for the `omero-server` user:

```
$ echo "date" >> script.sh
$ chmod a-r script.sh
$ chmod a+x script.sh
$ cat script.sh
cat: script.sh: Permission denied
$ sudo ./script.sh
Fri Dec 15 15:35:13 CET 2017
```

will-moore:
@joshmoore I can see you've got a ```script.sh``` that is executable (by the omero-server user) but not readable. So this could contain password for a user on a remote OMERO.server? But I'm not sure what NOPASSWD means?

joshmoore:
```
# joshuas-mbp:script jamoore$ cat Dockerfile
FROM centos:7

RUN yum install -y sudo
RUN echo "omero-server ALL= (root) NOPASSWD: /opt/command.sh" >> /etc/sudoers
RUN echo "#!/bin/bash"
RUN echo date >> /opt/command.sh
RUN chmod a+x /opt/command.sh
RUN chmod a-r /opt/command.sh

RUN useradd omero-server
RUN echo "omero-server:omero-server" | chpasswd
USER omero-server
CMD /bin/bash
# joshuas-mbp:script jamoore$ docker build -t test .
...
# joshuas-mbp:script jamoore$ docker run -ti --rm test
[omero-server@029e25ca1249 /]$ /opt/command.sh
/bin/bash: /opt/command.sh: Permission denied
[omero-server@dc8030f583df /]$ cat /opt/command.sh
cat: /opt/command.sh: Permission denied
[omero-server@dc8030f583df /]$ sudo cat /opt/command.sh
[sudo] password for omero-server:
Sorry, user omero-server is not allowed to execute '/bin/cat /opt/command.sh' as root on dc8030f583df.
[omero-server@029e25ca1249 /]$ sudo /opt/command.sh
Fri Dec 15 14:53:13 UTC 2017
[omero-server@dc8030f583df /]$ sudo date

We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

[sudo] password for omero-server:
Sorry, user omero-server is not allowed to execute '/bin/date' as root on dc8030f583df.
[omero-server@dc8030f583df /]$
```

joshmoore:
*But* this will do the same thing for all users. Potentially, there could be one such script per OMERO user or group (if necessary).

eilidh-t:
Thanks for your effort with this. Today I looked into storing the password on the user's computer.
I’m thinking of using python [keyring](https://pypi.python.org/pypi/keyring#installing-and-using-python-keyring-lib) to store the login to the remote OMERO server.

Install:
`pip install keyring`

User sets password using this command on the command line:
`keyring set omero username`

Then in the OMERO script the user would only have to enter their username, and the script could look up their password as follows:
```
python
>>> import keyring
>>> keyring.get_password("omero","username")
u'omero_password'
```

This would work, and be fairly convenient, but the security seems to be that if you're logged into the computer where you entered your password then it's accessible. I'll need to discuss this with my team. For example, on the command line:
```
% keyring set omero username
Password for 'username' in 'omero':
% keyring get omero username
omero_password
```

joshmoore:
@eilidh-t: likely you could use the `sudo` trick above to allow a user to *only* set their *own* password in the keyring:

```
ssh -qt remote-system sudo /opt/set-password.sh
```

where `set-password.sh` could use `$SUDO_USER` to determine which user to set the password for.

> Then in the OMERO script the user would only have to enter their username

If possible, I'd _not_ let users choose which username to use the password for but to handle the mapping between OMERO user and operating system user somewhere the users themselves cannot access.

PR #5600, opened 2017-12-12
filter_anns_json_by_ns

will-moore:
# What this PR does

Allows filtering by namespace ```ns``` in webclient/api/annotations
This is needed for Simple FRAP workflow, where we want to load only map annotations with specified namespace into Figure to create labels. See https://github.com/jburel/omero-example-scripts/pull/6

# Testing this PR

1. Add a map annotation to image with a different namespace. E.g. see https://docs.openmicroscopy.org/omero/5.4.0/developers/Python.html but use a different namespace.

2. Using the Image ID, to ```/webclient/api/annotations/?type=map&image=ID``` to see all map annotations.

3. Add ```?ns=your.namespace``` to filter by namespace. Should now only see your map annotation.

4. Test should be passing.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/tree.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/tree.py
@@ -1798,7 +1798,7 @@ def _marshal_exp_obj(experimenter):
 
 def marshal_annotations(conn, project_ids=None, dataset_ids=None,
                         image_ids=None, screen_ids=None, plate_ids=None,
-                        run_ids=None, well_ids=None, ann_type=None,
+                        run_ids=None, well_ids=None, ann_type=None, ns=None,
                         group_id=-1, page=1, limit=settings.PAGE):
 
     annotations = []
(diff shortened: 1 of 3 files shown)


mtbc:
Any way to select only those with no namespace?

mtbc (approved):
Code looks good, functional testing passes.

mtbc:
Maybe worth extending https://docs.openmicroscopy.org/omero/5.4.1/developers/json-api.html a bit.

will-moore:
No, sorry. No way to select only those with no namespace.
The https://docs.openmicroscopy.org/omero/5.4.1/developers/json-api.html docs are for URLs under ```/api/```, which doesn't include this PR.
We don't support annotations there yet - but will try to soon, especially since the ```/webclient/``` urls shouldn't really be considered a public API. Would prefer not to use them for e.g. the OMERO.figure Console examples etc.

mtbc:
https://10.0.51.143:32797/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_tree_annotations/TestTreeAnnotations/test_filter_by_namespace/ is green and stable.

PR #5603, opened 2017-12-13
Handle activities exceptions

will-moore:
# What this PR does

Improve handling of errors for scripts in the Activities panel.
See https://trello.com/c/NsfAOEX6/31-handle-activities-for-invalid-job-id

# Testing this PR

1. To simulate invalid callback string (which fails to find process on server) you can restart the server immediately after starting to run a script.
2. When server has restarted, login to webclient and check Activities panel. Should see error status for script job:

screen shot 2017-12-13 at 11 29 22

3. Other change in this PR is simply changing the logging level for ```proc.getResults()``` and display appropriate message to user. Not sure how to reproduce this error (seen from nightshade - see trello card).

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -3103,17 +3103,22 @@ def activities(request, conn=None, **kwargs):
     new_results = []
     _purgeCallback(request)
 
-    # If we have a jobId, just process that (Only chgrp supported)
+    # If we have a jobId (not added to request.session) just process it...
+    # ONLY used for chgrp dry-run in Chgrp dialog.
     jobId = request.GET.get('jobId', None)
     if jobId is not None:
         jobId = str(jobId)
-        prx = omero.cmd.HandlePrx.checkedCast(conn.c.ic.stringToProxy(jobId))
-        rsp = prx.getResponse()
-        if rsp is not None:
-            rv = chgrpMarshal(conn, rsp)
-            rv['finished'] = True
-        else:
-            rv = {'finished': False}
+        try:
+            prx = omero.cmd.HandlePrx.checkedCast(
+                conn.c.ic.stringToProxy(jobId))
+            rsp = prx.getResponse()
+            if rsp is not None:
+                rv = chgrpMarshal(conn, rsp)
+                rv['finished'] = True
+            else:
+                rv = {'finished': False}
+        except IceException:
+            rv = {'finished': True}
         return rv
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Ah, good catch, @will-moore. I never considered that using the Ice library like this is naturally going to throw different exceptions. Have you `git grep`'d for other uses? In Python, almost any import of a `Prx` class might should be considered suspect.

jburel:
Changing the target so it can be considered for 5.4.2

joshmoore:
Thanks, @jburel. Definitely good to have this in. @will-moore : did you have a chance to look into further such issues or do you want me to capture as a card?

will-moore:
I'll add a try/catch around one more for ```chgrp``` handling in the same method.
Don't know about Insight etc or where to look for that?

joshmoore:
In Java, the Ice exceptions must be explicitly caught: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/insight/SRC/org/openmicroscopy/shoola/env/data/ConnectionExceptionHandler.java#L100

rgozim:
Am I testing this wrong? I selected multiple images and ran an analysis script on them and whilst it was running, restarted the server. I got the following error when logging back, and after pressing the notification button 

```
UnknownLocalException at /webclient/activities/
exception ::Ice::UnknownLocalException
{
    unknown = Network.cpp:2357: Ice::ConnectionRefusedException:
connection refused: Connection refused
}

Request Method: GET
Request URL: http://127.0.0.1:8000/webclient/activities/?_=1516121374998
Django Version: 1.8.18
```

screen shot 2018-01-16 at 16 50 11

will-moore:
Tried this locally against today's
```
* cb453e3af9 (HEAD, snoopycrimecop/develop/merge/daily) merge develop -Dorg -Ssuccess-only
```
and it "works for me".
Happy to push this to 5.4.3 since the requirement is coming from me, not external.

joshmoore:
I've reproduced locally with the `bin/omero admin restart` workflow outlined in the description:

screen shot 2018-01-18 at 15 49 51

The bottom "No processor" came with an error report; the upper did *not*. Personally, I find this quite a nice improvement.

The code looks generally sensible, so merging for 5.4.2. I do think there's room for expansion here that we can discuss either for 5.4.3 or 5.x.x:

 * Should all `except` blocks log _something_ (even if at DEBUG)?
 * Should different Ice.Exceptions be handled differently?
 * Should non-Ice.Exceptions _also_ be handled?
 * Are there other locations needing similar handling?
 * Could that be done by a more generic decorator?

PR #5607, opened 2018-01-05
have import fail fast if into illegal container

mtbc:
# What this PR does

For when a user tries to import into a container that they do not have permissions for, this PR makes the import failure faster and clearer.

# Testing this PR

Try importing into, say, somebody else's dataset in a read-annotate group and make sure that file upload does not even start.

https://web-proxy.openmicroscopy.org/next-ci/job/OMERO-test-integration/lastCompletedBuild/testngreports/integration/ImporterTest/testImportImageIntoOthersDataset/

# Related reading

https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8054&p=17168#p17168
https://trello.com/c/dFxZnrag/65-failed-import-on-non-user-dataset


dominikl:
Tested it with CLI import, works like expected: Can import into own read-annotate datasets, but fails immediately when trying to import into other user's read-annotate dataset, before starting upload/import. Good to merge 👍

PR #5608, opened 2018-01-05
extend CLI admin email timeout

mtbc:
# What this PR does

Sending e-mail to a user of our demo server takes around 15s. This PR extends the timeout for `bin/omero admin email ...` so as to be more patient with such slow servers.

# Testing this PR

Use the CLI to send e-mail to a demo server user and check that the e-mail is received and `omero::LockTimeout` is not thrown at you. (I am `mtbcarroll` on that server and I am happy to make anyone else an account on it.) Can also try out the `--wait` option.

# Related reading

https://trello.com/c/bSyOy8Bh/34-cli-or-email-plugin-5s-timeout

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1391,7 +1391,7 @@ present, the user will enter a console""")
 
         try:
             cb = client.submit(
-                req, loops=10, ms=500,
+                req, loops=50, ms=500,
                 failonerror=True, failontimeout=True)
         except omero.CmdError, ce:
             err = ce.err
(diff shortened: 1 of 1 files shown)


joshmoore:
See `_add_wait` from cli.py: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/cli.py#L847

mtbc:
Ah, so close this PR and add `--wait` facility to the `email` subcommand?

joshmoore:
Keeping this as a bump of the default timeout makes sense, too.

dominikl:
Looks good, works like expected. Good to merge 👍

joshmoore:
@kennethgillen  : If there's anywhere in particular you think this should be advertised, please let us know.

kennethgillen:
It should probably be an example somewhere on https://docs.openmicroscopy.org/latest/omero/sysadmins/index.html

I missed this myself. Might have been worth a "sysadmin update" in the release notes.

PR #5609, opened 2018-01-08
fix big plate thumbs

will-moore:
This fixes slow loading of thumbnails on big plates, caused by slow jQuery building of plate Tables.

This was originally reported as slow thumbnail loading, https://trello.com/c/2fQFE3Lb/33-bug-thumbnail-caching, since the jQuery DOM manipulation happens right before thumbnail loading. Therefore, the responses from  ```webgateway/get_thumbnail/``` are not downloaded while DOM layout is still processing, making it appear that the requests themselves were taking a long time.

This was exacerbated on a plate of 20 columns and 35 rows (700 Wells), since we create a much larger table of 96 columns and 64 rows (6144 Cells)!

Now, instead of appending each ```<td>``` individually, we build the entire table html string before appending to the table in a single call.

# Testing this PR

1. User-3 on web-dev-merge: test loading of thumbnails on Damir's large plate.

2. Need regression testing of related plate-loading functionality: Load plate, click on Wells to load images below, switch Field, zoom plate then load another plate (should load at same sized thumbnails) etc.

Some functionality of the old ```for loop``` code has not been reproduced in the new code since we don't iterate through each thumbnail `````` as we did before so we don't have each in hand:

1. Triggering of the ```thumLoad``` event for every thumbnail image:
```
_this.self.trigger('thumbLoad', [$(this).parent(), $(this)]);
```
We could likely reproduce this behaviour by an additional ```$("img.loading", table).forEach()``` loop, but I can't see that this is used or tested anywhere so don't know if it's worth preserving?

2. Setting of ```wellpos``` data on each ``````. Can't see that this is used anywhere either:
```
.data('wellpos', data.rowlabels[i] + data.collabels[j]);
```

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.plateview.js
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.plateview.js
@@ -137,12 +137,14 @@ jQuery._WeblitzPlateview = function (container, options) {
     table.addClass('showWellLabel wellSize' + opts.width);
 
     var imgIds = [];
+    var html = "";
+    // Build table html and append below
     for (i=0; i < data.rowlabels.length; i++) {
-      tr = $('<tr></tr>').appendTo(table);
-      tr.append('<th>'+data.rowlabels[i]+'</th>');
+      html += '<tr>';
+      html += '<th>'+data.rowlabels[i]+'</th>';
       for (var j=0; j<data.grid[i].length; j++) {
         if (data.grid[i][j] === null) {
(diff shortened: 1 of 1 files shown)


mtbc:
This does seem to fix the regression. :+1:

will-moore:
Thumbnail loading certainly improved on web-dev-merge:

screen shot 2018-01-09 at 09 26 54

pwalczysko:
Cannot detect any regressions here. The speed of loading is good.

joshmoore:
Thanks all!

PR #5606, opened 2018-01-05
robot tests improvement

jburel:
This PR uses
 * ``robotframework-selenium2library==3.0.0``
* ``robotframework==3.0.2``

This PR:
* removes usage of deprecated keywords
* removes usage of reserve syntax
* splits the tests in smaller tests
* splits some test suite i.e. file into smaller suite
* adds new tasks to allow rerun failing tests only or failing suites
* adds new tasks to merge results
* adds new tasks to run the script in parallel using ``robotframework-pabot``
* re-activates commented out tests
* review search tests. Test not testing UI functionality will be converted in an integration test in another PR

Note that the number of tests has increased. The number in the job
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-robotframework/ was the aggregate of 2 runs.

We could split a bit more the tests but that could be done in follow up PRs
or improve passing one e.g. removing usage of Sleep (not great way to synch)
The ``robotframework-pabot`` is still in its early development

Devspace running at https://10.0.51.135:32784/

# Related PRs
 * https://github.com/openmicroscopy/ansible-role-devspace/pull/7
 * https://github.com/openmicroscopy/devspace/pull/88

# Related reading

https://trello.com/c/QgTG93Ow/1-devspace

cc @will-moore @rgozim @bramalingam

PR #5611, opened 2018-01-11
robot test fixes workflows

will-moore:
# What this PR does

Attempts to fix failure of this robot test:
https://10.0.51.135:32776/job/OMERO-robot/23/robot/Web%20&%20Web/Web_1/Webadmin%20User%20Settings/Owner%20Group%20Edit/
When testing Group Edit manually, "Save" takes some time. Seems robot test is asserting the location change immediately. This PR waits for this to succeed.

NB: The failure of the test at this point on Firefox means that the permissions of the group do not get reset, so the same test Subsequently fails on Chrome because the permissions are wrong:
https://10.0.51.135:32776/job/OMERO-robot/23/robot/Web%20&%20Web/Web/Webadmin%20User%20Settings/Owner%20Group%20Edit/

# Testing this PR

1. Both tests above should be passing.

cc @jburel

### DIFF
--- a/components/tests/ui/testcases/web/webadmin_user_settings.txt
+++ b/components/tests/ui/testcases/web/webadmin_user_settings.txt
@@ -64,9 +64,9 @@ Owner Group Edit
     Submit Form                     css=form.settings_form
 
     # Return to edit group, to check permissions were saved and reset
-    Location Should Be          ${WEBADMIN WELCOME URL}myaccount/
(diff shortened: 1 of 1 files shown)


jburel:
Cherry-picked the commit
Now in https://github.com/openmicroscopy/openmicroscopy/pull/5606
In case we need to change other files to avoid conflict
The test is passing with the change

PR #5613, opened 2018-01-12
remove product.license and product.final.name from config

mtbc:
# What this PR does

Removes a couple of long-unused properties from `etc/omero.properties`.

# Testing this PR

CI should remain green.


joshmoore:
All is green and I don't see anything via `git grep` either. :+1:

PR #5593, opened 2017-12-06
tables: retry repo_uuid lookup

joshmoore:
This adds on to 12646428a3f44908ae5451c88db710bebe32de94 (gh-5481)
but not just adding a wait to the .omero/repository lookup but as
well to the .omero/repository/*/repo_uuid lookup, so that the 20
REPEATS don't progress within 2-3 seconds but instead take long
enough for the integration server has time to configure itself

cF. https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-training/811/

# Testing this PR

1. table tables (incl. merge-training) should be green for an extended period
2. setting `omero.repo.wait` to a longer period (in MT) should help alleviate any issues

PR #5615, opened 2018-01-12
tables: use omero.repo.wait rather than SECONDS_SLEEP

joshmoore:
The previous loop checked for the elapsed time
to see whether waiting was necessary. The with-
wait method doesn't have access to the elapsed
time and so should run unconditionally.

# Testing this PR

Testing is difficult since tables initialization needs to fail. *If* it fails, then several seconds should be waited. I tested locally by modifying https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/unit/tablestest/test_servants.py#L27 and causing the method to fail.

# Related reading

See https://github.com/openmicroscopy/openmicroscopy/pull/5593

### DIFF
--- a/components/tools/OmeroPy/src/omero/tables.py
+++ b/components/tools/OmeroPy/src/omero/tables.py
@@ -28,7 +28,6 @@ tables = __import__("tables")  # Pytables
 
 VERSION = '2'
 RETRIES = 20
-SECONDS_SLEEP = 5
 
 
 def slen(rv):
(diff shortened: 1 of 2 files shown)


joshmoore:
Opening this because it _should_ prevent training etc. from failing. However, `omero.repo.wait` is no longer being used which means this needs some cleaning/simplifying.

rgozim:
Works fine from running local tests.

PR #5621, opened 2018-01-17
Close fileinput

dominikl:
# What this PR does

Implements @mtbc suggestion to close the `fileinput` in `obj.py` plugin after parsing.

Should fix the failed integration tests: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/782/testReport/ 

I couldn't replicate the failed integration tests locally, so couldn't check if that really solves the issue. But we'll see tomorrow on CI.

# Testing this PR

Check python integration tests.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -560,8 +560,8 @@ Bash examples:
                         line = line.strip()
                         if line and not line.startswith("#"):
                             actions.append(self.parse(state, line=line))
-                except Exception:
-                    self.ctx.die(999, "Cannot read file")
+                except IOError:
+                    self.ctx.die(337, "Cannot read file")
                 finally:
                     if fi is not None:
                         fi.close()
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -553,11 +553,18 @@ Bash examples:
         if not args.command:
             if args.file:
                 path = args.file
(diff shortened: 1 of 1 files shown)


components/tools/OmeroPy/src/omero/plugins/obj.py
```diff
@@ -557,6 +557,7 @@ def process(self, args):
                     line = line.strip()
                     if line and not line.startswith("#"):
                         actions.append(self.parse(state, line=line))
+                fileinput.close()
```
joshmoore: Perhaps in a try/finally block? I looked for best-practices. Perhaps cF. http://www.zopatista.com/python/2013/11/26/inplace-file-rewriting/

dominikl:
For the `self.ctx.die(...)`, is there a list of error codes somewhere? No idea what to use there.

joshmoore:
> For the self.ctx.die(...), is there a list of error codes somewhere?

No. Just try to pick similar but unique values as in the current file.

joshmoore:
Travis & ci-master are green and the code looks good. I think though in a follow up which reviews all uses of `fileinput` we could make use of a `with`:

https://docs.python.org/3/library/fileinput.html

```
with fileinput.input(files=('spam.txt', 'eggs.txt')) as f:
    for line in f:
        process(line)
```

since [__exit__](https://github.com/python/cpython/blob/master/Lib/fileinput.py#L244) calls close regardless of any exceptions. In our case, that'd likely suggest:

```
try:
    with fileinput... as f:
        for line in f:
            ...
exception ...:
    self.ctx.die(...)
   
```

PR #5620, opened 2018-01-17
fix Java emory import

mtbc:
# What this PR does

Fixes an attempted import of `java.util.Collections`.

# Testing this PR

`build-java` *may* pass in Travis.
--depends-on #5619

PR #5619, opened 2018-01-17
fetch artifacts from repos over HTTPS, not HTTP (rebased from develop)

mtbc:
# What this PR does

Has Ivy fetch third-party artifacts over HTTPS rather than simple HTTP.

# Testing this PR

OMERO should build fine even *without* a local maven cache already populated.

# Related reading

https://github.com/openmicroscopy/bioformats/pull/2992
--rebased-from #5569
--depends-on #5620

PR #5622, opened 2018-01-17
travis fixes

mtbc:
Combines #5619 and #5620 as `--depends-on` no longer works.


joshmoore:
`     [exec]         ERROR: Ice version mismatch!`

mtbc:
Yes, no idea what that is; this PR fixes other issues. If there's a fixing commit elsewhere then happy to cherry-pick it.

joshmoore:
Keeping this green has been a constant battle. Happy for this to go in without the Ice fix; I don't have one off-hand.

sbesson:
Successfully tested locally while testing a change to `omero.properties`. Agreed it would be nice to restore Travis but the cost of doing it might simply be too high given the fact our next large goal will be to kill this custom branch.

PR #5636, opened 2018-01-25
script to work with test-infra

jburel:
# What this PR does

Move upstream methods required to work with test-infra
Changes needed to handle non official script

# Testing this PR

Check that travis is green

# Related reading

See https://github.com/ome/omero-figure/pull/259

cc @joshmoore

### DIFF
--- a/components/tools/OmeroPy/test/integration/scriptstest/script.py
+++ b/components/tools/OmeroPy/test/integration/scriptstest/script.py
@@ -38,6 +38,13 @@ class ScriptTest(ITest):
             return -1
         return script.getId().getValue()
 
+    def get_script_by_name(self, path, name):
+        script_service = self.root.sf.getScriptService()
+        script = _get_script_by_name(script_service, path, name)
+        if script is None:
+            return -1
+        return script.getId().getValue()
+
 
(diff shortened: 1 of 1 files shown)


joshmoore:
No immediate objections. I begin to wonder if these should be pip installable though so we don't tie ourselves so tightly to the OMERO version.

cc: @sbesson

jburel:
having things pip installable will be certainly be better.
For now the same changes are in the figure repo (similar to scripts and CLI plugins)

sbesson:
Is anything within the OMERO repository still consuming these classes? If the only place where we are using this test API are the downstream components, I can totally conceive:

OMERO (ITest) -> {omero-test-{script,...}} -> omero-test-infra -> omero-figure, scripts

to allowing ourselves to make backwards-compatible test API additions as needed without the need to release a full OMERO stack.

joshmoore:
@jburel : is having OMERO **not** depend on `omero-dev-pytest` (or whatever it's named) ok with you?

jburel:
@joshmoore that will work.

jburel:
Closing this PR since we will be looking into other option

PR #4927, opened 2016-11-07
idr-0.3.0 release

joshmoore:
# What this PR does

Bump Bio-Formats version to use the newly deployed "0.3.0" tag in the "idr" group. (Note: the tag has only been pushed to http://github.com/idr to prevent issues with `git describe`)

cc: @sbesson 

# Testing this PR

1. The regular devspace builds should continue to function normally.
2. Additionally, local builds should be possible _without_ building a 5.1.11-SNAPSHOT

PR #5641, opened 2018-01-31
bioformats: make org in ivy files configurable

joshmoore:
In order to support forks of Bio-Formats, not just
the version number but also the organization name
itself needs to be configurable. This was previously
done for the IDR fork of Bio-Formats by hard-coding
`org="idr"`.

see: gh-4927

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -915,10 +915,15 @@ Ice.IPv6=1
 #############################################
 product.name=OMERO.server
 
+
 #############################################
 ## Library versions
 #############################################
+
+org.bioformats=ome
+versions.bioformats=5.7.3
 versions.ome-java=2007-Aug-07-r3052
+
 ##
 versions.JHotDraw=7.0.9
 versions.TableLayout=bin-jdk1.5-2009-08-26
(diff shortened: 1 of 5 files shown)


etc/omero.properties
```diff
@@ -915,10 +915,15 @@ Ice.IPv6=1
 #############################################
 product.name=OMERO.server
 
+
 #############################################
 ## Library versions
 #############################################
+
+org.bioformats=ome
```
sbesson, line 923: Is this block separating our maintained dependencies from external dependencies?

sbesson (approved):
Assuming all the tests are passing, the proposed way to make the Bio-Formats organization configurable via property/command-line looks reasonable and potentially extensible to other dependencies.

etc/omero.properties
joshmoore: That was my feeling. It "felt wrong" to have an unalphabetized section:

```
...
org.bioformats=
versions.bioformats
...
```

but it also felt wrong not to have them together. When I thought about having a maintained section, then it started getting complicated (omero.slice, etc) so I tried to keep it simple.

rleigh-codelibre:
Looks fine to me.

etc/omero.properties
sbesson: I had the same feeling while thinking at the list and the maintained section.  As this is largely internal and the alphabetical order is not strictly respected, I'd say either integrate as this is or putting these 2 lines consecutively into the main list are completely acceptable choices for the time being.

joshmoore:
Thanks, all. I'm going to go ahead and get this in so I can open up an integration branch making use of. Happy to address re-orderings, etc. in a follow-up for 5.4.x if desired.

PR #5647, opened 2018-02-06
Projection no channels

jburel:
# What this PR does

Handle case in renderProjectedMethod when all the channels are turned off 

# Testing this PR

Check that that integration tests are green

# Related reading
https://trello.com/c/kn5FWGWv/44-bug-webgateway-renderimage-maxint-with-channels-all-off
cc @will-moore

### DIFF
--- a/components/server/src/ome/services/RenderingBean.java
+++ b/components/server/src/ome/services/RenderingBean.java
@@ -609,6 +609,9 @@ public class RenderingBean implements RenderingEngine, Serializable {
                     projectedSizeC += 1;
                 }
             }
+            if (projectedSizeC == 0) {
+                return new int[0];
+            }
             Pixels projectedPixels = new Pixels();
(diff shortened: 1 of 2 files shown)


will-moore:
I'd expect to see a blank image with no active channels and projection, same as without projection.
Otherwise web/insight will still show an error.

will-moore:
Works fine for me in web now. Renders as a black image as expected.

pwalczysko:
It would have been nice to fix the whole workflow here. After clicking ``Project``, I have still an error-dialog experience (the error message does not explain why). I would suggest to disable the ``Project`` button when all channels are off. See line 2 of https://docs.google.com/spreadsheets/d/1lI4jVmocE-DuQhxS_EMqjcI2UeNrAoIq1TgM5HlIPnY/edit#gid=420881033

PR #5649, opened 2018-02-07
Bio-Formats: rollback to idr:*:0.4.4

joshmoore:
The simplest path to having a running IDR on OMERO 5.4
is to use the version of Bio-Formats in the current IDR.
Unfortunately, API differences in MetadataOptions require
some hard-coding. This commit may end up being temporary
until we can update the IDR version of Bio-Formats.


joshmoore:
This change has been frequently tested on `test50`. Memoizer files do *not* need re-generating. Merging to begin the process of putting together a first milestone build for prod50.

Note: an alternative to this PR would have been to modify the names of the classes on https://github.com/openmicroscopy/bioformats/tree/metadata54 such that they match the mainline. If we can't move to Bio-Formats 5.8+ soon, then this might be the next viable step.

PR #5650, opened 2018-02-07
Insight bug fixes

dominikl:
# What this PR does

Some bugfixes for Insight.

- If an image has been open for a while ( > 10min) in the image viewer and you try to save modified rendering settings when closing the image viewer an error dialog popped up (but the rendering settings are saved). Reported by @mporter-gre .
- If an image is open in the image viewer and the server is restarted, the rendering engine did not get reloaded, and on every change of the rendering settings (e.g. switch a channel on/off), Insight "thought" it lost connection to the server and tried to reconnect again. This was unfortunately a side effect of #5513 . Reported by @jburel on 5.4.2 release testing ( row 9 https://docs.google.com/spreadsheets/d/1iRCuIDLTA0k7QeZK1xLwSXbYdirUoNIpNXnDLBbdQLE/edit#gid=420881033 ).

I leave this PR open and gradually add some more bugfixes. As Insight is in maintenance mode this will be either critical ones or easy-to-fix ones. 

# Testing this PR

Basically try to replicate the bugs described in above section.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2017 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2018 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -387,6 +387,9 @@ class OMEROGateway
 	
 	private Gateway gw;
 	
+    // Keep track of the rendering engines currently in use
+    private Map<SecurityContext, Set<Long>> renderingEngines = new HashMap<SecurityContext, Set<Long>>();
+	
 	/**
 	 * Creates the query to load the file set corresponding to a given image.
 	 *
@@ -514,7 +517,17 @@ class OMEROGateway
 	void closeService(SecurityContext ctx,
             StatefulServiceInterfacePrx svc)
     {
-        if (ctx == null || svc == null) return;
+        if (ctx == null || svc == null)
+            return;
+        if (svc instanceof RenderingEnginePrx && renderingEngines.containsKey(ctx)) { 
+            try {
+                renderingEngines.get(ctx).remove(
+                        ((RenderingEnginePrx) svc).getPixels().getId()
+                                .getValue());
+            } catch (ServerError e) {
+                // do nothing
+            }
+        }
         gw.closeService(ctx, svc);
     }
 
@@ -1602,7 +1615,7 @@ class OMEROGateway
 	 */
 	Map<SecurityContext, Set<Long>> getRenderingEngines()
 	{
-		return gw.getRenderingEngines();
+		return renderingEngines;
 	}
 
 	boolean joinSession() {
@@ -2397,6 +2410,12 @@ class OMEROGateway
 			service.lookupPixels(pixelsID);
 			needDefault(pixelsID, service);
 			service.load();
+            Set<Long> pixIds = renderingEngines.get(ctx);
+            if (pixIds == null) {
+                pixIds = new HashSet<Long>();
+                renderingEngines.put(ctx, pixIds);
+            }
+            pixIds.add(pixelsID);
 			return service;
 		} catch (Throwable t) {
 		    log(t.getMessage());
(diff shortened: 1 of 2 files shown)


components/insight/SRC/org/openmicroscopy/shoola/agents/metadata/rnd/RendererModel.java
```diff
@@ -960,6 +960,7 @@ void setColorModel(String colorModel)
 		if (rndControl == null) return;
 		makeHistorySnapshot();
 		rndControl.setModel(colorModel);
+		System.out.println("jep!");
```
joshmoore: Really? :)

dominikl: Oops :-)

sbesson: feature!

components/insight/SRC/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -2397,6 +2410,12 @@ private RenderingEnginePrx generateRenderingEngine(
 			service.lookupPixels(pixelsID);
 			needDefault(pixelsID, service);
 			service.load();
+            Set<Long> pixIds = renderingEngines.get(ctx);
```
jburel, line 2413: for big images several rendering engines are created for the same set of pixels
Have you considered that case?

dominikl: Shouldn't be a problem, works like before. Only the pixels Id is needed to reload the RenderingEngines, it's only used here: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/insight/SRC/org/openmicroscopy/shoola/env/data/DataServicesFactory.java#L365 (respectively bit further down at line 388 to reload the RenderingControl for the pixels Id (which should take care of the different RenderingEngines)).

pwalczysko:
Note: if further fixes are to be implemented here as the header suggests, then https://trello.com/c/0c3vG5rs/49-insight-failed-login-message

pwalczysko:
Testing of 

>f an image has been open for a while ( > 10min) in the image viewer and you try to save modified rendering settings when closing the im

all works fine here, tested on Win7 laptop and Mac, small and big images.

pwalczysko:
Testing of

>If an image is open in the image viewer and the server is restarted, the rendering engine did not get reloaded, and on every change of the rendering settings (e.g. switch a channel on/off), Insight "thought" it lost connection

Did the tests locally on my localhost, as described in step 3 of https://docs.openmicroscopy.org/internal/testing_scenarios/ConnectionErrorHandling.html. I am a bit confused about the report because it says that the step 3 was tested, but it says "when a dataset is expanded". There is only a mention about full viewer and switching channels on and off in step 3. 

I did:

  - open a big or small image in Full Viewer (identical results obtained for both)
  - restart the server 
  - switch a channel on and off in the still opened Full Viewer
  
Result:
   
  - an intermittent "Trying to reconnect..." window popped up, but vanished after couple of seconds
  - the initial action was not executed (i.e. the channel switch-off which elicited the dialog got not executed) but every following action on the same Full Viewer was performed as expected

Another test:

   - expand a project
   - restart the server
   - expand a dataset -> result is that this just works, no popup or detrimental behaviour observed.

pwalczysko:
In summary:
The first test https://github.com/openmicroscopy/openmicroscopy/pull/5650#issuecomment-365586928 -> all as expected.
Although some question about the precise setup of https://github.com/openmicroscopy/openmicroscopy/pull/5650#issuecomment-365594600 remain, the behaviour seems to me as fully satisfactory.

Not sure if @dominikl wants to add more fixes to this PR, (such as https://github.com/openmicroscopy/openmicroscopy/pull/5650#issuecomment-365586620), but if not, ready to merge FMPOV.

dominikl:
Thanks @pwalczysko . The steps for testing point 2 were absolutely correct. Without the bugfix Insight would hang in the "Trying to reconnect..." loop forever. As you noticed there's still the (imho) minor problem, that the first rendering setting action on an open image after a server restart is not executed (fails silently), but that would be quite tricky to fix.

The failed login error message shouldn't be difficult to fix. Not sure if it's already too late for 5.4.4? I'll open a separate PR anyway, let you decide if it'll go in to 5.4.4 or later.

jburel:
Connection message handled in another PR https://github.com/openmicroscopy/openmicroscopy/pull/5658

PR #5651, opened 2018-02-08
Javagateway bugfixes

dominikl:
# What this PR does

A bugfix and an RFE for the Java gateway. Would be nice to get it into 5.4.4, if possible?

RFE (first commit): Very simple RFE, but requested a few times: "Why do you have to request each single pixel value from a Plane2D object? I just want the whole thing". The commit simply adds a getter for the pixel values array. /cc @bramalingam 

Bugfix (second commit): When using the `LoginCredentials(String[] args)` constructor it wasn't possible to switch the group (create a connector for a different `SecurityContext`), just noticed that when trying the workflow https://github.com/openmicroscopy/ome-documentation/pull/1844 .

PR #5652, opened 2018-02-08
Sudo training example

dominikl:
See https://github.com/openmicroscopy/ome-documentation/pull/1844

(needs/includes #5651 , only last commit relevant for this PR)


dominikl:
I was assuming the training job runs as root (or other admin account), but apparently it's not, SecurityViolations in https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/866/console .

jburel:
2 new users are created during the set up.
They are standard users

examples/Training/java/src/training/util/BytesConverter.java
```diff
@@ -1,71 +0,0 @@
-/*
```
bramalingam, line 1: Is this meant to be a breaking change? As in, even if we are not intending on using these methods: we could potentially be breaking user workflows, if they are using the BytesConverter.

dominikl: I don't think so. These classes are also part of the Gateway: https://github.com/openmicroscopy/openmicroscopy/tree/develop/components/blitz/src/omero/gateway/rnd . I think when I started the Gateway I copied them over, but just forgot to remove them from the training examples.

jburel: they were for convenience since they were not easily accessible previously.
No need  to keep them in training now they are in the gateway

jburel:
Waiting for tomorrow build since it was red today see https://github.com/openmicroscopy/management_tools/pull/669

examples/Training/java/src/training/util/BytesConverter.java
waxenegger: an absolutely minor point since most people would use the relevant code bits from the example but the second argument for the Sudo instantiation in the main method should be sudoUsername

jburel:
@dominikl the parent PR has now been merged

jburel:
``
Logged in as ladmin_user
Added 'training_user's tag' to dataset MyDataset on behalf of training_user
Added 'ladmin_user's tag' to dataset MyDataset as admin.
``
Everything looks good

PR #5654, opened 2018-02-12
Handle bot get requests

will-moore:
# What this PR does

Handles various GET requests that are not used by webclient but we see errors from bots trying to use these on nightshade.
See https://trello.com/c/FYWseOU8/23-get-webclient-action-addnewcontainer
Now, instead of raising Errors that will be reported we simply return a 'Not supported' response.

# Testing this PR

To test, visit various urls in your browser:

 - /webclient/action/addnewcontainer/
 - /webclient/load_searching/form/
 - /webclient/chgrp/
 - /webclient/load_history/
 - /webclient/api/links/

All should return a suitable message, mostly saying that you need to use ```POST``` instead of ```GET```.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/urls.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/urls.py
@@ -88,7 +88,7 @@ urlpatterns = patterns(
     # load history
     url(r'^load_calendar/(?:(\d{4})/(\d{1,2})/)?$', views.load_calendar,
         name="load_calendar"),
-    url(r'^load_history/(?:(\d{4})/(\d{1,2})/(\d{1,2})/)?$',
+    url(r'^load_history/(\d{4})/(\d{1,2})/(\d{1,2})/$',
         views.load_history, name="load_history"),
 
     # load search
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -1396,7 +1396,10 @@ def load_searching(request, form=None, conn=None, **kwargs):
     # form = 'form' if we are searching. Get query from request...
     r = request.GET
     if form is not None:
-        query_search = r.get('query').replace("+", " ")
+        query_search = r.get('query', None)
+        if query_search is None:
+            return HttpResponse("No search '?query' included")
+        query_search = query_search.replace("+", " ")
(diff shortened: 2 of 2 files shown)


will-moore:
I was going to add handling of more similar errors from https://trello.com/c/O9zFRzPT/12-loadchgrpgroups-keyerror but this got me thinking:
It seems that most of the errors we're seeing here were generated by bots or by some security testing e.g. https://www.openmicroscopy.org/qa2/qa/feedback/18338/ trying to access various URLs in ways that normal users wouldn't be able to (via clicking on webclient UI).
If we disable all error handling / reporting for these cases, then any bug in the webclient UI that causes the webclient to try and access e.g. GET webclient/action/addnewcontainer/ would fail silently and give the user no way to report the bug.

So, although handling these cases silently is nice for us (don't get so many e-mails and QAs) I'm worried that we may also miss useful bug reporting.
Thoughts? cc @kennethgillen @joshmoore @jburel

joshmoore:
It seems that minimally saying, "this method is POST only" is a more useful response than exploding. What type of situation are you foreseeing wanting debugging of?

will-moore:
OK, fair enough. I was thinking of JavaScript bugs where e.g. we somehow miss adding an essential argument to the request, e.g. don't add groupId to chgrp POST https://www.openmicroscopy.org/qa2/qa/feedback/18338/ but it seems that we catch most of these in testing and it's quite rare to have a user report these via QA. (I'm pretty sure that example was an automated bot / security testing in some way and it's not a webclient bug).

Last commit is tricky to test, (from https://trello.com/c/O9zFRzPT/12-loadchgrpgroups-keyerror) code should make sense.

rgozim:
Working on web-dev-merge but not my local build. I assume it's because I'm missing some other PR(s) that might be required for this work locally.

jburel:
``/webclient/load_history/`` returns a 404

will-moore:
@jburel That's expected (see https://github.com/openmicroscopy/openmicroscopy/pull/5654/commits/4ce8a8358500c3f8a8c96dc57af1ad8f3502d995) I think that's the correct behavior - just didn't mention it in the description.

jburel:
@will-moore thanks.

jburel:
Everything seems to work as expected
I will merge tomorrow unless someone still has questions/concerns

PR #5657, opened 2018-02-14
admin fixpyramids

jburel:
# What this PR does

Fix bugs found when working on the pyramid fix
Allow users with correct privileges to run command

bug was introduced during the lightadmin work
# Testing this PR

Run as an admin  ``bin/omero admin fixpyramids /OMERO --dry-run``
This should  no longer return an error

 cc @pwalczysko @mtbc

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -931,7 +931,7 @@ present, the user will enter a console""")
         else:
             self.ctx.call(command)
 
-    @admin_only()
+    @admin_only(AdminPrivilegeReadSession)
     @with_config
     def fixpyramids(self, args, config):
         self.check_access()


components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -931,7 +931,7 @@ def ice(self, args):
         else:
             self.ctx.call(command)
 
-    @admin_only()
+    @admin_only(AdminPrivilegeReadSession)
```
mtbc, line 934: This is a rather strong privilege. What error was being returned -- what powers are actually necessary? (Haven't found the card yet.)

mtbc: Ah -- it's by analogy with @joshmoore's caution with `cleanse`?

jburel: similar to cleanse

mtbc: Okay, no objections if the caution's warranted.

jburel: at the moment nobody can run it including root

mtbc: Ahh, I'm afraid I don't recall the permissible ways to write `@admin_only` and never understood the implementation -- @joshmoore, could the bug be in the decorator? Not blocking this PR anyway.

joshmoore: I think this could be achieved with `@admin_only(full_admin=True)` *however* that appears to be the default, which doesn't explain why this was failing previously.

mtbc (approved):

joshmoore:
I think the discussion above now has me worried that this is hiding something deeper.

jburel:
Something to look at in another PR

joshmoore:
Opened in https://github.com/openmicroscopy/openmicroscopy/pull/5661

PR #5658, opened 2018-02-14
Provide better error message for the user on version mismatch

dominikl:
# What this PR does

With this PR the user gets a better error message, in case he connects with an incompatible client. Before the dialog showed "Check username and password", which is obviously wrong and misleading.

screen shot 2018-02-14 at 14 57 08

# Testing this PR

Connect with a incompatible Insight/Importer to a server with a higher or lower version number.

Also check other error conditions (wrong server, port, username/password) and make sure they still show appropriate error messages.

# Related reading

https://trello.com/c/0c3vG5rs/49-insight-failed-login-message

Too late for 5.4.4 @jburel @pwalczysko ?

--

By the way, a very quick and easy method to spin up OMERO.servers with different versions (thanks @sbesson  !): Clone `https://github.com/openmicroscopy/omero-test-infra`, change the `OMERO.version` in `.env`, run `docker-compose up`.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
@@ -186,7 +186,8 @@ public class LoginServiceImpl
                     if (cause.getCause() instanceof SecurityViolation)
                         failureIndex = ACTIVE_INDEX;
                 }
-                else if (exception.getMessage().contains("is not compatible")) {
+                else if (exception.getMessage() != null
+                        && exception.getMessage().contains("is not compatible")) {
                     failureIndex = VERSION_MISMATCH;
                     failureDetails = exception.getMessage();
                 }
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.env.data.login.LoginServiceImpl
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2013 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2018 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
@@ -96,6 +96,9 @@ public class LoginServiceImpl
 
     /** The index set if an error occurred while trying to connect. */
     private int failureIndex;
+    
+    /** The error message */
+    private String failureDetails;
 
     /** Allows to easily access the service's configuration. */
     protected LoginConfig config;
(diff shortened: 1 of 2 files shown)


pwalczysko:
Listed for tomorrow's standup

components/insight/SRC/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
```diff
@@ -183,7 +186,12 @@ private int attempt()
                     if (cause.getCause() instanceof SecurityViolation)
                         failureIndex = ACTIVE_INDEX;
                 }
-            } else failureIndex = SYSTEM_FAILURE_INDEX;
+                else if (exception.getMessage().contains("is not compatible")) {
```
mtbc: Maybe a `null` check first?

mtbc: May need a note in `Gateway` to update this if the message wording is changed.

dominikl: That's true, have to check for `null` first. It's a bit dodgy anyway parsing the message string to find out the specifics of the Exception, would maybe need another flag/field in the `DSOutOfServiceException` itself to account for that (will add a note to the gateway card).

mtbc: Could also subclass it.

pwalczysko:
Tested wrong server, port, username/password. The message was always as appropriate and leading me to the right conclusion.
The only remark here is 
screen shot 2018-02-15 at 18 05 21
why should I try again later when my credentials are wrong ? The trying later definitely makes sense when server is down, but when the sever is down, I have a message saying

screen shot 2018-02-15 at 18 07 24
which makes perfect sense.

Meaning: if the server is down, I would get the second dialog, and all is well. If the credentials are wrong, I will get the first dialog, and in that situation I think
    - the server is up
    - the credentials are wrong
    - why should I ``...or try again later`` in such situation ? This will do me no good....

pwalczysko:
Connecting the insight built with this PR (5.4.3 insight) to a 5.3.0 server gives me a message analogous to the one in the header of this PR -> as expected.

Edit: Also tested connecting of this 5.4.3 insight to 5.4.0 server (minor version difference). Connected as expected.

pwalczysko:
In summary:

   - the incorrect credentials dialog could be improved https://github.com/openmicroscopy/openmicroscopy/pull/5658#issuecomment-366012955, but not a blocker if too difficult to fix
   - the version mismatch dialog is as expected in the situation new instight -> old server
ToDo:
@dominikl helpfully agreed to build a 5.3.x insight with this PR included, which would help me to connect to eel, having the ``old insight -> new server`` situation, which is the main situation we are trying to fix the problem for.

pwalczysko:
Testing the test from ToDo https://github.com/openmicroscopy/openmicroscopy/pull/5658#issuecomment-366014203 - 5.3 insight with this PR connecting to a 5.4 server. @dominikl helpfully build such insight and importer, which I used for this test.

All works as expected even in this situation.

The only minor fix: See the dialog below: It is ``OMERO.Insight``, but should be ``OMERO.insight`` (do not capitalize the ``I``)
screen shot 2018-02-16 at 10 00 31

dominikl:
Thanks. I'll add one more commit to get rid off the 'try again later' when the username/password is wrong, that indeed doesn't make sense.

manics:
Isn't there a corner case where the server has just started and is listening but has not fully initialised, and will reject logins?

pwalczysko:
@manics @dominikl : 


>sn't there a corner case where the server has just started and is listening but has not fully initialised, and will reject logins?

Yes, there is such case, just tested it and in this situation the dialog you get from insight is the one which talks about credentials. Indeed, this improved by me re-entering the same credentials later.

pwalczysko:
This works as expected. Tested insight, importer and insight-ij plugin. All messages as expected, the capitalization is fixed too. Tested also on Win7. 

Ready to merge FMPOV.

PR #5659, opened 2018-02-15
Remove pyramid

jburel:
# What this PR does

Offer a CLI option to remove pyramid of a given endianness
This should allow to remove the corrupted pyramids using ``removepyramids``
This follow the approach taken by ``cleanse`` or ``fixpyramids`` (we could look at some unification at some stage/or move to a plugin)

# Testing this PR
Several options are possible (Indicated with ``--dry-run`` so that the pyramids are not deleted)
1.
``
bin/omero admin removepyramids --dry-run
``
will list the pyramids with little-endian=false to be deleted

2. 
``
bin/omero admin removepyramids  --dry-run --endian=little
``
will list the pyramids with little-endian=true to be deleted

3. 
``
bin/omero admin removepyramids --dry-run --endian=little --imported-after 2018/02/10
``
will list the pyramids with little-endian=true and imported after a given date to be deleted
Check is a valid date and a date in the future.
Nothing should be printed out for a date in the future

one image generating an OMERO pyramid have been imported on 15/02 for root:

*  in ``fake`` with little-endian=true

* Delete the pyramids with little-endian=true. The thumbnail on disk will also be deleted
* Note the values returned e.g. ``255``. ``255`` is the image ID
* Refresh the dataset
* The "clock" should be displayed instead of the thumbnail.
* When the pyramid creation is done, refresh again, a valid thumbnail should be created. 
The pyramid and thumbnails have been regenerated

# Related reading

https://trello.com/c/o3Ro0GXT/24-omero-import-large-tiffs-issue

Note:
This was initially added to ``admin`` since it required local access but the implementation has changed since.
It is kept in admin so it is in synch with the other methods e.g. ``fixpyramids``


joshmoore:
The API here seems generally reasonable, i.e. being able to choose by type and date-range. In fact, that may be the type of selector that we will need more widely for permitting operations on large repositories.

My primary concern here is that unlike `cleanse`, this doesn't need to be client-side(`**`). `cleanse` and friends have been a consistent cause of concern, but they *must* be client-side since the issue they are dealing with is that the server doesn't know about the files. 

Would using this code as a generator of candidates by an option and then delegate to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/cmd/fs/ManageImageBinariesI.java#L252 for the deletion?

`**`: by client-side I mean code that is implemented directly using OS primitives like `os.remove` or Java's `File.delete` rather than using existing server methods.

jburel:
It uses os.remove to remove the pyramid and thumbnail similar to the method ``fixpyramids``
are you suggesting to review this entire approach? 
I will avoid that in this stage if possible

jburel:
Discussed with @joshmoore, now that we have the logic in place, I will try to push the "deletion" to the server. So that admin does not need direct access to the file system
I will re-use ManageImage class

jburel:
Still one TODO in ``findPyramids`` but the structure is now in place

components/blitz/src/omero/cmd/fs/FindPyramidsI.java
```diff
@@ -0,0 +1,197 @@
...
+    @Override
+    public Response getResponse() {
+        return helper.getResponse();
+    }
+
+    /**
+     * Finds the pyramids and prepares the response
+     */
+    private void findPyramids() {
+        File root = new File("/OMERO"); //TODO
```
joshmoore: I think you want this bean:

```
  <bean name="/OMERO/Pixels" class="ome.io.nio.PixelsService"
```

jburel:
@joshmoore no more todos
It will be straightforward to adjust the ``fixpyramids`` method.
Due to the bug with permissions it has not been used in 5.4 so we could take the opportunity to change it.

jburel:
``@admin_only(AdminPrivilegeReadSession)`` will have to be replaced by ``@admin_only()``when
#5661 is in

components/tools/OmeroPy/src/omero/util/cleanse.py
```diff
@@ -329,6 +334,61 @@ def fixpyramids(data_dir, query_service,
...
+            req.deletePyramid = True
+            req.deleteThumbnails = True
+            to_delete.append(req)
+
+    if len(to_delete) > 0:
+        submit(client, to_delete)
+        print "%s Pyramids removed" % len(to_delete)
+
+
+def submit(client, request):
```
joshmoore: Could you reuse https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/clients.py#L892 ?

components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -365,9 +375,20 @@ def __init__(this, name, help, wait=False):
                 help="Print out which files would be deleted")
             x.add_argument(
                 "data_dir", type=DirectoryType(),
-                help="omero.data.dir directory value (e.g. /OMERO")
+                help="omero.data.dir directory value e.g. /OMERO")
             x.add_login_arguments()
 
+        removepyramids.add_argument(
```
joshmoore, line 387: perhaps add the `wait` arguments? https://github.com/openmicroscopy/openmicroscopy/pull/5608/files#diff-57f0d9cb121726c674f02fd0bc23bd1cR220

joshmoore (commented):
Few omero.cmd related questions

components/tools/OmeroPy/src/omero/plugins/admin.py
jburel: I do not have access to that commit in the workflows branch
I can open  a follow up PR when this is merged into develop

joshmoore:
A follow-up is of course fine, but having access to 5.4.2 on the `workflows` branch seems reasonable enough if you want.

components/tools/OmeroPy/src/omero/util/cleanse.py
jburel: done

jburel:
--exclude

jburel:
--exclude

jburel:
For a reason I have not yet found 2 tests do not pass when run on devspace (1 similar is green)
running the command from my local build pointing to devspace
```
bin/omero admin removepyramids --little-endian --dry-run
Previous session expired for root on localhost:4064
Server: [localhost:4064]10.0.51.143:32780
Username: [root]
Password:

```
returns the correct message ``Would remove pyramid for image 2``

joshmoore:
Is this a "tests are not run local to server" issue?

jburel:
it does not explain why one pass and the other one does not:
``test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_imported_after_future PASSED
06:01:42 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_imported_after_now FAILED
``
since both tests follow the same pattern and it fails at the request for pyramid
```
request = omero.cmd.FindPyramids()
    request.littleEndian = little_endian
    request.importeAfter = -1
    if imported_after is not None:
        date = time.strptime(imported_after, "%d/%m/%Y")
        request.importeAfter = time.mktime(date)
    rsp = submit(client, request, wait)
```

joshmoore:
@jburel : you'll need to unwrap the CmdError to get the `omero.cmd.ERR`. Looking at the server logs it's likely:

```
2018-02-21 06:01:41,742 INFO  [        ome.services.util.ServiceHandler] (2-thread-4)  Excp:	omero.cmd.HandleI$Cancel: bad-step
```

which follows:

```
2018-02-21 06:01:41,740 INFO  [                ome.io.nio.PixelsService] (2-thread-4) Using Pyramid BfPixelBuffer: /home/omero/workspace/OMERO-test-integration/data/Pixels/Dir-001/1203_pyramid
2018-02-21 06:01:41,741 ERROR [                ome.io.nio.PixelsService] (2-thread-4) Error instantiating pixel buffer: /home/omero/workspace/OMERO-test-integration/data/Pixels/Dir-001/1203_pyramid
ome.conditions.ResourceError: FormatException: /home/omero/workspace/OMERO-test-integration/data/Pixels/Dir-001/1203_pyramid
Invalid TIFF file: /home/omero/workspace/OMERO-test-integration/data/Pixels/Dir-001/1203_pyramid
```

components/tools/OmeroPy/src/omero/util/cleanse.py
```diff
@@ -329,6 +334,56 @@ def fixpyramids(data_dir, query_service,
...
+        submit(client, to_delete, wait)
+        print "%s Pyramids removed" % len(to_delete)
+
+
+def submit(client, request, wait):
+    if isinstance(request, list):
+        request = DoAll(request)
+    ms = 500
+    loops = ceil(wait * 1000.0 / ms)
+    cb = client.submit(request, loops=loops, ms=ms, failonerror=True,
```
joshmoore: Here you can either pass `failonerror=False` to get back the `ERR` response or you can catch `omero.CmdError` and unwrap yourself (`cmd_error.err`). Also, the callback will need closing.

jburel:
```
16:02:18 test/integration/clitest/test_pyramids.py::TestRemovePyramids::test_removepyramids_admin_only PASSED
16:02:27 test/integration/clitest/test_pyramids.py::TestRemovePyramidsRestrictedAdmin::test_removepyramids_restricted_admin PASSED
16:03:13 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_little_endian PASSED
16:03:48 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_imported_after_future PASSED
16:04:23 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_imported_after_now PASSED
```

jburel:
https://10.0.51.143:32777/job/OMERO-test-integration/7/ is green

pwalczysko:
Tested the following workflow


  - run bin/omero admin removepyramids --dry-run --little-endian
  - first to check what will be deleted then (edited)

  - bin/omero admin removepyramids --little-endian
   - for server enter
   - 10.0.51.143:32780
   - put in credentials
   - log in web http://10.0.51.143:32783/web
   - expand dataset "fake"
   - you should see a clock
   - then after a while the thumbnail is back (i.e. pyramid has been regenerated)

This workflow works as expected, but, seeing the output of the ``-h`` of the ``removepyramids`` command, I tred also playing with other subcommands.

There is a ``no pyramids to remove`` message when I leave the ``--little-endian`` flag out. This implies that the vanilla ``removepyramids`` command removes some other pyramids than ``little-endian`` ones (possibly a big-endian ? (a wild guess)). This is not intuitive I would say and is not explained or even hinted at in the help output.

jburel:
There is a ``no pyramids to remove`` message when I leave the --little-endian flag out
the ``no pyramids to remove`` is only displayed if no matching pyramids are found. So it really depends on what is on the server. I should have been clearer in the description
if ``--little-endian`` is not specified then yes, big-endian pyramid will be removed

pwalczysko:
>if --little-endian is not specified then yes, big-endian pyramid will be removed

Okay, thanks. I think a sentence like this should be inside a description of the command in the output of``omero removepyramids -h`` 

Edit: (the instructions you gave me for testing were clear enough, there is not a problem with that, I just think the command itself should be better explained for our users)

jburel:
Adjusted the help message after discussion with @pwalczysko 
```
Examples:
  bin/omero admin removepyramids --dry-run
  bin/omero admin removepyramids --dry-run --little-endian
  bin/omero admin removepyramids --dry-run --imported-after dd/mm/YYYY
            

Optional Arguments:
  In addition to any higher level options

  -h, --help                        show this help message and exit
  --dry-run                         Print out which files would be deleted
  --little-endian                   Delete pyramid with little-endian equals to true. If not specified, pyramid with big-endian equals to true will be deleted
  --imported-after DATE             Delete pyramid imported after a given date. Expected format dd/mm/YYYY
  --wait WAIT                       Number of seconds to wait for the processing to complete (Indefinite < 0; No wait=0).
```

pwalczysko:
Help message looks good now. 
Ready to merge FMPOV

components/blitz/resources/omero/cmd/FS.ice
```diff
@@ -26,6 +26,43 @@ module omero {
...
+
+            /**
+             * Retrieves the pyramids with little endian true or false
+             **/
+            bool littleEndian;
+            /**
+             * Retrieves the pyramids created after a specified time.
+             * Pass -1 not to check date
+             **/
+            long importeAfter;
```
joshmoore: Typo. Drop the `e` or perhaps `importedAfter`?

components/blitz/src/omero/cmd/fs/FindPyramidsI.java
```diff
@@ -0,0 +1,226 @@
...
+                                if (n.startsWith("." + name) &&
+                                (n.endsWith(".tmp") || n.endsWith(".pyr_lock"))) {
+                                    toDelete = false;
+                                    break;
+                                }
+                                
+                            }
+                        }
+                    }
+                    if (toDelete) {
```
joshmoore: `toDelete` name is perhaps inappropriate.

components/blitz/src/omero/cmd/fs/FindPyramidsI.java
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Returns whether or not the pixels set is little endian.
+     * @param pixels The pixels set to handle
+     * @return See above.
+     */
+    private boolean isLittleEndian(Pixels pixels) {
+        try {
+            PixelBuffer pf = pixelsService._getPixelBuffer(pixels, false);
+            if (pf instanceof BfPixelsWrapper) {
```
joshmoore: This block still concerns me. Perhaps with the work on the OmeroPyramidWriter code, we can simplify it.

joshmoore (commented):
Barring a few minor comments, probably my largest concern is the lack of paging. Are there any limits on systems with a very large number of pyramids?

components/blitz/src/omero/cmd/fs/FindPyramidsI.java
jburel: I don't like it either but unfortunately as mentioned before  I could not introduce a ``isLittleEndian`` method at the I/F level (b/c of the way RomioPixelBuffer works)
I will add a TODO so we can go back to it

jburel:
If the time limit is used, option that makes sense, I don't think the number of pyramids will be that large (otherwise we will have heard earlier about that issue)

jburel:
@joshmoore I have introduced a ``--limit`` option. The max value is currently set to an arbitrary value of 500
Test added.

jburel:
https://10.0.51.143:32777/job/OMERO-test-integration/8/ is green

joshmoore:
A comment, an RFE, and a bug:

Comment: the text for little-endian wraps around. I'm not sure if we have an easy way to handle this at the moment. You can put some details into the help description ([here](https://github.com/openmicroscopy/openmicroscopy/pull/5659/files#diff-57f0d9cb121726c674f02fd0bc23bd1cR175))

```
  -h, --help                        show this help message and exit
  --dry-run                         Print out which files would be deleted
  --little-endian                   Delete pyramid with little-endian equals to true. If not specified, pyramid with big-endian equals to true will be deleted
  --imported-after DATE             Delete pyramid imported after a given date. Expected format dd/mm/YYYY
  --limit MAX_NUMBER                Set the limit of pyramids to remove in one call. The default value is set to 500
  --wait WAIT                       Number of seconds to wait for the processing to complete (Indefinite < 0; No wait=0).

```

RFE: date formatting is always an issue. Probably ok to just go with this one for now since it's clear in the docs, but longer term, we should decide if it should default to trying to match locale, etc.

```
Examples:
  bin/omero admin removepyramids --dry-run
  bin/omero admin removepyramids --dry-run --little-endian
  bin/omero admin removepyramids --dry-run --imported-after dd/mm/YYYY
```

Minor bug:

```
Removing pyramid for image 4306
Removing pyramid for image 4899
Removing pyramid for image 4987
Removing pyramid for image 4991
    rv = omero.cli.argv()
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1095, in invoke
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1172, in onecmd
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 644, in _check_admin
    return func(*args, **kwargs)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/plugins/admin.py", line 1003, in removepyramids
    limit=limit, wait=wait)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/util/cleanse.py", line 378, in removepyramids
    submit(client, to_delete, wait, ctx)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/util/cleanse.py", line 401, in submit
    cb.close(True)
UnboundLocalError: local variable 'cb' referenced before assignment
```

mtbc:
Perhaps YYYY-MM-DD would be safest if we want to avoid potential for confusion.

jburel:
I have reorganized the help and changed the date format to YYYY-mm-dd

jburel:
tests are green after the changes.

joshmoore:
Would `testintegration` have been updated? If so (from just now): 

```
No more than 500 pyramids will be removed
Removing pyramid for image 4306
Removing pyramid for image 4899
Removing pyramid for image 4987
Removing pyramid for image 4991
    rv = omero.cli.argv()
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1095, in invoke
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1172, in onecmd
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 644, in _check_admin
    return func(*args, **kwargs)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/plugins/admin.py", line 1004, in removepyramids
    limit=limit, wait=wait)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/util/cleanse.py", line 378, in removepyramids
    submit(client, to_delete, wait, ctx)
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/util/cleanse.py", line 400, in submit
    ctx.die(13, "Failed to remove pyramid:\n%s" % err)
AttributeError: 'EventContext' object has no attribute 'die'
```

jburel:
We should now have 
```
Removing pyramid for image 4306
Removing pyramid for image 4899
Removing pyramid for image 4987
Removing pyramid for image 4991
Failed to remove pyramid:
object #0 (::omero::cmd::ERR)
{
    category = ::omero::cmd::ManageImageBinaries
    name = pyramid-requires-fileset
    parameters = 
    {
    }
}
```
Another option is not to ``die``, and just print ``cannot remove pyramid``.
The ``die`` option is in the spirit of the ``email`` method
Maybe review that later

joshmoore:
* `    name = pyramid-requires-fileset` - this means that these files aren't being cleaned up. A run of `cleanse` should clean them.
 * `By default pyramids with big-endian equals to true will be deleted.` - I think I would have expected all pyramids to by deleted if this flag wasn't set.
 * `def submit(` - this might should be slated for addition to `omero/cli.py`

Let me prepare a commit.

jburel:
I opted for little-endian or big-endian but not all
I can see that being an option

for ``submit`` I agree. I am also planning to deprecate the fixpyramids and use only removepyramids
I have prep the new class for that. That will be done later

joshmoore:
Tested the above with:

 * `big&sizeX=3500&sizeY=3500&little=false.fake`
 * `lil&sizeX=3500&sizeY=3500&little=true.fake`
 * `touch /OMERO/Pixels/123_pyramids`

Was looking for a easy way to create a pre-FS pyramid. See https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/test_reimport.py#L117

jburel:
https://10.0.51.143:32777/job/OMERO-test-integration/16/ is green
tested added for:

 * big/little/both
 * manual addition i.e. ``touch /OMERO/Pixels/123_pyramids``
 * pre-fs data
 * handle  ``limit<=0``

jburel:
f2bcbef means that the callback is probably closed before and an exception is not thrown.
I proposed to investigate that in a follow-up PR (post release)

Edit: card https://trello.com/c/KBFsy1gR/353-cleanse-review

components/blitz/resources/omero/cmd/FS.ice
```diff
@@ -26,6 +26,49 @@ module omero {
...
+            /**
+             * Retrieves the pyramids with little endian true or false.
+             * If unset, both will be retrieved.
+             **/
+            omero::RBool littleEndian;
+            /**
+             * Retrieves the pyramids created after a specified time.
+             * Pass -1 not to check date
+             **/
+            long importedAfter;
```
joshmoore: Units need specifying.

components/blitz/src/omero/cmd/fs/FindPyramidsI.java
```diff
@@ -0,0 +1,232 @@
...
+        Pixels pixels = service.findByQuery(query, new Parameters().addId(pixelsId));
+        if (pixels == null) {
+            return -1;
+        }
+        Image image = pixels.getImage();
+        if (checkEmptyFile) {
+            return image.getId().longValue();
+        }
+        long creation = image.getDetails().getCreationEvent().getTime().getTime();
+        if (importedAfter < creation/1000) {
```
joshmoore: Though python uses seconds, `long time` in the API is often in milliseconds.

components/tools/OmeroPy/src/omero/util/cleanse.py
```diff
@@ -329,6 +334,72 @@ def fixpyramids(data_dir, query_service,
...
+                                   failontimeout=True)
+                print "Pyramid removed for image %s" % image_id
+            except omero.CmdError, ce:
+                print "Failed to remove for image %s: %s" % (
+                    image_id, ce.err.name)
+            finally:
+                if cb:
+                    try:
+                        cb.close(True)
+                    except Ice.NotRegisteredException:
```
joshmoore, line 394: Would be good to know why these are being caught. In general, catching all exceptions here is likely fine, but we will need to debug after release.

components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
```diff
@@ -0,0 +1,235 @@
...
+    def import_pyramid(self, tmpdir, name=None):
+        if name is None:
+            name = "test&sizeX=4000&sizeY=4000.fake"
+        fakefile = tmpdir.join(name)
+        fakefile.write('')
+        pixels = self.import_image(filename=str(fakefile), skip="checksum")[0]
+        # wait for the pyramid to be generated
+        id = long(float(pixels))
+        assert id >= 0
+        time.sleep(30)
```
joshmoore, line 99: `time.sleep` should be used only as a last resort. Would be good to replace these with a `for`-loop which checks to see if the pyramids have been generated yet. (A `omero.cmd` callback would be even better)

components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
```diff
@@ -0,0 +1,235 @@
...
+        self.args += ["--endian=little"]
+        self.args += ["--limit", "0"]
+        self.cli.invoke(self.args, strict=True)
+        out, err = capsys.readouterr()
+        output_start = "No more than 500 pyramids will be removed"
+        assert output_start in out
+
+    def test_remove_pyramids_manual(self, tmpdir, capsys):
+        """Test removepyramids for a file manually added under /Pixels"""
+        proxy, description = self.client.getManagedRepository(description=True)
```
joshmoore: These will need the `@pytest.mark.local_only` (or similar) marker.

joshmoore (commented):
I've noted a couple of things that should be handled in a follow up. To not block 5.4.4, the only issue that really needs a decision is the type of `importedAfter`. If we're ok with this being the result of `mktime` which we haven't used elsewhere, then we can just add the required documentation later. (At the moment, it's only used internally by the Python code)

components/tools/OmeroPy/src/omero/util/cleanse.py
jburel: card created

components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
jburel: Agree. We can improve that in follow-up PR

components/blitz/resources/omero/cmd/FS.ice
jburel: going for RTime

joshmoore:
Local testing (on devspace/testintegration) looks good.

jburel:
```
14:36:04 test/integration/clitest/test_pyramids.py::TestRemovePyramids::test_removepyramids_admin_only PASSED
14:36:14 test/integration/clitest/test_pyramids.py::TestRemovePyramidsRestrictedAdmin::test_removepyramids_restricted_admin PASSED
14:36:59 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_little_endian PASSED
14:37:34 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_imported_after_future PASSED
14:38:10 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_limit PASSED
14:38:45 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_not_valid_limit PASSED
14:38:45 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_manual SKIPPED
14:39:21 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove__pre_fs_pyramids PASSED
14:39:57 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_big_endian PASSED
14:41:07 test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids PASSED
```

jburel:
All tests introduced in this PR as green.
Merging

PR #5661, opened 2018-02-16
full_admin: fix exception on no arguments

joshmoore:
When no arguments were being passed, an exception was being throw
which trigger an early exit so that even root could not run the
method.

# Testing this PR

1. Without this PR, try `bin/omero admin fixpyramids /OMERO` as root -- **fail**
2. Try `bin/omero admin fixpyramids /OMERO` as root with this PR -- **ok**
3. Try add an admin with all privileges with this PR -- **ok**
4. Try add an admin with even one privilege removed -- **fail**

# Seealso

 * https://github.com/openmicroscopy/openmicroscopy/pull/5657
 * Note: in trying to write tests for this, the fact that this is a local-only plugin reared it's head again. I assume that's why this was not found sooner.

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -605,7 +605,10 @@ class Context:
 def admin_only(*fargs, **fkwargs):
     """
     Checks that the current user is an admin and has sufficient privileges,
-    or throws an exception.
+    or throws an exception. If no arguments are present or if full_admin is
+    passed and is True, then this method assumes that the user must be a
+    full admin and have all privileges. To disable this behavior, set
+    `full_admin` to False.
     """
     def _admin_only(func):
         @wraps(func)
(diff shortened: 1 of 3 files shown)


jburel:
The same restriction should apply to ``cleanse``

jburel:
tested the commands manually it works as indicated in the PR.
I have also opened a PR against this branch adding tests

Comment above, ``cleanse`` should follow the same approach

joshmoore:
@jburel : pushed

jburel:
```
./components/tools/OmeroPy/src/omero/plugins/admin.py:37:1: F401 'AdminPrivilegeReadSession' imported but unused
```

joshmoore:
Doh. That's what I get for editing with GitHub.

joshmoore:
Green, @jburel

jburel:
Travis is green

jburel:
Manual tests are passing. Merging so changes can be made in #5659
If required I will make adjustments in #5659

PR #5669, opened 2018-02-21
layout button css fix

will-moore:
# What this PR does

With the filter search div previously having ```width:90%```, we found that a small browser window would cause it to overlap with the layout buttons, meaning that clicks were ignored.

To test:
 - make your browser window quite small
 - Click layout buttons to switch between list and grid layout.

cc @jburel


jburel:
tested in devspace
it works as expected now
thanks @will-moore

PR #5676, opened 2018-03-07
Script param list default

will-moore:
# What this PR does

See https://github.com/ome/scripts/pull/139
If a script ```List``` parameter has a default set, we now populate this correctly in the script dialog as a comma-separated list.

# Testing this PR

1. Test the script from https://github.com/ome/scripts/pull/139 and check that the default value for Channels is ```1,2,3,4``` and that this is handled correctly without editing the field (up to first 4 channels of an image are used for ROI intensitities).

screen shot 2018-03-07 at 14 29 28

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -3592,7 +3592,7 @@ def script_ui(request, scriptId, conn=None, **kwargs):
         elif pt.__class__.__name__ == 'list':
             i["list"] = True
             if "default" in i:
-                i["default"] = i["default"][0]
+                i["default"] = ",".join([str(d) for d in i["default"]])
         elif isinstance(pt, bool):
             i["boolean"] = True
         elif isinstance(pt, int) or isinstance(pt, long):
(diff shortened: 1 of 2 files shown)


joshmoore:
* :+1: for showing more than the first value
 * Bit of a surprise that the list is longer than the actual channels (e.g. when testing a single channel fake)
    - This is ultimately an architectural problem with the server: the parameter could be "dynamic" but that would slow down the user interaction.
    - Are we already detecting "channel" parameters anywhere else? If we have a standard name or type (or annotation) then we could do the validation client-side. If not, it's probably not worth adding it now.
 * Are there any tests for this code converting parameters to user inputs and then converting them back to script inputs? If so, we could probably use some additions here. For example, if I change the channel to "x" the script passes and shows:

```
script_params:
{'Channels': [], 'Export_All_Planes': False, 'IDs': [95103L], 'Data_Type': 'Image', 'File_Name': 'roi_intensities.csv'}
Processing 1 images...
Image ID 95103...

will-moore:
Tests passed at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/813/testReport/OmeroWeb.test.integration.test_scripts/

will-moore:
@joshmoore We don't have anything for Channel validation client-side, but this wouldn't give much to the user after they hit Run, since they wouldn't know whether the client was checking Channel counts and passing 'valid' indexes to the script or if the validation was happening in the script itself (which is actually much easier to do since we don't want the client to have to load all the data).

manics (approved):
UI is as expected, CI tests pass.

PR #5677, opened 2018-03-09
Fix CLI list subcommand

sbesson:
## Description

This PR addresses a  bug in the CLI config command reported in http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2018-March/004159.html

## Testing

Without this PR e.g. using OMERO.py 5.4.4, `bin/omero config list` should fail with the exception reported in the email thread.

With the PR included, the command should return the same output as `bin/omero config get`. 

## Changes summary

cf9a1e6 extends the unit test to cover this subcommand (fails without the fix and passes with it).
7ab9620 allows any code to call `PrefsControl.get(args, control)` without requiring the password arguments (hiding password as the default)
233a2db implements the same password options in `bin/omero config list` as for `bin/omero config get`

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -297,7 +297,8 @@ class PrefsControl(WriteableConfigControl):
             if args.KEY and len(args.KEY) == 1:
                 self.ctx.out(config[k])
             else:
-                if is_password(k) and not args.show_password:
+                if is_password(k) and not getattr(
+                        args, "show_password", False):
                     self.ctx.out("%s=%s" % (k, '*' * 8 if config[k] else ''))
                 else:
                     self.ctx.out("%s=%s" % (k, config[k]))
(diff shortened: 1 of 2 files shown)


joshmoore:
No objections to this, but might make sense to also add the argument itself (https://github.com/sbesson/openmicroscopy/blob/7ab962086c4cd36775f932592035f61630d6841b/components/tools/OmeroPy/src/omero/plugins/prefs.py#L153 in a `for`-loop) and/or deprecate the method.

sbesson:
Deprecating a subcommand in a patch release without discussion seemed premature. Pushed 233a2db which turns `config list` an alias of `config get` without positional `KEY` arguments.

jburel:
```
bin/omero config get
omero.web.apps=["omero_parade"]
omero.web.ui.center_plugins=["Parade", "omero_parade/init.js.html", "omero_parade"]
omero.web.ui.top_links=["Parade", "parade_index", {"title": "Search and filter in OMERO.parade"}]
```
```
bin/omero config list
omero.web.apps=["omero_parade"]
omero.web.ui.center_plugins=["Parade", "omero_parade/init.js.html", "omero_parade"]
omero.web.ui.top_links=["Parade", "parade_index", {"title": "Search and filter in OMERO.parade"}]
```

jburel:
Everything works as expected.
@sbesson do you already have a card to discuss deprecation so it does not get forgotten?

sbesson:
Captured as https://trello.com/c/7KNfAJAD/373-cli-config-list-deprecation

jburel:
Thanks

PR #5678, opened 2018-03-09
fix Trac URLs in comments

mtbc:
trac.openmicroscopy.org.uk's SSL certificate is long-expired. This PR does *not* fix all the URLs but it should at least be an improvement.


joshmoore:
Code changes look trivially fine. The only consideration would be potential issues around backporting. None of the files look generally contested at the moment, though likely this is more centrally an issue for your own read-only work.

mtbc:
If we can get this in for 5.4.5 then I'll be happy to merge that into the read-only branch or rebase against it.

joshmoore:
Merging to provide a simpler integration point for the read-only branch.

PR #5679, opened 2018-03-09
Pyramid thumbnail

jburel:
# What this PR does

* Roll back a change in the ThumbnailBean making the caching void
* delete the rendering settings associated to a pyramid. This will allow the thumbnails to be regenerated correctly

# Testing this PR

* Browse a large dataset or a large plate and check the performance e.g. ``InCell``
* Import a fake pyramid ``big&sizeX=4000&sizeY=4000&little=false.fake``
* Browse the dataset containing the imported pyramid. Check that the thumbnail is correctly generated when the pyramid is ready.
* Delete the pyramid: ``omero admin removepyramids --endian=big --imported-after DATE-1day`` where ``DATE=YYYY-MM-DD``
* Browse the dataset containing the imported. Check that the "clock" is displayed. Wait then refresh. The thumbnail should be displayed.
 
# Related reading

https://trello.com/c/7FUOn3U7/13-thumbnails-loading-performance

cc @pwalczysko @chris-allan

### DIFF
--- a/components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
@@ -87,7 +87,7 @@ class TestRemovePyramidsFullAdmin(CLITest):
         self.args += ["admin", "removepyramids"]
         self.group_ctx = {'omero.group': str(self.group.id.val)}
 
-    def import_pyramid(self, tmpdir, name=None):
+    def import_pyramid(self, tmpdir, name=None, settings=False):
         if name is None:
             name = "test&sizeX=4000&sizeY=4000.fake"
         fakefile = tmpdir.join(name)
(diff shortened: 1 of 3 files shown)


chris-allan:
Can we have this rebased and force pushed so that the first commit is a `git revert` of 5dd3cad7031437ad2658f7f77843af3ea7aa7fea rather than a separate commit that just deletes the line?

Just for my own understanding, I don't think this is explained on #5659, under what conditions are clocks getting cached?

The reason I ask is that there are plenty of conditions where a pyramid might be missing, or corrupt and needs to be deleted. Am I correct in saying that if a valid `RenderingDef` is then available for the Pixels that a thumbnail is being requested for that a clock will be cached?

jburel:
The ``removepyramids`` method removes the pyramid and the associated thumbnails on disk.
Requesting a thumbnail, trigger the generation of the pyramid. 
You are correct: if a valid renderingDef is available, the clock is cached.
Another option (not tested) could be to delete the thumbnail in DB, I think that should work.
I have opted to delete the renderingDef since they might be invalid.

components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
```diff
@@ -98,9 +98,24 @@ def import_pyramid(self, tmpdir, name=None):
...
+                tb.getThumbnailByLongestSideSet(rint(16), [id])
+            finally:
+                tb.close()
+            # check the rendering settings
+            query = """
+            select rdef from RenderingDef as rdef
+            left outer join fetch rdef.pixels as p
+            where p.id =:id
+            """
+            rnd_defs = query_service.findAllByQuery(query, param)
```
mtbc, line 114: Is this happening in all-groups context or is that not an issue in this case?

jburel: method is admin only

mtbc: I may be misunderstanding. I need the `context` variable to get a result from this script:

```python
#!/usr/bin/env python

import omero

conn = BlitzGateway('root', 'omero', host='eel.openmicroscopy.org')
conn.connect()
query_service = conn.getQueryService()

image_id = 25956

query = 'SELECT name FROM Image WHERE id = :id'
params = omero.sys.ParametersI()
params.addId(omero.rtypes.rlong(image_id))
context = {'omero.group': '-1'}

image_names = query_service.projection(query, params, context)

for image_name in image_names:
    print(image_name[0].val)

conn._closeSession()
```

chris-allan:
While I'm sure that deleting associated `RenderingDef` instances will work can we not find the code path causes the problem in the first place? Handling the exact situation you're outlining is as I understand it the fundamental purpose of the `inProgress` instance variable.

This seems like something that is fundamental and papering over it is going to come back and bite us in the future.

jburel:
replaced by https://github.com/openmicroscopy/openmicroscopy/pull/5682

PR #5680, opened 2018-03-10
pyramid help

jburel:
# What this PR does

Clarify help for limit parameter

# Testing this PR

Check that ``omero admin removepyramids --help`` display the message correctly

cc @mtbc


mtbc:
Could you remind me which bit of code implements the refusal to support limits over 500? I seem to have lost the previous conversation.

jburel:
```
 value = long(limit)
 if value > 500 or value <= 0:
        value = long(500)
```

mtbc:
Ahh, it's off in `cleanse`! I was looking in the `admin` plugin and in `QueryImpl` and whatnot, thank you. It might thus also be worth a commit to add a comment with that code block to warn that if someone changes the limit they know to change it in this help text too so they don't fall out of sync. Needn't block this PR though.

mtbc (approved):

PR #5681, opened 2018-03-12
Increase OMEROweb log-size to 500MB (rebased onto develop)

kennethgillen:
This is the same as gh-5386 but rebased onto develop.

----

# What this PR does

OMERO.web logs are currently rotated at 5MB, in contrast to the Java servers such as Blitz which are rotated at 500MB. This PR sets it to 500MB, matching the `LOGSIZE` set for the Python servers https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.0-m2/components/tools/OmeroPy/src/omero/util/__init__.py#L33

# Testing this PR

Install OMERO.web. Interact with the webclient until `OMEROweb.log` is larger than 5MB. Setting `omero.web.debug=true` will help fill the logs up.

This PR is required as OMERO.web generates enough log entries that single 5MB file is rotated away easily in a day on e.g. Dundee's production OMERO.web server.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -88,6 +88,8 @@ FULL_REQUEST_LOGFORMAT = (
     ' HTTP %(status_code)d %(request)s')
 
 LOGGING_CLASS = 'omero_ext.cloghandler.ConcurrentRotatingFileHandler'
+LOGSIZE = 500000000
+
 
 LOGGING = {
     'version': 1,
@@ -114,7 +116,7 @@ LOGGING = {
             'class': LOGGING_CLASS,
             'filename': os.path.join(
                 LOGDIR, 'OMEROweb.log').replace('\\', '/'),
-            'maxBytes': 1024*1024*5,  # 5 MB
+            'maxBytes': LOGSIZE,
             'backupCount': 10,
             'formatter': 'standard',
         },
@@ -123,7 +125,7 @@ LOGGING = {
             'class': LOGGING_CLASS,
             'filename': os.path.join(
                 LOGDIR, 'OMEROweb.log').replace('\\', '/'),
-            'maxBytes': 1024*1024*5,  # 5 MB
+            'maxBytes': LOGSIZE,
             'backupCount': 10,
             'filters': ['require_debug_false'],
             'formatter': 'full_request',


kennethgillen:
--rebased-from #5386

chris-allan (approved):
My immediate reaction would be that I would prefer for this to be configurable rather than a global that may have to be edited in place and potentially lost on upgrades. However, that would mean changing the order in which configuration properties are read and may lead to a other unexpected consequences. This is a fine incremental improvement as is.

kennethgillen:
Thanks @chris-allan - in total agreement. To have a default which can be overridden via config for both web and server would be great.

joshmoore:
Looking at:

 * https://logback.qos.ch/manual/configuration.html#variableSubstitution and
 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/etc/logback.xml#L28

I could see changing `500MB` to `500000000` in logback.xml and allow a single environment variable. Something like (untested):


```
diff --git a/etc/logback.xml b/etc/logback.xml
index 774e1e1a60..10de2c35da 100644
--- a/etc/logback.xml
+++ b/etc/logback.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration scan="true" debug="false">

+  <property name="LOG_SIZE" value="${OMERO_LOG_SIZE:-5000000000" />
+
   <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
     <file>${omero.logfile}</file>
       <!-- Exclude debug level logging from ome.services.blitz.repo.ManagedImportRequestI -->
@@ -25,7 +27,7 @@
       <maxIndex>9</maxIndex>
     </rollingPolicy>
     <triggeringPolicy class="ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy">
-      <maxFileSize>500MB</maxFileSize>
+        <maxFileSize>${LOG_SIZE}</maxFileSize>
     </triggeringPolicy>
     <encoder>
       <pattern>%date %-5level [%40.40logger{40}] \(%10.10thread\) %msg%n</pattern>
```

then:

```
export OMERO_LOG_SIZE=100000
bin/omero admin restart
```

jburel:
@kennethgillen could you capture the configuration suggestion so it can be improved in a follow up PR?

kennethgillen:
Linked to @joshmoore's comment on https://trello.com/c/WTatDRHI/374-configurable-log-file-sizes

jburel:
Thanks @kennethgillen

PR #5682, opened 2018-03-12
Pyramid thumbnail

jburel:
# What this PR does

* Roll back a change in the ThumbnailBean making the caching void
* Delete in the thumbnail bean the "clock" thumbnail so that when the pyramid is ready.
The correct thumbnail is displayed. 
* Add integration test checking the pyramid/thumbnail regeneration is correct.

# Testing this PR

* Browse a large dataset or a large plate and check the performance e.g. ``InCell`` or compare browsing of data between merge and latest server
* Import a fake pyramid ``big&sizeX=4000&sizeY=4000&little=false.fake``
* Browse the dataset containing the imported pyramid. Check that the thumbnail is correctly generated when the pyramid is ready.
* Delete the pyramid: ``omero admin removepyramids --endian=big --imported-after DATE-1day`` where ``DATE=YYYY-MM-DD``
* Browse the dataset containing the imported. Check that the "clock" is displayed. Wait then refresh. The thumbnail should be displayed.
* Check that the tests are green
 
# Related reading

https://trello.com/c/7FUOn3U7/13-thumbnails-loading-performance

cc @pwalczysko @chris-allan

### DIFF
--- a/components/server/src/ome/services/ThumbnailBean.java
+++ b/components/server/src/ome/services/ThumbnailBean.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -1128,6 +1129,20 @@ public class ThumbnailBean extends AbstractLevel2Service
                 _createThumbnail();
             }
             byte[] thumbnail = ioService.getThumbnail(thumbnailMetadata);
+            //Thumbnails are copied to disk retrieved from disk including
+            //the "clock"
+            //When existing pyramids and thumbnails are deleted from disk
+            //using method like removepyramids, the "clock" will be, the first
+            //time saved to disk, then always as used as the thumbnail since the
+            //current implementation assumes that the image does not already
+            //have some rendering settings.
+            //In the check, the "clock" is deleted from disk after creation.
+            //This means that the clock will have to be created every time until
+            //the correct thumbnail is ready.
+            if (inProgress) {
+                ioService.removeThumbnails(
+                        Arrays.asList(new Long[] { thumbnailMetadata.getId() }));
+            }
             return thumbnail;
         }
         catch (IOException e)
--- a/components/server/src/ome/services/ThumbnailBean.java
+++ b/components/server/src/ome/services/ThumbnailBean.java
@@ -1129,16 +1129,23 @@ public class ThumbnailBean extends AbstractLevel2Service
                 _createThumbnail();
             }
             byte[] thumbnail = ioService.getThumbnail(thumbnailMetadata);
-            //Thumbnails are copied to disk retrieved from disk including
-            //the "clock"
-            //When existing pyramids and thumbnails are deleted from disk
-            //using method like removepyramids, the "clock" will be, the first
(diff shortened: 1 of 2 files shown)


mtbc:
Created [[trello] Make clock thumbnails detectable](https://trello.com/c/Og41Gz5R/782-make-clock-thumbnails-detectable) to note a larger issue that this PR works around.

components/server/src/ome/services/ThumbnailBean.java
```diff
@@ -1129,6 +1129,20 @@ protected void actOnOneGroup(Set<Long> pixelsIds) {
                 _createThumbnail();
             }
             byte[] thumbnail = ioService.getThumbnail(thumbnailMetadata);
+            //Thumbnails are copied to disk retrieved from disk including
```
chris-allan: Rewording below:

Thumbnails are always saved to disk and then retrieved when the call stack includes `retrieveThumbnail().`  This includes the "clock".  `inProgress` is not set early enough for `retrieveThumbnailDirect()` to be called.

The current implementation assumes that if an image has valid rendering settings for the current user then it will have a valid pyramid.  When existing pyramids, thumbnail metadata, and cached thumbnails are deleted from disk using a method like removepyramids it invalidates this assumption.

The "clock" will first be saved to disk.  `inProgress` will then be checked and the "clock" deleted from disk after creation.  The "clock" will have to be created, saved to disk, and then retrieved every time the thumbnail is requested until pyramid creation completes and the correct thumbnail can be created.  Otherwise the "clock" would stay as the cached thumbnail until the valid rendering settings for the current user are modified.

components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
```diff
@@ -237,3 +243,46 @@ def test_remove_pyramids(self, tmpdir, capsys):
...
+            assert len(thumbs) == 1
+            # The clock should be returned during the pyramid generation
+            digest = self.calculate_sha1(thumbs[id])
+            assert digest != thumb_hash
+            while digest != thumb_hash:
+                thumbs = tb.getThumbnailByLongestSideSet(rint(64), [id],
+                                                         {'omero.group': '-1'})
+                digest = self.calculate_sha1(thumbs[id])
+                # wait few seconds before the next check
+                time.sleep(5)
```
chris-allan: Rather than sleeping for an indeterminate amount of time what you could do is initialize a RawPixelsStore and check for exceptions on `setPixelsId()` at an interval of once a second with a maximum duration of 30-60 seonds. This will hopefully prevent test brittleness when pyramid generation takes a while.

chris-allan (changes requested):
Overall implementation looks good. We have tested it on our own systems with good success.

components/tools/OmeroPy/test/integration/clitest/test_pyramids.py
jburel: I was planning to review the various time.sleep used in this class. Just wanted to get something out asap for review

chris-allan (approved):
I'm happy with the implementation if you're happy with the status of the integration tests. Good to merge as far as I'm concerned.

rgozim:
Tested on web-dev-merge and everything looks to be working great. [Integration tests](https://ci.openmicroscopy.org/view/OMERO/job/OMERO-DEV-merge-integration-python/831/testReport/OmeroPy.test.integration.clitest.test_pyramids/) also passing.

PR #5683, opened 2018-03-13
omeroweb service close

jburel:
# What this PR does

* Close services
* Add tests and check that the services are closed

# Testing this PR

# Related reading

https://github.com/openmicroscopy/management_tools/issues/703

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -2760,7 +2760,7 @@ def _table_query(request, fileid, conn=None, **kwargs):
             )
         )
     finally:
-        if t is None:
+        if t is not None:
             t.close()
 
 table_query = login_required()(jsonp(_table_query))
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -2732,31 +2732,36 @@ def _table_query(request, fileid, conn=None, **kwargs):
     ctx.setOmeroGroup("-1")
 
     r = conn.getSharedResources()
-    t = r.openTable(omero.model.OriginalFileI(fileid), ctx)
-    if not t:
-        return dict(error="Table %s not found" % fileid)
+    t = None
+    try:
+        t = r.openTable(omero.model.OriginalFileI(fileid), ctx)
+        if not t:
+            return dict(error="Table %s not found" % fileid)
 
-    cols = t.getHeaders()
-    rows = t.getNumberOfRows()
+        cols = t.getHeaders()
+        rows = t.getNumberOfRows()
 
-    if query == '*':
-        hits = range(rows)
-    else:
-        match = re.match(r'^(\w+)-(\d+)', query)
-        if match:
-            query = '(%s==%s)' % (match.group(1), match.group(2))
-        try:
-            hits = t.getWhereList(query, None, 0, rows, 1)
-        except Exception:
-            return dict(error='Error executing query: %s' % query)
+        if query == '*':
+            hits = range(rows)
+        else:
+            match = re.match(r'^(\w+)-(\d+)', query)
+            if match:
+                query = '(%s==%s)' % (match.group(1), match.group(2))
+            try:
+                hits = t.getWhereList(query, None, 0, rows, 1)
+            except Exception:
+                return dict(error='Error executing query: %s' % query)
 
-    return dict(data=dict(
-        columns=[col.name for col in cols],
-        rows=[[col.values[0] for col in t.read(range(len(cols)), hit,
-                                               hit+1).columns]
-              for hit in hits],
+        return dict(data=dict(
+            columns=[col.name for col in cols],
+            rows=[[col.values[0] for col in t.read(range(len(cols)), hit,
+                                                   hit+1).columns]
+                  for hit in hits],
+            )
         )
-    )
+    finally:
+        if t is None:
+            t.close()
 
 table_query = login_required()(jsonp(_table_query))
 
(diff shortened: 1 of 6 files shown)


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8283,7 +8283,10 @@ def getThumbnail(self, size=(64, 64), z=None, t=None, direct=True,
                 # else:
                 #     pos = None
             if self.getProjection() != 'normal':
-                return self._getProjectedThumbnail(size, pos)
+                try:
```
chris-allan: I certainly appreciate the goal here. However, doing this has lots of potential implications on other code paths that may rely on the `RenderingEngine` being initialized.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8308,13 +8311,13 @@ def getThumbnail(self, size=(64, 64), z=None, t=None, direct=True,
             args += [ctx]
             rv = thumb(*args)
             self._thumbInProgress = tb.isInProgress()
-            tb.close()      # close every time to prevent stale state
             return rv
         except Exception:  # pragma: no cover
             logger.error(traceback.format_exc())
+            return None
+        finally:
             if tb is not None:
```
chris-allan: This one definitely makes sense.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -9131,14 +9134,17 @@ def exportOmeTiff(self, bufsize=0):
...
+        try:
+            e.addImage(self.getId())
+            size = e.generateTiff(self._conn.SERVICE_OPTS)
+            if bufsize == 0:
+                # Read it all in one go
+                return fileread(e, size, 65536)
+            else:
+                # generator using bufsize
+                return (size, fileread_gen(e, size, bufsize))
+        finally:
```
chris-allan: The idea here is certainly good but this has no effect on open number of filehandles or perceived "leakage". The exporter data retrieval methods are all atomic.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -2732,31 +2732,36 @@ def _table_query(request, fileid, conn=None, **kwargs):
...
+
+        return dict(data=dict(
+            columns=[col.name for col in cols],
+            rows=[[col.values[0] for col in t.read(range(len(cols)), hit,
+                                                   hit+1).columns]
+                  for hit in hits],
+            )
         )
-    )
+    finally:
```
chris-allan, line 2762: This one definitely makes sense.

chris-allan (commented):
In general this is a *very* tough thing to review. Especially in the context of OMERO.web as the `login_required()` decorator, which is used by basically every remote call, already enforces all of this via its connection cleanup logic.

I would suggest that this PR be renamed `BlitzGateway` service close.

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel: I don't think it is related to the "leakage" but I think it is better to close it as a general rule

chris-allan:
In addition to the above review I would just like to remind us all that we have already been down this road in #5545 and then had to roll back stuff in #5637 due to unintended consequences. The details of the appropriate review are outlined in the latter PR.

As @joshmoore already mentioned on #5545, these sorts of changes are invasive. I would extend the significance of that statement to say that they are **hugely** invasive. Integration tests that look at a very isolated vertical set of code are not going to be sufficient to judge the impact of these changes.

mtbc:
I suppose that the prospect of unintended consequences is especially exciting if the changes are made shortly before planned release so we get little chance to even notice the issue before others do also.

jburel:
Discussed with @chris-allan: for clarity I will split the PR into 2: an omeroweb one and omeropy one
and partially roll back e552efde9dfccda16e10cd39cd8f5c5f15aeb1b4

jburel:
omeropy commits moved to gh-5685

will-moore:
Testing on web-dev-merge, repeatedly refreshing an OMERO.figure that had 80 images in it, which was sufficient to show the issues in #5545 is working fine currently.

components/tools/OmeroWeb/test/integration/test_plategrid.py
```diff
@@ -419,10 +419,14 @@ def test_get_plate_table(self, django_client, plate_well_table, conn):
...
-                                     data={'query': 'Well-%s' % wellId})
-        rspJson = json.loads(response.content)
-        print rspJson
+        try:
+            response = django_client.get(request_url,
+                                         data={'query': 'Well-%s' % wellId})
+            rspJson = json.loads(response.content)
+        finally:
+            for v in self.client.getSession().activeServices():
+                assert 'TableProxy' not in v, 'Leaked table!'
```
will-moore: Can't test locally, but are you sure this is ```TableProxy```, not ```TablePrx``` or ```Table```?

jburel: typo you are correct

will-moore:
It looks like most of the changes here are additional tests and the code changes are limited to the ```try/finally close()``` in the tables service.
So this would be expected to have much less impact than lower level BlitzGateway changes discussed above.
Unfortunately can't test locally as tables haven't been working on my machine for a while
```
         table = conn.c.sf.sharedResources().newTable(1, tablename)
>       table.initialize(columns)
E       AttributeError: 'NoneType' object has no attribute 'initialize'
```
but changes look good and tests passed at e.g. https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/805/testReport/OmeroWeb.test.integration.test_plategrid/TestScreenPlateTables/test_get_plate_table/

jburel:
closing this investigation work.
I will open a simplified version of it later on if required

PR #5685, opened 2018-03-14
Omeropy close service

jburel:
# What this PR does

Extracted some commits from gh-5683
Close service

# To test:
Check that the updated tests remain green

cc @chris-allan

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
@@ -373,4 +373,4 @@ class TestImage (object):
             admin = gatewaywrapper.gateway.getAdminService()
         # Check that the exporter is closed
         for v in gatewaywrapper.gateway.c.getSession().activeServices():
-            assert 'ExporterPrx' not in v, 'Leaked exporter!'
+            assert 'Exporter' not in v, 'Leaked exporter!'
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
@@ -32,6 +32,11 @@ class TestImage (object):
     def setUp(self, author_testimg):
         self.image = author_testimg
 
+    def assert_no_leaked_exporter(self, client):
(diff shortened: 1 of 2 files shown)


will-moore:
Tests are green and code looks good.
Did you want to add a test for ```image.exportOmeTiff()```?

jburel:
There is already one ``testExport`` and there is a check to be sure that the service is closed

components/tools/OmeroPy/test/integration/gatewaytest/test_image.py
```diff
@@ -367,3 +371,6 @@ def testExport(self, gatewaywrapper):
         finally:
             gatewaywrapper.loginAsAdmin()
             admin = gatewaywrapper.gateway.getAdminService()
+        # Check that the exporter is closed
+        for v in gatewaywrapper.gateway.c.getSession().activeServices():
+            assert 'ExporterPrx' not in v, 'Leaked exporter!'
```
will-moore: I removed the ```close()``` calls in BlitzGateway and this test still passed.
When I added ```print v``` to this loop, I saw this
```
5b494fb5-be0d-4af2-980b-e9d3d0066dc8omero.api.Exporter
```
which doesn't contain ```ExporterPrx```

jburel: pushing a fix i.e. Exporter instead of ExporterPrx

will-moore: I think you also need to test for open services after the ```del gen``` above. Any services left open at that point are not picked up by by the checking at the end of the test since by that time we have logged in as a different user.

jburel:
@will-moore done

will-moore:
Thanks. Testing locally with that change it does now fail when I remove the ```fin.close()``` in ```fileread_gen```.
Looks good to merge.

jburel:
Note that for the "exporter" the ``close`` was already there, i only added a ``try/finally`` in case there is a problem when reading the data.

@chris-allan are you ok with those changes?

chris-allan (approved):

PR #5686, opened 2018-03-15
public_user_security_violation_fix

will-moore:
# What this PR does

Fixes at least one cause of ```SecurityViolation``` reported in https://trello.com/c/sBdAN4L6/1-public-user-workflow-crash

I reproduced the bug by enabling public user and setting timeout to 10 secs
```
$ omero config set omero.sessions.timeout 10000
```
Then logging in as non-public user, wait > 10 secs and refresh tree (or try to expand P/D), as described by @mtbc.
This was because the ```group_id``` is still held in the page ```JavaScript``` and is used to load data in urls with ```?group=2```.
That is fixed with this commit. Instead, you will get no data, or possibly incorrect data returned but no ```SecurityViolation```.

We could check earlier in the request handling that an incorrect group_id is being used and return a 403, causing the webclient to refresh (current not-logged-in behaviour). 

On the trello card @jburel described getting this error with ```Enter http://web-dev-merge.openmicroscopy.org/webclient/``` and I don't see any error in that case. 

# Testing this PR

1. Log in as non-public user, kill session or use short timeout, then refresh tree (see above).

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/tree.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/tree.py
@@ -30,6 +30,17 @@ from copy import deepcopy
 from omero.gateway import _letterGridLabel
 
 
+def get_group_id(conn, group_id):
+    """Get a valid group ID for use in queries."""
+    if group_id is None:
+        group_id = -1
+    if group_id != -1:
+        # Make sure we don't get security violation.
+        if group_id not in conn.getEventContext().memberOfGroups:
+            group_id = -1
+    return group_id
+
+
 def build_clause(components, name='', join=','):
     ''' Build a string from a list of components.
         This is to simplify building where clauses in particular that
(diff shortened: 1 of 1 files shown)


jburel:
```
 Instead, you will get no data, or possibly incorrect data returned
```
I don't think that is an improvement in that case. I would rather have an error than incorrect data

pwalczysko:
Yes, agree, getting wrong data is more dangerous than erroring.

will-moore:
This needs a lot more thought....
In many cases, adding the ```?group=5``` filter to the URL for loading tree data is not needed and it should probably be removed. E.g. if you are loading images from ```Dataset: 2``` then you wouldn't need the group filter. However, if we want to return a ```SecurityViolation``` then this would rely on keeping the group filter since the current query "give me any images in Dataset: 2" will not fail if you are logged into a different group - it will just give you Zero images.

PR #5689, opened 2018-03-16
 resolution level

jburel:
Do not attempt to set resolution level for romio pixels buffer

# Testing this PR

1. Open a big image and draw a rectangle 3kx3k.
1. Generate an image using the image from ROIs script
1. View the image using ``/webgateway/render_image_region/ImageID/0/0/?tile=0,0,0,512,512``
this should no longer return a 404.

# Related reading

https://trello.com/c/AdRIcaBJ/37-viewing-large-non-tiled-image-fails#comment-5aabb26c699037cedb89c236

cc @will-moore

### DIFF
--- a/components/rendering/src/omeis/providers/re/Renderer.java
+++ b/components/rendering/src/omeis/providers/re/Renderer.java
@@ -18,6 +18,7 @@ import org.slf4j.LoggerFactory;
 
 import ome.conditions.ResourceError;
 import ome.io.nio.PixelBuffer;
+import ome.io.nio.RomioPixelBuffer;
 import ome.model.core.Pixels;
 import ome.model.display.ChannelBinding;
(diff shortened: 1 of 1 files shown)


joshmoore:
If possible I'd be for avoiding use of `instanceof` since it's fairly brittle. Either throwing [UnsupportedOperation](https://github.com/jburel/openmicroscopy/blob/29c240622cb178fe2cd219dbc386922535a1049a/components/romio/src/ome/io/nio/RomioPixelBuffer.java#L963) should be part of the contract in which case we should catch it, or the RPB should not throw and instead be a no-op.

cc: @chris-allan

jburel:
will look a step further in the call
I don't like that approach either

jburel:
problem is that we pass a resolution`` level 0 in render_image_region``
looking into that

PR #5297, opened 2017-05-15
OMERO.tables Image and Well metadata support

emilroz:
This PR adds the ability to display Image and Well metadata in the `Tables` section for the same Plate.

# Testing this PR

1. required setup
  - OMERO 5.3 with changes from this PR.
  - two bulk_annotations attached to the same Plate, created with 'Populate_metadata' script as described here: http://help.openmicroscopy.org/scripts.html#metadata. One bulk_annotation should contain `Well` column, the other `Image` column.

2. actions to perform
- After creating bulk annotations, check if metadata is visible for both Well and Image (that were annotated with the csv files).

3. expected observations
Metadata should be visible in `Tables` tab.

# Related reading

Link to cards, tickets, other PRs:

1. background for understanding this PR
Currently only the latest OMERO.table is searched for Image or Well metadata which means that only Well or Image metadata will be shown in `Tables` section. With this PR a query is ran against attached OMERO.tables, from the latest, until match is found. This adds the ability to show metadata for Images and Wells from the same Plate.

2. what this PR assists, fixes, or otherwise affects

PR #5691, opened 2018-03-16
Project-Dataset bulk annotations

emilroz:
# What this PR does

This PR extends `populate_metadata.py` to work also for Project->Dataset hierarchy. The extension to OMERO.web was also added to display `Tables` in the right hand side panel in the fashion similar to the Screen->Plate.

Screen->Plate workflow is also fixed with this PR.

Since there is no DatasetColumn implementation (just definition) at the moment if Dataset column is needed it should be set to String on import (not to Dataset type). The code was adjusted to resolve the Datasets internally based on their names, the only requirement is to have a column named `Dataset` in the `csv` file.

# Testing this PR

1. Create a csv file with Image column (holding the image IDs) and some other columns with data, doubles or strings.
2. Run populate_metadata.py on the Dataset or Project to which the image ids belong to.
3. Run should complete successfully
4. The project/dataset should have a new file annotation called `bulk_annotation`
5. When image is selected in the OMERO.web the data from the csv file should appear in the `Tables` section in the right hand side panel.

# Related reading
https://github.com/openmicroscopy/openmicroscopy/pull/4455
https://github.com/openmicroscopy/openmicroscopy/pull/5297
https://github.com/openmicroscopy/openmicroscopy/pull/4446

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/populate_metadata.py
+++ b/components/tools/OmeroPy/src/omero/util/populate_metadata.py
@@ -937,7 +937,7 @@ class ParsingContext(object):
         image_column = None
         image_name_column = None
         for column in self.columns:
-            columns_by_name[column.name] = column
+            columns_by_name[column.name.lower()] = column
             if column.__class__ is PlateColumn:
                 log.warn("PlateColumn is unimplemented")
             elif column.__class__ is WellColumn:
@@ -964,8 +964,8 @@ class ParsingContext(object):
                 try:
                     well_id = well_column.values[i]
                     plate = None
-                    if "Plate" in columns_by_name:  # FIXME
-                        plate = columns_by_name["Plate"].values[i]
+                    if "plate" in columns_by_name:  # FIXME
+                        plate = columns_by_name["plate"].values[i]
                     v = self.value_resolver.get_well_name(well_id, plate)
                 except KeyError:
                     log.warn(
@@ -986,8 +986,9 @@ class ParsingContext(object):
                     log.debug(image_name_column.values)
                     iid = image_column.values[i]
                     did = self.target_object.id.val
-                    if "Dataset" in columns_by_name:
-                        did = columns_by_name["Dataset"].values[i]
+                    if "dataset" in columns_by_name:
+                        dname = columns_by_name["dataset"].values[i]
+                        did = self.value_resolver.wrapper.datasets_by_name[dname].id.val
                     log.debug("Using Dataset:%d" % did)
                     iname = self.value_resolver.get_image_name_by_id(
                         iid, did)
@@ -1004,8 +1005,8 @@ class ParsingContext(object):
                 iid = image_column.values[i]
                 log.info("Checking image %s", iid)
                 pid = None
-                if 'Plate' in columns_by_name:
-                    pid = columns_by_name['Plate'].values[i]
+                if 'plate' in columns_by_name:
+                    pid = columns_by_name['plate'].values[i]
                 iname = self.value_resolver.get_image_name_by_id(iid, pid)
                 image_name_column.values.append(iname)
                 image_name_column.size = max(
@@ -1015,7 +1016,7 @@ class ParsingContext(object):
                 log.info('Missing image name column, skipping.')
 
             if plate_name_column is not None:
-                plate = columns_by_name['Plate'].values[i]   # FIXME
+                plate = columns_by_name['plate'].values[i]   # FIXME
                 v = self.value_resolver.get_plate_name_by_id(plate)
                 plate_name_column.size = max(plate_name_column.size, len(v))
                 plate_name_column.values.append(v)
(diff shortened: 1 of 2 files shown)


dominikl:
Used local setup with today's merge build (5.4.4-rc2-76-0ae83a6-ice36-b918). Web running as `omero.web.application_server development`. The `bulk_annotations` file is created and attached to the project (or dataset respectively), no error or warning messages. But the 'Tables' tab in web doesn't show anything.
My test annotations csv:
```
ImageColumn, Data_1, Data_2, Data_3
1, Test 1, 1, 1.123
2, Test 2, 2, 2.123
3, Test 3, 3, 3.123
4, Test 4, 4, 4.123
```

Console output:
```
ls29139:util dlindner$ ./populate_metadata.py  -s localhost -p 4064 -u root Project:1 /Users/dlindner/test_annotations.csv 
Password: 
INFO:omero.util.Resources:Starting
INFO:omero.util.populate_metadata:Adding dataset:2 image:4
INFO:omero.util.populate_metadata:Adding dataset:2 image:3
INFO:omero.util.populate_metadata:Adding dataset:1 image:2
INFO:omero.util.populate_metadata:Adding dataset:1 image:1
INFO:omero.util.populate_metadata:Nothing to do during post processing.
INFO:omero.util.populate_metadata:Created new table OriginalFile:94
INFO:omero.util.populate_metadata:Table initialized with 4 columns.
INFO:omero.util.populate_metadata:Added 4 rows of column data (batch 1)
INFO:omero.util.Resources:Halted
```

(Just tested it on eel, too. Same there, `bulk_annotations` is there, but 'Tables' tab empty)

will-moore:
Need to check whether tables test failures are due to this PR: https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/809/testReport/

emilroz:
The current test failures are due to the fact that before this PR `Project->Dataset` was resolved using `Image Names` now it's all moved to Image Ids. I look into resolving the issue.

emilroz:
Found few issues which needs to be resolved first. Closing it until it's ready.

emilroz:
I expanded the scope to allow for all the permutation of image names, ids and dataset names and ids to be used to populate metadata I hope this will let the tests pass now.

@dominikl the Table view did not work because the queries used to retrieve the metadata from OMERO.tables expect the column with image id to be called `Image` (case sensitive).

All supported csv types for Project/Dataset hierarchy:

* Using image id:
```
Image,Feature 1,Feature 2
iid1,Value 1,Value A
iid2,Value 2,Value B
```
* Using image name
```
Image Name,Feature 1,Feature 2
iname1,Value 1,Value A
iname2,Value 2,Value B
```
* Using dataset id and image id:
```
Dataset,Image,Feature 1,Feature 2
did,iid1,Value 1,Value A
did,iid2,Value 2,Value B
```
* Using dataset name and image id:
```
Dataset Name,Image,Feature 1,Feature 2
dname,iid1,Value 1,Value A
dname,iid2,Value 2,Value B
```
* Using dataset id and image name
```
Dataset,Image Name,Feature 1,Feature 2
did,iname1,Value 1,Value A
did,iname2,Value 2,Value B
```
* Using dataset name and image name
```
Dataset Name,Image Name,Feature 1,Feature 2
dname,iname1,Value 1,Value A
dname,iname2,Value 2,Value B
```

pwalczysko:
I understand that the bigger plan here is to merge the usecase of the IDR (using the metadata CLI plugin) and this one ? What about the script which is to be found in the UI (see screenshot) ? 
@jburel asked me to test this PR, but this might be pre-mature considering the decisions above ?
In any case, the help parts of this script here need to be updated, to reflect the new changes.


screen shot 2018-04-24 at 11 06 22

joshmoore:
And now after far too much silence, a summary of various thoughts (correct as necessary):

 * There's an assumption that this is not immediately needed on the develop branch.
 * There's also an assumption that this fixes the breakage caused by forward-porting the metadata53 work to develop.
 * There's currently no difference between metadata54 and develop.
 * So it _might_ be that this breaks the IDR workflows (requiring changes to the CSVs) but that's acceptable and just needs doing.
 * However, it's generally assumed to be better to handle this work on https://github.com/ome/omero-cli-metadata rather than the mainline. Our steps to get there would be:
   - a separate PR to remove `bin/omero metadata` from the mainline
   - finalize @sbesson's refactoring of https://github.com/ome/omero-cli-metadata
   - open this work against https://github.com/ome/omero-cli-metadata
   - migrate the script that Petr mentions to https://github.com/ome/omero-cli-metadata (with the added difficulty of needing a way to deploy it)

cc: @pwalczysko & @emilroz

sbesson:
Following up on https://github.com/openmicroscopy/openmicroscopy/pull/5691#issuecomment-388320009, the current state of the populate metadata code in OMERO 5.4.7 has now been filtered under https://github.com/ome/omero-metadata, tagged as version 0.2.1 and released to PyPI - see https://pypi.org/project/omero-metadata/0.2.1/.

This means:
- the release cycle of this metadata plugin is now decoupled from the upstream OMERO release
- consumers can also install and control/pin the version of the metadata plugin they use independently of OMERO.py
- the populate metadata code will be deprecated in this repository

As per the last statement, I am going to close this contribution but as in https://github.com/openmicroscopy/openmicroscopy/pull/5691#issuecomment-388320009, I propose to re-open the bulk of these changes against `omero-metadata` so that we can discuss the roadmap towards a `omero-metadata:0.3.0` release with P/D/I support. The corresponding OMERO.web changes might need to be re-opened as a separate PR.

PR #5692, opened 2018-03-16
render_image_region

jburel:
# What this PR does

Ignore resolution level for non pyramid file
The problem is also present for small image as soon as the resolution specified was not 0.

# Testing this PR

Test 1:
1. Open a big image and draw a rectangle 3kx3k.
1. Generate an image using the image from ROIs script
1. View the image using ``/webgateway/render_image_region/ImageID/0/0/?tile=0,0,0,512,512``
this should no longer return a 404.

Test 2:
1. Open any small image e.g. dv
1. View the image using ``/webgateway/render_image_region/ImageID/0/0/?tile=1,0,0,512,512``
A 400 will be returned and prevent error like
```
repo/openmicroscopy/dist/lib/python/omero_api_PyramidService_ice.py", line 324, in setResolutionLevel
    return _M_omero.api.PyramidService._op_setResolutionLevel.invoke(self, ((resolutionLevel, ), _ctx))

```

Test 3:
* Open any pyramid
* View the image using ``/webgateway/render_image_region/ImageID/0/0/?tile=0,0,0,512,512``
* View the image using ``/webgateway/render_image_region/ImageID/0/0/?region=0,0,512,512``

Test 4:
Negative resolution
1. Open any small image e.g. dv
1. View the image using ``/webgateway/render_image_region/ImageID/0/0/?tile=-1,0,0,512,512``
A "invalid resolution message" will be returned

cc @will-moore @chris-allan 

problem test 1 reported https://trello.com/c/AdRIcaBJ/37-viewing-large-non-tiled-image-fails
test 2: found while investigating

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8982,11 +8982,8 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                 tile_height = round(tile_height / 2.0)
             width = int(tiles_wide * tile_width)
             height = int(tiles_high * tile_height)
-            level = 0
-            if not re.requiresPixelsPyramid():
-                level = None
             jpeg_data = self.renderJpegRegion(
-                z, t, x, y, width, height, level=level)
+                z, t, x, y, width, height, level=0)
             if size is None:
                 return jpeg_data
             # We've been asked to scale the image by its longest side so we'll
@@ -9034,8 +9031,14 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         regionDef.height = int(height)
         self._pd.region = regionDef
         try:
-            if level is not None and self._re.requiresPixelsPyramid():
-                self._re.setResolutionLevel(level)
+            if level is not None:
+                if self._re.requiresPixelsPyramid():
+                    self._re.setResolutionLevel(level)
+                else:
+                    if level != 0:
+                        logger.debug('On renderJpegRegion')
+                        logger.debug('Cannot set resolution %s' % level)
+                        return None
             if compression is not None:
                 try:
                     self._re.setCompressionLevel(float(compression))
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8982,8 +8982,11 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                 tile_height = round(tile_height / 2.0)
             width = int(tiles_wide * tile_width)
             height = int(tiles_high * tile_height)
+            level = 0
+            if not re.requiresPixelsPyramid():
+                level = None
             jpeg_data = self.renderJpegRegion(
-                z, t, x, y, width, height, level=0)
+                z, t, x, y, width, height, level=level)
             if size is None:
                 return jpeg_data
             # We've been asked to scale the image by its longest side so we'll
@@ -9031,7 +9034,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         regionDef.height = int(height)
         self._pd.region = regionDef
         try:
-            if level is not None:
+            if level is not None and self._re.requiresPixelsPyramid():
                 self._re.setResolutionLevel(level)
             if compression is not None:
                 try:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -9032,13 +9032,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         self._pd.region = regionDef
         try:
             if level is not None:
-                if self._re.requiresPixelsPyramid():
-                    self._re.setResolutionLevel(level)
-                else:
-                    if level != 0:
-                        logger.warn('On renderJpegRegion')
-                        logger.warn('Cannot set resolution %s' % level)
-                        return None
+                self._re.setResolutionLevel(level)
             if compression is not None:
                 try:
                     self._re.setCompressionLevel(float(compression))
@@ -9049,6 +9043,11 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                     return self.renderJpeg(z, t, None)
             rv = self._re.renderCompressed(self._pd, self._conn.SERVICE_OPTS)
             return rv
+        except omero.ApiUsageException:
+            logger.debug('On renderJpegRegion')
+            logger.debug('Cannot set resolution %s' % level)
+            logger.debug(traceback.format_exc())
+            return None
         except omero.InternalException:  # pragma: no cover
             logger.debug('On renderJpegRegion')
             logger.debug(traceback.format_exc())
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -9036,8 +9036,8 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                     self._re.setResolutionLevel(level)
                 else:
                     if level != 0:
-                        logger.debug('On renderJpegRegion')
-                        logger.debug('Cannot set resolution %s' % level)
+                        logger.warn('On renderJpegRegion')
+                        logger.warn('Cannot set resolution %s' % level)
                         return None
             if compression is not None:
                 try:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -9043,14 +9043,8 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                     return self.renderJpeg(z, t, None)
             rv = self._re.renderCompressed(self._pd, self._conn.SERVICE_OPTS)
             return rv
-        except omero.ApiUsageException:
-            logger.debug('On renderJpegRegion')
-            logger.debug('Cannot set resolution %s' % level)
-            logger.debug(traceback.format_exc())
-            return None
-        except omero.InternalException:  # pragma: no cover
-            logger.debug('On renderJpegRegion')
-            logger.debug(traceback.format_exc())
+        except (omero.ApiUsageException, omero.InternalException):
+            logger.debug('On renderJpegRegion', exc_info=True)
             return None
         except Ice.MemoryLimitException:  # pragma: no cover
             # Make sure renderCompressed isn't called again on this re,
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -946,10 +946,9 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
             x = int(zxyt[1])*w
             y = int(zxyt[2])*h
         except:
-            logger.debug(
-                "render_image_region: tile=%s" % tile, exc_info=True
-            )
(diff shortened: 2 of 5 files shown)


components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -926,6 +926,9 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
             level = levels-int(zxyt[0])
             x = int(zxyt[1])*w
             y = int(zxyt[2])*h
+            # check if the image requires pyramid
+            if not img._re.requiresPixelsPyramid():
+                level = None
```
will-moore: Might be nicer to have this check happen in the BlitzGateway ```renderJpegRegion()``` so that BlitzGateway users benefit from it too?

jburel: Need to check impact on method like ``renderBirdsEyeView``

jburel:
I have covered all cases I hope with tests unless you think there is a thing I am still missing

will-moore:
With large images from API (e.g. OMERO.figure export) the level in ```?tile=``` is ignored so these now render fine. E.g. http://web-dev-merge.openmicroscopy.org/webgateway/render_image_region/78503/0/0/?tile=10000,1,4,512,512 (user-3)
This means that these images now view fine in iviewer because it never asks for ```level=1``` but *if* it did then we'd silently display tiles at ```level=0``` instead of 404, which is maybe not the expected behaviour. If we don't support any levels other than 0 then we should probably raise a 404, as we currently do if the tile is outside the image bounds, (ideally with a helpful message).

Big images still view fine at all levels.
Small images render fine with ```tile=0,0,0,512,512```

jburel:
@will-moore changes discussed and tests updated accordingly. Could you have a look at it?

jburel:
I have also fixed the negative resolution case. an ``HttpResponseBadRequest`` will be returned in that case

will-moore:
Works for me nicely. All non-zero levels for non-tiled images are handled with 404 or invalid message etc. Tiled images working as before.

jburel:
@will-moore could you check the negative resolution level? I have just updated the description

will-moore:
e.g. /webgateway/render_image_region/72493/0/0/?tile=-1,1,2,512,512 gave me "Invalid resolution" in the browser.

jburel:
Message is the expected one

jburel:
All tests modified or added in this PR are green

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -9032,7 +9032,13 @@ def renderJpegRegion(self, z, t, x, y, width, height, level=None,
         self._pd.region = regionDef
         try:
             if level is not None:
-                self._re.setResolutionLevel(level)
+                if self._re.requiresPixelsPyramid():
+                    self._re.setResolutionLevel(level)
+                else:
+                    if level != 0:
```
chris-allan: This feels like it should be a hard error. Probably an HTTP 400.

chris-allan (changes requested):
While the sanity check here does make sense, I do feel that at some level the exception is accurate. There will be a performance hit as a result of this change and it will now apply across the board to every use of tile images.

Before we move forward with merging this it would be good to quantify the performance impact. Perhaps even making the change server side so that setting resolution level 0 on a non-tiled image does not fail hard.

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel: The method will return ``None`` and that will result in a ``404`` in views.py

chris-allan: Yes, which is not really an appropriate error in this case.

jburel: I read 404 in your comment

jburel:
We could ignore in Romio when the value is set instead of throwing an exception
That should not affect the performance

chris-allan:
That would certainly be worth investigating. However, I don't know what the impact of that change may be.

components/tools/OmeroPy/src/omero/gateway/__init__.py
jburel: in that case we will need a check too in views.py so it does not change the existing behaviour i.e. none => 404
that will have an impact on performance since we will have to handle web usage and gateway usage

chris-allan: Just looking at the overarching semantics, it looks like we have always been catching the `omero.InternalException` and returning `None`. While I don't really agree with that decision, it has been the prevailing semantic for some time now and I don't think it's a good idea to change that. Checking the level value preemptively is not really something you need to do here just for the sake of avoiding an exception.

Incidentally, these exceptions should probably be logged at `WARN` rather than `DEBUG` so that we actually see them in the logs.

will-moore: With the omero.InternalException handling, ```conn.renderJpegRegion()``` will also return ```None``` if you specify an invalid ```?tile=3,5,100,256,256``` (tile outside of image) for a tiled image, so having it return ```None``` for an invalid ```level``` (for a non-tiled image) seems consistent.

This hides the inconsistency on the server where the first will throw a ```ome.conditions.InternalException:  Wrapped Exception: (java.lang.RuntimeException):
Invalid Region, the height must be positive:0``` and the second will simply render something other than what you might expect.

Now that @waxenegger has reverted to using ```?region= ``` for large non-tiled images (https://github.com/ome/omero-iviewer/pull/186/commits/71d9f0d7e23a8dfd5e8cd3abf8a2d9a378e13fc6) this is no-longer so urgent.
Seems we can't move these changes to the server for a point release since it's an API change, but I would have thought that the changes in this PR are a good first step?

mtbc: Outwith this PR, do feel free to open a design issue that accumulates suggested server exception changes that would improve the API usability - maybe we can try to take on a bunch next year or whenever next time we're breaking everything.

jburel: The fix is not bound to iviewer.
There are few usage issues with the methods.
As @chris-allan pointed out, more check => potential performance issue

jburel:
@will-moore For every tile, the added check is:
```
public boolean requiresPixelsPyramid(Pixels pixels) {
        String type = pixels.getPixelsType().getValue();
        if ("float".equals(type) || "double".equals(type))
            return false;
        final long sizeX = pixels.getSizeX();
        final long sizeY = pixels.getSizeY();
        final boolean requirePyramid = (sizeX * sizeY) > (sizes.getMaxPlaneWidth()*sizes.getMaxPlaneHeight());
        return requirePyramid;
    }
```

jburel:
@will-moore @chris-allan I have removed the usage of ``requiresPixelsPyramid`` in the gateway, instead the method now handles ``ApiUsageException``.
We now check if it is a pyramid using the ``levels`` variable already used, this should no longer impact  performance
The only limitation is that the non pyramid image generated from API (Test 1) will not be rendered if directly retrieved from the gateway but it will be viewable using Web

That's the best we can do at the moment without potential performance impact

Tests updated accordingly

will-moore:
Testing with a large (non-tiled) api-generated OR imported image:
 - Valid ```?tile=0,1,2,512,512``` works fine.
 - Invalid ```?tile=0,100,2,512,512``` is 404.
 - Invalid level ```?tile=1,1,2,512,512``` returns "Invalid resolution"

Testing with Big (tiled) image:
  - Valid ```?tile=0,1,2,512,512``` works fine.
 - Invalid ```?tile=0,1000,2,512,512``` is 404.
 - Invalid level (max is 4) ```?tile=5,0,0,100,100``` is 404.

Looks good.

jburel:
``Invalid level (max is 4) ?tile=5,0,0,100,100 is 404.`` in that case it should be "Invalid resolution" to be consistent but that will be a change in the actual behaviour
What do you think?

Regardless I will add a test since it is not covered

jburel:
@will-moore it should no longer be a 404 when an invalid resolution is specified for pyramid
It should now match the rest. 
Test added

will-moore:
```?tile=5,0,0,100,100``` now gives "invalid resolution" instead of 404 for a tiled image.
Working as expected and code looks good.
Good to merge.

**Comment by jburel** [comment]
@chris-allan ok with the last set of changes?

**Comment by chris-allan** [comment]
I will need at least a day to have a chance to look at these seriously.

File: components/tools/OmeroPy/src/omero/gateway/__init__.py [hunk 3]
```diff
@@ -9043,6 +9043,11 @@ def renderJpegRegion(self, z, t, x, y, width, height, level=None,
                     return self.renderJpeg(z, t, None)
             rv = self._re.renderCompressed(self._pd, self._conn.SERVICE_OPTS)
             return rv
+        except omero.ApiUsageException:
+            logger.debug('On renderJpegRegion')
+            logger.debug('Cannot set resolution %s' % level)
```
chris-allan: This does not necessarily indicate that the resolution cannot be set as the `ApiUsageException` could have been thrown by any of the remote invocations here. I would not include this log line. You can also use the following form to avoid having to call the `traceback` module and end up logging separately:

```
...
logger.debug('On renderJpegRegion', exc_info=True)
...
```

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -923,7 +923,30 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
                     if tile_length > max_tile_length:
                         tile_size[i] = max_tile_length
                 w, h = tile_size
-            level = levels-int(zxyt[0])
+            v = int(zxyt[0])
+            if v < 0:
+                logger.debug(
+                    "render_image_region: invalid level %s" % v, exc_info=True
+                )
+                return HttpResponseBadRequest('Invalid resolution')
```
chris-allan: Probably useful to include the resolution and reason in the response, `'Invalid resolution level, %d < 0' % v`

Depending on what you want to do you might use the same message in both the log statement and the response. There is also no reason to include context (`render_image_region`) in the message as that is included, along with the line number, in the standard log format.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -923,7 +923,30 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
...
+
+            if levels == 0:  # non pyramid file
+                if v > 0:
+                    logger.debug(
+                        "render_image_region: invalid level %s" % v,
+                        exc_info=True
+                    )
+                    return HttpResponseBadRequest('Invalid resolution')
```
chris-allan: Probably useful to include the resolution and reason in the response, `'Invalid resolution level %d, non pyramid file' % v`

Depending on what you want to do you might use the same message in both the log statement and the response. There is also no reason to include context (`render_image_region`) in the message as that is included, along with the line number, in the standard log format.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -923,7 +923,30 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
...
+                else:
+                    level = None
+            else:
+                level = levels-v
+                if level < 0:
+                    logger.debug(
+                        "render_image_region: invalid level %s" % v,
+                        exc_info=True
+                    )
+                    return HttpResponseBadRequest('Invalid resolution')
```
chris-allan: Probably useful to include the resolution and reason in the response, `'Invalid resolution level, %d > number of available levels %d' % (v, levels)`

Depending on what you want to do you might use the same message in both the log statement and the response. There is also no reason to include context (`render_image_region`) in the message as that is included, along with the line number, in the standard log format.

chris-allan (changes requested):
Logic is sound. Some comments on return values and logging style.

jburel:
@chris-allan: done. I also changed log of tile/region to keep things in synch (I was following the pattern in place)

chris-allan (approved):

PR #5693, opened 2018-03-19
Remove testPerformanceAttachFile integration test

dominikl:
# What this PR does

Simply excludes the integration test, which tests the performance of parallel file attachments.

The test compares single and parallel file attachments. But when the CI performance itself isn't very stable this integration test becomes very flaky.

### DIFF
--- a/components/tools/OmeroJava/test/integration/gateway/DataManagerFacilityTest.java
+++ b/components/tools/OmeroJava/test/integration/gateway/DataManagerFacilityTest.java
@@ -319,8 +319,8 @@ public class DataManagerFacilityTest extends GatewayTest {
      * 
      * @throws Exception
      */
-    //@Test(groups = "broken") // marked as 'broken', works fine locally, but flaky on CI
-    @Test(dependsOnMethods = { "testSaveAndReturnObject" })
+    @Test(groups = "broken") // Flaky due to general CI performance issues
+    //@Test(dependsOnMethods = { "testSaveAndReturnObject" })
     public void testPerformanceAttachFile() throws Exception {
         long start = System.currentTimeMillis();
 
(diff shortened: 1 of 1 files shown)


components/tools/OmeroJava/test/integration/gateway/DataManagerFacilityTest.java
```diff
@@ -319,8 +319,8 @@ public void testAttachFile() throws Exception {
      * 
      * @throws Exception
      */
-    //@Test(groups = "broken") // marked as 'broken', works fine locally, but flaky on CI
-    @Test(dependsOnMethods = { "testSaveAndReturnObject" })
+    @Test(groups = "broken") // Flaky due to general CI performance issues
+    //@Test(dependsOnMethods = { "testSaveAndReturnObject" })
```
mtbc: Interesting, this causes a problem if left in?

joshmoore:
Is this long-term or short-term? Is the test broken? Do you want to see it fixed? If the latter, then I'd assume we'd need to perform the two actions in a loop and compare the set of times. (Is it conceivable that there's a concurrency issue that the test is pointing to?)

dominikl:
The problem is that the test compares two methods, and if the performance of the infrastructure is kind of random between the different runs, then the outcome of the test is random, too. It works fine locally (e.g. avgSingleUploadDuration=690, avgParallelUploadDurationPerFile=281) but not on the CI, so I don't think that test is broken or reveals some concurrency issue. 
I should just remove the test. As it is now actually (already allows the parallel file attachment to be 1.5 times slower than single upload before it fails) it doesn't really make sense anyway.
Any objections to just remove it?

joshmoore:
Sounds better than commenting it out. Long-term it would be nice to be able to support this kind of test, but I definitely understand that that's easier said than done.

manics (approved):
Existng tests are passing, and it clearly solves the problem of a flaky test!

PR #5696, opened 2018-03-21
Metadata plugin pixelsize

dominikl:
# What this PR does

Adds a `pixelsize` command to the metadata plugin, which allows setting a custom physical pixel size for an Image (or whole Dataset, Project, etc.)

# Testing this PR

Set a custom pixel size:
`./omero metadata pixelsize --x 5 --y 6 --z 7 --unit nanometer Dataset:123 `

Doesn't output anything, so use Insight or Web to check if it was successful.

# Related reading

https://github.com/IDR/idr0042-nirschl-wsideeplearning/issues/1

/cc @sbesson

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/metadata.py
+++ b/components/tools/OmeroPy/src/omero/plugins/metadata.py
@@ -24,7 +24,7 @@ from omero.util import populate_metadata, populate_roi, pydict_text_io
 from omero.util.metadata_utils import NSBULKANNOTATIONSCONFIG
 from omero.util.metadata_utils import NSBULKANNOTATIONSRAW
 from omero.grid import LongColumn
-
+from omero.model.enums import UnitsLength
 
 HELP = """Metadata utilities
 
@@ -148,6 +148,7 @@ class MetadataControl(BaseControl):
         rois = parser.add(sub, self.rois)
         populate = parser.add(sub, self.populate)
         populateroi = parser.add(sub, self.populateroi)
+        pixelsize = parser.add(sub, self.pixelsize)
 
         populate.add_argument("--batch",
                               type=long,
@@ -156,7 +157,7 @@ class MetadataControl(BaseControl):
         self._add_wait(populate)
 
         for x in (summary, original, bulkanns, measures, mapanns, allanns,
-                  rois, populate, populateroi):
+                  rois, populate, populateroi, pixelsize):
             x.add_argument("obj",
                            type=ProxyStringType(),
                            help="Object in Class:ID format")
@@ -210,6 +211,16 @@ class MetadataControl(BaseControl):
         populateroi.add_argument(
             "--measurement", type=int, default=None,
             help="Index of the measurement to populate. By default, all")
+
+        pixelsize.add_argument(
+            "--x", type=float, default=None, help="Physical pixel size X")
+        pixelsize.add_argument(
+            "--y", type=float, default=None, help="Physical pixel size Y")
+        pixelsize.add_argument(
+            "--z", type=float, default=None, help="Physical pixel size Z")
+        pixelsize.add_argument(
+            "--unit", default="um", help="Unit (e.g. nm, um, mm, ...) "
+                                         "(default: um)")
 
     def _clientconn(self, args):
         client = self.ctx.conn(args)
@@ -553,6 +564,87 @@ class MetadataControl(BaseControl):
                         continue
                 meas = ctx.get_measurement_ctx(i)
                 meas.parse_and_populate()
+
+    def pixelsize(self, args):
+        "Set physical pixel size"
+        md = self._load(args)
+        client, conn = self._clientconn(args)
+
+        units = {
+            'nm': UnitsLength.NANOMETER,
+            'um': UnitsLength.MICROMETER,
+            'mm': UnitsLength.MILLIMETER,
+            'cm': UnitsLength.CENTIMETER,
+            'm': UnitsLength.METER,
+            'km': UnitsLength.KILOMETER
+        }
+
+        if args.unit not in units:
+            self.ctx.die(100, "%s not recognized as valid unit." % args.unit)
+
+        if not args.x and not args.y and not args.z:
+            self.ctx.die(100, "No pixel sizes specified.")
+
+        if md.get_type() == "Screen":
+            q = """SELECT pix FROM Pixels pix, WellSample ws, Plate p,
+                   ScreenPlateLink spl WHERE
+                   spl.child=p AND pix.image=ws.image AND ws.well.plate=p AND
+                   spl.parent.id=:id"""
+        elif md.get_type() == "Plate":
+            q = """SELECT pix FROM Pixels pix, WellSample ws WHERE
+                   pix.image=ws.image AND ws.well.plate.id=:id"""
+        elif md.get_type() == "PlateAcquisition":
+            q = """SELECT pix FROM Pixels pix, WellSample ws WHERE
+                   pix.image=ws.image AND ws.plateAcquisition.id=:id"""
+        elif md.get_type() == "Well":
+            q = """SELECT pix FROM Pixels pix, WellSample ws WHERE
+                   pix.image=ws.image AND ws.well.id=:id"""
+        elif md.get_type() == "Project":
+            q = """SELECT pix FROM Pixels pix, DatasetImageLink dil,
+                   ProjectDatasetLink pdl WHERE dil.child=pix.image AND
+                   dil.parent=pdl.child AND pdl.parent.id=:id"""
+        elif md.get_type() == "Dataset":
+            q = """SELECT pix FROM Pixels pix, DatasetImageLink dil WHERE
+                   dil.child=pix.image AND dil.parent.id=:id"""
+        elif md.get_type() == "Image":
+            q = """SELECT pix FROM Pixels pix WHERE pix.image.id=:id"""
+        else:
+            raise Exception("Not implemented for type %s" % md.get_type())
+
+        params = omero.sys.ParametersI()
+        params.addId(md.get_id())
+        pixels = client.getSession().getQueryService()\
+                                    .findAllByQuery(q, params)
+        for pixel in pixels:
+            if args.x:
+                pixSize = pixel.getPhysicalSizeX()
+                if not pixSize:
+                    pixSize = omero.model.LengthI(args.x, units[args.unit])
+                    pixel.setPhysicalSizeX(pixSize)
+                else:
+                    pixSize.setValue(args.x)
+                    pixSize.setUnit(units[args.unit])
+
+            if args.y:
+                pixSize = pixel.getPhysicalSizeY()
+                if not pixSize:
+                    pixSize = omero.model.LengthI(args.y, units[args.unit])
+                    pixel.setPhysicalSizeY(pixSize)
+                else:
+                    pixSize.setValue(args.y)
+                    pixSize.setUnit(units[args.unit])
+
+            if args.z:
+                pixSize = pixel.getPhysicalSizeZ()
+                if not pixSize:
+                    pixSize = omero.model.LengthI(args.z, units[args.unit])
+                    pixel.setPhysicalSizeZ(pixSize)
+                else:
+                    pixSize.setValue(args.z)
+                    pixSize.setUnit(units[args.unit])
+
+            client.getSession().getUpdateService().saveAndReturnObject(pixel)
+
 
 try:
     if "OMERO_DEV_PLUGINS" in os.environ:
--- a/components/tools/OmeroPy/src/omero/plugins/metadata.py
+++ b/components/tools/OmeroPy/src/omero/plugins/metadata.py
@@ -604,20 +604,22 @@ class MetadataControl(BaseControl):
         else:
             raise Exception("Not implemented for type %s" % md.get_type())
 
+        ctx = {'omero.group': '-1'}
+
         params = omero.sys.ParametersI()
         params.addId(md.get_id())
-        pixels = client.getSession().getQueryService()\
+        pixels = client.getSession().getQueryService(ctx)\
                                     .findAllByQuery(q, params)
 
         for pixel in pixels:
             if args.x:
-                pixel.setPhysicalSizeX(omero.model.LengthI(args.x, unit))
+                pixel.setPhysicalSizeX(omero.model.LengthI(args.x, unit), ctx)
             if args.y:
-                pixel.setPhysicalSizeY(omero.model.LengthI(args.y, unit))
+                pixel.setPhysicalSizeY(omero.model.LengthI(args.y, unit), ctx)
             if args.z:
-                pixel.setPhysicalSizeZ(omero.model.LengthI(args.z, unit))
+                pixel.setPhysicalSizeZ(omero.model.LengthI(args.z, unit), ctx)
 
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -554,6 +565,87 @@ def populateroi(self, args):
...
+            raise Exception("Not implemented for type %s" % md.get_type())
+
+        params = omero.sys.ParametersI()
+        params.addId(md.get_id())
+        pixels = client.getSession().getQueryService()\
+                                    .findAllByQuery(q, params)
+        for pixel in pixels:
+            if args.x:
+                pixSize = pixel.getPhysicalSizeX()
+                if not pixSize:
```
jburel: to avoid copy/paste of similar code, you should introduce a method to the set the physical size

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -554,6 +565,87 @@ def populateroi(self, args):
...
+            q = """SELECT pix FROM Pixels pix, DatasetImageLink dil,
+                   ProjectDatasetLink pdl WHERE dil.child=pix.image AND
+                   dil.parent=pdl.child AND pdl.parent.id=:id"""
+        elif md.get_type() == "Dataset":
+            q = """SELECT pix FROM Pixels pix, DatasetImageLink dil WHERE
+                   dil.child=pix.image AND dil.parent.id=:id"""
+        elif md.get_type() == "Image":
+            q = """SELECT pix FROM Pixels pix WHERE pix.image.id=:id"""
+        else:
+            raise Exception("Not implemented for type %s" % md.get_type())
```
mtbc, line 605: This is fine but just to mention that another option is http://downloads.openmicroscopy.org/latest/omero5.4/api/slice2html/omero/cmd/FindChildren.html#FindChildren to find their IDs with `typesOfChildren = ['Pixels']` and `stopBefore = ['Roi']`.

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -554,6 +565,87 @@ def populateroi(self, args):
...
+                else:
+                    pixSize.setValue(args.y)
+                    pixSize.setUnit(units[args.unit])
+
+            if args.z:
+                pixSize = pixel.getPhysicalSizeZ()
+                if not pixSize:
+                    pixSize = omero.model.LengthI(args.z, units[args.unit])
+                    pixel.setPhysicalSizeZ(pixSize)
+                else:
```
mtbc: Are these `else` worthwhile or simpler just to always do the first clause? (Can instantiate the `LengthI` outside the loop so the pixels share it?)

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -554,6 +565,87 @@ def populateroi(self, args):
...
+                    pixSize.setValue(args.z)
+                    pixSize.setUnit(units[args.unit])
+
+            client.getSession().getUpdateService().saveAndReturnObject(pixel)
```
mtbc: Could instead save many in bulk with `saveArray` outside the loop.

components/tools/OmeroPy/src/omero/plugins/metadata.py
dominikl: Always create a new `LengthI` object? Is the old one automatically removed from the database then?

joshmoore: Yes, definitely. They are `structs`.

mtbc: They don't exist in the database as first-class objects.

components/tools/OmeroPy/src/omero/plugins/metadata.py
dominikl: Do we have a python example somewhere which uses `FindChildren`? Can't get my head around how to use that.

mtbc:
```python
from omero.callbacks import CmdCallbackI
from omero.cmd import FindChildren

conn = BlitzGateway(stuff)  # I suggest user-2 on eel
conn.connect()

req = FindChildren(targetObjects={'Project': [1]},
                   typesOfChildren=['Pixels'], stopBefore=['Roi'])
prx = conn.c.sf.submit(req)
print CmdCallbackI(conn.c, prx).loop(500, 500)

conn._closeSession()
```
and see many IDs of Pixels that are in Project 1.

dominikl: Thanks! That works. But I think I leave the the hql query for now, because it loads the Pixel objects directly. Otherwise I'd have to get the ids first with `FindChildren` and then load the Pixel objects from the id list.

dominikl: Btw, that example would be nice to have on https://docs.openmicroscopy.org/omero/5.4.4/developers/Python.html :-)

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -554,6 +565,73 @@ def populateroi(self, args):
...
                 meas.parse_and_populate()
 
+    def pixelsize(self, args):
+        "Set physical pixel size"
+        md = self._load(args)
+        client, conn = self._clientconn(args)
+
+        units = {
+            'nm': UnitsLength.NANOMETER,
+            'um': UnitsLength.MICROMETER,
```
mtbc: Ought we also allow 'μm'?

mtbc:
Not sure `_setPixelsize` buys us much but either way certainly looking better! :+1:

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -554,6 +565,73 @@ def populateroi(self, args):
                 meas = ctx.get_measurement_ctx(i)
                 meas.parse_and_populate()
 
+    def pixelsize(self, args):
+        "Set physical pixel size"
+        md = self._load(args)
+        client, conn = self._clientconn(args)
+
+        units = {
```
joshmoore: This lookup can be achieved via:

```
In [34]: omero.model.LengthI.SYMBOLS.keys()[omero.model.LengthI.SYMBOLS.values().index("mm")]
Out[34]: 'MILLIMETER'
```

so that no hard-coding is needed. You'd have to special case `um`, though, regardless and unicode is going to be challenging from the terminal.

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -211,6 +212,16 @@ def _configure(self, parser):
...
             help="Index of the measurement to populate. By default, all")
 
+        pixelsize.add_argument(
+            "--x", type=float, default=None, help="Physical pixel size X")
+        pixelsize.add_argument(
+            "--y", type=float, default=None, help="Physical pixel size Y")
+        pixelsize.add_argument(
+            "--z", type=float, default=None, help="Physical pixel size Z")
+        pixelsize.add_argument(
+            "--unit", default="um", help="Unit (e.g. nm, um, mm, ...) "
```
joshmoore: If the sizes need different units, the suggestion would be to call the plugin multiple times?

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -554,6 +565,73 @@ def populateroi(self, args):
                 meas = ctx.get_measurement_ctx(i)
                 meas.parse_and_populate()
 
+    def pixelsize(self, args):
+        "Set physical pixel size"
+        md = self._load(args)
+        client, conn = self._clientconn(args)
```
joshmoore: Might be worth doing this after validating the inputs.

joshmoore (commented):
A few minor comments added. In general, working with units and the related enums is tricky at best. Just trying to write a snippet for this PR, I already noticed 2-3 helper methods that should be created. But that's for down the road.

Long-term I think we should support 1 or 2 forms of _all_ enum values for a given unit:

 - MILLIMETER
 - NANOMETER
 - etc

and/or

 - mm
 - nm
 - etc.

As long as we're happy to formalize that later *and* continue to support any extra helpers introduced in this PR (like `um`) then no objections to not having complete support here.

If there are any concerns over the "one-unit-fits-all" issue, a next step after this PR might be to use the (yet another) helper suggested in https://trello.com/c/Z2NAGjBQ/379-cli-support-enums-for-objpy such that:

```
 --x=1mm
```
would be a valid argument.

dominikl:
I'd prefer only using the enum names like `--x=1MILLIMETER` (guess we could ignore the case, e. g.  ~ `--x=1millimeter` ). It's horrible to type and look at, but at least avoids any unicode issues. But especially then one should have at least the option to specify the unit only once and use it for x, y and z; which is the usual case anyway I'd assume.

dominikl:
I'll leave x, y, z parameter as value only parameter for now. A unit can be specified with `--unit` parameter and is applied to x, y and z, like before, but I removed the custom unit parsing and use the enum names directly instead ("nanometer", "micrometer", ...). It's just too much hassle to parse stuff like `--x=1mm` or `--x=1inch`  or could also be `--x=1.0ATTOMETER `, etc. and can be added later, without breaking current behaviour.

joshmoore:
This has the same issue as addressed in https://github.com/ome/omero-cli-render/pull/6 in that a non-owner thinks that the change has taken place but silently fails:

```
[jamoore@ome-c6100-1 dist]$ bin/omero hql -q --all "select i from Pixels i order by i.id desc" --limit=2
 # | Class   | Id    | sizeT | dimensionOrder    | sha1                                     | physicalSizeY             | physicalSizeX             | image        | physicalSizeZ            | sizeX | sizeY | sizeZ | details         | sizeC | pixelsType    | significantBits
---+---------+-------+-------+-------------------+------------------------------------------+---------------------------+---------------------------+--------------+--------------------------+-------+-------+-------+-----------------+-------+---------------+-----------------
 0 | PixelsI | 96151 | 200   | DimensionOrderI:1 | fd52d62a98eef4c2d525f176c18afdbb83ad8c47 | 0.112484999001 MICROMETER | 0.112484999001 MICROMETER | ImageI:98401 | 0.20000000298 MICROMETER | 256   | 256   | 1     | owner=5;group=8 | 2     | PixelsTypeI:6 | 16
...
[jamoore@ome-c6100-1 dist]$ OMERO_DEV_PLUGINS=1 bin/omero metadata pixelsize --unit micrometer --x 0.11 --y 0.11 --z 0.3 Image:98401
[jamoore@ome-c6100-1 dist]$ 
```

dominikl:
That means I'd need to set the context explicitly `{'omero.group': '-1'}` ?

joshmoore:
Exactly.

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -604,20 +604,22 @@ def pixelsize(self, args):
         else:
             raise Exception("Not implemented for type %s" % md.get_type())
 
+        ctx = {'omero.group': '-1'}
+
         params = omero.sys.ParametersI()
         params.addId(md.get_id())
-        pixels = client.getSession().getQueryService()\
+        pixels = client.getSession().getQueryService(ctx)\
                                     .findAllByQuery(q, params)
```
joshmoore: I think the `ctx` needs to be passed to the `findAllByQuery`. On eel I'm getting the same behavior of no change. Likely `if not pixels:` should also call `self.ctx.die` or at least `self.ctx.err`.

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -604,20 +604,22 @@ def pixelsize(self, args):
...
 
         for pixel in pixels:
             if args.x:
-                pixel.setPhysicalSizeX(omero.model.LengthI(args.x, unit))
+                pixel.setPhysicalSizeX(omero.model.LengthI(args.x, unit), ctx)
```
joshmoore: These don't hurt, but also have no effect.

components/tools/OmeroPy/src/omero/plugins/metadata.py
dominikl: Argh, of course, thanks @joshmoore .

joshmoore:
Doh. Unfortunately, this leads to:

```
    serverExceptionClass = ome.conditions.SecurityViolation
    message = Cannot read ome.model.core.Pixels:Id_96252
```

i.e. ctx needed on `saveCollection`. All I'm doing is:

 * login as root
 * bin/omero hql -q --all "select i from Pixels i order by i.id desc" --limit=1 (i.e. choose the newest image)
 * OMERO_DEV_PLUGINS=1 bin/omero metadata pixelsize --unit micrometer --x 0.11 --y 0.11 --z 0.3 Image:98502

dominikl:
Bit lost now... When I pass the `ctx` to the save method I get:
```
No valid permissions available! DUMMY permissions are not intended for copying. Make sure that you have not passed omero.group=-1 for a save without context
```
What is that supposed to mean?

joshmoore:
The server is failing to figure how what group you intended to save to, perhaps as a result of `saveCollection` which could potentially be saving from multiple groups in one go.

mtbc:
Ha, yes, `OMERO_DEV_PLUGINS=1` does seem to help with the PR review. :smiley:

mtbc:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_metadata/TestMetadataControl/test_pixelsize/ has been passing.

components/tools/OmeroPy/src/omero/plugins/metadata.py
```diff
@@ -622,7 +621,9 @@ def pixelsize(self, args):
             if args.z:
                 pixel.setPhysicalSizeZ(omero.model.LengthI(args.z, unit))
 
-        client.getSession().getUpdateService(ctx).saveCollection(pixels)
+        groupId = pixels[0].getDetails().getGroup().getId().getValue()
+        ctx = {'omero.group': str(groupId)}
```
jburel: conn.SERVICE_OPTS.setOmeroGroup(str(groupId)) is another option

dominikl: For some reason that doesn't work. If I set `conn.SERVICE_OPTS.setOmeroGroup(str(groupId))` instead of using the `ctx`, I'll get a `ome.conditions.SecurityViolation` when running the command as non-owner of the image.
Same further up: `pixels = conn.getQueryService().findAllByQuery(q, params, ctx)` works, but when setting `conn.SERVICE_OPTS` instead of using the `ctx`, it doesn't.

components/tools/OmeroPy/src/omero/plugins/metadata.py
joshmoore: Leaving someone else to record or migrate the Python example.

joshmoore:
Merging for 5.4.6-rc2. Considering this is still a DEV plugin, we might consider a round of refactoring before going full production.

components/tools/OmeroPy/src/omero/plugins/metadata.py
mtbc: https://trello.com/c/8oJXeUPY/625-findparents-findchildren-example

PR #5697, opened 2018-03-21
Services

jburel:
# What this PR does

Review closure of services

# Testing this PR

Check that the tests remain green


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -7222,7 +7223,8 @@ def getTiles(self, zctTileList):
                 exc_info=True)
             exc = e
         try:
-            rawPixelsStore.close()
+            if rawPixelsStore is not None:
+                rawPixelsStore.close()
```
will-moore, line 7227: So, this simply handles the case that ```_prepareRawPixelsStore()``` returns ```None``` or throws an exception?

jburel: it handles the case when an exception is thrown during setPixelsId (done in ``_prepareRawPixelsStore()``)

will-moore:
Changes look fine and tests are green. Good to merge.

PR #5698, opened 2018-03-23
Import library python

will-moore:
See https://trello.com/c/3zH1PSo7/311-omeropy-importlibrary

# What this PR does

Takes code from https://gist.github.com/will-moore/f5402e451ea471fd05893f8b38a077ce (see discussion at https://www.openmicroscopy.org/community/viewtopic.php?t=8407&p=18839) and attempts to organise it like [ImportLibrary.java](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/ome/formats/importer/ImportLibrary.java) with an ```importImage``` and ```createImport``` methods.

Also removes the directory traversing from the library, instead it takes generators of file chunks and file paths.
This allows the same logic to be used with files uploaded to the web.

Tested locally with this code that creates a fileset from all images in a given folder and imports it.
```
from omero.util.import_library import ImportLibrary

folder_path = "/path/to/images/folder"

def folder_gen(folder_path):
    for f in os.listdir(folder_path):
        if f.startswith('.'):
            continue
        yield os.path.join(folder_path, f)

def chunks_gen(open_file):
    while True:
        block = open_file.read(1000 * 1000)
        if not block:
            break
        yield block

def file_gen(folder_path):
    for abspath in folder_gen(folder_path):
        with open(abspath) as f:
            yield chunks_gen(f)

file_paths = folder_gen(folder_path)
file_chunks = file_gen(folder_path)

import_library = ImportLibrary(client)
rsp = import_library.importImage(file_paths, file_chunks)
for p in rsp.pixels:
    print 'Imported Image ID:', p.image.id.val
```

# Testing this PR

1. TODO: Add tests....

Maybe this code should live somewhere other than ```utils.py```?
(forum discussion was asking how to import using BlitzGateway).

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/import_library.py
+++ b/components/tools/OmeroPy/src/omero/util/import_library.py
@@ -1,0 +1,124 @@
+
+# From http://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8407
+# Uses code from https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/testlib/__init__.py
+
+import omero
+import platform
+import os
+from omero.model import ChecksumAlgorithmI
+# from omero.model import NamedValue
+from omero.model.enums import ChecksumAlgorithmSHA1160
+from omero.rtypes import rstring, rbool
+from omero_version import omero_version
+from omero.callbacks import CmdCallbackI
+from omero.gateway import BlitzGateway
+
+
+def create_fileset(folder_path):
+    fileset = omero.model.FilesetI()
+    # for f in folder.files():
+    for f in os.listdir(folder_path):
+        if f.startswith('.'):
+            continue
+        entry = omero.model.FilesetEntryI()
+        # entry.setClientPath(rstring(str(f.abspath())))
+        entry.setClientPath(rstring(os.path.join(folder_path, f)))
+        fileset.addFilesetEntry(entry)
+
+    # Fill version info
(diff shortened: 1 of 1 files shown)


jburel:
Could we discuss that first? This looks more like a prototype

jburel:
Excluding so it is not part of the daily build prior to the release

jburel:
This is  a candidate for a topic branch alongside the insight-offline

jburel:
Closing this PR, another one has been created.

will-moore:
@jburel @joshmoore This branch was an attempt to add import functionality for Python users, without including the web import.
It seems that this would be useful for cases such as https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8449&p=19525#p19525, and [Import with Python problems](https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8520) and (older) [Scripting image uploads with cli.invoke](https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=5149&p=10283)

NB: I did a bit more on the python import in the web-import PR: #5719.

joshmoore:
No doubt, @will-moore. For me there are a few outstanding items on the linked trello card that it would be good to work through. That being said, I could definitely see this being the beginning of `omero-import` in https://gitlab.com/openmicroscopy/incubator. That would let us roll out changes to people more quickly: `pip install omero-import==0.1.0` --> 0.2.0 etc.

PR #5700, opened 2018-03-27
Fix pageCount when childCount % PAGE_COUNT = 0

will-moore:
# What this PR does

Fixes https://trello.com/c/4KHktsx7/15-bug-third-page-of-pagination-empty

# Testing this PR

1. Check that when child-count exactly fits onto a number of pages (e.g. 400 images) that we don't get an extra empty page shown.
1. Check pageCount is correct for other child counts.


pwalczysko:
failed checks here @will-moore

will-moore:
@pwalczysko Yes, I think the build system has changed a lot. I cherry-picked the same commit to #5701 so it should be good to test there (although may be excluded until after release)?

joshmoore:
Thanks, @will-moore. @pwalczysko reports being satisfied with the `page_size` workaround (see: https://github.com/IDR/deployment/pull/104) Closing this in favor for your `develop` PR. We can re-open if need be.

PR #5702, opened 2018-03-28
warn in comments about Preference.Visibility.user not working

mtbc:
Warns in comments that `Preference.Visibility.user` works just like `hidden`.

### DIFF
--- a/components/common/src/ome/system/PreferenceContext.java
+++ b/components/common/src/ome/system/PreferenceContext.java
@@ -165,7 +165,7 @@ public class PreferenceContext extends PropertyPlaceholderConfigurer {
         case admin:
             return ec.isCurrentUserAdmin();
         case user:
-            /* TODO */
+            /* TODO: is a member of the "user" group? */
         case hidden:
         default:
             return false;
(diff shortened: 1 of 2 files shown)


components/common/src/ome/system/PreferenceContext.java
```diff
@@ -164,6 +164,8 @@ public boolean canRead(EventContext ec, String key) {
             return true;
         case admin:
             return ec.isCurrentUserAdmin();
+        case user:
+            /* TODO */
```
joshmoore: I would think the appropriate logic is "is member of `user` group"

mtbc: That's a thought. I'll add it to the to-do list!

mtbc: Makes sense. Implementation looks to be a larger change than I'd initially hoped.

joshmoore:
The changes seem fine. My primary concern would be keeping in sync with https://gitlab.com/openmicroscopy/incubator/omero-common/tree/upstream

Happy to discuss strategies.

mtbc:
I'm probably happy to fit in with whatever you prefer @joshmoore but feel free to list a Chat.

PR #5704, opened 2018-03-28
remove unused configuration property versions.ome-java

mtbc:
Removes the `versions.ome-java` config property from `omero.properties`: it is an unused vestige of the ancient past.

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -922,7 +922,6 @@ product.name=OMERO.server
 
 org.bioformats=ome
 versions.bioformats=5.7.3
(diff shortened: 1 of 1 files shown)


mtbc:
Did you get a good CI run from this before excluding? If you just want to merge it then I'm happy to open a #5713 replacement. Or how should we proceed?

joshmoore:
updated https://github.com/openmicroscopy/openmicroscopy/pull/5713/files#diff-670a96f0ce2a910f3469b131c80c52c9 and included this.

PR #5705, opened 2018-03-28
fix typos in Python documentation string

mtbc:
Fixes the documentation at https://docs.openmicroscopy.org/omero/5.4.5/sysadmins/config.html#std:property-omero.web.public.url_filter.


hflynn (approved):
Looks good to me

PR #5706, opened 2018-03-28
deprecated Repository::pixels(string) as apparently unused

mtbc:
Deprecates https://downloads.openmicroscopy.org/omero/5.4.5/api/slice2html/omero/grid/Repository.html#pixels as it appears to be unused and it represents a small maintenance burden. See https://trello.com/c/q5L39XPZ/5-bioformatscache-is-expecting-read-write#comment-5ab36609b5fc3a034e14bbfd.


joshmoore:
Discussed with Mark during the read-only work. This method is currently unused and makes some unfortunate assumptions internally. Deprecating for the next release after which we can discuss its future.

PR #5707, opened 2018-03-28
Allow clearing user passwords from CLI

mtbc:
Effects https://trello.com/c/EeM6Ekrp/45-allow-setting-empty-passwords by providing a `bin/omero db password --empty`.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/db.py
+++ b/components/tools/OmeroPy/src/omero/plugins/db.py
@@ -74,7 +74,8 @@ class DatabaseControl(BaseControl):
         pw_spec = pw.add_mutually_exclusive_group()
         pw_spec.add_argument("password", nargs="?")
         pw_spec.add_argument("--empty", action="store_true",
-            help="Remove the password, allowing any for login.")
+                             help=("Remove the password, "
+                                   "allowing any for login."))
         pw.add_argument("--user-id",
                         help="User ID to salt into the password. "
                         "Defaults to '0', i.e. 'root'",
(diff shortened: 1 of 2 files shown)


joshmoore:
As a side-note, I've had to manually modify the generate SQL to `INSERT` rather than `UPDATE` in some scenarios where the password has been deleted.

dominikl:
Well, seems to work, prints out an UPDATE statement. But probably someone else has to review this PR, I don't really get it what the purpose of this `password --empty` command should be.

mtbc:
Once that SQL has been applied to the DB then the user should be able to log in with any password, kind of like the guest user.

manics:
Why would you want to allow that?

mtbc:
To undo the 5.3 workaround for 2017-SV4 that sets a password on the `guest` user. I'd also guess it could be convenient for an OMERO hidden on a private network, authentication may matter less in some situations that aren't HIPAA or whatever, especially in a read-only group.

dominikl:
The update statement works as well, password hash is set to empty string. But that has the effect, that I can't login any longer with that user, can't enter an empty password with cli (or other client), if I enter any password the password check fails. Does that need extra configuration to allow an user with empty password to log in?

mtbc:
Thanks for checking, will investigate!

mtbc:
Sorry, I had forgotten some details over the past couple of weeks since I tested but https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.5/components/server/src/ome/security/auth/PasswordUtil.java#L282 reminds me of the relevance of the `guest` user. You'll find that this empty password hash setting for a user takes effect only when running with a server for which that user is set as the guest user via `omero.roles.user.guest` (set it to their username).

dominikl:
Thanks @mtbc , yes that worked, with `omero.roles.user.guest` set, I can login with the user who's password was removed.

mtbc:
Great, thank you. You should also find that if you do set them a password then even as the designated guest user they now need it. Sorry for the interim confusion!

joshmoore:
Merging for inclusion in 5.4.6-rc2.

PR #5709, opened 2018-03-28
fix thumbnail group checks

mtbc:
It looks to me as if there are some coding bugs among the group membership checks for delivering thumbnails. This PR fixes those but it is worth then checking thumbnail behavior for any regressions.


jburel:
Does it mean that our tests did not detect that?
In that case  we need to review the permissions ones

mtbc:
I think that the effect of the bug is simply to publish rather more `ContextMessage` group changes than necessary so at worst it may have slowed performance or cluttered logs.

will-moore:
Tried a bunch of thumbnail loading and saving actions for various users in their default and non-default groups, on their own or others' images, Saving and Saving-to-all on Datasets and Plates. All seems fine.

joshmoore:
Merging for inclusion in 5.4.6-rc2

PR #5710, opened 2018-03-28
public login reset group and user

will-moore:
# What this PR does

Another potential fix for SecurityVioloation seen in https://trello.com/c/sBdAN4L6/375-public-user-workflow-crash.

To test:
 - Need to have public user enabled.
 - Login as non-public user and switch group context in web UI (this sets 'active_group' in session).
 - Allow session to timeout or kill it. E.g. if computer sleeps so that webclient doesn't ping server. Or can simply restart OMERO, or set very short session timeout (options described in trello card)
 - Now visit /webclient/, logged in as public user. Should see no error and be displayed data in public group.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/decorators.py
@@ -324,6 +324,7 @@ class login_required(object):
                 userip=get_client_ip(request))
             request.session['connector'] = connector
             # Clear any previous context so we don't try to access this
+            # NB: we also do this in WebclientLoginView.handle_logged_in()
             if 'active_group' in request.session:
                 del request.session['active_group']
             if 'user_id' in request.session:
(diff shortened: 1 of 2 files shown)


manics (approved):
Reproduced on web-dev-latest, fixed on web-dev-merge

joshmoore:
Thanks, both. Is there there are any other relevant keys that need cleaning? i.e. any future proofing? Or would a comment somewhere else remind us to keep this location in sync?

will-moore:
Added comments to help when updating the login session cleanup.

joshmoore:
Thanks for the pointers, @will-moore. Assuming there are no further comments, I'd say let's get this into the next (hopefully final) 5.4.6 build.

joshmoore:
Or 5.4.6-rc2, whichever comes first! :smile:

PR #5717, opened 2018-04-03
Restore history

chris-allan:
# What this PR does

Restore the functionality of the "History" tab in OMERO.web which was broken by openmicroscopy/openmicroscopy#5654. This PR was released as part of 5.4.4.

# Testing this PR

1. Navigate to the "History" tab in OMERO.web

2. Page should load normally, calendar should display in the left-hand panel. If the response of the XHR to `/webclient/load_calendar/` is inspected `'/webclient/load_history'` should be present correctly in the JavaScript that the template renders.

3. Visiting `/webclient/load_history/` in your web browser without the required trailing part of the URL should result in a 404.

# Related reading

1. openmicroscopy/openmicroscopy#5654

2. https://trello.com/c/HOUhWHaQ/382-bug-get-loadcalendar

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/urls.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/urls.py
@@ -88,7 +88,7 @@ urlpatterns = patterns(
     # load history
     url(r'^load_calendar/(?:(\d{4})/(\d{1,2})/)?$', views.load_calendar,
         name="load_calendar"),
-    url(r'^load_history/(\d{4})/(\d{1,2})/(\d{1,2})/$',
+    url(r'^load_history/(?:(\d{4})/(\d{1,2})/(\d{1,2})/)?$',
         views.load_history, name="load_history"),
 
     # load search
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -3056,6 +3056,9 @@ def load_calendar(request, year=None, month=None, conn=None, **kwargs):
 def load_history(request, year, month, day, conn=None, **kwargs):
     """ The data for a particular date that is loaded into the center panel """
 
+    if year is None or month is None or day is None:
+        raise Http404('Year, month, and day are required')
+
     template = "webclient/history/history_details.html"
 
     # get page


will-moore:
Working locally for me.
I'll add a robot test in another PR.

will-moore:
Added some tests in #5718. Could add more for to test ```load_history``` url but they cover the ```/load_calendar/``` url that is fixed in this PR.

Ideally the ```calendar.html``` shouldn't have this line:
```
$("div#content_details").load('{% url 'load_history' %}'+y+'/'+m+'/'+d+'/');
```
since this requires Django to generate an invalid ```/webclient/load_history/``` url that is actually 404, just so we can add ```y/m/d/``` to it in JavaScript to generate a valid url.
Then we have to handle the 404 elsewhere.

Better would be:
```
$("div#content_details").load('{% url 'webindex' %}load_history/'+y+'/'+m+'/'+d+'/');
```

chris-allan:
That has the downside of us ending up with hard coded URL endpoints in the template. The outcomes of this PR vs. the original are the same: `/webclient/load_history/` becomes a 404. Django is purely doing the lookup in the mapping of the view to URL there's no generation of a URL here.

Incidentally, I don't think there's a "right" way to do this. Template tags are not designed to be used in this way in the first place.

will-moore:
The hard-coded URL endpoints in html templates or js files happens everywhere we need to generate URLs in JavaScript, which is true in almost all of our AJAX calls.

Allowing ```/webclient/load_history/``` in the ```urls.py``` regex and then returning a 404 from ```views.py``` seems very confusing to me (I don't know that we are doing this anywhere else), hence my mistake in removing this in #5654.

chris-allan:
The number of cases where we have hard coded endpoints (`{% webindex %}/...`) is significantly lower than the use of the URL template tag to get the exact URL to the view. We have several cases where incomplete URLs will result in a 404. We even do evil stuff like pass invalid values to the URL template tag and do a search and replace afterwards. I don't like any of these semantics and the ones addressed here are hardly exceptional.

Furthermore, the contract here is established by `urls.py`. The view has just not been validating input. This is hardly a new problem for us, we have similar issues in a great deal of locations.

Fundamentally, this PR is about making the smallest number of changes to expose us to the least amount of upheaval in order to fix this regression. If you want to debate sweeping changes, which I agree are needed, we can do that afterwards.

jburel:
Tested on web-dev-merge, everything works fine. This PR fixes a regression, found and fixed by external.
We could certainly improve the general URL situation but this is outside the scope of this PR.
I have captured the discussion on https://trello.com/c/4IAGGCor/389-hard-coded-web-url

PR #5719, opened 2018-04-04
Import web

will-moore:
# What this PR does

Prototype web import (on top of #5698).
Work in progress - this PR is just somewhere to put notes...

--exclude

So far:

 - Select a Dataset in webclient...
 - Can drag and drop several folders or files onto webclient and they will be uploaded to create a new fileset, preserving all relative paths of files in sub-folders.
 - During upload, we display a progress bar.
 - Once a fileset is created we get a callback handle and then start to ping Activities to monitor import
 - During this time, the Activities panel shows a link to import log file which can be opened and refreshed during import.
 - When import completes, we can refresh Dataset to see images

screen shot 2018-04-04 at 21 08 29

Problems:

 - If a directory is uploaded (instead of files) BioFormats fails to import and we get this in the import log file:
```2018-04-04 13:27:18,404 INFO  [  o.s.b.r.ManagedImportRequestI.@3f24993] (2-thread-1) Cancelled```
 - Sometimes, when import has completed (log file is complete and image has imported), the subsequent activities ping fails to connect with OMEROweb.log:
```
2018-04-04 15:12:05,847 DEBUG [                           omero.gateway] (proc.96074) connect():2135 Error: exception ::Glacier2::PermissionDeniedException
{
    reason = Password check failed for '9915e5aa-4609-43a6-8391-9b4473411607': [user=will, created=2018-04-04 14:28:11.01, closed=2018-04-04 15:12:02.89609]
}
2018-04-04 15:12:05,847 DEBUG [                     omeroweb.decorators] (proc.96074) get_authenticated_connection():419 Connector is no longer valid, destroying...
```
which then causes webclient to log out.
 - When I log in again, the next activities request fails to get the import status of that job (still 'in progress') with ```Ice::ObjectNotExistException```.
 - There's no logic on the client to know when to create more than 1 fileset. E.g. if the user selects a bunch of TIFFs, we can only create a single fileset, instead of 1 fileset per file.

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/import_library.py
+++ b/components/tools/OmeroPy/src/omero/util/import_library.py
@@ -1,0 +1,124 @@
+
+# From http://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8407
+# Uses code from https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/testlib/__init__.py
+
+import omero
+import platform
+import os
+from omero.model import ChecksumAlgorithmI
+# from omero.model import NamedValue
+from omero.model.enums import ChecksumAlgorithmSHA1160
+from omero.rtypes import rstring, rbool
+from omero_version import omero_version
+from omero.callbacks import CmdCallbackI
+from omero.gateway import BlitzGateway
+
+
+def create_fileset(folder_path):
+    fileset = omero.model.FilesetI()
+    # for f in folder.files():
+    for f in os.listdir(folder_path):
+        if f.startswith('.'):
+            continue
+        entry = omero.model.FilesetEntryI()
+        # entry.setClientPath(rstring(str(f.abspath())))
+        entry.setClientPath(rstring(os.path.join(folder_path, f)))
+        fileset.addFilesetEntry(entry)
+
+    # Fill version info
+    system, node, release, version, machine, processor = platform.uname()
+
+    # client_version_info = [
+    #     NamedValue('omero.version', omero_version),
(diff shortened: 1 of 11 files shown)


mtbc:
On my personal systems I don't use a desktop environment so I have no way to drag files around across applications. RFE: Let me use a file chooser dialog as an alternative. (Not a blocker because my lack of desktop environment is fairly unusual these days.)

mtbc:
I seem to be having a difficult time "aiming" at datasets so sometimes my images end up orphaned. RFE: Some visual feedback of target container before import starts. Overall, though, rather nicely designed.

*Update*: Aha,

> Select a Dataset in webclient...

Yes, that's working just fine, I was mistaken in thinking that it matters to where in the tree I drag the files.

mtbc:
If I select a project then the image goes into the previously selected dataset. Perhaps that is a feature; the import message in the activities window could mention the target.

will-moore:
@mtbc Glad that it's actually working for you ;)
Certainly picking the 'last-selected' Dataset could do with improving, but I didn't have an easy idea for that. I'll add the Dataset name and link to the Activities message anyway, which should help a lot.

In general, it's not the web UI issues that are most concerning but other "Problems" listed above.

mtbc:
Mmm, I did see that PermissionDeniedException logout issue once.

mtbc:
Maybe on the server the directory is being passed to `setId`.

joshmoore:
Closing this to prevent any confusion with the community. :+1: for keeping notes but likely that would be better in https://github.com/openmicroscopy/design/. I've added it to the Tuesday agenda to discuss.

PR #5723, opened 2018-04-12
deprecate server's TinyImportFixture

mtbc:
`TinyImportFixture` appears to be an unused server class.


dominikl:
Couldn't find any usage of this class either. Good to merge 👍

PR #5724, opened 2018-04-12
move OMEROImportFixture into test directory

mtbc:
Move `OMEROImportFixture` from `blitz/src/` to `blitz/test/` given that it is a test fixture.


dominikl:
Makes sense, good to merge 👍

jburel:
This could have been used by external even if it is for test
This might be better to be included in major release
Maybe indicate the file is deprecated for this reelase

mtbc:
See #5754.

PR #5725, opened 2018-04-12
in Python parse strings in creating wrapped Boolean values

mtbc:
@manics noticed that,
```
bin/omero obj new BooleanAnnotation boolValue=false
```
creates annotations whose Boolean value is set to "true".

This PR adjusts the Python RTypes handling to provide more intuitive handling of Boolean string values.

### DIFF
--- a/components/tools/OmeroPy/src/omero/rtypes.py
+++ b/components/tools/OmeroPy/src/omero/rtypes.py
@@ -3,8 +3,6 @@
 """
 ::
     /*
-     *   $Id$
-     *
      *   Copyright 2008 Glencoe Software, Inc. All rights reserved.
      *   Use is subject to license terms supplied in LICENSE.txt
      *
@@ -171,6 +169,8 @@ def rbool(val):
     """
     if val is None or isinstance(val, omero.RBool):
         return val
+    elif isinstance(val, basestring):
+        return rbool(val.lower() == 'true')
     elif val:
         return rtrue
     else:
--- a/components/tools/OmeroPy/test/unit/test_rtypes.py
+++ b/components/tools/OmeroPy/test/unit/test_rtypes.py
@@ -2,8 +2,6 @@
 # -*- coding: utf-8 -*-
 """
 /*
- *   $Id$
- *
  *   Copyright 2008-2014 Glencoe Software, Inc. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  */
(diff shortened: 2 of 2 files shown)


manics:
`obj new BooleanAnnotation ns=openmicroscopy.org/omero/bulk_annotations boolValue=someOtherValue` results in `false`. I think it should be either `true` or an error.

mtbc:
Ah, I was going by analogy with https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#Boolean(java.lang.String).

manics:
Hmmm... Python vs Java!

components/tools/OmeroPy/src/omero/rtypes.py
```diff
@@ -171,6 +169,8 @@ def rbool(val):
     """
     if val is None or isinstance(val, omero.RBool):
         return val
+    elif isinstance(val, basestring):
```
chris-allan, line 172: I'm really torn on the impact of this specific conditional type coercion here. Having the semantics be different than `bool(val)` seems to me to be particularly dangerous and may lead to additional confusion.

At some level I would almost prefer us drawing a line in the sand with:

```
...
elif isinstance(val, basestring):
    raise TypeError('unsupported %s' % type(value))
...
```

chris-allan (changes requested):
The changes proposed have far reaching implications. At best I would not include this in a patch release.

chris-allan:
In addition to what I stated in my above review I do not feel that a suitable condition for acceptance of something like this is "it works with all our stuff". The rtypes and their public API will be some of the most used, if not **the** most used, code in our entire API.

components/tools/OmeroPy/src/omero/rtypes.py
joshmoore: Having kicked off this discussion with this code snippet:

```
          boolValue=_field_info_data(wrapper=omero.rtypes.rbool, nullable=True),
```

I'll add I didn't intended to kick off any breaking changes. I'm happy for us to discuss what `rbool` _should_ have originally done and what it should do in the future, though this might not be the place.

For me, more immediately the question I was getting towards was:

> How can the cli (specifically, [obj.py](https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/plugins/obj.py)) work with string inputs?

i.e. we need some mechanism so that `wrapper` above can know it's dealing with string inputs and then do something smarter.

```
def rbool(val, parse_trueness=False):
```

is one quick idea though obviously it needs hammering on.

mtbc:
This is not something that I can quickly solve so I created [[trello] "obj set" can't handle Boolean False](https://trello.com/c/s0h6FxJL/). Some discussion / scoping is required.

PR #5733, opened 2018-04-18
current sessions request should ignore unloaded sessions

mtbc:
In read-only mode some in-memory sessions reach the proxy cleanup filter and `buildResponse` is thus handed some unloaded sessions. This PR should not prevent sessions from being returned that would have ordinarily been from a read-write server. See https://ci.openmicroscopy.org/job/OMERO-DEV-latest-docs/ws/src/omero/_build/html/developers/Server/Clustering.html#read-only for configuration and use `bin/omero sessions who` to test.


joshmoore:
Testing on prod50-omeroreadonly-2 with this change via https://github.com/IDR/omero-prometheus-tools/blob/master/sessions/metrics.py

```
root: 2
public: 2
```

and then after a demo login:

```
demo: 1
root: 2
public: 3
```

PR #5734, opened 2018-04-18
ignore unloaded sessions from in-memory session provider

joshmoore:
> The proxy cleanup filter unloads those that the Hibernate-based security
would have filtered out anyway. The provider returns loaded sessions but
some are cleaned up before the response can be built.

*Duplicate* PR of gh-5733 against metadata54.


joshmoore:
`Trigger-IDR/48` running on east-ci.

joshmoore:
Passed. To be closed, once develop is re-merged into metadata54.

PR #5737, opened 2018-04-19
testlib script check file_name

will-moore:
See https://github.com/ome/scripts/pull/139#discussion-diff-182703948R110

This ports changes to the ```def check_file_annotation()``` from the PR above.

No testing needed (not used currently) but check code has been accurately ported.

### DIFF
--- a/components/tools/OmeroPy/src/omero/testlib/script.py
+++ b/components/tools/OmeroPy/src/omero/testlib/script.py
@@ -103,15 +103,18 @@ def check_file_annotation(client, file_annotation,
     assert id > 0
 
     conn = BlitzGateway(client_obj=client)
-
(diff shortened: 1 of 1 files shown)


components/tools/OmeroPy/src/omero/testlib/script.py
```diff
@@ -105,8 +106,12 @@ def check_file_annotation(client, file_annotation,
...
-    conn.close()
     if is_linked:
         assert links == 1
     else:
         assert links == 0
+
+    if file_name is not None:
+        name = wrapper.getFile().getName()
+        assert name == file_name
+    conn.close()
```
jburel: Will the connection be closed if the assert fails?

will-moore:
Looks good, thanks

PR #5738, opened 2018-04-19
add to Java Gateway methods to check for read-only server

mtbc:
# What this PR does

Adds Java Gateway methods `getReadOnlyStatus` and `isAnyReadOnly` for testing the server's read-only status easily.

I am not very familiar with the Gateway but I hope that this PR at least provides most of what is needed. @dominikl is very welcome to adjust how this fits into the Gateway API, e.g., via a `Facility` or somesuch.

# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.gateway/ReadOnlyTest/

Could perhaps also try those methods against `test50-omero{readonly,readwrite}` from `idr-experimental.openmicroscopy.org` or against a local server.

# Related reading

https://docs.openmicroscopy.org/omero/5.4.6-rc2/developers/Server/Clustering.html#read-only
https://trello.com/c/1deeKI1D/50-gateway-methods-to-test-read-only

### DIFF
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2015-2017 University of Dundee. All rights reserved.
+ *  Copyright (C) 2015-2018 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -18,6 +18,7 @@
  *
  *------------------------------------------------------------------------------
  */
+
 package omero.gateway;
 
 import java.beans.PropertyChangeListener;
@@ -89,6 +90,8 @@ import Glacier2.PermissionDeniedException;
 import Ice.DNSException;
 import Ice.SocketException;
 
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimaps;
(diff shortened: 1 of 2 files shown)


dominikl:
Works as expected. Makes sense to have these methods on the `Gateway` level in my opinion! Good to merge 👍

joshmoore:
In a quick conversation with @jburel & @sbesson this morning, I raised an issue that I think I had discussed with @mtbc previously which is:

> how do we go about future proofing this API if we need other states than "read-only" down the line?

@jburel said that all current read-only bugs center about whether or not `canCreate` rights are present, similar to the `canAnnotate`, `canEdit`, `canDelete` privileges that are available from `PermissionsI`. The reason I didn't originally add `canCreate` is that the other privileges are queried from an object which isn't feasible for creation.

Looking at this PR, I could see reducing the scope of the addition for 5.4.6 to:

 * `BlitzGateway::canCreate` (or very similar name) which calls
 * `omero.client::canCreate` which internally calls
 * a _private_ version of the methods added here

in order to give us time to evaluate if we want "isReadOnly" to start infiltrating client code.

The primary difficulty would be in determining the current group for additions since a different group could be set on the object itself. Likely:

```
def canCreate(self, group=None):
```

would be necessary where `None` would use the current logged in group.

mtbc:
That `canCreate` would not cover probing in detail what is possible, e.g., the server may be in a state that permits tagging (`!isAnyReadOnly('db')`) but not uploading new attachments (`isAnyReadOnly('db', 'repo')`). Certainly open to leaving this out of 5.4.6 altogether but I fear that `canCreate` could ultimately confuse more than it helps.

joshmoore:
> canCreate would not cover probing in detail what is possible

Agreed, but I'm not yet sure we want overly much probing. I'm looking for something minimal that will still prevent all known read-only related exceptions.

mtbc:
`canCreate` would be misleading though: one'll get a `false` when some things can be created. Cautious clients can use `any(conn.getReadOnlyStatus().values())`.

Discussion with @jburel suggests maybe both `any`/`all` versions of that: `isFullReadOnly` vs `isFullReadWrite` where no more than one is `true`? Should provide these? And hide the current methods?

mtbc:
Adopting the suggested `canCreate` for both PRs (with an "experimental" warning) and hiding current methods.

mtbc:
Superseded by #5749.

PR #5739, opened 2018-04-23
add to Python Gateway methods to check for read-only server

mtbc:
# What this PR does

Adds Python Gateway methods `getReadOnlyStatus` and `isAnyReadOnly` for testing the server's read-only status easily.

# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest/test_config_service/ because it is via the configuration service that the server reports its read-only status.

Could perhaps also try those methods against `test50-omero{readonly,readwrite}` from `idr-experimental.openmicroscopy.org` or against a local server.

# Related reading

https://docs.openmicroscopy.org/omero/5.4.6-rc2/developers/Server/Clustering.html#read-only
https://trello.com/c/1deeKI1D/50-gateway-methods-to-test-read-only

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2715,8 +2715,8 @@ class _BlitzGateway (object):
         key_prefix = 'omero.cluster.read_only.runtime.'
         key_regex = '^' + key_prefix.replace('.', '\.')
         properties = self.getConfigService().getConfigValues(key_regex)
-        return {key[len(key_prefix):]: value.lower() == 'true'
-                for key, value in properties.items()}
+        return dict((key[len(key_prefix):], value.lower() == 'true')
+                    for key, value in properties.items())
 
     def isAnyReadOnly(self, *subsystems):
         """
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -2702,6 +2702,35 @@ class _BlitzGateway (object):
         """
         return ProxyObjectWrapper(self, 'createExporter')
 
+    ####################
+    # Read-only status #
+
+    def getReadOnlyStatus(self):
+        """
+        Get the read-only status of the server.
+
+        :return:  a map of server subsystems each of whose value indicates if
+                  it is read-only
+        """
+        key_prefix = 'omero.cluster.read_only.runtime.'
+        key_regex = '^' + key_prefix.replace('.', '\.')
+        properties = self.getConfigService().getConfigValues(key_regex)
+        return {key[len(key_prefix):]: value.lower() == 'true'
+                for key, value in properties.items()}
+
+    def isAnyReadOnly(self, *subsystems):
+        """
+        Check specific aspects of the read-only status of the server.
+
+        :param subsystems:  the server subsystems whose status is to be checked
+        :return:            if any of the given subsystems are read-only
+        """
+        subsystems = set(subsystems)
+        for key, value in self.getReadOnlyStatus().items():
+            if value and key in subsystems:
+                return True
+        return False
+
     #############################
     # Top level object fetchers #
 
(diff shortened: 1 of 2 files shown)


mtbc:
Gah, Python 2.6 does not support dictionary comprehensions. Maybe soon we can finally drop it.

dominikl:
Tested against eel (no readonly) and local (with readonly), no problems:
```
eel:
>>> conn.getReadOnlyStatus()
{'repo': False, 'db': False}
>>> conn.isAnyReadOnly('repo')
False

against local server with readonly flags set:
>>> conn.getReadOnlyStatus()
{'repo': True, 'db': True}
>>> conn.isAnyReadOnly('repo')
True
>>> conn.isAnyReadOnly('blup')
False
```

Good to merge 👍

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -2702,6 +2702,35 @@ def createExporter(self):
...
+        properties = self.getConfigService().getConfigValues(key_regex)
+        return dict((key[len(key_prefix):], value.lower() == 'true')
+                    for key, value in properties.items())
+
+    def isAnyReadOnly(self, *subsystems):
+        """
+        Check specific aspects of the read-only status of the server.
+
+        :param subsystems:  the server subsystems whose status is to be checked
+        :return:            if any of the given subsystems are read-only
```
will-moore: Maybe this is obvious, but if it was   ```:return:            True if any of....``` it's a bit clearer that this is returning a boolean.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -2702,6 +2702,35 @@ def createExporter(self):
...
+        :return:  a map of server subsystems each of whose value indicates if
+                  it is read-only
+        """
+        key_prefix = 'omero.cluster.read_only.runtime.'
+        key_regex = '^' + key_prefix.replace('.', '\.')
+        properties = self.getConfigService().getConfigValues(key_regex)
+        return dict((key[len(key_prefix):], value.lower() == 'true')
+                    for key, value in properties.items())
+
+    def isAnyReadOnly(self, *subsystems):
```
will-moore, line 2721: I'm not familiar with use-cases here but is this a typical use case:
```
conn.isAnyReadOnly( conn.getReadOnlyStatus().keys() )
```
?
If so, it might be handy to have that be the default behaviour if I do ```conn.isAnyReadOnly()``` ?

mtbc: That feels to me to too greatly confuse the semantics: for instance, sometimes adding an argument makes it more likely to be True, sometimes less likely. As it is it tries to be the least surprising in fitting Python's `any([]) == False`. For the use case that you mention one can simply write,
```python
any(conn.getReadOnlyStatus().values())
```

will-moore: OK, fair enough.

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -2702,6 +2702,35 @@ def createExporter(self):
...
+        """
+        Check specific aspects of the read-only status of the server.
+
+        :param subsystems:  the server subsystems whose status is to be checked
+        :return:            True if any of the given subsystems are read-only
+        """
+        subsystems = set(subsystems)
```
will-moore, line 2728: Does this line have any affect on behaviour?

mtbc: I figured it future-proofs the performance of doing repeated `key in subsystems` queries if in the longer term we have many.

will-moore:
Code looks good to merge.

joshmoore:
Seealso my comment on the Java PR --https://github.com/openmicroscopy/openmicroscopy/pull/5738#issuecomment-385365538 -- which discusses a `canCreate` method as a preliminary to introducing the public method(s) added here.

mtbc:
Superseded by #5749.

PR #5740, opened 2018-04-23
Linear quantization coefficient

will-moore:
# What this PR does

See https://trello.com/c/CaURcuor/60-bugs-resetquantizationmap

It seems that the rendering engine, Blitz Gateway and iviewer allow rendering with a coefficient different from ```1.0``` with the ```linear``` family, which seems wrong. We don't show the coefficient slider in iviewer when "linear" is selected, but a previously chosen coefficient is still used. This means that switching from linear to polynomial has no effect to the rendered image.
That is now fixed in this PR.

Also, the coefficient was restricted to being less than 1.0 in the Blitz Gateway, even though the coefficient slider in iviewer has a range of 0 - 4. This PR removes the maximum limit.

# Testing this PR

1. In iviewer, choose Advanced settings in the channel colour-picker, then choose Polynomial map. Rendering will be unchanged with Gamma at 1.0.
2. Choose a higher Gamma, should see that the image is rendered differently.
3. Revert to "linear" map. Should see that you revert to the linear rendered image.

cc @pwalczysko @jburel

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8848,29 +8848,9 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         :param family:          The family (string)
         :param coefficient:     The coefficient (float)
         """
-        if channelIndex < 0 or channelIndex >= self.getSizeC():
-            return
-
-        families = self.getFamilies()
-        f = families.get("linear")
-        try:
-            f = families.get(family.lower(), f)
-        except:
-            pass
-
-        c = 1.0
-        try:
-            c = float(coefficient)
-            if c < 0:
-                c = 0
-        except:
-            pass
-
-        # Rendering engine should probably handle this...
-        if family and family.lower() == "linear":
-            c = 1.0
-
-        self._re.setQuantizationMap(channelIndex, f._obj, c, False)
+        f = omero.model.FamilyI()
+        f.value = rstring(family)
+        self._re.setQuantizationMap(channelIndex, f, coefficient, False)
 
     @assert_re()
     def setQuantizationMaps(self, maps):
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8867,7 +8867,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
             pass
 
         # Rendering engine should probably handle this...
-        if family.lower() == "linear":
+        if family and family.lower() == "linear":
             c = 1.0
 
         self._re.setQuantizationMap(channelIndex, f._obj, c, False)
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -8772,6 +8772,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
 
         return BlitzObjectWrapper(self._conn, self._re.getModel())
 
+    @assert_re()
     def setGreyscaleRenderingModel(self):
         """
         Sets the Greyscale rendering model on this image's current renderer
@@ -8780,6 +8781,7 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
         rm = self.getRenderingModels()
         self._re.setModel(self._rm.get('greyscale', rm[0])._obj)
 
+    @assert_re()
     def setColorRenderingModel(self):
         """
         Sets the HSB rendering model on this image's current renderer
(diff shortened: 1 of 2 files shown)


will-moore:
Hopefully this will get into OMERO 5.4.6. But we then need to decide whether we *need* it for iviewer release.
The enforcing of coefficient of 1.0 for linear quantization is handled by https://github.com/ome/omero-iviewer/pull/201 but not the restriction of coefficient < 1.0.

will-moore:
Tests failing at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-python27/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest.test_rdefs/TestRDefs/
fixed by last 2 commits above

NB: Lots of tests are failing, even though only 2 are broken from this PR because these failing tests don't reach the ```self.image._closeRE()``` at the end of the test. Then other tests that expect no REs open also fail.

jburel:
see https://github.com/openmicroscopy/openmicroscopy/pull/5743

jburel:
Discussed with @chris-allan:

 * Checking for negative value will happen in the rendering engine. An exception will be thrown. To be added to #5743
 * ``setQuantizationMap`` method should only invoke ``self._re.setQuantizationMap(channelIndex, f._obj, c, False)`` . The rendering engine already throws an exception if an invalid family is specified.
 * ``getFamilies`` method does not need to initialise a rendering engine. The families can be retrieved using either the Type or Query service. Usage of ``getFamilies`` in ``setQuantizationMap`` should be removed, the method is used by iviewer but that could be easily replaced by a call in iviewer to one of the service mentioned and not use ``image.getFamilies``

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8758,9 +8758,10 @@ def getRenderingModels(self):
         """
 
         if not len(self._rm):
-            for m in [BlitzObjectWrapper(self._conn, m)
-                      for m in self._re.getAvailableModels()]:
-                self._rm[m.value.lower()] = m
+            for m in self._conn.getEnumerationEntries('RenderingModel'):
```
jburel, line 8760: @assert_re() is no longer required

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8833,9 +8834,10 @@ def getFamilies(self):
         :rtype:     Dict
         """
         if not len(self._qf):
-            for f in [BlitzObjectWrapper(self._conn, f)
-                      for f in self._re.getAvailableFamilies()]:
-                self._qf[f.value.lower()] = f
+            for f in self._conn.getEnumerationEntries('Family'):
```
jburel, line 8835: @assert_re() is no longer required

jburel:
```
./components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py:444:43: E231 missing whitespace after ','
```

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8758,9 +8757,10 @@ def getRenderingModels(self):
         """
 
         if not len(self._rm):
-            for m in [BlitzObjectWrapper(self._conn, m)
-                      for m in self._re.getAvailableModels()]:
-                self._rm[m.value.lower()] = m
+            query_service = self._conn.getQueryService()
+            for m in self._conn.getEnumerationEntries('RenderingModel'):
+                mod = query_service.get('RenderingModel', m.id)
```
chris-allan: `getEnumerationEntries()` already makes a query to retrieve these models, why are we doing it again here and once for every model?

components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -8833,9 +8832,10 @@ def getFamilies(self):
         :rtype:     Dict
         """
         if not len(self._qf):
-            for f in [BlitzObjectWrapper(self._conn, f)
-                      for f in self._re.getAvailableFamilies()]:
-                self._qf[f.value.lower()] = f
+            query_service = self._conn.getQueryService()
+            for f in self._conn.getEnumerationEntries('Family'):
+                family = query_service.get('Family', f.id)
```
chris-allan: `getEnumerationEntries()` already makes a query to retrieve these families, why are we doing it again here and once for every model?

chris-allan (changes requested):

components/tools/OmeroPy/src/omero/gateway/__init__.py
will-moore: OK, yes. Sorry - I had tried a bunch of things here (including using unloaded RenderingModel, which worked for rendering but not saving) and ended up getting mixed up between RenderingModels, Enumerations and various wrapper objects!
Sorted now...

chris-allan (approved):
New changes look good.

I think the only thing that's now rather frustrating is the fact that `getFamilies()` and `getRenderingModels()` which both retrieve enumerations have different styles of return value. These are a dictionary keyed off the enumeration value which has been present since `5.4.0` and a list of enumerations which has been present since `4.1.1` respectively. We really should try to avoid this sort of thing in the future and probably should have nipped it in the bud earlier on #5501 before it was merged.

/cc @jburel, @joshmoore

jburel:
as @chris-allan mentioned, it is unfortunate that 2 similar methods to not return the same time
Option will be to deprecate and introduce a new one (similar to ``setActiveChannels`` => ``set_active_channels``)
Last commit makes sense

PR #5742, opened 2018-04-25
with read-only server do not attempt to modify file table

mtbc:
Prevents read-only servers from attempting modifications to the script repository: see https://trello.com/c/ryGZ2boi/51-getscripts-fails-with-nextval.


joshmoore (approved):
Code looks generally sensible and matches the "script deleted" case below. Merging inclusion in the final IDR 0.5.0 build.

PR #5743, opened 2018-04-26
Linear map

jburel:
# What this PR does

This PR sets the coefficient to 1 for linear mapping
Currently if the family is linear and the coefficient is not 1, the map will be polynomial

Another option could be to throw an exception if an incompatible value is specified.
This problem has been around since the beginning

# Testing this PR
Make sure the tests are green
# Related reading

See https://github.com/openmicroscopy/openmicroscopy/pull/5740
https://trello.com/c/CaURcuor/60-bugs-resetquantizationmap

cc @will-moore @chris-allan @joshmoore

### DIFF
--- a/components/rendering/src/omeis/providers/re/Renderer.java
+++ b/components/rendering/src/omeis/providers/re/Renderer.java
@@ -817,7 +817,7 @@ public class Renderer {
         qs.setQuantizationMap(family, coefficient, noiseReduction);
         ChannelBinding[] cb = getChannelBindings();
         cb[w].setFamily(family);
-        cb[w].setCoefficient(coefficient);
+        cb[w].setCoefficient(qs.getCurveCoefficient());
         cb[w].setNoiseReduction(noiseReduction);
     }
 
--- a/components/rendering/src/omeis/providers/re/quantum/QuantumStrategy.java
+++ b/components/rendering/src/omeis/providers/re/quantum/QuantumStrategy.java
@@ -312,7 +312,7 @@ public abstract class QuantumStrategy {
         this.family = family;
         String value = family.getValue();
         if (value.equals(Family.VALUE_LINEAR)) {
-            //Make sure that linear is always set to one
+            //Make sure that for linear the coefficient is always set to 1
             k = 1.0;
         }
         curveCoefficient = k;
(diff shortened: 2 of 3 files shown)


mtbc:
New tests passing at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration/RenderingEngineTest/.

jburel:
To match the logic ofhttps://github.com/openmicroscopy/openmicroscopy/blob/develop/components/rendering/src/omeis/providers/re/quantum/QuantumStrategy.java#L132
we should probably throw an exception if the coefficient is not compatible with the family but that will be a potential breaking change
cc @sbesson

mtbc:
One could argue that the current lack of exception is a bug.

chris-allan:
The use of the coefficients has up until very recently been incredibly limited. I feel like we can set whatever set of semantics we want now. However, we probably only get to do this once and changing them, irrespective of correctness, will be incredibly painful.

Let's make absolutely sure we are all okay with these semantics, and those we will add in #5740, before we release anything. Doing otherwise will be very costly for ourselves and anyone consuming the API.

**Edit:** I would also suggest that we define and enforce whatever semantics we want at as low a level as possible. I'm especially concerned about doing bounds checking or condition validation in `BlitzGateway`.

/cc @will-moore

jburel:
@chris-allan yes we must sure of what we want.

jburel:
Discussed with @chris-allan:
 * Add check for negative coefficient. Throw an exception in that case
 * If linear and coefficient is not 1. Default to 1 i.e. keep the current implementation

sbesson:
Sounds good to me. In terms of model/specification, that's effectively only enforcing the positive nature of the coefficient in which case throwing an exception on negative values is a reasonable behavior. The handling of any nonsensical but syntactically valid coefficient is then the responsibility of the implementation. How is zero handled?

jburel:
I should have said negative or 0, since passing 0 will imply multiplying by a constant. I don't think it will add value

will-moore:
Working for me, throwing exceptions for negative coefficients and ignoring coefficients for "linear" family.

PR #5745, opened 2018-04-30
clarify the action of fs sets --check in help text

mtbc:
Makes the help text for `bin/omero fs sets --check` more specific. See discussion on [[trello] fs sets timeout](https://trello.com/c/cuwJ9NNz/53-fs-sets-timeout).


rleigh-codelibre:
The help text change reads more informatively, and looks good to merge.

PR #5746, opened 2018-04-30
OMERO.web centre panel cleanup

chris-allan:
# What this PR does

The current "Thumbnails" centre panel plugin assumes that it is always active. This PR attempts to minimise background event, and data loading churn when another plugin is active.

# Testing this PR

1. Conditions where there is, and is not, an additional active centre panel plugin should be available

2. All actions where `update_thumbnails_panel()` is called should be tested:
    * Tree cut/copy and paste
    * Tree node move; Image into another Dataset for instance
    * Tree node selection
    * Tree refresh
    * Change group
    * Delete

3. No change in behaviour should be observed

# Notes

`update_thumbnails_panel()` is a window wide function. As such it has effectively become part of the public API of OMERO.web and is deeply coupled to the "Thumbnails" plugin. Many thumbnail refresh and update actions call this function explicitly as outlined in the testing criteria above.

If a third-party centre panel plugin wishes to react to any of these conditions it must find a way to intercept or mimic the behaviour of this function. It also must ensure that anything that it does to facilitate this does not completely break when switching between plugins. The purpose of this PR is definitely not to refactor all `update_thumbnails_panel()` use to an event based model but it is quite sorely needed.

/cc @emilroz, @jburel, @joshmoore

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
@@ -89,10 +89,6 @@ OME.handle_tree_selection = function(data, event) {
     // and update_thumbnails_panel, just run them instead
 
     // Check the functions exist, they might not if the central panel has not been loaded
-    if (typeof(syncThumbSelection) === "function") {
-        // safe to use the function
-        syncThumbSelection(data, event);
-    }
     if (typeof(update_thumbnails_panel) === "function") {
         // safe to use the function
         update_thumbnails_panel(event, data);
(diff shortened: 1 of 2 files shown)


will-moore:
This is working fine for me, with various jsTree manipulations etc.
However, I noticed that it's possible to get the jsTree out of sync with ```parade``` by e.g. dragging and dropping images between Datasets in the tree. These are resolved as soon as there's a selection event to re-render parade, but if we want better sync then we need to listen for more events.
(Even webclient centre panel can still get out of sync, e.g. if you drag an image *into* a Dataset that's already selected. It listens for more events but the update logic is not as fool-proof as React)

chris-allan:
That is something that I noticed but deliberately did not attempt to handle as part of ome/omero-parade#20. Especially after I found out about the way events were being dealt with by the current Thumbnails plugin which was the impetus to begin this PR.

I'll definitely revisit it once I have a better handle on all the things that the Thumbnails plugin is doing and attempt to limit conflicting interactions.

PR #5749, opened 2018-05-03
for read-only add canCreate to gateways

mtbc:
Adds ternary `canCreate` to gateways for easy checking of the server's read-only status. Supersedes #5738 and #5739. See https://docs.openmicroscopy.org/omero/5.4.6-rc2/developers/Server/Clustering.html#read-only and https://trello.com/c/1deeKI1D/50-gateway-methods-to-test-read-only.


jburel:
Merging this PR now that we have examples of client using the method
#5753 and #5752 were built on top of this PR

PR #5750, opened 2018-05-03
fix logger class for raw file bean

mtbc:
Has the `RawFileBean` log as itself instead of pretending to be the `RawPixelsBean`.


joshmoore:
Oh, wow. Thanks, @mtbc

PR #5751, opened 2018-05-03
have read-only server's RawFileBean.close() be read-only

mtbc:
Allow raw file stores to be closed on a read-only server. See https://trello.com/c/wlML9Pdg/14-need-rawfilebeanreadonly.

### DIFF
--- a/components/server/resources/ome/services/service-ome.api.RawFileStore.xml
+++ b/components/server/resources/ome/services/service-ome.api.RawFileStore.xml
@@ -3,9 +3,7 @@
 <!--
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 # 
-# $Id$
-# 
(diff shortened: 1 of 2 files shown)


mtbc:
I think this helps OMERO.figure somehow @will-moore?

will-moore:
You couldn't previously close a RawFileStore on read-only server? If not then I think this would help, but wasn't aware of any problem.

joshmoore:
Added the full stacktrace to the linked card. Seems `file_ann.getFileInChunks()` was enough to trigger the email we saw.

will-moore:
This PR should also fix the download of images as zip files in web, which is failing with 
```
ome.conditions.ApiUsageException message = This instance is read-only
getFileInChunks store.close()
```
cc @jburel

jburel:
With this change it is possible to import:

 * single image e.g. tiff
 * image with companion file e.g. dv. In that case a zip is created

mtbc:
How / where is the zip created -- is this something that should be prevented?

jburel:
It is created in client.

PR #5752, opened 2018-05-03
Read only insight

jburel:
# What this PR does

Disable some features when the server is configured as read-only

Build on top of https://github.com/openmicroscopy/openmicroscopy/pull/5749
So we can test the feature exposed to the client

# Testing this PR
Check that 
* importer
* running script
* download
* creation of project/dataset/screen from top
* saving of settings
are disabled

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
@@ -606,7 +606,6 @@ public class DataServicesFactory
             msg.println(entry.getKey()+": "+entry.getValue());
         }
         registry.getLogger().info(this, msg);
-
         registry.bind(LookupNames.CURRENT_USER_DETAILS, exp);
         registry.bind(LookupNames.IMAGE_QUALITY_LEVEL, 
         		determineImageQuality(uc.getSpeedLevel()));
@@ -634,10 +633,12 @@ public class DataServicesFactory
         Set<GroupData> available;
         List<ExperimenterData> exps = new ArrayList<ExperimenterData>();
         String ldap = null;
+        long gid = exp.getDefaultGroup().getId();
+        SecurityContext ctx = new SecurityContext(gid);
+        boolean canCreate = omeroGateway.canCreate(ctx);
         try {
             GroupData defaultGroup = null;
-            long gid = exp.getDefaultGroup().getId();
-        	SecurityContext ctx = new SecurityContext(gid);
+        	registry.bind(LookupNames.CAN_CREATE, canCreate);
         	groups = omeroGateway.getAvailableGroups(ctx, exp);
         	registry.bind(LookupNames.SYSTEM_ROLES,
                     omeroGateway.getSystemRoles(ctx));
@@ -727,6 +728,7 @@ public class DataServicesFactory
 			agentInfo = (AgentInfo) kk.next();
 			if (agentInfo.isActive()) {
 				reg = agentInfo.getRegistry();
+				reg.bind(LookupNames.CAN_CREATE, canCreate);
 				reg.bind(LookupNames.CURRENT_USER_DETAILS, exp);
 				reg.bind(LookupNames.USER_GROUP_DETAILS, available);
 				reg.bind(LookupNames.USERS_DETAILS, exps);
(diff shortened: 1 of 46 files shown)


components/insight/SRC/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
```diff
@@ -634,10 +633,12 @@ public void connect(UserCredentials uc)
         Set<GroupData> available;
         List<ExperimenterData> exps = new ArrayList<ExperimenterData>();
         String ldap = null;
+        long gid = exp.getDefaultGroup().getId();
+        SecurityContext ctx = new SecurityContext(gid);
+        boolean canCreate = omeroGateway.canCreate(ctx);
```
mtbc, line 638: Do you mean `Boolean.TRUE.equals(omeroGateway.canCreate(ctx))` to guard against nulls?

mtbc: Ah, sorry, maybe not, probably getting my gateways mixed up!

jburel: ``Null`` is already handled in the internal gateway so we have to deal in the rest of the code with  either ``true`` or ``false``

joshmoore: @jburel : if we start to take more conditions into account, how will this change in the future? Will you break the API?

jburel: This internal gateway is now in many cases a wrapper around the Java gateway under blitz.
If we introduce more conditions, I think we will be adding entry in the lookup. It won't break the API.

pwalczysko:
missing bird's eye view on user-6 ID 21450


Edit: The same image works fine in iviewer. So I cannot guarantee that this is read-only problem, but I can guarantee that this is an insight specific problem

pwalczysko:
Insight from https://ci.openmicroscopy.org/job/OMERO-DEV-merge-build/lastSuccessfulBuild/ICE=3.6,jdk=8_LATEST,label=octopus/ was used

pwalczysko:
The menu of scripts from RHP is not disabled.

```
java.lang.Exception: java.lang.Exception: org.openmicroscopy.shoola.env.data.ProcessException: Cannot run script with ID:140162
	at org.openmicroscopy.shoola.env.data.views.ProcessBatchCall.doStep(ProcessBatchCall.java:101)
	at ...


```

screen shot 2018-05-04 at 12 17 13

pwalczysko:
Create Report has not been disabled and is working. Expected ?

pwalczysko:
Save thumbnails has not been disabled and is apparently working. Expected ?
screen shot 2018-05-04 at 12 21 13

pwalczysko:
4 problems

   - https://github.com/openmicroscopy/openmicroscopy/pull/5752#issuecomment-386569644 Fixed see 
be7e8e1
   - https://github.com/openmicroscopy/openmicroscopy/pull/5752#issuecomment-386571288 Fixed see 6773631
   - https://github.com/openmicroscopy/openmicroscopy/pull/5752#issuecomment-386571776 Expected see below
   - https://github.com/openmicroscopy/openmicroscopy/pull/5752#issuecomment-386571906 Expected see below

jburel:
``* Create Report has not been disabled and is working. Expected ?``
Yes this is done client side.
``* Save thumbnails has not been disabled and is apparently working. Expected ?``
Yes  this is done client side.

pwalczysko:
Bug: The Delete key on keyboard still works, opens dialogs confirming the delete, then the activities spinner spins endlessly on read-only server (eel merge)

pwalczysko:
Bug: Drag and drop on read-only in the tree (trying to D&D a Dataset into another Project) pops up a dialog (note, I am the owner of all the objects I am manipulating here)
screen shot 2018-05-04 at 13 58 46

jburel:
Interestingly for the problem reported https://github.com/openmicroscopy/openmicroscopy/pull/5752#issuecomment-386569644
the thumbnail shows up then disappears
I double-checked in web and it is correctly visible in Web i.e. this is **not** a server problem cc @mtbc which is good.

jburel:
@pwalczysko the problems reported should now all be fixed

dominikl:
Looks good. Only additional thing I could think of is to add an error message to the OMERO.importer client, which says that you can't import on an read-only server directly after the log in and then exit. If you try now, the import fails, but doesn't say why (can't even view the exception for some reason). But can be done as separate PR.

jburel:
@dominikl good point I did not check the importer standalone. I will do that in a follow up PR

jburel:
#5749 has been merged

jburel:
Thanks @dominikl and @pwalczysko for review.
Merging

PR #5753, opened 2018-05-03
Read only can create web

will-moore:
# What this PR does

NB: This is on top of https://github.com/openmicroscopy/openmicroscopy/pull/5749 since it uses ```conn.canCreate()``` added there.

This uses ```conn.canCreate()``` to populate the ```request.session``` on login, so it doesn't have to be called repeatedly.
Then we pass this into the rendering context and use it to set a global JavaScript variable ```WEBCLIENT.CAN_CREATE``` which is used to disable creating in jsTree and opening of scripts menu.

We also disable OME-TIFF export.

# Testing this PR

1. All create controls should be disabled in jsTree
1. Script menu won't launch
1. OME-TIFF export option is disabled.

cc @mtbc

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/decorators.py
@@ -247,7 +247,12 @@ class login_required(object):
         return False
 
     def load_server_settings(self, conn, request):
-        """Loads Client preferences from the server."""
+        """Loads Client preferences and Read-Only status from the server."""
+        try:
+            request.session['can_create']
+        except KeyError:
+            request.session.modified = True
+            request.session['can_create'] = conn.canCreate()
         try:
             request.session['server_settings']
         except:
--- a/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/experimenter_form.html
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/templates/webadmin/experimenter_form.html
@@ -310,7 +310,8 @@
         {% else %}
 
         <label>Password</label>
-            <a class="btn btn_text silver" id="change_password" {% if not can_modify_user %} style="pointer-events: none" {% endif %}>
+            <a class="btn btn_text silver" id="change_password"
+                {% if not can_modify_user or not ome.can_create %} style="pointer-events: none" {% endif %}>
                 <span {% if not can_modify_user %} style="color:Gray" {% endif %}>Change User's Password</span>
             </a>
         <span id="password_change_message"></span>
@@ -351,7 +352,7 @@
         <br />
         
         <input type="submit" 
-            {% if not can_modify_user %}
+            {% if not can_modify_user or not ome.can_create %}
                 disabled title="You don't have permissions to edit Users"
             {% endif %}
             value="{% trans 'Save' %}" />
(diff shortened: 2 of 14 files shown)


manics:
An issue noticed today on web-dev-merge configured as read-only:
screen shot 2018-05-04 at 09 44 09

The `Publishing Options` menu is still available. Clicking `Split View Figure` or `Thumbnail Figure` brings up a functional dialog, but clicking `Create Figure` fails.

Expected behaviour: Ideally a figure should be downloaded directly to the client. Otherwise disable the option.

jburel:
The option should be disabled since this is a script

manics:
The `Preview` `Save to all` icon is enabled even though it doesn't work, whereas the `Save` icon is disabled as expected.
screen shot 2018-05-04 at 09 53 31

pwalczysko:
``Move to Group`` is enabled in context menu. This is disabled in Insight (I think Insight is right there).
screen shot 2018-05-04 at 12 29 22

pwalczysko:
All looks good as described in the header. 


But https://github.com/openmicroscopy/openmicroscopy/pull/5753#issuecomment-386573329 is still there, not sure if this is planned for another PR.

Also, Figure has still problems, are these to be reported here ?


Edit:
Created https://trello.com/c/QB8XPKtA/188-bug-crash-on-export-read-only#comment-5aec4992af0e43611d5c5818 and https://trello.com/c/hiRysX5K/187-bug-read-only-save-button for the Figure problems.

jburel:
@pwalczysko figure is not covered

will-moore:
Re https://github.com/openmicroscopy/openmicroscopy/pull/5753#issuecomment-386573329
I think this needs a server-side fix since we use ```canChgrp``` permissions. @mtbc will look into it...

jburel:
Editing of users/groups should be disabled too
The easiest will be to disable the Admin button the toolbar

jburel:
tested against pub-omero (merge build deployed) (not read-only). Features are back

jburel:
* OME-TIFF: disabled. Expected
* Publishing script: not available. Expected
* Save to all: disabled. Expected

jburel:
Clicking on user settings leads to a crash
```
 File "/opt/hudson/workspace/WEB-DEV-merge-deploy/OMERO.py/lib/python/omero_api_IRepositoryInfo_ice.py", line 204, in getFreeSpaceInKilobytes
    return _M_omero.api.IRepositoryInfo._op_getFreeSpaceInKilobytes.invoke(self, ((), _ctx))

{
    serverStackTrace = ome.conditions.ApiUsageException: This instance is read-only
	at ome.security.basic.EventHandler.invoke(EventHandler.java:136)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	at 
```

jburel:
Shares:
* The ``Add Comment `` button is enabled when clicking on a share. Entering comment leads to an error as excepted. This should be disabled

jburel:
#5749 has been merged

jburel:
* Share comments not available. Expected
* No longer possible to create/edit users/group. Expected
* Crash when clicking on Statistics. This should be fixed by #5757

PR #5755, opened 2018-05-04
do not set chgrp, chown permissions if read-only database

mtbc:
# What this PR does

Prevents `canChgrp`, `canChown` from being viewed as `true` from clients if `omero.cluster.read_only.db=true`.

# Related reading

#5753


jburel:
``Move to Group`` is now correctly disabled in Web

PR #5758, opened 2018-05-07
Exit standalone importer when connected to read-only server

dominikl:
# What this PR does

Popup an error message and exit standalone importer when connected to a read-only server.

# Testing this PR

- Connect to read-only server, check that the error message pops up and the application exits.
- Connect to a non read-only server, check that the importer starts up as usual.

/cc @jburel


jburel:
tested against
 * eel (read-only). Message pops up and the application exits
 * pub-omero (not read-only). Connect correctly and import

PR #5759, opened 2018-05-07
Check PUBLIC_ENABLED before redirect

will-moore:
# What this PR does

Fix error from https://www.openmicroscopy.org/qa2/qa/feedback/21070/
```
File "/opt/omero/OMERO.server-5.4.3-ice36-b77/lib/python/omeroweb/webclient/views.py", line 383, in _load_template
if settings.PUBLIC_USER == conn.getUser().getOmeName():
File "/usr/lib/python2.7/site-packages/django/conf/__init__.py", line 49, in __getattr__
return getattr(self._wrapped, name)
AttributeError: 'Settings' object has no attribute 'PUBLIC_USER'
```

# Testing this PR

1. Test on a server where PUBLIC_USER is not set
2. Login to webclient
3. Go to /webclient/?show=dataset-ID where ID doesn't exist (enter a random number)
4. Should simply go to webclient page without error.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -381,7 +381,8 @@ def _load_template(request, menu, conn=None, url=None, **kwargs):
     # If we failed to find 'show'...
     if request.GET.get('show', None) is not None and first_sel is None:
         # and we're logged in as PUBLIC user...
-        if settings.PUBLIC_USER == conn.getUser().getOmeName():
+        if (settings.PUBLIC_ENABLED and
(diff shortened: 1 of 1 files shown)


mtbc:
Without this PR reproduced the,
> 'Settings' object has no attribute 'PUBLIC_USER'

mtbc:
This PR does fix the problem.

mtbc:
It's desired that when public user is enabled a link to a non-existent dataset takes me to the login page?

will-moore:
@mtbc Yes, because the assumption is that this Dataset does exist and that a registered user can access it via logging in, but the public user can't access it.

mtbc (approved):

PR #5760, opened 2018-05-08
5.4.6 text update

jburel:
Update text in history

### DIFF
--- a/history.txt
+++ b/history.txt
@@ -8,8 +8,13 @@ OMERO version history
 5.4.6 (May 2018)
 ----------------
 
-This introduces a significant new subsystem for read-only operation with which servers can be configured not to make changes to the database, the filesystem, or both. The goal is to permit horizontal scaling of OMERO by running multiple servers in parallel to increase the throughput of data and metadata for large-scale analysis or publishing.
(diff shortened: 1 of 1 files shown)


mtbc:
Could it be wrapped to a reasonable column width?

history.txt
```diff
@@ -8,10 +8,16 @@ OMERO version history
...
-the filesystem, or both. The goal is to permit the horizontal scaling
-of OMERO installations.
+This introduces a significant new subsystem for read-only operation with which
+servers can be configured not to make changes to the database, the filesystem, or both. 
+The goal is to permit horizontal scaling of OMERO by running multiple servers in parallel
+to increase the throughput of data and metadata for large-scale analysis or publishing.
+Additionally, a read-only copy of an existing OMERO can be opened safely to the public
+for experimentation. For example, this infrastructure supports the public OMERO web
+and the Jupyter environment of the 
+`Image Data Resource <https://idr.openmicroscopy.org/>`_.
```
mtbc, line 18: probably drop the markup

mtbc: ah, looks like `history.txt` has others like this, ugh

mtbc (approved):
Fine for this release, maybe someday we can revisit this habit of putting Sphinx markup in this file!

PR #5761, opened 2018-05-08
Shapedata bugfix

dominikl:
# What this PR does

Allows to remove the C, Z and T of a Shape. Previously when a negative value was passed to `ShapeData.setC()` (etc.) the value was set to `0`.  With this PR passing a negative value will set the underlying Ice Shape object's C, Z, T to `null`.  And for the other way round, if the Shape object's C, Z, T is `null`, `ShapeData.getC()` (etc.) will return `-1`, indicating that the value has not been set, so that getters and setters now also behave in a consistent way.

Had to modify the `ROIData` pojo, too, e.g. remove methods like `firstPlane()` (which isn't used anywhere), because these wouldn't work properly with `null` CZT ROIs (which are supposed to be present on all planes).

Also updated the integration test accordingly.

Another bug was noticed: The various setters (`setX()`, `setY()`, etc.) on the ShapeData subclasses (`RectangleData`, etc.) didn't set the `dirty` flag, that's why the ROIFacility didn't save them back to the server when `saveROIs` was called. Fixed the setters and added an unit test for all setters of all ShapeData classes.

# Testing this PR

- Check ROIFacility integration test.
- Check ShapeData unit test.
- Check that ROI tool in Insight still works as before.

# Related reading

http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2018-May/004199.html

/cc @awalter17

### DIFF
--- a/components/blitz/src/omero/gateway/model/ROIData.java
+++ b/components/blitz/src/omero/gateway/model/ROIData.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- * Copyright (C) 2006-2016 University of Dundee. All rights reserved.
+ * Copyright (C) 2006-2018 University of Dundee. All rights reserved.
  *
  *
  * This program is free software; you can redistribute it and/or modify
@@ -247,7 +247,9 @@ public class ROIData
      */
     public List<ShapeData> getShapes(int z, int t)
     {
-        return roiShapes.get(new ROICoordinate(z, t));
+        List<ShapeData> res = roiShapes.get(new ROICoordinate(z, t));
+        res.addAll(roiShapes.get(new ROICoordinate(-1, -1)));
+        return res;
     }
 
     /**
@@ -259,39 +261,6 @@ public class ROIData
     {
         return roiShapes.values().iterator();
     }
-
-    /** 
-     * Return the first plane that the ROI starts on.
-     *
-     * @return See above.
-     */
-    public ROICoordinate firstPlane()
-    {
-        return roiShapes.firstKey();
-    }
-
-    /** 
-     * Returns the last plane that the ROI ends on.
-     *
-     * @return See above.
-     */
-    public ROICoordinate lastPlane()
-    {
-        return roiShapes.lastKey();
-    }
-
-    /**
-     * Returns an iterator of the Shapes in the ROI in the range [start, end].
-     *
-     * @param start The starting plane where the Shapes should reside.
-     * @param end The final plane where the Shapes should reside.
-     * @return See above.
-     */
-    public Iterator<List<ShapeData>> getShapesInRange(ROICoordinate start,
-            ROICoordinate end)
-            {
-        return roiShapes.subMap(start, end).values().iterator();
-            }
 
     /**
      * Returns <code>true</code> if the object a client-side object,
--- a/components/blitz/src/omero/gateway/model/ROIData.java
+++ b/components/blitz/src/omero/gateway/model/ROIData.java
@@ -209,9 +209,11 @@ public class ROIData
         ROICoordinate coord = shape.getROICoordinate();
         List<ShapeData> shapeList;
         shapeList = roiShapes.get(coord);
-        shapeList.remove(shape);
-        roi.removeShape((Shape) shape.asIObject());
-        setDirty(true);
+        if (shapeList != null) {
+            shapeList.remove(shape);
+            roi.removeShape((Shape) shape.asIObject());
+            setDirty(true);
+        }
     }
 
     /**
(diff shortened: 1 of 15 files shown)


dominikl:
@awalter17 spotted another ROI related bug, will fix it within this PR, too.
( http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2018-May/004214.html )

mtbc:
> remove ... which isn't used anywhere

How do we know? Should we deprecate here and remove after 5.4?

*Update:* Ha, actually that's what you did, :+1:

components/blitz/src/omero/gateway/model/ROIData.java
```diff
@@ -247,7 +250,9 @@ public int getShapeCount()
      */
     public List<ShapeData> getShapes(int z, int t)
     {
-        return roiShapes.get(new ROICoordinate(z, t));
+        List<ShapeData> res = roiShapes.get(new ROICoordinate(z, t));
+        res.addAll(roiShapes.get(new ROICoordinate(-1, -1)));
+        return res;
```
mtbc: This also covers if only one of *z* or *t* is -1?

sbesson:
It seems that with this PR included, several UpdateService and RoiService integration tests underwent some regression - see https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/892/

dominikl:
Thanks @sbesson ! Strange that I didn't spot that earlier.

components/blitz/src/omero/gateway/model/ROIData.java
```diff
@@ -260,38 +271,46 @@ public int getShapeCount()
...
      */
     public Iterator<List<ShapeData>> getShapesInRange(ROICoordinate start,
-            ROICoordinate end)
-            {
-        return roiShapes.subMap(start, end).values().iterator();
-            }
+            ROICoordinate end) {
+        List<List<ShapeData>> res = new ArrayList<List<ShapeData>>();
+        res.addAll(roiShapes.subMap(start, end).values());
+        res.add(roiShapes.get(new ROICoordinate(-1, -1)));
```
mtbc: Can this end up adding a `null`? (If so, that's okay?)

dominikl: Yes, thanks, that needs a null check, too.

dominikl:
Integration tests are green, tested the Insight ROITool again, I think the PR is ready to merge.

components/blitz/src/omero/gateway/model/ROIData.java
```diff
@@ -247,7 +252,13 @@ public int getShapeCount()
...
     public List<ShapeData> getShapes(int z, int t)
     {
-        return roiShapes.get(new ROICoordinate(z, t));
+        List<ShapeData> res = roiShapes.get(new ROICoordinate(z, t));
+        if (res == null)
+            res = new ArrayList<ShapeData>();
+        List<ShapeData> allZT = roiShapes.get(new ROICoordinate(-1, -1));
+        if (allZT != null)
+            res.addAll(allZT);
+        return res;
```
mtbc: What about if only one of *z* or *t* is -1?

components/tools/OmeroJava/test/integration/gateway/ROIFacilityTest.java
```diff
@@ -128,18 +142,9 @@ public void testLoadROIs() throws DSOutOfServiceException,
...
-                if (r2.getId() == r.getId())
-                    it.remove();
-            }
-        }
-
-        Assert.assertTrue(myRois.isEmpty());
+        compare(myRois, rois);
     }
-
+
```
mtbc: avoid trailing space

components/blitz/src/omero/gateway/model/ROIData.java
dominikl: Argh, that's true, these cases have to be considered, too.

dominikl:
Fixed issue https://github.com/openmicroscopy/openmicroscopy/pull/5761#discussion_r200272329 and also deprecated `getShapesInRange` method, can't imagine anyone uses that anyway.

components/blitz/src/omero/gateway/model/ROIData.java
```diff
@@ -209,9 +209,11 @@ public void removeShapeData(ShapeData shape)
         ROICoordinate coord = shape.getROICoordinate();
         List<ShapeData> shapeList;
         shapeList = roiShapes.get(coord);
```
rgozim: `List<ShapeData> shapeList = roiShapes.get(coord);` in one line would look better

rgozim:
Looks okay to me

components/blitz/src/omero/gateway/model/ROIData.java
joshmoore: :+1: Longer-term for the API might be nice to say something like: `new ROICoordinate(new Optional(z), new Optional(t))` or `new LenientROICoordinate(z, t)` or `new ROICoordinate(z, t, lenient=true)`

joshmoore:
Merging for inclusion in 5.4.7

PR #5765, opened 2018-05-10
address all-groups-context issue for gateway queries

mtbc:
Sets the Java Gateway to object when `SecurityContext(-1)` is attempted when it is useless, so as to make people fix clients accordingly. The test captures some possible usage patterns surrounding `SecurityContext(-1)`: see the discussion on this PR and at https://trello.com/c/9W8F4Kmq/412-clarify-behavior-of-all-groups-context. It should be passing at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.gateway/GatewayUsageTest/testGroupContextForService/.

### DIFF
--- a/components/blitz/src/omero/gateway/Gateway.java
+++ b/components/blitz/src/omero/gateway/Gateway.java
@@ -1649,11 +1649,11 @@ public class Gateway implements AutoCloseable {
                 prx = client.createSession(login.getUser().getUsername(), login
                         .getUser().getPassword());
             }
-
-            if (ctx.getGroupID() >= 0)
-                prx.setSecurityContext(new ExperimenterGroupI(ctx.getGroupID(),
-                        false));
-            
+            if (ctx.getGroupID() >= 0) {
+                prx.setSecurityContext(new ExperimenterGroupI(ctx.getGroupID(), false));
+            } else {
+                throw new IllegalArgumentException("must set security context with a valid group ID");
+            }
             c = new Connector(ctx, client, prx, login.isEncryption(), log);
             for (PropertyChangeListener l : this.pcs
                     .getPropertyChangeListeners())
--- a/components/tools/OmeroJava/test/integration/gateway/GatewayUsageTest.java
+++ b/components/tools/OmeroJava/test/integration/gateway/GatewayUsageTest.java
@@ -25,6 +25,7 @@ import java.beans.PropertyChangeListener;
 import java.util.UUID;
 
 import integration.AbstractServerTest;
+import omero.SecurityViolation;
 import omero.gateway.Gateway;
 import omero.gateway.JoinSessionCredentials;
 import omero.gateway.LoginCredentials;
@@ -37,10 +38,13 @@ import omero.gateway.model.DatasetData;
 import omero.gateway.model.ExperimenterData;
 import omero.gateway.model.GroupData;
 import omero.log.SimpleLogger;
+import omero.model.LongAnnotation;
+import omero.model.LongAnnotationI;
 
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+import com.google.common.collect.ImmutableMap;
 
 /**
  * Tests the login options supported by gateway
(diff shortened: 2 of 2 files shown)


dominikl:
With respect to the Gateway: If you set a `SecurityContext` group Id >= 0 it just passes that on to 
the `ServiceFactory` If the id < 0 it doesn't set anything. https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/Gateway.java#L1653 
Don't know what happens under the hood of the `ServiceFactory`.

Maybe the `QueryService` ignores the "non-group" context? I think I came across a similar problem in the past, the `QueryService` always needed the `omero.group` set to `-1` explicitely.

mtbc:
Thank you very much, I'd missed that. A couple of questions then:

1. @dominikl: Would it be okay if I adjusted the gateway to return an `IllegalArgumentException` in this case rather than silently ignoring the given security context? Or perhaps I missed a note that the -1 is code for "stay in default group" and the current behavior is needed?

1. @joshmoore: It makes sense to you that `SessionManagerImpl.setGroupSecurityContext` can't cope with all-groups?

joshmoore:
Setting to `-1` is equivalent to saying "all groups". The Python gateway makes extensive use of this.

mtbc:
@joshmoore:
```
2018-05-10 11:05:22,703 INFO  [        ome.services.util.ServiceHandler] (l.Server-6)  Executor.doWork -- ome.services.sessions.SessionManagerImpl.setGroupSecurityContext[-1]
2018-05-10 11:05:22,703 INFO  [        ome.services.util.ServiceHandler] (l.Server-6)  Args:	[null, InternalSF@879319843]
2018-05-10 11:05:22,705 INFO  [         ome.security.basic.EventHandler] (l.Server-6)  Auth:	user=0,group=0,event=null(User),sess=9be7fe84-d149-4dfa-939b-2ee5287b0eb0
2018-05-10 11:05:22,706 INFO  [                 org.perf4j.TimingLogger] (l.Server-6) start[1525946722703] time[3] tag[omero.call.exception]
2018-05-10 11:05:22,707 INFO  [        ome.services.util.ServiceHandler] (l.Server-6)  Excp:	ome.conditions.ApiUsageException: No such group: -1
```
caused by the,
```java
s.getDetails().setGroup(sf.getAdminService().getGroup(id));
```

dominikl:
But in the Python it was the same, for queries had to set the 'omero.group' explicitely to '-1', e. g. 
https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/plugins/metadata.py#L607

joshmoore:
I'm thinking of the likes of:

```
plugins/fs.py:        gateway.SERVICE_OPTS.setOmeroGroup("-1")
```

mtbc:
Ah, am I effectively asking for a `SERVICE_OPTS` equivalent for the Java Gateway?

joshmoore:
Probably.

dominikl:
Will have a look into that, once I figured out what this `SERVICE_OPTS` actually is...

dominikl:
I still don't get it. So if the (Java) SecurityContext is < 0, the SecurityContext of the ServiceFactory is not set ( https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/Gateway.java#L1653 ) What implication does that have? It didn't cause any issues so far. Does it pick up the right context anyway, but what is the purpose of setting it then anyway?

And the `SERVICE_OPTS` in the Python gateway is just a map containing `omero.group: -1` and only used for the QueryService. Should be rather called `QUERY_OPTS`? So the QueryService seems to be a rather special thing, no?

mtbc:
> It didn't cause any issues so far.

As a user of the Java Gateway it fooled me into thinking that I had successfully set all-groups context on the query service. In fact my queries were being limited to the current group and my code unexpectedly failed in some situations.

I'll defer the `QUERY_OPTS` question to @joshmoore. :smiley:

dominikl:
Yes, I guess it's good to disallow setting it to `-1` then. This SecurityContext framework is totally confusing.

joshmoore:
Anything that can be done with `ServiceFactoryPrx.setSecurityContext` can also be done with the map. The reverse unfortunately is not true. There are things that currently can only be done by use of `{"omero.*": "..."}`.

see:
 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/ome/services/blitz/impl/ServiceFactoryI.java#L236
 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/services/sessions/SessionManagerImpl.java#L1255
 * https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/security/basic/BasicEventContext.java#L152

A server RFE/bug is likely: allow setting -1 via `setSecurityContext`. Currently, `SERVICE_OPTS` works around this by maintaining state that will be passed in each server call.

dominikl:
Just confirmed that too, setting it to `-1` doesn't do anything, and just uses the user's default group then. Better to disallow it in first place to avoid confusion 👍

dominikl:
I guess it makes sense to disallow setting it to `-1`. Good to merge with that respect 👍 

I'm still not sure, how to handle the cases where one explicitely wants a 'groupless' context, i. e. get all projects for a user no matter which group. ~~For some gateway methods (like `BrowseFacility.getHierarchy(...)`) one can pass through a `omero.sys.Parameters` object to achieve that, but that's not very user friendly, and for other methods (`getDatasets(...)` etc.) that's not even possible at the moment (at the java gateway level).~~ Should these methods then instead have an optional flag `ignoreGroup`, `acrossGroup`, or something like that?

Edit: Ignore that, I think not even `omero.sys.Parameters` can be used to do something across groups. It's just not possible on the java gateway level yet.

joshmoore:
seealso: https://github.com/imagej/imagej-omero/issues/84

joshmoore:
* This is a change in behavior. If omero.gateway were decoupled, I'd suggest a larger version bump for this.
 * Where do we want to capture the followup and what would be the timeline, since querying all will be a common tasks?

mtbc:
I've been trying to capture followup on the description's Trello card: I should add the ImageJ issue link.

I figured that it's a fairly minor behavior change given that it's basically now blocking a bug that was previously going undetected but happy to wait for 5.5 if preferred.

mtbc:
(The card's on the 5.x.x board so I'd like to imagine that timeline will be determined in board review.)

joshmoore:
Merging since this clarifies to the end user that it's _not_ supported. Happy to discuss best strategies for _supporting_ it next.

PR #5699, opened 2018-03-27
Leakage paranoia

chris-allan:
# What this PR does

After feedback regarding OMERO.web resource accumulation over time (openmicroscopy/management_tools#703) superficial research indicated that both `PIPE` and thread accumulation is possible under a variety of conditions. Especially in scenarios where the `Ice.Communicator` is **not** destroyed. This accumulation was reproduced in isolated scenarios of ascending complexity:

 1. Raw `omero.client` usage
 1. `BlitzGateway` usage
 1. OMERO.web usage

**Any** raw `omero.client` usage that does not follow the following pattern has the possibility to leak both pipes and threads; possibly until the process exits:

```
...
client =omero.client(...)
try:
   ...
finally:
   client.closeSession()
...
```

In order for `BlitzGateway` users to be compliant with this, they need to know about `.c` and to call `closeSession()` on it. As OMERO.web is an extensive user of `BlitzGateway` all uses there also need to conform.

This PR ensures that the canonical OMERO.web connection cleanup is available via `BlitzGateway.close()`. Users should now be in a safer scenario as not to unknowingly accumulate leaked resources. If this PR is approved **all** documented examples should be updated to reflect the aforementioned pattern.

This PR also addresses several cases where OMERO.web is not following the pattern and would leak pipes and threads under the following conditions:

 * Whenever a check for server version is made
 * Whenever a check for server up/down status is made
 * On every failed login
 * On every valid login

After some limited fuzzing I'm now of the opinion that **any** thread accumulation is a signal of not following the aforementioned pattern. Without this PR pipes and threads **will** accumulate in OMERO.web workers. `maxrequests` or regular, scheduled worker restarts are **essential** until this PR or a similar implementation is merged.

Finally, during the aforementioned research as to the causes of resource accumulation circular references were identified to be a significant contributor. This PR attempts to reduce the number of those circular references in the `BlitzGateway` implementation and cleans them up as part of `close()`.

# Testing this PR

As this PR makes changes to the fundamentals of `BlitzGateway` connection handling code and how it is interacted with in OMERO.web thorough testing of both valid and invalid user workflows will be required. Furthermore, it is essential that we attempt to monitor thread counts and see if we can find any other cases where valid gateway resource handling is not being performed.

/cc @emilroz, @stick, @kennethgillen

PR #5769, opened 2018-05-17
Do not encapsulate null connection behavior in try/finally block

sbesson:
See https://github.com/openmicroscopy/openmicroscopy/pull/5699 for the set of previous changes to this API

If `connector.create_gateway` fails to create a connection, the current contract is to return None. In the current implementation, this will cause the final `connection.close()` statement to fail.
Moving the nullity tests out of the try/finally block should restore the previous behavior e.g. when performing a failing version check while preserving the termination of connections when such a connection exist.

To test this PR, try to point a Web deployment to a server with a mismatching version (without overriding the version_check property). Without this PR e.g. OMERO.web 5.4.6, the connection should fail with a stack trace. With this PR, it should be handled more gracefully

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/connector.py
+++ b/components/tools/OmeroWeb/omeroweb/connector.py
@@ -209,9 +209,9 @@ class Connector(object):
 
     def is_server_up(self, useragent):
         connection = self.create_guest_connection(useragent)
+        if connection is None:
+            return False
         try:
-            if connection is None:
-                return False
             try:
                 connection.getServerVersion()
                 return True
@@ -223,9 +223,9 @@ class Connector(object):
 
     def check_version(self, useragent):
         connection = self.create_guest_connection(useragent)
+        if connection is None:
+            return False
         try:
-            if connection is None:
-                return False
             try:
                 server_version = connection.getServerVersion()
                 server_version = self.SERVER_VERSION_RE.match(server_version)


chris-allan (approved):

chris-allan:
Just re-examining the implementation and rationale for clarity. The problem here is that `Connector.create_guest_connection()` may return `None` not `Connector.create_gateway()`, correct?

joshmoore:
I assume this is also the cause for https://github.com/openmicroscopy/omero-web-docker/issues/24#issuecomment-397547408

sbesson:
@joshmoore yes the stack trace you pasted looks very much like the one this PR is intending to fix.

@chris-allan: sorry completely missed your comment. Yes I have been to reconstruct the contract of `Connector.create_guest_connection()` and my reading of it is that it would return either a valid output of `client_wrapper()` i.e. a `BlitzGateway` connection if succesful or `None` if an error occurs or the gateway failed to connect. A related question reading https://github.com/openmicroscopy/openmicroscopy/pull/5699/files#diff-8e7605ce49d4ebb2015c715980f85a15R182 is whether the contract of `Connector.create_guest_connection()` was slightly altered in 5.4.6. It looks like previously, you would only return `None` if an exception was propogated up the stack but you would have returned a `BlitGateway` even if the wrapper failed to connect.

will-moore:
See https://trello.com/c/6sCJsMH9/12-login-bug-when-server-not-running

PR #5770, opened 2018-05-21
wholly avoid trying to list mismatching scripts in read-only

mtbc:
Extends robustness of read-only servers of scripts in database not matching scripts in repository: pretends mismatching scripts do not exist.


mtbc:
This helped me get OMERO.server running with local repo when pointing to remote IDR test server DB.

joshmoore:
@mtbc : did you also consider if this check could be put at a higher-level? i.e. is there a behavior that the read-only servers should wholly abstain from?

mtbc:
Given how the script updating is wrapped into the load queries this change looks to me to be the best that adheres to our approach of keeping a light touch for the within-5.4 server work. In a new world where we work on omero-all for 5.5 or somesuch then there could well be some profitable refactoring to be done here.

joshmoore:
Here's to future refactoring

PR #5771, opened 2018-05-25
Inplace import

jburel:
# What this PR does

* Remove -- when not needed
* Update QA URL

Noticed during the preparation of the workshop
See
http://docs.openmicroscopy.org/omero/5.4.6/sysadmins/in-place-import.html#inplace-import


rleigh-codelibre:
These changes all look fine.  Was the dropping of `https` for `http` intentional?

jburel:
it is intentional it is not https for qa

joshmoore:
I think the intent here was to show using a non-default server. It may be better to go so far as to use `qa.example.com`.

jburel:
Thanks I will update that accordingly

joshmoore (approved):

PR #5772, opened 2018-05-28
adjust user plugin to first / last name terms

mtbc:
# What this PR does

Fixes CLI omero user help to be consistent about first / last name terms.

# Testing this PR

`bin/omero user add -h` and compare with this PR's code diff.

# Related reading

https://trello.com/c/g9wMfh5U/67-rfe-surname-name-and-lastname


joshmoore:
I don't see a particular issue with simplifying to use "first"/"last" and will go ahead and merge, and under the assumption that the other values were confusing, I'll go ahead and merge. I do notice that there's not much value in the doc strings at this point, but it's not likely to be worth too much consideration.

PR #5774, opened 2018-06-04
Disable fileannotation delete

will-moore:
# What this PR does

In order to allow linking of OMERO.figure FileAnnotations to Images (https://trello.com/c/3dlIno4Q/6-link-json-figure-files-to-images-dataset), we need to disable the Delete button so that users can't accidentally delete figure files.
We do this by specifying a configurable list of namespaces. Linked annotations with these namespaces will not be deletable in the webclient.

# Testing this PR

1. Create a figure file (save a figure)
2. Link this to an Image by Annotation - click [+] and find the figure file name to link to Image.
3. The Figure (with namespace ```omero.web.figure.json```) will not show a Delete X button.
4. The ```"disable_delete_for_namespaces"``` list is added to JSON from e.g. webclient/api/annotations/?type=file&image=9551 - Check URL in browser (image ID is not important)
5. Check that new test for this is passing.

 TODO:
 - Add Robot tests

### DIFF
--- a/components/common/resources/ome/config.xml
+++ b/components/common/resources/ome/config.xml
@@ -260,6 +260,11 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
+    <bean class="ome.system.Preference" id="omero.client.delete_disabled.namespaces">
+       <property name="db" value="false"/>
+       <property name="mutable" value="false"/>
+       <property name="visibility" value="all"/>
+    </bean>
     <!-- End preference list -->
             </list>
         </property>
--- a/components/common/resources/ome/config.xml
+++ b/components/common/resources/ome/config.xml
@@ -260,7 +260,7 @@
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
     </bean>
-    <bean class="ome.system.Preference" id="omero.client.delete_disabled.namespaces">
+    <bean class="ome.system.Preference" id="omero.client.disable_delete_for_namespaces">
        <property name="db" value="false"/>
        <property name="mutable" value="false"/>
        <property name="visibility" value="all"/>
--- a/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
@@ -138,20 +138,20 @@ class render_response(omeroweb.decorators.render_response):
         context['ome']['can_create'] = request.session.get('can_create', True)
         # UI server preferences
         if request.session.get('server_settings'):
-            s_settings = request.session.get('server_settings')
-            context['ome']['email'] = server_settings.get('email', False)
-            if s_settings.get('ui'):
+            context['ome']['email'] = request.session.get(
+                'server_settings').get('email', False)
+            if request.session.get('server_settings').get('ui'):
                 # don't overwrite existing ui
                 context.setdefault('ui', {'tree': {}})
                 context['ui']['orphans'] = \
-                    s_settings.get('ui', {}).get('tree', {}).get('orphans')
+                    request.session.get('server_settings').get('ui', {}) \
+                    .get('tree', {}).get('orphans')
                 context['ui']['dropdown_menu'] = \
-                    s_settings.get('ui', {}).get('menu', {}).get('dropdown')
+                    request.session.get('server_settings').get('ui', {}) \
+                    .get('menu', {}).get('dropdown')
                 context['ui']['tree']['type_order'] = \
-                    s_settings.get('ui', {}).get('tree', {}).get('type_order')
-            if s_settings.get('disable_delete_for_namespaces'):
-                context['ome']['disable_delete_for_namespaces'] = \
-                    s_settings.get('disable_delete_for_namespaces')
+                    request.session.get('server_settings').get('ui', {}) \
+                    .get('tree', {}).get('type_order')
 
         self.load_settings(request, context, conn)
 
--- a/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
@@ -138,20 +138,20 @@ class render_response(omeroweb.decorators.render_response):
         context['ome']['can_create'] = request.session.get('can_create', True)
         # UI server preferences
         if request.session.get('server_settings'):
-            context['ome']['email'] = request.session.get(
-                'server_settings').get('email', False)
-            if request.session.get('server_settings').get('ui'):
+            s_settings = request.session.get('server_settings')
+            context['ome']['email'] = server_settings.get('email', False)
+            if s_settings.get('ui'):
                 # don't overwrite existing ui
                 context.setdefault('ui', {'tree': {}})
                 context['ui']['orphans'] = \
-                    request.session.get('server_settings').get('ui', {}) \
-                    .get('tree', {}).get('orphans')
+                    s_settings.get('ui', {}).get('tree', {}).get('orphans')
                 context['ui']['dropdown_menu'] = \
-                    request.session.get('server_settings').get('ui', {}) \
-                    .get('menu', {}).get('dropdown')
+                    s_settings.get('ui', {}).get('menu', {}).get('dropdown')
                 context['ui']['tree']['type_order'] = \
-                    request.session.get('server_settings').get('ui', {}) \
-                    .get('tree', {}).get('type_order')
+                    s_settings.get('ui', {}).get('tree', {}).get('type_order')
+            if s_settings.get('disable_delete_for_namespaces'):
+                context['ome']['disable_delete_for_namespaces'] = \
+                    s_settings.get('disable_delete_for_namespaces')
 
         self.load_settings(request, context, conn)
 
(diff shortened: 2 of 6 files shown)


joshmoore:
It would be good to discuss the long-term plan here as well as the costs/benefits. I'm worried that this is another _client-side security_ feature that may leave users and their data at risk long-term since all clients will need to be updated to respect this property. Other options to that occur to me:

 * add the ns filtering into the delete command itself (cc @mtbc)
 * use a data structure which would prevent the delete
 * also link the annotations somewhere else to prevent delete

mtbc:
I wonder if it could somehow be in the gateways so then at least it might be a per-language implementation rather than per-client. With this PR is there still a different way that users can delete figures then? Would it be better to simply not list them among the normal annotations, as with measurements and companion files, rather than have them listed but undeletable? (I ask these things without really understanding the design/plan for handling figures.)

mtbc:
> link the annotations somewhere else

This may not help: I haven't verified but I think that multi-linking doesn't help if you directly target an annotation for deletion and if you delete just its link then I don't think the system ever automatically deletes the related attachment anyway.

will-moore:
We certainly still want to allow deleting when you're in the Figure app itself (just as you can now).
We want to attach figures to Images etc so that you can:
 - Open the figure, via API or UI (openwith needs to support file-annotations)
 - When you chgrp the image(s), the figure also chgrps

So, we don't want to hide this link from the UI, and we don't want to block deletion from the server or client libs. We simply want to prevent users from accidentally deleting the figure.

See screenshot: This image has a bunch of figures exported (pdf & tiff) and I want to delete them but if the figure file itself also had a X, it would be easy to delete it by mistake, especially since users won't know that the figure is linked here.

screen shot 2018-06-05 at 09 29 33

joshmoore:
> We certainly still want to allow deleting when you're in the Figure app itself (just as you can now).

I guess for me this sets off some flags. Adding properties to the core because one client needs different behavior feels like it could be a band-aid. How should other client developers achieve the same effect? At the moment, they'd have to document that sysadmins should set a property. Can we go a step further and have this auto-register somehow? Are there other namespaces that need to be taken into account?

> This may not help

Understood, @mtbc. Another idea is a more general use of the Namespace table.

will-moore:
For me, this is kinda like the custom behaviour we have for companion files, or enabling editing of some map annotations based on namespace, or omitting various scripts from the scripts menu. In all cases, we want the clients to not allow everything that the API allows.
At the moment, other client developers already have to tell sysadmins to set several properties to install the app, add to top_links, add to open-with etc. So this is just one more to add to the list.

pwalczysko:
All functionality works as described in the header of this PR.

For the integration test, I am guessing it is https://web-proxy.openmicroscopy.org/east-ci/job/OMERO-test-integration/ - but the tests there seem to be failing since long.

I will wait with further actions for the result of the discussion here.

joshmoore:
> For me, this is kinda like the custom behaviour we have for companion files, or enabling editing of some map annotations based on namespace, or omitting various scripts from the scripts menu.

Agreed, I don't think what you've done here is fundamentally different from what's been done in the past, but this is the most external client that we've done it for. What I mean by that is we've purposefully kept figure very much as a standalone code base with its own model etc. I think the changes here are in conflict with that. I'd like to see spelled out how we see moving forward with the development. Are the two moving closer? Or do we need to provide extension points so that _anyone_ can make these same types of changes without needing to introduce a new property, etc.

> In all cases, we want the clients to not allow everything that the API allows.

There's a caveat here that then _all_ clients need to be aware of this new property, i.e. essentially a breaking change.

> At the moment, other client developers already have to tell sysadmins to set several properties to install the app, add to top_links, add to open-with etc. So this is just one more to add to the list.

That's true, but again, this one sets off new flags for me since it's concerning loss of data.

joshmoore:
> concerning loss of data....

Sorry, I realized I only raised my warning flags here without offering any of the solutions that were in my head (I blame 😷). At a high-level, this PR says to me that the JSON files need a way to be made "more valuable".  Possible other options that occur to me:

|                                                                                             | *Pros*                                                                  | *Cons*                                                          |
|---------------------------------------------------------------------------------------------|-------------------------------------------------------------------------|-----------------------------------------------------------------|
| 1. teach the delete code about the new property                                             | Prevent all unintentional deletes                                       | Already tricky code; Requires extensive testing                 |
| 2. elevate your "do not delete" flag to the Namespace table so that handling is server-side | Same as 1 but queryable                                                 | Same as 1 plus requires a DB upgrade                            |
| 3. add some "external data" mechanism to the OMERO model that will be treated specially     | There will likely be similar use cases that we need to support.         | Requires DB upgrade; unclear what the mechanism would look like |
| 4. add the figure model to the OMERO model directly                                         | Brings figure into the core; probably provides the most value of figure | Requires DB upgrade; unclear how to handle cross-group links    |
| 5. store figure json in a separate PostgreSQL database/table                                | Fairly straight-forward including backups                               | Requires some configuration by admin                            |
| 6. store figure json in mongoDB (or similar)                                                |    Cool                                                                     | Requires a new server!                                          |


5 & 6 and to some extent 3 would require us to make some initial decisions regarding "federation", but we knew it would come eventually.

cc: @jrswedlow

jburel:
Note that the "question/issue" is not only limited to figure. Other annotation like modulo falls into the same category: if deleted there is no way to figure out how to handle correctly the data

Edit: modulo annotation are not companion files

mtbc:
>  custom behaviour we have for companion files

For what it's worth, server-side graph operations handle them differently in that if the client requests special treatment for orphaned annotations (e.g., "also delete all linked annotations") then those in companion-file-like namespaces remain excluded from that option, at least by default.

jrswedlow:
Thanks for all the great discussion and ideas, and especially to @joshmoore for writing out the various options and choices. @jburel's point that this issue is more general, which I think has been stated by others several times and ways, is very important.  While it takes longer, we should avoid band-aids that we have to fix later.  Can we have a discussion on the table of choices (Q1: is it complete?) and then decide together?

will-moore:
I find it hard to imagine an extension point that would allow the kinds of changes in this PR (including custom UI changes) without needing to introduce a new property.
Looking at all the properties we currently have to configure the webclient and/or insight and how they're handled, it doesn't seem possible to make this generic, so that we don't need separate properties.

But if we're trying to allow *anyone* to make these kind of changes with minimal breakage then it seems that considering DB upgrades is not the way to go.
E.g. we're now storing Orbit models and annotations (Polygons) in OMERO as file annotations with known namespaces. We don't want to need a DB upgrade or new server, we just want to disable accidental deletion in webclient/insight.

Happy to discuss.

pwalczysko:
I think we are missing one granularity in this discussion here. To paraphrase the table of @joshmoore where it says ``prevent all unintentional delete``.

I was wondering what that is in case of figures. Note: this is a very different case than the modulo, which should never be deleted except when the image is deleted as well.
But for figures, the delete is not unintentional, it might be even desired - note that there is no very nice workflow to delete figures in the figure itself. 
In the Orbit case mentoined here, the ``model`` of Orbit on the server -> sometimes the user might want to delete it, sometimes not. Other times the users do not know what they are doing. I think this is the case for Figures as well.

So: If there is a ``flag`` on the file which **prevents** me as a user to delete an Orbit model or a Figure (I mean prevents me to delete these objects from OMERO via UI, CLI, API...), how do I delete it then please ? Is it fair to say "You have to go to Orbit or Figure to delete it, otherwise sorry, no ??? I am afraid we are crafting here a set-in-stone feature which does not think about the usage too much.

will-moore:
@joshmoore Do you want to find a time to chat about this? Anytime is fine with me except 1pm on Friday

joshmoore:
Notes from conversation with @will-moore @mtbc @pwalczysko @jburel @sbesson:

 * Various comparisons ("Figure data is like...")
   * modulo
   * map annotations ("if the client didn't create them, it shouldn't delete them")
   * OMERO.tables
   * external data source linked via map annotation (url, description, icon, mimetype, tool suggestion)

 * Questions/Issues
   * Is figure internal or external?
   * Is figure group-agnostic or gnostic?
   * Why would you prevent me from deleting? Ok with creating a system?
   * As what do we consider figure data? (i.e. it's value) J-m: true for any thing that effects visualization (modulo, channel grouping) and can also be omitted from the UI.
   * Users do not understand what they are deleting.
   * What would be our ideal solution? (regardless of cost, etc.) e.g. subconsciously - "attached to image so that move to group is easier"

 * Design goals
   * Reducing the burden on future clients
   * Safety for "external" data
   * Maintain decoupling (suggests external route) unless clear MUST feature that requires it (compromise)

 * Additions or alternative solutions:
   * Different popup when deleting (prevent accidental delete)
   * Client-specific MapAnnotation per row of the Namespace (until we have a database upgrade) which drives a configurable right-hand panel (see below). cF. recent accordion request on the forums

```
GET /api/load_annotations_v2
{
  "default_map_annotations": [
      ...
      "open_with": https://orbit...,
      "canURL": "true",
   ]

}
```

Note: also think about third-party services which provide endpoints for providing _lists_ of things.

 * Additional high-level wishes
   * Indexing for search (e.g. labels on images) - only happens on export to PDF
   * Filter accordions that are not there
   * Add annotations to omero-marshal (but perhaps use a sub-`1.0` JSON API if necessary to get functionality to the user sooner)
   * Extensible right-hand panel (with the associated refactoring that was previously skipped)

mtbc:
Small notes that may help with the larger picture: namespaces have a `displayName` property and can be annotated with map annotations or whatever else. Further, attachments have a `mimetype` and can themselves be annotated.

will-moore:
TODO:
 - Use URL in image description or map-annotation to link to figure
 - Support Chgrp of figures from figure app (instead of relying on link to image)
 - Future: index figure JSON for searching, use image IDs in figure to index image labels -> image
 - Think about cleaning up right-hand panel - more extensible without hacks
 - For Image-of-a-Figure, use ```$ bin/omero config set omero.web.viewer.view omero_figure.views.open_image``` to open the figure linked to image.

PR #5775, opened 2018-06-07
Check settings.PUBLIC_ENABLED before settings.PUBLIC_USER in views.py

will-moore:
Fixes bug where you go to an image that doesn't exist when you don't have PUBLIC USER configured.

To test:
 - Without ```PUBLIC_USER``` configured, pick an Image ID that doesn't exist and go to ```/webclient/?show=image-ID```
 - Should show webclient without error.


will-moore:
This was already fixed in https://github.com/openmicroscopy/openmicroscopy/pull/5759

PR #5776, opened 2018-06-07
Add log plot to histogram

will-moore:
# What this PR does

Adds a Log plot to histograms. See https://github.com/openmicroscopy/design/issues/94

To test:
 - Show histogram
 - Log plot is shown by default
 - Can be hidden by unchecking the 'Log plot' checkbox

cc @ehrenfeu 

screen shot 2018-06-07 at 23 22 30

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_preview.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_preview.html
@@ -83,10 +83,6 @@
       .rdef .owner {
         color: blue;
       }
-      /* histogram svg */
-      .line {
-        fill: none;
-      }
     </style>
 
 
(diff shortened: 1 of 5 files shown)


jburel:
If we go for that option
we should have a similar change in iviewer

pwalczysko:
Worked as user-4.

On most histograms it is clear what is the log and what is the linear histogram, but on some, like
http://web-dev-merge.openmicroscopy.org/webclient/?show=image-6881
it is less clear (see screenshot below).
screen shot 2018-06-20 at 17 32 41

Also note, what is the strange vertical line on the histograms ? (another screenshot with vertical line see below, from http://web-dev-merge.openmicroscopy.org/webclient/?show=image-6885

screen shot 2018-06-20 at 17 34 20

I wonder what will happen when my histogram is in such a shape (unfortunately did not find a suitable image yet) that it covers the top right corner (and thus the checkbox itself)  ?

will-moore:
Fixed merge conflict...
@pwalczysko The 'strange vertical line on histograms' shows the position of the channel slider handles (this is more obvious and useful when you're dragging the sliders).
I think it would be quite rare to find an image where the plotted histogram overlaps the "Log plot" checkbox, but even if it did, it would lie behind the checkbox and label so I think they'd still be usable.
I don't know a better spot for the Log plot checkbox without adding an extra row in the UI which would look worse I think.

will-moore:
@joshmoore @jburel It would be nice to try and get this in 5.4.8 if no objections (could you remove the exclude label)?

jburel:
I still think we need to explore other mapping options as mentioned on the design issue
Adding it in the web is one thing but we will need to synch the other web viewers (figure, iviewer, etc.) 
I feel that this should come at once

will-moore:
Since the web apps are decoupled from OMERO, we should be able to release them separately? Like we did with LUTs etc.

jburel:
They are decoupled but with the number of things in the pipeline.
I would rather target another release. Also I think it is important to have potential support for other mapping if we decide to add them without changing the UI constantly

PR #5779, opened 2018-06-11
unlinkAnnotation.m components\tools\OmeroM\src\annotations

kouichi-c-nakamura:
# What this PR does

`unlinkAnnotation.m` works in the opposite way to `linkAnnotation.m` and allows you to unlink Annotation objects from the specified parent object. Without this function, it was quite tedious to achieve that.

You can see the discussion related to this.
https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8504&start=20

**SYNTAX**
unlinkAnnotation(session,parentType,parentID)
unlinkAnnotation(session,parentType,parentID,annt)

# Testing this PR

1. required setup
+ Need to know the image ID for an image
+ Need to know the tag ID for a tag

2. actions to perform

```matlab
imgID = 18868
tagID = 2676
tag1 = getTagAnnotations(session,tagID)

linkAnnotation(session,tag1,'image',imgID)

unlinkAnnotation(session,'image',imgID ,tag1)
```

3. expected observations

The last line should unlink the tag from the image.


components/tools/OmeroM/src/annotations/unlinkAnnotation.m
```diff
@@ -0,0 +1,166 @@
+function unlinkAnnotation(session,parentType,parentID,varargin)
+% unlinkAnnotation allows you to unlink all or specified
+% MapAnnotation (Key-Value Pairs in GUI) from parent in OMERO server.
```
sbesson: The short description probably undersells the scope of this function as it applies to all annotations

components/tools/OmeroM/src/annotations/unlinkAnnotation.m
```diff
@@ -0,0 +1,166 @@
...
+p.addRequired('parentType',@(x) ischar(x) && ismember(x, {objects.name}));
+p.addRequired('parentID',@(x) isnumeric(x) && isvector(x));
+
+p.addOptional('annt',[],@(x) isempty(x) || all(isa(x, 'omero.model.Annotation')));
+p.parse(session,parentType,parentID,varargin{:});
+
+annt = p.Results.annt;
+
+x = annt;
+assert(isa(x,'omero.model.MapAnnotationI') || isa(x,'omero.model.TagAnnotationI') ...
```
sbesson: What is the role of this `assert` since line 84 checked that `annt` was an Annotation? Note that it currently fails if `annt` is not passed

sbesson (commented):
Briefly tested using an extension of the `WriteData` training example to test the various usages of the function https://github.com/sbesson/openmicroscopy/commit/268fb4c0abe69c457bd375a7a6057e2a188808a5. Currently tested:
- the deletion of a single annotation associated to a project 
- the deletion of all annotations associated to a project.
Are there other use cases we should include into the daily training example?

Main functional blocker from this first round of testing is the assertion failure (line 90) when no annotation is passed.

A couple of general questions about the scope of the new function:
- as the default is to delete all annotations of a given object, maybe the function should be called `unlinkAnnotations` similar to what is done in the Python gateway
- someway related/inspired by the functions in the Python gateway, extensions of this functionality could include an optional namespace parameter (e.g. to delete all key/value pairs created via the UI). Probably a RFE.

In general looks like a good addition to the current toolbox. I have a few more minor stylistic comments that can be discussed once we sort out the contract and expectations of this new function.

In the future, we might want to invest into migrating OMERO.matlab to use the new `omero.gateway`, as a general question, is this functionality already covered by existing APIs @dom?

components/tools/OmeroM/src/annotations/unlinkAnnotation.m
kouichi-c-nakamura: For sure you can get rid of line 90. I personally cannot test all other types of Annotation objects so I narrowed it down just to be conservative.

components/tools/OmeroM/src/annotations/unlinkAnnotation.m
kouichi-c-nakamura: I agree that unlinkAnnotations may make better sense.

You can say that this works for all types Annotations once you tested.

components/tools/OmeroM/src/annotations/unlinkAnnotation.m
kouichi-c-nakamura: Thank you for the time and effort.

Should I change the code accordingly now?

sbesson: Yes please feel free to push the changes to this branch and we can retest. If we agree the example addition covers everything, let me know if you want to cherry-pick sbesson/openmicroscopy@268fb4c so that the function is tested by our daily Continuous Integration jobs. I can also open it as a separate PR.

kouichi-c-nakamura: I didn't know how to cherry pick to that branch so I added the changes manually. Hope they're in good shape.

jburel:
```
function unlinkAnnotations(session,parentType,parentID,varargin)
+% unlinkAnnotation allows you to unlink all or specified
+% MapAnnotation (Key-Value Pairs in GUI) from parent in OMERO server.
+%
+% SYNTAX
+% unlinkAnnotation(session,parentType,parentID)
+% unlinkAnnotation(session,parentType,parentID,annt)
```
The method has been renamed ``unlinkAnnotations`` , the description needs to be updated too

jburel:
The gateway could handle more gracefully the case when an annotation is already linked to an object

examples/Training/matlab/WriteData.m
```diff
@@ -672,6 +672,31 @@
     xas = getScreenXmlAnnotations(session, screenId1, 'include', ns);
     assert(hasAnnotation(xa, xas), 'WriteData: Could not find annotation');
 
+   % Unlink Annotations
+   % =====================
```
jburel: extra ``===``

components/tools/OmeroM/src/annotations/unlinkAnnotations.m
```diff
@@ -0,0 +1,161 @@
+function unlinkAnnotations(session,parentType,parentID,varargin)
+% unlinkAnnotations allows you to unlink all or specified
+% MapAnnotation (Key-Value Pairs in GUI) from parent in OMERO server.
```
sbesson: I would generalize this to any annotation as this function is more generic

components/tools/OmeroM/src/annotations/unlinkAnnotations.m
```diff
@@ -0,0 +1,161 @@
...
+% OriginalFile
+% PlaneInfo
+% Roi
+% Channel
+% Folder
+
+% https://github.com/openmicroscopy/openmicroscopy/blob/develop/examples/Training/matlab/ReadData.m#L334
+
+
+switch parentType
```
sbesson, line 103: The mapping between `parentType` and `ParentType` can also be retrieved directly from the dictionary created by `getObjectTypes` so I would vote for re-using this one if possible (can be a follow-up PR).
Also can the `parentType` vs `ParentType` cause issues in case-insensitive systems (Windows)?

components/tools/OmeroM/src/annotations/unlinkAnnotations.m
```diff
@@ -0,0 +1,161 @@
...
+        ParentType = 'PlateAcquisition';
+    case 'roi'
+        ParentType = 'Roi';
+end
+
+parents = session.getQueryService().findAllByQuery(...
+    [sprintf('select obj from %s as obj left outer join fetch obj.annotationLinks as link join fetch link.child as annotation where obj.id =  ',ParentType), ...
+    num2str(parentID)], []);
+
+% images = session.getQueryService().findAllByQuery(...
```
sbesson: I would simply delete these lines rather than having it commented without an explanation

sbesson (approved):
A couple of final comments to improve the maintenance and exposition of this new  function.
Overall, looks great, thanks for the work. The comments can be addressed either here or in a separate PR from my perspective. If you plan to push more commits in this PR, let us know @kouichi-c-nakamura.

sbesson (approved):
Thanks for the last changes. 

The new functionality makes sense and is successfully tested by https://ci.openmicroscopy.org/job/OMERO-DEV-merge-training/975/ with the latest version of this code. https://github.com/openmicroscopy/openmicroscopy/pull/5779/files#r198189525 is a mid-term maintenance improvement and can be handled separately. Unless @jburel wants to take a look at the latest commits, I propose to get this merged immediately for inclusion in OMERO 5.4.7. Thanks @kouichi-c-nakamura

kouichi-c-nakamura:
Thanks @sbesson for your responses. I'll wait to see the arrival of 5.4.7 with a bit of my code.

joshmoore:
I've relaunched the travis job since `pip install` failed.

sbesson:
The last commit generalizes the usage to support arrayes of IDs. This is not covered by the current training examples but introduces no regression on the tested use cases i.e. single-value parentID with and without optional annotation. Merging and improvements can be performed separately.

PR #5780, opened 2018-06-12
Fix loss of privileges when Full Admin edits themselves

will-moore:
# What this PR does

Fixes loss of privileges when Admin edits themselves: https://trello.com/c/oJPTE5vq/358-bug-edit-root-user

# Testing this PR

- Create a spare admin account for testing!
- Log in as Admin, and edit yourself (The Role controls will be disabled as before)
- Check that Save doesn't cause you to lose privileges (should still be full Admin after update.

I haven't had a chance to check this fix against ALL possible scenarios with various privileges, editing yourself or other users of various roles. Need a list of all testing scenarios to check.

cc @pwalczysko @mtbc

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webadmin/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/views.py
@@ -590,6 +590,7 @@ def manage_experimenter(request, action, eid=None, conn=None, **kwargs):
                 privileges = conn.get_privileges_from_form(form)
                 to_add = []
                 to_remove = []
+                # privileges may be None if disabled in form
                 if privileges is not None:
                     # Only update privileges that we have permission to set
                     # (prevents privilege escalation)
--- a/components/tools/OmeroWeb/omeroweb/webadmin/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/views.py
@@ -588,16 +588,18 @@ def manage_experimenter(request, action, eid=None, conn=None, **kwargs):
 
                 # Update 'AdminPrivilege' config roles for user
                 privileges = conn.get_privileges_from_form(form)
-                # Only process privileges that we have permission to set
                 to_add = []
                 to_remove = []
                 if privileges is not None:
+                    # Only update privileges that we have permission to set
+                    # (prevents privilege escalation)
                     for p in conn.getCurrentAdminPrivileges():
                         if p in privileges:
                             to_add.append(p)
                         else:
                             to_remove.append(p)
-                conn.updateAdminPrivileges(experimenter.id, to_add, to_remove)
+                    conn.updateAdminPrivileges(experimenter.id,
+                                               to_add, to_remove)
 
                 conn.updateExperimenter(
                     experimenter, omename, firstName, lastName, email, admin,
--- a/components/tools/OmeroWeb/omeroweb/webadmin/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webadmin/views.py
@@ -588,17 +588,15 @@ def manage_experimenter(request, action, eid=None, conn=None, **kwargs):
 
                 # Update 'AdminPrivilege' config roles for user
                 privileges = conn.get_privileges_from_form(form)
-                if privileges is None:
-                    privileges = []
                 # Only process privileges that we have permission to set
                 to_add = []
                 to_remove = []
-                for p in conn.getCurrentAdminPrivileges():
-                    if p in privileges:
-                        to_add.append(p)
-                    else:
-                        to_remove.append(p)
-
+                if privileges is not None:
+                    for p in conn.getCurrentAdminPrivileges():
+                        if p in privileges:
+                            to_add.append(p)
+                        else:
+                            to_remove.append(p)
                 conn.updateAdminPrivileges(experimenter.id, to_add, to_remove)
 
                 conn.updateExperimenter(
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -1143,7 +1143,10 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
 
         defaultGroup = ExperimenterGroupI(defaultGroupId, False)
 
-        if privileges is not None:
+        if isAdmin:
+            if privileges is None:
+                privileges = []
+
             if defaultGroupId not in otherGroupIds:
                 listOfGroups.append(ExperimenterGroupI(defaultGroupId, False))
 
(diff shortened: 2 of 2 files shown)


components/tools/OmeroWeb/omeroweb/webadmin/views.py
```diff
@@ -588,17 +588,15 @@ def manage_experimenter(request, action, eid=None, conn=None, **kwargs):
...
-                    else:
-                        to_remove.append(p)
-
+                if privileges is not None:
+                    for p in conn.getCurrentAdminPrivileges():
+                        if p in privileges:
+                            to_add.append(p)
+                        else:
+                            to_remove.append(p)
                 conn.updateAdminPrivileges(experimenter.id, to_add, to_remove)
```
mtbc: Maybe put this line in the outer `if` too?

components/tools/OmeroWeb/omeroweb/webadmin/views.py
```diff
@@ -588,17 +588,15 @@ def manage_experimenter(request, action, eid=None, conn=None, **kwargs):
...
                 to_add = []
                 to_remove = []
-                for p in conn.getCurrentAdminPrivileges():
-                    if p in privileges:
-                        to_add.append(p)
-                    else:
-                        to_remove.append(p)
-
+                if privileges is not None:
+                    for p in conn.getCurrentAdminPrivileges():
```
mtbc: How come it's relative to the current admin privileges, not those of the target user `experimenter`? Maybe I am just not understanding the gateway.

components/tools/OmeroWeb/omeroweb/webadmin/views.py
rgozim: `conn.updateAdminPrivileges(experimenter.id, to_add, to_remove)` should probably exist within the `if` block

pwalczysko:
Created https://trello.com/c/8FDcsXzI/455-restricted-admin-fails-silently-on-admin-downgr and https://trello.com/c/sb6GpUhf/454-restricted-admin-cannot-upgrade-user. 

Neither of the problems above is a regression, the problems are present on eel latest as well.

All the remaining workflows work as expected, the bug is fixed. Ready to merge FMPOV

mtbc:
I wonder if those problems relate to https://github.com/openmicroscopy/openmicroscopy/pull/5780#discussion_r195007375. But, yes, not a problem introduced by this PR.

components/tools/OmeroWeb/omeroweb/webadmin/views.py
will-moore: I think this is about privilege escalation. I can't give someone a permission that I don't have myself.

will-moore: With the code as it is now, I can only add or remove a privilege from someone if I have the privilege myself. Should you be able to remove a privilege from someone if you don't have the privilege yourself? cc @pwalczysko

pwalczysko: >Should you be able to remove a privilege from someone if you don't have the privilege yourself? cc @pwalczysko

IMHO the answer is no.  But the implementation as it stands now is different, which we simply took in our stride back in summer 2017 when this was first rolled out. Also, note the two problems specified in https://github.com/openmicroscopy/openmicroscopy/pull/5780#issuecomment-396896942 -> we are half allowing this, half failing in the process when this is acted upon in the UI

mtbc: I still don't understand the logic. For instance, why do we add a privilege as `toRemove` just because we don't have it?

will-moore: We don't add a privilege to remove unless we have it AND it's not chosen in the form for the user we're editing.
We go through all the privileges the I have ```for p in conn.getCurrentAdminPrivileges():``` since those are the only one's I can edit for another user.
For each privilege, if it was selected in the form, we add it, otherwise we remove it.

mtbc: Ah, thank you, that makes sense.

pwalczysko:
So https://github.com/openmicroscopy/openmicroscopy/pull/5780/commits/e3dc0aeea8d8b81dfada49fb8d7faa24bef21447 actually fixes https://trello.com/c/8FDcsXzI/455-restricted-admin-fails-silently-on-admin-downgr - now, restricted admin is able to downgrade a full admin all the way, and the UI is consistent.
All the other functionality is preserved as expected.

Although the restricted admin is IMHO empowered a bit too much here (see above), I think this is in line with the server behaviour.

Edit: the restricted admin in the above described case had 1. Create and Edit Users and 2. Add users to groups

Ready to merge FMPOV

joshmoore:
Which rights did the restricted admin have in this case?

will-moore:
This seems wrong (that last commit should have had NO effect on behaviour).
I'll try and reproduce.

pwalczysko:
@will-moore Sorry, my bad. The behaviour did not change in fact. I just masked it from myself by going:

   - restricted admin with Edit User and Add User to Group edits a full admin
   - makes a simple User from that full admin (action succeeds when checked later)
   - restricted admin from step 1 above now upgrades the simple User from step 2 above to restricted Admin -> this also works
   - a direct attempt of the restricted Admin to downgrade full admin to a restricted Admin will fail silently

Just for completeness, an attempt of restricted admin as described above to edit ``root`` and downgrade the ``root`` to restricted Admin fails silently. The attempt to downgrade ``root`` to a simple User fails with error ``ApiUsageException: exception ::omero::ApiUsageException
{
    serverStackTrace = ome.conditions.ApiUsageException: cannot remove light administrator privileges from the root user``

So all good here. Ready to merge.

sbesson:
One of the webadmin integration tests seems to have regressed since the introduction of this fix - see https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/885/. How easy is it to fix @will-moore ?

will-moore:
@sbesson Test fixed

joshmoore:
@will-moore : can you explain why a change is needed to the code to fix a test?

will-moore:
Previously, ```createExperimenter()``` was using the ```privileges``` parameter to determine whether to use ```createRestrictedSystemUserWithPassword()``` (when ```privileges is not None```) or to use ```createExperimenterWithPassword()```.

However, now that ```def get_privileges_from_form()``` returns an empty list instead of ```None``` for regular users, ```createExperimenter()``` was unexpectedly creating them admins (even when the ```isAdmin``` parameter was False). So, now I use the ```isAdmin``` parameter to decide whether to use ```createRestrictedSystemUserWithPassword()``` or not (which makes more sense).

joshmoore:
Thanks for the explanation, @will-moore. If I remember correctly these classes are all internal and so the change to the API isn't a concern, right? But even for ourselves it'd be good to have some of your explanation in:

 * the commit,
 * the code, and/or
 * the doc strings

so that when we come back to this code path in the future we have that context.

PR #5781, opened 2018-06-13
Close communicator after a failed session creation

joshmoore:
If an incorrect password or a bad session is entered, then the client object
was not being cleaned up. Previously, the finalizer was taking care of this,
but changes in newer Ice versions prevented that from happening. Rather than
relying on the Ice behavior, we can (try to) track down all of the unclosed
omero.client objects and close them.

# Testing this PR

1. bin/omero logout
2. bin/omero login (with the wrong password)
3. see if the communicator warning disappears

# Related reading

 * https://trello.com/c/DPv5q2Ec/421-command-line-intermittent-error

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/sessions.py
+++ b/components/tools/OmeroPy/src/omero/util/sessions.py
@@ -342,24 +342,28 @@ class SessionsStore(object):
         client = omero.client(props)
         client.setAgent("OMERO.sessions")
 
-        if sudo is not None:
-            sf = client.createSession(sudo, pasw)
-            principal = omero.sys.Principal()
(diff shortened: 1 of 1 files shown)


jburel:
Is it something that could be captured with an integration test? (assuming that it is not to bad to parse the output)

joshmoore:
A problem with output parsing is that you have to do it on a case by case basis since you can only capture it once, so any test that's already doing so (e.g. all import tests) will have to be modified to check for the string.

We can certainly write a new test for specifically this login/logout case, but doing this for the general case probably isn't worth it.

jburel:
as you said probably not worth it

PR #5784, opened 2018-06-18
Insight: add null-check to ratings editor

mtbc:
Adds a null check for getting user rating data. See [QA \#21545](https://www.openmicroscopy.org/qa2/qa/feedback/21545/).


rgozim:
Just wondering though if it's better to return `Collections.emptyMap()` from the `getAllUserRatingAnnotation()` rather than `null`? At least it would save having to do any `null` checks in any other uses of the `getAllUserRatingAnnotation()` that may occur in future, or already exist.

mtbc:
Probably. As the method can already return `null` and its Javadoc warns of that then I figured I'd keep my changes minimal as I don't know this code and Insight is maintenance-only anyway.

rgozim:
Good point. Lets not rock the boat. PR is good for merging!

PR #5786, opened 2018-06-19
datetime fix annotations

will-moore:
# What this PR does

See https://trello.com/c/HEXZNdSC/227-omeroweb-default-timezone-is-different-from-omeroserver

This fixes the bug where Annotations show time-stamps 1 hour out in Summer time.
However, it doesn't yet try to determine the timezone as suggested in [comment](https://trello.com/c/HEXZNdSC/227-omeroweb-default-timezone-is-different-from-omeroserver#comment-5b28bc6bc6ba992ea159cfa7) as there seems to be no way to do that.

To test:
 - Add a Comment
 - Check that the time-stamp on the Comment shows the current time.
 - Repeat this for servers in a bunch of different time-zones ;)

NB: I think the bug was due to my misunderstanding e.g. https://stackoverflow.com/questions/19654578/python-utc-datetime-objects-iso-format-doesnt-include-z-zulu-or-zero-offset

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/tree.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/tree.py
@@ -460,7 +460,8 @@ def _marshal_date(timestamp):
         def utcoffset(self, dt):
             # https://stackoverflow.com/questions/1111056/
             # /get-time-zone-information-of-the-system-in-python
-            offset = time.timezone if (time.localtime().tm_isdst == 0) else time.altzone
+            offset = time.timezone if (time.localtime().tm_isdst == 0) \
+                else time.altzone
             return timedelta(0, 0, 0, 0, 0, offset / 60 / 60 * -1)
 
         def tzname(self, dt):
--- a/components/tools/OmeroWeb/omeroweb/webclient/tree.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/tree.py
@@ -457,7 +457,7 @@ def marshal_datasets(conn, project_id=None, orphaned=False, group_id=-1,
 def _marshal_date(time):
     try:
         d = datetime.fromtimestamp(time/1000)
-        return d.isoformat() + 'Z'
+        return d.isoformat()
     except ValueError:
         return ''
 
--- a/components/tools/OmeroWeb/omeroweb/webclient/tree.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/tree.py
@@ -25,7 +25,7 @@ import omero
 from omero.rtypes import rlong, unwrap, wrap
 from django.conf import settings
 from django.http import Http404
-from datetime import datetime
+from datetime import datetime, tzinfo, timedelta
 from copy import deepcopy
 from omero.gateway import _letterGridLabel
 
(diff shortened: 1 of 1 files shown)


pwalczysko:
Tested:

   - adding Tags, Comments, KVPs and FileAttachment in RHP
   - importing fresh image (into fresh Dataset)
   - creation of new Project

All three actions show correct timestamps in Webclient.

Definitely improves the behaviour. (hopefully works in winter as well  :)

Ready to merge FMPOV

joshmoore:
> Definitely improves the behaviour. (hopefully works in winter as well :)

Shall we setup a server with winter time then?

pwalczysko:
>Shall we setup a server with winter time then?

Makes sense, if not too much hassle, thank you, or tell me how to do it and I can have a go @joshmoore

will-moore:
Based on https://stackoverflow.com/questions/1111056/get-time-zone-information-of-the-system-in-python and https://docs.python.org/2/library/datetime.html#tzinfo-objects I have tried to offset timestamps according to local time zone.

To test, we need to set up servers with different time zones!

I tried this locally, setting my Mac's timezone to New-York time. However, when I run Django dev server and ```print time.tzname``` in settings.py I see this (settings.py gets imported 4 times):

```
('EST', 'EDT')
('EST', 'EDT')
('GMT', 'BST')
('GMT', 'BST')
```
as if it's originally starting with the EST (New-York) and then correcting it to GMT.

joshmoore:
> To test, we need to set up servers with different time zones!

```
$ docker run -ti --rm --entrypoint=python openmicroscopy/omero-web-standalone:5.4 -c "import time; print time.tzname; print time.localtime()"
('UTC', 'UTC')
time.struct_time(tm_year=2018, tm_mon=6, tm_mday=21, tm_hour=7, tm_min=15, tm_sec=3, tm_wday=3, tm_yday=172, tm_isdst=0)

$ docker run -ti --rm --entrypoint=python -e TZ=America/Los_Angeles openmicroscopy/omero-web-standalone:5.4 -c "import time; print time.tzname; print time.localtime()"
('PST', 'PDT')
time.struct_time(tm_year=2018, tm_mon=6, tm_mday=21, tm_hour=0, tm_min=9, tm_sec=6, tm_wday=3, tm_yday=172, tm_isdst=1)

$ docker run -ti --rm --entrypoint=python -e TZ=Europe/London openmicroscopy/omero-web-standalone:5.4 -c "import time; print time.tzname; print time.localtime()"
('GMT', 'BST')
time.struct_time(tm_year=2018, tm_mon=6, tm_mday=21, tm_hour=8, tm_min=9, tm_sec=29, tm_wday=3, tm_yday=172, tm_isdst=1)
```

i.e. start with the docker-compose from http://www.openmicroscopy.org/explore/ and add `TZ`:

```
...
services:
  ...
  web:
    environment:
      - TZ=<whatever you want here>
     ...
```

sbesson:
Several integration tests also seem to have regressed with this change and need to be fixed accordingly - see https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/890/. Excluding so that we can try to get the integration builds green for now and since the manual testing will require some testing coordination.

pwalczysko:
Still finding this bug on outreach. It is quite a forward-facing one.

pwalczysko:
@manics suggests we should find first out where the server stores the timezone. From there, we could judge on validity of the changes here.

manics:
Expanding on the previous comment:
1. Figure out what timezone the server uses to store the datetime of the annotation in postgres
2. Figure out whether the saved datetime on the server changes depending on summer/winter time
3. Request the datetime of a given annotation from the server. Does it change depending on whether it's retrieved in summer or winter? Does it change depending on whether the annotation was created in summer or winter?
4. See how the displayed web datetime compares to the previous server datetimes

mtbc:
Last year's [[ome-devel] Timestamp over HQL](https://lists.openmicroscopy.org.uk/pipermail/ome-devel/2018-January/004109.html) was an interesting thread. Note we've collected some past discussion at [[trello] OMERO.web default timezone is different from OMERO.server](https://trello.com/c/HEXZNdSC/227-omeroweb-default-timezone-is-different-from-omeroserver).

pwalczysko:
@will-moore I think I am missing something here. Obviously, the change proposed here is 

    - breaking integration tests
    - needs very generous and complex testing 

But, coming from my original observation, which is

   - create a project on outreach server in Web, observe the creation timestamp is correct
   - link an attachemt to anything on outreach server in Web, observe the creation timestamp is one hour off

I cannot help but to conclude that the

   - the timestamping code is different for Project and Attachments in the Web
   - this seems to be illogical to me
   - if the timestamping code of Attachments is set to use the same logic as the timestamping code on Projects, the change would be smaller, it would not break the tests (maybe, just a hope of course), and would not need so much testing ???

manics:
I agree with @pwalczysko, timezones are complicated and we need to fix it in OMERO, but this discrepancy should be resolvable without it. Attachments should use the same date handling as projects (even if it's technically incorrect) for consistency.

manics:
I'm testing a possible fix

manics:
See https://github.com/openmicroscopy/openmicroscopy/pull/6009

PR #5787, opened 2018-06-19
import.py: generate scripts for use with parallel

joshmoore:
Remove the need to process the output of a dry-run. E.g.

```
bin/omero import --bulk X.yml
```
by instead setting the value of dry-run to a pattern as such `/tmp/%s.sh`. These scripts can then be used as the input to parallel.

# Testing this PR

1. Setup the following files:

bulk.yml
```
continue: "true"
transfer: "ln_s"
exclude: "clientpath"
checksum_algorithm: "File-Size-64"
logprefix: "logs/"
output: "yaml"
dry_run: "script%s.sh"
path: bulk.tsv
```

bulk.tsv
```
bulk.dir
```
(could also use bulk.fake)

2. Run `bin/omero import --bulk bulk.yml  -f`

3. Execute `script1.sh` manually

(Note: the `dry_run` flag itself is not yet enough to prevent a login from being required. Adding the `-f` flag works around this issue)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,24 @@ RUN apt-get update \
  && pip install tables "zeroc-ice>3.5,<3.7"
 # TODO: unpin pip when possible
 RUN adduser omero
-COPY . /src
+
+# TODO: would be nice to not need to copy .git since it invalidates the build frequently and takes more time
+COPY .git /src/.git
+
+COPY build.py /src/
+COPY build.xml /src/
+COPY components /src/components
+COPY docs /src/docs
+COPY etc /src/etc
+COPY ivy.xml /src/
+COPY lib /src/lib
+COPY luts /src/luts
+COPY omero.class /src/
+COPY setup.cfg /src/
+COPY sql /src/sql
+COPY test.xml /src/
+COPY LICENSE.txt /src/
+COPY history.txt /src/
 RUN chown -R omero /src
 USER omero
 WORKDIR /src
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,9 @@ ARG BUILD_IMAGE=openjdk
 # The produced /src directory will be copied the
 # RUN_IMAGE for end-use. This value can also be
 # set at build time with --build-arg RUN_IMAGE=...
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -43,11 +45,11 @@ RUN sed -i "s/^\(omero\.host\s*=\s*\).*\$/\1omero/" /src/etc/ice.config
 #
 #     https://trello.com/c/rPstbt4z/216-open-ssl-110
 #
-# RUN sed -i 's/\("IceSSL.Ciphers".*ADH\)/\1:@SECLEVEL=0/' /src/componenthreadts/tools/OmeroPy/src/omero/clients.py /src/etc/templates/grid/templates.xml
+RUN sed -i 's/\("IceSSL.Ciphers".*ADH[^"]*\)/\1:@SECLEVEL=0/' /src/components/tools/OmeroPy/src/omero/clients.py /src/etc/templates/grid/templates.xml
```
joshmoore: Except for uncommenting this, I think this PR is ready for testing once https://github.com/openmicroscopy/openmicroscopy/pull/5807 is merged and travis is re-run.

joshmoore:
Closing & re-opening

joshmoore:
If green (expected), then I leave someone else to decide if this functionality is sufficiently useful for inclusion.

joshmoore:
I won't have time to look into the unit test failure.

sbesson:
See https://github.com/sbesson/openmicroscopy/tree/bulk-fixes for commits fixing the unit test (and the case where `dry_run: true`). I have not written a unit test for `dry_run: <path>` yet but that might be a good addition.

dominikl:
Not sure what the docker stuff does, but the import.py change looks good and works 👍 .
These scripts could then simply be run with parallel like that:
`find pwd -name 'script*.sh' | parallel --jobs 2`

joshmoore:
Thanks @dominikl. The Docker bits are an attempt to re-activate integration testing in travis. I'll modify .travis.yml in a separate PR so that we can get this functionality in. It might be that @mtbc's work on parallelization will require some adjustments here, but we'll handle that in his PR.

PR #5793, opened 2018-06-26
Open with

jburel:
# What this PR does

Review the Open With option introduced a while back
At the time, all the images were downloaded as OME-TIFF
This PR adds the option to download the file in its original format

# Testing this PR

* Open an image using the Open With option
* Check that the original image is downloaded
* Set the option to false in the configuration file
* Check that an OME-TIFF is generated

# Related reading

https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8516

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/ui/OpenObjectLoader.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/ui/OpenObjectLoader.java
@@ -21,6 +21,11 @@
 package org.openmicroscopy.shoola.env.ui;
 
 import java.io.File;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 
 import org.openmicroscopy.shoola.env.config.Registry;
 
@@ -29,6 +34,10 @@ import omero.gateway.SecurityContext;
 import org.openmicroscopy.shoola.env.data.views.CallHandle;
 import org.openmicroscopy.shoola.util.filter.file.OMETIFFFilter;
 import org.openmicroscopy.shoola.util.ui.UIUtilities;
+
+import bitronix.tm.utils.CollectionUtils;
+
+import com.vividsolutions.jts.util.CollectionUtil;
 
 import omero.gateway.model.DataObject;
 import omero.gateway.model.FileAnnotationData;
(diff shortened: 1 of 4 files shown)


dominikl:
Works like expected, good to merge 👍

joshmoore:
Is there a substantial change in the screenshot? Any docs which need updating?

jburel:
There are no UI changes and this feature is not indicated in ome-help (only the web option)

components/insight/SRC/org/openmicroscopy/shoola/env/ui/OpenObjectLoader.java
```diff
@@ -93,14 +119,22 @@ public void load()
...
-    		name += "_"+image.getId();
-    		path += UIUtilities.replaceNonWordCharacters(name)+"."+OMETIFFFilter.OME_TIFF;
-    		f = new File(path);
-    		f.deleteOnExit();
-    		handle = ivView.exportImageAsOMETiff(ctx, image.getId(), f, null,
-    				this);
+    		if (originalImage) {
+    		    List<DataObject> objects = new ArrayList<DataObject>();
+    		    objects.add(image);
+    		    f = new File(folderPath);
```
joshmoore, line 126: Will this `folderPath` get cleaned up on exit?

jburel: I will need to check.
This is very old part of the code i.e. where we store things

jburel: I reviewed the code. The file is stored in the specified tmp directory
Nothing was deleted on exit

joshmoore:
Steps taken:

 * downloaded from ci-master
 * set OpenWith/Original to `true`
 * logged into eel and used open_with fiji.app on an image
 * saw that it was created as $TMPDIR/test_image.png
 * closed fiji & insight: the file wasn't deleted.

Note: I also tested with a multi-file download. Both files were downloaded to $TMPDIR correctly, but were also not deleted.

jburel:
It must have been the case since we introduce the feature since ``deleteOnExit`` was invoked at least in the fiji case

jburel:
Which platform are you using?
The file is deleted on mac

joshmoore:
OS X (10.12.6)

```
# joshuas-mbp:~ jamoore$ echo $TMPDIR
/var/folders/z5/txc_jj6x5l5cm81r56ck1n9c0000gn/T/
# joshuas-mbp:~ jamoore$ ls $TMPDIR/{open_with,test_import}*
/var/folders/z5/txc_jj6x5l5cm81r56ck1n9c0000gn/T//open_with.fake
/var/folders/z5/txc_jj6x5l5cm81r56ck1n9c0000gn/T//open_with.fake.ini
/var/folders/z5/txc_jj6x5l5cm81r56ck1n9c0000gn/T//test_import.fake
/var/folders/z5/txc_jj6x5l5cm81r56ck1n9c0000gn/T//test_import.png
```

jburel:
problem should be sorted now

joshmoore (approved):
Files are definitely deleted now. I'd say let's go for it.

For a potential follow-up, one thing I noted: since the last files hadn't been deleted, the new download had their names changed:

```
-rw-r--r--    1 jamoore  staff          0 Jun 29 13:16 open_with.fake(1).ini
-rw-r--r--    1 jamoore  staff          0 Jun 29 13:16 open_with(1).fake
```

This strategy will be unsafe for some file formats and instead a separate subdirectory per fileset download should be used.

jburel:
The "renaming" is something not directly related to the "open with" functionality
The same situation will happen when clicking on "download"
I went for the smallest modifications required in this PR
but certainly something to potentially look into

joshmoore:
In the download case, is the user given the chance to chose the location? If so, would that same mechanism _easily_ be an option here?

jburel:
The user can chose the location but a file with the same name could be there too
The download/open with and possibly other options use the same method to download. The renaming logic is implemented in one method so I think the strategy indicated above should not be difficult to implement

sbesson:
Captured follow-up todos as https://trello.com/c/mYgWd2vx/462-insight-rfe-prevent-renaming-on-download-open-with.

PR #5245, opened 2017-04-10
Social media absolute url config

manics:
# What this PR does

Allows absolute URL for social media to be overriden by the OMERO.web administrator, the default is to use Django's `build_absolute_uri` which probably uses the `Host` header sent by the client.

--depends-on https://github.com/openmicroscopy/openmicroscopy/pull/5244

In conjunction with https://github.com/openmicroscopy/openmicroscopy/pull/5244 this includes all required changes from https://github.com/openmicroscopy/openmicroscopy/pull/4744 (some commits were already rebased to develop, others are unnecessary following the addition of `omero.web.api.absolute_url` in mainline.

# Testing this PR

1. Configure a public web user, enable social media options, e.g. https://github.com/manics/deployment/blob/9bba3a5c7223acb6b9f01c16fc5435778796d2e8/ansible/group_vars/omero-hosts.yml#L109
2. Check the twitter and opengraph metadata is present in the HTML source of the full image viewer
3. Set a custom value for `omero.web.api.absolute_url`, check this is reflected in the hard-coded URL.

# Related reading

https://trello.com/c/z2U5K6ht/1-omero-5-3-migration

@aleksandra-tarkowska

PR #5795, opened 2018-06-27
Add support for twitter cards (full image viewer only) (rebased onto metadata52) (rebased onto metadata53) (rebased onto metadata54)

manics:
This is the same as gh-5216 gh-4613 gh-4260 but rebased onto metadata54.

---

Includes gh-5245 rebased onto metadata54.

---


manics:
--rebased-from #5216

manics:
--rebased-from #5245

joshmoore:
Is this a discrepancy between metadata54 and develop?

manics:
I can't find any sign of it on the current develop

joshmoore:
Then likely best to open that (instead/as well) against develop and we get it tested in the mainline. Then we can merge 5.4.7 into metadata54.

manics:
--rebased-to #5796

manics:
--rebased-to #5797

manics:
This should be upstream now

PR #5801, opened 2018-07-02
in webadmin gray out user role when editing root user

mtbc:
In webadmin adjusts user editing form so that one may not choose to downgrade root user from full admin. See https://trello.com/c/oJPTE5vq/2-bug-edit-root-user.


pwalczysko:
This works as expected in following sense:


   - root cannot change their privileges in the UI (expected)
   - root can change privileges of other admins in UI (expected)
   - other admin cannot change privileges of root in UI (expected)
   - other admin can change privileges of a third admin (expected)
   - other admin cannot change privileges of themselves (expected)

The above features are either good additions to the behaviour or are in lines with the old behaviour.

Ready to merge FMPOV

PR #5798, opened 2018-06-29
disable pyramid building on read-only servers

mtbc:
# What this PR does

For read-only servers disables the periodic triggering of pixel thread runs and ignores missing pyramid messages.

# Testing this PR

Check for functionality regressions in the pixels service or new complaints in logs, especially with importing images like `8kx8k.jpg`. Include a multi-server setup in which read-write and read-only coexist.

# Related reading

https://trello.com/c/xXQSXhzw/14-read-only-pixeldata-thread

PR #5802, opened 2018-07-03
disable pyramid building on read-only servers (rebased from develop)

mtbc:
Rebased from #5798 to aid testing. cc: @manics


mtbc:
Happy to reopen if useful for testing.

PR #5800, opened 2018-07-02
temporarily disable bin/omero fs rename

mtbc:
# What this PR does

Disables `fs rename` in CLI because it makes images' pixel data unreadable.

Takes a simple, expedient approach. We should fix this promptly and I suspect nobody uses it anyway.

# Testing this PR

Check that `fs rename` now raises an exception with a related message.

# Related reading

https://trello.com/c/J3LNquSH/1-bug-omero-fs-rename

PR #5807, opened 2018-07-06
travis: disable `integration`

joshmoore:
gh-5800 is currently failing on the integration step. gh-5787 started
trying to fix this but is not yet green. Disabling for the moment.

cc: @sbesson @mtbc @jburel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,6 @@ before_install:
     - export PATH=$PATH:$HOME/.local/bin
     - if [[ $BUILD == 'build-python' ]]; then travis_retry pip install --user flake8==2.4.0 pytest==2.7.3; fi
     - if [[ $BUILD == 'build-python' ]]; then ./components/tools/travis-build py-flake8; fi
-    - if [[ $BUILD == 'integration' ]]; then git clone git://github.com/openmicroscopy/omero-test-infra .omero; fi
 
 # retries the build due to:
 # https://github.com/travis-ci/travis-ci/issues/2507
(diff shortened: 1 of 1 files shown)


manics:
From #5800: `E           ln: failed to create hard link '/OMERO/ManagedRepository/11a98b27-9736-488b-bb49-d2c79f903764_2/2018-07/06/07-51-57.322/import_fake_file_1vWFp8d&series=1&extra_args=[\'--transfer=ln\'].fake' => '/opt/omero/server/omero/tmp/omero_omero-server/919/import_fake_file_1vWFp8d&series=1&extra_args=[\'--transfer=ln\'].fake': Invalid cross-device link`

Potential fixes:
- Run `df` in travis to get a list of mounts (maybe they moved `/opt` onto a separate partition?)
- Move the fake file to under `/OMERO/tmp` instead to guarantee it's on the same device as `/OMERO`

joshmoore:
Thanks, @manics. Linked to your comment from omero-test-infra, but for the moment since there are other change too, I'd rather just get this in so PRs go green again.

sbesson:
Thanks all. Agreed with everything above. Merging this immediately to prevent PRs from blocking ongoing PRs to be excluded because of upstream changes. I understand ways to address this problem are investigate independently. Once done, we can look into re-enabling these integration tests.

mtbc:
> Move the fake file to under `/OMERO/tmp`

looks like a good idea. Coming back to this issue is captured on a card or something?

joshmoore:
https://github.com/openmicroscopy/omero-test-infra/issues/20

PR #5809, opened 2018-07-11
add helper method for Java tests to log in users by name

mtbc:
Creates a new `AbstractServerTest.loginUser(String)` convenience method.


dominikl:
👍  Good to merge from my point of view!

joshmoore:
:+1: Surprised that this didn't exist previously. Perhaps long term we could plan on unifying Python's testlib and Java's base tests.

PR #5810, opened 2018-07-13
Preview pane large channel count features

chris-allan:
# What this PR does

A couple of useful OMERO.web features for the preview pane when working with large (10+) channel counts:

* Scroll the channel rendering settings rather than the entire preview tab
* Option to toggle on or off display of only the active channels rendering settings

# Testing this PR

1. required setup

Regular image and an image with a large number of channels. For our own testing we used two fake files:

 * `lotsachannels&sizeC=30.fake`
 * `threechannels&sizeC=3.fake`

2. actions to perform

The preview pane in the right hand panel should only have the list of channels scroll rather than the entire pane:

image

And the "Active" checkbox should toggle the display back and forth based on which channels are currently active:

image

3. expected observations

See above. No changes to the behaviour of other preview pane controls should be noticed and resizing of the right hand panel should behave as previously.

# Related reading

N/A

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
@@ -541,10 +541,21 @@
         //border-left: solid 1px hsl(210,10%,85%);
         min-width:240px;
     }
-    
-   
-   
-   
+
+    #preview_tab .right_tab_inner {
+        display: flex;
+        flex-direction: column;
+    }
+
+    #preview_tab .right_tab_inner > * {
+        flex-shrink: 0;
+    }
+
+    #preview_tab .right_tab_inner #rdef-container {
+        flex-shrink: 1;
+        overflow-y: auto;
+    }
+
     /* Seperator */
     
     hr {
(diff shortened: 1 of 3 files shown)


will-moore:
Looks good and is working fine. 
Tiny suggestion: I noticed that the tooltip for the Greyscale and Active controls only works over the checkbox itself and not the label. One solution to fix that without duplicating the tooltip is to nest them (also don't need 'for' attribute then)
```
<label  title="Show only active channels">
  <input id="showactive" type="checkbox" />
  Active
</label>
```

NB: I noticed while testing a 69-channel image http://web-dev-merge.openmicroscopy.org/webclient/?show=image-96302 (user-3) that the Save of rendering settings doesn't work in Preview panel or iviewer because the query string is too long. Also, rendering of the image for a 76-channel image doesn't work in Preview panel because of the same issue.
Card created at https://trello.com/c/RMUkeAfu/8-webclient-many-channel-images

jburel:
This is already an improvement. 
Tooltip is now showing for both checkbox and tooltip.
 
When scrolling a large number of channels, the controls like ``Grayscale`` are not always visible. This could be improved in a follow-up PR

chris-allan:
Agreed.

Want to do one additional thing where we limit the max height of the viewer so when the pane is expanded horizontally the controls don't get lost.

chris-allan:
When the right hand panel is made wide enough that its width exceeds ~80% of its height the controls end up beyond the bottom of the preview tab. Not a problem when we allowed the entire tab to scroll but now that we don't it is.

160bdd765c34cb6c2135c1cffa4170ee5f342014 pins the viewport so that its height will never be > 75% of the available height of the tab.

jburel:
Changes look good.
Capture comment on https://trello.com/c/C1qnSme4/479-web-preview-large-number-of-channels

jburel (approved):

PR #5812, opened 2018-07-26
deprecate bin/omero config list

mtbc:
# What this PR does

Deprecates OMERO.cli `config list`.

# Testing this PR

Try using `config list` and see a warning message, though it should still work.

# Related reading

https://trello.com/c/7KNfAJAD/373-cli-config-list-deprecation

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -280,7 +280,7 @@ class PrefsControl(WriteableConfigControl):
 
     @with_config
     def list(self, args, config):
-        self.ctx.err('warning: "config list" is deprecated, '
+        self.ctx.err('WARNING: "config list" is deprecated, '
                      'use "config get" instead')
         args.KEY = []
         self.get(args, config)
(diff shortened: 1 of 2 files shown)


jburel:
* run: ``bin/omero config list``
* output:  ``warning: "config list" is deprecated, use "config get" instead``
It works as expected

manics:
Could you also hide the command from `omero config --help`: https://stackoverflow.com/questions/11114589/creating-hidden-arguments-with-python-argparse

mtbc:
Probably partly through not much knowing either Python or argparse, I don't understand enough to get the suppression working.

manics:
Looks like it's a sub-parser not an argument which complicates things. It should be possible according to https://stackoverflow.com/questions/21692395/hiding-selected-subcommands-using-argparse but I'm not sure exactly how. If it's not easy maybe just add "deprecated" to the help message instead?

mtbc:
It already says it's deprecated?

manics:
I meant here: https://github.com/mtbc/openmicroscopy/blob/8ebfa1fab7a7a3c646b146ec76f0eecea4879567/components/tools/OmeroPy/src/omero/plugins/prefs.py#L143

mtbc:
Aha, yes, thank you.

mtbc:
Some interesting discussion in https://bugs.python.org/issue22848.

sbesson (approved):

PR #5813, opened 2018-07-31
Have Insight import dialog report empty files as such.

mtbc:
This PR adjusts Insight's import dialog so that when one adds empty files (e.g. fakes) for import then their size is listed as "empty" rather than "--" which I found confusingly indeterminate. Non-empty file sizes should be reported as before.


dominikl:
👍  Good to merge

PR #5815, opened 2018-08-01
Bring custom HTTP response in line with cleanup semantics

chris-allan:
Use the same cleanup semantics, which were introduced in order to avoid
filehandle and socket leakage (in #5699), as the rest of the codebase.

# What this PR does

More explicit gateway connection closure.

# Testing this PR

1. required setup

Valid session and the following available:

 * Loose, or attached via FileAnnotation, OriginalFile ID
 * FileAnnotation ID with associated OriginalFile
 * Image ID and Well ID in both single and multiples with archived files available

2. actions to perform

`ab -C 'sessionid=<session_id>' -n 1000 'http://host:port/<endpoint>'`

Endpoints, should be tested with both valid and **missing** identifiers:

 * `webclient/annotation/<file_annotation_id>/`
 * `webclient/get_original_file/<original_file_id>/`
 * `webgateway/archived_files/download/<image_id>/`
 * `webgateway/archived_files/download/`
 * `webgateway/archived_files/download/?image=<image_id>`
 * `webgateway/archived_files/download/?well=<well_id>`

When objects are missing HTTP 404s should be returned. When parameters are missing or incorrect HTTP 400s should be returned.

3. expected observations

No socket, pipe, or filehandle leakage.

# Related reading

 * https://trello.com/c/UYsZa42J/478-bug-omeroweb-open-files-leak-not-being-cleaned-up
 * #5699

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -2902,10 +2902,10 @@ def get_original_file(request, fileId, download=False, conn=None, **kwargs):
 @login_required(doConnectionCleanup=False)
 def download_annotation(request, annId, conn=None, **kwargs):
     """ Returns the file annotation as an http response for download """
-    ann = conn.getObject("Annotation", annId)
+    ann = conn.getObject("FileAnnotation", annId)
     if ann is None:
         rsp = ConnCleaningHttpResponse(StringIO(
-            "Annotation does not exist (id:%s)." % (annId)), status=404)
+            "FileAnnotation does not exist (id:%s)." % (annId)), status=404)
         rsp.conn = conn
         return rsp
 
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -34,6 +34,7 @@ import re
 import sys
 import warnings
 
+from StringIO import StringIO
 from time import time
 
 from omero_version import build_year
@@ -2877,8 +2878,10 @@ def get_original_file(request, fileId, download=False, conn=None, **kwargs):
 
     orig_file = conn.getObject("OriginalFile", fileId)
     if orig_file is None:
-        return handlerInternalError(
-            request, "Original File does not exists (id:%s)." % (fileId))
+        rsp = ConnCleaningHttpResponse(StringIO(
+            "Original File does not exist (id:%s)." % (fileId)), status=404)
+        rsp.conn = conn
+        return rsp
 
     rsp = ConnCleaningHttpResponse(
         orig_file.getFileInChunks(buf=settings.CHUNK_SIZE))
@@ -2901,8 +2904,10 @@ def download_annotation(request, annId, conn=None, **kwargs):
     """ Returns the file annotation as an http response for download """
     ann = conn.getObject("Annotation", annId)
     if ann is None:
-        return handlerInternalError(
-            request, "Annotation does not exist (id:%s)." % (annId))
+        rsp = ConnCleaningHttpResponse(StringIO(
+            "Annotation does not exist (id:%s)." % (annId)), status=404)
+        rsp.conn = conn
+        return rsp
 
     rsp = ConnCleaningHttpResponse(
         ann.getFileInChunks(buf=settings.CHUNK_SIZE))
--- a/components/tools/OmeroWeb/omeroweb/webgateway/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webgateway/views.py
@@ -2412,9 +2412,11 @@ def archived_files(request, iid=None, conn=None, **kwargs):
     wellIds = request.GET.getlist('well')
     if iid is None:
         if len(imgIds) == 0 and len(wellIds) == 0:
-            return HttpResponseServerError(
+            rsp = ConnCleaningHttpResponse(StringIO(
                 "No images or wells specified in request."
-                " Use ?image=123 or ?well=123")
+                " Use ?image=123 or ?well=123"), status=400)
+            rsp.conn = conn
+            return rsp
     else:
         imgIds = [iid]
 
@@ -2431,17 +2433,20 @@ def archived_files(request, iid=None, conn=None, **kwargs):
         for w in wells:
             images.append(w.getWellSample(index).image())
     if len(images) == 0:
-        logger.debug(
-            "Cannot download archived file becuase Images not found.")
-        return HttpResponseServerError(
-            "Cannot download archived file because Images not found (ids:"
-            " %s)." % (imgIds))
+        message = 'Cannot download archived file because Images not ' \
+            'found (ids: %s)' % (imgIds)
+        logger.debug(message)
+        rsp = ConnCleaningHttpResponse(StringIO(message), status=404)
+        rsp.conn = conn
+        return rsp
 
     # Test permissions on images and weels
     for ob in (wells):
         if hasattr(ob, 'canDownload'):
             if not ob.canDownload():
-                raise Http404
+                rsp = ConnCleaningHttpResponse(status=404)
+                rsp.conn = conn
+                return rsp
 
     for ob in (images):
         well = None
@@ -2450,12 +2455,15 @@ def archived_files(request, iid=None, conn=None, **kwargs):
         except:
             if hasattr(ob, 'canDownload'):
                 if not ob.canDownload():
-                    raise Http404
+                    rsp = ConnCleaningHttpResponse(status=404)
+                    rsp.conn = conn
         else:
             if well and isinstance(well, omero.gateway.WellWrapper):
                 if hasattr(well, 'canDownload'):
                     if not well.canDownload():
-                        raise Http404
+                        rsp = ConnCleaningHttpResponse(status=404)
+                        rsp.conn = conn
+                        return rsp
 
     # make list of all files, removing duplicates
     fileMap = {}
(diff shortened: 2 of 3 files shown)


kennethgillen:
Nice, thanks @chris-allan!

kennethgillen:
Is it enough to apply this diff to a release webclient at `/opt/omero/web/OMERO.web/lib/python/omeroweb/decorators.py`?

I can't replicate the original behaviour with `ab` on a local vagrant, but can with 

```
 for (( ann=1; ann<=1000; ann++)); do echo "http://127.0.0.1:8080/webclient/annotation/$ann" | tee -a ann_log_localhost.txt && curl -sL http://127.0.0.1:8080/webclient/annotation/$ann 2>&1 | tee -a ann_log_localhost.txt | grep "does not exist"; done;
```

Not sure if that's related to being on vagrant.

Anyway, with that diff applied I still get the FIFO (and total) climbing way way beyond the requests to `/webclient/`

Possibly just applying the diff to the file and restarting web is not sufficient?

chris-allan:
Applying the diff and restarting web should absolutely be sufficient. It could be this is down to not actually completing the download and the `close()` not being called.

The code literally does exactly what the decorators do so if we are having leakage then we'll have to find out what the specific conditions are. In particular, if it's related to the public user.

Since your `curl` executions have no cookie supplied I'd assume you have the public user enabled on that vagrant, @kennethgillen?

kennethgillen:
@chris-allan - Yes, public user is enabled. I'm using the  https://github.com/ome/ansible-examples-omero/tree/master/public-user

Steps to reproduce the environment are at: https://kennethgillen.github.io/ome-ansible-training/04-ome-public-example/

It'd be nice to use an easier-to-reproduce docker workflow - not sure if we have a public-user docker image kicking around.

And here's what I'm watching:

```
[vagrant@localhost ~]$ cat web-lsof.bash
 lsof -u omero-web | awk '{ print $5 }' | sort |uniq -c |sort -n | grep -v TYPE | awk '{sum += $1} {print $2"="$1} END {print "total=" sum}'
```

manics:
@chris-allan I suspect the problem is requesting non-existent annotations with the public user leading to some unclosed threads, I've been testing on WEB-DEV-merge and managed to kill the server
```
$ sudo ps -efT |grep python | wc -l
4071
```
It's been more than 10 minutes so any sessions should've timed out by now

chris-allan:
A ha! That would make complete sense. The connection closing is disabled at the decorator level and `close()` on the request is only going to be called if the annotation actually exists. I'll look into that.

chris-allan:
a32bc951b4c9a6bcc5858b2a946856c796e83bff has some pretty significant changes and expands the testing scope quite dramatically. I've added the relevant items to the PR description.

I don't think any of the HTTP status code changes are controversial. Using HTTP 500s when objects just can't be found or arguments are incorrect doesn't feel right. Also, I don't think there's good justification for leaking stacktraces to clients when we have an error during multi-file handling. The stack trace will end up in the log file where the administrator can attempt to act accordingly.

kennethgillen:
@chris-allan - is it still acceptable to change the files on local filesystem and restart web?

Replacing the views.py files in:
 /opt/omero/web/OMERO.web/lib/python/omeroweb/webgateway
 /opt/omero/web/OMERO.web/lib/python/omeroweb/webclient

and restarting web gives me:

```
Exception Value: | No module named webclient_utils
/opt/omero/web/OMERO.web/lib/python/omeroweb/api/views.py in <module>, line 61
``

**Comment by chris-allan** [comment]
Yes, that's always viable.

I have no idea why you're getting that exception. The file nor the module you're getting the exception on are related to the changes here.

**Comment by kennethgillen** [comment]
Cheers - probably replaced the wrong file locally. Sorry for the hassle. I've downloaded `/api/views.py` as well, and it's happier.

**Comment by kennethgillen** [comment]
@chris-allan 

The responses to the `webgateway/archived_files/download/?well=` endpoint looks wrong:

Specifically no `id` is parsed, and response states "Images" not "Well".

```
2018-08-02 15:06 LS27172:lsof kenny$ head -n 3 get_archived_files-download-querystring-image.txt
http://127.0.0.1:8080/webgateway/archived_files/download/?image=0
Cannot download archived file because Images not found (ids: [u'0'])
http://127.0.0.1:8080/webgateway/archived_files/download/?image=1
Cannot download archived file because Images not found (ids: [u'1'])

2018-08-02 15:05 LS27172:lsof kenny$ head -n 3 get_archived_files-download-querystring-well.txt
http://127.0.0.1:8080/webgateway/archived_files/download/?well=0
Cannot download archived file because Images not found (ids: [])
http://127.0.0.1:8080/webgateway/archived_files/download/?well=1
Cannot download archived file because Images not found (ids: [])
```

**Comment by chris-allan** [comment]
That's in keeping with the existing implementation. The length check only happens for the aggregate number of Images not Wells. I wasn't about to mess with that and start changing the scope of the error messages.

**Comment by kennethgillen** [comment]
@chris-allan - Sorry, trying to grok that - does that mean the endpoint `webgateway/archived_files/download/?well=<well_id>` doesn't work even with valid well IDs?

**Comment by chris-allan** [comment]
The logic works as expected for Wells and Images. The check for whether anything was loaded always happens at the Image level, as that is what the Filesets are linked to, so that is what is reported in the error message.

**Comment by kennethgillen** [comment]
Public user setup, some IDs valid, some not. 

Each endpoint iterated from `0`-`1000`.

LSOF all stable at +- 100 and fluctuating as expected during testing, but at end of test run close to levels at start of testing at all times. 

Endpoints tested, with summary of return codes:

- webclient/annotation/<file_annotation_id>/
```
./webclient-annotation-n-.txt
   2 < HTTP/1.1 200 OK
 998 < HTTP/1.1 404 NOT FOUND
   1 < HTTP/1.1 500 INTERNAL SERVER ERROR
```

- webclient/get_original_file/<original_file_id>/
```
./webclient-get_original_file-n-.txt
  86 < HTTP/1.1 200 OK
 887 < HTTP/1.1 404 NOT FOUND
  28 < HTTP/1.1 500 Internal Server Error
```

- webgateway/archived_files/download/<image_id>/
```
./webgateway-archived_files-download-n-.txt
  13 < HTTP/1.1 200 OK
 988 < HTTP/1.1 404 NOT FOUND
```

- webgateway/archived_files/download/
```
./webgateway-archived_files-download-.txt
1001 < HTTP/1.1 400 BAD REQUEST
```

- webgateway/archived_files/download/?image=<image_id>
```
./webgateway-archived_files-download--image-n.txt
  13 < HTTP/1.1 200 OK
 988 < HTTP/1.1 404 NOT FOUND
```

- webgateway/archived_files/download/?well=<well_id>
```
./webgateway-archived_files-download--well-n.txt
 1001 < HTTP/1.1 404 NOT FOUND
```

Great work, thanks @chris-allan

File: components/tools/OmeroWeb/omeroweb/webclient/views.py [hunk 1]
```diff
@@ -2877,8 +2878,10 @@ def get_original_file(request, fileId, download=False, conn=None, **kwargs):
 
     orig_file = conn.getObject("OriginalFile", fileId)
     if orig_file is None:
-        return handlerInternalError(
-            request, "Original File does not exists (id:%s)." % (fileId))
+        rsp = ConnCleaningHttpResponse(StringIO(
```
will-moore, line 2881: Sorry - coming to this a bit late...
As I understood it, we're using ```ConnCleaningHttpResponse``` in places where we don't want to call ```conn.close()``` until the response is closed (because we're downloading a file etc), but in these cases the response is not downloading anything so we could simply do this?
```
conn.close(hard=False)
return HttpResponseNotFound("Original File does not exist (id:%s)." % (fileId))
```

components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -2901,8 +2904,10 @@ def download_annotation(request, annId, conn=None, **kwargs):
     """ Returns the file annotation as an http response for download """
     ann = conn.getObject("Annotation", annId)
```
will-moore: You probably want to include #5816 here. I think that is trying to fix errors coming from testing this PR?
This line will try to find *any* type of annotation (not just "FileAnnotation") but if you try to get any other type of annotation then ```ann.getFileInChunks()``` subsequently fails (raises 500) and you will not have ```conn.close()``` called.

components/tools/OmeroWeb/omeroweb/webclient/views.py
chris-allan: Yes we could.

I was preferring the usage of the same response type and keeping all the close logic in one place.

will-moore: OK, cool. I was really just checking that I understood the logic.

will-moore:
Code looks good and @kennethgillen is happy with functionality so I think this is good to merge.

components/tools/OmeroWeb/omeroweb/webgateway/views.py
```diff
@@ -2431,17 +2433,20 @@ def archived_files(request, iid=None, conn=None, **kwargs):
...
             images.append(w.getWellSample(index).image())
     if len(images) == 0:
-        logger.debug(
-            "Cannot download archived file becuase Images not found.")
-        return HttpResponseServerError(
-            "Cannot download archived file because Images not found (ids:"
-            " %s)." % (imgIds))
+        message = 'Cannot download archived file because Images not ' \
+            'found (ids: %s)' % (imgIds)
+        logger.debug(message)
```
joshmoore, line 2438: Is this `logger.debug` intentional?

will-moore: I think @chris-allan just wanted to keep the logging the same here.

joshmoore: Sorry, makes sense.

PR #5817, opened 2018-08-13
ChannelProcessor: do not change color to white if Bio-Formats set a color

melissalinkert:
# What this PR does

Channels with no emission wavelength data are typically candidates for
having their colors reset to white.  This prevents the color from being
reset if the Bio-Formats reader supplied a color via ```MetadataStore```.

# Testing this PR

Using 5.4.7 or a develop build without this change, start the server and import ```data_repo/curated/nd2/jose/08_SO4_Fasc100_at_surface_w_average.nd2```.  Querying the database or opening the imported image in web should show that the channel colors are green (16711935), white (-1) and white (-1), in that order.  This differs from the channel colors reported by Bio-Formats, which are green (16711935), red (-16776961), and white (-1) as can be seen with ```showinf -nopix -omexml``` or by opening in ImageJ with color mode set to ```Composite``` or ```Colorized```.

Using an OMERO build that includes this PR, import the file again and confirm that the colors are now green, red, and white.

To verify that the channel color is still set to white when neither emission wavelength nor channel color is set by Bio-Formats, import ```data_repo/curated/fv1000/aryeh/1h.oib``` with and without this PR.  This file has 3 channels, none of which have a color and the last of which does not have an emission wavelength.  There should be no difference in channel colors with this PR; in particular, the last channel should be white.

# Related reading

The logic to reset a channel color to white was introduced in openmicroscopy/openmicroscopy@6a324bf.  That commit is approximately 9 years old (pre-4.4.x), and dates from a time when trusting the colors supplied by Bio-Formats was not realistic and colors were rarely supplied in any case.  We now have a little more confidence in Bio-Formats (at least I hope so!), so defaulting to whatever color is set by the reader makes more sense now.

### DIFF
--- a/components/blitz/src/ome/formats/model/ChannelProcessor.java
+++ b/components/blitz/src/ome/formats/model/ChannelProcessor.java
@@ -106,13 +106,9 @@ public class ChannelProcessor implements ModelProcessor {
   private void setSingleChannel(ChannelData channelData) {
     int channelIndex = channelData.getChannelIndex();
     Channel channel = channelData.getChannel();
-    Integer red = getValue(channel.getRed());
-    Integer green = getValue(channel.getGreen());
-    Integer blue = getValue(channel.getBlue());
-    Integer alpha = getValue(channel.getAlpha());
     RString name;
-    //color already set by Bio-formats
-    if (red != null && green != null && blue != null && alpha != null) {
+    if (hasColor(channel)) {
+      //color already set by Bio-formats
       return;
     }
     int[] defaultColor = ColorsFactory.newGreyColor();
@@ -163,13 +159,9 @@ public class ChannelProcessor implements ModelProcessor {
       return;
     }
 
-    Integer red = getValue(channel.getRed());
-    Integer green = getValue(channel.getGreen());
-    Integer blue = getValue(channel.getBlue());
-    Integer alpha = getValue(channel.getAlpha());
     RString name;
-    //color already set by Bio-formats
-    if (red != null && green != null && blue != null && alpha != null) {
+    if (hasColor(channel)) {
+      //color already set by Bio-formats
       //Try to set the name.
       log.debug("Already set in BF.");
       if (lc.getName() == null) {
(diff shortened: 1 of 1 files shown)


joshmoore:
**Client** --> **Server**
Merge --> Merge
Merge --> Latest
Latest --> Latest

all showed an image with green/red/white



screen shot 2018-08-21 at 11 40 15

joshmoore:
```
nd2/before: green/white/white
nd2/after:  green/red/white
oib/before: green/red/white
oib/after:  green/red/white
```
:+1:

Do you have a feeling for what percentage of readers will be impacted and for those what percentage of filesets will be impacted?

melissalinkert:
Should be limited to readers that set ```Channel.Color``` and at least one of ```Channel.EmissionWavelength``` and ```LightPath.EmissionFilterRef``` (referencing a valid filter with ```CutIn``` set per https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/ome/formats/model/ColorsFactory.java#L144).  At the moment, that should just be:

```
BaseZeissReader
ColumbusReader
ImarisHDFReader
LeicaReader
LIFReader
NativeND2Reader
OIRReader
ZeissCZIReader
ZeissLSMReader
```

i.e. < 10% of supported formats.  Percentage of affected filesets is harder to say, but I'd expect a small minority since it took me quite a while to find a file in our repo that actually showed the problem.

sbesson:
From my perspective, this goes in the general directions of our review of the OMERO import code to trust more of the metadata as read by Bio-Formats. The biggest risk in terms of UX is the mismatch for some datasets before and after this change. However with this code, datasets should match closer the experience from the acquisition software.

If a strong case arises for maintaining the previous behavior in some circumstances, I assume we could make the previous logic optional/configurable.

jburel:
The closer we are to the acquisition software the better the user experience/acceptance is
User could use methods like ``Apply settings to all`` to reset the previous behaviour if it is what they want.

joshmoore:
All agreed. Merging. We'll need to be careful both with the testing sheet as well with the changelog when it comes to this one.

cc: @pwalczysko

PR #5820, opened 2018-08-14
adjust ordering of Java "static" "final" modifier keywords

mtbc:
# What this PR does

Switches `final static` to `static final` in some code so it doesn't look distractingly odd.

Targets just some rather visible code and codegen; more can be done in other PRs.

# Testing this PR

Should not change behavior at all. No regressions in CI.

# Related reading

https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.1.1

### DIFF
--- a/components/blitz/resources/omero/api/IContainer.ice
+++ b/components/blitz/resources/omero/api/IContainer.ice
@@ -328,7 +328,7 @@ module omero {
                  *
                  * @param obj Can be <i>unloaded</i>.
                  * @param collectionName
-                 *            <code>public final static String</code> from the
+                 *            <code>public static final String</code> from the
                  *            IObject.class
                  * @param options
                  *            Parameters. Unused.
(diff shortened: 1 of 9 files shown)


joshmoore:
> then it is customary, though not required, that they appear in the order consistent with that shown above in the production for ClassModifier.

Merging, but let's discuss how many of these suggestions we're going to implement. I think the potential for conflicts may outweigh the benefit. (Happy to see a linter for new code/repositories though!)

mtbc:
Yes, rather than an all-at-once bulk change over a couple hundred files, this is a gradual effort done mindfully of other work: e.g., here I'm not touching anything remotely related to the import epic and not many of us are working on the Java code at one time anyway. I figure a series of PRs over time is the safest then adding something to Travis to prevent regressions. At least any conflicts that do still arise should be trivial to fix.

PR #5822, opened 2018-08-15
make data provider builder available to integration tests

mtbc:
# What this PR does

Exposes #5537's `DataProviderBuilder` and moves it to `components/common/test/` so that it can be used from `components/tools/OmeroJava/test/integration/`.

# Testing this PR

Travis should pass (i.e. `GraphPolicyRuleTest` unit test). I can also point to other code that uses this class.

# Related reading

https://trello.com/c/B9feHhY6/465-expose-dataproviderbuilder

### DIFF
--- a/components/common/test/ome/testing/DataProviderBuilder.java
+++ b/components/common/test/ome/testing/DataProviderBuilder.java
@@ -17,7 +17,7 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-package ome.services.graphs;
+package ome.testing;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -34,7 +34,7 @@ import org.testng.annotations.DataProvider;
  * @author m.t.b.carroll@dundee.ac.uk
  * @since 5.4.1
  */
-class DataProviderBuilder {
+public class DataProviderBuilder {
 
     private List<List<Object>> args = Collections.singletonList(Collections.emptyList());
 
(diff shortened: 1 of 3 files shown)


joshmoore:
In general, :+1: for getting making this and other testing code available in our jars. I'm slightly concerned about use of `ome.` since we typically try to avoid it's usage on the client side, though that perhaps doesn't apply here. In Python land, this would be `omero.testlib`.

mtbc:
We're also using it for unit tests of server code.

dominikl:
If I create the Eclipse project with the default `./build.py build-dev` command I can't use any `ome.testing.*` imports in the omero-server project, they are all flagged as "cannot be resolved". Does that still need to be added to the ivy config or something like that? Apart from that it looks fine.

mtbc:
Hmm, I've not had to do anything special in my Eclipse.

joshmoore:
> In Python land, this would be omero.testlib.

I take this back since it's used earlier in the graph than I thought. `omero.testlib` may still come.

> If I create the Eclipse project with the default ./build.py build-dev

I ran `build-dev` and the top-level project was fine, but I did see @dominikl's issue for `components/server` on its own:

screen shot 2018-08-20 at 17 31 09

mtbc:
With Oxygen.3a:

imports

> on its own

but maybe this is key? I have OMERO open as one big "ome.git" project but maybe you can close individual components within it? `components/server/test.xml` already lists a dependency on `common-test` though.

joshmoore:
@dominikl : is it possible that the individual `omero-server` component on its own was broken independently of this PR?

(Sorry, @dom whoever you are for the noise)

mtbc:
There's certainly plenty else in `components/server/test/ome/server/` that uses classes from `components/common/test/`. How does, say, `AbstractManagedContextTest` look in your Eclipse?

dominikl:
Oh yes, I don't think that is caused by this PR, sorry.

joshmoore:
> Oh yes, I don't think that is caused by this PR, sorry.

Gotcha. We'll handle additional Eclipse wonkiness later. Merging.

PR #5823, opened 2018-08-15
getFileName_return_string

will-moore:
# What this PR does

This fixes Blitz FileAnnotation.getFileName() for when there is no name or empty string.
See https://trello.com/c/18d8k6F1/7-fileanngetfilename

# Testing this PR

1. Test should be green
1. Web adding file annotation dialog: files with empty string Name should show "No name. ID 123"
screen shot 2018-09-03 at 12 42 31

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5207,7 +5207,7 @@ class FileAnnotationWrapper (AnnotationWrapper, OmeroRestrictionWrapper):
         fpath = f.getPath()
         if fpath is not None and len(fpath) > 0:
             return fpath
-        return f.id
+        return str(f.id)
 
     def getFileInChunks(self, buf=2621440):
         """
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -5201,13 +5201,7 @@ class FileAnnotationWrapper (AnnotationWrapper, OmeroRestrictionWrapper):
         f = self.getFile()
         if f is None or f._obj is None:
             return None
-        fname = f.getName()
-        if fname is not None and len(fname) > 0:
-            return fname
-        fpath = f.getPath()
-        if fpath is not None and len(fpath) > 0:
-            return fpath
-        return "No name (id:%s)" % f.id
+        return f.getName()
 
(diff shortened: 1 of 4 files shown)


will-moore:
@mtbc The error on the trello card was caused by returning a different type, so that's what this is trying to fix.

mtbc:
I wondered if the bug isn't that the gateway gives a bare ID but that `custom_forms.py` fails to cope when it does.

*Update:* Ah, but I guess that breaks the `:rtype:     String` contract, fair enough.

will-moore:
Test now passing at https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/936/testReport/junit/OmeroPy.test.integration.gatewaytest/test_annotation/testFileAnnotationNoName/

joshmoore:
If anyone's interested, see the additional conversation going on under https://github.com/openmicroscopy/openmicroscopy/commit/c5739a44db8c54f3971c632c1dc25e116c337978#comments

Short summary from my side is I find it difficult to know how we got into this situation so making safe suggestions is difficult. The only thing that occurs to me (since apparently throwing an exception if there's no name isn't appropriate) would be to return something more clearly _not_ the name of the annotation: `"No name (id:%s)" % f.id` but I don't know if that makes the situation substantially better.

manics:
Couldn't figure out how to test this, but `"No name (id:%s)` seems to be mixing a UI message with the model, e.g. Imagine if we added internationalization here. I think returning `str(id)`, or `''`, or `OriginalFile:id` would be better.

joshmoore:
How do we want to move forward here?

will-moore:
Apologies, my initial handling of this may have been a bit too simplistic (casting the ID to string) when it's not the job of the BlitzGateway to do that.
If the file name is an empty string then BlitzGateway should simply return that and if we want to improve how that's shown in the UI then that's webclient's job.
The 2 commits above follow that strategy.

will-moore:
Also added a web test to test that we don't get the original error now.

joshmoore:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_annotate/TestFileAnnotations/

:+1: Cheers @will-moore

joshmoore:
For reference see f6923b1a3f0b8129d9b30aec5eb38276be038c53 from gh-189

PR #5825, opened 2018-08-16
calculate StatsInfo when missing even for not-big images

mtbc:
# What this PR does

If [`StatsInfo`](https://docs.openmicroscopy.org/latest/omero5.4/developers/Model/EveryObject.html#statsinfo) is not available for an image then create it before attempting to render the image. This means that global min-max is used for default rendering settings instead of a hopeful guess at intensity range.

# Testing this PR

Any OMERO 5.4 clients should suffice for imports: this PR changes the server only.

Unzip [test-image.ome.xml.gz](https://github.com/openmicroscopy/openmicroscopy/files/2293788/test-image.ome.xml.gz) and import it into both eel latest and eel merge. Notice how it has five planes with the darkest being the middle one. This is a correct rendering of the image.

Now import it into both again but using,
```
bin/omero import -- --no-stats-info --no-thumbnails test-image.ome.xml
```
* On eel latest, without this PR, you will see that the planes all look the same because misleading rendering settings were chosen.
* On eel merge, with this PR, you will see a clock thumbnail. Refresh and it will become the image. When you view it then it should be rendered correctly.

Omitting `--no-thumbnails` gets the clock-fixing process moving on import rather than first waiting for viewing.

Is the correctness worth the delay?

### DIFF
--- a/components/romio/src/ome/io/nio/PixelsService.java
+++ b/components/romio/src/ome/io/nio/PixelsService.java
@@ -585,7 +585,7 @@ public class PixelsService extends AbstractFileSystemService
         }
         // Note: since the OMERO.fs work, a pixels pyramid is only required
         // when the pixels set meets big image criteria.
-        if (!pixelsFileExists && requirePyramid && originalFilePath != null)
+        if (!pixelsFileExists && originalFilePath != null)
         {
             handleMissingStatsInfo(pixels);
         }


mtbc:
If we still want to avoid `StatsInfo` calculation when rendering settings are already provided then perhaps the work of `handleMissingStatsInfo` could instead be done in `ThumbnailBean._resetDefaults`.

*Update:* Doesn't look as if it would be useful. Would need something more like `RenderingSettingsImpl.resetChannelBindings` publishing `MissingStatsInfoMessage` and somehow failing back toward a clock.

sbesson:
Main immediate thought was: would we still need a way to force min/max calculation to be skipped e.g. in the case where rendering settings have been calculated externally and are meant to be applied post import via `omero-cli-render`.

For the scope of the investigation, in addition to the sample file provided, I would also suggest testing a couple of fake and real large multidimensional data where the unconditional calculation the stats info should give an idea of the created delays e.g.

- a fake file e.g. `test&sizeX=3000&&sizeY=3000&sizeZ=500&sizeC=3&sizeT=1000.fake`
- `curated/tiff/pavel/Dmel/full.pattern` (use `--transfer=ln_s and `--skip=checksum` as this is a 3600 files fileset)

components/romio/src/ome/io/nio/PixelsService.java
```diff
@@ -585,7 +585,7 @@ public PixelBuffer _getPixelBuffer(Pixels pixels, boolean write)
         }
         // Note: since the OMERO.fs work, a pixels pyramid is only required
         // when the pixels set meets big image criteria.
```
jburel, line 587: The comment will need to be adjusted

mtbc:
> in the case where rendering settings have been calculated externally

Yes, I'd say so.

mtbc:
Publishing a `MissingStatsInfoMessage` *instead of* creating rendering settings looks to be a way to go. HQL to find pixels without the stats info is:
```sql
SELECT DISTINCT pixels.id FROM Channel WHERE statsInfo IS NULL AND pixels.id IN (:ids)
```

PR #5829, opened 2018-08-22
deprecate importer option useFullPath / userFullPath

mtbc:
Deprecates unused importer options `useFullPath` and `userFullPath`. They are ignored by our code and not mentioned in our documentation.


dominikl:
Can't find any location where this option is used either. 👍

PR #5831, opened 2018-08-22
Populate metadata deprecation

sbesson:
This deprecates the metadata modules distributed as part of OMERO.py. The content of these classes  has been migrated to https://github.com/ome/omero-metadata available from PyPI and future updates/fixes/improvements will be carried out there.

Upstream tests should remain functional and passing but any usage of the modules e.g. when using the CLI plugin via

```
OMERO_DEV_PLUGINS=true bin/omero metadata -h
```

should display the deprecation warning.

See also discussion at https://github.com/ome/omero-parade/issues/47#issuecomment-414400502

### DIFF
--- a/components/tools/OmeroPy/src/omero/util/populate_metadata.py
+++ b/components/tools/OmeroPy/src/omero/util/populate_metadata.py
@@ -33,6 +33,7 @@ from getpass import getpass
 from getopt import getopt, GetoptError
 from itertools import izip
 from collections import defaultdict
+import warnings
 
 import omero.clients
 from omero import CmdError
@@ -59,6 +60,11 @@ from populate_roi import ThreadPool
 
 log = logging.getLogger("omero.util.populate_metadata")
 
+warnings.warn(
+    "This module is deprecated as of OMERO 5.4.8. Use the module"
+    " in the omero-metadata project available from"
+    " https://pypi.org/project/omero-metadata/ instead.",
+    DeprecationWarning)
 
 def usage(error):
     """Prints usage so that we don't have to. :)"""
--- a/components/tools/OmeroPy/src/omero/util/populate_metadata.py
+++ b/components/tools/OmeroPy/src/omero/util/populate_metadata.py
@@ -66,6 +66,7 @@ warnings.warn(
     " https://pypi.org/project/omero-metadata/ instead.",
     DeprecationWarning)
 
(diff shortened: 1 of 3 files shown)


manics (approved):
The deprecation warning is printed twice since the warning is in two places, but that's not a big deal.
👍

components/tools/OmeroPy/src/omero/util/populate_metadata.py
```diff
@@ -59,6 +60,12 @@
 
 log = logging.getLogger("omero.util.populate_metadata")
 
+warnings.warn(
+    "This module is deprecated as of OMERO 5.4.8. Use the module"
+    " in the omero-metadata project available from"
+    " https://pypi.org/project/omero-metadata/ instead.",
+    DeprecationWarning)
```
sbesson, line 67: While testing an import, I just realized that this warning is displayed on every import of `omero.util.populate_metadata`. Concretely that implies it will show up every invocation of the CLI since the first step is to load the plugins even if they are not registered.

Although harmless, my feeling is that this is fairly annoying in terms of user experience. I will look at moving the deprecation warning under the main classes like `ParsingContext` instead so that the warning happens at run-time and does not create unnecessary noise.

sbesson: Ponderating the comment above, as per https://docs.python.org/2/library/warnings.html#warning-categories, `DeprecationWarning is  ignored by default` on Python 2.7. If we still wanted to reduce the CLI  noise on Python 2.6, https://docs.python.org/2/library/warnings.html#available-context-managers would certainly be a way to silence these warnings.

mtbc: I doubt that Python 2.6 is worth any avoidable effort at all.

joshmoore: Perhaps add this:
```
$ grep warnings bin/omero
import warnings
warnings.filterwarnings("ignore", category=DeprecationWarning)
```

joshmoore: Pushed.

sbesson: Thank you

PR #5832, opened 2018-08-22
Populate dataset webclient

sbesson:
## Summary 

Previously opened as https://github.com/openmicroscopy/openmicroscopy/pull/5691, but the metadata population logic has now been filtered out of OMERO.py into an independent component which can be released in isolation (see #5831).

This change cherry-picks the changes at the Webclient level for displaying Tables on P/D/I hierarchy. Minimally, when clicking on the Tables tab on an image, two additional queries are performed to load the bulk annotations at either the Project and/or at the Dataset level.

## Testing

Two IDR studies have been imported under the public user of https://web-proxy.openmicroscopy.org/east-web/webclient/: the experimentA of [idr0021](https://idr.openmicroscopy.org/webclient/?show=project-51) and the screenB of [idr0008](https://idr.openmicroscopy.org/webclient/?show=screen-206).

For each study, the bulk annotations have been populated using the annotation.csv files under the metadata repositories ([idr0021](https://github.com/IDR/idr0021-lawo-pericentriolarmaterial) and [idr0008](https://github.com/IDR/idr0008-rohn-actinome)) and [omero-metadata:0.2.2](https://pypi.org/project/omero-metadata/0.2.2/)

Tables should now load when expanding the corresponding tab in the right-hand panel for:
- Images under the `idr0021` project
- Wells under the `idr0008` screen (current behavior)

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -351,7 +351,7 @@
                     var plateQuery = "{% url 'webgateway_object_table_query' 'Plate.wells.wellSamples.image' manager.image.id %}";
                     var projectQuery = "{% url 'webgateway_object_table_query' 'Project.datasetLinks.child.imageLinks.child' manager.image.id %}";
                     var datasetQuery = "{% url 'webgateway_object_table_query' 'Dataset.imageLinks.child' manager.image.id %}";
-                    var query = "image-{{ manager.image.id }}";
+                    var query = "Image-{{ manager.image.id }}";
                     {% endif %}
 
                     var showBulkAnnTooltip = function(data) {
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -349,6 +349,8 @@
                     {% elif manager.image %}
                     var screenQuery = "{% url 'webgateway_object_table_query' 'Screen.plateLinks.child.wells.wellSamples.image' manager.image.id %}";
                     var plateQuery = "{% url 'webgateway_object_table_query' 'Plate.wells.wellSamples.image' manager.image.id %}";
+                    var projectQuery = "{% url 'webgateway_object_table_query' 'Project.datasetLinks.child.imageLinks.child.image' manager.image.id %}";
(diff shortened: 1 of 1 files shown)


sbesson:
Last two commits should contain a few additional fixes

- 3c0632c modifies the image query to use lowercase `image-`. This matches the current behavior of the metadata population tool e.g. see 
https://idr.openmicroscopy.org/webgateway/table/Project.datasetLinks.child.imageLinks.child/1884807/query/?query=image-1884807 vs https://idr.openmicroscopy.org/webgateway/table/Project.datasetLinks.child.imageLinks.child/1884807/query/?query=Image-1884807. The original behavior was implemented in https://github.com/openmicroscopy/openmicroscopy/pull/4929 so I am unclear whether the `Image-{{ manager.image.id }}` case should also be supported
- 84eae09 should prevent an error message about undefined `projectQuery` variable in the developer tools when trying to open the Tables tab on a well.

will-moore:
Working fine for Project and Screen on east-web as described.
Tried locally with a Table prepared for OMERO.parade and this isn't working as parade is using capitalised ```Image``` column (https://github.com/ome/omero-parade/blob/master/omero_parade/table_filters/data_providers.py#L101) instead of lowercase ```image``` column as expected.

Since we are using query="Well" elsewhere and WellColumns are called "Well", it is more consistent to use "Image" for ImageColumns.

sbesson:
No personal objection to unifying towards `Image-` instead of `image-`. The consistency argument holds completely. Major concrete points to consider with regard to any long-term decisions:


- we are currently looking at `omero-metadata`, `omero-parade` and `omerowebclient`. Are there other consumers relying on this semantics?
- we should assume that current production instances like IDR already have data populated with an `image` column. What is the expectation for those? Are we forcing a systematic re-annotation in order to benefit from the Webclient and Parade integration? An intermediate might be to support both APIs in the meantime i.e. handle both `image-` and `Image-` queries but deprecate the former.
- independently of the path, we need to update our metadata population tool to comply with the format specification - see https://github.com/ome/omero-metadata/issues/9

sbesson:
In order to test https://github.com/ome/omero-metadata/pull/5 and in absence of concrete discussion and agreement, 572c319 implements the intermediate solution discussed in https://github.com/openmicroscopy/openmicroscopy/pull/5832#issuecomment-416212771 i.e. acknowledges the fact Project/Dataset tabular data currently exists under two forms and handle both variations.

joshmoore:
@will-moore: is this good to go?

will-moore:
Yes, this is working for me.
I guess the only question is how long we keep polling for ```?query=image...``` and ```?query=Image...```.

joshmoore:
Thanks, @will-moore. I'll leave you to engage that conversation, creating cards as necessary.

PR #5833, opened 2018-08-23
add %thread% expansion option for omero.fs.repo.path

mtbc:
# What this PR does

Adds a new `%thread%` option for setting in `omero.fs.repo.path`.

# Testing this PR

Use `bin/omero config set omero.fs.repo.path ...` to set the path to other than the default of `%user%_%userId%//%year%-%month%/%day%/%time%`. Include `%thread%` in that new path, restart the server, do concurrent imports and see how the thread is reflected in the paths created in `ManagedRepository/`.

# Related reading

https://docs.openmicroscopy.org/omero/5.4.7/sysadmins/fs-upload-configuration.html#template-path (does not yet include this)
https://trello.com/c/OML4OoYE/232-bug-race-condition-on-import

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
@@ -1075,6 +1075,18 @@ public class ManagedRepositoryI extends PublicRepositoryI
         }
 
         /**
+         * Expand {@code %thread%} to the name of the current thread.
+         * @param prefix path component text preceding the expansion term, may be empty
+         * @param suffix path component text following the expansion term, may be empty
+         * @return entire replaced path component, may be unchanged to be revisited,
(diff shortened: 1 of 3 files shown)


dominikl:
Looks good. Tested locally with `%user%_%userId%//%year%-%month%/%day%/%thread%-%time%`
Result:
```
...
Blitz-0-Ice.ThreadPool.Server-0-14-39-11.780
Blitz-0-Ice.ThreadPool.Server-0-14-39-11.780.log
Blitz-0-Ice.ThreadPool.Server-1-14-35-13.241
Blitz-0-Ice.ThreadPool.Server-1-14-35-13.241.log
...
```

There's still the `%time%` weirdness, ie. `%thread%-%time%` works but `%time%-%thread%` doesn't (`%thread%` doesn't get substituted). **Not related to this PR**.

PR #5835, opened 2018-08-24
for import fix prefix template path expansion bug

mtbc:
# What this PR does

Fix the bug that template path terms are not expanded if they follow `%time%`, `%increment%`, `%subdirs%` in that same path component.

# Testing this PR

Adjust `omero.fs.repo.path` to paths that have been observed not to expand properly and see where newly imported files are put.

Note that  `%time%`, `%increment%`, `%subdirs%` may not be placed together in the same path component.

# Related reading

https://docs.openmicroscopy.org/omero/5.4.7/sysadmins/fs-upload-configuration.html#template-path
https://github.com/openmicroscopy/openmicroscopy/pull/5833#issuecomment-415763603

### DIFF
--- a/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
+++ b/components/blitz/src/ome/services/blitz/repo/ManagedRepositoryI.java
@@ -1130,7 +1130,7 @@ public class ManagedRepositoryI extends PublicRepositoryI
                     /* try to expand the term */
                     final String oldPattern = pattern;
                     final boolean isTryCreateDirectory = createDirectories &&
-                            !(TEMPLATE_TERM.matcher(prefix).matches() || TEMPLATE_TERM.matcher(suffix).matches());
+                            !(TEMPLATE_TERM.matcher(prefix).find() || TEMPLATE_TERM.matcher(suffix).find());
                     while (true) {
                         try {
                             if (parameters == null) {


dominikl:
Tried various combinations of `%time%`, `%increment%` and `%subdirs%` followed by something else. `%time%` and `%increment%` works, but `%subdirs%` doesn't. 
For example: `omero.fs.repo.path=%user%_%userId%//%year%-%month%/%day%/%subdirs%-%session%` doesn't expand the `%session%`. Imported file `ACA1-1000_AF594_03_R3D.dv` -> created directory `ManagedRepository/root_0/2018-08/28/ACA1-1000_AF594_03_R3D.dv`

Edit 1:
But maybe that's a problem of `%session%`... let me test something else...

Edit 2:
No, its a problem with `%subdirs%`. `%increment%-%sessionId%` works, but `%subdirs%-%sessionId%` doesn't expand `%sessionId%`.

mtbc:
This actually fits the documentation:

> expands to nothing until the preceding directory ...

To try this one out, do something like,
```
bin/omero config set omero.fs.repo.path '%user%_%userId%//%year%-%month%/%day%/%subdirs:1%-%session%'
```
then see if it starts to take effect once there are ten things in that `%day%` directory.

mtbc:
(Please do feel to create a card if you would have preferred to have been getting `-%session%` directories created though before the subdirectories started being expanded to more than nothing.)

dominikl:
The `%subdirs%` option is a bit strange, guess I just don't really understand it. Anyway, with respect to this PR, good to merge in my opinion 👍

mtbc:
:+1: I'll look at adding some clarification to its documentation.

PR #5838, opened 2018-08-29
import: Add mmap FileTransfer

joshmoore:
In additional the parallel-upload and parallel-fileset work,
the UploadFileTransfer class itself could be made faster.

A few attempts were tested briefly (over an SSH tunnel to eel)
to see what might work.

Comparison of java.nio speeds for upload:

| *`--transfer=`* | *Seconds for 500K* | *Seconds for 5MB* | *Seconds for 70MB* |
|-----------------|--------------------|-------------------|--------------------|
| upload          | 6.02               | 111.38            | 1118.84            |
| mmap            | 6.57               | 61.26             | 957.12             |
| ~parallel~        | 10.64              | 71.38             | 1454.66            |
| ~parallel mmap~   | 5.84               | 84.39             | 1629.17            |

Based on those (highly non-scientific) numbers, this PR adds a single mmap-based
(non-parallel) implementation which can be chosen via `--transfer=mmap`. The other
two implementations are not included for the sake of simplicity.

# Testing this PR

1. Import files of various sizes, over various networks, with and without `--transfer=mmap`.
2. The expectation is that for most configuration `mmap` will be faster and that we can target this as our default behavior in the future.

# Related reading
 * https://trello.com/c/BPrkScBc/36-file-upload-should-be-faster

### DIFF
--- a/components/blitz/src/ome/formats/importer/transfers/AbstractFileTransfer.java
+++ b/components/blitz/src/ome/formats/importer/transfers/AbstractFileTransfer.java
@@ -53,7 +53,8 @@ public abstract class AbstractFileTransfer implements FileTransfer {
         cp(CopyFileTransfer.class),
         cp_rm(CopyMoveFileTransfer.class),
         upload(UploadFileTransfer.class),
-        upload_rm(UploadRmFileTransfer.class);
+        upload_rm(UploadRmFileTransfer.class),
+        mmap(MMapUploadFileTransfer.class);
         Class<?> kls;
         Transfers(Class<?> kls) {
             this.kls = kls;
--- a/components/blitz/src/ome/formats/importer/transfers/MMapUploadFileTransfer.java
+++ b/components/blitz/src/ome/formats/importer/transfers/MMapUploadFileTransfer.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (C) 2014 University of Dundee & Open Microscopy Environment.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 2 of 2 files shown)


manics:
Does this require enough memory to load the whole file, or just part of it?

joshmoore:
https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#map(java.nio.channels.FileChannel.MapMode,%20long,%20long)

> Many of the details of memory-mapped files are inherently dependent upon the underlying operating system and are therefore unspecified.  ...

:) In general, it will _try_ to load the whole file, but shouldn't tank the client in doing so. Testing this in combination with `--parallel-upload` will be telling.

joshmoore:
@manics was right, for a 10GB file:

```
2018-08-30 11:21:02,231 1976       [3-thread-1] ERROR     ome.formats.importer.cli.ErrorHandler - FILE_EXCEPTION: /OMERO/fakes/10G.fake
java.lang.IllegalArgumentException: Size exceeds Integer.MAX_VALUE
	at sun.nio.ch.FileChannelImpl.map(FileChannelImpl.java:869) ~[na:1.8.0_181]
```

mtbc:
Going by `bin/omero fs importtime --summary` I'm not getting worthwhile gains from mmap locally but it's unscientific and may depend on all kinds of things like local, FUSE, network mount? Files already cached by kernel? Etc. Certainly after the testing my eCryptfs needs to recache a bunch.

I've been experimenting with combining it with `--parallel-{upload,fileset}=<n>` options. An early finding is that if I bump `omero.db.poolsize` up then I can avoid,
```
java.sql.SQLException: unable to get a connection from pool of a PoolingDataSource containing an XAPool of resource a010e33c-00f2-4d3e-83b3-d24f7ef09f74 with 10 connection(s) (0 still available)
```

mtbc:
With `bin/omero import --parallel-{upload,fileset}=100 --transfer=mmap` of 100 × 55MB files I can cause an OOM. With 5MB upload buffers that should be 500MB of buffers but 5500MB of mmap'ing. With small images then I instead fail server-side with `Flush during cascade is dangerous` which I guess just arises from the high parallelism.

mtbc:
Yeah, with `--parallel-{upload,fileset}=60` I can import small images but mmap'ing the 55MB ones OOMs.

mtbc:
No objection to merging if:

* we make it clear that it is purely experimental
* it does materially benefit IDR workflows.

joshmoore:
Closing for the moment.

PR #5844, opened 2018-09-04
Blitzgateway close all services

dominikl:
# What this PR does

Adds a `closeServices` method to the Python Blitzgateway.

Running the `render` plugin (which uses the Blitzgateway) in parallel on a large dataset fails after while with an error indicating that too many services are left open. The gateway can't be closed because it closes the session too. A method which only closes the open services is needed. 

There might already be a similar method in the Blitzgateway, but it's basically impossible to find anything in a 10k+ lines file.

I'm currently testing this (as part of the render plugin) on test51 to generate the thumbnails for the HPA dataset.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1931,6 +1931,14 @@ class _BlitzGateway (object):
         self._proxies = NoProxies()
         logger.info("closed connection (uuid=%s)" % str(self._sessionUuid))
 
+    def closeServices(self):
+        """
+        Terminates all services (proxies) of the current session.
+        """
+        for proxy in self._proxies.values():
+            proxy.close()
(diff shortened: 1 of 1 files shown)


joshmoore:
@dominikl : does `hard=False` do what you want? https://github.com/openmicroscopy/openmicroscopy/commit/c3338ece52025c94f3617de5f33cd4d527153304#diff-e4feecea30b2ee097639cbb7c9171fed

dominikl:
Not sure what consequences the line `self._connected = False` will have, I'll check.

joshmoore:
Fair enough. I've never been overly keen on use of `self._proxies.values()` so it might make sense to have your method be the _one_ place that doesn't the service cleanup.

dominikl:
Ok, `close(hard=False)` seems to do the job too 👍 I'll close this PR again.

PR #5845, opened 2018-09-04
Added isThumbnailCached to retrieveThumbnail

rgozim:
### What this PR does

Adds `ctx.isThumbnailCached` to `retrieveThumbnail` function. `ctx.isThumbnailCached` checks if 
the rendering settings have changed for a particular image and is useful for figuring out whether to
regenerate an images thumbnail.

### Testing this PR

1. Change an images rendering settings 
2. Refresh and view the dataset with that image present
3. Check to see if the thumbnail of the image you edited, matches

PR #5846, opened 2018-09-06
Bugfixes for RC2

dominikl:
# What this PR does

Fixes some bugs noticed during RC1 testing:

- Importer: Don't request thumbnails if the "skip thumbnails" flag has been set.
- Don't crash Insight if the ThumbnailService can't be initialized for a certain image (e. g. if there is no thumbnail available for an image in a read-only group), just show black thumbnail instead.

Still outstanding:

- ~~Thumbnail doesn't get updated after rendering settings change (save and save-to-all)~~ This seems to be already fixed by @rgozim PRs #5845 

# Related reading

https://docs.google.com/spreadsheets/d/1SDXHxHpWQQ3VBpsGAOXdpZ4H1JdSMTQZHEu3Eg1CAQk/edit#gid=1885911056

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
@@ -272,6 +272,8 @@ public class ThumbnailLoader extends BatchCallTree {
                 }
             } else {
                 thumbnail = WriterImage.bytesToImage(thumbnailData);
+                if (thumbnail == null) 
+                    thumbnail = Factory.createDefaultThumbnail("Error");
             }
             // Convert thumbnail to whatever
             currentThumbnail = new ThumbnailData(pxd.getImage().getId(),
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
@@ -22,6 +22,7 @@
 package org.openmicroscopy.shoola.env.data.views.calls;
 
 import ome.conditions.ResourceError;
+import omero.ApiUsageException;
 import omero.ServerError;
 import omero.api.IConfigPrx;
 import omero.api.RawPixelsStorePrx;
@@ -33,6 +34,7 @@ import omero.gateway.model.DataObject;
 import omero.gateway.model.ImageData;
 import omero.gateway.model.PixelsData;
 import omero.log.LogMessage;
+
 import org.openmicroscopy.shoola.env.data.OmeroImageService;
 import org.openmicroscopy.shoola.env.data.model.ThumbnailData;
 import org.openmicroscopy.shoola.env.data.views.BatchCall;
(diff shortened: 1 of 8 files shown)


joshmoore:
black thumbnail case:
 * looking at user-8's image 151802 as user-9 on eel
```
2018-09-06 13:23:22,216 WARN  [ o.o.s.e.d.views.calls.ThumbnailLoader$1] ( Thread-14) Couldn't initialize the ThumbnailStore for pixels id 149552omero.ApiUsageException
    serverStackTrace = "ome.conditions.ApiUsageException: Unable to reset rendering settings in a ...
2018-09-06 13:23:24,601 WARN  [     o.o.s.a.m.editor.AnnotationTaskPane] (entQueue-0) UI for displaying ROIS annotations not implemented yet! 
2018-09-06 13:23:24,907 DEBUG [       omero.gateway.facility.Facility$1] ( Thread-24) Created new ROIFacility 
2018-09-06 13:23:24,917 DEBUG [       omero.gateway.facility.Facility$1] ( Thread-24) Created new DataManagerFacility 
```
 * double-clicking opens the image :+1:
 * closing and refreshing leads to a *gray* image, ... I assume because the rendering def has been created? and the following log:
```
2018-09-06 13:23:30,564 ERROR [o.o.s.e.data.views.calls.ThumbnailLoader] ( Thread-33) null 
```

So this behavior is definitely better, but perhaps slightly confusing. Is it possible for the original black image to have text over it? Is the gray thumbnail another, new condition?

dominikl:
That's interesting. I'll have a look were the gray thumbnail comes from. Sure, could have text as well, but what text? ApiUsageException could be quite generic... just show "Error"?

joshmoore:
I haven't been able to reproduce the existing "Error" text that @rgozim put in place. Until then, it's probably safest to keep them distinct. Options include:

 - "Can't render"
 - "Can't reset"
 - "Failed"
 - ...

joshmoore:
@dominikl : if it makes it easier to define these conditions, they can *all* be "Error" but then let's add a prefix/suffix to show where it occurred. Something like:

 * `ERROR1016`
 * `E23: error`
 * `Error (#09)`

dominikl:
I've not added different error codes yet, only "Can't render" is displayed, but surely can still do that, shouldn't prevent any testing yet.

joshmoore:
No worries. This can also be a follow-up when we talk about mouseovers, etc.

components/insight/SRC/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
```diff
@@ -220,13 +222,24 @@ protected void buildTree() {
                 BatchCall call = new BatchCall("Loading thumbnails") {
                     @Override
                     public void doCall() throws Exception {
-                        ThumbnailStorePrx store = getThumbnailStore(pxd);
+                        ThumbnailStorePrx store = null;
                         try {
+                            store = getThumbnailStore(pxd);
                             handleBatchCall(store, pxd, userId);
+                        } catch (ApiUsageException ex) {
```
joshmoore: If I understand the diff correctly, in the previous version this was protected by at least `Exception` if not `Throwable` to handle various possibilities. I defer on how this is handled elsewhere in Insight, but any remote call can call a number of exceptions beyond `ApiUsageException`.

dominikl: Yes, the safest way to prevent Insight crashes would be catch everything and show a black thumbnail in any error case. I'll change that.

mtbc:
> Don't request thumbnails

... for light-UI case only of course.

mtbc:
I'll also check plates, maybe true for those too.

mtbc:
Works fine for plates even with non-light-UI.

mtbc:
With everything skipped the light import UI still generates thumbnails when importing non-SPW data.

mtbc:
> just show black thumbnail instead

I get:
thumbnail

Not lovely but far better than a crash.

dominikl:
That's true, the light ui still triggers the thumbnail request somewhere, that's definitely a blocker.
Re 'black thumbnail': The last commit adds an error code. Not sure if that makes it lovlier though.

joshmoore:
> that's definitely a blocker.

But for outside this PR, right?

mtbc:
I don't mind how fixes are distributed across PRs.

joshmoore:
Thanks, guys.

PR #5847, opened 2018-09-06
adjust section titles in Insight import dialog options tab

mtbc:
Before:

before

After:

after

Now the titles look a bit more consistent and adding tags isn't hidden away.


dominikl:
Looks good 👍  Don't know if @pwalczysko  wants to take a look as well.

pwalczysko:
``Speedup`` ? Not ``speed-up`` or ``speed up`` ? 
It seems that "before" it was spelled better.

pwalczysko:
yes, this works for me, the addition of "Tags" is a good catch.

PR #5848, opened 2018-09-06
CLI import: expose more advanced Java import arguments

sbesson:
While reviewing 5.4.8-rc1, we realized that some of the advanced import arguments used for very fast CLI import including `--no-stats-info`, `--no-thumbnails` are not exposed in the Python arguments i.e. not displayed when calling `bin/omero import -h`.

This effectively prevents bulk configuration files to include these import arguments i.e. meaning skipping these steps need to be executed as `bin/omero import --bulk bulk.yml --skip <argument>` at the moment.

This PR adds these Java arguments to the `import` argument parsing as well as unit tests verifying these keys are recognized and passed to the Java command.

To test this PR:
- ~check that `bin/omero import -h`  includes the new options~
- create a minimal bulk configuration file similar to `components/tools/OmeroPy/test/unit/clitest/bulk_import/test_javaargs/bulk.yml` and check the import effectively skips thumbnail generation, min/max calculatin and upgrade check

/cc @dominikl


components/tools/OmeroPy/src/omero/plugins/import.py
```diff
@@ -449,6 +449,16 @@ def add_advjava_argument(*args, **kwargs):
         add_advjava_argument(
             "--checksum-algorithm", nargs="?", metavar="TYPE",
             help="Alternative hashing mechanisms balancing speed & accuracy")
+        add_advjava_argument(
+            "--no-stats-info", action="store_true",
+            help="Disable calculation of minima and maxima when as part of the"
+            " Bio-Formats reader metadata")
+        add_advjava_argument(
+            "--no-thumbnails", action="store_true",
+            help="Do not perform thumbnailing after import")
```
mtbc: "during" rather  than "after"

components/tools/OmeroPy/src/omero/plugins/import.py
```diff
@@ -449,6 +449,16 @@ def add_advjava_argument(*args, **kwargs):
         add_advjava_argument(
             "--checksum-algorithm", nargs="?", metavar="TYPE",
             help="Alternative hashing mechanisms balancing speed & accuracy")
+        add_advjava_argument(
+            "--no-stats-info", action="store_true",
+            help="Disable calculation of minima and maxima when as part of the"
```
mtbc: Interesting, it doesn't disable it when not part of the reader metadata?

joshmoore:
Primary high-level concern is whether we should be advertising these, i.e. is the redundancy with `--skip` confusing.

sbesson:
Re `--skip`, I agree with the potential confusion. Would `help=SUPPRESS` be a good option and/or do we want to support `skip: <SKIP>` in `bulk.yml`?

On the two command-line help, I literally copy-pasted what is in `CommanLineImporter.java`. Better formulations welcome and I can adjust both places to be in sync depending on the decision above.

joshmoore:
> do we want to support skip: <SKIP> in bulk.yml

My assumption was that it *should* work.

sbesson:
I certainly concur that having everything we pass on the command-line supported in the bulk configuration file was also my expectation. Given that both forms are currently acceptable, I would propose to get this reviewed and I will work on a separate PR allowing `skip: <value>` in the configuration file.

joshmoore (approved):
* Nice tests
 * `bin/omero import --no-thumbnails -f` works manually
 * options don't appear in regular help:

```
usage: bin/omero import [-h] [-C] [-s SERVER] [-p PORT] [-g GROUP] [-u USER]
                        [--javahelp] [--bulk [BULK]] [--logprefix [LOGPREFIX]]
                        [--file [FILE]] [--errs [ERRS]]
                        [--clientdir CLIENTDIR] [--logback LOGBACK] [-n NAME]
                        [-x DESCRIPTION] [--report] [--upload] [--logs]
                        [--email EMAIL] [--annotation-ns ANNOTATION_NS]
                        [--annotation-text ANNOTATION_TEXT]
                        [--annotation-link ANNOTATION_LINK] [-f] [-c]
                        [-l READER_FILE] [-d DATASET_ID] [-r SCREEN_ID]
                        [-T TARGET] [--debug LEVEL] [--output TYPE]
                        [--encrypted TYPE] [--advanced-help]
                        [--transfer [TYPE]] [--exclude [TYPE]]
                        [--checksum-algorithm [TYPE]]
                        [--parallel-upload COUNT] [--parallel-fileset COUNT]
                        [--depth DEPTH]
                        [--skip {all,checksum,minmax,thumbnails,upgrade}]
                        [path [path ...]]
```

only advanced help:

```
[jamoore@ome-c6100-1 dist]$ bin/omero import --java-help 2>&1 | grep thumb
[jamoore@ome-c6100-1 dist]$ bin/omero import --advanced-help 2>&1 | grep thumb
  --no-thumbnails		Do not perform thumbnailing after import
  e.g. $ bin/omero import -- --no-thumbnails foo.tiff
       $ ./importer-cli --no-thumbnails bar.tiff
```

PR #5849, opened 2018-09-07
import: add --readers flag and unit tests

joshmoore:
This is in addition to the short `-l` form.

# Testing this PR

1. double check the unit tests and their status
1. create a readers.txt for testing, e.g `echo loci.formats.in.FakeReader > /tmp/readers.txt`
1. try out `bin/omero import -f --readers=/tmp/readers.txt` in addition to `... -l /tmp/readers.txt ...`

PR #5850, opened 2018-09-07
CLI import: add skip support to bulk files

sbesson:
Follow-up of #5848, this PR adds support for `skip` option in bulk import configuration files. The various YAML files under ` components/tools/OmeroPy/test/unit/clitest/bulk_import/test_skip` should test all the values currently supported by the import plugin. The command-line usage of `--skip` option should be unmodified by this PR so integration tests are expected to pass.


components/tools/OmeroPy/test/unit/clitest/bulk_import/test_skip/upgrade.yml
```diff
@@ -0,0 +1,3 @@
+---
+skip: "all"
```
dominikl: Shouldn't this be `upgrade`?

dominikl:
Would the integration test better fit into `test_import_bulk.py` instead of `test_import.py`?

joshmoore:
:+1: from me for splitting the test files into smaller portions moving forward.

dominikl:
Apart from that, looks good. Also tested manually, works fine 👍

components/tools/OmeroPy/test/unit/clitest/bulk_import/test_skip/upgrade.yml
sbesson: You're right. Thanks for spotting that

sbesson:
I'll push a commit to address https://github.com/openmicroscopy/openmicroscopy/pull/5850#discussion_r216266818. Splitting bulk tests from `test_import.py` into `test_import_bulk.py` make sense and mirror the integration tests. Happy to also do that as part in this PR although my main concern is to conflict with #5849.

joshmoore:
Happy for us to go ahead and merge #5849 if that helps. I tried to put the readers test far away from the bulk tests for just this reason.

joshmoore:
Merging based on travis status. Happy to have test file reworking happen down the road.

PR #5851, opened 2018-09-07
Show duplicate status again

dominikl:
Should fix the "companion file" issue, see bug reports line 10:
https://docs.google.com/spreadsheets/d/1SDXHxHpWQQ3VBpsGAOXdpZ4H1JdSMTQZHEu3Eg1CAQk/edit#gid=1885911056

Luckily that was just a display issue.

/cc @mtbc

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/env/data/util/StatusLabel.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/env/data/util/StatusLabel.java
@@ -67,6 +67,9 @@ public class StatusLabel extends JPanel implements PropertyChangeListener {
     /** The text indicating the scanning steps. */
     public static final String SCANNING_TEXT = "Scanning...";
 
+    /** The text indicating the scanning steps. */
+    public static final String CANCELLED_TEXT = "Cancelled...";
+
     /** The width of the upload bar. */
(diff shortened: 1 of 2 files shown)


dominikl:
The missing change from "Pending..." to "Cancelled..." (line 11) was basically the same issue, StatusLabel not getting notified when the Status has changed. Should be fixed now.

components/insight/SRC/org/openmicroscopy/shoola/env/data/util/StatusLabel.java
```diff
@@ -67,6 +67,9 @@
     /** The text indicating the scanning steps. */
     public static final String SCANNING_TEXT = "Scanning...";
 
+    /** The text indicating the scanning steps. */
+    public static final String CANCELLED_TEXT = "Cancelled...";
```
mtbc: no ellipsis suffix, was simply "Cancelled"

mtbc (approved):
In local testing both issues are now fixed for me.

mtbc:
Also worked fine with this morning's eel merge build against pub-omero.

joshmoore:
Assuming @rgozim doesn't have any objections.

PR #5852, opened 2018-09-10
Lookupnames

dominikl:
# What this PR does

Reduce the number of calls to the `ConfigService` by requesting the values for server version and max plane size once at the login stage and storing them as LookupNames.

PR #5855, opened 2018-09-12
Move default values for max_plane_w/h to container.xml

dominikl:
# What this PR does

Moves the client side default value for non pyramid max plane width / height to the container.xml.
Addresses @jburel 's comment on https://github.com/openmicroscopy/openmicroscopy/pull/5852#pullrequestreview-154494677

### DIFF
--- a/components/insight/config/container.xml
+++ b/components/insight/config/container.xml
@@ -115,10 +115,12 @@
     -->
     <entry name="/services/Thumbnailing/fetchMediumSpeed" type="double">0.5</entry>
     
-    <!-- Default value for the max plane width which doesn't require pyramids - can be overwritten by the server  -->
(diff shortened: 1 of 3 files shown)


components/insight/config/container.xml
```diff
@@ -115,6 +115,11 @@
     -->
     <entry name="/services/Thumbnailing/fetchMediumSpeed" type="double">0.5</entry>
     
+    <!-- Default value for the max plane width which doesn't require pyramids - can be overwritten by the server  -->
```
jburel: it will be overwritten by the server value if it is set server side

jburel:
The values are correctly read from the config file
and they will be used if  not set server side or if it cannot be parsed.
Ready to merge

joshmoore:
Thanks, both!

PR #5858, opened 2018-09-17
import: better handle parallel stdout/stderr

joshmoore:
The various calls to `System.{out,err}.print()` (no `ln`)
allow multiple different threads to interleave their logging,
leading to output like:

```
Image:Image:2460
2461
```

This PR reduces the individual calls to System.{out,err} and synchronizes the `update`  method.

# Testing this PR

 1. perform multiple parallel imports, e.g. `time for x in $(seq 11 50); do time omero import -T Dataset:name:fake-$x *.fake --parallel-fileset=3 >$x.out 2>$x.err; done`
 2. checkout the output for lines *not* matching `Image:\d+`
 3. with this PR, that should not happen; _without_ the PR, it does occasionally. 

# Related reading

1. https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-Python27/968/testReport/OmeroPy.test.integration.clitest.test_import/TestImport/testParallelFileset/

Note: this does not *yet* solve interleaving of YAML output, etc.

### DIFF
--- a/components/blitz/src/ome/formats/importer/cli/LoggingImportMonitor.java
+++ b/components/blitz/src/ome/formats/importer/cli/LoggingImportMonitor.java
@@ -54,7 +54,7 @@ public class LoggingImportMonitor implements IObserver
         return old;
     }
 
-    public void update(IObservable importLibrary, ImportEvent event)
+    public synchronized void update(IObservable importLibrary, ImportEvent event)
     {
         if (event instanceof IMPORT_DONE) {
             IMPORT_DONE ev = (IMPORT_DONE) event;
--- a/components/blitz/src/ome/formats/importer/cli/LoggingImportMonitor.java
+++ b/components/blitz/src/ome/formats/importer/cli/LoggingImportMonitor.java
@@ -223,15 +223,12 @@ public class LoggingImportMonitor implements IObserver
                 String exclude) {
 
             System.err.println("Other imported objects:");
-            System.err.print("Fileset:");
-            System.err.println(fid);
+            System.err.println("Fileset:" + fid);
             for (String kls : collect.keySet()) {
                 if (exclude == null || !kls.equals(exclude)) {
                     List<Long> ids = collect.get(kls);
                     for (Long id : ids) {
-                        System.err.print(kls);
-                        System.err.print(":");
-                        System.err.println(id);
+                        System.err.println(String.format("%s:%d", kls, id));
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


joshmoore:
A more correct solution may be to replace `LoggingImportMonitor lim = new LoggingImportMonitor();` with `new ParallelLoggingImportMonitor()` when parallel filesets is activated.

 * https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.7/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java#L283
 * https://trello.com/c/iIpGEb7X/39-rfe-log-fileset-during-parallel-output

I would say this is an easy addition for later, but it potentially breaks the stdout/stderr "API" and therefore might be worth (quickly) discussing.

cc: @sbesson @mtbc

mtbc:
This PR may also fix https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-integration-python/754/testReport/ (see its console log).

mtbc:
The code changes look good.

mtbc:
Would it suffice (also for YAML, etc.) to simply make a bunch of the methods `synchronized`? Or do we expect some relevant ones not to be quick?

joshmoore:
For YAML (which isn't overly frequently used as far as I know), I could imagine that'd be the easiest solution. `getObjectIdMap` might take a _bit_ but nothing should really be blocking. Happy to give it a go.

mtbc:
Without this PR I managed to get 29 failures with ten fakes and `--parallel-{upload,fileset}=5`.

joshmoore:
An example of using thread name as the prefix for stdout/stderr:

```
Ice.ThreadPool.Client-0:Imported objects:
Ice.ThreadPool.Client-0:---
Ice.ThreadPool.Client-0:- Fileset: 391
Ice.ThreadPool.Client-0:  Image: [930]
```

Not overly attractive. See the [patch.txt](https://github.com/openmicroscopy/openmicroscopy/files/2388082/patch.txt)


I've tested `public synchronized void update()`  locally. It works fine with fake datasets. Pushing that commit now.

mtbc:
In comparison with https://github.com/openmicroscopy/openmicroscopy/pull/5858#issuecomment-421946279, with this PR in, the imports reported perfectly to stdout, no interleaving failures. :+1:
```console
$ cat *.out | cut -f 1 -d : | uniq -c
    400 Image
$ 
```

mtbc:
Either the earlier commit or the last suffice:
```console
$ for i in without-PR original-PR last-commit-only without-PR-retest
> do echo -n $i:\  ; grep -Ex 'Image:[[:digit:]]+' $i/*.out | wc -l
> done
without-PR: 342
original-PR: 400
last-commit-only: 400
without-PR-retest: 352
$ grep -vEx 'Image:[[:digit:]]+' */*.out | cut -f 1 -d / | uniq -c
     58 without-PR
     48 without-PR-retest
$ 
```

joshmoore:
I assume the earlier commit is not detrimental, eh? So really it's a question if we're happy with the `synchronized update` strategy.

mtbc:
Indeed, not detrimental.

mtbc:
On https://github.com/openmicroscopy/openmicroscopy/pull/5858#issuecomment-421949469 we could probably accumulate a thread-safe cache of prettier names to which we translate.

joshmoore:
Interesting. Can certainly see how that would work, but I'd say despite the output-api-breakingness let's look into that down the road.

joshmoore:
Merging for the second build of 5.4.8

PR #5859, opened 2018-09-17
Birdeyeview

jburel:
# What this PR does

Insight generates a "second" thumbnail  (size not 96x96) when viewing a pyramid image
The user might not have the permissions to write a pyramid on disk
In that case, the ``getThumbnailDirect`` method is used

# Testing this PR
In a read-only group
* User 1 import an big image but does not view the Image (thumbnail can be generated)
* User 2 view the image of user-1
* Check that the bird eye view is displayed

# Related reading

See 5.4.8 testing row 9 for background

cc @kennethgillen @rgozim @joshmoore

### DIFF
--- a/components/server/src/ome/services/ThumbnailBean.java
+++ b/components/server/src/ome/services/ThumbnailBean.java
@@ -510,8 +510,7 @@ public class ThumbnailBean extends AbstractLevel2Service
      */
     private byte[] convertThumbnailToBytes(BufferedImage image, boolean inProgress)
             throws IOException {
-        ByteArrayOutputStream byteStream = new ByteArrayOutputStream();
-        try {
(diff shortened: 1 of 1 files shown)


components/server/src/ome/services/ThumbnailBean.java
```diff
@@ -1340,6 +1365,8 @@ protected void actOnOneGroup(Set<Long> pixelsIds) {
         }
     }
 
+
```
mtbc: Probably unintentional?

jburel:
@mtbc the NPE is due to the fact there is not a thumbnail with that ID
``String path = getThumbnailPath(thumbnail.getId());`` in the ThumbnailService

We could check that before and avoid the exception handling

joshmoore:
Tested locally by @joshmoore and @rgozim. Merging for the second build of 5.4.8

PR #5861, opened 2018-09-18
Simplified retrieveThumbnail and added call to iUpdate.saveObject

rgozim:
# What this PR does

Fixes a bug where changing rendering settings for the owner of an image results in a separate users from not being able to view the updated image.

### DIFF
--- a/components/server/src/ome/services/ThumbnailBean.java
+++ b/components/server/src/ome/services/ThumbnailBean.java
@@ -891,13 +891,7 @@ public class ThumbnailBean extends AbstractLevel2Service
                         _setMetadataVersion(thumbMetaData, inProgress);
                         dirtyMetadata = true;
                     } else {
-                        //new one for owner of the settings.
-                        final Dimension d = new Dimension(thumbMetaData.getSizeX(),
-                                thumbMetaData.getSizeY());
-                        thumbMetaData = ctx.createThumbnailMetadata(pixels, d);
-                        _setMetadataVersion(thumbMetaData, inProgress);
-                        thumbMetaData = iUpdate.saveAndReturnObject(thumbMetaData);
-                        dirtyMetadata = false;
+                        thumbMetaData = createThumbnailSettingsForUser(thumbMetaData);
                     }
                 } finally {
                     try {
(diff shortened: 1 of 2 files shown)


will-moore:
Built locally, testing with Insight b97 this is working now for me:
 Saving rendering settings changes in Insight and viewing thumbnails from web as non-owner in read-only group is working fine.

will-moore:
Related 'Known' bug (also present on 5.4.7):
If I "Save to All" for a multi-field Plate in Insight or Web, read-only group, but I don't request thumbnails for all fields, then a non-owner will see 'No Image' for all those fields.
screen shot 2018-09-18 at 16 17 52

Not a blocker.

jburel:
Regression:
 * Import a big image as tb1 (no thumbnail) Do no view
 * Log in as tb2, view image. The question mark is back. Also 128x128 is requested and 96x96 is returned (this is why we have 96x96 question mark)

pr_5861

jburel:
Bird eye now loads correctly in insight 

Side effect noticed:
* Tb1 import a big image no thumbnail
* tb2 view the image -> rnd settings owned by tb2 are created
* tb1 view the image -> rnd settings owned by tb1 are created and thumbnail
* tb1 modify tb1 settings and save
* tb2 loads the dataset and tb1 thumbnail is displayed => when viewing the image mismatch between what's in the viewer and the bird eye view

jburel:
Similar issue with this PR than the one described https://github.com/openmicroscopy/openmicroscopy/pull/5862#issuecomment-422923587

The problem is due to the fact that the thumbnail of tb1 is used for tb2 settings leading to the impression that the settings of tb2 are changed

Edit: The problem is partially visible in web.
i.e. if you go to web after doing the changes in insight then you will see the same thumbnail for tb1 and tb2. 
It "corrects" itself after changing the settings in Web.
The way to load the thumbnail is different between web and insight

jburel:
See https://github.com/openmicroscopy/openmicroscopy/pull/5862#issuecomment-423073714
It is a regression

joshmoore:
Excluding this form the 5.4.8 builds.

jburel:
Closing this PR, the code has been migrated to another repo

PR #5869, opened 2018-09-25
have server dispose() of graphics contexts it creates

mtbc:
# What this PR does

Unless an AWT `Graphics` instance was provided by another owner (e.g., AWT asked for a repaint) they should be `dispose()`d so that resources are recovered promptly, otherwise it can be a while before finalizers run. Insight is riddled with this problem (e.g., it may not have been realized that `BufferedImage.getGraphics()` can eat memory in this way, it's hardly intuitively named) but this PR focuses on correcting just the server's few usages.

# Testing this PR

Thumbnails should still generate and render just fine.

### DIFF
--- a/components/server/src/ome/logic/AWTScaleService.java
+++ b/components/server/src/ome/logic/AWTScaleService.java
@@ -1,7 +1,5 @@
 /*
- * ome.logic.QueryImpl
- *
- *   Copyright 2006 University of Dundee. All rights reserved.
+ *   Copyright 2006-2018 University of Dundee. All rights reserved.
  *   Use is subject to license terms supplied in LICENSE.txt
  */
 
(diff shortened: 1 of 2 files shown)


mtbc:
Created https://trello.com/c/bV2UR8ZM/518-insight-and-graphicsdispose for the Insight side.

components/server/src/ome/logic/AWTScaleService.java
```diff
@@ -58,6 +56,7 @@ public BufferedImage scaleBufferedImage(BufferedImage image, float xScale,
         // graphics2D.setRenderingHint(RenderingHints.KEY_INTERPOLATION,
         // RenderingHints.VALUE_INTERPOLATION_BILINEAR);
         graphics2D.drawImage(image, 0, 0, thumbWidth, thumbHeight, null);
+        graphics2D.dispose();
```
joshmoore, line 59: Worth `try/finally`s?

mtbc: I thought probably not given that they do get cleaned up anyway, it's not like leaking file descriptors, and it's pretty unlikely that `drawImage` and its like are going to throw, so just trying to balance code clutter with likely benefit.

dominikl:

PR #5821, opened 2018-08-15
perf4j: log import timings in import log

joshmoore:
By adding use of Slf4J's MDC to ManagedImportProcessI and
ManagedRepositoryI, we are able to get stop watch info as
to various internal stages of the import process.

Note: this also required adding perf4j to the SIFT logger in etc/logback.xml

# Testing this PR

1. with a new server, perform an import (I've used here the 300 fakes from the linked trello card)
2. examine the perf4j lines in the import log

```
bin/omero fs logfile 351 | grep perf4j | grep omero.import | java -jar lib/server/perf4j.jar -r -t 120000
Performance Statistics   2018-08-15 08:40:00 - 2018-08-15 08:42:00
Tag                                                  Avg(ms)         Min         Max     Std Dev       Count
omero                                                   47.0           0        1965        91.9         605
omero.import                                            47.0           0        1965        91.9         605
omero.import.process                                    47.0           0        1965        91.9         605
omero.import.process.checksum                            0.0           0           0         0.0         301
omero.import.process.create                           1965.0        1965        1965         0.0           1
omero.import.process.init                              255.0         255         255         0.0           1
omero.import.process.opener                             86.6          48         226        28.1         301
omero.import.process.verify                            161.0         161         161         0.0           1
```

# Related reading

1. https://trello.com/c/BPrkScBc/36-file-upload-should-be-faster

PR #5870, opened 2018-09-25
reduce Memoizer logging

mtbc:
# What this PR does

Reverts commit f8d159ed463ef920666deb06f019ca8ade3a6860 to reduce the large amount of memoizer logging when we import plates, etc. The memoizer is now quite stable and understood and logging can still easily be increased by those who wish to.

# Testing this PR

Compare import logs for plates imported with and without this PR and see that this PR stops the `loci.formats.Memoizer` DEBUG lines from appearing.

### DIFF
--- a/etc/logback.xml
+++ b/etc/logback.xml
@@ -106,7 +106,6 @@
   <!-- Quieting blitz startup -->
   <logger name="net.sf.ehcache" level="ERROR"/>
   <logger name="loci" level="INFO"/><!-- Bio-Formats -->
-  <logger name="loci.formats.Memoizer" level="DEBUG"/>
   <logger name="ucar" level="WARN"/><!-- NetCDF -->
 
   <!-- BUILD ////////////////////////////////////////////////////////////////// -->


joshmoore:
I'm pretty sure there are times when I do check this logger on historical files, but on the flip side, I know it's verbose. The code paths of interest tend to be import and rendering. Can we review if there's at least some statement that shows a memo file was loaded and how long it took but perhaps from a new location? If that's still too verbose, happy to see this reduced.

sbesson:
Having looked at the typical statements from a production server, the most redundant lines for every call are:

1. the `perf4j` statements for `loci.formats.Memoizer.loadMemo` and `loci.formats.Memoizer.setId`
2. the `loaded memo file` statement

I have similar thoughts as the ones expresssed https://github.com/openmicroscopy/openmicroscopy/pull/5870#issuecomment-425815566. I'll add that https://github.com/openmicroscopy/openmicroscopy/pull/5821 moved us towards putting more timing logging information in the server log rather than less. Worth adding  an OMERO level `per4j` INFO call to the `setId` call as a first step ?

mtbc:
The time obviously isn't as ripe as I'd thought: other work needs doing first.

PR #5877, opened 2018-10-01
ImportCandidates bugfix

dominikl:
# What this PR does

Prevents that files are getting ignored, although they might still be needed by other filesets.
E.g. this happens if a `*.dv.log` file is associated with more than one `*.dv` files (in which case the following dv files could be silently skipped).

Adds a unit test using artificially created nhdr files to test this scenario.

# Testing this PR

- Check that the unit test makes sense, that it fails without the bugfix commit 9d29ea4 (see travis build log before the commit) and that it passes with the bugfix commit.
- Check the files from QA 21768
- We should also manually test some other file formats which can have multiple associated files to make sure this change doesn't have any unexpected side effects.

PR #5879, opened 2018-10-02
ImportCandidates bugfix

dominikl:
Replaced by #5877
----------

# What this PR does

Prevents that files are getting ignored, although they might still be needed by other filesets.
E.g. this happens if a `*.dv.log` file is associated with more than one`*.dv` files (in which case the following dv files could be silently skipped).

# Testing this PR

Check that the integration test #5877 makes sense, that it fails without this PR and that it passes with this PR included.
We should also manually test some other file formats which can have multiple associated files to make sure this change doesn't have any unexpected side effects.

# Related reading

https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8595


dominikl:
Added the bugfix commit to the PR with the tests: #5877

PR #5881, opened 2018-10-03
More ImportCandidatesTests

dominikl:
# What this PR does

Follow-up to #5877 .
Adds some more tests to the ImportCandidatesTest

# Testing this PR

Check the unit tests.

/cc @sbesson


sbesson:
Thanks for the additional tests. I think this covers a few of the obvious use cases and should reduce the risks of regression in the import candidate detection logic. For extra confidence, it might be we want to test both the folder level detection as well as the individual files detection with all permutations for all use cases but that can be captured as a follow-up.

Travis is passing and displaying all the individual tests as successful. LGTM.

sbesson (approved):

PR #5882, opened 2018-10-04
Importerui bugfix

dominikl:
# What this PR does

Two small bugfixes of the lightweight importer ui:

- The content of the failed import dialog was added to the root pane instead of the content pane, which causes problems on Windows and Linux (drop down of the file selection combo box didn't work).
- If multiple files are selected of which some fail to import and some are skipped the spinners in the lightweight ui kept spinning although the import was already finished.

# Testing this PR

Import the files from QA 21768 selecting all files (not the directory) .After the import finished:
- Check that the upload and process spinners stop.
- Click on "Show failed" and check that the dialog works.

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/FailedImportDialog.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/FailedImportDialog.java
@@ -161,8 +161,8 @@ public class FailedImportDialog extends JDialog {
             }
         });
 
-        getRootPane().setLayout(new BorderLayout());
-        getRootPane().add(con, BorderLayout.CENTER);
+        getContentPane().setLayout(new BorderLayout());
+        getContentPane().add(con, BorderLayout.CENTER);
 
         pack();
 
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/ImporterUIElementLight.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/view/ImporterUIElementLight.java
@@ -238,7 +238,7 @@ class ImporterUIElementLight extends ImporterUIElement {
         // if there are many imports to cancel this might take a while; in that
         // case until all cancellations went through there might have been another
         // upload happened so this could sum up to something > super.totalToImport
-        if (uploaded + cancelled + super.countFailure >= super.totalToImport) {
+        if (super.isDone()) {
             uploadBusy.setBusy(false);
             uploadBusy.setIcon(IconManager.getInstance().getIcon(IconManager.APPLY));
         }  else {
@@ -248,7 +248,7 @@ class ImporterUIElementLight extends ImporterUIElement {
         processed.setValue(complete);
         processed.setMaximum(uploaded);
         processedBusy.setText(complete + "/" + uploaded);
-        if (complete + cancelled + super.countFailure >= super.totalToImport) {
+        if (super.isDone()) {
             processedBusy.setBusy(false);
             processedBusy.setIcon(IconManager.getInstance().getIcon(IconManager.APPLY));
         } else {


dominikl:
If there is time, I'd like to change one more thing. To make the UI more compatible to the CLI output, I'd change the current UI...

screen shot 2018-10-04 at 11 03 46

- To count all files as uploaded (including skipped and error) so that at the end it'll show (120/120).
- Changed "Processed" to "Imported"
- Add a field "Skipped" to show the number of skipped files.

Would that still be doable for 5.4.9 ?

This would basically look like that:

screen shot 2018-10-04 at 11 12 55

jburel:
@dominikl that will be useful
But that UI will only show up when you do a large import
What will show up if I select a folder with a small number, same as usual?

dominikl:
If you select the folder, the lightweight UI will still be shown (as it contains more than 100 files), but only the 48 imports will be picked up, because in that case the other files are already getting sorted out by the ImportCandidates.
screen shot 2018-10-04 at 11 26 09

With a folder with less than 100 files the old UI is shown.

dominikl:
I'll add the commit so people can have a look themselves. Can remove it again if there are objections.

sbesson:
@dominikl: immediate thoughts:

- :+1: for having consistency between the metrics reported the different import clients
- :+1: for having folder import vs individual files upload give the same results from the UI perspective
- 💯  for not conveying the incorrect impression that half of the files were missed from the upload


My primary concern is that as we introduce these changes in OMERO 5.4.8, we need to find a compromise between fixing confusing (and scary) bugs and keeping changing the UI elements i.e. modifying from the expectations of the consumers. Happy to hear thought's from people involved in training/outreach here /cc @jburel @pwalczysko.

If we wanted to assess the last UI change, a couple of initial questions come to mind:

- are these changes restricted for the lightweight importer UI, cf @jburel's comment? Currently the old UI only shows "Imported"?
- we might need to check the behavior across a few scenarios here e.g. n files <-> m filesets <-> o images including MIFs, SPW
- given the lightweight importer size, should it be a bit more verbose and be explicit about what these numbers represent (`files`, `filesets`...) ?

dominikl:
"are these changes restricted for the lightweight importer UI, cf @jburel's comment? Currently the old UI only shows "Imported"?"
Yes, this affects the lightweight ui only. The old UI doesn't show a number of imported files explicetely, it lists each single seperately (and marks it as skipped, failied or complete).

jburel:
Since we have a lot of space in that UI
it might be useful to define the meaning of terms used in the summary Imported, skipped etc.
People will be confused by 120/120 uploaded and 48/48 imported

jburel:
I also think that we should validate with the person who reported the issue if such UI will not generate more confusing
I don't think we should rush to include it in 5.4.9

dominikl:
Ok, can remove the last commit again. Just worried that the 48/120 progress bar will scare people.

dominikl:
By the way, the issue with QA 21768 wouldn't have been picked up by any UI. Because when you select the directory to import, the files are already silently skipped in the ImportCandidates, Insight never knew anything about them.

dominikl:
Removed the commit with the UI changes. So it's now only the two bugfixes mentioned in the description.

mtbc:
Yes, on Linux without this fix: The spinners still spin and the dropdown doesn't work.

mtbc:
On Linux with this fix: The spinners turn to green checkmarks and the dropdown works.

pwalczysko:
In my opinion, nobody claims here to have explicitly tested this PR on Windows. I am ready to perform the test, will consult with @dominikl

jburel:
Reviewing on mac, things are working. No regression. One RFE (outside the scope of this PR), it will be good to indicate the file not considered. This is different from "skipped". I have in the folder a file that it is not recognised by Importcandidates (this is correct since it is an IJ macro)

pwalczysko:
On Win7 both with and without this PR:

   - after  uploading the first 3 files, the UI gets a bit frozen, the spinners are there but not spinning, and the button "Show Failed" is not clickable.
   - this  goes away after some time, but the UI progresses in fits rather than a smooth process

pwalczysko:
All the other fixes intended in this PR are as epected (green tick and the stopping of the spinners)

sbesson:
Thanks all. Given the time, I am in favor of carding separately the issue reported in https://github.com/openmicroscopy/openmicroscopy/pull/5882#issuecomment-427804384. Merging this for inclusion in 5.4.9

PR #5884, opened 2018-10-10
Ignore owner when checking omero.data.dir is accessible

manics:
# What this PR does

The owner/permissions check on the OMERO data dir is too strict, for example it will fail if owned by root even if it is group accessible.

# Testing this PR

1. Create a new OMERO data dir
2. Make it owned by `root`
3. Change the group of the data dir to one that the user that will be running OMERO.server is a member of
4. Make the OMERO data dir group readable/writeable
5. Start the server and it should work!

# Related reading
- https://trello.com/c/qndSyho7/532-omero-admin-datadir-owner-check-is-too-strict

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -16,7 +16,6 @@
 import re
 import os
 import sys
-import stat
 import platform
 import datetime
 import Ice
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1513,18 +1513,9 @@ present, the user will enter a console""")
             self.ctx.die(8, "FATAL: OMERO directory does not exist: %s"
                          % pathobj)
 
-        owner = os.stat(filepath)[stat.ST_UID]
-        if owner == 0:
-            msg = ""
-            msg += "FATAL: OMERO directory which needs to be writeable"\
-                " belongs to root: %s\n" % filepath
-            msg += "Please use \"chown -R NEWUSER %s\" and run as then"\
-                " run %s as NEWUSER" % (filepath, sys.argv[0])
-            self.ctx.die(9, msg)
-        else:
-            if not os.access(filepath, mask):
-                self.ctx.die(10, "FATAL: Cannot access %s, a required"
-                             " file/directory for OMERO" % filepath)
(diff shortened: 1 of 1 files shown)


joshmoore:
@manics : did you consider the other uses of can_access here?
```
$ git grep -n can_access admin.py | grep -v def
admin.py:1537:            self.can_access(var, mask)
admin.py:1541:            self.can_access(omero_data_dir, mask)
admin.py:1546:                self.can_access(subpath, mask)
admin.py:1620:        if not cfg_xml.exists() and self.can_access(grid_dir):
admin.py:1621:            if cfg_tmp.exists() and self.can_access(cfg_tmp):
admin.py:1700:            self.can_access(omero_data_dir)
```
It may well be that the weirdness was introduced to deal with root-owned `var` directories which I vaguely remember being a real issue at one point.

manics:
No, `var` and `etc` are currently embedded under the OMERO.server directory which all our instructions say should be owned by the OMERO user, the data dir is the only directory that currently lives outside.

joshmoore:
Sure, but the check is (also) there to enforce just that.

manics:
Isn't that handled by `os.access()`? In what situation would the owner be `root` and `os.access()` return `True` due to group perms but the directory still be unreadable/writeable?

chris-allan:
We've had some similar issues where `os.access()` returns `False` but the user has access to the folder in cases where there are extended ACLs in place, for example via NFSv4, on the binary repository directory.

manics:
Here's a concrete use case: https://github.com/manics/kubernetes-omero/issues/1

dominikl:
Doesn't seem to work. 
Created a omero.data.dir, set owner to 'root', group 'omero', set permission to 770. Start up as user 'dominik' (also member of group 'omero'), get: `FATAL: Cannot access /home/dominik/Testing/dev_data, a required file/directory for OMERO`. Then set permission to 774, this time I got in Blitz.log:
```
...
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'serverDirectoryCheck' defined 
in class path resource [ome/services/startup.xml]: Invocation of init method failed; nested exception is org.springframework.bean
s.factory.BeanCreationException: required directory /home/dominik/Testing/dev_data/FullText cannot be read but repository is read
-only
```

Note: Built directly from this branch (lastet commit 783297573 ). Maybe it has to be rebased or something!?

manics:
Ported to https://github.com/ome/omero-py/pull/208

PR #5885, opened 2018-10-25
Python gateway: add contextmanager for connection

joshmoore:
# What this PR does

Add a new method `open()` similar to `connect()` and `close()` together so that `with` can be used for login.

# Related reading

 * https://twitter.com/totti0223/status/1055352420417462272

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -17,6 +17,7 @@ import os
 
 import warnings
 from collections import defaultdict
+from contextlib import contextmanager
 from types import IntType, LongType, UnicodeType, ListType
 from types import BooleanType, TupleType, StringType, StringTypes
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


manics:
It'd be nice if this could work with the `__init__` parameters, e.g.
```
with Blitzgateway(client_obj=client) as conn:
    conn.getObject(...)
```
Unfortunately `__init__` is inconsistent, constructing using a `client_obj` means the returned object is already connected, whilst constructing with host/user/pass requires an explicit call to `connect()`. Could you use this as an opportunity to provide a unified interface when using this as a contextmanager, i.e.
```
with Blitzgateway(user, password, host="localhost") as conn:
    conn.getObject(...)
```
is automatically connected?

See https://trello.com/c/zebaFaV2/315-blitzgatewayuser-password-hosthost-doesnt-work#comment-5a53a0f87568cf14764cd556

joshmoore:
I thought about `with Blitzgateway(client_obj=client) as conn:` but having `contextmanager` on the class/constructor wasn't immediately clear to me. Happy to take an alternative PR. For this PR, I could definitely see throwing an exception if the object has been created with `client_obj`.

manics:
See https://github.com/openmicroscopy/openmicroscopy/pull/5887

joshmoore:
Closing in favor of gh-5887

PR #5887, opened 2018-10-26
Add context-manager to BlitzGateway

manics:
# What this PR does

Turns BlitzGateway into a context manager. Alternative implementation of https://github.com/openmicroscopy/openmicroscopy/pull/5885

# Testing this PR

BlitzGateway has two constructors which may behave unintuitively when it comes to whether they're connected or not https://trello.com/c/zebaFaV2/315-blitzgatewayuser-password-hosthost-doesnt-work#comment-5a53a0f87568cf14764cd556

All-in-one:
```
with BlitzGateway('user', 'pass', host='example.openmicroscopy.org') as conn:
    print list(conn.getObjects('Project'))
    assert conn._connected
assert not conn._connected
```
Using an existing client:
```
import omero
client = omero.client('example.openmicroscopy.org')
client.createSession('user', 'pass')

with BlitzGateway(client_obj=client) as conn:
    print list(conn.getObjects('Project'))
    assert conn._connected
assert not conn._connected
# client.getSession() should throw "ClientError: No session available"
```

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1573,6 +1573,26 @@ class _BlitzGateway (object):
         # The properties we are setting through the interface
         self.setIdentity(username, passwd, not clone)
 
+    def __enter__(self):
+        """
+        Enter a context manager, connect if necesaary
+        Raise Exception if not connected
+
+        with BlitzGateway('user', 'password', host='host') as conn:
+            print list(conn.getObjects('Project'))
+        """
+        if not self._connected:
+            r = self.connect()
+            if not r:
+                raise Exception("Connect failed")
+        return self
+
+    def __exit__(self, *args):
+        """
+        Exit a context manager, close connection
+        """
+        self.close()
+
     def _register_service(self, service_string, stack):
         """
         Register the results of traceback.extract_stack() at the time


components/tools/OmeroPy/src/omero/gateway/__init__.py
```diff
@@ -1573,6 +1573,26 @@ def __init__(self, username=None, passwd=None, client_obj=None, group=None,
...
+            r = self.connect()
+            if not r:
+                raise Exception("Connect failed")
+        return self
+
+    def __exit__(self, *args):
+        """
+        Exit a context manager, close connection
+        """
+        self.close()
```
joshmoore, line 1594: Does close call the object even if not _connected?

manics: You can call close() as many times as you want without errors:
```
In [3]: conn._connected
Out[3]: True

In [4]: conn.close()

In [5]: conn._connected
Out[5]: False

In [6]: conn.close()

In [7]: conn._connected
Out[7]: False
```

joshmoore: Sorry, that's not my worry, but rather if the `client_obj` is passed in the the _caller_ owns the object and not the `BlitzGateway` object.

manics: I can think of two reasonable options:
1. There's no point using `BlitzGateway(client_obj)` as a context manager unless you want it to be automatically closed, so we assume it's intentional, and that the user was passed the `client` from a source but has responsibility for closing it (for example `omero shell --login`).
2. Throw an exception if `BlitzGateway` is passed as a context manager and it's already connected.

manics: Option 1 is also consistent with the behaviour of the existing [`omero.cli.cli_login`](https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.9/components/tools/OmeroPy/src/omero/cli.py#L1545).

1. Create a CLI session with `omero login ...`
2. Run a script that includes `with cli_login() as cli: ...`
3. When the script exits you will be logged out of the session created in step 1

joshmoore: I can see getting behind Option 1. Can we just get the fact that the close will happen into the docs somewhere? I'd assume some example for exposing to users will need adding. cc: @pwalczysko @jburel

pwalczysko: @joshmoore so does that mean that with Option1 the lines like https://github.com/ome/training-scripts/blob/master/practical/groovy/analyse_particles_for_another_user.groovy#L218 will not be necessary anymore ? Maybe I am still missing the bigger picture here.

pwalczysko: What does that mean that a script ``exits`` ?

joshmoore: > does that mean that with Option1 the lines like 

Yes, except at the moment, only in pure Python. We'd have to examine making BlitzGateway `Closeable` (if it's not already) on the Java side. cc: @dominikl 

> What does that mean that a script exits

In this case, it means when the `with` block ends:

```
with BlitzGateway() ...
   in the block
   in the block
   in the block
out of the block; no session
```

pwalczysko: Aha, so once this PR is merged, then the ``with BlitzGateway()...`` pattern is compulsory for all scripts ?  Not anymore https://github.com/ome/training-scripts/blob/master/practical/jython/analyse_particles_for_another_user.jy#L81 ?

joshmoore: > pattern is compulsory for all scripts ?

Definitely not compulsory. This is just a convenience that was suggested by a workshop attendee in Kobe.

manics: Since we're agreed on the behaviour I've updated the docs: https://github.com/openmicroscopy/ome-documentation/pull/1915

joshmoore:
Merging. Addition to tests, training material and elsewhere will follow.

PR #5888, opened 2018-10-29
improve error if attempting graph operation with no targets

mtbc:
When OMERO.cli users attempt chgrp, chown, etc. but without specifying any target object, do not give a nasty stack trace. See https://trello.com/c/LHtMPy8y/69-bug-chmod-with-no-arguments-fails for details.

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1944,6 +1944,8 @@ class GraphControl(CmdControl):
             if exc:
                 opt.excludeType = exc
 
+        if args.obj is None or not args.obj:
+            self.ctx.die(440, "no object targets supplied for graph operation")
         commands, forces = zip(*args.obj)
         show = not (args.force or args.dry_run)
         needsForce = any(forces)


sbesson:
An alternate approach would be to modify `GraphControl._objects` and set `nargs` to `+` instead of `*`  i.e. fail at the argument parsing stage if not at least one argument is passed. Are there concrete implementations of this control which support no object argument passed to the command-line (I assume they would need to override `main_method`)?

mtbc:
`bin/omero delete --list` is one.

joshmoore:
That actually holds for _all_ graph controls. This looks fine and works as expected. I'll leave to you if you'd prefer this strategy:

```
[hudson@ome-c6100-1 dist]$ diff cli.py.old lib/python/omero/cli.py
1882a1883
>         self.__help = parser.print_help
1947a1949
>             self.__help()
```

i.e. `if no_args and no_list: print_help and die`

mtbc:
I think probably I wouldn't but thank you for the suggestion: I'd worry that the error message could get lost among the rest.

joshmoore:
Ok. Merging. I'll ponder on showing helping and perhaps add a `self.die_with_help` method for all to use in these more complicated situations.

PR #5889, opened 2018-10-29
check syntax of key names in loading configuration from file

mtbc:
Adds some checks for configuration property keys for obviously wrong characters. Review https://trello.com/c/o33zUcPZ/6-valid-chars-in-omero-config-keys where @manics found that they could accidentally apply `bin/omero config load` to set all kinds of nonsense. This PR does not fix all possible ways of setting daft keys but it does address that particular workflow for until we get around to something more comprehensive.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -576,7 +576,7 @@ re-run"""
         _new = parts[1]
         if _key in config.keys():
             _old = config[_key]
-        elif search(r'[\s=]', _key):
+        elif search(r'[\s]', _key):
             self.ctx.die(506, 'Illegal property name: {}'.format(_key))
         else:
             _old = None
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -260,6 +260,12 @@ class PrefsControl(WriteableConfigControl):
         except Exception, e:
             self.ctx.die(113, str(e))
 
+    def assert_valid_property_name(self, key):
+        from re import search
+
+        if search(r'[\s]', key):
+            self.ctx.die(506, 'Illegal property name: {}'.format(key))
+
     @with_config
     def all(self, args, config):
         for k, v in config.properties(None, True):
@@ -326,6 +332,7 @@ class PrefsControl(WriteableConfigControl):
             else:
                 f = args.file
             try:
+                self.assert_valid_property_name(args.KEY)
                 config[args.KEY] = (''.join(f)).rstrip()
             finally:
                 f.close()
@@ -334,6 +341,7 @@ class PrefsControl(WriteableConfigControl):
             if args.report:
                 self.ctx.out('Changed: Removed %s' % args.KEY)
         else:
+            self.assert_valid_property_name(args.KEY)
             config[args.KEY] = args.VALUE
             if args.report:
                 self.ctx.out('Changed: Set %s:%s' % (args.KEY, args.VALUE))
@@ -558,8 +566,6 @@ re-run"""
         config["omero.config.upgraded"] = "4.2.0"
 
     def handle_line(self, line, config, keys):
-        from re import search
-
         line = line.strip()
         if not line or line.startswith("#"):
             return None
@@ -576,9 +582,8 @@ re-run"""
         _new = parts[1]
         if _key in config.keys():
             _old = config[_key]
-        elif search(r'[\s]', _key):
-            self.ctx.die(506, 'Illegal property name: {}'.format(_key))
         else:
+            self.assert_valid_property_name(_key)
             _old = None
 
         if keys and _key in keys and _new != _old:
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -443,6 +443,8 @@ class PrefsControl(WriteableConfigControl):
         if not args.q:
             keys = config.keys()
 
+        # Handle all lines before updating config in case of error.
+        new_config = dict(config)
         try:
             for f in args.file:
                 if f == "-":
@@ -455,7 +457,7 @@ class PrefsControl(WriteableConfigControl):
                     for line in f:
                         if previous:
                             line = previous + line
-                        previous = self.handle_line(line, config, keys)
+                        previous = self.handle_line(line, new_config, keys)
                 finally:
                     if f != "-":
                         f.close()
@@ -463,6 +465,8 @@ class PrefsControl(WriteableConfigControl):
             raise
         except Exception, e:
             self.ctx.die(968, "Cannot read %s: %s" % (args.file, e))
+        for key, value in new_config.items():
+            config[key] = value
 
     @with_rw_config
     def edit(self, args, config, edit_path=edit_path):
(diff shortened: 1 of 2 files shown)


manics:
As expected it'll print out an error message if an incorrect key is found
```
$ ./bin/omero config load etc/hibernate.properties
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
Illegal property name: hibernate.search.default.directory_provider
```
Though by this point several other keys which happen to match the require regex may have already been set. This is still an improvement, but I'm unsure whether this should be merged this or if it should be put on hold so the key naming policy can be applied centrally.

mtbc:
Fair point: happy for others to decide. Maybe have to parse all lines before acting on any. Still, that's curious: Surely `hibernate.search.default.directory_provider` doesn't match the regex? I feel as if I am missing something.

manics:
It's the space before the `=` https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.9/etc/hibernate.properties#L19
`hibernate.search.default.directory_provider = org.hibernate.search.store.FSDirectoryProvider`

mtbc:
Aha, thank you. :+1:

mtbc:
Ah, I'm inferring that this `config` variable isn't a normal Python dictionary but some automagical backed-by-a-file thing.

manics:
Definitely a lot of automagic.... For mysterious reasons `omero config edit` with an invalid key seems to wipe the entire config!
```
$ bin/omero config set a.b c

$ bin/omero config get
a.b=c

$ bin/omero config edit  # Set an invalid property in the editor e.g. `e f=g`
Illegal property name: e f

$ bin/omero config get
# All properties are gone!
```

mtbc:
Wow, glad you noticed that, ta.

components/tools/OmeroPy/src/omero/plugins/prefs.py
```diff
@@ -482,11 +482,18 @@ def edit(self, args, config, edit_path=edit_path):
...
+        old_config = dict(config)
         self.drop(args, config)
         args.file = [open(str(temp_file), "r")]
         args.q = True
-        self.load(args, config)
-        remove_path(temp_file)
+        try:
+            self.load(args, config)
+        except Exception as e:
+            for key, value in old_config.items():
```
joshmoore, line 500: This could be: `config.update(old_config)`:

```
>>> a = {1:2}
>>> b = {1:10, 3:4}
>>> a.update(b)
>>> a
{1: 10, 3: 4}
```

joshmoore (commented):
A big :+1: on the `edit` catch. (_Mea culpa_) I'm not completely sure about the limitations in the regex. i.e. I can't be sure whether or not this will impact users. Potentially valid (though weird) characters might include: @, /, \. The only two other ideas I have to mitigate this are:
 * Add an argument to be less strict
 * Use a blacklist of values that we _know_ to be dangerous

components/tools/OmeroPy/src/omero/plugins/prefs.py
mtbc: `AttributeError: 'ConfigXml' object has no attribute 'update'`

joshmoore: My apologies. That would have had to been on the underlying dictionary.

mtbc:
Definitely inclined more toward a whitelist than blacklist to surely constrain what we need handle and maximize detection of what happened to @manics. Want me to ask on ome-devel if anybody's using weirdly named configuration properties to see if we need a `--lenient-parsing` or somesuch? (Would prefer not to offer it if nobody needs it, otherwise it might give them ideas!)

components/tools/OmeroPy/src/omero/plugins/prefs.py
mtbc: np, this `config` thing is tricksy!

joshmoore:
> . Want me to ask on ome-devel if anybody's using weirdly named configuration properties to see if we need a 

I definitely understand the motivation, but I'm not sure it's worth a breaking change in a point release.

> Want me to ask on ome-devel

I don't expect that to work.

mtbc:
> not sure it's worth a breaking change

Aha, fair enough, will close and can resurrect when we're thinking about a breaking release.

joshmoore:
I disagree. Definitely let's get the `config edit` change in, and a middle ground for black/white-listing.

mtbc:
I don't know what such a middle ground would be. `config edit` is safe in current mainline, no?

joshmoore:
> config edit is safe in current mainline, no?

Ah, only this PR introduced the problem? So no `mea culpa` was needed?? :smile:

> I don't know what such a middle ground would be

Perhaps just eliminating `=` and ` `?

mtbc:
Can perhaps revert that last for breaking release.

sbesson:
I tried to capture my understanding of the behavior that is addressed by this PR as unit tests currently failing against `develop` - see https://github.com/sbesson/openmicroscopy/commit/8faa34199309ea3b21a14b205feb2138a236e4c7 and https://travis-ci.org/sbesson/openmicroscopy/jobs/453872913.

Can you cherry-pick this commit so we confirm the unit tests work with the changes?

sbesson:
Pushed an extra commit to fix unit test failures although the `E=F=G` test case still fails to throw.

I realized we actually have some production use cases where values have an equal sign in it e.g. https://docs.openmicroscopy.org/omero/5.4.9/sysadmins/server-ldap.html#minimum-configuration.
It is probably safe to assume equal signs should be allowed in the property value only when loading them via `config load`?

mtbc:
Certainly makes sense to allow `=` in configuration values.

mtbc:
`bin/omero config set foo bar=baz` works just fine too.

manics:
`config set` still allows you to set invalid keys. This prevents `config edit` being used to add keys since it will detect the already existing invalid key and refuse to update.
```
$ bin/omero config set 'a b' c

$ bin/omero config get
a b=c

$ bin/omero config edit # Add a new valid key, or make no changes
 Illegal property name: a b
```
Though we could just accept this

sbesson:
Thanks @manics.  My vote is for further generalizing the new `testLoadWhitelist` and `testLoadInvalidKey` to test both paths (`load` and `set`, maybe `edit` ?) and enforce the same behavior.

joshmoore:
`$ bin/omero config set a=b c=d` didn't fail like I'd expect it to, no? (In fact, it looks to be a no-op)

cF
```
$ bin/omero config set "a b" "c d"
Illegal property name: a b
```

mtbc:
Interesting, yes, it's a no-op for 5.4.9 too.

mtbc:
I propose that we separate https://github.com/openmicroscopy/openmicroscopy/pull/5889#issuecomment-439416479 as being outwith this PR which worries only about spaces in property names. The other behavior seems to arise from #3567 and I'd rather step back and create a card for reviewing the intent rather than trying to attempt a quick fix here that might miss something important about the thinking behind that other PR.

joshmoore:
Works for me. Happy to have this in the next milestone.

mtbc:
Thanks: created https://trello.com/c/QyD1awAD/550-cli-config-set-no-op-bug so we can merge this.

jburel:
Merging
This will be included in m2

mtbc:
Can we tighten up the regex again here now if we're targeting 5.5.0 rather than 5.4.10?

manics:
Yes! Is there already a proposal somewhere in this PR?

mtbc:
Kind of the opposite of https://github.com/openmicroscopy/openmicroscopy/pull/5889/commits/7f238ca4043aa9579ced6314dec6e0f16b507ba5.

PR #5892, opened 2018-11-01
Public login button

chris-allan:
# What this PR does

When logged in as the public user the standard user profile is displayed in the top right hand corner of the main OMERO.web user interface. If a user wants to log in they have to semi-confusingly log out the public user in order to get to the log in page. This PR adds public user status to the OMERO.web request context and conditionally renders the user profile button as a more user friendly "Login" button instead when using the public user.

image

# Testing this PR

1. OMERO.web with a public user enabled should be available

2. When logged in as the public user a "Login" button should be displayed, otherwise the standard profile dropdown.

image

# Related reading

N/A

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -635,6 +635,13 @@ CUSTOM_SETTINGS_MAPPINGS = {
           "and link to it with the OMERO logo.")],
     "omero.web.login_view":
         ["LOGIN_VIEW", "weblogin", str, None],
+    "omero.web.user_dropdown":
+        ["USER_DROPDOWN",
+         "true",
+         parse_boolean,
+         ("Whether or not to include a user dropdown in the base template."
+          " Particularly useful when used in combination with the OMERO.web"
+          " public user where logging in may not make sense.")],
     "omero.web.staticfile_dirs":
         ["STATICFILES_DIRS",
          '[]',
--- a/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
@@ -212,3 +212,5 @@ class render_response(omeroweb.decorators.render_response):
             c_plugins.append({
                 "label": label, "include": include, "plugin_id": plugin_id})
         context['ome']['center_plugins'] = c_plugins
+
+        context['ome']['user_dropdown'] = settings.USER_DROPDOWN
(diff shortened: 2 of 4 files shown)


manics:
Would you consider it within scope of this PR to have a property to disable this, for example in public-only deployments like the IDR where login isn't allowed?

chris-allan:
Don't see why not. What would you like to see? No button at all?

manics:
I think no button would be good if it's not too much work to get it looking nice.

jburel:
@chris-allan Will you have time to work on the "no button" feature so this PR can be included in 5.4.10?

sbesson:
Chiming in late. The current state of things is certainly confusing and clarifying the state of the authentication make sense to me. Re-using the`Login` semantics in the top-right corner as proposed here is in agreement with the `webclient/login` UI. 

To add further to the confusion, as experienced by @francesw recently, if the data is owned by the public user itself, performing operations such as adjusting and saving the rendering settings require to login as the public user i.e. using the public user username/password from the main login page. In addition to the changes proposed here, this raises the question of whether our login page overloads the concept of `Log in/Login` and if we can find another terminology for the public connection. Note this might very well be an edge case outwith the scope of OMERO 5.4.10 as generally the public user will not own the data himself.

chris-allan:
@jburel: Might have time today depending on how a couple things go. I think this is an improvement regardless so depending on the opinion of @manics we might go ahead and get this in and then follow up with another PR. Up to you guys.

@sbesson: My general position would be that having the public user own *any* data is not a good idea. Lacking the ability to both prevent a user from writing anything (true read only) nor to make data world readable that's probably unrealistic right now.

/cc @joshmoore

jburel:
@chris-allan My vote will be to add the RFE in a follow-up PR, @manics what do you think?

manics:
My main concern is on the IDR this will lead to a `Login` button (that is blocked at the Nginx level so will redirect to the index page) instead of the drop-down which helps obscure logout/login.

Edit: Though if it'll be fixed in 5.4.11 we could temporarily revert this PR on the metadata54 branch.

chris-allan:
image

Behaviour when the user dropdown is completely turned off by setting the new `omero.web.user_dropdown` configuration property to `False`.

kennethgillen:
Possibly out of scope, but I'd expect to read "Log In", as opposed to "Login". i.e. we're not talking about a login as an entity, but the act of logging in. https://www.google.co.uk/search?q=log+in+vs+login for various examples.

chris-allan:
I agree with you @kennethgillen. However, we are using "Login" in several places as a verb and this PR is not about establishing grammatical correctness for verb vs. noun usage of that phraseology.

manics (approved):
Thanks, this looks great.

jburel:
Thanks all, merging

PR #5893, opened 2018-11-02
Add help message for login

jburel:
# What this PR does

Add a help button indicating that the username or session ID can be used to connect to omero

# Testing this PR

* Start insight and check that a help icon is displayed next to the text field
* Click on the help button and check that dialog pops up

Repeat the test using insight as an ImageJ plugin

cc @yukako @DrKenHo


components/insight/SRC/org/openmicroscopy/shoola/util/ui/login/ScreenLogin.java
```diff
@@ -71,6 +73,8 @@
 //Third-party libraries
 
 
+
+
```
mtbc: unintentional?

dominikl:
Works both in Insight standalone as well as IJ plugin 👍

PR #5895, opened 2018-11-02
Don't set trustStore properties if empty

manics:
# What this PR does

[`etc/omero.properties`](https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.9/etc/omero.properties#L142-L150) sets `omero.security.trustStore` and `omero.security.trustStorePassword` to empty. OMERO sets the corresponding system properties to empty.

Setting the system properties appears to override the default system truststore, leading to errors when a client attempts to verify a server.

This PR simply prints out an error instead of setting these properties to empty.

# Testing this PR

The easiest way is to get an OMERO build that supports AWS S3 imports since the minio client attempts to verify the S3 server. Without this PR S3 imports will fail on CentOS 7 (you'll need to check the server side import log to confirm the error) unless you set:
- `omero.security.trustStore=/etc/pki/ca-trust/extracted/java/cacerts`
- `omero.security.trustStorePassword=changeit`

These are the defaults for the Java trust-store on CentOS 7. You will have to figure out the corresponding values for other operating systems.

If this PR is included the two properties above are not required. The Blitz log will contain
```
2018-11-02 16:58:32,925 ERROR [                          omero.security] (
main) trustStore property is empty, not setting
2018-11-02 16:58:32,925 ERROR [                          omero.security] (
main) trustStorePassword property is empty, not setting
```

Note. east-ci is currently setup with these two properties. It should be possible to remove them if this PR is included.

### DIFF
--- a/components/server/src/ome/security/KeyAndTrustStoreConfiguration.java
+++ b/components/server/src/ome/security/KeyAndTrustStoreConfiguration.java
@@ -42,9 +42,17 @@ public class KeyAndTrustStoreConfiguration implements InitializingBean {
                 log.warn("Overwriting existing trust store: " + oldTrustStore);
             }
         }
-        System.setProperty(JAVAX_NET_SSL_TRUST_STORE, trustStore);
-        System.setProperty(JAVAX_NET_SSL_TRUST_STORE_PASSWORD,
-                trustStorePassword);
+        if (trustStore == null || trustStore.length() < 1) {
+            log.error("trustStore property is empty, not setting");
+        } else {
+            System.setProperty(JAVAX_NET_SSL_TRUST_STORE, trustStore);
+        }
+        if (trustStorePassword == null || trustStorePassword.length() < 1) {
+            log.error("trustStorePassword property is empty, not setting");
+        } else {
+            System.setProperty(JAVAX_NET_SSL_TRUST_STORE_PASSWORD,
+                               trustStorePassword);
+        }
 
         if (oldKeyStore != null) {
             if (oldKeyStore.equals(keyStore)) {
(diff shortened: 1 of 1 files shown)


components/server/src/ome/security/KeyAndTrustStoreConfiguration.java
```diff
@@ -42,9 +42,17 @@ public void afterPropertiesSet() throws Exception {
                 log.warn("Overwriting existing trust store: " + oldTrustStore);
             }
         }
-        System.setProperty(JAVAX_NET_SSL_TRUST_STORE, trustStore);
-        System.setProperty(JAVAX_NET_SSL_TRUST_STORE_PASSWORD,
-                trustStorePassword);
+        if (trustStore == null || trustStore.length() < 1) {
```
mtbc: note strings also have an `isEmpty()` and there're even utils like null-safe `StringUtils.isBlank`

mtbc:
So does this log a startup ERROR for a fairly default OMERO installation?

manics:
What would you recommend for logging?

mtbc:
I think I'd defer to @joshmoore and @kennethgillen here on the severity: certainly I've not noticed my OMEROs being deficient but maybe it is still really bad that they've not been setting values for this stuff and we need to improve our installation documentation accordingly?

components/server/src/ome/security/KeyAndTrustStoreConfiguration.java
manics: Thanks, I've changed it to use `StringUtils.isEmpty` so we've still got the option of using `[space]` to distinguish between empty and unset/default.

manics:
No one suggested a level so I've changed it to `WARN`:
```
2018-11-08 15:49:26,793 WARN  [                          omero.security] (      main) trustStore property is empty, not setting
2018-11-08 15:49:26,793 WARN  [                          omero.security] (      main) trustStorePassword property is empty, not setting
```

mtbc (approved):

PR #5898, opened 2018-11-07
ImageJ manual workflow

jburel:
# What this PR does

Add option to save roi and results on the original image

# Testing this PR

* Open an image in ImageJ
* Crop the image for example
* Draw a shape on the cropped image and add it to the data manager
* Click Save ROIs
* Message should indicate to save the image on the newly created if yes is selected. If no is selected the ROI should be saved on the original image. The same is true for the measurements.

# Related reading

See https://trello.com/c/t2O0BP32/19-fiji-manual-workflow-review

cc @pwalczysko

### DIFF
--- a/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/chooser/LocationDialog.java
+++ b/components/insight/SRC/org/openmicroscopy/shoola/agents/fsimporter/chooser/LocationDialog.java
@@ -21,6 +21,8 @@
  */
 package org.openmicroscopy.shoola.agents.fsimporter.chooser;
 
+import ij.IJ;
+
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Container;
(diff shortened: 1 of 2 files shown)


pwalczysko:
This feature works as described for both "Yes" and "No" answers in the dialogue.

Also tested the option of "Import images to OMERO" and it still works as expected.

What we need for this PR is to fix the dialog itself

screen shot 2018-11-08 at 10 14 15


I would suggest:

   - remove "Do you wish to" (saves space)
   - add "Yes - ..." and "No" in front of the options, such as ``Yes - the ROIs and results will be saved on the new images`` and ``No - ....``
   - ROI should be ROIs (plural)

pwalczysko:
Bug: This is a bug which was there prior to this PR:

   - answer "Yes" in the dialog descirbed in the header of the PR
   - import location window will open (see screenshot below)
   - in order to have the import started, nothing else helps but to click the red button in the top left of the dialog (then you see the import dialog appear and the import getting in progress immediately)
   - the Close and Add to Queue buttons seem not to do anything, the only way to get the workflow running again is to close the dialog window

We should definitely fix that I think, if you do not know what I write above, you will declare the workflow broken. But definitely this is not a fault of this PR

screen shot 2018-11-08 at 10 52 13

jburel:
@pwalczysko I finally managed to find the problem. It is not included in today's build

pwalczysko:
Built locally and tested all the workflows, also the "No" answer again, as well as "Close" after "Yes" and "Add to Queue" after "Yes. Also had a check on "Import Images to OMERO" menu item again.

All works fine and makes sense.

Ready to merge.

PR #5899, opened 2018-11-07
CLI handles being given directory instead of file argument.

mtbc:
# What this PR does

Fixes the nasty stack trace when giving OMERO.cli the name of a directory where it expected a file.

# Testing this PR

`bin/omero config load /etc`

should now better match the kind of error as from,

`bin/omero admin cleanse /etc/hosts`

# Related reading

https://trello.com/c/S0h8tKMv/527-cli-bin-omero-config-load-dir-fails


manics (approved):
Looks good to me

PR #5900, opened 2018-11-07
Have importer-cli more concisely report bad server hostname.

mtbc:
# What this PR does

Suppresses the stack trace printed for when `importer-cli` cannot look up a host name in the DNS.

# Testing this PR

Try `importer-cli` with `-s` options for good host names and bad ones like `-s wibble.wobble`. See that the bad host name error is not attached to a ginormous stack trace.

# Related reading

https://trello.com/c/THscwVDC/32-cli-report-wrong-server-nicely
https://docs.openmicroscopy.org/latest/omero5.4/users/cli/import.html#command-line-importer


joshmoore (approved):
LGTM

PR #5901, opened 2018-11-12
CLI diagnostics: make clear when OMERO log file is empty

mtbc:
# What this PR does

CLI diagnostics report log files that have a bit of text in them as "0.0 KB" which can be tempting to misread as "empty file" when just glancing through. This PR adjusts the output so that empty files look clearly different.

# Testing this PR

Try having some `var/log/*` files with nothing in and some with just a bit in and see that you can tell the difference from `bin/omero admin diagnostics`.

# Related reading

https://trello.com/c/qxJpamya/60-highlight-master-log-entries

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1015,6 +1015,8 @@ OMERO Diagnostics (%s) %s
         else:
             if not p.exists():
                 self.ctx.out("n/a")
+            elif not p.size:
+                self.ctx.out("empty")
             else:
                 warn_regex = ('(-! )?[\d\-/]+\s+[\d:,.]+\s+([\w.]+:\s+)?'
                               'warn(i(ng:)?)?\s')


manics (approved):
Looks fine:
```
Log files:  Blitz-0.log                    n/a
Log files:  DropBox.log                    n/a
Log files:  FileServer.log                 n/a
Log files:  Indexer-0.log                  n/a
Log files:  MonitorServer.log              n/a
Log files:  Processor-0.log                n/a
Log files:  Tables-0.log                   n/a
Log files:  TestDropBox.log                n/a
Log files:  master.err                     empty
Log files:  master.out                     0.0 KB
Log files:  Total size                     0.00 MB
```

Future suggestion: It looks like there's already a mix of units in use, so to make it even more obvious that 0.0 and 0.00 don't mean 0 maybe we could use use bytes for small log files?

jburel:
@manics Could you capture the RFE in a card?

manics:
https://trello.com/c/LFJpyDfr/44-omero-admin-diagnostics-log-size-use-human-readable-units

jburel:
Thanks @manics

PR #5902, opened 2018-11-12
Have import log include specific report of target container.

mtbc:
# What this PR does

https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8518&start=50#p19654 -

> Is there a way how I can get the dataset ID back after the bulk import?

This PR includes that information in what the CLI importer reports to the user.

# Testing this PR

Try CLI imports using the `-T` option. Check that the specific target container is printed to the console during import. For bonus points, check that the container is reported even if the importing user does not have permission to put images into that container. On `eel` an easy way is to have `user-2` import into `read-only-1` then have `user-4` try importing to that same target container.

# Related reading

https://docs.openmicroscopy.org/latest/omero5.4/users/cli/import-target.html

### DIFF
--- a/components/blitz/src/ome/formats/importer/ImportLibrary.java
+++ b/components/blitz/src/ome/formats/importer/ImportLibrary.java
@@ -291,6 +291,12 @@ public class ImportLibrary implements IObservable
                                     try {
                                         IObject obj = target.load(store, ic);
                                         if (!(obj instanceof Annotation)) {
+                                            Class <? extends IObject> targetClass = obj.getClass();
+                                            while (targetClass.getSuperclass() != IObject.class) {
+                                                targetClass = targetClass.getSuperclass().asSubclass(IObject.class);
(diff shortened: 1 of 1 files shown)


joshmoore:
```
$ bin/omero import -T Dataset:name:josh-pr-5902 /tmp/a.fake
...
2018-11-14 09:39:00,973 2126       [      main] INFO          ome.formats.importer.ImportConfig - Using import target: Dataset:name:josh-pr-5902
2018-11-14 09:39:01,047 2200       [2-thread-1] INFO         ome.formats.importer.ImportLibrary - Import target specifies container: Dataset:498
```

Seems fine. Something that we _might_ consider is whether "Dataset:498" should get printed to standard out/err like the other created objects, but I'd say let's save that for a follow-on. (There are several other issues around targets if I remember correctly)

Merging for 5.4.10.

PR #5903, opened 2018-11-12
CLI: obj map-set can delete keys

mtbc:
# What this PR does

Extends the CLI plugin to be able to delete keys from map annotations.

# Testing this PR

`omero obj map-set MapAnnotation:1234 mapValue mykey` now deletes any `mykey` entry from the map annotation with ID 1234. It is not an error if the key does not exist in the map.

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_obj/TestObj/test_map_mods/ should pass.

# Related reading

https://trello.com/c/XHVymmdY/43-delete-a-map-key

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/obj.py
+++ b/components/tools/OmeroPy/src/omero/plugins/obj.py
@@ -284,24 +284,29 @@ class MapSetTxAction(NonFieldTxAction):
 
         from omero.model import NamedValue as NV
 
-        if len(self.tx_cmd.arg_list) != 5:
-            ctx.die(335, "usage: map-set OBJ FIELD KEY VALUE")
+        argc = len(self.tx_cmd.arg_list)
+        if argc not in [4, 5]:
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/src/omero/plugins/obj.py
```diff
@@ -284,24 +284,29 @@ def on_go(self, ctx, args):
...
-            if nv and nv.name == name:
-                nv.value = value
-                state = "SET"
-                break
-
-        if state != "SET":
-            current.append(NV(name, value))
+        if argc == 4:
+            name = self.tx_cmd.arg_list[3]
+            current = filter(lambda nv: nv and nv.name != name, current)
```
manics: A list comprehension is more "pythonic":
`current = [nv for nv in current if (nv and nv.name != name)]`

manics (commented):
One suggested code change, but works as expected.

manics (approved):

PR #5896, opened 2018-11-05
Idr 0.6.0 rc1

sbesson:
- Merge `v5.4.9` tag into the IDR OMERO integration branch
- Bump IDR Bio-Formats version to 0.5.0 - see https://trello.com/c/bUgDRt5b/2-idr-bio-formats-050

PR #5904, opened 2018-11-12
Replace all usages of DefaultMetadataOptions by DynamicMetadataOptions

sbesson:
## What this PR does

Noticed while https://github.com/openmicroscopy/openmicroscopy/pull/5896 and reviewing the current divergence between the `metadata54` branch of openmicroscopy and the latest 5.4.9 release - see https://github.com/openmicroscopy/openmicroscopy/compare/v5.4.9...metadata54.

File differences between the IDR branch and the latest mailine tag can be classified into 3 groups:
1. scripts are disabled into IDR
1. IDR still uses a custom Bio-Formats version - related to the `ImportCandidates` unit tests exclusion
1. IDR uses `DynamicMetadataOptions` rather than `DefaultMetadataOptions`. The former is a more generic implementation of the MetadataOptions allowing support for key/value pairs.

This PR proposes to further reduce this divergence by switching the mainline OMERO `ImportCandidates` API to use `DynamicMetadataOptions`.  The class is identical on the mainline Bio-Formats as well as IDR Bio-Formats - https://github.com/openmicroscopy/bioformats/compare/v5.9.2...IDR:IDR-0.5.2.

## Testing this PR

Changes should only be happening client-side and there is currently no mechanism to allow key/value pair options to be passed to `ImportCandidates`. Thus I expect passing unit and integration tests to be the main item to review. Memo files of a 5.4.9 server should not be affected (although this might be worth checking).

As per https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.9/components/blitz/src/ome/formats/importer/ImportCandidates.java#L113, there is a way to control the `MetadataLevel` within the ImportCandidates via system property although I do not know of a reader where the list of files would vary depending on this level.

### DIFF
--- a/components/blitz/src/ome/formats/importer/ImportCandidates.java
+++ b/components/blitz/src/ome/formats/importer/ImportCandidates.java
@@ -25,7 +25,7 @@ import loci.formats.IFormatReader;
 import loci.formats.MissingLibraryException;
 import loci.formats.UnknownFormatException;
 import loci.formats.UnsupportedCompressionException;
-import loci.formats.in.DefaultMetadataOptions;
+import loci.formats.in.DynamicMetadataOptions;
 import loci.formats.in.MetadataLevel;
 import ome.formats.ImageNameMetadataStore;
 import ome.formats.importer.util.ErrorHandler;
@@ -423,7 +423,7 @@ public class ImportCandidates extends DirectoryWalker
                 reader.close();
                 reader.setMetadataStore(new ImageNameMetadataStore());
                 reader.setMetadataOptions(
-                        new DefaultMetadataOptions(METADATA_LEVEL));
+                        new DynamicMetadataOptions(METADATA_LEVEL));
                 reader.setId(path);
                 format = reader.getFormat();
                 usedFiles = getOrderedFiles();
--- a/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
+++ b/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
@@ -19,7 +19,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import loci.formats.in.DefaultMetadataOptions;
+import loci.formats.in.DynamicMetadataOptions;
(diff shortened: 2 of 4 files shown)


mtbc:
> Changes should only be happening client-side ...

It is indeed my impression that this PR does *not* affect any server-side code paths. :+1:

joshmoore:
The simple rename matches my expectations.

jburel:
Did you check if the memo files were not affected?

mtbc:
I didn't.

manics:
No idea whether this is relevant, but whilst refactoring [`FilePatternReader`](https://github.com/openmicroscopy/bioformats/blob/v6.0.0-m2/components/formats-bsd/src/loci/formats/in/FilePatternReader.java) in https://github.com/openmicroscopy/bioformats/pull/3250 I noticed it does not override [`FormatHandler.setMetadataOptions`](https://github.com/openmicroscopy/bioformats/blob/v6.0.0-m2/components/formats-api/src/loci/formats/FormatHandler.java#L111) to pass it to the helper. Should it, or is it reader specific?

joshmoore:
Not exactly the answer you're looking for, but if I remember correctly, propagation via `setMetadataOptions` is pretty critical for the `wrap`ping logic.

jburel:
Checked the following steps as suggested by @sbesson:
without this PR:
* Imported an image
* Memofile created
* opened Image 
```
2018-11-19 13:48:17,116 DEBUG [                   loci.formats.Memoizer] (l.Server-9) loaded memo file: /OMERO/BioFormatsCache/OMERO/ManagedRepository/root_0/2018-11/19/13-47-45.282/.4D-series.ome.tif.bfmemo (71519 bytes)

```
 with this PR
* Re-built the server
* Opened the same image

```
2018-11-19 13:54:26,101 DEBUG [                   loci.formats.Memoizer] (l.Server-6) loaded memo file: /OMERO/BioFormatsCache/OMERO/ManagedRepository/root_0/2018-11/19/13-47-45.282/.4D-series.ome.tif.bfmemo (71519 bytes)
```

PR #5910, opened 2018-11-19
Allow cli_login to be used without closing session

manics:
If a CLI session is created outside of the script calling `cli_login` it's annoying for the session to be closed. `cli_login` is still a useful function as it handles all the usual CLI login steps.

Note this may be broken: For some reason using `with cli_login(close=True) as cli:` in a python script causes the script to hang when it's finished (I've added `print "Finished"` to verify it hangs at the very end and not just after the cli_login step).

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1554,10 +1554,12 @@ def cli_login(*args, **kwargs):
     passed to onecmd
 
     kwargs:
-      - keep_alive
+      - keep_alive (keep alive interval, default 300)
+      - close (close afterwards, default True)
     """
 
     keep_alive = kwargs.get("keep_alive", 300)
+    close_after = kwargs.get("close", True)
     try:
         cli = omero.cli.CLI()
         cli.loadplugins()
@@ -1573,7 +1575,8 @@ def cli_login(*args, **kwargs):
                 raise Exception("Failed to login")
         yield cli
     finally:
-        cli.close()
+        if close_after:
+            cli.close()
 
 
 def argv(args=sys.argv):
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1577,6 +1577,7 @@ def cli_login(*args, **kwargs):
         yield cli
     finally:
         cli.close()
+        client.stopKeepAlive()
 
 
 def argv(args=sys.argv):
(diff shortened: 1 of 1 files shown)


joshmoore:
That'll be the result of `keep_alive`, since it creates threads that don't exit. You may need to shut that down regardless. Additionally, you can use `detachOnDestroy` so that the session isn't harmed.

manics:
I couldn't see a way to stop the keep-alive. `detachOnDestroy` with `close` seems to work though

joshmoore:
Did `stopKeepAlive` not work? (see https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/clients.py#L646)

manics:
Thanks, I was looking at the Java client which doesn't have that https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.0-m1/components/blitz/src/omero/client.java#L819

joshmoore:
Do you want to relist this?

sbesson:
Syntactically the proposed changes make sense, look useful and backwards compatible i.e. an extra opt-in parameter that can be passed to the context manager allowing to detach your sessions before closing.

I am having a hard time reproducing the behavior this PR tries to address. I tried to create a few test cases where a session is first created and then passed to multiple successive invocations of `cli_login` e.g.:

```
from omero.cli import cli_login

conn = BlitzGateway("public", "public", host="localhost", port=4064)
conn.connect()

key = str(conn.getSession().uuid.val)
print key

for i in range(4):
    with cli_login("-s", "localhost", "-u", "public", "-k",  key) as cli:
        cli.invoke("hql -q 'select count(p) from Project p'")
```

From the description, I was expecting the first invocation to pass but the session to be closed at the second pass and to be prompted for a password, but I got:

```
[sbesson@test64-omeroreadonly-1 ~]$ python test.py 
00f82362-eaf1-4ab5-96db-1ea3947e7d6e
 # | Col1 
---+------
 0 | 24   
(1 row)
 # | Col1 
---+------
 0 | 24   
(1 row)
 # | Col1 
---+------
 0 | 24   
(1 row)
 # | Col1 
---+------
 0 | 24   
(1 row)
!! 05/24/19 12:24:33.374 error: communicator not destroyed during global destruction.
```

Also tested a variant of the script above by creating a session using `bin/omero login` instead of the BlitzGateway and I got the same result.

Do you have a minimal test that reproduces the initial issue (probably something we would like to formalize as an integration test too)?

jburel:
Test required. Comment from @sbesson indicates the difficulty to reproduce the point addressed by this PR. Excluded for 5.5.

joshmoore:
@manics : do you see this still being transferred? Was this closed by accident?

manics:
Oops, I was cleaning up old branches and assumed `patch-1` was irrelevant 😀

joshmoore:
I.e. re-opening later?

manics:
Maybe? I'm wondering if our cli/login/blitzgateway is already complicated enough without more additions

joshmoore:
K. Leaving that consideration to you.

PR #5912, opened 2018-11-20
OMERO.web default log can be changed to stdout

manics:
OMERO.web can be configured to send its default logs to stdout instead of a file. This is useful when running in the foreground, for example in Docker. The default behaviour is unchanged.

For example:
```
omero config set omero.web.default_log_handlers '["stdout"]'
omero web start --foreground
```
You can also have multiple loggers `'["default", "stdout"]'`

Ideally this would be reviewed by someone who understands exactly how settings and logging works in Django.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -140,6 +140,11 @@ LOGGING = {
             'class': 'logging.StreamHandler',
             'formatter': 'standard'
         },
+        'stdout': {
+            'level': 'DEBUG',
+            'class': 'logging.StreamHandler',
+            'formatter': 'standard'
+        },
         'mail_admins': {
             'level': 'ERROR',
             'filters': ['require_debug_false'],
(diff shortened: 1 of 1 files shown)


joshmoore:
I like the last commit, but it might require a review of the behavior before and after https://github.com/openmicroscopy/openmicroscopy/commit/6223dd595d41a6b480925b5534520c6808e2c02e to understand potential consequences.

manics:
I haven't seen `OMEROweb_brokenrequest.log` on any server for years, nor is it generated by omero-weberror. Are you sure that's still enabled?

joshmoore:
> Are you sure that's still enabled?

Guess that's what I mean by a review.

joshmoore:
Worth removing references and moving forward with this? Would definitely be nice to have this in docker.

jburel:
Review required before including. Not considered for 5.5

manics:
I've rewritten this so there are no changes to the existing logging config, instead an `stdout` section is added which mostly duplicates the `console` section. It should now be clear from static code inspection that the default value for `omero.web.default_log_handlers` should result in the same logging config as without this PR.

jburel:
Removing the exclude label so that this PR can be retested. The label was added during the 5.5.0 PR filtering.

joshmoore:
@manics : do you see this still being transferred? Was this closed by accident?

manics:
It was .... reopened at https://github.com/ome/omero-web/pull/196

PR #5913, opened 2018-11-20
web check_version defaults to ignoring minor version

manics:
Installation of OMERO.web is decoupled from OMERO.server, so we shouldn't be so strict about comparing versions. This compares the major version only, ~and changes the property `omero.web.check_version` to be a boolean (for backwards compatibility) or integer.~

~- `0`/`false`: same as before (no check)~
~- `1`/`true`: new default (compare major versions only)~
~- `2`: old default (strict check)~

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/connector.py
+++ b/components/tools/OmeroWeb/omeroweb/connector.py
@@ -221,11 +221,9 @@ class Connector(object):
         finally:
             connection.close()
 
-    def check_version(self, useragent, strict=False):
+    def check_version(self, useragent):
         """
-        Check whether the client and server version match.
-        If strict this is an exact match, otherwise only compare major
-        versions
+        Check whether the server version is compatible with this client
         """
         connection = self.create_guest_connection(useragent)
         if connection is None:
@@ -240,8 +238,9 @@ class Connector(object):
                 client_version = client_version.group(1).split('.')
                 logger.info("Client version: '%s'; Server version: '%s'"
                             % (client_version, server_version))
-                if strict:
-                    return server_version == client_version
+                # Compatibility is determined by matching the major version
+                # If OMERO moves to semver only the first element will need
+                # to be checked
                 return server_version[:2] == client_version[:2]
             except:
                 logger.error('Cannot compare server to client version.',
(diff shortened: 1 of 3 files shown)


sbesson:
In general :+1: for extending this logic to express the fact patch releases of the same minor version should be compatibe i.e. OMERO.web 5.4.x and OMERO.server 5.4.y. While we currently allow ourselves to break the API on minor version increments of OMERO, we might want to evolve towards semver in the near future. While this is not applicable immediately, how would this affect this delination? Would we keep three levels of validation i.e. no check, major version check, full check?

manics:
Personally I'd prefer to keep it as a boolean, and just have no-check vs major-check, but I went with an integer in case people want the option of an strict check. If everyone's happy to not have a strict check I'll change it back to a boolean.

I think the default should always be to check the major version, so `5.5.*` with the current versioning, `6.*.*` with semantic versioning. If we kept the integer levels then you could have 2=minor-check, 3=patch-check, etc.

sbesson:
Personally, I would also be happy with keeping it as a boolean i.e. check or no check and leave the question about how many digits should be compared as an implementation detail of the API (i.e. relax it to comparing only the first 2 digits for now and when we move to semver, further relax the check).

From the consumer perspective, this solution would be advantageous as it involves no change to the configuration of production deployments and things will just work when upgrading independently the web deployment or server to a compatible version.

joshmoore:
Travis loves the irony:

```
     [exec] E           Exception: Unsupported OMERO version: 5.5.0-m2-ice35-SNAPSHOT
```

joshmoore:
The change at this point is quite straight-forward, but having not been the one to add the strict check I'd suggest we get a final sign off from @openmicroscopy/webapps folks.

Are there plans to bring the tertiary logic back at some point?

manics:
> Are there plans to bring the tertiary logic back at some point?

No. If there's a good use case then I think we should add it now, and deprecate the boolean (but keep backwards compatibility).

will-moore:
I think the strict check was originally there to mimic the behaviour of Insight. Then Ola added the option to disable it, primarily for devs since we often found we had to comment out the version check to connect to various servers.

Re "we currently allow ourselves to break the API on minor version increments of OMERO" above: we don't allow ourselves to break the OMERO API on minor versions, only the web api, and then only on some undocumented ```/webclient/``` URLS.

sbesson:
I meant minor version as defined by https://semver.org/.

will-moore:
Code looks and works fine.
I can't think of a need to support more strict (patch) version checking.
Good to merge.

PR #5915, opened 2018-11-23
Keep a large base font size configured in browser from breaking tag dialog

knabar:
# What this PR does

When a minimum font size of 12px or larger is configured in the browser, the second list box in the tagging dialog wraps and appears misaligned below where it should be. This PR changes the CSS to keep the list boxes aligned independent of any configured minimum font size.

Dialog with 13px minimum font size configured:
screen shot 2018-11-23 at 09 25 04

With this PR included:
screen shot 2018-11-23 at 09 37 59

# Testing

To set the minimum font size in Chrome, go to Settings>Appearance>Customize Fonts>Minimum font size

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
@@ -2646,7 +2646,6 @@ button#id_add_new_tag {
     height: 25px;
 }
 div#id_move_tags {
-    float: left;
     width: 40px;
     padding: 120px 0 40px;
     margin: 10px;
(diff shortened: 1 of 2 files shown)


dgault:
Without the PR I was able to reproduce the wrapping effect as mentioned when increasing the font size.

With the PR included I can increase the font size way up without any alignment issues:

screen shot 2018-11-26 at 15 28 15


(Tested on Macbook with Chrome and Firefox)

chris-allan:
Thanks, @dgault. I should emphasise that there are lots of *other* parts of OMERO.web that react pretty poorly to even a minimum font size of 12 and we'll have to consider what, if anything, we want to do about that another time.

/cc @jburel, @pwalczysko

jburel:
Thanks for fixing this issue.
Something to adjust on a case-by-case basis

PR #5839, opened 2018-08-30
Add Executor.Priority.BACKGROUND

joshmoore:
The addition of parallel import has made it quite easy
to use up the omero.db.pool_size number of PostgreSQL
connections available to the server. To prevent this,
the ManagedImportRequests that are submitted on import
now go to a lower priority pool of threads.

# Testing this PR

1. with parallel-fileset, import a number of filesets concurrently
2. watch that other clients/users can still login regardless of the import progress
3. if desired, set `omero.threads.background_threads` to one and watch that import is *serial*
4. if desired, set `omero.threads.background_timeout` to something small to see client errors

PR #5918, opened 2018-11-30
move clients' graph requests onto background threads

mtbc:
When clients submit async requests without a priority set they are assigned the usual default. This PR instead defaults graph requests like move and delete to be confined to the background threads that #5839 introduced to the server. See https://trello.com/c/fEtAzcrB/528-move-chgrp-chown-to-executorprioritybackground.

### DIFF
--- a/components/blitz/src/omero/cmd/SessionI.java
+++ b/components/blitz/src/omero/cmd/SessionI.java
@@ -147,7 +147,14 @@ public class SessionI implements _SessionOperations {
 
     public void submit_async(AMD_Session_submit __cb, omero.cmd.Request req,
             Ice.Current current) {
-        submit_async(__cb, req, current, null); // null == USER priority
+        final Executor.Priority priority;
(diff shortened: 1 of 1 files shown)


joshmoore (commented):
Code looks fine for the interim, but longer-term, I'd suspect we can/should make use of the fact that as of [line 171](https://github.com/openmicroscopy/openmicroscopy/pull/5918/files#diff-6cfab6ac2cf1a7a63cf55e3abdcfb296R171) you're guaranteed to have a [IRequest](https://github.com/openmicroscopy/openmicroscopy/blob/c8a5e40af0ec5d4da369e91d03d9ab571a912360/components/blitz/src/omero/cmd/IRequest.java) and therefore that interface could give you the priority.

mtbc:
Cool, was hoping to do something in time to help annotation deletion not disable IDR staging servers for the next round. Will leave the card as-is on the board so we can revisit when we have time to think about how best to change the API.

PR #5919, opened 2018-11-30
CLI: further restrict names of configuration properties

mtbc:
# What this PR does

Limits config key names to only certain characters.

# Testing this PR

First commit should fail Travis, next commit should pass.

# Related reading

Discussion on #5889.

### DIFF
--- a/components/tools/OmeroPy/src/omero/plugins/prefs.py
+++ b/components/tools/OmeroPy/src/omero/plugins/prefs.py
@@ -263,7 +263,7 @@ class PrefsControl(WriteableConfigControl):
     def assert_valid_property_name(self, key):
         from re import search
 
-        if search(r'[\s]', key):
+        if search(r'[^A-Za-z0-9._-]', key):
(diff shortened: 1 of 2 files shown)


joshmoore:
Two separate formats that we deal with to consider:
 * https://doc.zeroc.com/ice/3.6/properties-and-configuration/configuration-file-syntax
 * https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader-

mtbc:
It's worth noting that the `\`-EOL continuation etc. is handled outwith `assert_valid_property_name`.

manics (approved):
Looks good to me

joshmoore:
@mtbc : you'll also handle the upgrade documentation page to warn users?

mtbc:
Sure: you're thinking the "what's new for developers"?

joshmoore:
That, and also "sysadmins". But no, not "users" in that sense.

joshmoore:
Thanks for https://trello.com/c/To2eki0T/51-limit-chars-for-config-keys

PR #5922, opened 2018-12-01
Drop ice35

jburel:
# What this PR does

Remove support for Ice 3.5.

# Testing this PR

Check that Travis is green.

# Related reading

https://trello.com/c/aCLtBiHl/17-ice-35-omero


mtbc:
Maybe can delete `components/tools/OmeroWeb/requirements-py27-*ice35.txt`.

jburel:
Thanks I thought I got rid of it :-)

mtbc:
Probably outwith this PR but I wonder if we still need to set Ice.Default.EncodingVersion to 1.0.

joshmoore:
There should be a note somewhere about when we can drop that; could well be it's now.

jburel:
https://doc.zeroc.com/ice/3.7/the-ice-protocol/protocol-and-encoding-versions

components/tools/OmeroPy/src/omero/clients.py
```diff
@@ -187,39 +187,30 @@ def _initData(self, id):
...
-                self._optSetProp(id, "IceSSL.Ciphers", "NONE (DH_anon.*AES)")
-            else:
-                self._optSetProp(id, "IceSSL.Ciphers", "ADH")
+        if sys.platform == "darwin":
+            self._optSetProp(id, "IceSSL.Ciphers", "NONE (DH_anon.*AES)")
         else:
             self._optSetProp(id, "IceSSL.Ciphers", "ADH")
+
         self._optSetProp(id, "IceSSL.VerifyPeer", "0")
         self._optSetProp(id, "IceSSL.Protocols", "tls1")
```
mtbc, line 208: Wondering if this should be `TLS1_2`!

mtbc:
Thank you, looks to me to be good to merge.

components/tools/OmeroPy/src/omero/clients.py
jburel: This will require testing

PR #5923, opened 2018-12-04
Disable refresh button while existing refresh in progress

chris-allan:
# What this PR does

Having multiple refreshes of the tree in flight at one time can result
in JavaScript errors, duplicate elements in the tree, and other race
condition like behaviour.  This disables the button while the refresh is
in progress.

Previously it was possible to create situations such as the following
by being rather aggressive with the mouse clicks:

image

# Testing this PR

1. OMERO.web with a reasonable amount of data loaded. Preferably with a read-only or read-annotate group with data from multiple users.

2. Double or multi-click the "Refresh" button in the left hand panel above the tree

3. No duplicate sets of the hierarchy

**Known issue:** This PR does not guard against refreshes that can be triggered while the tree is in the middle of performing other loading actions. Hitting refresh during these periods may result in JavaScript errors being printed to the console.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.tree.js
@@ -351,6 +351,10 @@ $(function() {
         });
         // Clear refreshPathsReverse after selection has been restored
         refreshPathsReverse = [];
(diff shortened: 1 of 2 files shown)


chris-allan:
https://github.com/openmicroscopy/openmicroscopy/pull/5923

rgozim:
Looks good to merge 👍

PR #5933, opened 2018-12-16
Web requirements

jburel:
# What this PR does

Introduce new files so we can use wheel

# Testing this PR

Check that travis is green
# Related reading

While working on omeroweb-install, I realised that we cannot use the new zeroc-ice-py wheels and the requirements file. 
This PR introduces new requirements files so we do not install python bindings via pip

### DIFF
--- a/components/tools/OmeroWeb/requirements-py27-all-noice.txt
+++ b/components/tools/OmeroWeb/requirements-py27-all-noice.txt
@@ -1,8 +1,9 @@
 # Python installation requirements for OMERO.web
 # ==============================================
 #
-#     pip install -r requirements-py27.txt
+#     pip install -r requirements-py27-all-noice.txt
 #
 
-zeroc-ice>3.5,<3.7
+numpy>=1.9
+Pillow<3.4
 -r requirements-py27-noice.txt
--- a/components/tools/OmeroWeb/requirements-py27-noice.txt
+++ b/components/tools/OmeroWeb/requirements-py27-noice.txt
@@ -1,0 +1,12 @@
+# Python installation requirements for OMERO.web
+# ==============================================
+#
+#     pip install -r requirements-py27-noice.txt
+#
+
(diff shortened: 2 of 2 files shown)


joshmoore:
@jburel : where are each of these 3 files going to be used from? It feels like one too many.

jburel:
The new files can now be used with the wheel and simplify the installation requirements
The original file is kept so it does not break any existing installation
or can be used without the wheel if desired

manics:
I thought the original requirements files should still work as long as the wheel is already installed:
```
[root@0818f7db6b37 /]# pip install https://github.com/openmicroscopy/zeroc-ice-py-centos7/releases/download/0.0.3/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
Collecting zeroc-ice==3.6.4 from https://github.com/openmicroscopy/zeroc-ice-py-centos7/releases/download/0.0.3/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
  Downloading https://github.com/openmicroscopy/zeroc-ice-py-centos7/releases/download/0.0.3/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl (23.7MB)
    100% |################################| 23.8MB 54kB/s
Installing collected packages: zeroc-ice
Successfully installed zeroc-ice-3.6.4

[root@0818f7db6b37 /]# pip install 'zeroc-ice>3.5,<3.7'
Requirement already satisfied (use --upgrade to upgrade): zeroc-ice<3.7,>3.5 in /usr/lib64/python2.7/site-packages
```

jburel:
I was trying to potentially avoid confusion
i.e. installing wheel then having another "pip install ice" command

jburel:
Closing this PR. I still think we need to avoid the confusion. 
Maybe a comment in the requirements file will be enough

PR #5936, opened 2019-01-05
redirect to login_view after logout

atarkowska:
# What this PR does

Fix invalid redirection after logout. This is only exposed when developer has to customise the view

# Testing this PR

There were no test prior to that fix.

Without fix test is should fail as follow
```
>       assert rsp['Location'].endswith(reverse('test_weblogin'))
E       assert <built-in method endswith of str object at 0x112f11db0>('/test_login/')
E        +  where <built-in method endswith of str object at 0x112f11db0> = 'http://testserver/webclient/login/'.endswith
E        +  and   '/test_login/' = reverse('test_weblogin')
```

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -333,10 +333,10 @@ def logout(request, conn=None, **kwargs):
                 logger.error('Exception during logout.', exc_info=True)
         finally:
             request.session.flush()
-        return HttpResponseRedirect(reverse("weblogin"))
+        return HttpResponseRedirect(reverse(settings.LOGIN_VIEW))
     else:
         context = {
-            'url': reverse('weblogout'),
+            'url': reverse(settings.LOGIN_VIEW),
             'submit': "Do you want to log out?"}
         t = template_loader.get_template(
             'webgateway/base/includes/post_form.html')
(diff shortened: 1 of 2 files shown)


components/tools/OmeroWeb/omeroweb/webclient/views.py
```diff
@@ -333,10 +333,10 @@ def logout(request, conn=None, **kwargs):
                 logger.error('Exception during logout.', exc_info=True)
         finally:
             request.session.flush()
-        return HttpResponseRedirect(reverse("weblogin"))
+        return HttpResponseRedirect(reverse(settings.LOGIN_VIEW))
     else:
         context = {
-            'url': reverse('weblogout'),
+            'url': reverse(settings.LOGIN_VIEW),
```
will-moore, line 339: I think this needs to remain unchanged, since this is the URL that the ```webclient/logout``` page tries to POST in order to logout. If this uses ```settings.LOGIN_VIEW``` and this is not set, default is ```"weblogin"```. Then, clicking on the "Do you want to log out?" on the  ```webclient/logout``` page tries to POST to the weblogin page with no form data and you get redirected to the login page with all the error messages displayed (and we're not actually logged out).

With this unchanged as ```reverse('weblogout'),``` the logout behaves as expected and you are then redirected to ```settings.LOGIN_VIEW```.

will-moore:
Apart from comment above, this works fine and test passes.

joshmoore:
https://web-proxy.openmicroscopy.org/west-ci/job/OMERO-training/397/console is still failing. Before merging, can someone definitely say that this PR is not related?

atarkowska:
I am not sure if this PR is included as the original branch was deleted. Replaced by #6024

PR #5289, opened 2017-05-10
Cors settings

will-moore:
# What this PR does

--cors

This replaces #5252 and instead of installing ```django-cors-headers``` app, it merely adds the essential settings, allowing users to install https://github.com/ottoyiu/django-cors-headers and configure remaining settings themselves.

The relies on #5269 to allow setting of middleware.

# Test setup:

These steps should be performed and deployed by devspace at http://10.0.51.146/web/api/
so you can skip this for testing.

1. Install app: ``` pip install django-cors-headers```
2. Add app: ``` bin/omero config append omero.web.apps '"corsheaders"' ```
    - NB: This is called for in app docs above, but doesn't seem to be necessary?!
3. Add middleware. Middleware order with respect to existing middleware must follow [rules](https://github.com/ottoyiu/django-cors-headers), e.g:
```
bin/omero config append omero.web.middleware '{"index": 0.5, "class": "corsheaders.middleware.CorsMiddleware"}'
bin/omero config append omero.web.middleware '{"index": 10, "class": "corsheaders.middleware.CorsPostCsrfMiddleware"}'
```

4. Configure middleware to allow connections from specified origins:
```
bin/omero config append omero.web.cors_origin_whitelist '"localhost:8000"'
```
OR from all origins
```
bin/omero config set omero.web.cors_origin_allow_all True
```

# Test CORS connection:

 - Checkout this repo to get the example code:

```
cd openmicroscopy/examples/Training/javascript/

python -m SimpleHTTPServer
```

 - Go to http://localhost:8000/, open devtools and follow the Network tab to see what is happening.
 - Enter the url http://10.0.51.146/web/ and click "Connect" (this devspace deploy has been configured to accept "localhost:8000" as shown above)
 - This will GET available servers, CSRF token and login url and show a login screen.
 - Login uses POST (should see the pre-flight OPTIONS request) and then lists Projects.
 - Connecting from origins other than "localhost:8000" should fail. e.g. different port:
``` python -m SimpleHTTPServer 8080```

# Related reading

Link to cards, tickets, other PRs:

1. background for understanding this PR

2. what this PR assists, fixes, or otherwise affects

PR #5937, opened 2019-01-07
omero.web.django_additional_settings

manics:
# What this PR does

Adds a new web property `omero.web.django_additional_settings`. This is a list of `[DJANGO_PROPERTY_NAME, json-value]` pairs that can be used to set or override Django settings that aren't managed by OMERO.web.

# Testing this PR

For example, if you install an additional non-ome web app and need to set some settings [(e.g. advanced whitenoise settings)](http://whitenoise.evans.io/en/stable/django.html#available-settings):
- without this PR you need to make code changes to `omeroweb/settings.py`
- with this PR you just add the property to `omero.web.django_additional_settings`:
```
omero config append omero.web.django_additional_settings '["WHITENOISE_STATIC_PREFIX", "/static/"]'
```

Another example: https://github.com/openmicroscopy/openmicroscopy/pull/5289 added settings to support CORS headers in the optional `django-cors-headers` app. This could have been done without a PR using this property.

One more example: completely override the OMERO.web log configuration and log to stdout:
```
omero config append --set omero.web.django_additional_settings '["LOGGING",
    {
      "version": 1,
      "disable_existing_loggers": false,
      "formatters": {
        "standard": {
          "format": "%(asctime)s %(levelname)5.5s [%(name)40.40s] (proc.%(process)5.5d) %(funcName)s():%(lineno)d %(message)s"
        }
      },
      "handlers": {
        "console": {
          "level": "DEBUG",
          "class": "logging.StreamHandler",
          "formatter": "standard"
        }
      },
      "loggers": {
        "": {
          "handlers": ["console"],
          "level": "DEBUG",
          "propagate": true
        }
      }
    }
  ]'
```

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -777,6 +777,14 @@ CUSTOM_SETTINGS_MAPPINGS = {
          parse_boolean,
          ("If True, cors_origin_whitelist will not be used and all origins "
           "will be authorized to make cross-site HTTP requests.")],
+
+    "omero.web.django_additional_settings":
+        ["DJANGO_ADDITIONAL_SETTINGS",
+         "[]",
+         json.loads,
+         ("Additional Django settings as list of key-value tuples. "
+          "Use this to set or override Django settings that aren't managed by "
+          "OMERO.web. E.g. ``[\"CUSTOM_KEY\", \"CUSTOM_VALUE\"]``")],
 }
 
 DEPRECATED_SETTINGS_MAPPINGS = {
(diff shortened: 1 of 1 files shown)


will-moore:
Sounds like a great idea!

joshmoore:
Sorry for having dropped this. Also a :+1: for not needing to open a PR for every new variable. What's the intent if both values are set? i.e. what's the precedence? Also, moving forward to we see _preferring_ this method or is it temporary until a new release?

manics:
`omero.web.django_additional_settings` is parsed at the end and always has priority over everything else, so that there's the option of overriding the omero.web built-ins if necessary (e.g. as in the `LOGGING` example above).

New `omero.web.*` property vs `django_additional_settings`:
- New `omero.web` property: a Django property that's used by OMERO.web and is generally useful to others
- `django_additional_settings`: Django properties for Django add-ons that aren't part of OMERO.web, or properties which no-one else is likely to change

joshmoore:
No further objections yesterday. Merging. :+1:

PR #5938, opened 2019-01-08
in server log note when integration tests happen

mtbc:
# What this PR does

Has the server log note which integration test happens when. This helps to figure out what server activity relates to which test. *However* it does also increase log size and clutter: is it worth it?

# Testing this PR

On eel see `/opt/hudson/workspace/OMERO-DEV-merge-integration/src/dist/var/log/Blitz-0.log`.

### DIFF
--- a/components/tools/OmeroJava/test/integration/AbstractServerTest.java
+++ b/components/tools/OmeroJava/test/integration/AbstractServerTest.java
@@ -6,6 +6,7 @@ package integration;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


joshmoore:
Thinking about this, no objection to having this in place to give us immediate feedback. The only immediate impact will be the ~2000 x 2 x runtime overhead and a slightly larger log file, but no worries.

To do this more completely though, it will be useful to think about how to combine this with a tracing package (like zipkin or jaeger) such that the client (here the integration tests) does something to communicate the context (here the class and method names) to the server so that all of the related backend threads can be queried.

mtbc:
```console
[mtbcarroll@ome-c6100-1 ~]$ grep ' log cmd.* - test ' /opt/hudson/workspace/OMERO-DEV-merge-integration/src/dist/var/log/Blitz-0.log | wc -l
4950
```

joshmoore:
Was a pretty good estimate... ;)

PR #5942, opened 2019-01-16
Fix configuration unicode usage

joshmoore:
`bin/omero config set` and `load` calls failed spectacularly when
presented with unicode characters. Hand-editing etc/grid/config.xml
allowed the values to propagate to Java. This PR attempts to allow
these values to be set via the command-line.

# Testing this PR (for the keen)

Following the IRC conversation:

 1. setup an organization unit with a unicode value
 2. try to set omero.ldap.base to match the OU from the CLI
 3. start the server and create a user in that OU

# Related reading

see: https://trello.com/c/WO3qatIc/589-bug-bin-omero-config-doesnt-handle-unicode-values

### DIFF
--- a/components/tools/OmeroPy/src/omero/config.py
+++ b/components/tools/OmeroPy/src/omero/config.py
@@ -25,8 +25,7 @@ sys = __import__("sys")
 
 import xml.dom.minidom
 
-from xml.etree.ElementTree import XML, Element, SubElement, Comment
-from xml.etree.ElementTree import tostring
+from xml.etree.ElementTree import XML, Element, ElementTree, SubElement, Comment
 from omero_ext import portalocker
 import json
 
@@ -329,7 +328,9 @@ class ConfigXml(object):
     def write_element(self, icegrid):
         temp_file = path.path(self.filename + ".temp")
         try:
-            temp_file.write_text(tostring(icegrid, "utf-8"))
+            # Copying etree usage from ome-model
+            with open(temp_file, "w") as o:
+                ElementTree(icegrid).write(o, encoding="UTF-8")
             if sys.platform == "win32":
                 os.remove(self.filename)
             temp_file.rename(self.filename)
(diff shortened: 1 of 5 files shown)


mtbc:
:+1: for the tests given the risk of regression in moving to Python 3.

components/tools/OmeroPy/src/omero/config.py
```diff
@@ -25,8 +25,7 @@
 
 import xml.dom.minidom
 
-from xml.etree.ElementTree import XML, Element, SubElement, Comment
-from xml.etree.ElementTree import tostring
+from xml.etree.ElementTree import XML, Element, ElementTree, SubElement, Comment
```
mtbc: This file needs some flake8.

manics:
`omero config edit` fails with
`UnicodeEncodeError: 'ascii' codec can't encode characters in position 54-56: ordinal not in range(128)`

joshmoore:
@manics : when setting a unicode during the edit or when is one already set? That being said, this currently only covers get/set/load.

manics:
When one is already set:
```
$ bin/omero config get
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
a=£££
omero.web.debug=true
omero.web.server_list=[["eel.openmicroscopy.org",4064,"oɹǝɯo"]]
omero.web.ui.top_links=[["ɐʇɐ◖", "webindex", {"title": "Browse Data via Projects, Tags etc"}], ["ʎɹoʇsıH", "history", {"title": "History"}], ["dʃǝH", "https://help.openmicroscopy.org/", {"title":"Open OMERO user guide in a new tab", "target":"new"}]]
```
```
$ bin/omero config edit
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
  File "bin/omero", line 130, in <module>
    rv = omero.cli.argv()
...
UnicodeEncodeError: 'ascii' codec can't encode characters in position 54-56: ordinal not in range(128)
```

joshmoore:
Pushed an edit fix. Still pondering the best way to test.

manics (approved):
Looks good.

Unrelated bug found whilst testing this: https://trello.com/c/zxBq8U9C/608-omero-config-edit-incorrectly-handles-backslash

PR #5934, opened 2018-12-19
Decoupled

joshmoore:
This PR removes the component jars and their build systems in favor of https://github.com/ome/omero-build

 * [x] server-jars
 * [x] Python: see https://github.com/ome/omero-blitz/issues/3 and https://github.com/ome/omero-blitz/issues/4
 * [x] Insight
 * [x] tests

PR #5943, opened 2019-01-18
Insight and CLI remove "anon" from disabled algorithms.

mtbc:
# What this PR does

On startup has Insight and command-line importer remove "anon" from among the value of the `jdk.tls.disabledAlgorithms` security property.

# Testing this PR

Insight should be able to connect to servers even with current Java installed: version 8u201 or later.

`bin/omero import ...` should work similarly.

# Related reading

https://trello.com/c/q7we5yYn/68-insight-ssl-algorithms

### DIFF
--- a/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
+++ b/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
@@ -1,5 +1,5 @@
 /*
- *   Copyright (C) 2009-2018 University of Dundee & Open Microscopy Environment.
+ *   Copyright (C) 2009-2019 University of Dundee & Open Microscopy Environment.
  *   All rights reserved.
  *
  *   Use is subject to license terms supplied in LICENSE.txt
@@ -13,6 +13,7 @@ import gnu.getopt.LongOpt;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.security.Security;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -40,9 +41,13 @@ import omero.grid.ImportProcessPrx;
 import omero.model.Annotation;
 import omero.model.CommentAnnotationI;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.time.StopWatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
 
 /**
  * The base entry point for the CLI version of the OMERO importer.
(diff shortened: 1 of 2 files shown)


mtbc:
Best to copy-and-paste code until after the dust from #5934 settles.

mtbc:
For `setProperty` could have caught `SecurityException` but:
* Possible that this isn't something we're likely to run into. (Maybe would have been an issue for webstart.)
* Not easy to handle in any *useful* way.
* Seemed best to keep changes as simple as possible for an emergency fix.

dominikl:
It's tricky to test. Tested so far:
- Windows 10, Java 8 201: Cannot connect to nightshade with 5.4.9. Can connect with OMERO-DEV-merge build to eel though, so fix seems to work (unforunatenly can't test nightshade with the 5.5.0 OMERO_DEV merge build).

jburel:
Connecting to eel is enough I would think

pwalczysko:
>Connecting to eel is enough I would think

But I thought we want to do 5.4.10 insight -> we need to build it anyway ?

pwalczysko:
Tested
  - Mac OS to eel with the java 8 update 201, went fine

pwalczysko:
Tested insight

   - Windows 7 laptop to eel with the java 8 update 201, went fine
   - Windows 7 laptop to eel with java 8 update 60, went fine (to test working with older java updates)
   - Mac OS to eel with the java 8 update 201, went fine

jburel:
enough to test the fix I mean

dominikl:
Tested Insight on Linux (Debian) was well 👍 Unfortunately can't test CLI, no idea how to get Ice 3.6 for Debian.

jburel:
@dominikl https://github.com/ome/omero-install/blob/develop/linux/step01_debian9_ice_deps.sh

pwalczysko:
Tested:

   - Mac OS CLI import to eel using java 8 update 201, went fine
   - Mac OS CLI import to eel using java 8 update 65, went fine

dominikl:
Ok, finally managed to get the right Java (11.0.2) installed on debian. The fix doesn't seem to work for the CLI:
screen shot 2019-01-21 at 14 47 31

pwalczysko:
@dominikl Isn't that maybe because you are running the linux in VB ? I would guess @mtbc had tested his fix on Linux in the first place ?

jburel:
@dominikl https://docs.openmicroscopy.org/omero/5.4.9/sysadmins/troubleshooting.html
See OpenSSL section

dominikl:
Thanks. With the debian OpenSSL patch applied it works (5.4.9 fails, 5.5.0 dev build works) 👍

pwalczysko:
I think we have expected results on 3 OS-es, Insight and CLI. 

Ready to merge FMPOV

joshmoore:
Sorry for the late comment, but would this have been better placed directly in `omero.client` or was your goal to hit `main` methods to make it only happen once?

mtbc:
I tend to be leery of touching that, especially as it's complex and mirrored in Python and C++; I also wasn't sure if it would always be wanted by every client instantiation (could be a utility method called from main methods?). Time's short and this is tested so probably okay for a 5.4.10? Also happy to refactor as you like for 5.5.0: was going to solicit suggestions on that.

joshmoore:
Understood. The downside is that other consumers will _not_ get the fix, but that may be acceptable. Fine to see refactoring in 5.5.

mtbc:
Superseded by #5947.

PR #5945, opened 2019-01-22
add failing integration test for import target canLink()

mtbc:
Adds an integration test that should pass once https://trello.com/c/kWgFImSC/450-bug-import-target is fixed. See https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-broken/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_import/. At present we should expect a failure iff `is_owns_newest` is False.


joshmoore:
Purposes seems straight-forward enough. Happy to have this merged at any time.

PR #5645, opened 2018-02-05
CLImporter: allow forcing encrypted connection

joshmoore:
By default, the CLI importer uses the TCP port (unencrypted)
for performance reasons. If no TCP port is available, however,
clients should be able to use the SSL port (encrypted). This
is now possible with the `--encrypted=true` flag.

# Testing this PR

1. Find a host which only has the 4064 port (potentially with prefix) available. (This may be most manageable via ssh port forwarding. `ssh -L 4064:localhost:4064 somelocation`)
2. Attempt to import with `--encrypted=true`: **pass**
3. Attempt to import with no `--encrypted` flag: pass but prints warning below
4. Attempt to import with `--encrypted=false` (the default): pass but prints warning below

```
2018-02-19 10:45:02,981 1947       [      main] INFO       ome.formats.OMEROMetadataStoreClient - Insecure connection requested, falling back
2018-02-19 10:45:03,085 2051       [      main] WARN       ome.formats.OMEROMetadataStoreClient - Cannot fallback. Using secure connection
```

PR #5946, opened 2019-01-23
Python gateway secure

manics:
# What this PR does

Python BlitzGateway and OMERO.web default to using secure connections.

Previously the defaults for secure/insecure were confusing: https://trello.com/c/pJuzsmKU/441-bug-blitzgateway-connconnect

This makes both default to secure. The main advantage is you can connect to any OMERO.server with just the standard 4064 port exposed or forwarded, you no-longer need to remember to set secure.

# Testing this PR

Everything should continue to just work. You should be able to connect to servers from the BlitzGateway and OMERO.web with or without 4063 exposed with no change in configuration, previously the default configuration would have failed if 4063 was blocked.

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -1488,7 +1488,7 @@ class _BlitzGateway (object):
 
     def __init__(self, username=None, passwd=None, client_obj=None, group=None,
                  clone=False, try_super=False, host=None, port=None,
-                 extra_config=None, secure=False, anonymous=True,
+                 extra_config=None, secure=True, anonymous=True,
                  useragent=None, userip=None):
         """
         Create the connection wrapper.
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -453,7 +453,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
           "sessions/#using-cached-sessions>` for more details.")],
     "omero.web.secure":
         ["SECURE",
-         "false",
+         "true",
          parse_boolean,
          ("Force all backend OMERO.server connections to use SSL.")],
     "omero.web.session_cookie_age":


will-moore:
Code change looks fine, and the rationale makes sense.
But I would think it needs a wider discussion about changing the default behaviour of web and BlitzGateway.
E.g. What are the performance implications of using ```secure=True```?
cc @chris-allan @joshmoore

chris-allan:
The performance implications, especially under high load, are significant as OMERO.web creates new connections and destroys them on every request. I would be **very** careful about this change. It has always been the expectation that TCP/4064 and TCP/4063 are required opened ports for proper OMERO client operation.

joshmoore:
The only exception I can think of is the importer which we taught to suffice with only 4064 if 4063 is not accessible: see https://github.com/openmicroscopy/openmicroscopy/pull/5645

I imagine the same strategy would work here.

manics:
If the concern is that production instances of OMERO.web with a large number of users may be slowed would it be enough to add it to the production OMERO.web docs?

joshmoore:
I could see adding it to the docs one way or the other, but do you have an objection to the fallback mechanism?

manics:
A fallback mechanism would work, but goes against the general trend of everyone moving to requiring secure connections by default.

jburel:
The insight PR has been closed. There is no import in Web but some (quick) tests reveal a major degradation for import.

joshmoore:
Ok. There seem to be several things going on here. Trying to summarize my understanding:

 * @simon says: "main advantage is you can connect to any OMERO.server with just the standard 4064 port exposed or forwarded," which can be achieved either by this change or a fallback.
 * but @chris-allan and @jburel have expressed performance concerns.
 * @simon suggests updating the docs to recommend that others set this for themselves.
 * and more generally, @simon is proposing the general policy of SSL by default. 
 * @jburel has pointed out that a decision to do so in insight has been postponed.

Assuming we agree on the above, is everyone more for tying the future of this PR to https://github.com/ome/omero-insight/pull/29 or for trying to achieve intermediate steps?

jburel:
My vote is to do that work on both java and python at the same time.

rgozim:
Not really much I can add to this, but if performance is degraded on imports, when that is something we're conscious should be faster (therefore as fast as we can get it to go) then I would not recommend setting `secure` to `True`. Not at least until we find a way to make imports faster regardless of security settings.

jburel:
The corresponding insight PR https://github.com/ome/omero-insight/pull/29  has been closed. 
Testing reveals some significant impacts on some calls e.g. import
The suggestion was to tackle both clients at the same time. This PR won't be considered for 5.5.

manics:
Ported to
- https://github.com/ome/omero-py/pull/189
- https://github.com/ome/omero-web/pull/142

PR #5944, opened 2019-01-18
disable TestHdfStorage.testMaskColumn unit test

mtbc:
Disables `TestHdfStorage.testMaskColumn` to get Travis passing.

PR #5947, opened 2019-01-24
have Java clients remove "anon" from disabled algorithms

mtbc:
# What this PR does

On initialization has clients remove "anon" from among the value of the `jdk.tls.disabledAlgorithms` security property.

# Testing this PR

Insight should be able to connect to servers even with current Java installed: version 8u201 or later.

`bin/omero import ...` should work similarly.

# Related reading

https://trello.com/c/q7we5yYn/68-insight-ssl-algorithms
https://github.com/openmicroscopy/openmicroscopy/pull/5943#issuecomment-456857831

### DIFF
--- a/components/blitz/src/omero/client.java
+++ b/components/blitz/src/omero/client.java
@@ -27,6 +27,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
+import java.security.Security;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -68,6 +69,11 @@ import Glacier2.CannotCreateSessionException;
 import Glacier2.PermissionDeniedException;
 import Glacier2.SessionNotExistException;
 import Ice.Current;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
+
+import org.apache.commons.lang.StringUtils;
 
 /**
  * Central client-side blitz entry point. This class uses solely Ice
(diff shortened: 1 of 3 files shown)


joshmoore:
Only quick thought based on ongoing conversations is whether or not `omero.security.anonymous=true` should disable the new workaround. (A more complicated version would be `omero.security.allow_ciphers=a,b,c`) Another option would be to do that outside of ICE_CONFIG and use a specialized environment variable.

mtbc:
It's a good idea but for an emergency patch I'd keep things real simple so I'd suggest targeting that for 5.5.

joshmoore:
Understood. Someone can always downgrade back to 5.4.9 if this has any issues. Just waiting on the doc push, then let's merge and build.

sbesson:
c3d8ba7 cherry-picked the single commit from https://github.com/openmicroscopy/openmicroscopy/pull/5944 to get at least Travis passing on the newly created `dev_5_4` branch, prior to merging this and building a 5.4.10 release candidate

joshmoore (approved):
The same code worked well in insight and the importer, so :+1: for moving it to all Java clients. "Disabling" the workaround will be equivalent to not using 5.4.10 which seems fair enough.

sbesson:
Thanks all. Merging for the creation of first 5.4.10 release candidate.

chris-allan:
Just as a bit of feedback after the fact here and with #5949 on the go these changes result in there being a Guava dependency now for clients. While using `com.google.common.base.Splitter` I'm sure made this implementation a tad easier, it does seem a bit excessive to be pulling in that entire dependency for those purposes.

Anyone who has been using exclude lists to keep their dependency tree in check as an OMERO client will need to take this into account now.

mtbc:
https://github.com/openmicroscopy/openmicroscopy/pull/5947#issuecomment-457159417 noted on https://trello.com/c/2LUZ8TPg/597-make-anon-ciphers-configurable.

PR #5949, opened 2019-01-28
Java SSL fix for command-line importer

mtbc:
Extends the 5.4.10 Java SSL fix to `bin/omero import ...`. See #5947 and https://trello.com/c/MywCIQL9/60-omero-cli-import.

* Moves the `omero.client` fix even earlier in execution.
* Has import initialize `omero.client` before doing upgrade check.

### DIFF
--- a/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
+++ b/components/blitz/src/ome/formats/importer/cli/CommandLineImporter.java
@@ -159,14 +159,6 @@ public class CommandLineImporter {
                 // config.requestFromUser(); // stdin if anything missing.
                 usage(); // EXITS TODO this should check for a "quiet" flag
             }
-
-            if (config.checkUpgrade.get()) {
-                config.isUpgradeNeeded();
-            }
-            else
-            {
-                log.debug("UpgradeCheck disabled.");
-            }
             store = config.createStore();
             store.logVersionInfo(config.getIniVersionNumber());
             reader.setMetadataOptions(
@@ -174,6 +166,12 @@ public class CommandLineImporter {
 
             library = new ImportLibrary(store, reader,
                     transfer, exclusions, minutesToWait);
+
+            if (config.checkUpgrade.get()) {
+                config.isUpgradeNeeded();
+            } else {
+                log.debug("UpgradeCheck disabled.");
+            }
         }
 
         Runtime.getRuntime().addShutdownHook(new Thread() {
--- a/components/blitz/src/omero/client.java
+++ b/components/blitz/src/omero/client.java
@@ -70,11 +70,6 @@ import Glacier2.PermissionDeniedException;
 import Glacier2.SessionNotExistException;
 import Ice.Current;
 
-import com.google.common.base.Joiner;
-import com.google.common.base.Splitter;
-
-import org.apache.commons.lang.StringUtils;
-
 /**
  * Central client-side blitz entry point. This class uses solely Ice
  * functionality to provide access to blitz (as opposed to also using Spring)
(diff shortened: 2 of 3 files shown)


joshmoore:
@mtbc : did you find out why the non-static version didn't work?

mtbc:
It probably does. I just figured it was better to do it earlier and only once but I think simply doing the initialization before the upgrade check suffices.

joshmoore:
> I just figured it was better to do it earlier and only once

:+1:

> I think simply doing the initialization before the upgrade check suffices.

Interesting, so something in the SSL subsystem is getting initialized with the wrong algorithms when the HTTP call is made?

mtbc:
> ... something in the SSL subsystem is getting initialized with the wrong algorithms when the HTTP call is made?

That theory agrees with my observations.

mtbc:
See the static initializer following https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.10/components/common/src/ome/system/UpgradeCheck.java#L57.

sbesson:
If I understand https://github.com/openmicroscopy/openmicroscopy/pull/5949#issuecomment-458128311 correctly, does that mean that calling `bin/omero import --skip=upgrade` might be a workaround to the Java SSL update issue using OMERO 5.4.10?

mtbc:
Good question. Just empirically verified: I *can* import using that workaround where without `--skip=upgrade` I get `SSLHandshakeException: No appropriate protocol`.

pwalczysko:
Happy to do the review, but do we have this built somewhere ? I am noticing this branch is 5_4, so this is not the ci merge build, is it ?
If to be built locally, do I use the old build system ? The new one ?

mtbc:
@sbesson will have to comment on CI artifacts but locally you can build with the old build system.

mtbc:
Any strong opinions on if this is worth a 5.4.11 soon or can wait for an eventual 5.5.0?

jburel:
@pwalczysko to build you will need the "old" build system.

Having a 5.4.11 will probably be useful for people using CLI. In that case it is probably good to include some insight fixes too (imageJ plugin problem) ahead of the workshop. 
It depends on people bandwidth (limited at the moment)

Those changes will need to be ported to the new repo

mtbc:
Porting noted on https://trello.com/c/q7we5yYn/68-insight-ssl-algorithms#comment-5c4f2831af8ca16d30f2a89e.

manics:
This will be needed on the IDR `metadata54` branch so it'll be good to have sign-off on this PR when it's ready.

mtbc:
Can fix the 5.4.11 history entry either if releasing or in rebase to 5.5.0.

pwalczysko:
Mac OS X:

  - Negative test with java 201 and old CLI client ended with import into outreach ``Ice.SecurityException: javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate``
  - positive test with the CLI built from this PR passed the import test on outreach for 
     - java 8 update 65 and
     - java 8 update 201

This passes the test

pwalczysko (approved):

joshmoore (approved):
* :+1: for removal of guava
 * :+1: for the static block
 * 😖 by the SSL weirdness in general, but thanks for sleuthing that

Only thing I can thing to figure out is whether or not to have the README updated in terms of user expectations of a release. Otherwise, LGTM

sbesson:
:+1: for having this fix in place. We have not heard from anyone hitting the CLI importer bug so far however no objection to having this merged into `dev_5_4` line and rebased onto `develop`. Regarding distribution, as discussed on Tuesday, this is largely dependant on the OMERO 5.5.0 timeline i.e. we can decide whether the fix will be shipped in OMERO 5.5.0 or schedule an OMERO 5.4.11 release with the fix in the upcoming weeks.

Main question is related to 11f6f60. So far, the `history.rst` file was  updated at release time and never included unreleased notes. Would that be rebased as it is on `develop` or combined into the 5.5.0 history? I don't mind switching behavior and starting to pre-populate this file but this is something new so we probably want the entire dev team to agree on the expectation.

mtbc:
It seemed better to compose the release text at the time everything was freshly in mind. If no 5.4.11 then I think it just becomes part of the latest milestone's 5.5.0 text upon rebase. (I won't be rebasing anyway until the decoupling effort settles.) I figured for the IDR rebase the code commits alone suffice.

sbesson:
Thanks. Merging and dabee37 can be reverted if we decide to go ahead with OMERO 5.4.11.

Issue #4657, opened 2016-05-12
[feature request] download link for correct version of client on webclient login page


carandraug:
The webclient login page could have a link to download the insight client for the correct server version.

Assuming that omero users shouldn't have to worry about what is the underlying server version, it becomes the job of the omero admin to keep updating any internal documentation pointing to the right version of the client.

This would remove the need for such internal documentation or save the user from googling for omero client download and ending up downloading the wrong version.

joshmoore:
@will-moore @rgozim : see an easy enough way to add this?

manics:
The login page includes the OMERO.web version so it should be easy to add a link https://github.com/openmicroscopy/openmicroscopy/blob/v5.4.10/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/login.html#L137

joshmoore:
Maybe `[linux] [mac] [win]` with the redirects?

```
RewriteRule ^/latest/omero5.4/insight-win.zip$ /omero/5.4.10/artifacts/OMERO.insight-5.4.10-ice36-b105-win.zip [R=seeother,L]
RewriteRule ^/latest/omero5.4/insight-mac.zip$ /omero/5.4.10/artifacts/OMERO.insight-5.4.10-ice36-b105-mac.zip [R=seeother,L]
RewriteRule ^/latest/omero5.4/insight-linux.zip$ /omero/5.4.10/artifacts/OMERO.insight-5.4.10-ice36-b105-linux.zip [R=seeother,L]
```

PR #5954, opened 2019-02-15
Add client downloads links to omeroweb login page

manics:
# What this PR does

Adds download links to the latest compatible insight release. Enabled by default, can be disabled by setting `omero.web.login.show_client_downloads=false`. The download base URL can be changed using `omero.web.login.client_downloads_base`

Closes https://github.com/openmicroscopy/openmicroscopy/issues/4657

- [x] CSS styling

\[I would've tried the new [draft PR feature](https://github.blog/2019-02-14-introducing-draft-pull-requests/) but it's not working on this repo\]

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -678,6 +678,18 @@ CUSTOM_SETTINGS_MAPPINGS = {
           "For example: ``'{\"redirect\": [\"webindex\"], \"viewname\":"
           " \"load_template\", \"args\":[\"userdata\"], \"query_string\":"
           " {\"experimenter\": -1}}'``")],
+
+    "omero.web.login.show_client_downloads":
+        ["SHOW_CLIENT_DOWNLOADS",
+         "true",
+         parse_boolean,
+         ("Whether to link to official client downloads on the login page")],
+    "omero.web.login.client_downloads_base":
+        ["CLIENT_DOWNLOAD_LATEST_BASE",
+         'https://downloads.openmicroscopy.org/latest/omero',
+         str,
+         ("Base URL for latest client downloads")],
+
     "omero.web.apps":
         ["ADDITIONAL_APPS",
          '[]',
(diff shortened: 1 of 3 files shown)


will-moore:
One simple option is to add the links like this (which is not too in-your-face):

screen shot 2019-02-18 at 04 20 29


See https://github.com/will-moore/openmicroscopy/commit/4a228bb3168a620eeb2b8c8e6e99d29a32b0f85a

mtbc:
If we adopt this then our workflows for production server upgrades should gain a step of updating `omero.web.login.client_downloads_base` unless we can instead default to using the latest for the installed major release.

manics:
@mtbc It's automatically determined: https://github.com/openmicroscopy/openmicroscopy/blob/814dd65394ff17d7f025da192ef5c6887f11d54f/components/tools/OmeroWeb/omeroweb/webclient/views.py#L264-L265

manics:
@will-moore Cherry-picked, thanks!

manics:
Note since the links are to the major version of the running OMERO.web they won't work until 5.5 is released.

joshmoore:
screen shot 2019-02-19 at 10 50 04

Looks good. See the mouseover. Only thing I can think of relating to your last comment, @manics, is whether or not the addition of `5.5` to the URL should happen at the level of `omero.web.login.client_downloads_base` i.e. that the "API" so to speak is just that at whatever URL there must be a `insight-{mac,win,linux}.zip`

manics:
It should be a *compatible* `insight-{mac,win,linux}.zip` that present at the URL, otherwise we're exchanging a download error for a runtime error.

I could make `omero.web.login.client_downloads_base` into a template with optional fields such as `https://downloads.openmicroscopy.org/latest/omero{major}.{minor}`. Do you think it's worthwhile, or unnecessary complexity that's unlikely to be used?

joshmoore:
> It should be a _compatible_ `insight-{mac,win,linux}.zip` that present at the URL, otherwise we're exchanging a download error for a runtime error.

I don't understand. From my reading, the `$major.$minor` gets appended regardless, and I'm asking if that should be moved up one level to the property.

joshmoore:
I like the template solution and the end-result looks the same. (We can separately consider if we want to hard-code `omero5.4` until 5.5 is ready)

Thanks @manics!

joshmoore:
(It also occurs to me that this likely needs a documentation PR)

PR #5955, opened 2019-02-18
OMERO Python gateway: allow to set rendering of inactive channels

sbesson:
## What this PR does

This addresses an issue noticed while applying the CLI render plugin to recent IDR studies.  The `render set` command wraps the `set_active_channels` API of the Python gateway. The current behavior of this API is that if a channel is disabled (by setting `-index` in the `channels` list), its rendering settings are never updated by this API.

While this in inline with the docstring contract (although incorrect the example fixed here), the original motivation is unclear. All the arguments for setting channel rendering  (`colors`, `windows`, `invertMaps`) handle `None` as way to skip updating the relevant settings. Thus the following API call unambiguously communicates the fact a second channel should be disabled with no other rendering change:

```
gateway.set([1, -2], windows=[[0.0, 255.0], None], colors=['FFF0000', None])
```

A contrario, if a caller made the effort to pass rendering settings for the disabled channel, these parameters are completely ignored at the moment:

```
gateway.set([1, -2], windows=[[0.0, 255.0], [10.0, 20.0], colors=['FFF0000', '00FF00'])
```

Additionally there is no other high-level way using the gateway to modify rendering settings of disabled channels

## Proposed changes

61f4cf9 proposes a minimal non-breaking approach by adding an extra key/value argument allowing the consume to choose  the behavior for disabled channels. If `false` (default), the current behavior is unchanged and disabled channels are only disabled. If `true`, all valid rendering settings are also applied to the disabled channels if applicable.

Other options might be worth discussing including adding additional `ImageWrapper` or `Channelwrapper` API to be able to manipulate channel rendering settings like lookup tables.

## Testing this PR

1e6c8b3 adds an integration tests covering the expectation of this API (incl. the simple testing of the channel disabling functionality). https://github.com/search?q=set_active_channels+user%3Aome+user%3Aopenmicroscopy+user%3AIDR&type=Code lists the known internal places that consume this API and should be reviewed with these changes in mind.

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py
@@ -522,6 +522,46 @@ class TestRDefs (object):
         g = gatewaywrapper.gateway
         assert not g._assert_unregistered("testSetActiveChannelsWithRE")
 
+    @pytest.mark.parametrize('set_inactive', [True, False])
+    def test_set_active_channels_set_inactive(
+        self, gatewaywrapper, set_inactive):
+        """
+        Tests set_active_channels method with rendering engine
+        """
+        # Clean potentially customized default
+        self.image.clearDefaults()
+        self.image._closeRE()
+        self.image = gatewaywrapper.getTestImage()
+        c0wmin = self.image.getChannels()[0].getWindowMin()
+        self.channels = self.image.getChannels()
+        assert self.c0color != 'F0F000'
+        assert self.c1color != '000F0F'
+        assert c0wmin != 0
+        self.image.set_active_channels(
+            [1, -2], windows=[[0.0, 1631.0], [409.0, 5015.0]],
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/test/integration/gatewaytest/test_rdefs.py
```diff
@@ -522,6 +522,46 @@ def testSetActiveChannelsWithRE(self, gatewaywrapper):
         g = gatewaywrapper.gateway
         assert not g._assert_unregistered("testSetActiveChannelsWithRE")
 
+    @pytest.mark.parametrize('set_inactive', [True, False])
+    def test_set_active_channels_set_inactive(
+        self, gatewaywrapper, set_inactive):
```
dominikl: flake8 seems to complain about the indention here.

dominikl:
Looks good, apart from the flake8 complaint. Tested with a local python script, works like expected. With having `set_inactive=False` as default, can't see that anything could break, so good to merge from my point of view.

PR #5956, opened 2019-02-19
Deprecate the CLI upload module and plugin

sbesson:
The plugin and module is getting decoupling and will be installed from https://pypi.org/project/omero-cli-upload.

See discussion on https://github.com/ome/omero-cli-upload/pull/1#issuecomment-465117050


manics:
Since 5.5.0 is a major release how about remove it completely instead of deprecating it?

sbesson:
Apologies for the delay in answering https://github.com/openmicroscopy/openmicroscopy/pull/5956#issuecomment-467800057. Similar to what we did with `omero-metadata`, my vote is for officially deprecating this module in the OMERO.py 5.5.0 release and dropping it in the following major (semver-wise) release.

jburel:
The approach taken here is the one we have previously followed. We cannot drop it without previous warning major release or not.
I agree with @sbesson

jburel:
Ran ```bin/omero upload fileName``` against west-ci
no warning printed out
i.e.
```
Created session for root@idr1-slot3.openmicroscopy.org:54064. Idle timeout: 10 min. Current group: system
OriginalFile:211508
```

joshmoore:
:+1:

re: removing / decoupling - what's missing from my understanding is how do we guarantee that _every_ OMERO install has a set of given plugins. If that were the case, I'd assume deleting would be equivalent to just moving the jars to new repositories. It might be that `pip install omero-py` or `pip install omero-cli` is the right time to do that.

manics:
Make it mandatory in the installation instructions? And switch the omero-server Ansible role to install it into a virtualenv along with a default set of plugins?

joshmoore:
That's certainly an option.

joshmoore:
I can confirm what @jburel saw that no warning was printed:

```
bash-4.2$ OMERO_DEV_PLUGINS=1 bin/omero -d1 upload foo
Using session for root@localhost:4064. Idle timeout: 10 min. Current group: system
File: foo does not exist
bash-4.2$ bin/omero upload foo
usage: bin/omero [-h] [-v] [-d DEBUG] [--path PATH] [-C] [-s SERVER] [-p PORT]
                 [-g GROUP] [-u USER] [-w PASSWORD] [-k KEY]
                 [--sudo ADMINUSER] [-q]
                 <subcommand> ...
bin/omero: error: argument <subcommand>: invalid choice: 'upload'

choose from:
	admin, chgrp, chown, config, db, delete, download, errors, export,
	fs, group, help, hql, import, ldap, load, login, logout, node,
	obj, perf, quit, script, search, server, sessions, shell, tag,
	testengine, user, version, web
```

At the same time, I hadn't noticed that OMERO_DEV_PLUGINS was required. I'd think requiring OMERO_DEV_PLUGINS (for a non-future plugin) is roughly equivalent to removing without deprecation. The only other idea I have is OMERO_DEPRECATED_PLUGINS, but I haven't thought that one through really...

sbesson:
Updated. With the last commit, the following should be tested:

- `bin/omero upload -h` should return the help
- `bin/omero upload file` should print a warning indicating the deprecation
- `OMERO_NO_DEPRECATED_PLUGINS=1 bin/omero upload -h` should fail discovering the `upload` plugin

manics:
To remove the need for extra env-vars would something like the following work?
- Add a new internal plugin `_upload.py` to https://github.com/ome/omero-upload/tree/master/src/omero/plugins that imports everything from `upload.py`, i.e. it's an alias
- Rename `components/tools/OmeroPy/src/omero/plugins/upload.py` to ` components/tools/OmeroPy/src/omero/plugins/_deprecated_upload.py`
- `components/tools/OmeroPy/src/omero/plugins/upload.py` becomes something like
```python
try:
    from _upload import * # or omero.plugins._upload?
except:
    print Deprecation warning
    from _deprecated_upload import *
```

joshmoore:
Interesting. I could also imagine something like `omero/deprecated_plugins/` or `omero/plugins/deprecated`.

sbesson:
https://github.com/openmicroscopy/openmicroscopy/pull/5956#issuecomment-496465657 is an interesting way to generalize plugin deprecation as we decouple them. I will look into that.
Any reason not to use`from omero_upload.cli import *` within the try/catch block as in https://github.com/ome/omero-upload/blob/3dca359d7ef9305ce96f4305200c70d59e54ffea/src/omero/plugins/upload.py#L14?

joshmoore:
Thinking about this more, I'm unsure who would run `from _deprecated_upload import *`. It might be that a new directory is the beginning of plugin-loading "_phases_". Ones I can think of are:

 * 00: "required" plugins
 * 10: user-chosen (via `--path` flag)
 * 30: dev plugins (if OMERO_DEV_PLUGINS is set)
 * 50: load from python path (includes standard libraries)
 * 90: deprecated plugins

And first one wins.

joshmoore:
The only caveat regarding the current state of this PR is that we might want to get rid of the "OMERO_NO_DEPRECATED_PLUGINS" environment variable, right?

sbesson:
That's right. The module is marked as deprecated but we still need to explore how to selectively disable plugins installed via PyPI at runtime. I won't be able to come back to that in time for OMERO 5.5.0 so my vote goes towards merging and reviewing the mechanism in OMERO 5.5.x

jburel:
Merging Thanks

PR #5957, opened 2019-02-20
OMERO.web feedback can be disabled

manics:
# What this PR does

The IDR is receiving a lot of spam through the feedback form. This adds two properties to disable it:
- `omero.web.feedback.comment.enabled`: controls feedback through the top-right `Send Feedback` menu option, default `true`
- `omero.web.feedback.error.enabled`, controls feedback when an error occurs, default `true`

# Testing this PR

- Set `omero.web.feedback.comment.enabled` `false`. The `Send Feedback` menu item should take you to a "please contact" page
- Set `omero.web.feedback.error.enabled` `false`. Generate a 500 error, submitting the error should lead to a "please contact" page

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/feedback/templates/disabled.html
+++ b/components/tools/OmeroWeb/omeroweb/feedback/templates/disabled.html
@@ -23,7 +23,7 @@
     {% block content %}
       <div>
         <h1>Feedback is disabled on this server</h1>
-        <p>Please contact your system administrator or the <a href="https://www.openmicroscopy.org/support/">Openmicroscopy Environment</a></p>
+        <p>Please contact your system administrator or the <a href="https://www.openmicroscopy.org/support/">Open Microscopy Environment</a></p>
     {% endblock %}
     </div>
 
--- a/components/tools/OmeroWeb/omeroweb/feedback/templates/disabled.html
+++ b/components/tools/OmeroWeb/omeroweb/feedback/templates/disabled.html
@@ -1,0 +1,32 @@
+{% extends "webgateway/core_html.html" %}
+{% load i18n %}
+
+{% block link %}
+    {{ block.super }}
+    <link rel="stylesheet" href="{% static "feedback/css/layout.css" %}" type="text/css" />
+{% endblock %}
+
+{% block title %}
+    {% trans "OMERO.web - support" %}
+{% endblock %}
+
+{% block head %}
+    {{ block.super }}
+    {% include "webgateway/base/includes/shortcut_icon.html" %}
+{% endblock %}
+
+{% block body %}
+<div class="bodyWrapper">
+    <div><h1>
+        <a href="#" onClick="window.close()">{% trans "Close" %}</a>
+    </h1>
+    {% block content %}
+      <div>
+        <h1>Feedback is disabled on this server</h1>
+        <p>Please contact your system administrator or the <a href="https://www.openmicroscopy.org/support/">Openmicroscopy Environment</a></p>
+    {% endblock %}
+    </div>
+
+    </div>
+</div>
+{% endblock %}
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -643,17 +643,19 @@ CUSTOM_SETTINGS_MAPPINGS = {
           " Particularly useful when used in combination with the OMERO.web"
           " public user where logging in may not make sense.")],
     "omero.web.feedback.comment.enabled":
-        ["FEEDBACK_COMMENT_ENABLE",
+        ["FEEDBACK_COMMENT_ENABLED",
          "true",
          parse_boolean,
          ("Enable the feedback form for comments. "
-          "These comments are sent directly to the OME team.")],
+          "These comments are sent to the URL in ``omero.qa.feedback`` "
+          "(OME team by default).")],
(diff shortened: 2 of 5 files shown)


components/tools/OmeroWeb/omeroweb/feedback/templates/disabled.html
```diff
@@ -0,0 +1,32 @@
...
+
+{% block body %}
+<div class="bodyWrapper">
+    <div><h1>
+        <a href="#" onClick="window.close()">{% trans "Close" %}</a>
+    </h1>
+    {% block content %}
+      <div>
+        <h1>Feedback is disabled on this server</h1>
+        <p>Please contact your system administrator or the <a href="https://www.openmicroscopy.org/support/">Openmicroscopy Environment</a></p>
```
mtbc: Microscopy is a separate word.

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -642,6 +642,18 @@ def leave_none_unset_int(s):
          ("Whether or not to include a user dropdown in the base template."
           " Particularly useful when used in combination with the OMERO.web"
           " public user where logging in may not make sense.")],
+    "omero.web.feedback.comment.enable":
```
sbesson: Unfortunately, we do have a mixture of `<property>.enable` and `<property>.enabled` but the recently added properties seem to comply with the second scheme. Thought about using `.enabled` instead?

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -642,6 +642,18 @@ def leave_none_unset_int(s):
...
+        ["FEEDBACK_COMMENT_ENABLE",
+         "true",
+         parse_boolean,
+         ("Enable the feedback form for comments. "
+          "These comments are sent directly to the OME team.")],
+    "omero.web.feedback.error.enable":
```
sbesson: Same comment as above

components/tools/OmeroWeb/omeroweb/feedback/templates/disabled.html
```diff
@@ -0,0 +1,32 @@
...
+
+{% block head %}
+    {{ block.super }}
+    {% include "webgateway/base/includes/shortcut_icon.html" %}
+{% endblock %}
+
+{% block body %}
+<div class="bodyWrapper">
+    <div><h1>
+        <a href="#" onClick="window.close()">{% trans "Close" %}</a>
```
sbesson, line 21: Unclear whether we would like to have a "Go back to OMERO.web" test instead or in addition like in the base template.

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -642,6 +642,18 @@ def leave_none_unset_int(s):
          ("Whether or not to include a user dropdown in the base template."
           " Particularly useful when used in combination with the OMERO.web"
           " public user where logging in may not make sense.")],
+    "omero.web.feedback.comment.enabled":
+        ["FEEDBACK_COMMENT_ENABLE",
+         "true",
+         parse_boolean,
+         ("Enable the feedback form for comments. "
+          "These comments are sent directly to the OME team.")],
```
sbesson: Not necessarily the case. Maybe something along the lines of "to the URL specified by `omero.qa.feedback` (OME team by default)"?

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -642,6 +642,18 @@ def leave_none_unset_int(s):
...
+    "omero.web.feedback.error.enabled":
+        ["FEEDBACK_ERROR_ENABLE",
+         "true",
+         parse_boolean,
+         ("Enable the feedback form for errors. "
+          "These errors are sent directly to the OME team.")],
```
sbesson: Same comment as above.

sbesson (commented):
Tested first with a Web deployment using the default `omero.web.feedback` configuration i.e.e comments and errors enabled. Testing was performed using the `/feedback/comment` URL as well as the `/weberror/error_500` URL with the `omero-weberror` app installed and configured.

When disabling both comments and errors via configuration and restarted the Web deployment, the two workflows eventually redirect me to the disabled page:

screen shot 2019-02-21 at 13 44 40

Functionally, the PR works as expected as no feedback was sent in either case.

The only slightly unexpected behavior is the error disabling. Even with this configuration turned on, I was directed towards an error form when calling the `/weberror/error_505` URL. After filling it out, the redirection to the disabled page happens only once the Send button is hit. Main interrogation is whether we could block earlier in the process e.g. by greying out the `Send` button.

A few other minor comments. Otherwise, I think having such an option is useful.

components/tools/OmeroWeb/omeroweb/feedback/templates/disabled.html
manics: It's a popup window, so no need for a "go back" link. This is the same as when feedback is enabled.

sbesson: Understood. This is probably an artifact of my testing where I entered the URLs directly (notably to generate the 500 error page).

sbesson (approved):
screen shot 2019-02-25 at 08 59 49

Good to merge from my perspective unless some OMERO.web person might want to clear out the changes.

manics:
--rebased-to #5982

PR #5961, opened 2019-02-27
Drop all targets and usage of OME Artifactory for publishing

sbesson:
Following the extraction of the Java components out of this repository, the deployment logic is also unnecessary. All read usages of the Artifactory for retrieving JARs should be preserved.

The OMERO release jobs which were previously pushing artifacts will need to be modified accordingly.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -396,11 +396,6 @@ To get started using Eclipse, execute "ant build-dev" and import the top-level
         </copy>
     </target>
 
-    <target name="release-hudson" depends="check-ivy"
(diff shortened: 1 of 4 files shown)


joshmoore:
Green. The changes make sense to me. I'd propose getting this in for the m3 tag as well.

jburel (approved):

joshmoore:
Merging and tagging. I'll look into what changes I can manage in terms of the release jobs for m3.

PR #5964, opened 2019-02-28
OmeroPy: don't look for bin/omero in .git

joshmoore:
The `travers` method walks up from the current directory looking
for a subdirectory that contains `bin/omero`. Unfortunately, that
holds true for .git as well. A better fix would likely be to not
check more deeply than 1 or 2 subdirectories.

see: https://github.com/manics/openmicroscopy/commit/040878c07f05a77855fd483315279908fba4ad61#commitcomment-32511555

### DIFF
--- a/components/tools/OmeroPy/src/omero/testlib/__init__.py
+++ b/components/tools/OmeroPy/src/omero/testlib/__init__.py
@@ -172,7 +172,7 @@ class ITest(object):
                     # find OMERO dist by searching for bin/omero
                     # exclude OmeroPy as is a source
                     if (str(t.dirname().dirname().basename())
-                            not in ('OmeroPy', 'build', 'target') and
+                            not in ('OmeroPy', 'build', 'target', '.git') and
                             str(t.basename()) == 'omero' and
                             str(t.dirname().basename()) == 'bin'):
                         dist = t.dirname().dirname()


will-moore:
I found this didn't work when I created my own branch ```json•additional•settings``` because when the path ```t``` is e.g. ```/Users/wmoore/Desktop/OMERO/openmicroscopy/.git/logs/refs/heads/json•additional•settings``` 
then ```t.dirname().dirname().basename()``` is ```refs``` so it fails the check for ```.git```.
Same will be true for
```/Users/wmoore/Desktop/OMERO/openmicroscopy/.git/logs/refs/remotes/simon/json•additional•settings``` 

It needs something like:
```
             dist = None
             for root, dirs, files in os.walk(p):
+                if '.git' in root:
+                    continue
                 for f in files:
                     t = path(os.path.join(root, f))
```

**Comment by will-moore** [comment]
NB: This took a long time to debug because trying to print anything e.g. ```print root``` seemed to make the exception vanish (probably changing some str() behaviour) so I instead had to add a ```try/except``` to print out various values when it failed.

**Comment by joshmoore** [comment]
> It needs something like:

Even if the top-level ends with `.git`? I'll need to double-check.

**Comment by will-moore** [comment]
This bug has got me again, having checked out @manics ```omero•web•client•settings``` branch to test https://github.com/openmicroscopy/openmicroscopy/pull/6053

```
rm -rf /Users/wmoore/Desktop/OMERO/openmicroscopy/.git/logs/refs/remotes/simon/omero•web•client•settings
git branch -d json•additional•settings
```
Still didn't work, so I used my fix above https://github.com/openmicroscopy/openmicroscopy/pull/5964#issuecomment-494009766 so I can run tests...

joshmoore:
@will-moore : if you have something you can test, can you try two further things:

 * rename your `openmicroscopy` to `openmicroscopy.git` and see what happens
 * try changing your fix to test for `/.git/` (cross-platform) as opposed to just `".git" in`

joshmoore:
Taken care of by https://github.com/ome/openmicroscopy/pull/6200

PR #5968, opened 2019-03-01
Enable omero-build to use this repository

joshmoore:
In an effort to allow the decoupled submodule repositories
to also be able to use server integration tests in travis,
the relationship between omero-build and this repository
will need to be changed. For example, a PR in omero-server
will need a matching PR in omero-build which will build
all the component jars and then inject them into the Dockerfile
in this repository. This is done by:

 (1) appending the necessary versions to etc/omero.properties:

```
    echo versions.omero-blitz >> etc/omero.properties
```

 (2) using the omero-build image as the base image

```
    --build-arg=BUILD_IMAGE=...name of omero-build image...
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,23 +64,7 @@ RUN sed -i "s/^\(omero\.host\s*=\s*\).*\$/\1omero/" /src/etc/ice.config
 #
 #     https://trello.com/c/rPstbt4z/216-open-ssl-110
 #
-# RUN sed -i 's/\("IceSSL.Ciphers".*ADH[^"]*\)/\1:@SECLEVEL=0/' /src/components/tools/OmeroPy/src/omero/clients.py /src/etc/templates/grid/templates.xml
-
-
-# Temp: Build jars locally
-########RUN git clone git://github.com/ome/omero-gradle-plugins /tmp/omero-gradle-plugins
-########RUN cd /tmp/omero-gradle-plugins && git submodule update --init
-########RUN cd /tmp/omero-gradle-plugins && ./build.sh
-
-########RUN git clone git://github.com/ome/omero-build /tmp/omero-build
-########WORKDIR /tmp/omero-build
-########RUN git submodule update --init
-########RUN ./build.sh
-########WORKDIR /src
-########USER root
-########RUN apt-get update -y && apt-get install -y vim
-########USER omero
-# End Temp
+RUN sed -i 's/\("IceSSL.Ciphers".*ADH[^"]*\)/\1:@SECLEVEL=0/' /src/components/tools/OmeroPy/src/omero/clients.py /src/etc/templates/grid/templates.xml
 
 # Reproduce jenkins build
 RUN env BUILD_NUMBER=1 OMERO_BRANCH=develop bash docs/hudson/OMERO.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ ARG RUN_IMAGE=openmicroscopy/omero-${COMPONENT}:latest
 
 
 FROM ${BUILD_IMAGE} as build
+USER root
 RUN apt-get update \
  && apt-get install -y ant \
       python-pip python-tables python-virtualenv python-yaml python-jinja2 \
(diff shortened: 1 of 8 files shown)


joshmoore:
@sbesson : I could use your suggestion here. When build with M3 jars, `ome-resolver` fails as on travis in https://travis-ci.org/openmicroscopy/openmicroscopy/builds/500214378 since only user-maven is attempted:

```
	==== user-maven: tried
	  /home/build/.m2/repository/org/openmicroscopy/omero-blitz/5.5.0-m3/omero-blitz-5.5.0-m3-python.zip
``` 

which is why I previously hard-coded to `ome-artifactory`. But for the "use .m2 for SNAPSHOTS", I want *only* `user-maven`. I can turn this into "Yet Another Ivy Property" but that seems overkill. Do you know of an alternative?

jburel:
Without the challenge to ``ivysettings``, it is currently not possible to build locally
 with snapshot.
So we may want to extract that depending on the timeline

joshmoore:
@jburel : sorry, I don't understand. What are you trying to build? If you're saying that this PR isn't working, I know. That's what I'm trying to figure out.

jburel:
Sorry I am testing the "m4" PRs from @rgozim locally so I am using the artifacts generated locally

joshmoore:
Couple of additional thoughts:

 * the ivy/pom lines might need `(-[classifier])` but it's unclear if that will work
 * this may also help with "every other build fails" issues
 * but it's failing since ivy seems to only find `user-maven` in the `5.5.0-m3` case

joshmoore:
cc: @jburel and @sbesson (see commit message)

jburel:
With the ultimate goal in mind (getting rid of ivy), the (a bit ugly) solution  that allows everything to work  together seems acceptable

sbesson:
Tested locally a build with the last commit from a fresh `~/.m2` repository and I can confirm the compilation passes.

Having parallelly tried to look into Ivy artifact resolution logic, I suspect most of our problems arises from the nature of the `omero-blitz` artifact which escapes the classical `POM + JAR (+ Javadoc + sources)` composition with the existence of an additional `python` artifact of type `zip`.

https://github.com/openmicroscopy/openmicroscopy/pull/5968/commits/a44ffe6108d2be97ed8409efd8ad098895c0e8ed#diff-f17d700ee90f1db676bcc7f85037fb3eR36 looks like a way to tell Ivy that `omero-blitz` is more than a JAR the first time the artifact is retrieved even if the `python` is unnecessary for this component. We might want to note this.

joshmoore:
Note: current travis failure is going to require a merge on omero-test-infra.

Dockerfile
```diff
@@ -22,17 +22,18 @@ ARG RUN_IMAGE=openmicroscopy/omero-${COMPONENT}:latest
...
       zlib1g-dev python-pillow python-numpy python-sphinx \
       libssl-dev libbz2-dev libmcpp-dev libdb++-dev libdb-dev \
       zeroc-ice-all-dev \
- && pip install --upgrade 'pip<10' setuptools
+ && pip install --upgrade 'pip<10' setuptools flake8==2.4.0 pytest==2.7.3
 # TODO: unpin pip when possible
 # openjdk:8 is "stretch" or Debian 9
 RUN pip install https://github.com/ome/zeroc-ice-py-debian9/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
-RUN adduser omero
+RUN id 1000 || useradd -u 1000 -ms /bin/bash build
```
sbesson, line 36: Nice way to conditionally add the user and enable the base image to be overriden at build time.

components/tools/travis-build
```diff
@@ -123,18 +123,7 @@ build_cpp()
...
-    else
-        FILES=$(git diff --name-only HEAD...$TRAVIS_BRANCH)
-    fi
-    for x in $(echo "$FILES" | grep OmeroPy/test/integration); do
-        TEST=${x#components/tools/OmeroPy/}
-        export PYTHONPATH=$PYTHONPATH:../target/lib/python
-        ( cd components/tools/OmeroPy/;
-          python setup.py test -t $TEST -m "not broken"
-        )
-    done
```
sbesson, line 137: Is this covered in the `test-integration` target or something we want to come back to later ?

sbesson (commented):
Excluding the diff of the changes already merged in #5971, the changes in `Dockerfile` and `components/travis` look fine as a step to couple this to `omero-build` and other upstream repositories.

Travis is green so I am fine with getting it merged as starting to consume it actively.

components/tools/travis-build
joshmoore: `test-integration` is a superset of this garbage. Note also that `integration` was only ever used from omero-test-infra --> `travis-build`. The trade-off is that just building touched tests built _fast_ but not every PR touched the tests.

joshmoore:
Ok. After an extended call, merging. What this *doesn't* yet deal with:

 * how to include an ant build into the gradle-based omero-build repo
 * how to use version numbers rather than `SNAPSHOT` for pushing to nexus
 * how to verify PRs against this repository independently of `omero-build`

cc: @rgozim @manics @mtbc @jburel @sbesson

PR #5971, opened 2019-03-04
Coupled build

sbesson:
See also #5968

This PR cherry-picks the relevant build system changes allowing to couple omero-build and openmicroscopy builds without the Travis integration.

With this change it should be possible to do the following:

(0) optionally clean your local `~/.m2` cache

(1) build a version of `omero-blitz`locally e.g. 5.5.0-SNAPSHOT. Make sure the artifact is under `~/.m2/repository/org/openmicroscopy`

(2) append the version to `etc/omero.properties` or add to build `properties`

    echo versions.omero-blitz=5.5.0-SNAPSHOT >> etc/build.properties

(3) build the components from this repository and check the artifacts consume the correct versions

    ./build.py build-dev


mtbc (commented):
This has been working well for me locally today after this morning's puzzling surprise, thank you.

PR #5977, opened 2019-03-12
WIP: Build openmicroscopy using new build jars from repository

manics:
I've no idea why this doesn't work with just one pom.xml.

Example: `docker build -t openmicroscopy --build-arg MAVEN_SNAPSHOTS_REPO_URL=http://internal.host/nexus/repository/maven-internal/ .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,10 +53,22 @@ COPY test.xml /src/
 COPY LICENSE.txt /src/
 COPY history.rst /src/
 RUN chown -R 1000 /src
+
+# TODO: Required to download snapshot omero-build artifacts
+COPY download-repo-jars /src/download-repo-jars
+RUN chown -R 1000 /src/download-repo-jars && \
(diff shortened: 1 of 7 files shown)


joshmoore:
Not sure if this has been tried, but something similar to:

```
git grep local-maven2
etc/ivysettings.xml:  <property name="local-maven2-dir" value="${user.home}/.m2/repository/"
```

might work here based on the "feature" of ant that once a property has been set it won't be reset, i.e.

 * add a property pointing to some (usually) non-extant maven server (`local-maven-url-=http://localhost:8081`)
 * add a resolver that is typically not used like `local-maven-server-resolver` which chains on top of ome-resolver
 * override ome.resolver to point to local-maven-server-resolver (in `*.properties` or on the command-line)
 * override local-maven-url to point to nexus (in `*.properties` or on the command-line)

manics:
I think this is a lost cause

PR #5978, opened 2019-03-13
Bump to m4

joshmoore:
Move from released m3 to m4 jars.


joshmoore:
Only failures on west-ci are either search related (known due to indexObject issues) or related to the previous m4 build failing due to JDK8 v JDK9 javac issues. Merging and updating omero-build to get a last green state before moving forward.

See: https://github.com/ome/omero-build/pull/7

PR #5982, opened 2019-03-15
OMERO.web feedback can be disabled (rebased onto metadata54)

manics:
This is the same as gh-5957 but rebased onto metadata54.

----

# What this PR does

The IDR is receiving a lot of spam through the feedback form. This adds two properties to disable it:
- `omero.web.feedback.comment.enabled`: controls feedback through the top-right `Send Feedback` menu option, default `true`
- `omero.web.feedback.error.enabled`, controls feedback when an error occurs, default `true`

# Testing this PR

- Set `omero.web.feedback.comment.enabled` `false`. The `Send Feedback` menu item should take you to a "please contact" page
- Set `omero.web.feedback.error.enabled` `false`. Generate a 500 error, submitting the error should lead to a "please contact" page

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/feedback/templates/500-nosubmit.html
+++ b/components/tools/OmeroWeb/omeroweb/feedback/templates/500-nosubmit.html
@@ -1,0 +1,16 @@
+{% extends "base_error.html" %}
+{% load i18n %}
+
+{% block content %}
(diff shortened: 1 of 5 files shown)


manics:
--rebased-from #5957

sbesson:
Thanks @manics. Merging this as Travis is green and I will create an IDR-OMERO build with the Java SSL fix as well as this Web feature.

PR #5959, opened 2019-02-22
Reduce pause in Java integration search tests.

mtbc:
# What this PR does

The indexer should not need very long to index new data. This PR reduces the search facility test's wait for the setup data to become available.

# Testing this PR

https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-java/lastCompletedBuild/testngreports/integration.gateway/SearchFacilityTest/testSearch/ should remain green.

# Related reading

#4483

PR #5958, opened 2019-02-22
Have Python integration tests for search briefly pause for indexing.

mtbc:
# What this PR does

The Python integration tests add new data to OMERO then search for it. They are flaky. This PR adds brief pauses to tests to give OMERO time to index the new data.

# Testing this PR

See if this PR improves https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration-Python27/lastCompletedBuild/testReport/OmeroPy.test.integration.test_search/TestSearch/history/.

# Related reading

https://trello.com/c/wPBnTfPU/50-flaky-tests

PR #5986, opened 2019-03-20
Update tests to properly use services for search.

mtbc:
# What this PR does

Updates search tests to both tidy them up a bit and get them passing reliably with OMERO 5.5.

# Testing this PR

- https://web-proxy.openmicroscopy.org/west-ci/job/OMERO-test-integration/lastCompletedBuild/testngreports/integration.gateway/SearchFacilityTest/
- https://web-proxy.openmicroscopy.org/west-ci/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.test_search/TestSearch/history/
- https://web-proxy.openmicroscopy.org/west-ci/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_search/TestSearch/history/

# Related reading

Supersedes #5958 and #5959 and relies on fix from https://github.com/ome/omero-server/pull/25 and https://github.com/ome/omero-server/pull/22 to address https://trello.com/c/wPBnTfPU/50-flaky-tests.

### DIFF
--- a/components/tools/OmeroJava/test/integration/gateway/SearchFacilityTest.java
+++ b/components/tools/OmeroJava/test/integration/gateway/SearchFacilityTest.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2015 University of Dundee. All rights reserved.
+ *  Copyright (C) 2015-2019 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -18,6 +18,7 @@
  *
  *------------------------------------------------------------------------------
  */
+
 package integration.gateway;
 
 import java.util.ArrayList;
@@ -25,6 +26,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import omero.ServerError;
+import omero.api.IUpdatePrx;
 import omero.gateway.SecurityContext;
 import omero.gateway.exception.DSAccessException;
 import omero.gateway.exception.DSOutOfServiceException;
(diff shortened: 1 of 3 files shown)


jburel:
Does it mean that the indexing is not reliable on Annotation?
Or it should not be included from the start

mtbc:
`indexObjects` was never intended to be used for types that are not in `omero.search.include_types` and by default annotation is not. With that setting my indexer reimplementation refuses to index annotations directly but it will index them if related to some object (like an annotated project) whose type *is* included. You should expect indexing to be reliable on annotations but for default `include_types` the way to get them indexed is to have `indexObjects` index a different type that has the target annotations linked. When annotations are adjusted (e.g., linked and unlinked) there are database triggers to cause the reindexing of the annotated object.

jburel:
So it means that figure for example will not be indexed (assuming we want to index them)

mtbc:
If they aren't linked to anything then indeed they won't be (and I'd expect aren't in 5.4).

mtbc:
A look at `FullTextBridge` suggests to me that simply adding `FileAnnotation` to `include_types` would get them indexed, assuming that figures are attachments(?), except of course our graphical clients probably don't expect such a return type from search.

jburel:
I am not suggesting to index figure. It was just an example of annotation not linked to any object

joshmoore:
All the linked tests passed in today's build 349. :+1:

Leaving the workflow minded (cc: @pwalczysko @sbesson @will-moore) to discuss the indexing of other types. From my side, I don't see a reason not to _except_ that the UI doesn't permit searching for them.

PR #5989, opened 2019-04-01
CLI import tests check the read-only status afterward.

mtbc:
Tests https://github.com/ome/omero-blitz/pull/35 by extending https://web-proxy.openmicroscopy.org/west-ci/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_import/TestImport/. Both written as parameterized in case we find a way to do integration testing for `--transfer=ln` or if it's thought best to test more methods.


mtbc:
Can return to https://github.com/ome/omero-blitz/pull/35 at a quieter time.

PR #5990, opened 2019-04-03
Web cyberessentials

manics:
# What this PR does

Adds Django properties so OMERO.web can be configured to be compliant with Cyberessentials without patching the code. 

These defaults should work in all cases (i.e. no change to behaviour of OMERO.web and standard web-apps).
For production-level security you should set
```
omero.web.session_cookie_secure=true
omero.web.csrf_cookie_secure=true
```
to instruct the browser to only send the CSRF and session cookies over https.

See:
- https://github.com/openmicroscopy/prod-playbooks/blob/cc9146ca91a904ee2699cfbfc14e0e01e4416c67/omero/omero-web-patch-settings.yml
- https://github.com/openmicroscopy/prod-playbooks/pull/154/files
- https://github.com/openmicroscopy/prod-playbooks/pull/153/files
- https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-SESSION_COOKIE_SECURE
- https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-CSRF_COOKIE_SECURE
- https://docs.djangoproject.com/en/1.8/ref/settings/#x-frame-options

# Testing:
- `*_cookie_secure`: Either manually compare this to the changes made on nightshade and demo, or test locally by setting up Nginx with a fake SSL certificate, setting `*_COOKIE_SECURE`, and checking the headers.
- `xframeoptions`: check embedding OMERO.web in an iframe is prohibited by the browser.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -471,6 +471,18 @@ CUSTOM_SETTINGS_MAPPINGS = {
          None,
          leave_none_unset,
          "The name to use for session cookies"],
+    "omero.web.session_cookie_secure":
+        ["SESSION_COOKIE_SECURE",
+         "false",
+         parse_boolean,
+         ("Restrict session cookies to HTTPS only, you are strongly "
+          "recommended to set this to ``true`` in production.")],
+    "omero.web.csrf_cookie_secure":
+        ["CSRF_COOKIE_SECURE",
+         "false",
+         parse_boolean,
+         ("Restrict CSRF cookies to HTTPS only, you are strongly "
+          "recommended to set this to ``true`` in production.")],
     "omero.web.logdir":
         ["LOGDIR", LOGDIR, str, "A path to the custom log directory."],
     "omero.web.secure_proxy_ssl_header":
@@ -1247,6 +1259,10 @@ PIPELINE_JS = {
         'output_filename': 'omeroweb.viewer.min.js',
     }
 }
+
+SESSION_COOKIE_HTTPONLY = True
+# TODO: CSRF_COOKIE_HTTPONLY breaks OMERO.web javascript POSTs
+# CSRF_COOKIE_HTTPONLY = True
 
 CSRF_FAILURE_VIEW = "omeroweb.feedback.views.csrf_failure"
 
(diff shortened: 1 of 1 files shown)


components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -1248,6 +1269,10 @@ def report_settings(module):
     }
 }
 
+SESSION_COOKIE_HTTPONLY = True
```
joshmoore, line 1281: Can I suggest that these are included as configuration properties but with descriptions that include what you've learned about what breaks? This will break people from re-trying what you've already attempted.

joshmoore (changes requested):
:+1: Looks good when adding the properties to the OMERO-web job configuration on west-ci. Minor change request when it comes to the commented out properties.

components/tools/OmeroWeb/omeroweb/settings.py
manics: I can do that for `CSRF_COOKIE_HTTPONLY default False`. I don't think `SESSION_COOKIE_HTTPONLY` should be configurable though, there's no reason it should be `False`.

joshmoore:
Discussion summary:

 * **status**: production (in use on IDR, nightshade, etc.)
 * **docs**: A "Running in production" section needs to explain that these properties should be turned on once nginx is https only.
 * **testing**: configuration properties to be set on devspace by default

Additionally, if/when we want to prioritize it, `CSRF_COOKIE_HTTPONLY` can be turned to true, but at that time all of our Ajax posts will need to be reviewed to pass the CSRF token. cc: @will-moore @jburel

joshmoore:
Thanks, @manics. Do you have a playbook which matches your general "should be done in production" settings that we could point to?

manics:
No, there's https://github.com/openmicroscopy/prod-playbooks/blob/cc9146ca91a904ee2699cfbfc14e0e01e4416c67/omero/omero-web-patch-settings.yml which can be replaced with these new settings but only after OMERO.server is released.

joshmoore (approved):

joshmoore:
Merging to unblock gh-6010. Leaving you to find the best definition of "things to do in production" for adding these properties. :+1:

PR #5993, opened 2019-04-04
Swingtest

jburel:
# What this PR does

* remove the UI tests for insight.
* reduce dependencies list in omero.properties
A full review of the omero.properties file will happen in another PR

# Testing this PR

Check that ``./build.py clean build-dev`` still works


mtbc:
Looks good to me. There's enough here that somebody else may also want to look over it but I don't see any problems. :+1:

Might also want to clean up the stuff about Jython and whatnot in `components/tests/ui/README.md` but that can be done later.

jburel:
@mtbc readme updated

mtbc:
Looks good to me, thank you.

PR #5994, opened 2019-04-04
WIP: Jinja2 for Web server templates

manics:
# What this PR does

Replaces python string formatting with Jinja2 when generating web-server configurations. This also allows the three separate Nginx templates to be combined into one, replacing duplicated config stanzas by some Jinja2 conditionals.

In future this may allow the OMERO.web Ansible role to deterministically generate and update the Nginx config. Currently this requires some logic to guess when the config needs to be regenerated: https://github.com/openmicroscopy/ansible-role-omero-web/blob/2.0.1/tasks/web-nginx.yml#L12-L37
For example: currently if you make an OMERO config change after the nginx config has already been generated it won't be updated until you upgrade OMERO.web. (Note that Ansible does not support remote templates so it would require an extra step: https://www.reddit.com/r/ansible/comments/7r0eno/using_a_template_on_the_remote_host/dste442/)

TODO: Currently the templated variable names are unchanged, though if this is incorporated into the Ansible role they will need to be renamed.

This PR is on top of https://github.com/openmicroscopy/openmicroscopy/pull/5995 to make it easier to review.

# Testing this PR

It should be enough to verify the unit tests pass with only whitespace changes to the reference templates as shown by this filtered view: https://github.com/openmicroscopy/openmicroscopy/pull/5994/files?file-filters%5B%5D=.conf&w=1

# Related reading

- https://trello.com/c/ZTvqyLH7/602-make-generated-omero-web-%E2%87%84-nginx-config-part-of-the-api


joshmoore:
Closing in favor of https://github.com/ome/omero-web/pull/63

PR #5995, opened 2019-04-04
Remove apache configs from OMERO.web

manics:
# What this PR does

Apache support in OMERO.web has been deprecated since 5.3, and the configs have not been tested for several years. The configs have been removed, but the wsgi OMERO.web option remains so expert users can manually configure Apache if they wish.

# Testing this PR

It should be sufficient to review the unit tests since these cover the generation of the Nginx configs.

# Related reading

This was broken out of https://github.com/openmicroscopy/openmicroscopy/pull/5994 to make review easier


joshmoore:
Thanks, @manics.

PR #5927, opened 2018-12-12
glacier2: support ice websockets for client connections

manics:
Adds Ice 3.6+ websockets to the list of protocols that Glacier2 listens on. This should allow the OMERO binary protocol to be proxied via a standard web-server. Details: https://doc.zeroc.com/ice/3.7/client-side-features/proxies/proxy-and-endpoint-syntax#id-.ProxyandEndpointSyntaxv3.7-WSEndpointSyntax

Example of connecting directly to the OMERO websocket port from the Python client :
```
import omero.clients
c = omero.client(args=['--Ice.Default.Router=OMERO.Glacier2/router:ws -p 4465 -h localhost'])
c.createSession('root','omero')
```

More interesting example: proxy omero server via Nginx location `/omero`. Replace `<OMERO-SERVER-IP>` in `nginx-proxy-wss.conf`:
```
user  nginx;
worker_processes  1;

error_log  /var/log/nginx/error.log warn;
pid        /var/run/nginx.pid;

events {
    worker_connections  1024;
}

http {
    include       /etc/nginx/mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    access_log  /var/log/nginx/access.log  main;

    sendfile        on;
    #tcp_nopush     on;

    keepalive_timeout  65;

    #gzip  on;

    map $http_upgrade $connection_upgrade {
        default upgrade;
        '' close;
    }
    server {
        listen 80 default_server;
        server_name  $hostname;
        location /omero {
            proxy_pass http://<OMERO-SERVER-IP>:4065;
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection $connection_upgrade;
            proxy_read_timeout 86400;
        }
    }
}
```
Run nginx:
`docker run -it --rm --link omero:omero -v $PWD/ngi-proxy-wss.conf:/etc/nginx/nginx.conf -p 10080:80 nginx`

And configure your client:
```
import omero.clients
c = omero.client(args=['--Ice.Default.Router=OMERO.Glacier2/router:ws -p 10080 -h localhost -r /omero'])
c.createSession('root','omero')
```
[...]

PR #5998, opened 2019-04-05
Glacier2 IceSSL properties can be overridden or extended

manics:
templates.xml Glacier2Template IceSSL properties can be overridden or extended using omero config properties `omero.glacier2.IceSSL.*`
E.g.: Workaround the [Debian `templates.xml` openssl problem](https://docs.openmicroscopy.org/omero/5.4.10/sysadmins/unix/server-debian9-ice36.html#patching-omero-server):

    omero config set omero.glacier2.IceSSL.Ciphers "ADH:!LOW:!MD5:!EXP:!3DES:@STRENGTH:@SECLEVEL=0"

This also enables more interesting stuff like setting [additional IceSSL properties](https://doc.zeroc.com/ice/3.6/property-reference/icessl) e.g. for certificate verification.

To reduce the number of places the default properties are stored this includes a new `parse_omero_properties` method that reads defaults from `omero.properties` when substituting xml template variables. A future PR could remove more duplication, for instance the OMERO ports:
- https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.0-m5/components/tools/OmeroPy/src/omero/plugins/admin.py#L1092-L1096
- https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.0-m5/etc/omero.properties#L89-L100


components/tools/OmeroPy/src/omero/plugins/prefs.py
```diff
@@ -93,6 +94,45 @@ def with_rw_config(func):
     return wraps(func)(_make_open_and_close_config(func, False))
 
 
+class _OmeroPropertiesFileObj(object):
```
joshmoore: Can you unify this with the existing code?

 * https://github.com/openmicroscopy/openmicroscopy/blob/08e8a91b480bccb6e5bc181baa172fdd16f9924a/components/tools/OmeroPy/src/omero/plugins/prefs.py#L463
 * https://github.com/openmicroscopy/openmicroscopy/blob/08e8a91b480bccb6e5bc181baa172fdd16f9924a/components/tools/OmeroPy/src/omero/install/config_parser.py#L231

components/tools/OmeroPy/src/omero/plugins/admin.py
```diff
@@ -1040,6 +1043,13 @@ def jvmcfg(self, args, config):
                 sb += " # %s" % settings
             self.ctx.out("%s=%s" % (k, sb))
 
+    def _glacier2_icessl_xml(self, config_props):
+        # Convert omero.glacier2.IceSSL.* properties to IceSSL.*
+        glacier2_icessl = [(k[15:], v) for (k, v) in config_props.items()
+                           if k.startswith('omero.glacier2.IceSSL.')]
```
joshmoore: Nice strategy. If similar needs arise, I could see having this generalized for other prefixes.

manics:
Added:
- A test of the existing `omero.install.config_parser.PropertyParser`. I think there's a bug in it but it's not an obvious fix and it seems to work anyway so I've added it with an xfail so we can return in future https://github.com/openmicroscopy/openmicroscopy/pull/5998/commits/cb05b91ccbb15029ad88c6ac29d381c9ec643048
- Tests for the new `omero.glacier2.IceSSL` properties
- Support for OS x which has a separate template and different Ciphers. These are the remaining differences:
```diff
$ diff -U 5 etc/templates/grid/templates.xml etc/templates/grid/osxtemplates.xml
--- etc/templates/grid/templates.xml    2019-04-08 13:53:18.000000000 +0100
+++ etc/templates/grid/osxtemplates.xml 2019-04-08 16:04:24.000000000 +0100
@@ -28,11 +28,11 @@
     <properties id="Repository">
       <property name="omero.repo.wait" value="300"/><!-- seconds -->
     </properties>

     <properties id="Basics">
-      <property name="Ice.MessageSizeMax" value="250000"/> <!-- 250 MB -->
+      <property name="Ice.MessageSizeMax" value="65536"/> <!-- 64 MB -->
       <property name="Ice.CacheMessageBuffers" value="0"/>
       <property name="Ice.Override.ConnectTimeout" value="5000"/>
     </properties>

     <properties id="SingleThreaded">
@@ -267,10 +267,15 @@
         <option>MEMORY:pixeldata</option>
         <option>-Djava.awt.headless=true</option>
         <option>-Dlogback.configurationFile=${OMERO_ETC}logback-indexing.xml</option>
         <option>-Domero.logfile=${OMERO_LOGFILE}</option>
         <option>-Domero.name=PixelData-${index}</option>
+        <!-- TODO: review when omero-blitz has a manifest
+        <option>-jar</option>
+        <option>${OMERO_JARS}omero-blitz.jar</option>
+        <option>ome.pixeldata</option>
+        -->
         <option>ome.services.blitz.Entry</option>
         <option>ome.pixeldata</option>
         <env>CLASSPATH=${OMERO_JARS}*</env>
         <adapter name="PixelDataAdapter" endpoints="tcp"/>
         <properties>
@@ -294,18 +299,13 @@
         <option>-Djava.awt.headless=true</option>
         <option>-Dlogback.configurationFile=${OMERO_ETC}logback.xml</option>
         <option>-Domero.logfile=${OMERO_LOGFILE}</option>
         <option>-Domero.name=Repository-${index}</option>
         <option>-Domero.repo.dir=${dir}</option>
-        <!-- TODO: review when omero-blitz has a manifest
         <option>-jar</option>
         <option>${OMERO_JARS}omero-blitz.jar</option>
-        <option>ome.pixeldata</option>
-        -->
-        <option>ome.services.blitz.Entry</option>
         <option>OMERO.repository</option>
-        <env>CLASSPATH=${OMERO_JARS}*</env>
         <adapter name="RepositoryAdapter" endpoints="tcp">
           <object identity="Repository-${index}" type="::omero::grid::InternalRepository"/>
         </adapter>
         <properties>
           <properties refid="Basics"/>
@@ -461,11 +461,11 @@
          <adapter name="IcePatch2" endpoints="${endpoints}">
            <object identity="${instance-name}/server" type="::IcePatch2::FileServer"/>
          </adapter>
          <properties>
             <properties refid="Profile"/>
-            <property name="IcePatch2.Admin.Endpoints" value="tcp -h @omero.master.host@"/>
+            <property name="IcePatch2.Admin.Endpoints" value="tcp -h @Ice.Default.Host@"/>
             <property name="IcePatch2.Admin.RegisterProcess" value="1"/>
             <property name="IcePatch2.InstanceName" value="${instance-name}"/>
             <property name="IcePatch2.Directory" value="${directory}"/>
          </properties>
        </server>
```
Are these intentional, or are they due to someone modifying templates.xml but forgetting about osxtemplates.xml? If it's the latter they could be combined.

joshmoore:
I think the latter. I think the only difference was for openssl issues which you seem to have taken care of nicely.

manics:
I've changed the property and logic so the darwin default is internal, setting `omero.glacier2.IceSSL.Ciphers` should work on all systems.

dominikl:
Works on Debian 9. Can't connect with Insight after basic server setup, like expected. But after running `./omero config set omero.glacier2.IceSSL.Ciphers 'ADH:!LOW:!MD5:!EXP:!3DES:@STRENGTH:@SECLEVEL=0'` the connection can be established.

etc/omero.properties
```diff
@@ -952,6 +952,33 @@ omero.client.browser.thumb_default_size=96
...
+# Glacier2Template SSL allowed protocols
+omero.glacier2.IceSSL.Protocols=tls1
+
+#############################################
+## Darwin (OS X) specific defaults for templated configuration files only.
+## ``darwin.xxx.yyy`` properties will override xxx.yyy properties in this file.
+## Do not set these properties in ``omero config``, instead set the original
+## ``xxx.yyy`` property
+#############################################
+darwin.omero.glacier2.IceSSL.Ciphers=(AES)
```
joshmoore: Can I propose using `darwin` _suffix_ to match jvmcfg? https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/unit/test_jvmcfg.py#L82

manics: Can do, as long as we're sure they'll never be a third party property that ends in `.darwin`.

I think it messes up the hierarchy though:
- OMERO is running on a Darwin system.
- Glacier2 is run by OMERO
- IceSSL is a component of Glacier2
- Ciphers is an attribute of IceSSL

joshmoore: > as long as we're sure they'll never be a third party property

Don't imagine we can ever be sure of that, though we can be sure we won't need these particular properties for ever. `omero.glacier2.darwin` would be an option, but it's likely hard to detect.

> I think it messes up the hierarchy though:

I understand, but it also seems odd to be doing basically the same type of override two different ways.

manics: How does this look?

joshmoore:
(Re-ran travis; now green. Download of wlz had failed)

etc/omero.properties
joshmoore: Thanks, and I like the idea of making the property "file-private" with the prefix. As long as that is made clear in the property name (perhaps also with the added benefit of getting across the fact, "this could change, please don't rely on it"), then :+1: from me. Other options that occurred to me while trying to evaluate the impact of this:

```
_darwin_omero
__darwin_omero
_omero...darwin
__omero...darwin
__omero_overrides...darwin
```

mtbc:
With this PR in but no extra configuration of `omero.glacier2.IceSSL.*` or whatever would you expect OMERO.web to still be able to connect? (For me it reports, "Server is not responding".)

manics:
With no extra `omero.glacier2.IceSSL.*` properties set the templates should be equivalent to before, I'd everything to work

mtbc:
Hmmm. With this PR I'm also getting,

```
$ omero config get | grep Ice
Ice.Default.Host=127.0.0.1
$ omero login
Previous session expired for user-2 on localhost:4064
Server: [localhost:4064]
Username: [user-2]    
Password:
Ice.ConnectionRefusedException: localhost:4064 isn't running
```

mtbc:
Ah, I do routinely run the command at https://docs.openmicroscopy.org/omero/5.4.10/sysadmins/troubleshooting.html#openssl-version - perhaps that could be a bad interaction?

manics:
> Ah, I do routinely run the command at https://docs.openmicroscopy.org/omero/5.4.10/sysadmins/troubleshooting.html#openssl-version 

It might be! This PR should replace the templates.xml sed command, can you try `omero config set omero.glacier2.IceSSL.Ciphers "ADH:!LOW:!MD5:!EXP:!3DES:@STRENGTH:@SECLEVEL=0"`?

Unfortunately this does not affect clients.py so you'll still need to run sed on that file.

mtbc:
Okay, with Debian 9.8 I ran,
```console
$ sed -ie 's/\("IceSSL.Ciphers".*"ADH\b\)/\1:@SECLEVEL=0/' dist/lib/python/omero/clients.py
$ dist/bin/omero config get | grep glacier2
omero.glacier2.IceSSL.Ciphers=ADH:!LOW:!MD5:!EXP:!3DES:@STRENGTH:@SECLEVEL=0
```
Now,
```console
$ omero login
Previous session expired for user-2 on localhost:4064
Server: [localhost:4064]
Username: [user-2]
Password:
Created session for user-2@localhost:4064. Idle timeout: 10 min. Current group: private-1
```
and web login works fine! :cake:

I guess we'll need some what's-new documentation then for those who were using the previous troubleshooting advice.

mtbc:
Also having some trouble figuring out with this PR what combination of changes will allow me to run the Python integration tests without hitting "no ciphers available".

manics:
@mtbc Do you want to give https://github.com/openmicroscopy/openmicroscopy/pull/6007 a try? (edit: this was in reply to the earlier comment, not the latest one)

mtbc:
Sure: with this PR merged in as well? And no special `glacier2` config or troubleshooting tweaks?

mtbc:
With #5998 and #6007 and no further `sed` or `config set omero.glacier2` or whatever, CLI login gives me,
```
Ice.ConnectionRefusedException: localhost isn't running
```

manics:
Sorry for not being clear. https://github.com/openmicroscopy/openmicroscopy/pull/6007 replaces the sed command for `clients.py`. You still need to set `omero.glacier2.IceSSL.Ciphers "ADH:!LOW:!MD5:!EXP:!3DES:@STRENGTH:@SECLEVEL=0"` to update the server.

It would be possible to automatically make this change in the server template too if openssl 1.1 is detected, but given this PR provides a non-hacky workaround I think it's better for server admins to be fully aware of what's configured (and that we're reducing the default openssl security level). Regardless, I think if that change were to be made it should go in a separate PR from this one.

mtbc:
:+1: Works excellently with the `config set` but without any seddery. Great!

jburel:
Excluding so we can review the "reduced" omero.properties #6004
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-DEV-merge-push#1314](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/1314/). See the [console output](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-push/1314/consoleText) for more details.
Possible conflicts:
  - PR #5927 manics 'glacier2: support ice websockets for client connections'
    - etc/omero.properties
  - Upstream changes
    - etc/omero.properties

~~--conflicts~~ Conflict resolved in build [OMERO-push#487](https://web-proxy.openmicroscopy.org/west-ci/job/OMERO-push/487/). See the [console output](https://web-proxy.openmicroscopy.org/west-ci/job/OMERO-push/487/consoleText) for more details.

manics:
Rebased without changes

joshmoore (approved):
Discussion summary:

status: production (tested by @mtbc)
docs: configuration properties but likely also a main docs page as well
testing: @joshmoore to try to make the Dockerfile in this repo fail and then use these properties to get it passing again

joshmoore:
I'll open a separate PR for testing the Dockerfile. (Currently looking at the same issue in other repositories) Leaving @manics to drive the documentation  on return.

PR #5999, opened 2019-04-09
Ensure date is padded with 0

manics:
The day component of annotation dates (`On: <date>`) was missing a leading 0 when the day has one digit, see screenshot:
tmp


pwalczysko:
yes, i can see the change here working as expected on eel. 
Compare with outreach (5.4.10 server without this PR), where the leading zero on days is indeed missing in Web.

will-moore:
👍 Looks good.

PR #6002, opened 2019-04-11
Batch map annotation

will-moore:
# What this PR does

See https://trello.com/c/VZqmJCyP/41-batch-add-map-annotations

This supports creating new Map Annotations for multiple selected objects at once.
We create a separate Map Annotation for each object (no 1-to-many links).
When multiple objects with *identical* Map Annotations are selected, the annotations are displayed
as a single annotation to allow editing of all of them in sync.

# Testing this PR

1. Select multiple objects in webclient
1. Under Key-Value pairs tab, you should see a placeholder for creating new Map Annotations.
1. Enter a bunch of Key-Value pairs etc. editing as normal.
1. Select each single object in turn to check that the Key-Value pairs are on each one.
1. Select multiple objects again (possibly even including some that were not selected before, and/or excluding some that were selected before)
1. Identical Map Annotations should appear as a single one, with text above to say the number of objects they are on.
1. The tooltip on these duplicate annotations lists the objects they are linked to.
1. Editing this duplicate again should apply the changes to all objects
1. Editing these Map Annotations individually, when 1 object is selected (not in batch panel), means that they are no-longer identical to the other copies and will not be combined with them when you return to Batch Annotate mode.
1. However, if they are once-again edited to become identical then they will be combined in Batch Annotate panel again.
1. We also allow editing of all other "client" Map Annotations in Batch Annotate panel. Previously, no Key-Value pairs were editable in Batch Annotate panel.

Screen Shot 2019-04-11 at 14 08 22

cc @evenhuis

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
@@ -49,6 +49,19 @@ var MapAnnsPane = function MapAnnsPane($element, opts) {
         return isClientMapAnn(ann) && ann.owner.id == WEBCLIENT.USER.id;
     };
 
+    var annsEqual = function(annA, annB) {
+        var valuesA = annA.values;
+        var valuesB = annB.values;
+        if (annA.id === annB.id) return true;
+        if (valuesA.length != valuesB.length) return false;
+        for (var i=0; i<valuesA.length; i++) {
+            if (valuesA[i][0] !== valuesB[i][0] || valuesA[i][1] !== valuesB[i][1]) {
+                return false;
+            }
+        }
+        return true;
+    }
+
     this.apiAnnotationUrl = function apiAnnotationUrl(url) {
         if (!url) { return WEBCLIENT.URLS.webindex + "api/annotations/";}
         return url;
@@ -110,11 +123,6 @@ var MapAnnsPane = function MapAnnsPane($element, opts) {
                     var map_annotations = [];
 
                     anns.forEach(function(ann){
-                        if (batchAnn) {
-                            // Don't allow editing in batch annotate panel
-                            // Get error if all rows are deleted then try to add new map
-                            ann.permissions.canEdit = false;
-                        }
                         if (isMyClientMapAnn(ann)) {
                             my_client_map_annotations.push(ann);
                         } else if (isClientMapAnn(ann)) {
@@ -123,6 +131,31 @@ var MapAnnsPane = function MapAnnsPane($element, opts) {
                             map_annotations.push(ann);
                         }
                     });
+
+                    if (batchAnn) {
+                        var unique_anns = [];
+                        // Try to collapse any IDENTICAL map anns...
+                        my_client_map_annotations.forEach(function(ann){
+                            var duplicate = false;
+                            for (var u=0; u<unique_anns.length; u++) {
+                                var unique_ann = unique_anns[u];
+                                if (annsEqual(unique_ann, ann)) {
+                                    // combine anns
+                                    unique_ann.id = unique_ann.id + "," + ann.id;
+                                    if (!unique_ann.parentNames) {
+                                        unique_ann.parentNames = [unique_ann.link.parent.name];
+                                    }
+                                    unique_ann.parentNames.push(ann.link.parent.name);
+                                    duplicate = true;
+                                    break;
+                                }
+                            }
+                            if (!duplicate) {
+                                unique_anns.push(ann);
+                            }
+                        });
+                        my_client_map_annotations = unique_anns;
+                    }
 
                     // Update html...
                     var html = "";
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js
@@ -128,7 +128,7 @@ var MapAnnsPane = function MapAnnsPane($element, opts) {
                     var html = "";
                     var showHead = true;
                     // If not batch_annotate, add placeholder to create map ann
-                    if (canAnnotate && !batchAnn) {
+                    if (canAnnotate) {
                         if (my_client_map_annotations.length === 0) {
                             showHead = false;
                             my_client_map_annotations = [{}];   // placeholder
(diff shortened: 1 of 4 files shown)


pwalczysko:
Works as described. I like this workflow, with the knowledge that all the lines of the Key-Value pairs might be deleted, which deletes the MapAnnotation, this gives the user full flexibility IMHO.

I guess we might still need some scripts which would transfer the info from the KVPs into CSV sheets and vice versa, but this would be a separate PR I think.

joshmoore:
> We create a separate Map Annotation for each object (no 1-to-many links).

> When multiple objects with identical Map Annotations are selected, the annotations are displayed
as a single annotation to allow editing of all of them in sync.

This asymmetry might lead to confusion. i.e. there's no way for a user to create two map annotations are the same or detect it other than comparing object IDs, creating a map on two objects creates one annotation, but editing two at the same time edits _both_. Does editing one make a copy?

will-moore:
@joshmoore Sorry, that's a long sentence that I can't quite understand.

> there's no way for a user to create two map annotations are the same

This is not the case. You can create two map annotations that are the same: Simply select 2 images and enter your key-value pairs into the placeholder table.

> creating a map on two objects creates one annotation

That's not how it works in the UI just now. If I select 2 objects and enter Key-Value pairs into the placeholder table, this creates TWO identical map annotations.

> Does editing one make a copy?

No. If I edit a single Map Annotation (where it doesn't say ```To: n objects```) then it just updates that Map Annotation (same as existing behaviour in the single-selection right panel).

manics:
I agree with @pwalczysko , this works quite nicely

joshmoore:
| created by           	| annotations that exist 	| editing via  	| result                     	|
|----------------------	|------------------------	|--------------	|----------------------------	|
| one object with `+`  	| 1                      	| single-click 	| 1 edit; ok                 	|
| two object with `+`  	| 2                      	| single-click 	| 1 edit; ok                 	|
| two objects with `+` 	| 2                      	| multi-select 	| both change?               	|
| via bulk             	| 1                      	| single-click 	| both appear to change; ok? 	|
| via bulk             	| 1                      	| multi-select 	| both appear to change; ok  	|

will-moore:
Row 3: If 2 *identical* annotations exist on the 2 selected objects. Both change ✅ 
Rows 4 and 5: If you have created a Map Annotation via some method other than via the client UI (and linked it to multiple objects) then it probably shouldn't use the "client" namespace and so it won't be editable in the webclient or Insight. If you did use the "client" namespace then Row 4 is correct (existing behaviour even without this PR) and Row 5 is now also correct.

joshmoore:
Thanks, @will-moore. I think I'm starting to understand. Does this workflow apply _only_ to map annotations with the "client" namespace or will it apply to any? If the latter, I still think we should at least test through what happens in those scenarios since we are telling people to create via bulk.

```
Added by: root root
To: 2 objects
```
Though I know one can figure it out via the IDs, this to me sounds like the same annotation is on multiple objects (which would be good to know, as well).

Would it be alright to make this more explicitly, namely by stating "currently editing two annotations" and if possible make some visual distinction that that's what's happening? Perhaps `(?)` with a "identical annotations were detected...".

will-moore:
Yes, only "client" map annotations can be created & edited in webclient and Insight.

I've added more info to headers and tooltips to give users a better idea what's going on:

Screen Shot 2019-04-12 at 13 41 03

Screen Shot 2019-04-12 at 13 39 43

Look better? Suggestions for tweaking the text?

evenhuis:
Hi Will and Josh,

I never got my head around namespaces and how they work.  When I add a map annotation from a python script would this in the client namespace?  

It's been a while since I looked into it, but I thought that that the set of the kv's attached to an object is a single immutable map annotation so when I want to add a single to kv to an object with existing kv's I need to create new mapannotation with the existing kvs the additional then replace it.

Have I got this the wrong way round?

Cheers,

Chris

will-moore:
Hi Chris.
 The "client" namespace ```omero.constants.metadata.NSCLIENTMAPANNOTATION``` or ```"openmicroscopy.org/omero/client/mapAnnotation"``` is simply a constant that is used to denote annotations that are editable in the clients. It *can* be added to any Map Annotation (e.g. created by a script) but it is not added automatically anywhere. 

Map Annotations are not immutable. You can update them. E.g. when editing in webclient: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/views.py#L2314

I'm hoping this PR replaces your ```Key_Val_from_Description.py``` functionality (although you have to select all images at once instead of edit on the Dataset). Also, this creates new (separate) Map Annotations instead of adding Key-Values to existing ones.

Cheers,
  Will.

evenhuis:
sorry, I might have got confused from [old documentation](https://docs.openmicroscopy.org/omero/5.4.1/developers/Model/StructuredAnnotations.html)

> Immutability
>The actual content value of an annotation – the text, long, double, file value, etc – is immutable. >Links to and from the annotation, however, can be modified.
>
>Currently the namespace field of annotations is mutable. See #878 for discussion.

will-moore:
Apologies Chris. That documentation is way out of date. The latest documentation (5.4.10) https://docs.openmicroscopy.org/omero/5.4.10/developers/Model/StructuredAnnotations.html is a bit better and @mtbc has just opened a PR to remove "immutable" statement from that. https://github.com/openmicroscopy/ome-documentation/pull/1943

will-moore:
To test that last commit:
 - Select multiple objects
 - Start editing Key-Value pairs, then delete all the Keys and Values
 - Then fill out Key-Value pairs again.
 - Check that the objects receive the latest Key-Values correctly.

pwalczysko:
Will have a look tomorrow.

manics:
There an inconsistency for non-client map annotations, and map-annotations added by another user- they're both displayed individually instead of being grouped:

Non-client namespace:
Screen Shot 2019-04-16 at 10 23 04

`user-6` viewing `user-3`'s annotations in group `read-annotate-1`:
Screen Shot 2019-04-16 at 10 27 21

pwalczysko:
>user-6 viewing user-3's annotations in group read-annotate-1:

@manics @will-moore : Please note that user-6 is a full administrator in OMERO. 

I think I can understand now the point @manics is making : Lets have the MAnns grouped wherever possible, whether these are editable or not. I think this is a sound idea.

pwalczysko:
This https://github.com/openmicroscopy/openmicroscopy/pull/6002#issuecomment-482590800 bug is fixed.

I think there is one ToDo here, the comment of @manics https://github.com/openmicroscopy/openmicroscopy/pull/6002#issuecomment-483584228

will-moore:
That commit should also group map annotations from other users and other 'non-client' namespaces. NB: I still don't group them if owners or namespaces don't match since then they are coming from a different source and should be displayed separately (but I think this is an edge case).
- Tooltip should say 'viewing' instead of 'editing' if you're not able to edit:

Screen Shot 2019-04-18 at 10 20 01

pwalczysko:
the last commits work as described in the instruction https://github.com/openmicroscopy/openmicroscopy/pull/6002#issuecomment-484422936

I still have a RFE though

Looking now at the tooltips in RHP when images are multiselected, I see that, in case there is only one single MAnn attached to one single image, I get comparably very rudimentary tooltip, not telling me for example to which image is this single MAnn attached.

Screen Shot 2019-04-19 at 11 43 13

Whereas if there are multiple images with the same annotation, I get a full info.
I would like the full info also when the tooltip is on single MAnn - I think this is a logical extension of the user-friendlinness.

will-moore:
@pwalczysko If it's only a single annotation, the object it's linked to is in bold above the key-value table, not in the tooltip.
Screen Shot 2019-04-19 at 18 15 24

joshmoore:
I tested through all of https://github.com/openmicroscopy/openmicroscopy/pull/6002#issuecomment-482535060 with:

```
  490  for x in $(seq 1 5); do touch row_${x}_a.fake row_${x}_b.fake row_${x}_c.fake; done
  492  bin/omero import -T Dataset:28401 /tmp/*fake
  503  bin/omero login user-3@localhost
  504  bin/omero obj new MapAnnotation
  505  bin/omero hql -q "select i from Image i order by i.id desc"
  506  bin/omero obj new ImageAnnotationLink parent=Image:164110 child=MapAnnotation:48406
  507  bin/omero obj new ImageAnnotationLink parent=Image:164111 child=MapAnnotation:48406
  510  bin/omero obj map-set MapAnnotation:48406 mapValue a b
  511  bin/omero login root@localhost
  512  bin/omero sessions group read-only-1
  513  M=$(bin/omero obj new MapAnnotation)
  514  bin/omero obj new ImageAnnotationLink parent=Image:164111 child=$M
  515  bin/omero obj new ImageAnnotationLink parent=Image:164112 child=$M
  516  bin/omero obj map-set $M mapValue a b
```

all in http://web-dev-merge.openmicroscopy.org/webclient/?show=dataset-28401 if anyone else wants to review. Everything works as expected.

jburel:
I checked the examples created by @joshmoore. Looks good
merging

PR #5653, opened 2018-02-09
Lucene search

will-moore:
# What this PR does

See https://trello.com/c/ILIGB0vf/117-rfe-search-cross-feature
NB: This isn't currently targeted for OMERO 5.4.4

TODO: Add robot tests

# Testing this PR

1. Search using Advanced tab to input raw lucene queries.
1. Switching between Advanced and regular search by switching tabs (without needing to remove search terms from each input)
1. Other form elements should work as before: Search for Project, Dataset Image etc and filter by Group & Owner.
1. Suggestions to improve the UI? Want to give the user some clue as to what advanced search is and how to use it without having to visit the docs pages. Example query shows:
    - usage of ```AND```
    - Searching by different fields and separation of field:query with colon.
    - Usage of double quotes

screen shot 2018-02-12 at 15 31 39

cc @joshmoore I believe you've been requesting this feature for some time?

PR #6008, opened 2019-04-17
Lucene search

will-moore:
# What this PR does

This was previously opened as https://github.com/openmicroscopy/openmicroscopy/pull/5653
but that couldn't be re-opened so it is replaced by this PR.

See https://trello.com/c/ILIGB0vf/117-rfe-search-cross-feature

# Testing this PR

1. Search using Advanced tab to input raw lucene queries.
1. Switching between Advanced and regular search by switching tabs (without needing to remove search terms from each input)
1. Other form elements should work as before: Search for Project, Dataset Image etc and filter by Group & Owner.
1. Suggestions to improve the UI? Want to give the user some clue as to what advanced search is and how to use it without having to visit the docs pages. Example query shows:
    - usage of ```AND```
    - Searching by different fields and separation of field:query with colon.
    - Usage of double quotes

screen shot 2018-02-12 at 15 31 39

cc @joshmoore I believe you've been requesting this feature for some time?

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4515,7 +4515,7 @@ class _BlitzGateway (object):
         :param created:     :class:`omero.rtime` list or tuple (start, stop)
         :param useAcquisitionDate: if True, then use Image.acquisitionDate
                                    rather than import date for queries.
-        :param rawQuery     If True, the text is passed directly to byFullText()
+        :param rawQuery     If True, text is passed directly to byFullText()
                             without processing. fields is ignored.
         :return:            List of Object wrappers. E.g. :class:`ImageWrapper`
         """
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4505,7 +4505,7 @@ class _BlitzGateway (object):
 
     def searchObjects(self, obj_types, text, created=None, fields=(),
                       batchSize=1000, page=0, searchGroup=None, ownedBy=None,
-                      useAcquisitionDate=False):
+                      useAcquisitionDate=False, rawQuery=True):
         """
         Search objects of type "Project", "Dataset", "Image", "Screen", "Plate"
         Returns a list of results
@@ -4575,10 +4575,14 @@ class _BlitzGateway (object):
             for t in types:
                 def actualSearch():
                     search.onlyType(t().OMERO_CLASS, ctx)
-                    search.byLuceneQueryBuilder(
-                        ",".join(fields),
-                        d_from, d_to, d_type,
-                        text, ctx)
+                    if rawQuery:
+                        # TODO: need to handle dates: d_from, d_to, d_type
+                        search.byFullText(text, ctx)
+                    else:
+                        search.byLuceneQueryBuilder(
+                            ",".join(fields),
+                            d_from, d_to, d_type,
+                            text, ctx)
 
                 timeit(actualSearch)()
                 # get results
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4515,6 +4515,8 @@ class _BlitzGateway (object):
         :param created:     :class:`omero.rtime` list or tuple (start, stop)
         :param useAcquisitionDate: if True, then use Image.acquisitionDate
                                    rather than import date for queries.
+        :param rawQuery     If True, the text is passed directly to byFullText()
+                            without processing. fields is ignored.
         :return:            List of Object wrappers. E.g. :class:`ImageWrapper`
         """
         if not text:
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -4578,7 +4578,8 @@ class _BlitzGateway (object):
                 def actualSearch():
                     search.onlyType(t().OMERO_CLASS, ctx)
                     if rawQuery:
-                        # TODO: need to handle dates: d_from, d_to, d_type
+                        if created is not None and len(created) > 1:
+                            search.onlyCreatedBetween(created[0], created[1])
                         search.byFullText(text, ctx)
                     else:
                         search.byLuceneQueryBuilder(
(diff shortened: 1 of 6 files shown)


manics:
There's a travis flake8 error `./components/tools/OmeroPy/src/omero/gateway/__init__.py:4518:80: E501 line too long (80 > 79 characters)`

manics:
I think there's a bug in the client side validation. If the default search box (on the non-advanced search tab) is empty when I submit an advanced search I get a `Search must contain some text.` error. I can workaround this by putting something in the original search box e.g. `.`, at which point the advanced lucene search seems to work e.g. `details.owner.omeName: "user-12"` will return objects owned by `user-12`

manics:
> Suggestions to improve the UI? Want to give the user some clue as to what advanced search is and how to use it without having to visit the docs pages

My initial thoughts are that a simplified explanation of a Lucene query is probably still too complicated for a typical user, and it'd be better to make it clear that it requires an advanced query syntax and link out to a docs page, perhaps a user docs "advanced search" page instead of the developer docs.

manics:
For comparison here's the Kibana search UI (the `Not ready yet? Find our lucene docs here.` link is https://www.elastic.co/guide/en/elasticsearch/reference/6.7/query-dsl-query-string-query.html#query-string-syntax):
Screen Shot 2019-04-18 at 11 32 14-fullpage

mtbc:
> link out to a docs page

If it's of any help, the relevant Lucene page is probably https://lucene.apache.org/core/3_1_0/queryparsersyntax.html.

will-moore:
@manics I fixed the ```Search must contain some text``` error. Now you should only see that if the current search field is empty.
I think what we're doing is very similar to the Kibana screenshot. They have ```Search... e.g. status:200 AND extension:PHP``` whereas we have ```Enter raw search term, for example: name:Control AND tag:"GFP H2B"```, with links to more extensive docs. But I agree that we need a more advanced user docs instead of dev docs.

pwalczysko:
https://docs.openmicroscopy.org/omero/5.4.10/developers/Modules/Search.html#leading-wildcard-searches
seems not to be true to me. It is allowed now in the "Advanced" tab. Is the doc wrong ?

pwalczysko:
Screen Shot 2019-04-24 at 18 16 15

What does that mean ? how would I have to construct a query for MapAnnotations in the Advanced tab then ?

pwalczysko:
Bug, regression:

It seems that the "Fields"  checkboxes in the Search UI, such as "Name" , "Description", ...
do not work as expected with this PR.  This is true both in the "Advanced" tab as well as in the "Normal" tab. Basically, they do nothing (no narrowing down of search results to the "string found only in Name" etc., instead, the result table with the "Name" checked or not is almost the same. This is very different from web-dev-latest, where the checkboxes work.

This problem might be also combined with the fact that these checkboxes are somehow doubled in the UI, see screenshot.

Screen Shot 2019-04-24 at 18 29 38

pwalczysko:
I think enough problems have been found, this needs some more work, stopping testing for now.

will-moore:
It seems we do allow leading wildcard at
https://github.com/openmicroscopy/openmicroscopy/blob/f03b5313a04ca3dba2694670adbea0a750e7431f/components/tools/OmeroPy/src/omero/gateway/__init__.py#L4552
The doc says if you're using the API, you need to call ```search.setAllowLeadingWildcard(True)``` in order to search with leading wildcard. The doc is not designed for client users.

will-moore:
In the Advanced tab, you can search for Map Annotations via ```key: value```. Instead of needing to do e.g. ```tag: GFP``` to search for values of "tag" field, the Key Value e.g. ```temperature: 37``` will have created a "dynamic field" allowing you to search for ```temperature: 37```.

will-moore:
Apologies for the form fields duplication. I think that was a rebase conflict that I didn't fix properly. Should be fixed now and it should work for e.g. restricting search to Description etc. for non-Advanced search.

pwalczysko:
Bug

``details.owner.omeName: user-12`` delivers items not belonging to user-12.

pwalczysko:
Indeed, you can do complex bracketed searches like

``annotation.ns:gene AND annotation.type:MapAnnotation AND (tag:blah OR (tag:huhly AND tag:foo))``

or

``NOT file.contents:figure AND file.format:pdf``

pwalczysko:
All looks good here. We will need a doc accompanying this.

As discussed with @will-moore , I can start a help.o..py chapter with a user-facing descriptions of queries like https://github.com/openmicroscopy/openmicroscopy/pull/6008#issuecomment-487070770.

pwalczysko:
Created https://trello.com/c/vP1Z3tju/139-bug-tag-dialog-in-search for the tagging dialog bug

pwalczysko:
Created https://github.com/openmicroscopy/ome-help/pull/324 for the doc.

joshmoore:
Sorry for coming so late to the party:

 * found my map annotation: `pr6008:hBDcFtWfJgNCPA` :tada: ([link](http://web-dev-merge.openmicroscopy.org/webclient/?show=image-164115))
 * import date doesn't look to be applied via `AND`, i.e. moving end date before today doesn't exclude the above (works for regular search), though I wouldn't consider this a momentous blocker.

Thanks, @will-moore

will-moore:
That seems quite easy to support date range for creationDate (https://github.com/openmicroscopy/openmicroscopy/pull/6008/commits/3940511d9b1f8d754f72ae79687911d0ce793491), with
https://downloads.openmicroscopy.org/omero/5.4.10/api/slice2html/omero/api/Search.html#onlyCreatedBetween but not sure how to handle acquisition date (for images). I don't see anything at https://docs.openmicroscopy.org/omero/5.4.10/developers/Modules/Search.html#field-names about acquisition dates?

pwalczysko:
Failed travis

will-moore:
Travis failed with ```No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.``` Restarted...

will-moore:
To test:
 - Filtering with advanced search should respect the Date range (or just start date)
 - Advanced search doc link goes to https://help.openmicroscopy.org/search.html#advanced (will be supported in https://github.com/ome/ome-help/pull/324/ )

jburel:
I can enter a start date to search for value >date
but I cannot enter a end date to search for value < date

pwalczysko:
>I can enter a start date to search for value >date

I can confirm this observation. This is the same behaviour as the present behaviour though. Maybe @will-moore wants to enhance this feature for both the General and Advanced tab ?

The link to help is as advertised.

will-moore:
The test is not about UI input, it is to check whether the results you get from Advanced search are restricted by the date range you enter.

jburel:
@will-moore I got that nonetheless a UI RFE was noticed.

The results returned for import date were matching the entered date range. I will do further testing this am

joshmoore:
My previous test using `pr6008:hBDcFtWfJgNCPA` now gets filtered as expected with earlier end dates and later start dates. :+1:

PR #2585, opened 2014-06-06
making time_zone and first_day_of_week configurable

aleksandra-tarkowska:
--exclude, on hold

@zeb requested TIME_ZONE being configurable via ```bin/omero config set omero.web.time_zone``` to avoid local code changes.

```bin/omero config set omero.web.time_zone```
TIME_ZONE - https://docs.djangoproject.com/en/1.6/ref/settings/#time-zone
```bin/omero config set omero.web.use_tz```
USE_TZ - https://docs.djangoproject.com/en/1.6/ref/settings/#use-tz
```bin/omero config set omero.web.first_day_of_week```
FIRST_DAY_OF_WEEK - 0 (Monday) to 6 (Sunday) is the int value used by the calendar in History Tab.

PR #6009, opened 2019-04-18
Treat annotation datetimes as UTC for consistency with metadata

manics:
`metadata_general.html` displays datetimes "as is", i.e. it appears to assume the datetime returned by OMERO.server is exactly what should be displayed. In contrast `OME.formatDate` assumes the date returned by OMERO.server is UTC and should be converted for local display. By using javascript `Date.getUTC*` instead of the non-UTC methods this ensures the datetime is displayed unchanged, so as to match metadata_general.html

In practice this will often not be the expected time, but in the absence of a known timezone in OMERO.server this should ensure OMERO.web datetimes are consistently wrong.

# Testing:
1. Create a new project
2. Create an annotation (e.g. a comment) on the project
3. Check that the creation times on the project and annotation are similar (i.e. there isn't a timezone discrepancy).

Note this PR does not attempt to resolve the general OMERO problem with timezones.

# Additional info
Background screenshot of debugging the annotation datetime without this PR. It shows that Javascript is treating the datetime as UTC and converting it to localtime:
Screen Shot 2019-04-18 at 10 32 04

- https://github.com/openmicroscopy/openmicroscopy/pull/5786#issuecomment-484155143
- https://trello.com/c/HEXZNdSC/227-omeroweb-default-timezone-is-different-from-omeroserver

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
@@ -710,9 +710,13 @@ OME.formatDate = function formatDate(date) {
         }
         return n;
     }
+    // For consistency with the datetimes in metadata_general.html we have to
+    // assume dates should not be converted using the local timezone. This is
+    // because the datetime returned by OMERO may already be in "localtime",
+    // treating it as UTC ensures no local timezone adjustment is made.
     var d = new Date(date),
-        dt = [d.getFullYear(), padZero(d.getMonth()+1), padZero(d.getDate())].join("-"),
-        tm = [padZero(d.getHours()), padZero(d.getMinutes()), padZero(d.getSeconds())].join(":");
+        dt = [d.getUTCFullYear(), padZero(d.getUTCMonth()+1), padZero(d.getUTCDate())].join("-"),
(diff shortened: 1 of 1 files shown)


pwalczysko:
Works as expected on eel 


Screen Shot 2019-04-19 at 10 30 14

Screen Shot 2019-04-19 at 10 30 25


Ready to merge FMPOV.

will-moore:
Thanks for figuring that out. It's certainly an improvement.
If we did all the DateTime rendering in JavaScript, would it be easier to be consistent AND correct? 
When you say "consistently wrong", this occurs when the server and client are in different timezones, right?
Nevertheless, I think this is good to merge.

manics:
@will-moore yes, that's what I'm trying in https://github.com/openmicroscopy/openmicroscopy/pull/6010 (currently excluded).

However given the [last attempt at fixing timezones was unmerged after 4 years](https://github.com/openmicroscopy/openmicroscopy/pull/2585) I thought I'd get this minimal change merged as an uncontroversial bug-fix, then I'll rebase and unexclude #6010.

joshmoore (approved):
:+1: merging. Only thing I can think of, @manics, is that you may want to find the appropriate developer documentation and specify the path to consistency that you've laid out here. i.e. all plugins etc. should be using the same methods, especially if that will make future fixes "just work".

Edit: or considering #6010 is _reverting_ perhaps this is a matter for the release notes.  (cc @openmicroscopy/release)

PR #6012, opened 2019-04-19
Remove build-importer target: moved to omero-insight repo.

mtbc:
# What this PR does

Removes the `build-importer` target.

# Related reading

https://github.com/ome/omero-insight/blob/master/src/config/containerImporter.xml


jburel:
Changes make sense 
Merging

PR #6013, opened 2019-04-19
CLI: disable --foreground reindexing

mtbc:
# What this PR does

Disables the `--foreground` option of the CLI reindexer.

# Testing this PR

The material from https://docs.openmicroscopy.org/latest/omero5.4/sysadmins/search.html#re-indexing should still generally work except that `--foreground` should cause immediate failure.

# Related reading

https://trello.com/c/XdwEzrvs/73-indexer-log-hibernate-error


jburel:
Ran ``bin/omero admin reindex --foreground``
A message is returned immediately
``foreground indexing is not available``

jburel:
Ran also other commands e.g. ``bin/omero admin reindex --dryrun``. No change in behaviour

PR #6021, opened 2019-04-29
Client jar

jburel:
# What this PR does

Remove client_jar
this was only used by the matlab toolbox and not distributed


joshmoore:
West was green this morning and none of these changes particular concern me. Very much looking forward to encouraging users to consume `omero-gateway`!

jburel:
Merging in that case

PR #5963, opened 2019-02-27
Web logout redirects to configured login page

manics:
OMERO.web allows the login page to be replaced with an alternative Django view: https://docs.openmicroscopy.org/omero/5.4.10/sysadmins/config.html#omero-web-login-view

This view is used by the default login check https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.0-m3/components/tools/OmeroWeb/omeroweb/decorators.py#L122
but the logout view doesn't take this property into account

# Testing this PR

1. Set an alternative login view `omero config set omero.web.login_view someotherviewname`
2. Logout
3. Check that you are directed to the configure view, and not the default login view

I've also added some docs for the `omero.web.login_view` property

PR #6024, opened 2019-05-06
Bug fixes

atarkowska:
# What this PR does

replace #5936
Fix invalid redirection after logout. This is only exposed when developer has to customise the view

# Testing this PR

There were no test prior to that fix.

Without fix test is should fail as follow
```
>       assert rsp['Location'].endswith(reverse('test_weblogin'))
E       assert <built-in method endswith of str object at 0x112f11db0>('/test_login/')
E        +  where <built-in method endswith of str object at 0x112f11db0> = 'http://testserver/webclient/login/'.endswith
E        +  and   '/test_login/' = reverse('test_weblogin')
```


will-moore:
Seems that these changes already got merged in https://github.com/openmicroscopy/openmicroscopy/pull/5963?

PR #6028, opened 2019-05-14
Drop -java from omero-java-gateway

joshmoore:
see: https://github.com/ome/omero-java-gateway/pull/8


joshmoore:
Merging based on green travis. Jar is being pulled from artifactory.

PR #6030, opened 2019-05-17
resolve dependencies

jburel:
# What this PR does

Make sure that some BF dependencies are found
Starting with an empty ``.m2``, the change was required in order to build

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -196,6 +196,6 @@ versions.velocity=1.4
 ###
 
 ## Internal dependencies
-versions.omero-blitz=5.5.0-m9
-versions.omero-common-test=5.5.0-m8
-versions.omero-gateway=5.5.0-m6
+versions.omero-blitz=5.5.0
+versions.omero-common-test=5.5.0
+versions.omero-gateway=5.5.0
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -138,6 +138,8 @@
     <module organisation="edu.ucar" name="bufr" resolver="maven-resolver"/>
     <module organisation="edu.ucar" name="thredds-parent" resolver="unidata-resolver"/>
     <module organisation="edu.ucar" name="netcdf" resolver="unidata-resolver"/>
+    <module organisation="edu.ucar" name="cdm" resolver="unidata-resolver"/>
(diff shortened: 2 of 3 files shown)


sbesson (commented):
The changes read well although the current HEAD of `develop` points to a blitz version which does not include the `netcdf` upgrade so I wasn't able to reproduce the failure using `./build.py build-dev -Dmaven.repo.local=/tmp/.m2/repository`.  Do we expect to have another milestone with the Bio-Formats 6.1.0 version bump this week? If so, I can certainly retest both the failure as well as the impact this PR with it.

An alternative to declaring all artifacts would be to include a generic module for all `edu.ucar` artifacts except `bufr` and `grib` e.g. as in https://github.com/openmicroscopy/openmicroscopy/blob/cad1883781d37ab9b9b4dba2b96c8167d27191ac/etc/ivysettings.xml#L139-L141.

jburel:
I should have indicated that I was building everything locally with SNAPSHOT

I will prepare probably a final milestone this week with BF 6.1 and slice doc generation

jburel:
Same error https://travis-ci.org/ome/omero-build/builds/536358571

etc/omero.properties
```diff
@@ -196,6 +196,6 @@ versions.velocity=1.4
 ###
 
 ## Internal dependencies
-versions.omero-blitz=5.5.0-m9
-versions.omero-common-test=5.5.0-m8
-versions.omero-gateway=5.5.0-m6
+versions.omero-blitz=5.5.0
```
sbesson, line 199: Isn't that pulled transitively by `omero-gateway` ?

sbesson (approved):
Tested locally by reverting 2398023 and running a clean build `./build.py build-dev -Dmaven.repo.local=/tmp/.m2/repository`. I could reproduce the failure mentioned in the description.

With the `ivysettings.xml` fix, the build complete as expected. The CI tests show that there is no regression in the ability to start and the behavior. Happy to see this merged including the bump to `omero-gateway:5.5.0`

etc/omero.properties
joshmoore: Most of blitz should be excluded by gateway so both are needed.

PR #6031, opened 2019-05-21
Remove the generation of Ivy dependencies.html files

sbesson:
With the recent work on OMERO 5.5, these files currently weigh 169M out of 480M for the complete server bundle. In addition to being extremely large, their size makes them almost non-usable.

## Proposed changes
These generation of `dependencies.html` using the [Ivy report](https://ant.apache.org/ivy/history/2.4.0/use/report.html) tasks was introduced in 5a8db9c7be8ccc527b4b5ac5d501eb1e74c423a8 as part of the decoupling of Bio-Formats. For developers and especially for reviewing dependencies, having the dependency tree under some form and being able to review the version evictions still remains a useful resource. However, having it large HTML files as part of every bundle feels unnecessary.

## Testing

Without this PR, a build server (or the last 5.5.0 milestone) zip should be ~160MB or 480MB unzipped. With it included, the size should be largely reduced.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -446,10 +446,6 @@ To get started using Eclipse, execute "ant build-dev" and import the top-level
         <!-- sync="true" deletes any other files like services.jar or extensions.jar which may be under lib -->
         <ivy:resolve file="ivy.xml" type="jar,egg,bundle,zip" conf="client" settingsRef="ivy.toplevel" log="quiet"/>
         <ivy:retrieve conf="client" pattern="${dist.dir}/lib/client/[artifact].[ext]" sync="false" log="quiet" settingsRef="ivy.toplevel"/>
-        <mkdir dir="${dist.dir}/share/client"/>
-        <ivy:report conf="client" todir="${dist.dir}/share/client"
-            outputpattern="dependencies.html" graph="false"
-            settingsRef="ivy.toplevel"/>
     </target>
 
     <target name="copy-server" depends="init">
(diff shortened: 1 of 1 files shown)


jburel:
From the discussion we had, this PR did not have the expected size reduction.

sbesson:
Finally managed to spend a bit of time on it:

```
sbesson@ls30630:Downloads $ du -csh OMERO.server-5.5.0-rc1-*
310M	OMERO.server-5.5.0-rc1-56-71c02ab-ice36-b39
272M	OMERO.server-5.5.0-rc1-56-71c02ab-ice36-b39.zip
481M	OMERO.server-5.5.0-rc1-ice36-b116
273M	OMERO.server-5.5.0-rc1-ice36-b116.zip
```

So this PR does not save much in terms of bundled server size (assuming most of the generated HTML are easily compressed) but it still saves ~170M when uncompressing the bundles.

Size being one aspect, another point for discussion is the usability of these files. I initially tried to use them to troubleshoot some dependency conflict/eviction. But it turns out the size makes these reports barely usable as they are (and I could not find an option to shorten the report).

mtbc:
I've found them useful for checking for conflicts and their outcome so it's nice they can be generated but I don't see why they need to be packaged in the artifact and I guess Ivy's days are numbered anyway.

mtbc:
(I just do something else while the browser opens the page, then I use ^f to search for what I'm checking.)

jburel:
@mtbc ok to merge this size reduction PR?

mtbc:
Okay to merge it. RFE to generate but *not* package it though because the report *is* useful: for example, just yesterday when I had set OMERO dependencies to 5.5.0-SNAPSHOT instead of the new 5.5.1-SNAPSHOT in testing https://github.com/ome/omero-server/pull/51 `dependencies.html` substantially assisted the diagnosis.

sbesson:
If the report is still helpful, proposing to move these calls to a separate target that does not copy into the server bundles. Proposing `generate-dependencies-report` as the target name, other ideas? And should it included in the default `build-dev` or opt-in?

mtbc:
All sounds good to me. (Don't mind if it's in default or not.) Fine if it all happens in some follow-up PR next month too. Cheers!

jburel:
Let's open an issue for the generation of report but not packaging of the report

jburel:
Issue created see https://github.com/openmicroscopy/openmicroscopy/issues/6038

PR #6034, opened 2019-05-27
add Jinja2

jburel:
# What this PR does

Add dependencies

See https://trello.com/c/QaxIg0AH/156-jinja-warning

cc @joshmoore

### DIFF
--- a/components/tools/OmeroWeb/requirements-py27.txt
+++ b/components/tools/OmeroWeb/requirements-py27.txt
@@ -11,3 +11,4 @@ gunicorn>=19.3
 
 omero-marshal==0.5.4
 django-redis>=4.4,<4.9
+Jinja2


jburel:
Discussed with @mtbc, the general approach has been to used th package manager see https://docs.openmicroscopy.org/omero/5.4.10/sysadmins/unix/server-centos7-ice36.html
so I will update omeroweb-install instead of the requirements file

jburel:
Re-opening this PR ansible-role-omero-web installed the dependencies using the requirement file
see https://github.com/openmicroscopy/ansible-role-omero-web/blob/master/tasks/web-install.yml
cc @joshmoore

joshmoore:
As long as web.py imports jinja and lives in the server repo, this will also need to be done for server installations as well. cc: @manics

jburel:
Actually  https://github.com/ome/omero-web-docker/blob/master/requirements.yml
and https://github.com/ome/omero-server-docker/blob/master/requirements.yml use https://github.com/openmicroscopy/ansible-role-omero-python-deps
which installed ``python-jinja2`` https://github.com/openmicroscopy/ansible-role-omero-python-deps/blob/master/tasks/main.yml

So this change should not needed.

PR #6036, opened 2019-05-28
error handling

jburel:
# What this PR does

Catch possible exception in case someone does not have the attribute available
This is a safety net

Check that you can open a script on eel (python 2.6)


mtbc:
We don't support 2.6?

jburel:
sorry I was thinking of a "gentle" transition but yes we are no longer supporting 2.6
closing to avoid confusion

jburel:
As discussed this morning, merging so we include in rc2 so we can test on eel.

PR #6042, opened 2019-06-04
Workaround indexing of image name

will-moore:
# What this PR does

As discussed with @jburel, this works around an indexing bug for Image Name in OMERO 5.5.
Instead of searching for ```"foo"``` we search for ```"foo OR name:foo"``` if no fields are specified and we're searching for images and the query doesn't already contain ```:```.

This applies to standard search and advanced search (IF the user doesn't enter any ```:```).

To discuss: Should we avoid modification of the query if the user is using Advanced search?
I want to avoid case where:
 - Search for "foo" under regular search -> find images named "foo"
 - Search for "foo" under advanced search -> Don't find images named "foo".

But, maybe we shouldn't modify search term if it contains ```AND``` keyword?
E.g. ```"foo AND bar"``` -> ```"foo AND bar OR name:foo AND bar"``` doesn't make sense.

# Testing this PR

1. e.g. search for ```dm2``` on eel-latest. Should find images named ```dm2```

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/controller/search.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/controller/search.py
@@ -61,6 +61,10 @@ class BaseSearch(BaseController):
             fields = fields.union(("file.name", "file.path", "file.format",
                                    "file.contents"))
             fields.discard("file")      # Not supported
+        # workaround indexing bug in OMERO 5.5.0. Image Name is not indexed
+        # If no fields are specified, also search by Name
+        if len(fields) == 0 and "images" in onlyTypes and ':' not in query:
+            query += ' OR name:' + query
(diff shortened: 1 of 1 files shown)


will-moore:
@jburel Just discussed with @mtbc. This looks a bit more complex than first thought. If I search for ```GFP H2B``` we really need to search for ```name:GFP or GFP or name:H2B or H2B```.
Also, if anyone uses the Search API then this fix won't help. Any chance this could be implemented server-side to fix that?

jburel:
I agree that if the changes are in the client then of course with someone uses the search API this will have different results. 
Exploring all avenues since we do know where the problem is

jburel:
Closing this PR. Thanks @will-moore for helping it. We think we have found the source of the problem.

PR #6044, opened 2019-06-06
web: incorrect login text can be customised

manics:
If a user's OMERO.web credentials are incorrect a custom error message can be displayed. This is useful when a non-standard process is required for checking credentials.

# Testing this PR

1. `omero config set omero.web.login_incorrect_credentials_text 'Please check your user name and password. If login still fails please check your details at https://lsd.lifesci.dundee.ac.uk/'`

2. Attempt to login to a valid server with incorrect credentials.

Note this does not work when `omero.web.check_version=false`.

Screen Shot 2019-06-06 at 14 31 56


mtbc:
Works well for me. Perhaps I should attempt a nightshade playbook PR!

mtbc (approved):

PR #6055, opened 2019-06-13
Table download

will-moore:
This PR has now been ported to https://github.com/ome/omero-web/pull/3 but tests remain in this one.

In several training workshops, explaining what an OMERO.table is and what data is contains is made difficult because it is very hard to show a table. Even when downloaded, it is not easy see the contents of a table since it's an HDF file.
For users of tables, inspecting the data they contain for trouble-shooting purposes etc is also hard. It is only possible to see the data for a single object/row at a time.

This allows users to view an OMERO.table file as an html table and to download it as a CSV.

# Testing this PR

1. Find an OMERO.table attached to an object e.g. Project, Screen etc. Tables are recognised by the ```openmicroscopy.org/omero/bulk_annotations``` namespace (you can find objects annotated with this namespace by searching for 'bulk'. E.g. http://web-dev-merge.openmicroscopy.org/webclient/search/?search_query=bulk (user-3)
2. Files will show an 'eye' (view) icon. When clicked, this opens the file as an html table in another tab.
3. The table page also has links to download the table as CSV (or show as JSON)
4. Click to download CSV and open in Excell.
5. Also click to view JSON (might be useful for devs?)
6. Check new web test is passing

NB: the default action of clicking the OMERO.table annotation itself to download the HDF file remains the same. Do we want the default action to open the html table (with links there allowing download as HDF or CSV) since download as HDF is rarely very useful?


will-moore:
Test failing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/56/testReport/junit/OmeroWeb.test.integration.test_plategrid/TestScreenPlateTables/test_get_plate_table/

manics:
Testing on merge-ci:

Icon `https://merge-ci.openmicroscopy.org/static/webgateway/img/icon_view.png` is broken (404 error)
Screen Shot 2019-06-14 at 14 56 00

Clicking on the icon opens a new tab:
Screen Shot 2019-06-14 at 14 59 36-fullpage

If you create an OMERO.table through the TablesAPI `session.sharedResources().newTable(0,'/test.h5')` the OriginalFile has a mimetype `OMERO.tables`. Could you use that instead of or as well as looking at the annotation namespace?

manics:
I think there needs to be limit on how much data is displayed, I tried to open a large table and it timed out.

dominikl:
That's pretty cool 👍 Always bit a hassle to download hdf file then open in HDFCompass etc.

will-moore:
@manics I'll look into adding pagination...

pwalczysko:
Works as expected.

I would suggest maybe a center position in the column numbers instead of left-bound ?

Screen Shot 2019-06-17 at 15 20 02


Regarding the display of very large tables - I think it would be of some advantage to display just some 50 rows or so from the top if the table is very large. Insisting on displaying the whole very large table seems not appropriate imho - after all, the tables are containing sometimes very large data and that is why the structure and limitations of this container are as they are. 
But, always display at least the headers, even if the table is very big, would be my bet.

pwalczysko:
After further discussion with @will-moore , maybe have some pagination, but either offer changing of the size in the url or make the page size generous (500 - 1000). There is possibly some value in having the pagintation as opposed to showing just first xxx rows.

pwalczysko:
not sure about the json. We could not come up with a good workflow where this would be useful.

pwalczysko:
As for changing the default downloads behaviour of the attachment when clicked - I would guess no, leave it as it is, this does not seem to be justified change (big change for a small functionality)

joshmoore:
This came up several times in Cologne, and I think it could be a big win. A few concerns:

 * definitely concerned about pagination for all the end points, but I'm not sure what we can do about the CSV case.
 * For the JSON, perhaps a json-ld style link could provide a link to the next download. (This is similar to @pwalczysko's suggestion for understanding the use-case first.) e.g. [atlassian](https://developer.atlassian.com/server/confluence/pagination-in-the-rest-api/), or [github](https://developer.github.com/v3/guides/traversing-with-pagination/)
 * I assume we will need to support these URL endpoints moving forward. If we foresee them needing to adapt (e.g. for the microservices world) it'd be good to design that flexibility in now.
 * Finally, is there any way to make this extensible so someone could plugin in their own viewer a la open_with?

joshmoore:
Note: I also saw a number of these:

```
  File "/opt/hudson/workspace/WEB-DEV-merge-deploy/OMERO.py/lib/python/omero/gateway/__init__.py", line 4700, in __call__
    return self.f(*args, **kwargs)
  File "/opt/hudson/workspace/WEB-DEV-merge-deploy/OMERO.py/lib/python/omero_api_IQuery_ice.py", line 362, in find
    return _M_omero.api.IQuery._op_find.invoke(self, ((klass, id), _ctx))
SecurityViolation: exception ::omero::SecurityViolation
{
    serverStackTrace = ome.conditions.SecurityViolation: Cannot read ome.model.core.OriginalFile:Id_48927
        at ome.security.basic.BasicACLVoter.throwLoadViolation(BasicACLVoter.java:300)
```

pwalczysko:
@will-moore tables for testing https://outreach.openmicroscopy.org/webclient/?show=project-872 (Fiji)
https://outreach.openmicroscopy.org/webclient/?show=project-1273 (R)
https://outreach.openmicroscopy.org/webclient/?show=plate-55 (two tables, 5 wells only)
https://outreach.openmicroscopy.org/webclient/?show=plate-102 (one table, all 96 wells)

will-moore:
To test:
 - visit the table page as above (click on eye in table file)
 - Should see pagination controls that show the first 200 rows by default.
 - We also support query in the URL e.g. ```?query=ColumnName>123```. Pagination should work on the results of each query and the hits from the query should be shown.

joshmoore:
:+1: for the pagination addition, but unless this can be clearly marked experimental, etc. I'd be concerned about merging this even _with_ pagination unless we discuss some of https://github.com/openmicroscopy/openmicroscopy/pull/6055#issuecomment-502749630

will-moore:
Re: https://github.com/openmicroscopy/openmicroscopy/pull/6055#issuecomment-502749630

 - [x] Pagination (except for csv)
 - JSON link I have removed. It was just thought based on a similar feature I saw at e.g. https://www.ebi.ac.uk/biostudies/studies/S-BSST261. And we have the link anyway (we get it for free because it simply returns the dict as JSON instead of rendering to template). But we don't do it anywhere else in the webclient so it's not something users would expect.
 - Re: "support these URL endpoints moving forward" and "flexibility to adapt to microservices". I don't know that this feature and URLs are any different from other webclient URLs in terms of future support or potential migration to microservices? What sort of flexibility is needed to migrate to microservices?
 - We have considered ```FileAnnotation -> Open with...``` feature when I originally introduced the Open-with for P/D/I etc. It's still a potentially useful feature, but outwith the scope of this PR. However, you could already do Open-with for a Project or Screen etc that has a Table attached (similar to what Parade does) and that wouldn't rely on this PR.

will-moore:
To test:

 - visit the table page as above (click on eye in table file)
 - Should see pagination controls that show the first 200 rows by default.
 - We also support query in the URL e.g. ```?query=ColumnName>123```. Pagination should work on the results of each query and the hits from the query should be shown.
 - We show a couple of example queries with a short explanation. Do these work OK?
 - Added some more tests of pagination and querying - should be passing

joshmoore:
re: https://github.com/openmicroscopy/openmicroscopy/pull/6055#issuecomment-503552183

So, I think my concerns fall under the general umbrella of what will this look like in the future? It might not be that moving to a microservice significantly changes the API here, but if the API then change, the cost of keeping everything in sync will rise. In general, I think this will excite people a lot. But that in turn means we're likely to get lots of suggestions that we haven't thought about yet, and I'm trying to get us to think about that _some_ so we're not left unable to respond quickly. 

Random questions include: What would `open_table` look like if it were a plugin? Would you do anything differently? If someone suggests an argument that doesn't match the current API, would you add an `open_table2`? Will `downloader` need to reproduce this format in Java code? Could both of them use the same base service?

will-moore:
I can imagine users wanting more functionality both in the html page and in the JSON data. E.g. "only show named columns" or "sort by column x". Optional features like these would not break existing behaviour and could be added to the URL parameters e.g. ```?columns=Results,Image&order_by=Image```. (although currently pagination happens before we read the data from the table so sorting and pagination may not be compatible).
If the html page gets more functionality, it might need to include JavaScript functionality and AJAX etc. 
I'm not sure what you mean by "What would ```open_table``` look like if it were a plugin?" But it sounds like Parade in a stand-alone window. If a plugin wanted to use this API, I could imagine there could be additional features required (e.g. parade would only need to load named columns for filtering) but I don't think we need to implement every feature we can imagine. OMERO.web plugins can also access the Python API directly so they are not reliant on this JSON api.
If downloader were to download OMERO.table as csv, then there's very little code needed on top of the tables API to convert the data into csv. Doesn't seem to justify the creation of a new base service.

pwalczysko:
```

error | "Error executing query: >"
-- | --


```
After clicking onto 
Screen Shot 2019-06-21 at 16 39 36
when editing as 

user-3 on merge-ci the table on https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-2955

will-moore:
@pwalczysko I think the last commit hadn't been merged for some reason when you tested there (it is merged now). I now don't try to show an example query if the table doesn't have any ```DoubleColumns```.
You can see the column types for a table in the JSON response. e.g. for that table it is 
https://merge-ci.openmicroscopy.org/web/webclient/omero_table/188164/json/ 
```
column_types": [
"WellColumn",
"StringColumn",
"StringColumn",
"StringColumn",
"StringColumn",
"StringColumn",
"StringColumn",
"StringColumn",
"StringColumn"
],
```
Whatever created that table should probably try to create ```DoubleColumn``` where the values are numbers.

components/tools/OmeroWeb/test/integration/test_table.py
```diff
@@ -0,0 +1,164 @@
...
+            col_data = [row[idx] for row in rows]
+            if col_type == StringColumn:
+                data.append(StringColumn(name, '', 64, col_data))
+            else:
+                data.append(col_type(name, '', col_data))
+
+        table.addData(data)
+        table.close()
+
+        orig_file = table.getOriginalFile()
```
joshmoore, line 92: All calls to `getOriginalFile` will need to be moved _before_ `table.close()` Cf. https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_table/TestOmeroTables/test_table_html/

pwalczysko:
>Whatever created that table should probably try to create DoubleColumn where the values are numbers.

@will-moore it was our own Populate Metadaata script... cc @manics

will-moore:
Need to update failing tests, e.g. https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_table/TestOmeroTables/test_table_html/ to close table AFTER getting orig file.

joshmoore:
Added `--exclude` to the description in order to test the closing fix for 5.5.1.

will-moore:
This is going to conflict with https://github.com/openmicroscopy/openmicroscopy/pull/6070.
Closing until that is merged.

pwalczysko:
the ``Download as CSV`` button, in case of a multi-page table, downloads just one page, not the whole table. Would like to donwload the whole table in one go.

pwalczysko:
One RFE. Otherwise works fine.

will-moore:
With that commit, I added link to download the "Whole Table" or the "Current View" (filtered and paginated). This means that even for users with very large tables, you will be able to download some of it to CSV, and is also useful to output the result of a query.

pwalczysko:
As discussed in the office, it would be nice to have a tooltip showing what type (double, string, float) the column is.

will-moore:
To test last commits:
 - Download as CSV: Whole Table | This View - links should work as expected. Try filtering and pagination - 'This View' should download CSV that corresponds to the rows currently shown
 - Column headers have tooltip to indicate the column type.

will-moore:
To test last commits:

- Download as CSV: Whole Table | This View - links should work as expected. Try filtering and pagination - 'This View' should download CSV that corresponds to the rows currently shown
- Column headers have tooltip to indicate the column type.
- Well column and Image column should contain links to the Well/Image by ID e.g. webclient/?show=image-123 and open that in a new browser tab.

manics:
`user-3` https://merge-ci.openmicroscopy.org/web/webclient/omero_table/77506/ 
This can be opened, but downloading the full table as a CSV failed

manics:
Also any thoughts on my earlier suggestion about allowing tables with mimetype `OMERO.tables` to be opened regardless of namespace? https://github.com/openmicroscopy/openmicroscopy/pull/6055#issuecomment-502120420

pwalczysko:
>user-3 https://merge-ci.openmicroscopy.org/web/webclient/omero_table/77506/
>This can be opened, but downloading the full table as a CSV failed

Yes, this is gateway timeout. I think this is fair enough as the table has over 18 thousand rows. I do not see any difference with the image downloading behaviour on web here. You can download single pages. Or maybe (in future ?) use OMERO.downloader cc @mtbc

will-moore:
Tests failing at:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/82/

pwalczysko:
FMPOV, after the comment of @manics https://github.com/openmicroscopy/openmicroscopy/pull/6055#issuecomment-510805767 is dealt with, this is good to go

will-moore:
Those commits add support to files with mimetype ```OMERO.tables```.
They should now show the 'eye' icon in right panel to view the table.

mtbc:
> maybe (in future ?) use OMERO.downloader

File attachment download is one of several things I got halfway through before something more urgent intervened, happy to resume once enough prioritized again. Could certainly be worth adding a Data Out card suggesting tables could also be exported into some other format (especially if they're readable from Java, no idea if they're a special Python thing) if that's your thinking - I guess HDF5, CSV, XML or somesuch?

pwalczysko:
travis failed

will-moore:
Tests now passing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/85/

manics:
👍  Setting the mimetype with no namespace works as expected

Behaviour with a `LongArrayColumn` is a bit unexpected, the viewer looks OK
Screen Shot 2019-07-16 at 11 58 54

But the downloaded CSV contains:
```csv
string columns,double column,log array column
a,-0.1,[1L. 2L. 3L]
bc,2.3,[-10L. -20L. -30L]
d e f,400000.0,[400L. 500L. 600L]
```

https://merge-ci.openmicroscopy.org/web/webclient/omero_table/325536/ `user-3`

manics:
newlines aren't handled in the downloaded CSV
Screen Shot 2019-07-16 at 12 08 23

```csv
string columns,double column,log array column
a,-0.1,[1L. 2L. 3L]
bc,2.3,[-10L. -20L. -30L]
d e f,400000.0,[400L. 500L. 600L]
..""'',0.0,[1L. 2L. 3L]
new
line,2.0,[-1L. -2L. -3L]
```

will-moore:
@manics Thanks for that. I can see that there's a bit more work to do on the CSV writing front to handle less common data types like ```LongArrayColumn``` and values containing new lines.
For this I'll need to switch to using the ```csv``` writer.
I also wanted to look at streaming csv instead of loading it all into memory to handle bigger tables.
Since this PR is now quite big (47 commits) and I'm not likely to get that done before I'm away, that could be handled in a follow-up PR.

pwalczysko:
Agree with @will-moore - this csv functionality is not crucial for the focus of this pr, and principally it does work - possibly this could be done in a follow-up PR ?

joshmoore:
Pushed a fix for https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/40/testReport/OmeroWeb.test.integration.test_table/TestOmeroTables/test_table_html/

will-moore:
Closing temporarily while #6171 is closed so tests aren't broken at https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/46/testReport/

will-moore:
Tests passing on py-3 at https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/96/testReport/OmeroWeb.test.integration.test_table/TestOmeroTables/

PR #5763, opened 2018-05-09
IDR: use new logo (rebased onto metadata54)

manics:
This is the same as gh-4989 gh-5230 but rebased onto metadata54.

Place the logo in the repository to prevent it being lost between upgrades.

PR #6057, opened 2019-06-17
Webclient top-bar logo can be customised

manics:
# What this PR does
New properties to customise the top-left webclient logo after login:
- `omero.web.top_logo`: URL to externally hosted logo
- `omero.web.top_logo_link`: Hyperlink for the logo

# Testing this PR
Set a custom logo (image must be hosted outside OMERO.web) and link for when you click on the logo. Default behaviour is unchanged.
```
omero config set omero.web.top_logo_link https://idr.openmicroscopy.org
omero config set omero.web.top_logo https://idr.openmicroscopy.org/static/webgateway/img/logo2.png
```

# Related reading
- https://github.com/openmicroscopy/openmicroscopy/pull/5763

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -668,6 +668,18 @@ CUSTOM_SETTINGS_MAPPINGS = {
          ("The error message shown to users who enter an incorrect username "
           "or password.")],
 
+    "omero.web.top_logo":
+        ["TOP_LOGO",
+         "",
+         str,
+         ("Customize the webclient top bar logo. The recommended image height "
+          "is 23 pixels and it must be hosted outside of OMERO.web.")],
+    "omero.web.top_logo_link":
+        ["TOP_LOGO_LINK",
+         "",
+         str,
+         ("The target location of the webclient top logo, default unlinked.")],
+
     "omero.web.user_dropdown":
         ["USER_DROPDOWN",
          "true",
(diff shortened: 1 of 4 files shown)


mtbc (approved):

will-moore:
Works fine and the code looks good too.
Good to merge.

components/tools/OmeroWeb/omeroweb/settings.py
```diff
@@ -668,6 +668,18 @@ def leave_none_unset_int(s):
          ("The error message shown to users who enter an incorrect username "
           "or password.")],
 
+    "omero.web.top_logo":
+        ["TOP_LOGO",
+         "",
+         str,
+         ("Customize the webclient top bar logo. The recommended image height "
```
joshmoore, line 675: Is there a recommendation on the format of the logo URL? i.e. is it safest to use `//my.host/logo.png` to deal with both http and https deployments? Or better to suggest always using https?

mtbc: Incidentally, I was kind of surprised with my own testing that https: to a different host worked but file: URLs didn't. Maybe that was a Firefox "feature".

manics: I think sysadmins can figure out what's best for themselves. We have at least one other place where a url can be specified and I don't think we should add detailed basic advice to every property. The image can also be hosted on the same server in another application so the `protocol://hostname` isn't always necessary

PR #6058, opened 2019-06-18
CLI ldap setdn can enable and disable LDAP for users.

mtbc:
Fixes `bin/omero ldap setdn -h` so the one-line summary includes that it can work both ways for enabling/disabling LDAP. See diff.


joshmoore:
:+1:

PR #6062, opened 2019-06-19
tables: load sessions quietly

joshmoore:
TablePrx instances which are not closed by the user
should not keep sessions alive.

see: https://github.com/ome/omero-server/pull/56


joshmoore:
Merging along with ome/omero-server#56. Note this won't _do_ anything until ome/omero-blitz#60 is merged and the version in this repo has been bumped.

PR #6063, opened 2019-06-19
web login: dynamically lookup client downloads

manics:
Insight has moved to GitHub releases. The `https://downloads.openmicroscopy.org/latest/omero5.5/insight-*` download links are no longer supported so instead this looks up the required download from GitHub by matching the first two components of the version i.e. it looks for the latest tag matching `v{major}.{minor}.[^-]+` (the `[^-]+` _should_ avoid matching prereleases).

This should be failsafe- in the event of a failure the default URL is https://www.openmicroscopy.org/omero/downloads/

# Testing this PR

The download links at the bottom of the login page should download the latest 5.5 insight clients from GitHub. If there's a failure (e.g. you've blocked requests to external resources) it should link to https://www.openmicroscopy.org/omero/downloads/

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -732,12 +732,10 @@ CUSTOM_SETTINGS_MAPPINGS = {
          parse_boolean,
          ("Whether to link to official client downloads on the login page")],
     "omero.web.login.client_downloads_base":
-        ["CLIENT_DOWNLOAD_LATEST_BASE",
-         'https://downloads.openmicroscopy.org/latest/omero{major}.{minor}',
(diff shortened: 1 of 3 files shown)


joshmoore:
Not sure how to simulate a failure to test that code path, but otherwise this looks to be working. Anyone have thoughts re: inclusion in 5.5.1?

will-moore:
Not tested, but looks good.

joshmoore:
see [conversation](https://openmicroscopy.slack.com/archives/C0K5EED3R/p1562339005489000) about rc releases. How do we do the equivalent of building a rc version of insight such that we can consume it but users don't automatically start downloading it?

manics:
Isn't that already handled by the `[^-]+$` in the regex?

joshmoore:
Awesome. It might "just work". Thanks.

PR #6066, opened 2019-06-21
examples: no remote calls on TablesPrx after close

joshmoore:
As part of https://github.com/ome/omero-blitz/pull/57, table.close()
is now much more definitive. No remote calls will be possible on the
instance once the wrapper has removed the servants server-side.

# Related reading

 * https://github.com/ome/omero-blitz/issues/56


joshmoore:
This PR is safe with or without any of the underlying code. Merging so that the autogenerated docs can be updated for review.

PR #6068, opened 2019-06-22
Fix layout of my_account form

will-moore:
# What this PR does

Fixes my-account page. See https://trello.com/c/lrHDbfX9/658-bugregerssion-css-on-user-settings-in-web-550

# Testing this PR
 - check that the "User settings" form looks not-broken.

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
+++ b/components/tools/OmeroWeb/omeroweb/webclient/static/webclient/css/layout.css
@@ -2190,6 +2190,9 @@ ul.criteria_right input{
 		overflow:hidden;
 		height: 400px;
 	}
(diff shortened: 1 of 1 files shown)


joshmoore:
@will-moore : do you have any idea what caused the regression?

mtbc:
Regardless of Travis this PR works well for me and the code change looks sensible.

Before:

before

After:

after

joshmoore:
Thanks, @mtbc. It would still be interesting to know if there's anything we can do (place a warning on the template, etc) to keep these things in sync. Merging.

will-moore:
@joshmoore Sorry, I started trying to check older versions of web for when this broke, but ran into trouble trying to run and build older servers.

PR #6075, opened 2019-07-03
Right panel performance fixes

will-moore:
See https://trello.com/c/vv4mqsAi/5752-omero-omeroweb-on-demo-is-slow
This PR improves the performance of the right-hand metadata panel, particularly for Images and noticeable with large numbers of users in the current group.

Calling ```conn.getObjects()``` or ```conn.getObject()``` with group context ```-1``` is slow with large numbers of users in the group. We *could* use the current group of the webclient (```session['active_group']```) to set the group context, but it could result in 404 if user has e.g webclient open in 2 browser tabs with a different group in each etc.
Instead, this PR reduces the number of times this is called to 1, instead of 3. So it should be 3 x faster.
Also, to avoid the risk of any additional calls to ```getObject()``` being added in the future, we set the group context once we have loaded the Object and know the group.

For example, using ```{{ ome.user.id }}``` in a template was causing ```getObject("Experimenter", id)``` to be called. Setting the context improved the speed of this, although I also replaced this with ```{{ ome.user_id }}``` which removes the call entirely.

To test functionality (not broken):
 - check that the right panel still loads and displays metadata for P/D/I etc (not broken)
 - Try adding comments etc.
 - check that right panel still works for Shares, adding Comment to the Share.

To test performance (learning server with large number of users):
- Log in to the 'learning' test server from https://merge-ci.openmicroscopy.org/web/webclient/login/

I guess if we want to quantify speed improvements, we could time the right panel loading with and without this PR merged?

cc @mtbc, @pwalczysko

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/decorators.py
@@ -133,7 +133,10 @@ class render_response(omeroweb.decorators.render_response):
         context['ome']['eventContext'] = eventContextMarshal(
             conn.getEventContext())
         context['ome']['user'] = conn.getUser
-        context['ome']['user_id'] = conn.getUserId()
+        context['ome']['getUserId'] = conn.getUserId
+        # Current active user_id (who's data are we looking at)
+        context['ome']['user_id'] = request.session.get('user_id',
+                                                        conn.getUserId())
         context['ome']['group_id'] = request.session.get('group_id', None)
         context['ome']['active_group'] = request.session.get(
             'active_group', conn.getEventContext().groupId)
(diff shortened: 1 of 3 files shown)


rgozim:
Provided the number of users is similar to what was on demo back in 2018, then this looks like the performance is better. Previously it was taking longer than 10 seconds to view content on the right hand panel, and although not instantaneous, it is substantially quicker with this fix/"quick win".

rgozim:
On a side note, clicking on the `Acquisition` tab in the right hand menu is a bit slow (not terribly so), and maybe could be do with being added as fix to this PR - if that is something that is also affected by the "group id -1" issue.

mtbc:
Good to know that OMERO 5.5.2 should be a further improvement for Paul then!

will-moore:
Ported to https://github.com/ome/omero-web/pull/17/

PR #6077, opened 2019-07-03
Get object set group context if none

will-moore:
This is a proof of concept for improving the loading speed of many OMERO.web pages, based on suggestion from @rgozim at https://trello.com/c/vv4mqsAi/5752-omero-omeroweb-on-demo-is-slow#comment-5bedae4959582c6b524d8d6e

Instead of using group context: ```-1``` by default, we use ```@login_required(setGroupContext=True)``` to improve the performance of ```conn.getObject()``` when we are in a group with many other users.

This could lead to the possibility of 404 errors if you are loading objects in a different group from the main webclient. E.g. open webclient in 2 different browser tabs, switch to a different group in tab-2, then browse the tree in tab-1.
We avoid these 404 errors by overriding the ```conn.getObject()```. If this returns None on the first attempt, we try again with groupContext ```-1```.

If we have loaded an object on the second attempt, we also set ```conn.SERVICE_OPTS.setOmeroGroup(gid)``` so that subsequent queries for objects in the same group and successful first time.

Currently in this PR I have only used ```@login_required(setGroupContext=True)``` on the right-hand Preview panel (and on the ```imgData``` JSON it loads).

To test:
 - The Preview panel should be much faster when viewing images in the current group.
 - Even when viewing images in a different group (using 2 browser tabs as above) it should still be a bit faster since we only have to do one ```group:-1``` query per request.

NB: if this approach looks good, we can add ```@login_required(setGroupContext=True)``` to other methods, testing each carefully. We don't yet handle 404 for ```conn.getObjects()``` (batch loading).

cc @mtbc @chris-allan

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -1580,7 +1580,7 @@ def load_metadata_details(request, c_type, c_id, conn=None, share_id=None,
     return context
 
 
-@login_required()
+@login_required(setGroupContext=True)
 @render_response()
 def load_metadata_preview(request, c_type, c_id, conn=None, share_id=None,
                           **kwargs):
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -131,7 +131,7 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
 
         super(OmeroWebGateway, self).__init__(*args, **kwargs)
         self._shareId = None
-    
+
     def getObject(self, *args, **kwargs):
         """
         We override this to handle cases where the Object is not found
--- a/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/webclient_gateway.py
@@ -144,12 +144,12 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
         if obj is None:
             self.SERVICE_OPTS.setOmeroGroup('-1')
             obj = super(OmeroWebGateway, self).getObject(*args, **kwargs)
-        if obj is not None:
             # set the group context for subsequent queries
-            group = obj.getDetails().group
-            if group is not None:
-                gid = group.id.val
-                self.SERVICE_OPTS.setOmeroGroup(gid)
+            if obj is not None:
+                group = obj.getDetails().group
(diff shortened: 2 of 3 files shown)


chris-allan:
I'm quite skeptical about this approach. It feels very error prone and subject to a lot of unintended consequences.

If there legitimately places where `setGroupContext=True`, and consequently making queries only across a single group, is significantly more performant it really feels to me like we would be better off spending our time being more explicit and efficient about the queries themselves. That will improve the situation for everyone.

On the Trello card you talk about queries with/without specific group contexts having substantially different performance. Is that because what is being returned is substantially different? Is it the query itself that takes more time?

will-moore:
@chris-allan The exact same query, returning the exact same thing is much slower if there are many users in the group and the context is -1. Therefore on the demo server we now limit the size of the private default group, which improves performance.
It's not what I expected, when you're doing a small query, e.g. get an Image by ID, but that seems to be the case - see https://trello.com/c/vv4mqsAi/5752-omero-omeroweb-on-demo-is-slow#comment-5d1c7e3f5a16666d9739d592

chris-allan:
Saw the query on the Trello card. I assume you're talking about:

```
select obj from Image obj
    join fetch obj.details.owner as owner
    join fetch obj.details.creationEvent
       where obj.id in (:ids)
```

???

Are you sure that the entire `image.details.owner`, `image.details.group`, object hierarchies are the same between the two styles? What happens if the events aren't loaded? What happens if the owner isn't loaded?

chris-allan:
Just did a quick test with `{'omero.group': '-1'}` vs. `{'omero.group': '<group_id>'}` on my local test system with a Project. When under `-1` `details.group.linkedExperimenterList()` is available due to the experimenter group map being filled out. I expect this is where your performance difference is but you'll have to validate that for yourself in your environment.

will-moore:
Wow, you're right.
Simply by setting group context ```-1``` this means that ```image.details.group.copyGroupExperimenterMap()``` lists all members of the group (even when you're loading your own image in a private group. But they are not loaded if you set the group context to the same group as the image. Is this expected behaviour?

We don't explicitly load the ```details.group``` for the object. Is there a way we can ensure that the ```groupExperimenterMap``` is not loaded?

cc @joshmoore

chris-allan:
Just a little further information: `details.owner` is also loaded when under `-1` conditions whether you ask for it or not.

joshmoore:
I won't have much time to look at this before leaving, but in general, I assume we should dig into https://github.com/ome/omero-server/issues/62 first before making too many client-side changes.

will-moore:
Agreed, better to fix this server-side...

PR #6081, opened 2019-07-04
Webclient performance fixes

will-moore:
Performance improvements for loading the webclient home page.

E.g. see https://trello.com/c/vv4mqsAi/5752-omero-omeroweb-on-demo-is-slow#comment-5bec2f993576763746ab1f21

We now load only the Groups and Experimenters needed for the page.
In particular, we don't load all the Experimenters in the current group (unless needed when loading the search page).

Added parameters for ```conn.getObject('ExperimenterGroup', id, opts={'load_experimenters':False})``` to avoid loading experimenters (and the same for ```getObject('Experimenter')```.

To test:
 - Groups and Experimenter functionality of webclient main page should not be broken.
     - Switch User - webclient displays their data as expected
     - Switch to All Members.
     - Switch Group etc.
     - webclient/?show=image-123 should work when image is not in current user/group
 - Should be much faster in Groups with large numbers of Experimenters.
 - Search page should still show Groups and Experimenters in form choosers.
 - Tests should be passing.

To test performance on a server with a large number of users, connect to the 'learning' server from https://merge-ci.openmicroscopy.org/web/webclient/login/

### DIFF
--- a/components/tools/OmeroPy/src/omero/gateway/__init__.py
+++ b/components/tools/OmeroPy/src/omero/gateway/__init__.py
@@ -6111,16 +6111,20 @@ class _ExperimenterGroupWrapper (BlitzObjectWrapper):
     def _getQueryString(cls, opts=None):
         """
         Returns string for building queries, loading Experimenters for each
-        group.
+        group by default.
         Returns a tuple of (query, clauses, params).
+        Supported opts: 'load_experimenters': If True, we load experimenters
 
         :param opts:        Dictionary of optional parameters.
-                            NB: No options supported for this class.
         :return:            Tuple of string, list, ParametersI
         """
-        query = ("select distinct obj from ExperimenterGroup as obj "
-                 "left outer join fetch obj.groupExperimenterMap as map "
-                 "left outer join fetch map.child e")
+        load_experimenters = True
+        if opts is not None:
+            load_experimenters = opts.get('load_experimenters', True)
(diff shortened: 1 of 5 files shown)


joshmoore:
see https://github.com/openmicroscopy/openmicroscopy/pull/6077#issuecomment-508720883

rgozim:
Provided that functionality or expected behaviour hasn't been fundamentally changed by this, I would say this PR is okay (just being mindful of the discussion @joshmoore linked).

joshmoore:
I'd suggest a round of wider testing with/without public, with admins, etc. I'm pretty sure there are no impacts on other clients, but it might be good to get a confirmation of that.

cc: @pwalczysko

will-moore:
As discussed today, PRs that have changes to ```omero-py``` and ```omero-web``` first need to be ported to ```omero-py```. When that is merged, tagged and released, the changes to ```omero-web``` should include bumping the dependency to the corresponding ```omero-py``` release.

pwalczysko:
@joshmoore @will-moore : reading the two last comments of each of you subsequently, https://github.com/ome/openmicroscopy/pull/6081#issuecomment-513190159 and https://github.com/ome/openmicroscopy/pull/6081#issuecomment-533146785 I wonder at what point and on which server should I perform the wider tests suggested by Josh

will-moore:
@pwalczysko This PR has not been ported to ```omero-py``` yet, but I'll list it when it's ready. Thx

PR #6082, opened 2019-07-04
tables: call delete callback

joshmoore:
In testing the table closing logic for 5.5.1, I set up a tight loop to create, delete, and close a table. At 10000 servants it failed because the callbacks were not being cleaned up. This adds a try/finally block in order to do so. Two things to note:

 1. Making use of `client.submit()` is not possible since we only have a `ServiceFactoryPrx`
 2. I don't believe the nulling of the `file_obj` was appropriate since that's part of the close logic

### DIFF
--- a/components/tools/OmeroPy/src/omero/tables.py
+++ b/components/tools/OmeroPy/src/omero/tables.py
@@ -317,16 +317,16 @@ class TableI(omero.grid.Table, omero.util.SimpleServant):
             raise
 
         try:
-            callback.loop(20, 500)
-        except LockTimeout:
+            try:
+                callback.loop(20, 500)
+            except LockTimeout:
+                raise omero.InternalException(None, None, "delete timed-out")
+
+            rsp = callback.getResponse()
+            if isinstance(rsp, omero.cmd.ERR):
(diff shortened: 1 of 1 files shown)


mtbc:
Code change looks plausible to me.

joshmoore:
If this hasn't caused any issues in tests over the last while, I'd propose merging. In general, a review of all callbacks, and perhaps even warning on persisting callbacks on session close would be sensible in the integration tests.

joshmoore:
Looks like there are a number of leaks. If we get close to the next release, I suggest getting this in. (cc @jburel @sbesson) Otherwise, I will try to add __enter__ and __exit__ methods to the Python version of CmdCallbackI so that something like this will work:

```
with someMethod() as cb:
   cb.getResponse()
```

since the current uses of this form all leave a callback in the session:

```
@@ -339,6 +339,7 @@ def forgotten_password(request, **kwargs):
                     try:
                         conn._waitOnCmd(handle)
                     finally:
+                        # This likely leaks a callback servant
                         handle.close()
```

cc: @manics @will-moore

joshmoore:
transferred to: https://github.com/ome/omero-py/pull/152

PR #6083, opened 2019-07-12
web: omero.web.http_response_headers to set custom headers

manics:
Add property to set custom headers in Django OMERO.web response

# What this PR does

This allows custom headers to be added to the HTTP response sent to the client by OMERO.web.
These header(s) will only be added if they're not already present in the response.

# Testing this PR

Set one or more custom headers, e.g.

```
omero config append omero.web.http_response_headers '["X-custom-header", "my value"]'
omero web start
```
Check the headers sent by OMERO.web, e.g. `curl -I localhost:4080`. You should see the custom headers you configured.


manics:
Ported to https://github.com/ome/omero-web/pull/143

PR #6085, opened 2019-07-16
For omero.web.open_with's supported_objects use "images".

mtbc:
https://docs.openmicroscopy.org/omero/5.5.1/sysadmins/config.html#omero-web-open-with should maybe have a default of,
```
[[“Image viewer”,
  “webgateway”,
  {“supported_objects”: [“images”],
   ”script_url”: “webclient/javascript/ome.openwith_viewer.js”
}]]
```
or maybe the documentation is wrong?

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/settings.py
+++ b/components/tools/OmeroWeb/omeroweb/settings.py
@@ -609,7 +609,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
     # OPEN WITH
     "omero.web.open_with":
         ["OPEN_WITH",
-         ('[["Image viewer", "webgateway", {"supported_objects": ["image"],'
+         ('[["Image viewer", "webgateway", {"supported_objects": ["images"],'
           '"script_url": "webclient/javascript/ome.openwith_viewer.js"}]]'),
          json.loads,
          ("A list of viewers that can be used to display selected Images "


mtbc:
Now though I also notice https://github.com/ome/prod-playbooks/blob/5b9f1999/templates/omero-web-config-for-webapps.j2#L13

manics:
https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.1/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html#L85-L90

What happens if you have `"supported_objects": ["imageRaNdOmExTrAtExT"]`?

mtbc:
@pwalczysko explained the difference. Could possibly expand documentation someday.

will-moore:
The full documentation is at https://docs.openmicroscopy.org/omero/5.5.1/developers/Web/LinkingFromWebclient.html#open-with

mtbc:
Aha, interesting, thank you for the pointer! "Linking from Webclient" makes sense in retrospect but I been too slow to think of "open with" as being "linking". Now I read it all it explains everything. An awkward issue also was that although the source has,

```
Open with
---------
```
then
```
Label, name and supported objects
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
I was again deceived by https://trello.com/c/qrHbPlV4/111-firefox-font-scaling-issue into not seeing that the latter is within the former.

PR #6091, opened 2019-07-29
CmdError: implement __str__

joshmoore:
Provide a better representation of the CmdError instance
even without unwrapping the `err` instance.

# Testing this PR

```
$ bin/omero shell --login
import omero
import omero.all
fp = omero.cmd.FindChildren()
h = client.submit(fp)
# Manually check h.getResponse()
print(str(omero.CmdError(h.getResponse())))
```


joshmoore:
Migrated to https://github.com/ome/omero-py/pull/151

PR #6093, opened 2019-07-31
admin diagnostics: show BF (and optionally other) jar versions

manics:
# What this PR does

Displays some jar manifest information for the Bio-Formats jars.

https://github.com/IDR/deployment/pull/168 replaces the bundled BF jars with a custom version. Since development versions may be used it'd be useful to check the versions if problems occur.

# Testing this PR

```
$ bin/omero admin diagnostics

...

Jar:        lib/client/formats-api.jar     Bio-Formats API      6.1.1   7 June 2019
Jar:        lib/client/formats-bsd.jar     BSD Bio-Formats readers and writers 6.1.1    7 June 2019
Jar:        lib/client/formats-gpl.jar     Bio-Formats library  6.1.1   7 June 2019
Jar:        lib/server/formats-api.jar     Bio-Formats API      6.1.1   7 June 2019
Jar:        lib/server/formats-bsd.jar     BSD Bio-Formats readers and writers 6.1.1    7 June 2019
Jar:        lib/server/formats-gpl.jar     Bio-Formats library  6.1.1   7 June 2019
```

```
$ bin/omero admin diagnostics --all-jars
```
Recursively find and show info for all `*.jar`

### DIFF
--- a/components/tools/OmeroPy/src/omero/cli.py
+++ b/components/tools/OmeroPy/src/omero/cli.py
@@ -1042,6 +1042,7 @@ class DiagnosticsControl(BaseControl):
         diagnostics.add_argument(
             "--no-logs", action="store_true",
             help="Skip log parsing")
+        return diagnostics
 
     def _diagnostics_banner(self, control_name):
 
--- a/components/tools/OmeroPy/src/omero/plugins/admin.py
+++ b/components/tools/OmeroPy/src/omero/plugins/admin.py
@@ -1511,6 +1511,7 @@ present, the user will enter a console""")
             'Implementation-Title',
             'Implementation-Version',
             'Implementation-Date',
+            'Implementation-Build',
         )
         self.ctx.out("")
         for jar in jars:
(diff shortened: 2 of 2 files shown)


sbesson:
Overall 💯 for adding extra utilities allowing to introspect the composition of a server. See https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-server/109/console for an example of output

A few comments related to the implementation:

- the `admin diagnostics` is probably as good as any to include this reporting. This code can probably be refactored if the logic is used in another command
- there will be a lot of duplication between `lib/server` and `lib/client`. Within the context of the `admin` command, I would propose only to list the content of `lib/server`. For people who want to know the client version, `bin/omero import -f` will return some information
- If only the server JARs are liste, maybe `Server JAR:\tformats-api.jar\t...`
- the other key/value I tend to use occasionally from the MANIFEST is `Implementation-Build` which contains the build SHA1 especially in development mode while the build date is usually less relevant

manics:
I've changed the default behaviour to only show `lib/server/formats-*.jar`. The `--all-jars` argument finds *all* jars under `OMERO.server`, it doesn't limit itself to `lib/`. Since this is for advanced diagnostic use and isn't the default I don't think it should be restricted

manics:
Ported to https://github.com/ome/omero-py/pull/188

PR #6096, opened 2019-08-02
Script gui file upload

will-moore:
See https://github.com/ome/openmicroscopy/issues/6095

Currently, some scripts that require a File use the convention of a parameter named ```File_Annotation``` to allow the user to choose a previously created ```FileAnnotation ID```.

This PR adds a File chooser input to parameter named ```File_Annotation``` which allows users to pick a local file from the Run-Script dialog in webclient. This will be used to create a File Annotation in OMERO and the ID will be passed to the ```File_Annotation``` script parameter.

To test:
 - Choose a script that has a ```File_Annotation``` parameter, such as "Populate Metadata" (https://github.com/ome/scripts/blob/develop/omero/import_scripts/Populate_Metadata.py)
 - The script dialog will allow you to choose a file:

Screen Shot 2019-08-02 at 23 53 57

 - This will allow you to pick a local File to upload
 - When the script is run, a new ```FileAnnotation``` is created in OMERO (not linked to anything) and the ID is passed to the ```File_Annotation``` script parameter (so the script should work as expected)

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/scripts/include_param.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/scripts/include_param.html
@@ -90,6 +90,12 @@
 
 {% if i.min %}<span style="color: #666"> Min: {{ i.min }} </span>{% endif %}
 {% if i.max %}<span style="color: #666"> Max: {{ i.max }} </span>{% endif %}
+
+<!-- If "File Annotation", allow user to choose file to upload, to create a File-Annotation -->
+{% ifequal i.name 'File Annotation' %}
+    OR <input name='file_annotation' type='file'/>
+{% endifequal %}
+
 </td></tr>
 </table>
 </div>
--- a/components/tools/OmeroWeb/omeroweb/webclient/views.py
+++ b/components/tools/OmeroWeb/omeroweb/webclient/views.py
@@ -4247,10 +4247,13 @@ def script_run(request, scriptId, conn=None, **kwargs):
     if fileupload is not None and fileupload != "":
         manager = BaseContainer(conn)
         fileAnnId = manager.createFileAnnotations(fileupload, [])
-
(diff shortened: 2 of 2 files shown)


jburel:
The upload is not specifically for file annotation. 
The script they are writing is to import an image
so we will need a new type I think

will-moore:
This would work for an image file too.
The only limitation is that this supports just 1 file per script (since you can only have 1 ```File_Annotation``` parameter).
However, it is still a big improvement to the usability of scripts with a ```File Annotation``` parameter. Previously, users had to know to create a File Annotation before-hand and to know how to pick it. Now, users can use this with no other instructions or preparation.

jburel:
I mean an image to import not an image as an attachment
I can image the script in that case uploading the file to dropbox 
this will only work for single file image

will-moore:
OK, so after reading https://forum.image.sc/t/file-browser-in-omero-script-gui/28117 more carefully I see what's wanted is a UI for browsing a drive mounted on the server.
But regardless, this PR is useful for the workflow described above.

rgozim:
Tested and works fine.

jburel:
* Select populate metadata 
* Choose a file. I deliberately selected a non CSV file. 
* Click Run script. The button is greyed out but nothing happens. The dialog does not close. Nothing in the activity. 
Feedback should be given if the type is not supported. Ideally the dialog used to select the file should only allow to upload the supported type if specified.

will-moore:
@jburel We have no way for the script to tell the UI script dialog about which files are supported.
This PR does not change the handling of files but is simply a short-cut to populating the existing ```File_Annotation``` input (which simply accepts an ID).
I don't know what the issue you had with the script dialog, but it wasn't caused by you picking a non CSV file since the file would not have been opened until the script started running on the server.

will-moore:
Ported to https://github.com/ome/omero-web/pull/128

PR #6099, opened 2019-08-08
Use chosen plugin for script UI if large number of options

will-moore:
# What this PR does

See https://forum.image.sc/t/file-browser-in-omero-script-gui/28117/16

If we have many options in a script UI field with ```defaults``` set then it becomes tricky to use.
This PR uses the chosen plugin (which we use elsewhere) to allow users to type to filter the list of options. We only use this if the number of options is large (propose to use limit of 20).

NB: The limit is currently set to 2 to allow testing (since we don't have any scripts with > 20 inputs
After testing, this can be set to e.g. 20?

To test:
 - Choose a script with a bunch of parameters with default options, e.g Batch_Image_Export.py
 - Currently, any parameters with > 2 options will use the chosen plugin, allowing you to type to select from the list
 - Others (e.g. Data_Type) will still use the regular select

### DIFF
--- a/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/scripts/script_ui.html
+++ b/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/scripts/script_ui.html
@@ -35,6 +35,8 @@
 
 {% block link %}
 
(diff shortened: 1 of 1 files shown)


will-moore:
Ported to https://github.com/ome/omero-web/pull/135

PR #6105, opened 2019-09-19
Enable use of decoupled omero-py

### LINKED ISSUE #6103: Decoupled Python servers should be used if omero-py from pypi is installed
OMERO.server Python servers are run using a fixed path:
https://github.com/ome/openmicroscopy/blob/a6977678df03046ea95a9256d33bac85f26f7431/etc/templates/grid/templates.xml#L352
https://github.com/ome/openmicroscopy/blob/a6977678df03046ea95a9256d33bac85f26f7431/etc/templates/grid/default.xml#L19

This means if a decoupled omero-py is installed the starting scripts from that package aren't used. If `OMERO.server/lib/python` is deleted the servers won't be run at all.

joshmoore:
fix #6103 

# What this PR does

In OMERO 5.5.1, the grid/templates.xml ran Python subprocesses via:

```
python lib/python/X.py
```

In order to also support a file on the PYTHONPATH, now use:

```
python -m X
```

# Testing this PR

1. integration tests should all pass
2. diagnostics should show Processor and Tables active

### DIFF
--- a/etc/templates/grid/default.xml
+++ b/etc/templates/grid/default.xml
@@ -18,7 +18,7 @@
 
     <variable name="OMEROPY_HOME"  value="lib/python/"/>
     <variable name="OMEROPY_SERV"  value="lib/python/omero/"/>
-    <variable name="PYTHONPATH"    value="PYTHONPATH=${OMEROPY_HOME}:$${PYTHONPATH}"/>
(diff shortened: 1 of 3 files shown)


joshmoore:
This has been in the build for quite some time. If someone is maintaining their own `templates.xml` they will have issues, but I assume that to be a minimal impact in comparison to everything else.

Merging.

joshmoore:
@ome/python -- note: this will likely only work where `lib/python` has been deleted. The setting of PYTHONPATH in default.xml takes precedence.

PR #6107, opened 2019-09-23
Use reverse(name) for webgateway URLs in integration tests

will-moore:
# What this PR does

This, along with https://github.com/ome/omero-web/pull/10/commits/b16716165c1c7dc31482e8e9f2a0f0e9d4758418 
should fix failing webgateway integration tests at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/154

I think this illustrates a breaking change in OMERO.web ```5.5.1dev``` which we first saw in the ```unit_tests``` at https://github.com/ome/omero-web/pull/10#issuecomment-527169895 and required https://github.com/ome/omero-web/pull/10/commits/2fff52d8dabfb0087493a1f2dc7068594912d1c4 to fix unit tests.

For other omero.web apps (figure, iviewer etc and 3rd-party apps), if they previously did this in views.py
```
reverse('webgateway.views.render_roi_thumbnail', args=['0'] )
```
in 5.5.1, this no-longer works with decoupled OMERO.web.
You now need to use the new 'name':
```
reverse('webgateway_render_roi_thumbnail', args=['0'] )
```
But in the Django html template, this still works with decoupled OMERO.web
```
{% url 'webgateway.views.render_roi_thumbnail' 0 %}
```

We don't use
```
reverse('webgateway.views.X')
```
in OMERO.figure or OMERO.iviewer. So they don't break with decoupled OMERO.web.
They only use
```
reverse('webgateway')
```
for the named webgateway base url, then generate other URLs via JavaScript, not python.

### DIFF
--- a/components/tools/OmeroWeb/test/integration/test_csrf.py
+++ b/components/tools/OmeroWeb/test/integration/test_csrf.py
@@ -194,7 +194,7 @@ class TestCsrf(IWebTest):
         session['fromid'] = img.id.val
         session.save()
 
-        request_url = reverse('webgateway.views.copy_image_rdef_json')
+        request_url = reverse('copy_image_rdef_json')
         data = {
             'toids': img.id.val
         }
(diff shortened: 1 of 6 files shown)


will-moore:
This has fixed 28 integration tests: https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/155/

joshmoore:
Thanks, @will-moore

PR #6108, opened 2019-10-02
Don't use wrap(id) for long IDs

will-moore:
# What this PR does

Fixes ```ClientError: Cannot handle conversion from: <type 'long'>``` in rtypes for ```TestGetObject.testOrderById()```.


joshmoore:
Green. Re-launching merge-ci with this in.

will-moore:
Don't know why travis is failing now on the same code that was previously passing (last commit above reverts the previous one).
The exclusion of this PR is causing a couple more tests to fail at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/175/

joshmoore:
Merging proactively so that everyone is working from the same set of commits.

PR #6111, opened 2019-10-18
use mox3

jburel:
# What this PR does

Remove usage of Mox. This has been removed from omero-py see https://github.com/ome/omero-py/pull/46
The class is deprecated but we have not decided yet if it will be removed


joshmoore:
Merging along with https://github.com/ome/omero-py/pull/46#issuecomment-545368367

PR #6114, opened 2019-10-21
Backwards compatibility test for OMERO.server tables 5.3.4

manics:
Adds a table created under OMERO 5.3.4 that can be used to test the Python 3 and subsequent work is compatible with tables created under an older server.

`service-reference-dev_5_3_4.h5.bz2 ` was created using the `openmicroscopy/omero-server:5.3.4` Docker image ([the oldest we've got on Docker Hub](https://hub.docker.com/layers/openmicroscopy/omero-server/5.3.4/images/sha256-d7b172bfbf3c7f7c03c530488ebf13c95c4db5bf404af112d533e708589f0e4f)):
```
$ python -c 'import tables; print tables.__version__'
3.2.0rc3
```

The original PR which added the metadata methods is older than 5.3.4: https://github.com/ome/openmicroscopy/pull/3249

The main thing to note is that the 5.3.4 Docker image treats both `int` and `long` in the metadata map as `RLong`, no idea whether this is expected or not.


joshmoore:
Code changes generally look fine. Considering the code tests are currently failing, I'm not exactly sure what to do about merging. Having had this in `latest-ci` _before_ all the changes would likely have been the sounded schema, no?

manics:
I've tested it against a 5.3.4 server.

Ideally it would've gone in with the original PR, something to remember for the next time OMERO.tables changes?

joshmoore:
Holding off on merging for https://github.com/ome/omero-py/pull/59

manics:
I don't think it needs to wait for ome/omero-py#59 since it's documenting existing behaviour

joshmoore:
It's more a matter of adding a failing test. If we're going to have a specific tables-focused tag coming up, then it feels like it's most bisect-able to have them together.

joshmoore:
Is this expected?

```
E       AssertionError: assert 'Invalid vers.../Dir-002/2345' == 'Tables metad...O.tables >= 2'
E         - Invalid version attribute (version=b'v1') in path: /home/omero/workspace/OMERO-test-integration/data/Files/Dir-002/2345
E         + Tables metadata is only supported for OMERO.tables >= 2

test/integration/tablestest/test_backwards_compatibility.py:329: AssertionError
```

https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/41/testReport/junit/OmeroPy.test.integration.tablestest.test_backwards_compatibility/TestBackwardsCompatibility_4_4_5/testMetadataException/

manics:
That looks like a byte/string conversion problem in the handling of the tables version string ...
https://github.com/ome/omero-py/blob/91a94ac1440cb922f605080526579cda78e77e63/src/omero/hdfstorageV2.py#L294-L313


In the absence of any better ideas how about add your
```
            if isbytes(t):
                t = bytes_to_native_str(t)
```
from https://github.com/ome/omero-py/pull/104 to https://github.com/ome/omero-py/pull/104/files#diff-e693b902f9c719a94b6bd0f71ec29e1aL302 ?

joshmoore:
Merging now that https://github.com/ome/omero-py/pull/99 (successor to 59) is in.

PR #6122, opened 2019-10-24
futurize all remaining tests (OmeroWeb)

joshmoore:
Like https://github.com/ome/openmicroscopy/pull/6120, but the futurize output for all tests under OmeroWeb/ which are not already modified on snoopy's merge-ci branch.


joshmoore:
Tests now running on py3 :+1:

PR #6125, opened 2019-10-30
timeline test

jburel:
Rewrite test so it does what i think it was supposed to do

https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/27/testReport/OmeroPy.test.integration.test_itimeline/TestITimeline/test1173/

Check that the test is green


joshmoore:
Any idea why it stopped passing?

jburel:
the only thing I could think of was a change in testlib but there is nothing obvious there
I had the test failing in both py2 and py3

joshmoore:
py2 and py3 still passing. And I _still_ can't understand what the test was previously trying to accomplish. _shrug_

PR #6126, opened 2019-10-30
fix clitest/test_download

will-moore:
# What this PR does

Fixes a bunch of tests like:
https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/27/testReport/OmeroPy.test.integration.clitest.test_download/TestDownload/testNonExistingOriginalFile__/


joshmoore:
Test passed in merge (217) and py3 (31). The solution seems fine but I imagine along with the discussion in https://github.com/ome/omero-py/pull/69#issuecomment-548302232 we could decide to be lenient in what we accept even if we're stricter in what we return  (cf. [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). Since these changes are all in a test it's not overly important. Merging.

PR #6143, opened 2019-10-31
Unpin pytest<3

joshmoore:
Pytest dropped support for Python 2 syntax leading us
to pin the version. Since the syntax has been updated,
this can now be unpinned.

### DIFF
--- a/components/tools/OmeroFS/setup.py
+++ b/components/tools/OmeroFS/setup.py
@@ -35,4 +35,4 @@ setup(name="OmeroFS",
       package_dir={"": "target"},
       packages=[''],
       cmdclass={'test': PyTest},
-      tests_require=['pytest<3'])
+      tests_require=['pytest'])
--- a/components/tools/OmeroFS/setup.py
+++ b/components/tools/OmeroFS/setup.py
@@ -35,4 +35,4 @@ setup(name="OmeroFS",
       package_dir={"": "target"},
       packages=[''],
       cmdclass={'test': PyTest},
-      tests_require=['pytest'])
+      tests_require=['pytest', 'pytest-xdist'])
--- a/components/tools/OmeroPy/setup.py
+++ b/components/tools/OmeroPy/setup.py
@@ -69,4 +69,4 @@ setup(
         'omero.gateway': ['pilfonts/*'],
         'omero.gateway.scripts': ['imgs/*']},
     cmdclass={'test': PyTest},
-    tests_require=['pytest'])
+    tests_require=['pytest', 'pytest-xdist'])
--- a/components/tools/OmeroPy/setup.py
+++ b/components/tools/OmeroPy/setup.py
@@ -69,4 +69,4 @@ setup(
         'omero.gateway': ['pilfonts/*'],
         'omero.gateway.scripts': ['imgs/*']},
     cmdclass={'test': PyTest},
-    tests_require=['pytest<3'])
+    tests_require=['pytest'])
--- a/components/tools/OmeroPy/test/conftest.py
+++ b/components/tools/OmeroPy/test/conftest.py
@@ -9,6 +9,7 @@
 from builtins import range
 from builtins import object
 import pytest
+import os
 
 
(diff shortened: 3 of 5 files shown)


sbesson:
The state of the Py2/Py3 integration tests with this PR included seem to indicate uncapping pytest does not create regression. Some tests were affected by dropped functionalities but the fix are relatively easy - see https://github.com/ome/openmicroscopy/pull/6151 - and uncapping this allows to start enabling parallel tests at least locally.

The primary issue I have been encountering is an environment setup one as I was expecting `setup.py test` to auto-install the appropriate dependency. However on Python 2, it looks like this pull `pytest 5.2.2` which is Python 3 only.

Rather than trying to fix `setup.py`, I will look into whether a `tox` environment similar to https://github.com/ome/omero-py/blob/master/tox.ini can be easily set up for these integration tests.

joshmoore:
Looks like even the newest setuptools can't handle the python_version < 3` syntax:

```
(py27) /opt/ome9/components/tools/OmeroPy $which pip
/opt/anaconda/envs/py27/bin/pip
(py27) /opt/ome9/components/tools/OmeroPy $pip install -U setuptools
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Collecting setuptools
  Downloading https://files.pythonhosted.org/packages/d9/de/554b6310ac87c5b921bc45634b07b11394fe63bc4cb5176f5240addf18ab/setuptools-41.6.0-py2.py3-none-any.whl (582kB)
     |████████████████████████████████| 583kB 1.2MB/s
Installing collected packages: setuptools
  Found existing installation: setuptools 41.4.0
    Uninstalling setuptools-41.4.0:
      Successfully uninstalled setuptools-41.4.0
Successfully installed setuptools-41.6.0
(py27) /opt/ome9/components/tools/OmeroPy $./setup.py test test/x.py
ez_setup.py is deprecated and when using it setuptools will be pinned to 33.1.1 since it's the last version that supports setuptools self upgrade/installation, check https://github.com/pypa/setuptools/issues/581 for more info; use pip to install setuptools
WARNING: '' not a valid package name; please use only .-separated package names in setup.py
error in omero_client setup command: 'tests_require' must be a string or list of strings containing valid project/version requirement specifiers; Invalid requirement, parse error at "'if pytho'"
(py27) /opt/ome9/components/tools/OmeroPy $git diff setup.py
diff --git a/components/tools/OmeroPy/setup.py b/components/tools/OmeroPy/setup.py
index 0a1d51c63d..f7e0b82a92 100755
--- a/components/tools/OmeroPy/setup.py
+++ b/components/tools/OmeroPy/setup.py
@@ -69,4 +69,7 @@ setup(
         'omero.gateway': ['pilfonts/*'],
         'omero.gateway.scripts': ['imgs/*']},
     cmdclass={'test': PyTest},
-    tests_require=['pytest', 'pytest-xdist'])
+    tests_require=[
+        'pytest < 5 if python_version < 3',
+        'pytest',
+        'pytest-xdist'])
```

Only other option would be to use `if sys.version_info < (3, 0, 0)` but that may cause other issues.

joshmoore:
Note: one downside of the conftest.py change is that xdist _must_ be installed.

sbesson (approved):
This has been successfully included in the integration tests for the last weeks. Additionally, enabling `pytest-xdist` allowed us to investigate parallelization of tests, at least for a subset of them since the current implementations have a lot of single-threaded assumptions.

As noted above, the main downside is https://github.com/ome/openmicroscopy/pull/6143#issuecomment-548831450 i.e. related to the setup. A review of https://docs.openmicroscopy.org/omero/5.5.1/developers/testing.html#running-tests-directly should probably be done further down the line as we migrate towards a release. Additionally, this only affects the soon-not-recommended Python 2 environment.

LGTM

PR #6145, opened 2019-10-31
gatewaytest: test helper

jburel:
handle py2 and py3
``omero_type`` is not really helping anymore for long/int case

Adjust the tests to check both cases py2 and py3 
This should fix
*  https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/29/testReport/OmeroPy.test.integration.gatewaytest.test_helpers/TestHelperObjects/testOmeroType/

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_helpers.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_helpers.py
@@ -11,7 +11,7 @@ from builtins import object
 import omero
 import omero.gateway
 try:
-    int
+    long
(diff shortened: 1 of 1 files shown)


joshmoore:
Test passed on merge (217) and py3 (31). Merging. I assume that when we go back and clean up anything matching `sys.version_info` or `python\s(2|3)` we can just drop the redundant assert. I'm not sure if something should be added to `omero_type` to permit long-y behavior. Perhaps it's just a documentation issue. cc: @will-moore

PR #6149, opened 2019-11-01
cli test-metadata

jburel:
Use long instead of int

This should fix tests like https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/217/testReport/OmeroPy.test.integration.clitest.test_metadata/TestMetadata/test_get_identifiers/
Tests should remain green on both py2 and py3


sbesson:
22 tests fixed and passing on merge (218) and py3 (32).

PR #6169, opened 2019-11-06
Bump omero-py, omero-web and omero-marshal versions

sbesson:
These should include the latest changes for Py3 compatibility


joshmoore (approved):
LGTM and only a (very) few unrelated test failures today. :+1:

PR #6172, opened 2019-11-07
Add OmeroWeb test_create_tag()

will-moore:
Adding new test as suggested at https://github.com/ome/omero-web/pull/50#issuecomment-550978557

This test fails if I revert the fixes in that PR.


joshmoore:
Merging along with omero-web#50

PR #6173, opened 2019-11-08
Check html activities output from running script

will-moore:
Adding test to check Activities html for script to catch bug at https://github.com/ome/omero-web/pull/61 that was found in manual testing.

Work in progress (having trouble running scripts locally)...


components/tools/OmeroWeb/test/integration/test_scripts.py
```diff
@@ -49,11 +49,13 @@ def upload_script(self):
 
         script = """
 import omero
-from omero.rtypes import rstring, rint, wrap
+from omero.rtypes import rstring, rint, wrap, robject
```
jburel: maybe time to order alphabetically the import

joshmoore: adding

```
[flake8]
import-order-style = smarkets
```

to `.flake8` would be one method to enforce the imports. see https://pypi.org/project/flake8-import-order/ for more info.

will-moore: I'm not sure that flake8 would catch these imports since they are in a script string.

will-moore:
This test now failing as expected with the fix https://github.com/ome/omero-web/pull/61 excluded due to travis

https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/53/testReport/OmeroWeb.test.integration.test_scripts/TestScripts/test_script_inputs_outputs_inputs0_/

PR #6175, opened 2019-11-12
Bump scripts to 5.6.0-m1

joshmoore:
Propose to merge when green and tag as 5.6.0-m1.


joshmoore:
This also needs a CHANGELOG entry for potentially adding to the upcoming image.sc post.

PR #6176, opened 2019-11-12
Restore CLI DB integration tests

sbesson:
As part of the omero-py decoupling, the tests for the `db` CLI plugin were skipped as they require extra infrastructure including the SQL scripts and the JAR.

This PR migrates these tests as integration tests of the `ome/openmicroscopy` repository so that they can be executed using the `OMERO.server` bundle with `OMERODIR` set.

Some of the  test logic dropped since the OMERO 5.5 refactoring work is removed and the `db password` tests are marked as `xfail` as per https://github.com/ome/omero-py/issues/112. I will open a separate PR to reactivate them and remove the markers.


joshmoore:
Excluded temporarily to let #6201 in.

jburel:
This PR has been in the build for a long time.
Proposing to include it
Any objections?

joshmoore:
None from me.

PR #6181, opened 2019-11-20
Allow to specify prefix for scripts archive

sbesson:
This is a prerequisite to be able to consume script integration branches from GitHub.

Unlike the other Python dependencies, the OMERO scripts are currently unpacked from GitHub source archives under `lib`. In order to consume archives from integration branches using a custom `versions.omero-github` location, it is also necessary to override the `v` prefix used for release tags.

Alongside https://github.com/ome/devspace/pull/157m this feature should be primarily be tested in the context of the merge devspaces, consuming the https://github.com/snoopycrimecop/scripts/tree/py3_ci and/or https://github.com/snoopycrimecop/scripts/tree/merge_ci integration branches and deploying servers including open script PRs.

### DIFF
--- a/components/tools/python.xml
+++ b/components/tools/python.xml
@@ -89,9 +89,9 @@ build_year = "${python_build_year}"
         <delete dir="../target/downloads/py/omero-py-${versions.omero-py}"/>
 
 
-        <download pkg="scripts" file="v${versions.scripts}.tar.gz" expected="${versions.scripts-md5}"
+        <download pkg="scripts" file="${versions.scripts-prefix}${versions.scripts}.tar.gz" expected="${versions.scripts-md5}"
                   where="../target/downloads/scripts"/>
-        <untar src="../target/downloads/scripts/v${versions.scripts}.tar.gz" dest="../target/lib/scripts" compression="gzip">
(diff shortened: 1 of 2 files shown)


joshmoore:
I remember struggling with this same thing. Not optimal that there needs to be an entire variable to define a single letter, but I didn't have a better suggestion either. :+1: Hoping that we can remove this soon anyway.

sbesson:
Having briefly investigated going beyond GitHub, I think we might be relatively close to consuming `omero-scripts` source distributions  (created via `sdist`) as opposed to repo archives. This would at least unify the behavior with the other core OMERO Python packages.

Roughly I think the steps might be: 

- fix the `scripts` `setup.py` to actually package the scripts as opposed to creating a shallow source distribution as of now
- update `omero.properties` and `python.xml` to unpack from a source distribution like `omero-py` etc 
- test consuming the [Jenkins artifacts](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-build/) as well as a pre-release pushed to PyPI
- rename the `scripts` repository as `omero-scripts` to match the package name might be a requirement or at least recommended

Note this would only address the problem of packaging scripts within an OMERO.server bundle (i.e. phase 1). Consuming fully pip-installed scripts would require additional thoughts.

PR #6184, opened 2019-11-21
Update Dockerfile for python3

sbesson:
Travis CI has been failing since `omero-test-infra` was updated to Python 3 - see https://github.com/ome/omero-test-infra/pull/44.

This commit should ensure the 5.6 Docker image are used.


sbesson:
Green. Merging and I will go through open PRs to get the fix in.

PR #6188, opened 2019-11-27
examples

jburel:
Adjusts the remaining python scripts under examples
We can check what we want to migrate/rescue from the examples folder later on

### DIFF
--- a/examples/OmeroClients/collectionmethods.py
+++ b/examples/OmeroClients/collectionmethods.py
@@ -8,14 +8,14 @@ DatasetI = omero.model.DatasetI
 EventI = omero.model.EventI
 PixelsI = omero.model.PixelsI
 
-image = ImageI(long(1), True)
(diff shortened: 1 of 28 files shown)


examples/OmeroClients/collectionmethods.py
```diff
@@ -8,14 +8,14 @@
 EventI = omero.model.EventI
 PixelsI = omero.model.PixelsI
 
-image = ImageI(long(1), True)
-image.getDetails().setUpdateEvent(EventI(1L, False))
+image = ImageI(int(1), True)
```
joshmoore: No objection, but the `int` here is almost certainly no longer needed.

joshmoore (approved):
Changes all look fine. :+1:

examples/OmeroClients/collectionmethods.py
jburel: removed

jburel:
Changes approved. Merging

PR #6190, opened 2019-11-28
Consume omero-scripts source distributions

sbesson:
This updates the Python downloading logic to consume source distributions of the `omero-scripts` component from PyPI like `omero-py`, `omero-web` rather downloading the GitHub archive previously.


joshmoore:
Scripts look fine today on py3-ci. :+1:

PR #6193, opened 2019-12-04
Bump python deps for 5.6.0-m4

joshmoore:
Bump to the latest releases of omero-py and omero-web. Travis status should suffice.


joshmoore:
Merging based on the greenness of travis and the daily builds.

PR #6189, opened 2019-11-28
Update testAllColumnsSameTable to test Unicode with StringColumn

sbesson:
While working on the support for  CSV files with Unicode in `omero-metadata`, I encountered an issue with the OMERO 5.6 server on Python 3.
This PR updates the default StringColumn integration test to test strings with unicode characters in it. It should initially pass on Python 2 but fail on Python 3.

PR #6194, opened 2019-12-05
Bump omero-py to 5.6.dev9

sbesson:
See https://github.com/ome/omero-py/blob/v5.6.dev9/CHANGELOG.md

This should contain the fix required for #6189


joshmoore:
Merging for tagging as 5.6.0-m4

PR #6200, opened 2020-01-06
Remove bin/omero

joshmoore:
Remove all references to `bin/omero`, incl.:

- bin from omero-py zip is not unpacked
- an exe of `omero` is expected everywhere
- all references (except history) are removed

Depends on:
 - [x] https://github.com/ome/omero-py/pull/157 so that testlib will use the correct `omero` executable
 - [x] https://github.com/ome/omero-server-docker/pull/40 to not use the `python OMERO.server/bin/omero` idiom


joshmoore:
General sign-off via slack. Re-ran deployment via https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/127/ (tests still running). Merging for rc-1. Follow-up commits will simplify build.xml to remove OMERO.py.zip and likely requirements.txt.

PR #6201, opened 2020-01-09
Cleanup build

joshmoore:
- [x] drop web requirements: https://github.com/ome/omero-web/issues/96
 - [x] drop OMERO.py zip
 - [x] misc cleanup


will-moore:
As far as I can understand these changes, they look good.

PR #6202, opened 2020-01-10
Clean up pypi version properties

joshmoore:
The omero-py, web, and dropbox versions are no longer used
in this repository. The scripts version *is* used and should
have been bumped to 5.6.0.


joshmoore:
Today's builds are green. General LGTM to readers.

Issue #6206, opened 2020-01-24
More non-parallel test assumptions


joshmoore:
Following on from https://github.com/ome/openmicroscopy/pull/6156, I tried running the integration tests as background jobs:

```
./build.py ... python &
./build.py ... java &
...
wait
```

but https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/287/#showFailuresLink failed with CLI assumptions:

```

Stacktrace

self = <test.integration.clitest.test_user.TestUser object at 0x7f52e4b75198>
capsys = <_pytest.capture.CaptureFixture object at 0x7f520acbbd68>
sort_key = None, group_format = None

    @pytest.mark.parametrize("sort_key", sort_keys)
    @pytest.mark.parametrize("group_format", [None, "count", "long"])
    def testList(self, capsys, sort_key, group_format):
        self.args += ["list"]
        if sort_key:
            self.args += ["--sort-by-%s" % sort_key]
        if group_format:
            self.args += ["--%s" % group_format]
        self.cli.invoke(self.args, strict=True)
    
        # Read from the stdout
        out, err = capsys.readouterr()
        ids = get_user_ids(out, sort_key=sort_key)
    
        # Check all users are listed
        if sort_key == 'login':
            sorted_list = sorted(self.users, key=lambda x: x.omeName.val)
        elif sort_key == 'first-name':
            sorted_list = sorted(self.users, key=lambda x: x.firstName.val)
        elif sort_key == 'last-name':
            sorted_list = sorted(self.users, key=lambda x: x.lastName.val)
        elif sort_key == 'email':
            sorted_list = sorted(self.users, key=lambda x: (
                x.email and x.email.val or ""))
        else:
            sorted_list = sorted(self.users, key=lambda x: x.id.val)
>       assert ids == [user.id.val for user in sorted_list]
E       assert [0, 1, 2, 3, 4, 5, ...] == [0, 1, 2, 3, 4, 5, ...]
E         Left contains 2 more items, first extra item: 3127
E         Full diff:
```

PR #6208, opened 2020-01-27
test_user: don't assume single-threaded (see #6206)

joshmoore:
`omero user list` was compared to the output of iadmin.lookupExperimenters().
Depending on the interval, other users could have been added. Now we assume
monotonically increasing users.

# Testing this PR

1. integration tests should remain green
2. additionally, parallel integration tests (pytest-xdist) should not fail on this test (others may exist)
3. the code should still reasonable test the assumptions of `omero user list`

### DIFF
--- a/components/tools/OmeroPy/test/integration/clitest/test_user.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_user.py
@@ -68,22 +68,35 @@ class TestUser(CLITest):
         self.cli.invoke(self.args, strict=True)
 
         # Read from the stdout
+        before = self.sf.getAdminService().lookupExperimenters()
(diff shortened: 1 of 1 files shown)


mtbc:
Is there a good tweak to `./build.py -f components/tools/OmeroPy/build.xml test -DTEST=test/integration/clitest/test_user.py` to make it go parallel?

joshmoore:
I'm less clear on getting a single test file to go parallel. However, either running the above N times, or:

```
pip install pytest-xdist
pytest -n 10 
test/integration/clitest/test_user.py
```

should do.

mtbc:
Working fine, indeed.

mtbc (approved):

PR #6209, opened 2020-01-29
Add space after period within  CLI test warning messages.

mtbc:
Simply fixes whitespace omission that easily slips into Python string concatenation.


jburel:
Thanks

PR #6212, opened 2020-02-11
add test to import images

jburel:
# What this PR does

Add tests for new method introduced in https://github.com/ome/omero-gateway-java/pull/30

cc @dominikl


jburel:
--exclude

PR #6216, opened 2020-02-18
Add gradle project support to OmeroJava

jburel:
# What this PR does

Give this PR a descriptive title then expand on that here.

This PR replaces https://github.com/ome/openmicroscopy/pull/6060

Running ``gradle test`` will now run the unit tests and the integration tests

Edit:

To run unit tests only run ``gradle test -Punit ``


components/tools/OmeroJava/settings.gradle
```diff
@@ -0,0 +1,9 @@
+pluginManagement {
+    repositories {
+        mavenLocal()
+        gradlePluginPortal()
+        maven { url "https://artifacts.openmicroscopy.org/artifactory/maven/" }
+    }
+}
+
+rootProject.name = "omero-java"
```
joshmoore: Might want to include `-test` somewhere in the name.

joshmoore (commented):
Any thoughts on the duplication with build.xml?

jburel:
for build.xml I was going to look at that afterwards if/when we were happy with first step

jburel:
to run the unit test only: 
run ``gradle test -Punit``

dominikl:
Looks good so far 👍

jburel:
I had a closer look while reviewing the docs. OmeroJava is the only remaining components with unit test. They should move to the gateway where they belong.
With the decoupling, it does not make sense to keep ``./build.py test-unit`` command either.

jburel:
I have moved the unit test to the gateway where it belongs see
https://github.com/ome/omero-gateway-java/pull/31
I will remove it in this PR and review ``build.gradle``

jburel:
Looking at the build.xml
We need to clean the top one, since it is also currently broken for OmeroFs, OmeroPy etc.
I will do that in another PR, for now I will add a no-op for the ``test`` target in OmeroJava

dominikl:
Looks good re OmeroJava, probably good to merge 👍

jburel:
Thanks Merging

PR #6219, opened 2020-02-21
Bump TestNG version to match other components of OMERO.

mtbc:
Seems odd to be using a different version of TestNG than most of its components. Can make testing with IDEs a bit of a pain.

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -188,7 +188,7 @@ versions.janino=2.5.10
 
 versions.reportng=1.1.1
 
-versions.testng=6.8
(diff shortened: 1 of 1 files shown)


jburel:
more legacy than anything else

jburel:
Thanks

jburel:
The upgrade seems to be responsible for the failure of the test https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/317/
I am looking into it

mtbc:
Interesting, thank you.

PR #6226, opened 2020-03-18
More Ivy configurability

chris-allan:
# What this PR does

In our push for more alignment with upstream this backports, in a modernised way, some of the Ivy hacks we have done over time to handle OMERO Plus builds.

There should be no changes to base functionality from this.

### DIFF
--- a/ivy.xml
+++ b/ivy.xml
@@ -13,5 +13,7 @@
     <dependency org="org.openmicroscopy" name="omero-gateway" rev="${versions.omero-gateway}"/>
     <!-- runtime dependencies from dsl/ivy.xml -->
     <dependency org="janino" name="janino" rev="${versions.janino}"/>
+    <!-- Useful for globally overriding the Bio-Formats version, empty version is ignored by default -->
+    <dependency org="ome" name="formats-gpl" rev="${versions.bioformats}"/>
   </dependencies>
 </ivy-module>
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,10 @@ addons:
     - git
 
 install:
-    - echo "ivy.retrieve.log=default" >> etc/local.properties
-    - echo "artifactory.host=artifacts.glencoesoftware.com" >> etc/local.properties
-    - echo "artifactory.baseurl=https://${artifactory.host}/artifactory" >> etc/local.properties
-    - echo "artifactory.repo=repo" >> etc/local.properties
+    - echo 'ivy.retrieve.log=default' >> etc/local.properties
+    - echo 'artifactory.host=artifacts.glencoesoftware.com' >> etc/local.properties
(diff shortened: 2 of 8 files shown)


chris-allan:
PR opened as per Slack conversation. We can discuss the specifics here based on a concrete proposal of changes.

/cc @sbesson, @joshmoore, @jburel

ivy.xml
```diff
@@ -13,5 +13,7 @@
     <dependency org="org.openmicroscopy" name="omero-gateway" rev="${versions.omero-gateway}"/>
     <!-- runtime dependencies from dsl/ivy.xml -->
     <dependency org="janino" name="janino" rev="${versions.janino}"/>
+    <!-- Useful for globally overriding the Bio-Formats version, empty version is ignored by default -->
```
joshmoore, line 16: Ah, good to know. Then I might investigate adding in omero-model etc. as well.

chris-allan: @joshmoore Yeah, limited testing so far but the assumption seems to hold as far as I can see. I don't think it's explicitly declared anywhere to be the case.

chris-allan:
Force pushing one commit at a time so I can try to find out what's causing the Ivy retrieve phase to hang.

Sorry for the noise @joshmoore

joshmoore:
No worries. Will be good to know.

chris-allan:
@sbesson / @joshmoore: That's Glencoe mirror up to date and build passing. Going to revert just to check and then I'll force push this again with all the original changes and tack the "Temporarily use Glencoe mirror" commit on top.

chris-allan:
Yep, confirmed, fails with artifacts.o.org. Going to force push now as mentioned previously and see how things go.

chris-allan:
Green, wooo! ~13 mins. Reverting the usage of the Glencoe mirror and seeing how that goes.

joshmoore:
@chris-allan: shall we try to get this in and see if we can stabilize travis? Or do you have any other changes planned?

chris-allan:
I had just force pushed away some of the crap before the weekend there to see if things would have stabilized which it looks like they have. I don't have any other changes planned right now.

joshmoore:
For everyone following along, one of the major (though straight-forward) wins here is that Ivy logging is increased which keeps travis from timing out. Propose merging this tomorrow.

sbesson (approved):
Quickly reviewed the proposed changes:

- :+1: for making the Ivy retrieval logging level configurable and consuming this in Travis CI
- :+1: for making the artifactory configurable as well as allowing to override the Bio-Formats without rebuilding the whole stack

The only potential downside of the log change in Travis would be to hit the [log limit](https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded). Assuming we are  currently still well below this limit, no objection to seeing this merged.

joshmoore:
All agreed. If we start to see log limit issues, we'll reconsider.

PR #6230, opened 2020-03-31
Test the reset password functionality

mtbc:
# What this PR does

Adds integration testing for [ResetPasswordRequest](https://docs.openmicroscopy.org/omero-blitz/5.5.6/slice2html/omero/cmd/ResetPasswordRequest.html).

# Testing this PR

Tests should pass against OMERO 5.6.0.
Tests should pass against OMERO 5.6.1 except for `testPasswordResetEmailGood` which can be seen failing on [OMERO-test-integration](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testngreports/integration/ResetPasswordTest/). *Update:* With https://github.com/ome/omero-blitz/pull/85 they should all pass.

# Related reading

https://forum.image.sc/t/35701
https://www.openmicroscopy.org/security/advisories/2019-SV3/

### DIFF
--- a/components/tools/OmeroJava/test/integration/AbstractServerTest.java
+++ b/components/tools/OmeroJava/test/integration/AbstractServerTest.java
@@ -897,14 +897,14 @@ public class AbstractServerTest extends AbstractTest {
     /**
      * Logs in the user.
      *
-     * @param ownerName
+     * @param omeName
      *            The OME name of the user.
      * @throws Exception
      *             Thrown if an error occurred.
      */
-    protected void loginUser(String ownerName) throws Exception {
+    protected void loginUser(String omeName) throws Exception {
         omero.client client = newOmeroClient();
-        client.createSession(ownerName, ownerName);
+        client.createSession(omeName, omeName);
         init(client);
     }
 
--- a/components/tools/OmeroJava/test/integration/ResetPasswordTest.java
+++ b/components/tools/OmeroJava/test/integration/ResetPasswordTest.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) 2020 University of Dundee & Open Microscopy Environment.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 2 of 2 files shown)


joshmoore:
I would assume that this means that these tests aren't being run?

```
/opt/omero-build/omero-blitz $ git grep ResetPasswordRequest
...
src/test/java/omero/cmd/admin/ResetPasswordRequestTest.java:public class ResetPasswordRequestTest extends AbstractServantTest {
src/test/java/omero/cmd/admin/ResetPasswordRequestTest.java:            final ResetPasswordRequestI req, Map<String, String> ctx) {
src/test/java/omero/cmd/admin/ResetPasswordRequestTest.java:        ResetPasswordRequestI req = new ResetPasswordRequestI(
```

mtbc:
It indeed seems likely they aren't. @jburel had been giving tests some care at some point, I don't know how far he got. I also still don't know why Travis runs different unit tests from `gradle test` locally for me. Maybe something to do with test groups? `unit.testng.xml` is presumably key.

jburel:
I will have a look again at the blitz tests. I think some could not be run at the time, but I could not remember the reason

jburel:
Only tests under ``omero.cmd.graphs.*`` are currently run I will look at enabling others

jburel:
ok now things start to come back to me after having a closer look
We have unit tests like ``ResetPassword`` using class i.e. ``AbstractServantTest`` which belongs to the ``integration`` group. so they are not run.
Changing that leads to error in the ``setup`` method

mtbc:
[OMERO-test-integration](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testngreports/integration/ResetPasswordTest/) is indeed as presently expected.

mtbc:
Next time it should pass.

jburel:
looking at the integration.testng.xml
```
<packages>
      <package name="ome.formats.*"/>
      <package name="ome.services.*"/>
      <package name="omero.util.*"/>
    </packages>
```
none of the cmd tests are included
So it is not a surprised that if are not run

joshmoore:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/384/ has definitely gone green. :+1:

PR #6231, opened 2020-04-02
xdist: fix import to disable pytest_configure_node

joshmoore:
`import xdist` apparently still passes even after
`pip uninstall pytest-xdist` but that's not the
case (currently) for `import xdist.plugin`.

### DIFF
--- a/components/tools/OmeroPy/test/conftest.py
+++ b/components/tools/OmeroPy/test/conftest.py
@@ -96,7 +96,7 @@ def pytest_configure(config):
 
 
 try:
-    import xdist  # noqa
+    import xdist.plugin  # noqa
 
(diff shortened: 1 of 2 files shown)


components/tools/OmeroPy/test/conftest.py
```diff
@@ -96,7 +96,7 @@ def pytest_configure(config):
 
 
 try:
-    import xdist  # noqa
+    import xdist.plugin  # noqa
```
will-moore, line 99: Works for me! - need the same fix in ```OmeroWeb``` too.

joshmoore: Done

PR #6239, opened 2020-06-17
Add test_shape() for paths_to_object

will-moore:
# What this PR does

Adds a test for paths_to_object with shape ID from https://github.com/ome/omero-web/pull/178


jburel:
A test should be added to SPW path too

PR #6240, opened 2020-06-23
Add integration test covering bulk import with log flags

sbesson:
This PR adds an integration test covering the scenario reported in https://github.com/ome/omero-py/issues/40

The test should be failing with `omero-py 5.7.0` or earlier but passing with the changes introduced in https://github.com/ome/omero-py/pull/223


sbesson:
See https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_import/TestImport/testBulkImportLogs/

PR #6241, opened 2020-07-13
Changelog 56 next

jburel:
# What this PR does

Add entry for next release
Version to be discussed 
cc @sbesson

### DIFF
--- a/history.rst
+++ b/history.rst
@@ -9,8 +9,9 @@ OMERO version history
 -----------------
 
 This release adds installation documentation for server and web on
-CentOS 8 and Ubuntu 20.04, and it removes installation instructions
-for Ubuntu 16.04.
+CentOS 8 and Ubuntu 20.04.
+We have dropped support for Ubuntu 16.04 and removes the corresponding
+installation instructions.
 
 This version has been tested with:
 
--- a/history.rst
+++ b/history.rst
@@ -5,6 +5,24 @@
 OMERO version history
 =====================
 
+5.6.2 (July 2020)
+-----------------
+
+This release adds installation documentation for server and web on
+CentOS 8 and Ubuntu 20.04, and it removes installation instructions
+for Ubuntu 16.04.
+
+This version has been tested with:
+
+- omero-blitz 5.5.7
+- omero-gateway-java 5.6.4
+- omero-py 5.7.1
+- omero-web 5.7.0
+- omero-dropbox 5.6.2
+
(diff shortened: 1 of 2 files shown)


mtbc (approved):

history.rst
```diff
@@ -5,6 +5,24 @@
 OMERO version history
 =====================
 
+5.6.2 (July 2020)
+-----------------
+
+This release adds installation documentation for server and web on
+CentOS 8 and Ubuntu 20.04, and it removes installation instructions
+for Ubuntu 16.04.
```
sbesson: Are we selling this as the main feature of the release or are there other high level improvements worth adding? As such it feels like there is little benefit to upgrading a 5.6.x server on a supported platform for instance.

For Ubuntu 16.04, should mention drop support?

history.rst
```diff
@@ -5,6 +5,24 @@
...
+
+This version has been tested with:
```
sbesson, line 16: Unrelated to this PR as i realize it is the wording of all 5.6.x releases. While it makes sense for omero-py and omero-web which are managed separately, these versions of the Java components are shipped with the new bundles.

Maybe we want separate paragraphs in the future.

history.rst
```diff
@@ -5,6 +5,24 @@
...
+
+- omero-blitz 5.5.7
+- omero-gateway-java 5.6.4
+- omero-py 5.7.1
+- omero-web 5.7.0
+- omero-dropbox 5.6.2
```
sbesson, line 22: Rfe it might be useful to link to the components changelogs or maybe summarize the major features.

sbesson (commented):
A few inline comments. Overall my main uncertainty is how we (under?)promote this release. I will have a look at the release annoucement.

history.rst
jburel: Links to the changelogs are in https://github.com/ome/ome-documentation/blob/develop/omero/changelog.rst

history.rst
jburel: Without replicating the content of the changelogs of each "component", this history file is becoming less verbose and less relevant.

sbesson: Understood. So is the announcement the best reference place to get an overview of all end-users features associated with each release

history.rst
joshmoore: I could imagine linking to all the versions (e.g. even omero-model) especially if they were links to the changelogs, though we'd likely want that to be auto-generated.

PR #6250, opened 2020-09-08
Add integration test for other-group duplication.

mtbc:
Adds broken test case for https://github.com/ome/omero-blitz/issues/103.


joshmoore (approved):
In general, :+1: for having a test for an open issue. https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/ws/src/components/tools/OmeroJava/target/reports/broken/html/index.html shows me:

```
java.lang.AssertionError: Found GraphException when pass==true: ::omero::cmd::Duplicate (graph-fail) params={stacktrace=ome.services.graphs.GraphException(message=cannot read ome.model.core.Image[9353])
at ome.services.graphs.GraphTraversal.getDetails(GraphTraversal.java:1089)
```

which also makes sense. :+1:

PR #6251, opened 2020-09-09
Test new command-line options for CLI duplicate.

mtbc:
Provides integration testing for https://github.com/ome/omero-cli-duplicate/pull/12.

### DIFF
--- a/components/tools/OmeroPy/test/integration/clitest/test_duplicate.py
+++ b/components/tools/OmeroPy/test/integration/clitest/test_duplicate.py
@@ -227,7 +227,7 @@ class TestDuplicate(CLITest):
         self.link(proj, dset)
 
         self.args += ['Project:%s' % proj.id.val]
-        self.args += ['--ignore-classes=ILink']
+        self.args += ['--ignore=ILink']
         self.args += ['--report']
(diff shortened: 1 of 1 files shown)


joshmoore:
Are there any complex corner cases when using multiple flags?

mtbc:
Good question. Nothing special comes to mind because the underlying code paths are fairly general. It's a good sign if the basic combinations work and it already augurs well that both `DuplicateTest` and `SpecificityClassifierTest` pass: the former also tries combinations and that latter's very much about combining type arguments like that.

joshmoore:
:+1: thanks. Can you describe / point to a conceivable multi-flag use case?

mtbc:
I am not sure how much I'm pushing at the envelope of "conceivable" but,
```
--reference-classes=Annotation
--duplicate-classes=BasicAnnotation,CommentAnnotation
--ignore-classes=Roi
```
should adjust duplication to,
* multiply link to annotations, reusing the existing,
* except it will duplicate some simple not-sharable ones like comments and ratings
* and omit RoIs from the duplication.

joshmoore (approved):
Looks good. Suggest merging with the rest of the PRs.

PR #6252, opened 2020-09-15
No longer need sql_run in "prepare" lifecycle.

mtbc:
Drop obsolete section of build config. Everything should run just as before.


joshmoore:
:+1:

PR #6253, opened 2020-09-23
Prepare changelog for releasing OMERO 5.6.3.

mtbc:
Update the changelog in preparation for releasing OMERO 5.6.3.

Stated versions are those to which I propose setting dependencies then naming the tag.


joshmoore (approved):
:+1:

PR #6258, opened 2020-10-01
Only assert_objects if len(exp_ids) > 0

will-moore:
# What this PR does

Fix test failure at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/543/testReport/junit/OmeroWeb.test.integration.test_api_experimenters_groups/TestExperimenters/test_groups_experimenters/

This failure can happen if the Group for which we check experimenters, has 0 experimenters.
Now we only `assert_objects()` if we have more than zero experimenters.
But to be sure that we call `assert_objects()`, we check all the groups that were previously loaded (up to 10) instead of just checking the first group.

### DIFF
--- a/components/tools/OmeroWeb/test/integration/test_api_experimenters_groups.py
+++ b/components/tools/OmeroWeb/test/integration/test_api_experimenters_groups.py
@@ -138,18 +138,21 @@ class TestExperimenters(IWebTest):
 
         assert_objects(conn, groups_json, groups, dtype="ExperimenterGroup")
 
-        # Check we can follow link to Experimenters for first Group
-        expimenters_url = groups_json[0]["url:experimenters"]
-        rsp = get_json(django_client, expimenters_url)
-        exps_json = rsp['data']
-        exp_ids = [e['@id'] for e in exps_json]
+        # Check experimenters_url for all groups above
+        for group_json in groups_json:
+            # Check we can follow link to Experimenters for first Group
+            expimenters_url = group_json["url:experimenters"]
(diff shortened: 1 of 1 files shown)


components/tools/OmeroWeb/test/integration/test_api_experimenters_groups.py
```diff
@@ -138,18 +138,22 @@ def test_groups_experimenters(self, user1):
...
-        # Check if eids are same for group (won't be ordered)
-        grp = conn.getObject("ExperimenterGroup", groups_json[0]['@id'])
-        eids = [link.child.id.val for link in grp.copyGroupExperimenterMap()]
-        assert set(eids) == set(exp_ids)
-
-        assert_objects(conn, exps_json, exp_ids, dtype="Experimenter")
+        # Check experimenters_url for all groups above
+        for group_json in groups_json:
+            # Check we can follow link to Experimenters for first Group
+            expimenters_url = group_json["url:experimenters"]
```
mtbc, line 144: Maybe "experimenters" rather than "expimenters". (Not a blocker though.)

sbesson (approved):
The CI job has been consistently passing since the introduction of this PR - see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/550/testReport/OmeroWeb.test.integration.test_api_experimenters_groups/history/
The test time is unmodified i.e. the looping over all groups does not impact the performance and makes sense compared to testing the first group only.

Given we saw intermittent failures every ~10 builds recently, proposing to leave this PR opened for the rest of the week and get this merged.

PR #6262, opened 2020-12-14
Documentation for https://github.com/ome/omero-py/pull/272

manics:
# What this PR does

https://github.com/ome/omero-py/pull/272 adds a new property `omero.server.nodedescriptors`. This adds some docs that should end up on https://docs.openmicroscopy.org/omero/5/sysadmins/config.html.

# Testing this PR

See the examples linked from https://github.com/ome/omero-py/pull/272

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -156,7 +156,9 @@ omero.glacier2.IceSSL.Protocols=tls1
 # set this to ``master:Blitz-0,Tables-0``.
 # Also use this to distribute OMERO services across multiple nodes,
 # for example:
-# ``master:Blitz-0,Tables-0 worker1:Processor-0``
+# ``master:Blitz-0,Tables-0 worker1:Processor-0``.
+# See
+# https://docs.openmicroscopy.org/omero/latest/sysadmins/grid.html#deployment-examples
 omero.server.nodedescriptors=
 
 #############################################
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -150,6 +150,15 @@ omero.glacier2.IceSSL.VerifyPeer=0
 # Glacier2Template SSL allowed protocols
 omero.glacier2.IceSSL.Protocols=tls1
 
+# Override the default set of OMERO servers.
+# For example, to run OMERO.server with Blitz and Tables only
+# (i.e. disable Processor, DropBox, Indexer, PixelData)
+# set this to ``master:Blitz-0,Tables-0``.
+# Also use this to distribute OMERO services across multiple nodes,
+# for example:
+# ``master:Blitz-0,Tables-0 worker1:Processor-0``
+omero.glacier2.IceSSL.Protocols=
+
 #############################################
 ## Darwin (OS X) specific defaults for templated configuration files only.
 ## ``_xxx.yyy.darwin`` properties will override xxx.yyy properties in this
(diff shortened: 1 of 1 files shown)


etc/omero.properties
```diff
@@ -150,6 +150,15 @@ omero.glacier2.IceSSL.VerifyPeer=0
...
 omero.glacier2.IceSSL.Protocols=tls1
 
+# Override the default set of OMERO servers.
+# For example, to run OMERO.server with Blitz and Tables only
+# (i.e. disable Processor, DropBox, Indexer, PixelData)
+# set this to ``master:Blitz-0,Tables-0``.
+# Also use this to distribute OMERO services across multiple nodes,
+# for example:
+# ``master:Blitz-0,Tables-0 worker1:Processor-0``
+omero.glacier2.IceSSL.Protocols=
```
joshmoore: This seems wrong... 🙂

manics:
```suggestion
omero.server.nodedescriptors=
```

manics: Fixed!

etc/omero.properties
```diff
@@ -150,6 +150,15 @@ omero.glacier2.IceSSL.VerifyPeer=0
 # Glacier2Template SSL allowed protocols
 omero.glacier2.IceSSL.Protocols=tls1
 
+# Override the default set of OMERO servers.
```
will-moore: Are these `OMERO servers` or `services`?

will-moore:
Just wondering what the `-0` is for after each service name? Do we need any explanation of that? Is there any reason to use any other value?
Once other comment above, but otherwise looks good.

etc/omero.properties
manics: If it was me I'd call them services, but the ice teminology is `server`:
```
$ omero admin ice server list
Blitz-0
DropBox
FileServer
Indexer-0
MonitorServer
OMERO.Glacier2
OMERO.IceStorm
PixelData-0
Processor-0
Tables-0
TestDropBox
```
There's a separate `omero admin ice service ...` command but I can't work out what it does. What do you reckon is best?

manics:
The `-0` is because you can run multiple copies of some services, e.g. you can have multiple `processor-N` spread across multiple nodes. It's partly documented on https://docs.openmicroscopy.org/omero/5.6.3/sysadmins/grid.html#deployment-examples

I'm reluctant to put too much technical detail into this file because it's hard enough keeping our existing docs synchronized, but I'll add a link to `https://docs.openmicroscopy.org/omero/latest/sysadmins/grid.html#deployment-examples`

etc/omero.properties
joshmoore: Pretty sure services only work in IceBox which we don't use: https://doc.zeroc.com/ice/3.7/icebox

manics: The problem is what do we use in the docs, `service` since that's what intuitively makes sense or `server` since that's the ice terminology?

will-moore: Yes, even in the same paragraph below it's "distribute OMERO services across multiple nodes" which is confusing.

manics: No one else said anything so I've changed it to `services`

joshmoore:
I'd vote time to get this in and prepare the omero-py release.

joshmoore:
Ok. Merging. I'll spin up the omero-py release ASAP (likely today) and then we can discuss a new docs release.

PR #6263, opened 2021-01-28
Test create image from numpy seq

will-moore:
# What this PR does

Adds tests for `conn.createImageFromNumpySeq(planeGen...)` with big images, as supported by https://github.com/ome/omero-py/pull/276

I was hoping that this method would "just work" exactly the same with big images and small images, so you didn't have to worry about whether you were creating a big image or not.
Unfortunately, the `TileLoop` for populating big images with `setTile()` wants to consume planes in the order:

```
for t in range(size_t):
    for c in range(size_c):
       for z in range(size_z):
           ...
```
so the `planeGenerator` passed to ``conn.createImageFromNumpySeq()` needs to provide planes in that order.
But for small images, we need:

```
for z in range(size_z):
    for c in range(size_c):
       for t in range(size_t):
           ...
```

So the tests in this PR need to use these different orders for creating big vv small images using the above PR.

Obviously this is not ideal. But I couldn't think of any way to populate Big Images with the order of planesthat is already required by `conn.createImageFromNumpySeq()`.

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_create_image.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_create_image.py
@@ -1,0 +1,61 @@
+
+import numpy as np
+from itertools import chain
+import pytest
+
+import omero
+from omero.gateway import BlitzGateway
+from omero.testlib import ITest
+
+
+class TestCreateImage(ITest):
+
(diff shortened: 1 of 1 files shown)


joshmoore:
> But for small images, we need:

Why's that the case?

will-moore:
That's just the API as it stands now. Changing that will be a breaking change, but you won't *know* it's broken until you find that your new image dimensions have been swapped.

PR #6264, opened 2021-02-19
Add more lines to test_get_line_data()

will-moore:
# What this PR does

Extends an existing test to cover the bug fixed in https://github.com/ome/omero-py/pull/278


jburel:
Thanks for adding the tests

PR #6273, opened 2021-04-08
Add tests for getObjectsByMapAnnotations

dominikl:
Add integration test for https://github.com/ome/omero-py/pull/285 .

### DIFF
--- a/components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
+++ b/components/tools/OmeroPy/test/integration/gatewaytest/test_get_objects.py
@@ -24,7 +24,8 @@ import pytest
 from omero.gateway.scripts import dbhelpers
 from omero.rtypes import wrap, rlong
 from omero.testlib import ITest
-from omero.gateway import BlitzGateway, KNOWN_WRAPPERS
+from omero.gateway import BlitzGateway, KNOWN_WRAPPERS, DatasetWrapper, \
+    ProjectWrapper, ImageWrapper, ScreenWrapper, PlateWrapper
 from omero.model import DatasetI, \
     ImageI, \
     PlateI, \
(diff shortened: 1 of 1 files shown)


will-moore:
Thanks @dominikl - That looks like pretty good coverage.
Tests all passing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/732/testReport/OmeroPy.test.integration.gatewaytest.test_get_objects/TestGetObject/

I guess one thing that could be added is to confirm that without a leading (or trailing) wild-card returns nothing.
```
# returns nothing
wc = value[2:12]+"*"
# returns something
wc = value[0:12]+"*"
```

And could also test that the limit works:
```
# should only return 2 objects (and don't need key, value or ns)
conn.getObjectsByMapAnnotations(datatype, opts={"limit": 2})
```
Actually, I just tried that myself and found a bug!
Since I'm initially querying for `ImageAnnotationLinks` I was getting duplicate Images in that list, but then the returned objects list were removing the duplicates. So for the `opts={"limit": 2}` - that meant that only 2 `ImageAnnotationLinks` were found, but if they were both for the same Image, then I'd only get 1 Image returned!
I only found that issue because in my local server I have many Images with multiple MapAnnotations.

That is now fixed by https://github.com/ome/omero-py/pull/285/commits/d27f2f0013bee7f089ed17bfe0b2e16bd98556fb

So maybe if your `createTarget()` actually added 2 identical Map annotations to each object, then test if `conn.getObjectsByMapAnnotations(datatype, opts={"limit": 4})` actually gets 4 objects.

Thanks!

dominikl:
👍 I'll add the additional tests.

will-moore (approved):
Looks good. Passing locally (but fails if I don't have last commit https://github.com/ome/omero-py/commit/d27f2f0013bee7f089ed17bfe0b2e16bd98556fb) 👍

PR #6275, opened 2021-04-27
Add OMERO5.4__1 for omero-model#71

joshmoore:
Bumps the databse to `OMERO5.4__1`. This is likely a better strategy than performing the indexing in the background since it potentially takes a considerable amount of time. However, this is up for discussion.

See: https://github.com/ome/omero-model/pull/72

--exclude


sbesson:
Closing as per ome/omero-model#72 as we have not identified a workflow where this index would significantly improve performance.

joshmoore:
Is that the right link?

PR #6277, opened 2021-04-29
Remove Python 2.7 requirement files

joshmoore:
Noticed after a dependabot alert.


sbesson (approved):
Makes complete sense given OMERO is now Python 3+.
Looking at the places where this might still be consumed, https://github.com/search?q=org%3Aome+requirements-py27&type=code and https://github.com/search?q=org%3Aome+requirements-redis&type=code, I do not see anything that would require an immediate update and we can review it as we start re-enabling these legacy code (e.g. the robot tests)

PR #6283, opened 2021-06-24
Add integration test for searching columns with spaces in their names

sbesson:
Depends on the new API introduced in https://github.com/ome/omero-py/pull/287

This adds a new integration test demonstrating the usage of the internal variables mapping to access columns with spaces in their names. A few various use cases are tested including different comparison operators, two column types (string & long) as well as combined expressions.


joshmoore (approved):
:+1:

PR #6288, opened 2021-08-19
Test download fileset

will-moore:
Update test_download.py for https://github.com/ome/omero-py/pull/298

 - Handle download of Image to target dir instead of single file
 - Test download of image with companion files (now supported by download plugin)


sbesson (approved):
Merging alongside https://github.com/ome/omero-py/pull/298

PR #6290, opened 2021-08-26
Fix polygon ROI points string format

Yu-AnChen:
# What this PR does

The current point string format (`"X1,Y1, X2,Y2, X3,Y3"`) in [python training example](https://github.com/ome/openmicroscopy/blob/develop/examples/Training/python/ROIs.py) is causing an error in PathViewer and resulting in a "locked" state in the annotation tools. Changing the point string format to `"X1,Y1 X2,Y2 X3,Y3"` fixed the issue.

# Related reading

- See [Polygon/@Points](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Polygon_Points) section in OME Schemas


will-moore:
Thanks for the fix. Looks good to merge 👍

PR #6291, opened 2021-08-26
Robot fixes

will-moore:
As discussed, https://github.com/ome/omero-web/pull/180 needs significant testing and would benefit from running robot tests, which haven't been run in a while (not since Python 2).

Updating instructions to work with recent libs and Python 3.

I ran all tests manually against local server running merge-ci branch.

Some tests needed fixing due to element change (`a` -> `span`) in jsTree https://github.com/ome/omero-web/pull/257
and other DOM changes in https://github.com/ome/omero-web/pull/270 (not merged).

Only 1 bug found due to jQuery (and jQuery UI) update, causing a button text not to update from "No" to "Cancel". Fixed in https://github.com/will-moore/omero-web/commit/ed7f2f02d13a2efbf40f82ac0cc087e9e5ef81ef

A couple of other (long-standing) issues discovered: https://github.com/ome/omero-web/issues/314, https://github.com/ome/omero-web/issues/315.

Functionality has removed for Shares. Tests updated.
Old webgateway viewer no-longer has access to `window.parent` (webclient window) so the Prev/Next image buttons are disabled. Haven't discovered why yet, but maybe not worth fixing since everyone uses iviewer now.


components/tests/ui/README.md
```diff
@@ -48,14 +48,14 @@ If you are running the tests on Mac OS X, you can install with the following
 brew install geckodriver
 ```
 
+NB: this took too long for me, and manually installing geckodriver
```
joshmoore: Maybe change this to "If installing takes too long, ..."

joshmoore (commented):
The high-level changes (Python 3, changes in robot) all seem fine. It's difficult to know what the impact of all the many changes to the xpaths, etc. will mean for users.

components/tests/ui/README.md
will-moore: Fixed in b4700164af

will-moore:
Changing the DOM structure (and therefore the xpath to specified elements) shouldn't affect any users because they don't depend on the DOM.
Certainly, UI users don't care, and devs *shouldn't* be depending on the DOM outside of their `<div>` container. The DOM is not an API.

jburel:
During the setup phase, we have the following message
```
omero metadata testtables
This plugin is deprecated as of OMERO 5.4.8. Use the plugin available from https://pypi.org/project/omero-metadata/ instead.
```
This should probably be upgraded too

will-moore:
@jburel My versions:
```
$ pip freeze | grep robot
robotframework==4.1
robotframework-pythonlibcore==3.0.0
robotframework-selenium2library==3.0.0
robotframework-seleniumlibrary==5.1.3
```

jburel:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-robot/32/ is green

jburel:
Also the instructions in the README need to be adjusted a bit since they were written before the decoupling work

jburel:
```
./components/tests/ui/plugins/robot.py:80:80: E501 line too long (85 > 79 characters)
```

will-moore:
Thanks @jburel - flake8 fixed

joshmoore:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-robot/34/ is certainly green. (I don't know anything about the "non-critical" classification)

jburel:
The failures are due to the fact that a window/dialog was slow to show up on screen.
The "double-run" (Google&Firefox) will reduce that number
Merging. Thanks @will-moore

PR #6296, opened 2021-12-03
Update omero-scripts to 5.6.1

joshmoore:
see: https://forum.image.sc/t/crash-with-images-from-rois-script-with-rois-3k/60538/4?u=joshmoore


jburel:
thanks

PR #6301, opened 2022-01-28
Use the copy

jburel:
Copy the file if found locally above src

Background:

During the daily build process, an entry is added to ``etc/omero.properties``  indicating where to download the artifacts from, e.g. https://merge-ci.openmicroscopy.org/jenkins//job/OMERO-python-superbuild-build/lastSuccessfulBuild/artifact/PACKAGE/dist/
This only works when using DNS.

In the case of the temporary devspace (no DNS the url is set to for example https://github.com/snoopycrimecop/openmicroscopy/blob/snoopycrimecop_ci/etc/omero.properties#L211 but the url points to nothing real ("real" is something like idr1-slot3.o.org:PORT/jenkins/...)

To support the non DNS case, artifacts are copied using the "CopyArtifacts" option in the jenkins job and moved to the directory where ``src`` is (location can be modified if there is a more suitable place)
The proposed change first looks if the artifact is available locally before using the ``versions.omero-pypi`` property.

### DIFF
--- a/components/tools/build.xml
+++ b/components/tools/build.xml
@@ -36,8 +36,18 @@
 			side code for use with OMERO.
 		</echo>
 		<mkdir dir="target/lib/server"/>
-		<download pkg="omero-scripts" file="omero-scripts-${versions.omero-scripts}.tar.gz" expected="${versions.omero-scripts-md5}"
+		<!-- Look for file above src -->
+		<property name="filegz" value="../../../omero-scripts-${versions.omero-scripts}.tar.gz" />
+		<if>
+			<available file="${filegz}"/>
+            <then>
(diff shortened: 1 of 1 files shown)


joshmoore:
@jburel : can you add some more of the background to the description here? What's triggering the need for the change?

jburel:
@joshmoore background added to the description

jburel:
Edit: removing exclude label to work on new CI

jburel:
This PR has now been opened for a long time. I suggest that we merge it otherwise the strategy to deploy new ci will not work

cc @pwalczysko

pwalczysko:
The build is green now. https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-build/

pwalczysko (approved):

PR #6302, opened 2022-01-29
Build

jburel:
This PR

* Adds dependencies to fix fresh build failure. The problem only occurs when running ``build-dev``
* Removes ``zeroc`` resolver. Breaks the build
* Removes target no longer required following the split. Breaks the build

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -264,7 +264,7 @@ To get started using Eclipse, execute "ant build-dev" and import the top-level
     </target>
 
     <target name="release-all" description="Produce all binary release artifacts (unsigned)"
-        depends="release-clients,release-zip,release-docs">
+        depends="release-clients,release-zip">
    </target>
 
     <target name="release-zip" description="Zip the dist directory into ${product.name}-${omero.version}.zip">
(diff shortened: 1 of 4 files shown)


sbesson:
Looking at the downloads for OMERO 5.6.3, the OMERO.apidocs artifact is effectively missing all content. I agree this target does not make sense anymore (probably since OMERO 5.6.0) and removing it from this build is probably the most sensible option. MOving forward I assume:
- some of the release workflow might need to be adjusted e.g. to skip the unpacking of  under https://downloads.openmicroscopy.org/omero/5.6.3/api/python/
- I assume the generation of this reference API documentation should happen in OMERO.py moving forward. Should this be captured as an issue (or is there already onyl?)

Same thoughts regarding the removal of the ZeroC Nexus. Given this resource is no longer available, it is primarily a source of noise and warnings and it's a good idea to remove the dependency assuming the relevant dependency are all available in secondary repositories.

In which part of the code are the `jcommander` and `joda-time` dependencies used? I assume they were previously pulled transitively so do we know what changed and caused these new failures?

jburel:
* zeroc. On a fresh infra it is more than a warning, it does not build at all
* for the 2 dependencies, older version are used but the jars do not get downloaded. I investigated the dependencies tree in some repositories but I have not been able to find the exact source of the failure
* For the docs. Yes adjustments to be made. I have not created an issue regarding the omero-py api

jburel:
Replaced by #6305 and #6306

PR #6303, opened 2022-01-31
delete the roi

jburel:
Delete the newly created roi


sbesson (approved):
Straightforward change.

PR #6307, opened 2022-02-22
update ice version to match omero-py

jburel:
Nothing to test
see https://github.com/ome/omero-py/blob/master/src/omero_version.py

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -185,7 +185,7 @@ product.name=OMERO.server
 #############################################
 ##
 versions.findbugs=1.2.1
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Unifies the patch version with the value in https://github.com/ome/omero-py/blob/14b4ae38be7aa172accd152ceac52aa95b130e38/src/omero_version.py#L2 so it feels appropriate.

Was anything broken previously due to the mismatch or is that primarily for consistency?

jburel:
nothing broken. Mainly consistency

PR #6308, opened 2022-02-22
remove usage of LooseVersion

jburel:
When build is done using Python 3.8
The ``omero-version`` is not set correctly and it is then not possible to run for example ``./build.py clean``

i hit this problem again when doing the following:
* Create a local conda environment with Python 3.8. I wanted to use omero-py
* Activate the environment
* Try to run ``./build clean``
* Error: ``omero_version = "parse_version:99: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead. ``

### DIFF
--- a/components/antlib/scripts/parse_version
+++ b/components/antlib/scripts/parse_version
@@ -95,8 +95,8 @@ def incr_version(omero_version):
     """
     try:
         # First we try to use distutils
-        from distutils.version import LooseVersion
-        version = LooseVersion(omero_version).version
+        from packaging.version import Version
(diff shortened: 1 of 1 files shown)


sbesson (approved):
`distutils.version` is deprecated and `packaging.version` is the natural replacement.

The only unknown is whether the loose requirements of the `LooseVersion` API were a requirement here. Looking at the history, I cannot find  more information about the context where this is used. Possibly all of this logic is anterior to the extraction of the server components. The CI builds have been passing with this included so merging in the absence of obvious error.

PR #6309, opened 2022-03-04
add try block

jburel:
PR #6303 fixed an initial problem
Adding try block to avoid build failure


joshmoore:
No real objection, here, @jburel, if it keeps tests passing, but it isn't really a style that we want others to copy, eh? Did this happen recently?

jburel:
It happens from time to time. It happens last time when https://github.com/ome/openmicroscopy/pull/6303 was opened

sbesson:
Closing to see if the recent CI work has fixed the issue

PR #6311, opened 2022-04-14
Prep 5.6.4 release

jburel:
* Bump dependencies version 
* Add entry in history

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -211,7 +211,7 @@ versions.omero-scripts-url=${versions.omero-pypi}
 versions.omero-blitz=5.5.10
 versions.omero-common-test=5.5.9
 versions.omero-gateway=5.6.9
-versions.omero-scripts=5.6.1
+versions.omero-scripts=5.6.2
 versions.OMEZarrReader=0.1.0
 ## Global overrides, if empty ignored
 versions.bioformats=
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -208,9 +208,9 @@ versions.omero-scripts-url=${versions.omero-pypi}
 ###
 
 ## Internal dependencies
-versions.omero-blitz=5.5.9
-versions.omero-common-test=5.5.8
-versions.omero-gateway=5.6.8
+versions.omero-blitz=5.5.10
+versions.omero-common-test=5.5.9
+versions.omero-gateway=5.6.9
 versions.omero-scripts=5.6.1
 versions.OMEZarrReader=0.1.0
 ## Global overrides, if empty ignored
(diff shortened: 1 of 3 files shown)


jburel:
Announcement will be in the spirit of https://www.openmicroscopy.org/2020/10/01/omero-5-6-3.html

sbesson (approved):
If the official announcement will be drafted separately, no objections to getting this to start generating the artifacts.

For the announcement, I would say:
- all considerations regarding the Django 3.2.x testing are irrelevant for consumers although it might be worth adding a note reminding people to upgrade their OMERO.web environment if they haven't as part of the upgrade
- certainly want to mention the Bio-Formats upgrade alongside the memo invalidation
- I assume we are don't necessarily want to promote the ZarrReader integration for this release. Otherwise, we might want to bump https://github.com/ome/openmicroscopy/blob/develop/etc/omero.properties#L215 to use the latest patch release immediately

jburel:
I don't think we need to promote the zarr reader but it does not hurt to bump the version.

etc/omero.properties
```diff
@@ -208,9 +208,9 @@ versions.omero-scripts-url=${versions.omero-pypi}
 ###
 
 ## Internal dependencies
-versions.omero-blitz=5.5.9
-versions.omero-common-test=5.5.8
-versions.omero-gateway=5.6.8
+versions.omero-blitz=5.5.10
+versions.omero-common-test=5.5.9
+versions.omero-gateway=5.6.9
 versions.omero-scripts=5.6.1
```
sbesson: Should this also be bumped to `version.omero-scripts=5.6.2`  which includes several bug fixes - see https://github.com/ome/omero-scripts/compare/v5.6.1...v5.6.2 ?

jburel: yes. I thought i did bump it

jburel:
back to green

history.rst
```diff
@@ -5,6 +5,22 @@
...
+
+This release improves and expands integration testing to handle Django 3.2.x.
+It also removes obsolete code.
+More importantly, it has been tested with:
+
+- omero-blitz 5.5.10
+- omero-gateway-java 5.6.9
+- omero-py 5.11.1
+- omero-web 5.14.0
+- omero-dropbox 5.6.2
```
sbesson, line 19: Should the `omero-scripts` bump be mentioned here?

sbesson (approved):
One amendment, otherwise LGTM

history.rst
jburel: It won't hurt

sbesson (approved):

jburel:
Merging and testing the build

PR #6312, opened 2022-04-15
Build fix

jburel:
The introduction of the zarr reader leads to a build failure on fresh installation
Problem noticed while preparing the doc
see https://github.com/ome/omero-documentation/actions/runs/2170402809 for example

To test:
* Run locally ``./build.py clean && ./build.py``. You may want to clean the maven cache too.

Edit: [30a7127](https://github.com/ome/openmicroscopy/pull/6312/commits/30a7127cd2ff2b2b88ce73a577ec92ddfbd6c936) builds from source. Check that it is green

### DIFF
--- a/.github/workflows/omero_plugin.yml
+++ b/.github/workflows/omero_plugin.yml
@@ -9,7 +9,7 @@
 #
 # 2. Adjust the cron schedule as necessary
 
-name: OMERO
+name: OMERO Plugin
 on:
   push:
   pull_request:
@@ -31,30 +31,6 @@ jobs:
           repository: ome/omero-test-infra
           path: .omero
           ref: ${{ secrets.OMERO_TEST_INFRA_REF }}
-      - uses: actions/setup-python@v2
-        with:
-          python-version: 3.8
-      - name: Set up JDK 11
-        uses: actions/setup-java@v1
-        with:
-          java-version: 11
-      - uses: actions/cache@v1
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-
-      - name: Install slice2java
-        run: |
-          sudo apt-get install -y libmcpp-dev
-          wget -q https://github.com/ome/zeroc-ice-ubuntu2004/releases/download/0.2.0/ice-3.6.5-0.2.0-ubuntu2004-amd64.tar.gz
-          tar xf ice-3.6.5-0.2.0-ubuntu2004-amd64.tar.gz
-          mv ice-3.6.5-0.2.0 ice-3.6.5
-          mv ice-3.6.5 /opt
-          rm ice-3.6.5-0.2.0-ubuntu2004-amd64.tar.gz
-          echo "${{ env.ICE_HOME }}/bin" >> $GITHUB_PATH
-      - name: Build
-        run: ./build.py
       - name: Set up log level
         run: |
           echo 'ivy.retrieve.log=default' >> etc/local.properties
--- a/.github/workflows/omero_plugin.yml
+++ b/.github/workflows/omero_plugin.yml
@@ -22,7 +22,6 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       STAGE: srv # app, cli, lib, scripts, srv
-      ICE_HOME: /opt/ice-3.6.5 # location where Ice is installed
     steps:
       - uses: actions/checkout@v2
       - name: Checkout omero-test-infra
--- a/.github/workflows/omero_plugin.yml
+++ b/.github/workflows/omero_plugin.yml
@@ -34,5 +34,5 @@ jobs:
       - name: Set up log level
         run: |
           echo 'ivy.retrieve.log=default' >> etc/local.properties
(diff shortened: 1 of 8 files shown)


jburel:
[30a7127](https://github.com/ome/openmicroscopy/pull/6312/commits/30a7127cd2ff2b2b88ce73a577ec92ddfbd6c936) builds the source. 
We did not build at all or run the tests (everything is based on the docker images)
This should avoid having a possible broken build without using realising. 
We can decide to review the strategy.

.github/workflows/omero_plugin.yml
```diff
@@ -30,6 +31,30 @@ jobs:
           repository: ome/omero-test-infra
           path: .omero
           ref: ${{ secrets.OMERO_TEST_INFRA_REF }}
+      - uses: actions/setup-python@v2
```
joshmoore: Would this do what you need as a separate workflow?

joshmoore (commented):
Few other comments:
 - I would assume that the Glencoe nexus is mirroring this. Maybe since we sometimes need a mirror _anyway_, it would be an idea to use that as the additional repo? cc: @chris-allan 
 - Do we know why the docker build didn't fail? Is perhaps some flag missing?

.github/workflows/omero_plugin.yml
jburel: I have not tried it to be honest, time was not on my side :-)

jburel:
I need to investigate why it is not working with the docker build itself or why it was commented out. I opted for an option that will give us the confidence

For the dependency, I have followed https://github.com/ome/ZarrReader/blob/main/pom.xml#L44
we could adjust that as required

jburel:
I think I remember now why it has been disabled. When a new method is added for example in the gateway-java, and a test is added to this repository. The version used is a release version of the java-gateway not a specific branch like in CI. It then leads to an error.

.github/workflows/omero_plugin.yml
joshmoore: Ok. Pushed.

joshmoore:
> For the dependency, I have followed https://github.com/ome/ZarrReader/blob/main/pom.xml#L44 we could adjust that as required

Fair enough. Then we can discuss upstream and propagate back here.

.github/workflows/omero_plugin.yml
joshmoore: Interestingly this is now failing with:

```
copy-client:
Created dir: /home/runner/work/openmicroscopy/openmicroscopy/dist/lib/client
:: problems summary ::
:::: WARNINGS
		[NOT FOUND  ] io.netty#netty-transport-native-epoll;4.1.46.Final!netty-transport-native-epoll.jar (0ms)
	==== user-maven: tried
	  /home/runner/.m2/repository/io/netty/netty-transport-native-epoll/4.1.46.Final/netty-transport-native-epoll-4.1.46.Final.jar
		[NOT FOUND  ] org.kohsuke#file-leak-detector;1.13!file-leak-detector.jar (0ms)
	==== user-maven: tried
	  /home/runner/.m2/repository/org/kohsuke/file-leak-detector/1.13/file-leak-detector-1.13.jar
		::::::::::::::::::::::::::::::::::::::::::::::
		::              FAILED DOWNLOADS            ::
		:: ^ see resolution messages for details  ^ ::
		::::::::::::::::::::::::::::::::::::::::::::::
		:: io.netty#netty-transport-native-epoll;4.1.46.Final!netty-transport-native-epoll.jar
		:: org.kohsuke#file-leak-detector;1.13!file-leak-detector.jar
		::::::::::::::::::::::::::::::::::::::::::::::
```

sbesson: Interesting, `file-leak-detector` is a dependency from the `test-suite` component of ome/bioformats for detecting unclosed file handles.

I suspect it is being pulled as a result of https://github.com/ome/ZarrReader/blob/20ccbe52ff68bb8f7af0ed460c6cc29b0e31ffb8/pom.xml#L124-L127. Assuming nothing depends on it (which arguably would be a bug), either this dependency should be removed completely from the POM or it should be declared within the `test` scope`.

jburel:
@joshmoore I am working on fixes to re-activate the docker build and hopefully get something working. Obviously the error above is a problem. Strange that it did not happen before

jburel:
Interesting that one build and the other one does not

.github/workflows/omero_plugin.yml
jburel: Adding ``<scope>test</scope>`` leads to an error when running ``mvn install``
```
ZarrReader.java:[640,15] cannot find symbol
  symbol:   variable StringUtils
```
the apache dependency is then missing.

jburel:
One more to go
```
		::::::::::::::::::::::::::::::::::::::::::::::
		::              FAILED DOWNLOADS            ::
		:: ^ see resolution messages for details  ^ ::
		::::::::::::::::::::::::::::::::::::::::::::::
		:: io.netty#netty-transport-native-epoll;4.1.46.Final!netty-transport-native-epoll.jar
```
Coming from https://github.com/ome/ZarrReader/blob/main/pom.xml#L85
The version used has numerous CVEs

jburel:
Replaced by https://github.com/ome/openmicroscopy/pull/6313

PR #6313, opened 2022-04-20
Remove zarr reader

jburel:
Discussed with @sbesson and @dgault.
This PR 
* reactivate the build via GHA
* bump omero dependencies version to pass the build
* remove the ZarrReader for now due to https://github.com/ome/openmicroscopy/pull/6312#issuecomment-1103083040

The reader will be turned back on post release

Make sure the GHA builds are green

### DIFF
--- a/.github/workflows/omero_plugin.yml
+++ b/.github/workflows/omero_plugin.yml
@@ -9,7 +9,7 @@
 #
 # 2. Adjust the cron schedule as necessary
 
-name: OMERO
+name: OMERO Plugin
 on:
   push:
   pull_request:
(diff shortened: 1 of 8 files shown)


jburel:
Any objection to merge this PR?
Final version bump will happen in https://github.com/ome/openmicroscopy/pull/6311

sbesson (approved):
As discussed yesterday, I am fine with the proposed dependency changes. No objection to removing the `OMEZarrReader` dependency for the OMERO 5.6.4 release. As several components of the Java dependency stack still need to be reviewed/unified, this feels like a sensible approach to deliver value to the community immediately. The reader can be re-included  post-release.

On the build changes, I have spent minimal time reviewing it, so it might require someone else's input. Does the new workflow tries to fix failures which were not detected previously? Is the vision to unify with the `OMERO` workflow or are the two workflows separate?

jburel:
One workflow (original) is reactivated, the new one detect error not found when using test-infra. Ideally it should have failed in test-infra we can investigate that afterwards

jburel:
the changes in the python files are required to pass the GHA checks

jburel:
Merging so the autogen doc job can turn green.
GHA can be reviewed later

PR #6315, opened 2022-04-25
Turn ZarrReader back on

jburel:
* Turn zarr back on
* Fix GHA build when Zarr is included.


.github/workflows/omero_plugin.yml
```diff
@@ -33,5 +33,5 @@ jobs:
       - name: Set up log level
         run: |
           echo 'ivy.retrieve.log=default' >> etc/local.properties
-      - name: Build and run OMERO tests
-        run: .omero/docker $STAGE
+      #- name: Build and run OMERO tests
+      #  run: .omero/docker $STAGE
```
sbesson: Does the command now fail when the new dependency is introduced? I was this under the impression we had a false negative previously

jburel: The command does not fail. I removed it since it takes time, running some integrations tests should be done in the CI infra and not necessarily on each commit. The test-infra does not solve the problem when a change is made in the java-gateway for example and a test is modified. It will fail every time in GHA but obviously not in CI. The flake8 is missing, this needs to be checked.

sbesson: So from your side, is there a value in retaining the workflow at all?

jburel: The only thing we should keep is flake8 check for python code e.g. test but that could be added to the other one.
I have not looked at it yet

jburel: I think I am going to remove the GHA in this PR and add the flake8 in a follow-up PR since some Python examples need to be adjusted. 
I don't want to push the flake8 related commits to this PR which about the zarr reader and the build

jburel: Removed

jburel:
This is crazy. Now netty issue is back!

.github/workflows/omero_plugin.yml
jburel: https://github.com/ome/openmicroscopy/pull/6317 flake8 added.

sbesson (approved):

PR #6316, opened 2022-04-28
Release workflow

jburel:
Follow up from 5.6.4 release.
This PR adds a release workflow to avoid the manual migration of the artifacts from downloads.o.org to GitHub.

The workflow does the following, When a tag is pushed:
* The artifacts are built
* A release is created
* The artifacts are uploaded

I have opted for one single ``SHASUMS`` and ``MD5`` collecting the various values to simplify the selection and reduce the number of uploaded files.

This can replaced the "release" job on CI

See https://github.com/jburel/openmicroscopy/releases/tag/v5.6.4-m1

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,12 @@ jobs:
           mv ice-3.6.5 /opt
           rm ice-3.6.5-0.2.0-ubuntu2004-amd64.tar.gz
           echo "/opt/ice-3.6.5/bin" >> $GITHUB_PATH
+      - name: Set version
+        run: |
+          tag_name="${GITHUB_REF##*/}"
+          tag_value="${tag_name:1}"
+          echo tag_value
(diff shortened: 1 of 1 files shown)


sbesson:
At first reading, it looks like this workflow copies over most of the setup/building logic of the source build workflow introduced in #6313. Would it make sense to merge the two workflows into a single one with some conditional task  execution?

jburel:
I think we can merge the 2 but I wanted to open this one for validation first. It is going to conflict with https://github.com/ome/openmicroscopy/pull/6315

sbesson:
So the priority order is review and ideally merge #6315 first then focus on this one to enhance the GH workflow?

jburel:
Merge #6315 and #6317 then apply the logic introduced in this PR on top. This should be straightforward.

jburel:
Replaced by https://github.com/ome/openmicroscopy/pull/6318

PR #6322, opened 2022-05-30
bump zarr reader version

jburel:
# What this PR does

bump zarr reader version


joshmoore (commented):
MSTM. Let's see what tomorrows build looks like though. Raises the perennial question of how/where to add tests for OME-Zarr imports into OMERO.

PR #6325, opened 2022-06-10
remove slice2java

jburel:
Remove slice2java.
Check that the build is green

### DIFF
--- a/components/antlib/scripts/parse_version
+++ b/components/antlib/scripts/parse_version
@@ -82,7 +82,8 @@ def snapshot_logic(omero_version):
     if m:
         # Non-tag version
         base = incr_version(m.group("BASE"))
-        omero_version = "%s-SNAPSHOT" % base
+        omero_version = "%s-SNAPSHOT" % (
+            base, m.group("ICE"))
     else:
         return omero_version
     return omero_version
--- a/components/antlib/scripts/parse_version
+++ b/components/antlib/scripts/parse_version
@@ -82,8 +82,7 @@ def snapshot_logic(omero_version):
     if m:
         # Non-tag version
         base = incr_version(m.group("BASE"))
-        omero_version = "%s%s-SNAPSHOT" % (
-            base, m.group("ICE"))
+        omero_version = "%s-SNAPSHOT" % base
     else:
         return omero_version
     return omero_version
--- a/components/antlib/scripts/parse_version
+++ b/components/antlib/scripts/parse_version
@@ -82,7 +82,7 @@ def snapshot_logic(omero_version):
     if m:
         # Non-tag version
         base = incr_version(m.group("BASE"))
-        omero_version = "%s-SNAPSHOT" % (
+        omero_version = "%s%s-SNAPSHOT" % (
             base, m.group("ICE"))
     else:
         return omero_version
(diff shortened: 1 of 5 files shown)


sbesson (changes requested):
Although the build passes without issue with this PR included, both in the GitHub Actions context as well as locally (without slice2java installed), the server failed to start - see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-server/1234/

Looking quickly at the logs, the services did not get started. Looking at a few configuration files, il looks like a few variables did not get subsituted e.g. in `etc/omero.properties`:

```
omero.version=5.6.3-242-d748290-ice${versions.ice_lib}-b1224
```

jburel:
Since we are using only one ice version, that could be removed.

jburel:
Back to green https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-server/1242/ with this PR included

sbesson:
OMERO.script integration tests (Python + Java) have been failing for the last two days - see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/1137/. This is possibly related to these changes

jburel:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/1144/ is green with this PR included

sbesson:
In general, :+1: for getting rid of the unnecessary `slice2java` dependency for building. 

The biggest change looking at the nightly artifacts is the removal of the `-ice36` suffix from `omero.version` and hence the OMERO.server artifact name.

This raises essentially the same question as in https://github.com/ome/openmicroscopy/pull/6318#issuecomment-1113366461. Do we have an authoritative list of all the downstream locations that would be impacted by the change and will need to be reviewed? And do you have a feeling of the cost/time of updating these post server release?

joshmoore:
Briefly:
 * I don't have an idea of what impact dropping `-ice36` might have.
 * Is slice2py still needed?

jburel:
* omego and the ansible roles are the ones to review. 
I have rolled back the name change. This does not need to be included in that body of work.

jburel:
@joshmoore I think we can also drop ``slice2py`` to build 
I have run ``./build.py clean && ./build.py`` in a conda environment without slice2py and it works fine.
The question is: do we want to have env to build only and env to run only?

jburel:
The last 2 commits remove:
*  slice2py 
* ice-action

the build remains green

components/antlib/scripts/parse_version
```diff
@@ -82,7 +82,7 @@ def snapshot_logic(omero_version):
     if m:
         # Non-tag version
         base = incr_version(m.group("BASE"))
-        omero_version = "%s%s-SNAPSHOT" % (
```
sbesson: Should this change be reverted if the `-ice` separator is maintained?

jburel: well spotted

sbesson (approved):
Thanks, the latest version of this PR feels like the minimal set of changes allowing to drop the unnecessary Ice requirement at the level of this repository without modifying artifact names and downstream consumptions.

No objection to merging from my side

joshmoore (approved):
Also no objections. I imagine eventually, we strip out the rest of strip2cpp, etc. but happy to wait until we drop ant & ivy :smile:

PR #6329, opened 2022-06-29
update dependencies

jburel:
This PR introduces a new job that will be run daily to check if some ``ome`` dependencies have been released.
If this is the case, the version(s) in ``omero.properties`` will be updated and a PR will be open


sbesson (changes requested):
Conceptutally, this like a useful action which implements an automated version bump similar to what GitHub dependabot does for the purpose of our custom build system with multiple dependency sources.

Probably the biggest unknown is out to validate the logic. Would it be doable to extract all the update logic into a standalone script in the repository. This would allow to test the script in isolation but also for the GitHub action to consume it?

jburel:
I have added bash script and it is used by the GHA
Note that to run locally ``jq`` will need to be installed e.g. ``conda install -c conda-forge jq ``. Also ``-e`` flag is required on MacOS

sbesson (approved):
Tested locally by changing a few of the last versions declared in `omero.properties`. Running the `update_dependencies.sh` script restores the current content of the file.

Merging and we'll monitor whether the next upstream releases are automatically picked up

PR #6335, opened 2022-10-24
switch to new output command

jburel:
Check that the build is green
Background https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/source_build.yml
+++ b/.github/workflows/source_build.yml
@@ -29,8 +29,8 @@ jobs:
           git clone https://github.com/ome/bioformats /tmp/bioformats --depth 1
           cd /tmp/bioformats
           mvn install -DskipTests -q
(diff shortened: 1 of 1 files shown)


joshmoore (approved):
Thanks, @jburel. This certainly matches my expectations from:

 * https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/  and
 * https://github.com/datacarpentry/image-processing/pull/234/files 

I'll try to get through as many of these as I can before AL, but I assume that for all matching PRs, if they are green, they are good to go. :+1:

jburel:
I noticed a couple that did not turn green but it is unrelated to the change. I will look at them
For the other PRs, if they are green they can be merged

PR #6344, opened 2023-03-15
changelog 5.6.7

jburel:
Preparation for 5.6.7

### DIFF
--- a/history.rst
+++ b/history.rst
@@ -29,6 +29,7 @@ Improvements include:
 
 **Note:** This upgrade will invalidate the Bio-Formats Memoizer cache. Please
 see the upgrade guide for further information.
+We also recommend to re-index the database especially if the OMERO server has a large number of High Content Screening data.
 
 This version of the OMERO.server has been tested with:
 
--- a/history.rst
+++ b/history.rst
@@ -22,9 +22,9 @@ This release includes the following upgrade of the OMERO.server Java components:
 Improvements include:
 
 - address performance issues when indexing fileset
-- add ``omero.search.max_fileset_size`` parameter to indicated the maximum size of the fileset to be considered for indexing
-- run the PixelDataThread Application events in SYSTEM Thread pool.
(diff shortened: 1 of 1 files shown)


history.rst
```diff
@@ -5,6 +5,37 @@
...
+- omero-renderer 5.5.12
+- omero-romio 5.7.2
+- omero-common 5.6.1
+- omero-model 5.6.10
+
+
+Improvements include:
+
+- address performance issues when indexing fileset
+- add option to configure fileset indexing
```
sbesson: maybe let's name the property explicitly like in the previous releases

history.rst
```diff
@@ -5,6 +5,37 @@
...
+- run the PixelDataThread Application events in SYSTEM Thread pool.
+- Declare logback-classic as explicit dependency and set to 1.2.x.
+- an upgrade of Bio-Formats to version 6.12.0
+
+**Note:** This upgrade will invalidate the Bio-Formats Memoizer cache. Please
+see the upgrade guide for further information.
```
sbesson, line 31: In addition with the memoizer cache regeneration, the only other point possibly worth discussing is that people might want to consider reindexing their database. 

The choice will be obviously domain/application-specific and is complex to communicate. It would be the case in the case of an instance with filesets with very large number of files e.g. an HCS OMERO where the indexing is likely to be unable to catch up. It could also be relevant in an instance where at least some of the filesets exceed the max size specified by the property and it is important to have consistency in terms of the search results between data indexed prior to and after the upgrade.

sbesson (changes requested):
A few comments and primarily a question about communicating the optional reindexing requirement.

While looking at the underlying changes, I found that the `master` branch of `omero-gateway-java` does not include the v5.7.2 tag.

history.rst
jburel: @sbesson adjustments made

sbesson (approved):

PR #6353, opened 2023-07-20
remove dependencies

jburel:
test classes included in omero-common
see https://github.com/ome/omero-common/pull/42

### DIFF
--- a/components/tools/OmeroJava/ivy.xml
+++ b/components/tools/OmeroJava/ivy.xml
@@ -27,7 +27,7 @@
 	    <artifact name="omero-blitz" type="zip" ext="zip" e:classifier="python"/>
     </dependency>
     <dependency org="org.openmicroscopy" name="omero-gateway" rev="${versions.omero-gateway}"/>
-    <dependency org="org.openmicroscopy" name="omero-common-test" rev="${versions.omero-common-test}"/>
+    <!--<dependency org="org.openmicroscopy" name="omero-common-test" rev="${versions.omero-common-test}"/>-->
   </dependencies>
 </ivy-module>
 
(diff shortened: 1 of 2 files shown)


components/tools/OmeroJava/ivy.xml
```diff
@@ -27,7 +27,7 @@
 	    <artifact name="omero-blitz" type="zip" ext="zip" e:classifier="python"/>
     </dependency>
     <dependency org="org.openmicroscopy" name="omero-gateway" rev="${versions.omero-gateway}"/>
-    <dependency org="org.openmicroscopy" name="omero-common-test" rev="${versions.omero-common-test}"/>
+    <!--<dependency org="org.openmicroscopy" name="omero-common-test" rev="${versions.omero-common-test}"/>-->
```
sbesson, line 30: Proposing to remove these lines rather than commenting them as there is no plan to reintroduce them

components/tools/OmeroJava/test.xml
```diff
@@ -11,7 +11,7 @@
   </publications>
   <dependencies defaultconfmapping="test->*">
     <dependency org="org.openmicroscopy" name="omero-blitz" rev="${versions.omero-blitz}"/>
-    <dependency org="org.openmicroscopy" name="omero-common" rev="${versions.omero-common-test}"/>
+    <!--<dependency org="org.openmicroscopy" name="omero-common" rev="${versions.omero-common-test}"/>-->
```
sbesson, line 14: Same comment as above

sbesson (changes requested):
As indicated in the description, this depends on https://github.com/ome/omero-common/pull/42 being included transitively into a set of releases of the `omero-*` server components.

In addition to the inline comments, I found a few other references to `omero-common-test` that should be cleaned up as part of this work
- `etc/omero.properties`
- `components/tools/OmeroJava/build.gradle`
- `update_dependencies.sh`

PR #6354, opened 2023-07-26
Update import to support Django 4

knabar:
# What this PR does

Update `url` import to `re_path` to support Django 4

This fixes the integration test:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/(root)/(empty)/OmeroWeb_test_integration_test_login/

### DIFF
--- a/components/tools/OmeroWeb/test/integration/test_api_login.py
+++ b/components/tools/OmeroWeb/test/integration/test_api_login.py
@@ -24,7 +24,6 @@ Tests logging in with webgateway json api
 import pytest
 from omeroweb.testlib import IWebTest, get_json, _response, post
 from django.urls import reverse, NoReverseMatch
-from django.middleware import csrf
 from omeroweb.api import api_settings
 from django.test import Client
 from omero_marshal import OME_SCHEMA_URL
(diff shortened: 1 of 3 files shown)


sbesson:
Barring the `flake8` issue the change is inline with all the Django 4 updates in the OMERO.web apps.

sbesson (approved):
Assuming tests pass with Django 4.2

sbesson:
Looks like OMERO.web integration tests were executed using Django 4.2. However https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/175/ flags a few failures

- https://github.com/ome/openmicroscopy/blob/f69811eae3039b839f8faba4be66155d9a263828/components/tools/OmeroWeb/test/integration/test_api_login.py#L167 uses an internal method which has been renamed as `_does_token_match` in Django 4.x. This also raises the question of whether we want to can avoid depending on a private API
- https://github.com/ome/openmicroscopy/blob/f69811eae3039b839f8faba4be66155d9a263828/components/tools/OmeroWeb/test/integration/test_config.py#L77 the constructor now requires a mandatory `get_response` argument - https://docs.djangoproject.com/en/4.2/releases/4.0/#features-removed-in-4-0

knabar:
@sbesson Trying to see if passing in a dummy response function will fix the issue with the `SessionMiddleware`.

Not sure how to best address the use of the internal function.

sbesson:
The usage of the internal API was introduced in https://github.com/ome/openmicroscopy/pull/6165/ and already updated as the method got renamed in Django 3.2 in https://github.com/ome/openmicroscopy/pull/6304. @jburel @will-moore: any recommendation?

jburel:
I could not find at the time a better way to compare the tokens other that using an internal function
Maybe there is an option now.

knabar:
Reviewing the Django code, the method has not only been renamed, but also behaves differently now: https://github.com/django/django/commit/5d80843ebc5376d00f98bf2a6aadbada4c29365c

Previously both tokens passed in were assumed to be masked, now the second token is assumed to be unmasked already, with the first optionally being masked.

This will require more than just renaming the function.

knabar:
The three options I can think of right now:

0. Remove the `assert`, since an invalid CSRF token should break other tests anyway

0. Use an instance of `CsrfViewMiddleware` instead of just the internal method, which may require significant amounts of work to mock up the environment needed for the middleware.

0. Write our own CSRF check function, based on the previous or current internal code in Django.  This might become more work as the Django CSRF implementation changes, to make sure we test the right things.

@jburel @will-moore

jburel:
@knabar as discussed today. Let's go for option 0 for now so the other tests can be reviewed

PR #6356, opened 2023-07-30
Fixes flake8

jburel:
This PR fixes flake8 issue reported after merging https://github.com/ome/openmicroscopy/pull/6355
Note that without this PR, no need PR will be green


jburel:
Merging so the "Scheduled" action will get back to green

PR #6361, opened 2023-09-12
Add int test for getOriginalMetadata method

dominikl:
Add integration test for https://github.com/ome/omero-gateway-java/pull/72 .
Simply checks if the method returns something.


jburel:
Build failure https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-build/306/

dominikl:
Sorry. Should be fixed now. How could the previous commit get a green tick, if it's already a compile error?

jburel:
We will need to check, one job might exclude the tests?

dominikl:
Closing again. Don't know how to test. I guess the artifical test images created for the integration test don't have any "original metadata", hence the method returns just an empty string for them.

sbesson:
If the images created for the integration tests use the API to create the objects and then write the planes using the PixelsService, then there will indeed be no original metadata.

To test this API, you'll likely need to import an image. You should be able to use the `.fake` format which allows to create original metadata key/value pairs programmatically for testing - https://bio-formats.readthedocs.io/en/stable/developers/generating-test-images.html

PR #6369, opened 2023-10-27
Remove all usages of mox3 in OMERO.py/OMERO.dropbox integration tests

sbesson:
Due to compatibility issues with Python 3.11, these are replaced by pytest-mock which wraps the standard library mock infrastructure.

See also https://github.com/ome/omero-py/pull/385 and https://github.com/ome/omero-dropbox/pull/22 which contain similar changes to the unit tests

Initially opening as a draft as I think the `test_requires` changes are ignored by the current logic of the Jenkins OMERO integration jobs - see https://github.com/ome/devspace/blob/master/home/jobs/OMERO-test-integration/config.xml#L26

### DIFF
--- a/components/tools/OmeroFS/test/drivers.py
+++ b/components/tools/OmeroFS/test/drivers.py
@@ -29,7 +29,6 @@ except ImportError:
     # Python 2
     from path import path
 from omero.util import ServerContext
-from mox3 import mox
 from functools import wraps
 from omero.util.temp_files import create_path
 from fsDropBoxMonitorClient import MonitorClientI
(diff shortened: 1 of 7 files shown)


jburel:
i have adjusted both ``OMERO-training`` and ``OMERO-test-integration`` jobs on ``merge-ci``

jburel:
ZarrReader should now be re-activated.

sbesson:
> ZarrReader should now be re-activated.

[374195d](https://github.com/ome/openmicroscopy/pull/6369/commits/374195d03a9277820df6c4d986a49478731cc678) force pushed away

jburel (approved):
Build is green and so are the test on merge-ci

PR #6381, opened 2024-04-02
Session creation

jburel:
Add test to support https://github.com/ome/omero-py/pull/400


jburel:
The omero-py PR has been merged and the tests are green.
Merging

PR #6382, opened 2024-04-15
tmp

jburel:
Replace ``omero-scripts`` to ``omero_scripts`` so other components can be tested

### DIFF
--- a/components/tools/build.xml
+++ b/components/tools/build.xml
@@ -37,20 +37,20 @@
 		</echo>
 		<mkdir dir="target/lib/server"/>
 		<!-- Look for file above src -->
-		<property name="filegz" value="../../../omero-scripts-${versions.omero-scripts}.tar.gz" />
+		<property name="filegz" value="../../../omero_scripts-${versions.omero-scripts}.tar.gz" />
 		<if>
 			<available file="${filegz}"/>
             <then>
                <copy file="${filegz}" todir="target/downloads/scripts"/>
             </then>
 			<else>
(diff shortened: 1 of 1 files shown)


sbesson:
Definitely related to the changes in https://setuptools.pypa.io/en/latest/history.html#v69-3-0 to canonicalizing the package name in source distribution. As shown by the failure, this causes problem with the latest release of `omero-scripts` (https://pypi.org/project/omero-scripts/5.8.0/#files) which still uses the previous form for the package name.

Capping `setuptools` does not feel like a reliable solution as the next release of `omero-scripts` will most likely use the new naming scheme.

jburel:
The same applies to all our Python packages
The only one not affected is ``omero_marshal``. Name has always been like that


We can 
* either rename all the packages and release a new version
 * or add a ``pyproject.toml``  
 ```
[build-system]
requires = ["setuptools < 69.3.0"]
```
to make sure the name is kept

sbesson:
> We can
> - either rename all the package and release a new version
> - or add a pyproject.toml

Let's discuss options tomorrow. At least from my side, capping `setuptools` by whatever mechanism can only be considered as a short term fix.

> The same applies to all our Python packages

But within the context of this repository, the only Python artifact downloaded and extracted via filename is `omero-scripts`, no? I understand there is are wider implications for the rest of the OME CI infrastructure. However looking at https://github.com/ome/devspace/blob/e91bc0e1420da617f05eb93d827af00e60bfe663/home/jobs/OMERO-server/config.xml#L140-L142, the usage of wildcards might make the logic more robust to the latest `setuptools` (breaking) changes.

jburel:
It is.
I was thinking of a wider fix

PR #6384, opened 2024-04-18
Scripts update

jburel:
Due to a recent change in setuptools cf.  https://setuptools.pypa.io/en/latest/history.html#v69-3-0, 
The name of the ``tar.gz`` is now ``omero_scripts`` instead of `omero-scripts``
The change of name only happens when ``python -m build`` is used (This is currently the command used in the CI) but not when ``python setup.py sdist`` is used.

### DIFF
--- a/components/tools/build.xml
+++ b/components/tools/build.xml
@@ -37,20 +37,20 @@
 		</echo>
 		<mkdir dir="target/lib/server"/>
 		<!-- Look for file above src -->
-		<property name="filegz" value="../../../omero-scripts-${versions.omero-scripts}.tar.gz" />
+		<property name="filegz" value="../../../omero_scripts-${versions.omero-scripts}.tar.gz" />
 		<if>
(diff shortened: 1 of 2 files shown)


sbesson (approved):
> The change of name only happens when `python -m build` is used (This is currently the command used in the CI) but not when ``python setup.py sdist`` is used.

Thanks, I had missed this aspect and that explains a few things. The issues seen in the OME CI environment were caused by the fact the Jenkinsfile logic had been updated in https://github.com/ome/omero-python-superbuild/pull/258 to use `python -m build` while the GitHub workflows publishing the release were still using `python setup.py sdist`.

Assuming both workflows are updated to be consistent, this looks good to include but it would be useful to have a validation in the nightly CI builds and https://merge-ci.openmicroscopy.org/jenkins seems to be unreachable this morning

jburel:
Docker was down again. Restarted now
I will check with IT since the same happened last Friday

PR #6389, opened 2024-05-28
Handle archive status being populated (See ome/omero-web#555)

chris-allan:
Adjust test cases for ome/omero-web#555.


sbesson (approved):
My understanding of the status of https://github.com/ome/omero-web/pull/555 is that the only remaining todo might be the icon to use for archived images. The addition of the `archived` status to the marshalled response is uncontroversial and this PR adjusts the integration tests accordingly

PR #6391, opened 2024-06-10
Use cookie to get csrf token in JSON API Login.py

will-moore:
# What this PR does

Fixing this example as part of question at https://forum.image.sc/t/testing-omero-json-api-with-curl-token-login/97400

# Testing this PR

1. Run `python Login.py` with the correct credentials (edit the constants in the file or use `Parse_OMERO_Properties`)

2. Should be able to login to the specified server and should list Projects etc.


dominikl (approved):
Thanks Will. Looks good. With these changes the scripts works again 👍

PR #6393, opened 2024-06-19
Add documentation for the omero.logging properties

### LINKED ISSUE #6225: Document Python logging properties
Add these `omero.logging` properties: https://github.com/ome/openmicroscopy/blob/v5.6.0/etc/templates/grid/templates.xml#L86-L90

to https://docs.openmicroscopy.org/omero/5.6.0/sysadmins/config.html

sbesson:
Initial step towards fixing #6225 

All these properties are defined in OMERO.py so changing the default value in this file has no effect but allows for the configuration to be included in the reference documentation

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -147,7 +147,7 @@ omero.logging.logsize=5000000
 omero.logging.lognum=9
 
 ## (For documentation only)
-# Specifies the threshold for the server log files. Logging messages
+# Specifies the threshold for the Python servers log files. Logging messages
 # which are less severe than this value will be ignored. For a list of
 # available values, see https://docs.python.org/3/library/logging.html#levels.
 omero.logging.level=20
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -132,16 +132,16 @@ omero.logging.directory=var/log/
 # interval. If true, log files will be rolled over at midnight.
 # Otherwise, log files will be rolled over at a predetermined size,
 # see omero.logging.logsize for more information.
-omero.logging.timedlog=False
+omero.logging.timedlog=false
 
 ## (For documentation only)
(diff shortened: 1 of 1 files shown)


will-moore:
Looks good. Text makes sense.

So this PR is purely to update the docs at https://omero.readthedocs.io/en/latest/sysadmins/config.html?
Trivial point: I noticed there that all booleans tend to be lowercase `false` whereas you have a mix (although I think it doesn't matter since they are all lowercased when comparing, but just for consistency).

sbesson:
> So this PR is purely to update the docs at https://omero.readthedocs.io/en/latest/sysadmins/config.html?

Absolutely, the default logging values are defined in https://github.com/ome/omero-py/blob/8d8108de02104c654741d65f0ecbbf11c85e329d/src/omero/util/__init__.py#L32-L38 so modifying `etc/omero.properties` has no functional effect.

But it is currently undocumented that the server logs can be configured using the standard `omero config` approach and this PR aims to update the documentation to include these properties.

> Trivial point: I noticed there that all booleans tend to be lowercase false whereas you have a mix (although I think it doesn't matter since they are all lowercased when comparing, but just for consistency).

Pushed b73a17c5f0e8f7b3f8f6d3b036152ce73fd0638f to use the current lowercase convention consistently

will-moore (approved):
Thanks

sbesson:
While using these properties as part of the review of https://github.com/ome/omero-py/pull/417, one important aspect worth noting is that the `omero.logging.level` property only affects the Python servers (like `Processor` or `Tables`). This is consistent with what is described in https://omero.readthedocs.io/en/stable/developers/logging.html.

The last commit should update the property description to be explicit about this.

joshmoore (approved):

PR #6395, opened 2024-07-05
Update IceGrid templates to make the OMERO.tables module configurable

sbesson:
Add a new top-level TABLES_MODULE variable used by the Tables template to start the new server.
Functionally, this will need to be tested alongside a corresponding OMERO.py with the logic to replace @omero.tables.module@ by either the value of the omero.tables.module configuration or the default runTables value.

Together with https://github.com/ome/omero-py/pull/419, the following scenarios should be tested

- the default behavior should be unchanged i.e. starting the server either via `omero admin start` or the systemd service file should rewrite the IceGrid templates using the default `runTables` value and start the `Tables` service
- setting `omero config set omero.tables.module runTables` in the OMERO configuration and restarting the server should have an identical behavior
- setting `omero config set omero.tables.module notexistingTables` in the OMERO configuration and restarting the server should not start the `Tables` service and `master.err` should contain a line similar to `/opt/omero/OMERO.venv/bin/python: No module named notexistingTables`

### DIFF
--- a/etc/omero.properties
+++ b/etc/omero.properties
@@ -168,7 +168,9 @@ omero.server.nodedescriptors=
 #############################################
 _omero.glacier2.IceSSL.Ciphers.darwin=(AES)
 
-
+## (For documentation only)
+# Name of the Python module to use when starting the Tables service
+omero.tables.module=runTables
 
 #############################################
 ## Server product name for release artifacts
--- a/etc/templates/grid/default.xml
+++ b/etc/templates/grid/default.xml
@@ -21,6 +21,7 @@
     <variable name="ROUTERPORT"    value="@omero.ports.prefix@@omero.ports.ssl@"/>
     <variable name="INSECUREROUTER" value="OMERO.Glacier2/router:tcp -p @omero.ports.prefix@@omero.ports.tcp@ -h @omero.host@"/>
     <variable name="SEP"           value=":"/>
+    <variable name="TABLES_MODULE" value="@omero.tables.module@"/>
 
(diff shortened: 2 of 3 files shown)


chris-allan (approved):

sbesson:
Probably the outstanding question is whether this mechanism is in scope for the upcoming OMERO.server 5.6.12 @jburel?

If so, an initial documentation step would be to add the property to `etc/omero.properties` in this repository and add a new `Tables` header similar to https://github.com/ome/omero-py/pull/416 to register the new property in the configuration glossary of the reference OMERO documentation at https://omero.readthedocs.io/en/latest/sysadmins/config.html

jburel:
I think it will make sense to include this strategy in the upcoming release

sbesson:
As discussed this morning with @knabar and @jburel, OMERO.py 5.19.4 has just been released with the underlying infrastructure allowing to replace the token in the templates. 

I quickly tested the behavior of a server restart with the templates modified as per with this PR with OMERO.py 5.19.3 (or earlier). In that case, the server components will start as expected except for the Tables service which will be marked as `inactive`. In addition `master.err` should include an error similar to:

```
/opt/omero/OMERO.venv/bin/python: Error while finding module specification for '@omero.tables.module@' (ModuleNotFoundError: No module named '@omero')
```

I assume the release notes of OMERO 5.6.12 should document the requirement to upgrade OMERO.py as part of the server upgrade but let me know if the above should also be captured e.g. in the troubleshooting section of the documentation.

PR #6396, opened 2024-07-08
Add a test case for ome/omero-py#418

chris-allan:
# What this PR does

Add a test case for ome/omero-py#418

### DIFF
--- a/components/tools/OmeroPy/test/integration/tablestest/test_service.py
+++ b/components/tools/OmeroPy/test/integration/tablestest/test_service.py
@@ -714,7 +714,7 @@ class TestTables(ITest):
         table.delete()
         table.close()
 
-    def tesIncludeRowNumbers(self):
+    def testIncludeRowNumbers(self):
         grid = self.client.sf.sharedResources()
         table = grid.newTable(1, "/testIncludeRowNumbers")
         assert table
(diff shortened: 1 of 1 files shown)


components/tools/OmeroPy/test/integration/tablestest/test_service.py
```diff
@@ -714,4 +714,40 @@ def testColumnSpaceNames(self):
         table.delete()
         table.close()
 
+    def tesIncludeRowNumbers(self):
```
sbesson:
```suggestion
    def testIncludeRowNumbers(self):
```
is likely required to be picked up by `pytest`

sbesson (changes requested):
CI execution has been failing overnight due to unrelated infrastructure issues. One syntactic issue to allow the test to be picked up by `pytest`.

Only other wonder if whether we also want to expand the test semantics to ensure that the output

```
table.read(colNumbers, start, stop, context)
table.readCoordinates(rowNumbers, context)
```

also respect the new `omero.tables.include_row_numbers` key/value pair

will-moore:
Code looks good, apart from typo mentioned above.

jburel:
test is green, upstream PR has now been. Requested changes were made to allow the test to be picked up

sbesson (approved):
All three new integration tests were included in the nightly CI builds, executed successfully and should cover all OMERO.tables API that have support for the new boolean

PR #6397, opened 2024-07-19
Tests for get_where_list and slice table API calls

knabar:
Add tests for new API calls from https://github.com/ome/omero-web/pull/564


jburel:
One remaining failure https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/

knabar:
Thanks @jburel - this one means that the different backends handle row overflow differently (silently ignored vs. disallowed), so the API call should check for that as well.

Will make the change to the API and to this test.

sbesson (approved):

PR #6402, opened 2024-07-26
Supplement TestTables.testColumnSpaceNames with advanced queries

sbesson:
Variable mapping passed to getWhereList can be used both for the column names and the values. The current test logic was focusing on the checking the replacement of column names with spaces. This commit adds a few additional advanced scenarios where the integer value used for the condition is also mapped using the supplied dictionary

Testing-wise, the test should keep passing using the current PyTables OMERO.tables implementation, there are no associated functional changes.

### DIFF
--- a/components/tools/OmeroPy/test/integration/tablestest/test_service.py
+++ b/components/tools/OmeroPy/test/integration/tablestest/test_service.py
@@ -702,14 +702,32 @@ class TestTables(ITest):
         table.addData(cols)
 
         assert [1, 2, 3] == table.getWhereList(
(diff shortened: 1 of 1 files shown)


joshmoore (approved):

chris-allan (approved):

chris-allan:
@jburel: Integration tests passing with these additional criteria. Anything else we should do prior to merging?

jburel:
No i have a round of PRs to merge so they are included in the release. i will go over that now that OMERO.web is ready to go

chris-allan:
Sounds good.

Issue #6405, opened 2024-08-06
Distribution contents regression with Ivy 2.5.2


chris-allan:
After the release of 5.6.12 it was reported [^1] that old artifacts, some with critical security vulnerabilities such log4j 1.2.17, were ending up in the OMERO.server distribution. This appears to an unintended consequence of #6388. The diff of JARs from v5.6.12 with 94c9b166741ca01ebf2bffcdb21c43f10bb93b6a reverted is as follows:

```
--- 5_6_12_server_jars  2024-08-06 09:43:14.683495565 +0000
+++ latest_jars 2024-08-06 09:42:52.459392636 +0000
@@ -40,8 +40,6 @@
 batik-xml.jar
 btm.jar
 bufr.jar
-byte-buddy-agent.jar
-byte-buddy.jar
 c3p0.jar
 calcite-core.jar
 calcite-linq4j.jar
@@ -102,7 +100,6 @@
 jblosc.jar
 jcip-annotations.jar
 jcl-over-slf4j.jar
-jcommander.jar
 jgoodies-common.jar
 jgoodies-forms.jar
 jhdf5.jar
@@ -118,7 +115,6 @@
 jzarr.jar
 kryo.jar
 listenablefuture.jar
-log4j.jar
 logback-classic.jar
 logback-core.jar
 lucene-analyzers.jar
@@ -139,8 +135,6 @@
 metrics-logback.jar
 minio.jar
 minlog.jar
-mockito-core.jar
-mockito-inline.jar
 native-lib-loader.jar
 objenesis.jar
 okhttp.jar
@@ -169,8 +163,6 @@
 serializer.jar
 sketches-core.jar
 slf4j-api.jar
-slf4j-jdk14.jar
-slf4j-log4j12.jar
 snakeyaml.jar
 solr-analysis-extras.jar
 solr-commons-csv.jar
@@ -193,7 +185,6 @@
 spring-tx.jar
 sqlite-jdbc.jar
 subethasmtp.jar
-testng.jar
 turbojpeg.jar
 udunits.jar
 xalan.jar
```

My running theory is this is due to changes in dependency resolution that result in artifacts from testing targets leaking into the runtime ones.

[^1]: https://forum.image.sc/t/critical-old-and-unsafe-log4j-jar-file-in-omero-5-6-12/100016

chris-allan:
Diff of the client JARs:

```
[...]

PR #6409, opened 2024-08-21
Bump jul-to-slf4j and jcl-over-slf4j to version 2.0.16

sbesson:
Motivated primarily by the regression in #6405 and https://github.com/ome/openmicroscopy/issues/6405#issuecomment-2283339623, this bumps the version of `jul-to-slf4j` and `jcl-over-slf4j` to the latest release of the 2.0.x series, currently 2.0.16. These should be compatible with the version of `slf4j-api` which is now shipped in OMERO.server.

As an additional feature, the latest version of `jul-to-slf4j` now depends on `reload4j` (rather than `log4j`) for testing purposes. This does not resolve the dependency problems raised in #6405 but it further reduces the chances of `log4j.jar` unexpectedly creeping into the binaries.

Testing should confirm that logging statements using either JCL (Apache Commons Logging, formerly Jakarta Commons Logging) or JUL (java.util.logging) as still handled via `slf4` and ultimately `logback`.

Unfortunately the biggest usage of these logging frameworks will be in third-party dependencies since the majority of the OMERO code base has been converted to using `slf4j` and `logback` over a decade ago - https://github.com/ome/openmicroscopy/pull/1006.

Within OMERO, there are still a few classes making internal use of  either JCL or JUL:
- the JVM checks in [ome.service.util.JvmSettingsCheck](https://github.com/ome/omero-server/blob/8eb311d45642163b79463c38c4b06884a637923d/src/main/java/ome/services/util/JvmSettingsCheck.java#L79) are using JCL -L85. These should be easy to test as they are logged in `Blitz-0.log` at the server startup
- `ome.util.Utils` uses JUL in its [closeQuietly](https://github.com/ome/omero-model/blob/7e89fbd253f75213d1c3c0dc9fb202ecf3296b6c/src/main/java/ome/util/Utils.java#L158-L168) API, `Utils.closeQuietly` is used in a few components to close input streams however the logging message will only appear if `close()` throws an exception so I can't think of a good scenario off-hand
- [omero.util.Resources](https://github.com/ome/omero-blitz/blob/bf9a738b0417973de01be3b2f7ccf45ee2eacd19/src/main/java/omero/util/Resources.java) uses JUL internally when adding/removing/cleaning managed entries. Here again most logging statement are at the FINEST (aka TRACE) level and only unexpected behaviors are logged at a WARNING level. This class is used in `omero.client` and the import code base primarily when calling keepAlive - https://github.com/search?q=org%3Aome+omero.util.Resources+language%3AJava&type=code&l=Java

As a follow-up all the classes above could probably be updated to use the `slf4j` API and get rid of JCL and JUL at least in the OME codebase. I'll hold on opening these until this has been tested as it could make the testing more complicated.

### DIFF
--- a/ivy.xml
+++ b/ivy.xml
@@ -11,8 +11,8 @@
   <dependencies>
     <dependency org="org.openmicroscopy" name="omero-blitz" rev="${versions.omero-blitz}"/>
     <dependency org="org.openmicroscopy" name="omero-gateway" rev="${versions.omero-gateway}"/>
-    <dependency org="org.slf4j" name="jul-to-slf4j" rev="1.7.30"/>
-    <dependency org="org.slf4j" name="jcl-over-slf4j" rev="1.7.30"/>
+    <dependency org="org.slf4j" name="jul-to-slf4j" rev="2.0.16"/>
+    <dependency org="org.slf4j" name="jcl-over-slf4j" rev="2.0.16"/>
     <!-- runtime dependencies from dsl/ivy.xml -->
     <dependency org="janino" name="janino" rev="${versions.janino}"/>
     <!-- Useful for globally overriding the Bio-Formats version, empty version is ignored by default -->


jburel:
@sbesson Do you need/have a ticket to capture the work to follow-up?

sbesson:
> @sbesson Do you need/have a ticket to capture the work to follow-up?

I have not done anything as I was waiting for feedback on the proposed changes first. Happy to capture as issues if this work would be up for consideration. In that case, would it be preferable to have a single issue in this repository or three issues in `omero-model`, `omero-server` and `omero-blitz`?

jburel:
A single issue will make sense.

jburel (approved):

sbesson:
> A single issue will make sense.

Opened as https://github.com/ome/openmicroscopy/issues/6416

PR #6411, opened 2024-09-10
Check query without matches

knabar:
# What this PR does

Adding a test to verify the behavior of the bitmask table query API call when a query is given that does not match any rows in the table.

### DIFF
--- a/components/tools/OmeroWeb/test/integration/test_table.py
+++ b/components/tools/OmeroWeb/test/integration/test_table.py
@@ -195,6 +195,8 @@ class TestOmeroTables(IWebTest):
         ['query=SmallNumbers>0.5&col_name=IDs', "0000101"],
         # a query that does not match anything
         ['query=SmallNumbers>100', "0000000"],
+        # a query that matches everything
+        ['query=SmallNumbers>0&col_name=Well', "011111"],
     ])
     def test_table_bitmask(self, omero_table_file,
                            django_client, query_result):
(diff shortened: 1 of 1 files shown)


will-moore:
It looks like this test is failing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_table/TestOmeroTables/test_table_bitmask_query_result4_/

knabar:
@chris-allan While looking into a bug we first thought to be in Pageant, @kyleBalisWest and I tracked down this issue in the bitmask API call.  If there are no matches, all rows are returned as matching.  I think the issue is that if `slice` is called without any rows, it returns everything.

With the coming updates I don't think the bitmask call is used anymore in any of our projects.

sbesson (changes requested):
See https://github.com/ome/omero-web/pull/580 for the proposed fix in OMERO.web which should fix the nightly CI tests.

For completeness, I would also suggest to add a scenario where a query returns all rows.

chris-allan:
Just did a little sleuthing to realize that this is a regression coming from ome/omero-web#554. Previously, when no "hits" were found even though a slice was being performed no paging would take place so the rows returned from `perform_table_query()` would have been empty.

chris-allan:
Also, this read everything when the indexes are empty behavior is also not consistent with PyTables itself but rather "special" as it relates to the `omero.column` implementations. Specifically, `omero.column.AbstractColumn.readCoordinates()`.

@sbesson: Once we've settled on the test cases here we should update the OMERO.tables documentation so this is clear; neither `read()` or `readCoordinates()` behave like this.

sbesson (approved):
New tests are now passing as of https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/167/

PR #6423, opened 2025-03-13
Use \d

jburel:
Upgrading the GitHub action to run on Ubuntu 24.04 with Python 3.12 to synchronize the documentation from various repositories has led to some failures
See https://github.com/ome/omero-documentation/actions/runs/13825561909

This is the first part of the puzzle.
Changes are required in other repositories e.g. omeroweb-install and ansible

### DIFF
--- a/components/antlib/scripts/parse_version
+++ b/components/antlib/scripts/parse_version
@@ -73,7 +73,7 @@ def snapshot_logic(omero_version):
     m = re.match((
         "^"
         "(?P<BASE>.*?)"
-        "(?P<STRIP>([-]DEV)?-\d+-[a-f0-9]+?(-dirty)?)"
+        "(?P<STRIP>([-]DEV)?-\\d+-[a-f0-9]+?(-dirty)?)"
         "(?P<ICE>-ice[0-9]+)?"
         "$"), omero_version)
     if m:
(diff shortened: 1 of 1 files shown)


joshmoore:
Interesting. What changed to make this start failing, @jburel ?

jburel:
```
documentation/OMERO.server/components/antlib/scripts/parse_version:76: SyntaxWarning: invalid escape sequence '\d'   "(?P<STRIP>([-]DEV)?-\d+-[a-f0-9]+?(-dirty)?)" /home/runner/work/omero-documentation/OMERO.server/components/antlib/scripts/parse_version:118: SyntaxWarning: invalid escape sequence '\d'   m = re.match("^([^\d]*\d+[.]\d+[.])(\d+)-SNAPSHOT$", omero_version) 5.6.14-ice36-SNAPSHOT). Use forcedeliver or update.
```
This is due to the fact that we are running with 3.12

joshmoore:
> This is due to the fact that we are running with 3.12

Gotcha.

sbesson (commented):
This matches the language changes announced in https://docs.python.org/dev/whatsnew/3.12.html#other-language-changes

I would suggest we follow the official recommendation and use raw strings for regular expressions

components/antlib/scripts/parse_version
```diff
@@ -70,7 +70,8 @@ def snapshot_logic(omero_version):
 
     See gh-67 for the discussion.
     """
-    m = re.match((
+    re.compile()
```
sbesson, line 73: Should this be reverted?
```suggestion
```

sbesson: @jburel converted my comment into a suggestion

sbesson (changes requested):
See https://github.com/ome/openmicroscopy/pull/6423/files#r2151360584

PR #6436, opened 2025-07-24
fix Training Read_Data.py filter by group

will-moore:
# What this PR does

Fixes code example, corresponding to change at https://github.com/ome/omero-documentation/pull/2496


sbesson (approved):
This is consistent with https://github.com/ome/omero-py/issues/466 and https://github.com/ome/omero-documentation/pull/2496

PR #6438, opened 2025-09-21
Update OMERO.py gateway test_wrapper integration test

sbesson:
Depends on the upstream OMERO.py changes proposed in https://github.com/ome/omero-py/pull/443


sbesson:
Merging this test-only change as the corresponding OMERO.py changes have been integrated a month ago.

Issue #6439, opened 2025-10-08
Handling /tmp partition mounted as noexec


sbesson:
# Background 

Widely recognized authorities including the [Center for Internet Security (CIS)](https://www.cisecurity.org/) and [NIST](https://www.nist.gov/) are regularly publishing and updating sets of security best practices for the configuration of IT systems like [CIS Benchmarks](https://www.cisecurity.org/cis-benchmarks) and [NIST 800-53](https://csrc.nist.gov/pubs/sp/800/53/r5/upd1/final). One of these recommendations is to have `/tmp` mounted as a separate partition with a number of mount options, including the noexec flag [^1] [^2] [^3] [^4] [^5].

Java programs can access native shared libraries using Java Native Access (JNA). Bio-Formats and OMERO make extensive use of this approach for the handling tiles compressed using JPEG, JPEG-XR or Blosc. Practically, this logic involves the creation of temporary files/folders where platform-specific libraries (e.g. `.so`, `.aarch64`) are extracted and loaded dynamically. By default, these files are created and executed from the Java temporary directory which points to `/tmp` on Unix systems. As a result, some of the reading functionalities of OMERO are directly impacted on systems where /tmp is mounted as a noexec partition. 

For end-users, this issue will result into tile loading failures [^6] [^7]. The application logs will include errors similar to:

```
java.lang.NoClassDefFoundError: Could not initialize class com.sun.jna.NativeLong
```

or

```
java.lang.UnsatisfiedLinkError: 'void org.libjpegturbo.turbojpeg.TJDecompressor.init()'
```

# Workaround

There are two approaches to fix Java usage of native shared libraries in an environment where `/tmp` is mounted as noexec:

1. modify the mount options of the `/tmp` partition. This option should be highly discouraged as it goes against security practices and institutional policies.
2. specify an alternate temporary directory by setting the `java.io.tmpdir` Java system property [^8].

In bioformats2raw, setting `java.io.tmpdir` appropriately is already the recommended workaround [^9]. Given the variety of environments and platforms from where the conversion utility can be invoked, it is the responsibility of the caller to choose a temporary directory with appropriate permissions.

[...]

PR #6441, opened 2025-10-30
Ice Grid: set java.io.tmpdir to var/tmp for Java servers

### LINKED ISSUE #484: omero admin: set java.io.tmpdir on the application startup
See https://github.com/ome/openmicroscopy/issues/6439

Update the CLI `admin` commands starting Java to:

1. check for the existence of` ${OMERODIR}/var/tmp` and create it if necessary
2. set `java.io.tmpdir` to `${OMERODIR}/var/tmp`

### LINKED ISSUE #6440: Create empty var/tmp directory as part of the server bundle
See https://github.com/ome/openmicroscopy/issues/6439

Allows various components of the OMERO ecosystem to use this well-known location under OMERO.server for temporary files used by the application

sbesson:
Fixes #6440 
Fixes https://github.com/ome/omero-py/issues/484

# What this PR does

This updates the Java servers started by Ice Grid to use `${OMERODIR}/var/tmp` as the temporary Java directory

- Create `var/tmp` directory at build time
- Define `OMERO_TMP` variable in the Ice Grid default configuration
- Use `OMERO_TMP` as the value of `-Djava.io.tmpdir` in Java services (Blitz, Indexer, Repository, PixelData)

2c0a18da0bcb5af220194d88f38e555faf3168d8 overrides the Java temporary directory for all Java servers. A reviewer question is whether it would be cautious to restrict this setting to the servers that are using Bio-Formats (`Blitz` & `PixelData` initially) since we have no concrete report of other functionalities affected by `/tmp` being mounted as `noexec`

# Testing this PR

Using the OMERO.server bundle built from this PR:

- start the application either via `omero admin start` or `systemctl start omero-server` 
  - check the applicatoin is properly initialised
  - check you can authenticate
- display the processes started by `icegridadmin` using `ps`
   -  check that `-Djava.io.tmpdir` is set to point at `var/tmp`.
- perform an operation requiring Bio-Formats ant the native library loading functionality e.g. import an NDPI file using one of the [public samples](https://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/)
  - check that a temporary native lib loader folder associated with the process is created under `var/tmp`
  - check the image is properly imported and the thumbnails are generated
  - check the image is indexed via `Indexer-0.log`
- stop the application either via `omero admin stop` or `systemctl stop omero-servr` 
  - check the temporary folder under `var/tmp` is cleaned up on restart

# Related reading

See #6439 for a full description of the rationale behind this change

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -365,6 +365,7 @@ To get started using Eclipse, execute "ant build-dev" and import the top-level
     <target name="create-workdirs" depends="init">
         <mkdir dir="${dist.dir}/etc/grid"/>
         <mkdir dir="${dist.dir}/var"/>
+        <mkdir dir="${dist.dir}/var/tmp"/>
     </target>
 
     <target name="copy-luts" depends="init">
--- a/etc/templates/grid/default.xml
+++ b/etc/templates/grid/default.xml
@@ -12,6 +12,7 @@
     <variable name="OMERO_ETC"     value="etc/"/>
     <variable name="OMERO_JARS"    value="lib/server/"/>
     <variable name="OMERO_LOGS"    value="var/log/"/>
+    <variable name="OMERO_TMP"     value="var/tmp/"/>
     <variable name="OMERO_LOGFILE" value="${OMERO_LOGS}$${omero.name}.log"/>
     <!-- Note: you will also need to modify the etc/master.cfg property
          file if you want var/log to be completely unused -->
--- a/etc/templates/grid/templates.xml
+++ b/etc/templates/grid/templates.xml
@@ -203,6 +203,7 @@
             <option>-agentlib:hprof=cpu=samples,cutoff=0,thread=y,interval=1,depth=50,force=y,file=${OMERO_LOGS}Blitz-${index}.hprof</option>
         </target>
         <option>-Djava.awt.headless=true</option>
+        <option>-Djava.io.tmpdir=${OMERO_TMP}</option>
         <option>-Dlogback.configurationFile=${OMERO_ETC}logback.xml</option>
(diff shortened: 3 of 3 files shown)


kkoz (approved):
Worked as expected

sbesson:
On the Glencoe side, we will apply this deployment change in the context of an upcoming round of Bio-Formats testing in OMERO Plus.

This change has been integrated into the nightly OME CI builds. The absence of failures suggest the usage of a local temporarty directory causes no outstanding regression in the behavior of the application. It would be useful to assign some reviewers from the OME development teams given the impact on all OMERO.server deployments. /cc @pwalczysko @joshmoore @jburel

joshmoore (approved):
In general, this looks good to me. I was a bit surprised that `OMERO_TMP` was not already defined, to be honest.

Regarding the question about limiting to Bio-Formats-related (Java) servers, I don't see a large risk. Or rather, I don't know of anything that would break on the use of var/tmp rather than /tmp. The only issue I can foresee is if large files like hdumps show up there, but probably better under var/ than randomly in /tmp which can be on the main mount point.

My only other thought would be trying to keep this inline with the behavior of the temp file logic added to Python which was likely originally based on the Java behavior.

sbesson:
> The only issue I can foresee is if large files like hdumps show up there, but probably better under var/ than randomly in /tmp which can be on the main mount point.

Definitely, we have also seen standalone `/tmp` partitions with low disk space (1-2G) which could easily be filled by large files and would affect other processes. Part of the motivation of using `/opt/omero` is that we can reasonably expect this mount point to be appropriately provisioned for the application including the logs, our default recommendation being at least 50G.

PR #6451, opened 2026-04-04
WIPUse beforeMethod instead of beforeClass

jburel:
Adjust the infra for gateway-test

### DIFF
--- a/components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
+++ b/components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
@@ -46,7 +46,7 @@ import omero.model.OriginalFile;
 import omero.model.FileAnnotation;
 import omero.sys.ParametersI;
 import org.testng.Assert;
-import org.testng.annotations.BeforeClass;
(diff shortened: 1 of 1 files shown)


components/tools/OmeroJava/test/integration/gateway/TablesFacilityTest.java
```diff
@@ -73,7 +73,7 @@ public class TablesFacilityTest extends GatewayTest {
     private final String ns = "some Namespace";
 
     @Override
-    @BeforeClass(alwaysRun = true)
+    @BeforeMethod(alwaysRun = true)
```
dominikl, line 76: Indeed that probably causes the integration test failure in `testGetAvailableTables`:
```
    @Test(dependsOnMethods = { "testAddTable" })
    public void testGetAvailableTables() throws Exception {
```
With the `@BeforeMethod` it probably removes the table again which was added by `testAddTable`. Is the intention of the PR to have completely independant test methods?

jburel: Since the PR did not have the expected result. Closing

PR #6456, opened 2026-05-28
Add image child count for expected tags

knabar:
This fixes tests for the changes in https://github.com/ome/omero-web/pull/676

The number of images linked to a tag is hardcoded since the fixtures are not currently set up in a way that the number of ImageAnnotationLinks can be easily determined from the fixture.

I tried to break up the creation of of the tagset and tag hierarchy in `tagset_hierarchy_userA_groupA`, but this quickly devolved into issues with objects not saving properly, etc. Note that there is also a tag annotation created for a different user that is not returned anywhere.

### DIFF
--- a/components/tools/OmeroWeb/test/integration/test_tree.py
+++ b/components/tools/OmeroWeb/test/integration/test_tree.py
@@ -242,7 +242,6 @@ def expected_tags(user, tags):
             'id': tag.id.val,
             'value': tag.textValue.val,
             'ownerId': tag.details.owner.id.val,
-            'childCount': len(tag.linkedAnnotationList()),
             'permsCss': get_perms(user, tag, "TagAnnotation")
         }
         if tag.description is not None:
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Sorry for the delay here, happy for this to go in alongside the accompanying OMERO.web PR which fixes a long-standing issue (count and paging in the tags view).

Reading the tests, am I correct in assuming part of the issue is that we have an overlap between several fixtures with the same `TagA`  value? If so, that could certainly be something we could look into fixing as I suspect the tests might depend on the test order. But this is not introduced by this change and is completely acceptable to review outside the scope of this PR

knabar:
> Reading the tests, am I correct in assuming part of the issue is that we have an overlap between several fixtures with the same `TagA` value? If so, that could certainly be something we could look into fixing as I suspect the tests might depend on the test order. But this is not introduced by this change and is completely acceptable to review outside the scope of this PR

The main issue I ran into was that changing the creation of some of the fixture objects, for example trying to create and save the tags in a separate fixtures, causes security violations when saving other dependent objects.  I'm sure this could be resolved, but outside the scope of this PR.

will-moore (approved):
All tests are passing here https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/158/testReport/junit/OmeroWeb.test.integration.test_tree/
LGTM 👍

PR #6460, opened 2026-07-02
Tests for new public URL filter setting

knabar:
# What this PR does

Adds tests for previous public URL configuration and a new setting introduced in https://github.com/ome/omero-web/pull/679


knabar:
Closing as the approach to handle `POST` requests has changed in https://github.com/ome/omero-web/pull/679

PR #6464, opened 2026-07-15
Fix test_pyramids.py little and big_endian

pwalczysko:
Fixes the [persistent failures](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.clitest.test_pyramids/TestRemovePyramidsFullAdmin/test_remove_pyramids_little_endian/) of the following 3 tests
-  ``test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_little_endian``.
-  ``test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_big_endian``
-  ``test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids``

# What this PR does

The test is creating pyramids first, relying on pyramids being written fast and then testing the removal. After move of the devspace testing framework to NFS, the speed of writing pyramid files dropped. This PR adds checks for really checking whether the files have been written to disk first, before the (actually to-be-tested) removal of them starts.
For that purpose, a new helper ``wait_for_pyramids`` was introduced in https://github.com/ome/openmicroscopy/pull/6464/commits/e4565f6c3d55dc641245678a78fa55017dcf3d19

# Testing this PR
Jenkins test ``test/integration/clitest/test_pyramids.py::TestRemovePyramidsFullAdmin::test_remove_pyramids_little_endian`` should pass.

# Related reading

The problematic is similar (definitely not fully identical though) to the already opened https://github.com/ome/openmicroscopy/pull/6462. On NFS, the speed of writing and deleting files is simply slower than the tests assume.

ToDo: 

Possibly remove the hardcoded ``pixels_dir = "/home/omero/workspace/OMERO-test-integration/data/Pixels"`` -> I think I would need another round of investigation to find our how to replace with an env variable (cc @jburel ) .

Thanks for any help. Lets see if it works tomorrow during the build.

cc @sbesson @joshmoore  @dominikl

PR #6471, opened 2026-07-27
Rewrite pyramid tests

pwalczysko:
# What this PR does

Rewrites pyramid tests in the OmeroPy suite contained inside  ``pytest test/integration/clitest/test_pyramids.py`` .
As these tests are operating strictly on OMERO-pyramids (under /OMERO/Pixels) and not on pyramids which are part of the imported file, the strategy taken here had to be to skip the tests with clarification, see comment of @sbesson https://github.com/ome/openmicroscopy/pull/6471#issuecomment-5142066579

This PR:
- Reverts the PR #6464 which was merged already and which used polling tactics to force the tests passing
- Skips the pure pyramid tests using the ``@pytest.mark.skipif``
- Adds comments to test classes explaining the character of the tested ``admin removepyramids`` cmd which is operating on OMERO pyramids only
- Renames the tests to ``test_omero_pyramids.py`` to make the situation obvious

# Testing this PR

Check if the tests in the OmeroPy suite contained inside  ``pytest test/integration/clitest/test_pyramids.py`` either pass or are skipped.

# Related reading

Note this is related to https://github.com/ome/openmicroscopy/issues/6470


pwalczysko:
This needs a total rethink. The replacement of the single-plane-fake file import with an import of a fake file with pyramids included has grave consequences on the whole philosophy of this test suite, which is mainly testing the CLI command ``removepyramids`` https://github.com/ome/omero-py/blob/5d0974d097e25d98892b9b419528edf9256274b0/src/omero/util/cleanse.py#L383. This cmd is relying for its basic functioning on pyramid files being present under /OMERO/Pixels. Such omero-pyramids are not present when I import a fake file with pyramids included. My attempts up till now to fix the tests were nonsensical, as one cannot delete "undeletable" pyramids which are a part of the image file itself.

Atm, the only reasonable approach seems to be skipping the failing tests here. 

cc @sbesson

sbesson:
Agreed, this is an example of tests which specifically test and depend on OMERO pyramids. I think skipping them (conditionally to either the configuration or environment) is the best course of action.

An additional clarification idea would be to rename the tests e.g. as `test_omero_pyramids` and add some comments clarifying their scope.

pwalczysko:
> Agreed, this is an example of tests which specifically test and depend on OMERO pyramids. I think skipping them (conditionally to either the configuration or environment) is the best course of action.
> 
> An additional clarification idea would be to rename the tests e.g. as `test_omero_pyramids` and add some comments clarifying their scope.

Thank you @sbesson . Fixes pushed accordingly see the 4 commits above. Also edited the header of this PR to make the situation clear and adding the Include to confirm on Jenkins.

pwalczysko:
The build https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/ is not having any failures from the tests in this PR, all pass or skipped as expected.

Screenshot 2026-08-01 at 21 43 02

sbesson (approved):

PR #6474, opened 2026-08-03
Thumb version adjustments: no omero pyramids

pwalczysko:
# What this PR does

Adjusts the ``OmeroPy/test/Integration/test_thumbs.py`` tests for env with no PixelData service.

# Testing this PR
- Check Jenkins passes. 
- Check sanity of exclusion and clarity of comments.
- Judge if some additional tests should be created to replace ``testThumbnailVersion`` -> such as tests importing an image WITH pyramids and testing a generation and a version of its thumb. 

# Related reading

See https://github.com/ome/openmicroscopy/issues/6470

PR #6475, opened 2026-08-03
Thumb version test adjustments - replaces #6474

pwalczysko:
# What this PR does

- Adjusts the ``OmeroPy/test/Integration/test_thumbs.py`` tests for env with no PixelData service.
- Skips only one test in this file, because it is the only one fully dependent on omero pyramid generation.

# Testing this PR
- Check Jenkins passes. 
- Check sanity of exclusion and clarity of comments.
- Judge if some additional tests should be created to replace ``testThumbnailVersion`` -> such as tests importing an image WITH pyramids and testing a generation and a version of its thumb. 

# Related reading

See https://github.com/ome/openmicroscopy/issues/6470


sbesson (approved):
Tests previously failing due to the lack of pyramid generation are now skipped as expected https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/43/testReport/OmeroPy.test.integration.test_thumbs/TestThumbs/

I think this is fine to merge as it. Additional scenarios could be added to test thumbnail generation for pyramidal file formats but this is outside the scope of this work.

PR #6450, opened 2026-03-25
Add omero-zarr-pixel-buffer

dominikl:
I'd strongly suggest to add the omero-zarr-pixel-buffer to the omero.server build. Like mentioned in https://github.com/glencoesoftware/omero-zarr-pixel-buffer?tab=readme-ov-file#installation .

The current version 0.6.1 only has 5 additional dependencies. But with Zarr v3 support and move to AWS SDK v2 (https://github.com/ome/omero-zarr-pixel-buffer/pull/13 ) there will be 61 (mostly because AWS SDK v2 is highly modularised in contrast to v1). It would also make it much easier to test in CI.

/cc @jburel

PR #6478, opened 2026-08-21
Add omero-zarr-pixel-buffer

dominikl:
Same as #6450 .

Had to close the other one, because it seemed to have some kind of hidden "failed" flag which prevented scc from merging it, even after several all successful rebuilds.

### DIFF
--- a/etc/ivysettings.xml
+++ b/etc/ivysettings.xml
@@ -85,6 +85,10 @@
           usepoms="true" useMavenMetadata="true"
           m2compatible="true"
           root="https://nexus.senbox.net/nexus/content/groups/public/"/>
+      <ibiblio name="gs-artifactory" cache="maven"
+          usepoms="true" useMavenMetadata="true"
+          m2compatible="true"
+          root="https://artifacts.glencoesoftware.com/artifactory/omero-plus/"/>
 
     <!-- Main resolver which has as its first resolver the location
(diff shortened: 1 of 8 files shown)


etc/ivysettings.xml
```diff
@@ -85,6 +85,10 @@
           usepoms="true" useMavenMetadata="true"
           m2compatible="true"
           root="https://nexus.senbox.net/nexus/content/groups/public/"/>
```
sbesson, line 87: We should be able to clean up the `bc.releases` resolver and the associated `chain` resolver below

etc/omero.properties
```diff
@@ -242,6 +242,9 @@ versions.velocity=1.4
 versions.omero-pypi=https://pypi.io/packages/source/o/PACKAGE
 versions.omero-scripts-url=${versions.omero-pypi}
 
+# hard coded, should not be picked up by scc for now!
+versions.omero-zarr-pixel-buffer=0.7.1
```
sbesson, line 246: Should this moved to the section below (alongside `versions.omero-blitz` and `versions.omero-gateway` for consistency?

sbesson (commented):
A few additional cleanup/organization suggestions but overall the proposed changes are consistent with the way Glencoe has been building OMERO.server with omero-zarr-pixel-buffer and all its dependencies included.

